PR #108, opened 2019-09-15
Convert to using Chapel's nilability feature

BryantLam:
**Do not merge this PR if #93 is merged. This PR is mutually exclusive from #93!**
**Do not merge this PR until Chapel 1.20.0 is released.**

**This PR is based on top of branch `ref-from-addEntry` #107 .** I thought I'd be able to redirect the PR to master, but I probably can't (I suppose it's a GitHub constraint). In any case, I can create another PR to merge against master once #107 is merged. I just didn't want the diff of this PR to be polluted with the required changes from #107.

This PR introduces the alternative way of converting the codebase to use Chapel 1.20.0's nilability feature. In a few words, nilability is represented as "can this class be nil" (syntax: `MyClass?`). If it can be nil, the user has to do some extra steps to check that the stored object is not nil before using it.

The primary change is deciding the return value of `SymTab.lookup()`. There's a few other places that nilability affects the code, but this one function decides the bulk of the changes. This function's behavior is "look up the name in the SymTab and either succeed or fail". The failure case can be handled by two approaches. Both approaches are valid and it is a matter of preference for how to handle errors.

- This commit takes the approach of returning a borrowed nilable. The
  borrow is nil if the lookup failed. The caller must check for nil.
  This behavior is similar to what the current code already does.
  The majority of the changes is simply calling the lookup function as
  usual, then converting it to a non-nilable if lookup succeeded (via
  another variable declaration). There are a few points in the code,
  notably GenSymIO.chpl, where the variable declaration cannot be split
  from the variable assignment/initialization because that variable's
  type is non-nilable (thus requiring initialization). So the lines
  of code moved around a bit to avoid the non-nil variable declaration.

- The other approach #93 is to do the nil-check within the `lookup`
  function and return a borrowed non-nilable (and throw if the lookup
  failed). In my opinion, this solution is cleaner to follow and
  requires less explicit handling of the failed lookups because they are
  instead handled as exceptions. The code is then able to assume that
  lookups succeed, and if they instead fail, will just be handled by the
  exception catcher.

Regardless of the decision, the introduction of nilability affects the entire codebase wherever the `lookup` function is called. The difference is simply which way you prefer to handle the failed lookup.

In doing this conversion, I found a few places where the lookup was assumed to always succeed. I preserved this behavior by simply decorating the call with e.g., `st.lookup(...)!` (using postfix `!`).

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -22,7 +22,7 @@ module ArgSortMsg
     use UnorderedCopy;
     use UnorderedAtomics;
 
-    //use Sort only;
+    use Sort only;
     use RadixSortLSD;
     
     // thresholds for different sized sorts
@@ -379,8 +379,9 @@ module ArgSortMsg
         var ivname = st.nextName();
         if v {try! writeln("%s %s : %s %s".format(cmd, name, ivname));try! stdout.flush();}
 
-        var gEnt: borrowed GenSymEntry = st.lookup(name);
-        if (gEnt == nil) {return unknownSymbolError(pn,name);}
+        var gEnt_: borrowed GenSymEntry? = st.lookup(name);
+        if (gEnt_ == nil) {return unknownSymbolError(pn,name);}
+        var gEnt = gEnt_!;
 
         select (gEnt.dtype) {
             when (DType.Int64) {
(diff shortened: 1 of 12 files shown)


mhmerrill:
I think we are going with a version of Brad's PR #152

PR #120, opened 2019-09-21
Fix compilation for Chapel 1.20.0

BryantLam:
I added a compatibility script `chpl-compat.bash` that will generate compiler flags depending on what version of `chpl` you are using. Right now, it only generates `--legacy-classes` for version 1.20 and higher. Once nilability support is merged in, this script should be deactivated because nilability is a breaking change so compatibility with 1.19.0 won't make any sense at that point.


mhmerrill:
I am going to close this PR because we are going to address this with a more current PR #153

PR #151, opened 2019-10-22
Python package

reuster986:
* Refactored the python client into a true package that is installable with `pip` (only installs the python client; I'm not sure whether it's possible to have `pip` build the server, but currently the user has to do it manually.)
* Ran all applicable python-based tests in `tests/`
* Updated documentation with `pip3 install arkouda` usage
* Created a VERSION file, currently with date-based version number
* Consolidated a few old files that were in the top-level dir under `tests/deprecated` and `toys/`
* Uploaded arkouda to the PyPI (again, only the python client installs from PyPI, not the server)

Note to developers: I recommend installing an editable version of the arkouda python client linked to your local copy of the github repo with

```bash
pip3 install -e /path/to/local/arkouda/repo/
```

That way you can make changes to python without having to re-install arkouda every time.


reuster986:
@BryantLam do you know if there is a way to hook `pip` so that it downloads the arkouda source and runs `make`, possibly after creating `Makefile.paths`? Then `pip` would be a one-stop shop for arkouda installation. On the other hand, some users might want to decouple client installation from server build, especially with how long the latter takes. Thoughts?

mhmerrill:
@reuster986 the README.md needs a little more updating...there are a couple of things like where is  `ak_test.py` and how to setup to run it.

reuster986:
@mhmerrill wait, I thought `ak_test.py` was deprecated in favor of `check.py`? Are we still using both?

reuster986:
@mhmerrill I reworked the section of the README.md on testing. It now says to use `tests/check.py`, which I think is more up to date.

BryantLam:
> @BryantLam do you know if there is a way to hook `pip` so that it downloads the arkouda source and runs `make`, possibly after creating `Makefile.paths`? Then `pip` would be a one-stop shop for arkouda installation. On the other hand, some users might want to decouple client installation from server build, especially with how long the latter takes. Thoughts?

I don't personally know how to do this (my Python experience is completely lacking), but I know it is certainly possible.

PR #152, opened 2019-10-23
Convert arkouda to nilable (one option for making Arkouda Chapel 1.20-ready)

bradcray:
This is a first stab at converting Arkouda over to using nilable
vs. non-nilable classes.  Generally speaking, the approach I took was
to use non-nilable classes for the most part and to rely on throwing
and catching errors to handle failed lookups or conversions.  This
had the upside of living primarily in the non-nilable world of features,
but the downside of requiring catches to be put throughout the
code (and, being a bit paranoid, I always did this with a very specific
catch block for the error I anticipated and a catch-all "just in case"
which made things messier...  I'm definitely open to different ideas
here).

Note that I didn't worry much about getting indenting "right" in order to
try and minimize the diff against master for the sake of comparison by
human reviewers of this PR. If we go forward with this PR, I can update the
indentation either before or after merging.

At this point, I'm opening this PR to get feedback on the approach
taken and whether the arkouda keepers would like to see a different
approach or any other minor changes.

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -422,18 +422,24 @@ module ArgSortMsg
       var size: int;
       // Check that all arrays exist in the symbol table and have the same size
       for (name, i) in zip(names, 1..) {
+        try {
 	// arrays[i] = st.lookup(name): borrowed GenSymEntry;
 	var g: borrowed GenSymEntry = st.lookup(name);
-	if (g == nil) { return unknownSymbolError(pn, name); }
 	if (i == 1) {
 	  size = g.size;
 	} else {
 	  if (g.size != size) { return incompatibleArgumentsError(pn, "Arrays must all be same size"); }
 	}
+        } catch (e: UndefinedSymbolError) {
+          return unknownSymbolError(pn, name);
+        } catch {
+          return unknownError(pn);
+        }
       }
       // Initialize the permutation vector in the symbol table with the identity perm
       var rname = st.nextName();
       st.addEntry(rname, size, int);
+      try {
       var iv = toSymEntry(st.lookup(rname), int);
       iv.a = 0..#size;
       // Starting with the last array, incrementally permute the IV by sorting each array
@@ -449,6 +455,12 @@ module ArgSortMsg
 	  return try! "Error: %s unknown cause".format(pn);
 	}
       }
+      } catch e: UndefinedSymbolError {
+        return unknownSymbolError(pn, rname);
+      } catch {
+        return unknownError(pn);
+      }
+
       return try! "created " + st.attrib(rname);
     }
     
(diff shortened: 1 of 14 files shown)


bradcray:
@mhmerrill: OK, I think this PR should now "work" with the changes to privacy settings.

mhmerrill:
@reuster986 @bradcray @BryantLam I think we should go with an approach much like this with a few differences.
Things to consider:
 * look at wrapping the top level select in the main of `arkouda_server.chpl` in a `try {} catch {}` block
 * how so we get the `proc fooMsg` name into the error which is `throw`n, do we have to `catch` and re-`throw` it?
 * define `Error` objects for all the error strings? or just one server error which holds any string?
 * other ideas?

bradcray:
> look at wrapping the top level select in the main of arkouda_server.chpl in a try {} catch {} block

On my way home last week, I was also wondering about pushing the `try...catch` blocks further up in the call chain to avoid making the code so ugly and fragmented.    The main challenge, as your second bullet above alludes to:

> how do we get the proc fooMsg name into the error which is thrown, do we have to catch and re-throw it?

is losing the procedure name that is currently being embedded into the unknownSymbolError using patterns like this:

>           return unknownSymbolError(pn, name);                                  

My main thoughts here are:

* we could have calls to `lookup()` (optionally?) pass the procedure name in as an argument so that the error can embed the procedure name until it's converted into a string at the outermost select.  
* we could decide that the procedure name implementing the lookup() isn't all that interesting to an end-user and not worry about preserving it?  (i.e., today it's something like `argsortMsg` which arguably isn't all that user-friendly or identifiable anyway?).
* we could implement something like https://github.com/chapel-lang/chapel/issues/14343 such that lookup() could know the calling procedure name without it needing to be passed in.  But I worry about the cost of a (general-purpose) implementation for #14343 (i.e., one that isn't limited to just the parent scope).

BryantLam:
> we could decide that the procedure name implementing the lookup() isn't all that interesting to an end-user and not worry about preserving it

This is the approach I would go with due to simplicity. In most cases, it doesn't matter where the lookup failed as long as it's reported back. The server's main loop already knows which function is called when the lookup failed, so the handling could be done there if it truly mattered.

The only case where this arguably matters is if the frontend or the server somehow sent a wrong identifier into the lookup function. If this happens, a throw would occur but there's no traceability (today) as to *how* the the wrong identifier was generated. I don't think this is important enough to worry about dealing with, especially since we can instrument the main loop later as mentioned above.

bradcray:
Bryant's comment reflects my thinking, but I wanted to keep the door open to other options (and am arguably not as invested in arkouda as Mike and Bill, so may not be the best judge).  Since it seems like the procedure name in failed lookups is most useful to a developer, it seems like a more general tracing mechanism may be the way to make such information available to developers when it isn't trivially obvious which lookup failed (my guess is that for most operations, given the variable names, it should be).  I'm open to the other approaches (or something I haven't thought of) as well, though.

bradcray:
Closing in favor of #165.

PR #153, opened 2019-10-24
Add building arkouda server to pip install

Spartee:
#### Description

This PR adds methods to build the ``Arkouda`` server when performing a pip install of ``Arkouda``. It works by overriding the setuptools.command.install class with a ``InstallArkouda`` class. In the ``InstallArkouda`` class the run method is overridden so that the chapel and arkouda builds are called during installation. 

#### TODO 
 - [x] Install on mac
 - [x] Recognize existing chapel installation if present
 - [x] Recognize environment/modules and compile on Cray
 - [x]  push to TestPyPi and pip install from there


Spartee:
The latest commit adds ``installers.py`` which makes the ``ArkoudaInstaller`` recognize an existing chapel installation. I have now tested this on an XC and on Mac. A few things to note when installing through pip

### On XC/Cray
 - Before call to ``pip install arkouda`` users need to
   - source existing chapel and/or load chapel module
   - ``module load cray-hdf5-parallel``
   - put zmq into ``LD_LIBRARY_PATH`` see notes below on this

### Putting ZMQ into LD_LIBRARY_PATH
  - once Chapel and Mason are loaded, call ``mason system pc libzmq`` which will provide the path to the zmq ``lib`` dir
  - Then just ``export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:zmq/lib/dir/path``

### On Mac
  - if the user has there own Chapel, we currently do not reconfigure and re-make it for them. We expect any user that already has Chapel to build and configure it correctly for arkouda prior to pip install
  - brew installed Chapel will not build arkouda, it is easiest to just let arkouda build its own chapel version, configure, and build that to build arkouda

Spartee:
Still open question: How do we want the user to have access to ``arkouda_server``? symlink to ``usr/local/bin``? what about XC? open a user facing python method?

reuster986:
@Spartee great work! I didn't know you could do all this with setuptools!

I would like to look over the code and test it on linux -- hopefully I will be able to do it this weekend, but I might not have time until Tuesday. In the meantime, I'd like to ask @BryantLam about whether he forsees any issues with using `LD_LIBRARY_PATH`, as opposed to the current approach of `--rpath`.

On the question about where to put `arkouda_server`, would it make sense to put it in the `bin` directory of the python installation? I have done this in packages with a pre-existing executable, but I'm not sure how you'd do it with an executable built at installation time. Just an idea.

Thanks for contributing!

Spartee:
@reuster986 That is one option. We could build it in that path and then it would be available in the ``PYTHONPATH``. We'd just build it and move it.

reuster986:
@Spartee I'm testing on linux now, and I'm having trouble finding where the compiled `arkouda_server` executable ends up. 

Here's what I did

```bash
git clone https://github.com/Spartee/arkouda.git Spartee/arkouda
cd Spartee/arkouda
git checkout build_arkouda_server
cd ~
pip uninstall arkouda
pip install Spartee/arkouda
```

The last command looked like it was building `arkouda_server` -- it paused for several minutes, during which I could see `chpl` running at 100% in top -- but now I can't find `arkouda_server` in any of the obvious places: ~, Spartee/arkouda, Spartee/arkouda/build, Spartee/arkouda/dist, etc. Any ideas?

Spartee:
@reuster986 So I was mistakenly testing the install using ``python setup.py install`` to test as I believed that was all that pip called upon setup. However, that actually seems to be the fallback to ``python setup.py bdist_wheel`` according to [This thread](https://discuss.python.org/t/pep-517-and-projects-that-cant-install-via-wheels/791/38). They are also talking about deprecating the fallback which would be bad for us as the fallback is what we rely on in the override to build Arkouda. I need to do more research to see if we can override the ``bdist_wheel`` command as well. 

We may have to rely on the user having Chapel already installed and create Arkouda through the process of building a wheel. H5py does a similar thing [here](https://github.com/h5py/h5py/blob/master/setup_build.py)

reuster986:
@Spartee I think it's fine to rely on the user installing and configuring chapel themselves (much like pip assumes the user has installed gcc for some things), especially since the README and docs describe how to build chapel.

Spartee:
Currently testing out a new branch that relies on the user having Chapel installed already. Installs the arkouda server into the prefix of the current python interpreter. Pushed to TestPypi for others to test out. [Test Package](https://test.pypi.org/project/arkouda-test-spartee/2019.10.22/)

Steps to install:
 - build and source Chapel
 - (optionally) create new virtual environment
 - `` pip install -i https://test.pypi.org/simple/ arkouda-test-spartee==2019.10.22``
 - should be able to call ``arkouda_server``

reuster986:
@Spartee this looks good! Two requests:
* Is there a way for `make_arkouda_server()` to detect failure in the arkouda build process and consequently make the `pip install` fail? When I did the `pip install`, I accidentally had the wrong environment variables for including ZMQ, so the arkouda build failed silently behind the scenes, but the `pip install` succeeded, and it was hard to tell what had happened (I ended up passing three `-v` flags to `pip` to trace everything). I think we should make it so that if the arkouda build fails, the pip install fails and prints the errors from the build.
* Can you change the binary install path from `sys.prefix` to the `bin` subdirectory of `sys.prefix`?

Spartee:
@reuster986 Done! the latest should work as you described. The most recent updates have been pushed to testPypi and can be found [here](https://test.pypi.org/project/arkouda-test-spartee/2019.11.2/). Besides stylistic comments I think this should be ready to go. 


Updated notes:

### General
 - We rely on the user having configured and built or module loaded chapel correctly
 - ``arkouda_server`` is built within the bin directory of ``sys.prefix``


### On XC/Cray
 - Before call to ``pip install arkouda`` users need to
   - (optionally but suggested) start a new virtual environment
   - source existing chapel and/or load chapel module
   - ``module load cray-hdf5-parallel``
   - put zmq into ``LD_LIBRARY_PATH`` see notes below on this


### Putting ZMQ into LD_LIBRARY_PATH
  - once Chapel and Mason are loaded, call ``mason system pc libzmq`` which will provide the path to the zmq ``lib`` dir
  - Then just ``export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:zmq/lib/dir/path``

### On Mac
  - brew installed Chapel will not build arkouda

reuster986:
@Spartee I like the error handling. @BryantLam for some reason, I can't get linux to recognize my `LD_RUN_PATH`, so even when I have it set correctly, `chpl` can't find ZMQ and HDF5 without a `Makefile.paths` (which is not an option with `pip install`). Do either of you have any ideas on how we can fix this? I think it would suffice to pass all paths in `LD_LIBRARY_PATH` as arguments to `-rpath`, but I don't know how to do that in the `Makefile`. Any help appreciated.

PR #166, opened 2019-10-30
Change get hostname technique from spawn() to locale query

### LINKED ISSUE #164: Make getting the host name more elegant
The following Arkouda code is an admittedly clunky way of getting the hostname of locale 0:

https://github.com/mhmerrill/arkouda/blob/a99a38e525f591615043c5e0db184d5a1943cfb2/src/ServerConfig.chpl#L41-L54

Some ways to improve this:

* read directly from the spawned command's stdout?
* call out to an extern C routine that returns the hostname?
* introduce a Chapel library routine that returns a node's hostname?

Note that I was originally thinking that `Locales[0].name` was the way to do this, but I don't think that would give a sufficiently fully qualified name for the client to be able to connect to the server as Arkouda is currently written.

bradcray:
This converts the spawn-based approach used to get the hostname
of the current locale into a ~query on the locale value, which is far
more Chapeltastic~.  Elliot points out that `locale.name` is not always
a valid hostname (specifically when using local spawning with GASNet
in some configurations), so I've changed this PR to use a workaround
as well.  Separately, I have a change intended for 1.21 in which locale
supports a `.hostname` query, so we can change to that at that time.

Even though this workaround uses a non-public interface, I think it
remains preferable to the previous approach.

Resolves #164.

### DIFF
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -29,7 +29,15 @@ module ServerConfig
     var serverHostname: string = try! get_hostname();
 
     proc get_hostname(): string {
-      return here.name;
+      // Want:
+      //   return here.hostname;
+      // but this isn't implemented yet; could use:
+      //   return here.name;
+      // but this munges the hostname when using local spawning with GASNet
+      // so the following is used as a temporary workaround:
+      extern proc chpl_nodeName(): c_string;
+      var hostname = chpl_nodeName(): string;
+      return hostname;
     }
 
     proc getConfig(): string {
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -29,30 +29,7 @@ module ServerConfig
     var serverHostname: string = try! get_hostname();
 
     proc get_hostname(): string {
-        /* The right way to do this is by reading the hostname from stdout, but that 
-           causes a segfault in a multilocale setting. So we have to use a temp file, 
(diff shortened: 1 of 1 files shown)


bradcray:
I was initially concerned that this would not yield a sufficiently qualified hostname, but my team has argued that I'm wrong (and it seems to work for me on my Mac).  @mhmerrill and @reuster986: You should probably test this on key platforms before merging to make sure we're correct.

ronawho:
@bradcray sorry, I led you astray. I forgot that for gasnet-smp and gasnet-udp with local spawning we append a unique id to locale.name [here](https://github.com/chapel-lang/chapel/blob/54c5cdb4884d97df8c172d486ef209e35a390543/modules/internal/LocaleModelHelpSetup.chpl#L106-L131). 

I'd probably do something like:

```chpl
extern proc chpl_nodeName(): c_string;
var name = chpl_nodeName():string;
return name;
```

bradcray:
> I forgot that for gasnet-smp and gasnet-udp with local spawning we append a unique id to locale.name

Hmm, that's a bummer (though also understandable).  What would you think about our adding a `hostname` query to the `locale` type that would give an unmodified hostname?  (wouldn't help with this code for now, but down the line...).

ronawho:
I'd be in favor of that.

bradcray:
I've updated the PR to use your suggested workaround Elliot, and have a simple change that adds `.hostname` to the `locale` type.

PR #167, opened 2019-10-30
Have get_config() report details on each of the locales

### LINKED ISSUE #163: Extend getConfigName to support all locales
As I understand it, the original intention of getConfig() was to collect a certain amount of configuration information from each of the locales, not just the 0th.  However, when this was first tried, the conversion to json failed.  The goal of this task would be to have the uncommented version of the following lines work s.t. the configuration of the entire set of locales can be captured:

https://github.com/mhmerrill/arkouda/blob/a99a38e525f591615043c5e0db184d5a1943cfb2/src/ServerConfig.chpl#L86-L94

bradcray:
Some variation on this apparently caused problems before, but
it doesn't seem to be now, at least as I've conceived of it.

Resolves #163

### DIFF
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -73,9 +73,11 @@ module ServerConfig
             var numPUs: int;
             var maxTaskPar: int;
             var physicalMemory: int;
+            var LocaleConfigs: [LocaleSpace] LocaleConfig =
+                        [loc in LocaleSpace] new owned LocaleConfig();
         }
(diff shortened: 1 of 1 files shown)


bradcray:
@mhmerrill: This seems to work for me, at least on one locale.  It could be that we've fixed a bug, or it could be that I haven't exactly reproduced what you were trying to do that was causing an error.  Want to take a look and see if this does what you were hoping, or not?

PR #171, opened 2019-10-30
Add support for src/TAGS for Chapel source

bradcray:
This adds a `make tags` rule to the top-level Arkouda Makefile
which attempts to run $CHPL_HOME/util/chpltags on the src/
directory in order to create a TAGS file for use with emacs
and/or vim.  This makes it much easier, when editing Arkouda
source, to jump to and find the definitions of symbols without
relying on grep or foreknowledge of the code organization.

If it were me, I'd probably have this command fire automatically
anytime the general `make` rule was executed to ensure that
the TAGS file was always up-to-date, but doing so in this PR
felt presumptuous on my part, and it'd probably make sense to
make sure this works for people other than me before taking
that step.

Open question: Is the path taken to find `chpltags` reasonable,
or should we assume the user has it in their path, or search for
it in a few locations, or ...?


bradcray:
@mhmerrill, @reuster986: For your consideration.  I've been meaning to implement this for weeks and have already profited from it enormously in the hour since I implemented it.

BryantLam:
@bradcray -- I pushed some changes to your branch and tested it. Tags support is neat.

Makefile
```diff
@@ -75,10 +78,16 @@ ARKOUDA_MAIN_SOURCE := $(ARKOUDA_SOURCE_DIR)/$(ARKOUDA_MAIN_MODULE).chpl
 $(ARKOUDA_MAIN_MODULE): $(ARKOUDA_SOURCES) $(ARKOUDA_MAKEFILES)
 	$(CHPL) $(CHPL_DEBUG_FLAGS) $(CHPL_FLAGS) $(ARKOUDA_MAIN_SOURCE) -o $@
 
-CLEAN_TARGETS += arkouda-clean
-.PHONY: arkouda-clean
-arkouda-clean:
-	$(RM) $(ARKOUDA_MAIN_MODULE) $(ARKOUDA_MAIN_MODULE)_real
```
bradcray: I'm probably missing something obvious, but I don't understand why these lines went away.

bradcray: Ah, I reviewed too quickly... you've restored them now.

PR #177, opened 2019-11-03
Fix tests

mhmerrill:
changed a few things in the tests to make python codes more uniform.
 * made all tests call `ak.shutdown()`
 * made all python tests use `#!/usr/bin/env python3`


mhmerrill:
@reuster986 I will make them all disconnect, I just wanted them to be consistent.

PR #183, opened 2019-11-07
Try to fix setting PYTHONPATH for travis testing

ronawho:


mhmerrill:
@ronawho you might want to try something like:
```bash
> pip3 install /path/to/arkouda/repo
```
this will put the arkouda package into the python env and you won't have to set `PYTHONPATH`

ronawho:
I think the original issue was just that something new needed to be in `PYTHONPATH` since I originally added Travis testing. I swapped to the `pip install .` route (which is really nice.) 

This led to new problems because Arkouda now requires Python 3.6, but the Chapel docker image is based on debian:9.5, which can only get python 3.5.

This PR currently throws `--ignore-requires-python` to ignore the requirement and use Python 3.5 This seems to "work", but feels wrong. I think the current options are:
 - Upgrade the Chapel docker image base to debian:10 (I'm not sure what is involved in this.)
 - Drop Arkouda to only require python 3.5 (I'm assuming this is not an option since presumably there's a good reason to require 3.6.)
 - Use the current `--ignore-requires-python` until the next chapel release, and make sure that release uses a newer debian base.

Issue #162, opened 2019-10-29
Sort performance thoughts


ronawho:
Misc thoughts and suggestions for improving the performance of sort. 

The current sort implementation is limited by the speed of unorderedCopy. A majority of the time is spent putting elements into their correct position during the "calc new positions and permute" step, and if there are negative values reordering at the end.

Here's the microbenchmark I've been using to evaluate performance. It measures how fast raw unorderedCopy is and how fast sort for keys/ranks is. By default it operates on arrays that use 1/50 of memory and generates random 64-bit values. It has options to change the size, max numbers of bits used, and if only positive numbers should be used.

<details>

```chpl
use BlockDist;
use Memory;
use Random;
use Time;
use RadixSortLSD;
use UnorderedCopy;

config type elemType = int;
param elemSize = numBytes(elemType);

const totMem = here.physicalMemory(unit = MemUnits.Bytes);
config const memFraction = 50;
config const nElems = totMem / elemSize / memFraction * numLocales;

var t: Timer;
inline proc startDiag() {
  t.start();
}
inline proc endDiag(name) {
  t.stop();
  //writeln(name, " time : ", t.elapsed());
  const mbPerNode = (nElems * elemSize):real / (1024.0*1024.0) / numLocales:real;
  writef("%20s MB/s per node: %.2dr\n", name, mbPerNode/t.elapsed());
  t.clear();
}

config const positiveValues = false;
config const maxBits = numBits(elemType);
const modV = ((1 << maxBits) - 1): elemType;

proc main() {
  var D = newBlockDom({0..#nElems});
  var A, revA: [D] elemType;

  fillRandom(A, seed=314159265);
  if positiveValues then A = abs(A);
  if isIntegral(elemType) && maxBits < numBits(elemType) then A %= modV;

  startDiag();
  forall i in D do unorderedCopy(revA[nElems-1-i], A[i]);
  endDiag("unorderedCopy");

  startDiag();
  var B = radixSortLSD_keys(A);
  endDiag("radixSortLSD_keys");
  assert(isSorted(B));

  startDiag();
  var C = radixSortLSD_ranks(A);
  endDiag("radixSortLSD_ranks");
}
```

</details>

When sorting positive 16-bit values performance for radixSortLSD_keys  is pretty close to raw unorderedCopy:

```
./sort-perf -nl 16 --positiveValues=true --maxBits=16 --memFraction=50
       unorderedCopy MB/s per node: 481.49
[...]

PR #185, opened 2019-11-11
Optimize sorting for non-cray networks

ronawho:
The current sorting code uses `unorderedCopy` for small messages. This
idiom is only optimized for Cray networks (comm=ugni) and is very slow
on other platforms.

This PR adds initial support for aggregating copies (`dst=src`) and uses
it for sorting on non-ugni configurations. There is a lot more that can
be done to optimize aggregation, and longer-term aggregation will
outperform unorderedCopy under ugni as well. For now, I just wanted to
get some initial aggregation committed since it has such a large benefit
for non-cray networks.

Here are some sort performance results for an FDR InfiniBand network
with gn-ibv-large in MB/s per node. This is the benchmark from #162 with
`./sort-perf --positiveValues=true --maxBits=16 --memFraction=50`

| nodes | unordered   | aggregated |
| ----- | ----------- | ---------- |
|   1   | 520.00 MB/s | 565 MB/s   |
|   2   |   1.65 MB/s | 420 MB/s   |
|   4   |   0.36 MB/s | 350 MB/s   |
|   8   |   0.34 MB/s | 320 MB/s   |
|  16   |   0.49 MB/s | 300 MB/s   |
|  32   |   0.53 MB/s | 250 MB/s   |

Performance is also significantly improved for Ethernet networks. I
don't have a real Ethernet cluster to run on, but I was able to use the
out-of-band Ethernet on that same IB cluster. At 16 nodes, performance
goes from 0.01 MB/s to 8 MB/s. This is not a production network and I
would expect results to be much better on a real Ethernet cluster.

There is a non-trivial memory overhead for aggregation. Each task has
per-destination buffers and uses `bufferSize * numLocales * elemType`
space. Sorting `int`s on 32 nodes with 36 cores per node this requires
~128 MB per node for aggregation buffers. At 512 nodes this would
require ~2 GB per node.

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -1,0 +1,158 @@
+module CommAggregation {
+
+  // TODO these parameters need to be tuned and size should be user-settable at
+  // creation time. iters before yield should be based on numLocales & buffSize
+  private config const maxItersBeforeYield = 4096;
+  private config const dstBuffSize = 4096;
+  private config const srcBuffSize = 4096;
+
+  /*
+   * Aggregates copy(ref dst, src). Optimized for when src is local.
+   * Not parallel safe and is expected to be created on a per-task basis
+   * High memory usage since there are per-destination buffers
+   */
+  record DstAggregator {
+    type elemType;
+    const bufferSize = dstBuffSize;
+    const myLocaleSpace = LocaleSpace;
+    var itersSinceYield: int;
+    var buffer: [myLocaleSpace][0..#bufferSize] (c_void_ptr, elemType);
+    var bufferIdxs: [myLocaleSpace] int;
+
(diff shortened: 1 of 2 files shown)


ronawho:
This will probably be easiest to review ignoring whitespace -- https://github.com/mhmerrill/arkouda/pull/185/files?w=1

This is a cleaned-up version of the prototype from https://github.com/mhmerrill/arkouda/issues/162#issuecomment-549540542. I dropped the default buffer size from 32,768 to 4,096. This hurts performance but lessons the memory overhead a lot.

Over time I plan to tune more and limit memory overheads, but I think this is a good start. If you want to experiment on your IB clusters, the important knob to tune is `dstBuffSize`.

Issue #184, opened 2019-11-07
fix division and modulus operators to behave just like numpy with respect to divide by zero


mhmerrill:
1. Currently the `operator_test.py` catches 18 errors where arkouda `pdarrays` don't behave like numpy `ndarrays`.
1. These all seem to involve `//`,`/`, and `%` operators and dividing by zero. 
1. Arkouda currently throws a runtime exception on divide by zero which numpy doesn't seem to do.
1. In numpy it appears that you can control this behavior with `numpy.seterr(divide='ignore')` which seems to be the default (default return is `inf`).

mhmerrill:
@bradcray @mppf @ronawho I would like to get the Chapel code to behave like numpy in the divide-by-zero case. I think the default for numpy is to return `inf` (infinite).

mhmerrill:
truediv
np.int64(0) / np.int64(0) = nan
np.int64(1) / np.int64(0) = inf
floordiv
np.int64(0) // np.int64(0) = 0
np.int64(1) // np.int64(0) = 0

bradcray:
@mhmerrill: For the truediv cases, I think what you're seeing is consistent with (my understanding) of IEEE floating point given that Python defines division between two integers as being a float. In contrast, Chapel preserves the integer nature, as in C and as discussed a bit on https://github.com/chapel-lang/chapel/issues/14306

[This TIO session](https://tio.run/##dY5BCsIwEEX3OcXQldk4ybbBw4SS2oHQkbEo5PKRVMQk1uW8/wbetPhbiDlPvM50hYcXSAQXME7VSAo6t5CLZ1tUPFs89RTaQlxPg0EzDtopxA9KhIm0A0CExcftrr7LKMHHfVaNvvMOSn1L9yTtrusk@5PEf5P4KImPkrhL4i6J30k5vwA) suggests to me that for cases involving floats in Chapel (including casting an int argument to real), the results are the same as for truediv.  To make Arkouda and NumPy / Python behave similarly in this case, would it be sufficient to have the translation of the operation to Chapel introduce the cast to real for int, int division?

I'm less familiar with floordiv and would need to dig into it further to offer an opinion.

mhmerrill:
PR #187 fixes this issue ;-)

PR #187, opened 2019-11-13
fixed divide and modulus by zero to match numpy

mhmerrill:
fixed all operators to match numpy except scalar//vector (floordiv)
there appears to be a discrepancy between numpy floordiv for vector//vector and scalar//vector.
numpy returns `inf` for `[non-zero]//[0]` and `nan` for `scalar//[0]`

### DIFF
--- a/src/OperatorMsg.chpl
+++ b/src/OperatorMsg.chpl
@@ -162,7 +162,10 @@ module OperatorMsg
                     } 
                     when "//" { // floordiv
                         var e = st.addEntry(rname, l.size, real);
-                        e.a = floor((l.a / r.a));
+                        ref ea = e.a;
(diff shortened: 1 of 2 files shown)


mhmerrill:
need another commit to get floordiv to produce matching values for zero divide. numpy outputs `nan` for all cases of zero division both scalar and vector, unlike truediv which produced `inf` for 1/0 and `nan` for 0/0

bradcray:
Pro tip if you're not aware of it:  In creating a PR if you put "Resolves #184" into the message (or a couple of other similar phrasings), merging the PR will automatically close the issue (and GitHub also does some special parsing/hyperlinking of the reference as well I think, though I don't use these).

PR #188, opened 2019-11-14
Another `private use` by default update

bradcray:
Chapel is moving toward having `use` statements be private (i.e.,
non-transitive) by default.  This is a case that needs a `use` which I
missed in my previous PR #173 (or perhaps the world changed since
then).


bradcray:
For your consideration as a candidate "100th merged PR" :)

PR #189, opened 2019-11-14
Add `use IO` to get stdout, stderr, etc.

bradcray:
Chapel is teasing IO into two categories:

* things that are supported by default (e.g., default, channel-less
  writeln(), writef()), and

* things that require `use IO` to get at (channels, files, errors,
  etc.)

This adds `use IO` in a few key places to keep Arkouda working under
these assumptions for consideration.


bradcray:
@lydia-duncan has an alternative that adds `use IO` to each file if that style would be preferred.  Here, I relied on adding `use IO` to ServerConfig.chpl to get most files in one fell swoop.

PR #190, opened 2019-11-14
added --ccflags="-Wno-incompatible-pointer-types" to makefile

mhmerrill:
incompatible-pointer-type warning started showing up again with HDF5 on our IB cluster so I added it back to the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ CHPL_DEBUG_FLAGS += --print-passes
 ifndef ARKOUDA_DEVELOPER
 CHPL_FLAGS += --fast
 endif
+CHPL_FLAGS += --ccflags="-Wno-incompatible-pointer-types"
 CHPL_FLAGS += -lhdf5 -lhdf5_hl -lzmq
 
 # --cache-remote does not work with ugni in Chapel 1.20
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ CHPL_DEBUG_FLAGS += --print-passes
 ifndef ARKOUDA_DEVELOPER
 CHPL_FLAGS += --fast
 endif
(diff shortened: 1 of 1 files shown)


bradcray:
Bummer!  Want to add a comment to the line adding the flag so that next time we remember why it was added?  (though I suppose we could also look back at this PR now).  I'd be curious if you could post an excerpt of what the error message looks like.  I'm guessing that it's similar to the following longstanding issue on the Chapel repo: https://github.com/chapel-lang/chapel/issues/7983

mhmerrill:
> Bummer! Want to add a comment to the line adding the flag so that next time we remember why it was added? (though I suppose we could also look back at this PR now). I'd be curious if you could post an excerpt of what the error message looks like. I'm guessing that it's similar to the following longstanding issue on the Chapel repo: [chapel-lang/chapel#7983](https://github.com/chapel-lang/chapel/issues/7983)

Yes this is what the warning looks like.

ronawho:
The warning shows up in the Travis runs too -- https://travis-ci.com/mhmerrill/arkouda/builds/136442838

If you want to do this for gnu only you can do something like:

```make
TARGET_COMP = $(shell $(CHPL_HOME)/util/chplenv/chpl_compiler.py --target 2>/dev/null)
ifneq (, $(findstring gnu,$(TARGET_COMP)))
CHPL_FLAGS += --ccflags="-Wno-incompatible-pointer-types"
endif
```

ronawho:
Just FYI that we made a similar change upstream in https://github.com/chapel-lang/chapel/pull/14446

Also fyi that this will fail on non gnu/clang compilers (I use the intel backend semi-frequently, but this may not be an important use-case for you)

PR #196, opened 2019-11-19
Mem usage

mhmerrill:
This PR is an attempt to predict and prevent the server from running out of memory.
It may not be perfect and catch all memory allocations which may make the server run out of memory but it should prevent most of the them.

All big memory users like argsort, sort, concatenate, unique, etc.


bradcray:
@mhmerrill / @reuster986 : Note that in order for Arkouda to work with recent versions of Chapel master, the `use IO;` statements that were commented out in this PR will have to be re-enabled.  My understanding is that they were likely commented out while chasing down the bug related to first-class functions (https://github.com/chapel-lang/chapel/issues/14467), but were ultimately unrelated to the core issue.  In addition, some other minor tweaks to `use IO` statements are necessary.  I'll open a new PR soon with these changes, and will ensure they are backward-compatible with 1.20 as well.

Issue #199, opened 2019-11-27
Increase log file precision on function time stamp and function delta 


jwcrandall:
Increase log file precision on function time stamp and function delta that gets printed to an arkouda log file.

mhmerrill:
fixed this with PR #200

PR #200, opened 2019-11-27
increased precision of timestamps

mhmerrill:
increased the precision of timestamps in the arkouda_server log/trace
this fixes issue #199

### DIFF
--- a/src/arkouda_server.chpl
+++ b/src/arkouda_server.chpl
@@ -67,7 +67,7 @@ proc main() {
             else {
                 writeln("reqMsg: ", reqMsg);
             }
-            writeln(">>> ",cmd," started at ",t1,"sec");
+            writeln(">>> %s started at %.17r sec".format(cmd, t1));
             try! stdout.flush();
         }
 
@@ -159,7 +159,7 @@ proc main() {
         if (memTrack) {writeln("bytes of memoryUsed() = ",memoryUsed()); try! stdout.flush();}
 
         // end timer for command processing
-        if v{writeln("<<< ", cmd," took ", Time.getCurrentTime() - t1,"sec"); try! stdout.flush();}
+        if v{writeln("<<< %s took %.17r sec".format(cmd, Time.getCurrentTime() - t1)); try! stdout.flush();}
     }
 
     writeln("requests = ",reqCount," responseCount = ",repCount);
--- a/src/arkouda_server.chpl
+++ b/src/arkouda_server.chpl
@@ -35,12 +35,12 @@ proc main() {
 
     var reqCount: int = 0;
     var repCount: int = 0;
+    var t1 = new Timer();
+    t1.clear();
+    t1.start();
     while !shutdownServer {
         // receive requests
         var reqMsg = socket.recv(string);
-
-        // start timer for command processing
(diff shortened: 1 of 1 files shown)


src/arkouda_server.chpl
```diff
@@ -67,7 +67,7 @@ proc main() {
             else {
                 writeln("reqMsg: ", reqMsg);
             }
-            writeln(">>> ",cmd," started at ",t1,"sec");
+            writeln(">>> %s started at %.17r sec".format(cmd, t1));
```
BryantLam: The format specifier `%.17r` will work but says "give me 17 significant digits". It might be easier to use the [generic number formats](https://chapel-lang.org/docs/master/modules/standard/IO/FormattedIO.html?highlight=print%20specifier#generic-numeric-conversions) if you only want so many digits after the decimal point.

For example, using format specifier `%{.###}` gives you three digits after the point. [See example in TIO.](https://tio.run/##hcw9CoNAFATg3lMMSsA0K5LCQlLlCrnAos/1ga6y@4xFyNk3/jQhIVjOMPNVrR6pC2HyhDv3VEbRQzvYYcZ1K5QhuU3OkZU1pucymh0LdTZdRh8plpZQ7UvIMgV7nFReuFg1g@u1bIdzCSDLkBfwbBo2HjKI7o6d@hfanZoNi4duhBzW9ziwlQPwqZIkeX2LC4jLHzCENw)

For info, the `Time` module calls `chpl_now_time()` -> `gettimeofday` which has up-to-microsecond resolution.

src/arkouda_server.chpl
```diff
@@ -159,7 +159,7 @@ proc main() {
         if (memTrack) {writeln("bytes of memoryUsed() = ",memoryUsed()); try! stdout.flush();}
 
         // end timer for command processing
-        if v{writeln("<<< ", cmd," took ", Time.getCurrentTime() - t1,"sec"); try! stdout.flush();}
+        if v{writeln("<<< %s took %.17r sec".format(cmd, Time.getCurrentTime() - t1)); try! stdout.flush();}
```
BryantLam: The `Time.getCurrentTime()` function returns the number of elapsed (default) seconds since midnight local time, not Unix epoch.

In my opinion, this function is not generally useful; actually it is deceiving and more of an anti-pattern such as in this usage here. Subtracting between two `Time.getCurrentTime()` timepoints will roll over if you time something across midnight and you'll get negative elapsed time. You should use the `Timer` class provided by the `Time` module instead of this function.

src/arkouda_server.chpl
mhmerrill: I wanted to allow for all the precision of floating point number from the IEEE 754 spec which states that 17 digits of precision captures all the precision of a 64 bit floating point number.

BryantLam: Ah fair enough. Good point.

src/arkouda_server.chpl
mhmerrill: ill make the change for the main timer

mhmerrill: @BryantLam change is made and pushed.

Issue #201, opened 2019-12-02
remove NotImplemented


reuster986:
A couple places (e.g.  [here](https://github.com/mhmerrill/arkouda/blob/e1df53466f5948726103e112f9a77392f6091d53/arkouda/pdarrayclass.py#L111)) in the python client return `NotImplemented`, which is a python pass-through that allows implementers to integrate new classes with existing python classes. However, we should not be using it, because nothing else knows about the `pdarray` class. Instead, we should raise a `ValueError` like numpy.

mhmerrill:
fixed with PR #203

PR #203, opened 2019-12-04
Fix pda operator not implemented fixes issue #201

mhmerrill:
Fixes issue #201 
I made the changes to fix the bad behavior when operators aren't implemented in the arkouda server. Instead of returning `NotImplemented` we are now raising an error with `raise TypeError`

This is difficult to test. I ended up starting up a python3 interpreter and causing the exceptions manually, we should probably devise a way to test the exceptional cases for the arkouda client and server.


mhmerrill:
> looks good.
@reuster986 is this good enough to merge?

reuster986:
I approved it. I think maybe it's waiting on @BryantLam ?

reuster986:
Ok, that time worked for some reason...

PR #207, opened 2019-12-06
Add a sorting test and fix some bugs

ronawho:
This adds a pretty carefully crafted test that has been really useful in
some work I've been doing to optimize sorting. There are 3 main
correctness variants:
 - Sort int/real arrays where the initial values are permuted indices
   There are variants that reverse the indices, shift the indices so 1/2
   are neg, make every other value neg, etc. This is a pretty simple and
   deterministic test.
 - Sort int/real arrays of random values between min(T) and max(T) where
   T is all int/uint types. e.g. it creates an int/real array where the
   values fit in 16-bit ints. This is representative of the typical
   input I would expect for sorting.
 - Sort int arrays where we take random 64-bit values and mask to create
   different interesting "active bit ranges". For instance we might mask
   so only the first or last 32-bits are active, or mask so that every
   other bit or only the first and last bits are active. This is a weird
   test, but it has uncovered a lot of bugs in my optimizations. Radix
   sort treats values as "bags of bits" and this stresses weird bit
   patterns.

The correctness test uses 10_000*numLocales elements. There are a lot of
variations, especially for the activeBitRange checks. Those can be
skipped with `--mode=testMode.correctnessFast`

The performance test is basically what I used in #162. It sorts int(64)
arrays populated with uint(16) values using 1/50 of memory. The array
type, value range, and memory fraction are configs. The default was
chosen because it will only go through the main radix sort loop once, so
it gives a good baseline.

This also fixes bugs uncovered by the sort test.

 - Ensure getBitWidth is bounded by `numBits(T)`. If min(int) was
   present it would return 65, which isn't valid.
 - Fix how negative numbers were being checked for. It was broken for
   `-0.0`, so swap to using `signbit` for reals.
 - `getDigit` for reals was trying to convert the real to a uint and
   then shift that. However, it was converting with casts that resulted
   in type-punning which is UB. Instead use memcpy to cast. This is the
   right way to convert types, and all C compilers will optimize this
   pattern (we have used this idiom and verified the asm is optimal in
   our runtime.)

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -46,29 +46,15 @@ module RadixSortLSD
         return ((key >> rshift) & maskDigit);
     }
 
-    proc shiftDouble(in key: real(64), rshift: int(64)) {
-      const ptrToReal = c_ptrTo(key);
-      const ptrToULL = ptrToReal: c_ptr(int(64));
-      const intkey = ptrToULL.deref();
-      return (intkey >> rshift);
+    inline proc realToUint(in r: real): uint {
+        var u: uint;
+        c_memcpy(c_ptrTo(u), c_ptrTo(r), numBytes(r.type));
+        return u;
     }
-    pragma "no doc" // Bug: chapel-lang/chapel#14250
-    /*
-    extern {
-      static inline unsigned long long shiftDouble(double key, long long rshift) {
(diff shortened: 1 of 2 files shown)


ronawho:
This is split off from some upcoming sort optimizations (optimizing sort for negative values, and improving aggregation.)

Both those efforts are taking longer than I expect so I'm peeling off parts to make them easier to manage.

PR #213, opened 2019-12-10
only use arkouda version on the server compile line

mhmerrill:
fixed Makefile to only use arkouda version number on the server compile and not the unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,13 +85,15 @@ ifneq ("$(wildcard $(VERSIONFILE))","")
 else
 	VERSION=$(shell date +'%Y.%m.%d')
 endif
-	# Version needs to be escape-quoted for chpl to interpret as string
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -85,13 +85,15 @@ ifneq ("$(wildcard $(VERSIONFILE))","")
 else
 	VERSION=$(shell date +'%Y.%m.%d')
 endif
-	# Version needs to be escape-quoted for chpl to interpret as string
-    CHPL_FLAGS += -sarkoudaVersion="\"$(VERSION)\""
+
+# Version needs to be escape-quoted for chpl to interpret as string
+CHPL_FLAGS_WITH_VERSION = $(CHPL_FLAGS)
+CHPL_FLAGS_WITH_VERSION += -sarkoudaVersion="\"$(VERSION)\""
```
BryantLam: Also acceptable to collapse the two lines into one.

PR #216, opened 2019-12-11
Simplify aggregation usage

ronawho:
Instead of having to select between aggregation and unorderedCopy at the
call site, wrap that logic inside the aggregation code. The only
functional impact this should have is always enabling the DstAggregator
under ugni (before it wasn't used for the metadata in sorting.) That has
a negligible performance impact.

This cleans up the aggregation calls in the sorting code, and should
make using aggregation easier in the future.

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -1,11 +1,32 @@
 module CommAggregation {
   use SysCTypes;
+  use UnorderedCopy;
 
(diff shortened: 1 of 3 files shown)


ronawho:
This is easiest to review ignoring whitespace: https://github.com/mhmerrill/arkouda/pull/216/files?w=1

PR #221, opened 2019-12-11
Add makefile targets for installing hdf5 and zmq

ronawho:
It's easy to install Arkouda dependencies on macs with Homebrew, but on
Crays and Linux clusters, it's harder to install dependencies. On Crays
there is a cray-hdf5 module available, but I've still needed to install
zeromq manually. On InfiniBand clusters I've had to install zeromq and
hdf5 manually. This adds `make install-hdf5` and `make install-zmq`
targets to make installing dependencies easier for users in a similar
boat. There is also a `make deps` that will install both.

This curls the latest versions (zmq 4.3.2 / hdf5 1.10.5), untars,
configures with optimizations, runs make and make install, and adds the
install dirs to Makefile.paths.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,43 @@ endef
 #                               ^ no space after comma
 -include Makefile.paths # Add entries to this file.
 
+.PHONY: install-deps
+install-deps: install-zmq install-hdf5
+
+DEP_DIR := dep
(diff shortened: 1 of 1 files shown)


BryantLam:
I think this is fine, but it's also somewhat of a slippery slope versus simply providing per-platform instructions on how to do it. These steps would work for the majority of users.. except when it doesn't. :)

PR #223, opened 2019-12-11
Add a mechanism to write/read the hostname and port to/from a file

ronawho:
Add `--serverConnectionInfo` (`ARKOUDA_SERVER_CONNECTION_INFO`) that
will have the server write connection info (`hostname:port`) out to a
file. Also have the client read from the file indicated by the env var
if the var is set.

When running the server and client on a machine with a shared file
system this makes make it a lot easier to convey hostname/port
information automatically instead of having to manually copy the values.

This then adds a `make check` that uses this functionality and starts
the server, runs checks.py, and shutdown the server.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -250,14 +250,8 @@ CLEAN_TARGETS += doc-clean
 doc-clean:
 	$(RM) -r $(DOC_DIR)
 
-COMM = $(shell $(CHPL_HOME)/util/chplenv/chpl_comm.py 2>/dev/null)
-ifneq ($(COMM),none)
-NUM_LOCALES := -nl 2
-endif
-SERVER_CONNECTION_INFO := $(ARKOUDA_PROJECT_DIR)/ak-server-info
 check:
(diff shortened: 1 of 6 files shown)


ronawho:
Opened as a draft for discussion. 

This is mostly motivated by some work I'm doing to add nightly performance testing, but I also find that copying the hostname/port for my manual runs gets old after a while.

If this seems attractive, I think the next step is to wrap the client side logic in some common routine that all the benchmarks/tests can use.

ronawho:
This will help with setting up automated testing, and I think can be useful even for regular users. 

Passed the new `make check` on OSX (comm=none), Cray-CS (gasnet-ibv), Cray-XC (ugni)

ronawho:
If this seems worthwhile, the tests/benchmarks will need to be adjusted to allow not passing a hostname/port. 

I'm thinking it might be cleaner to leave `connect()` as it was, and instead add `get_default_server()`/`get_default_port()` functions that are then used from tests/benchmarks.

ronawho:
Closing in favor of https://github.com/mhmerrill/arkouda/pull/243, which I think is a cleaner solution.

PR #224, opened 2019-12-11
disambiguate private config const in SegStringSort.chpl

mhmerrill:
renamed `private config const v` in SegStringSort.chpl to prevent conflict on the command line with the normal one in ServerConfig.chpl

### DIFF
--- a/src/SegStringSort.chpl
+++ b/src/SegStringSort.chpl
@@ -6,10 +6,14 @@ module SegStringSort {
   use CommAggregation;
   use PrivateDist;
 
-  private config const v = true;
-  private config const numTasks = here.maxTaskPar;
-  private config const MINBYTES = 8;
-  private config const MEMFACTOR = 5;
+  private config const SSS_v = true;
+  private const v = SSS_v;
+  private config const SSS_numTasks = here.maxTaskPar;
+  private const numTasks = SSS_numTasks;
+  private config const SSS_MINBYTES = 8;
+  private const MINBYTES = SSS_MINBYTES;
+  private config const SSS_MEMFACTOR = 5;
+  private const MEMFACTOR = SSS_MEMFACTOR;
 
   record StringIntComparator {
     proc keyPart(a: (string, int), i: int) {


reuster986:
I don't understand why this is necessary... I thought the `private` keyword already disambiguates the variables, i.e. you already have to specify either `--RadixSortLSD.numTasks=8` or `--SegStringSort.numTasks=8`, right?

BryantLam:
I certainly think that should be the case, but `private` is much more recent than `config` variables have been around. This is likely a now-known bug. @bradcray?

mhmerrill:
we need to disambiguate it with respect to `--ServerConfig.v` so we can use `--v=false` on the command line.

https://github.com/chapel-lang/chapel/issues/14622

> I don't understand why this is necessary... I thought the `private` keyword already disambiguates the variables, i.e. you already have to specify either `--RadixSortLSD.numTasks=8` or `--SegStringSort.numTasks=8`, right?

mhmerrill:
@BryantLam it is not a bug it is a feature.

> I certainly think that should be the case, but `private` is much more recent than `config` variables have been around. This is likely a now-known bug. @bradcray?

BryantLam:
Right, not strictly a bug. I meant to say that this would likely behave better as a feature request to make `private config` variables not be ambiguous with `public config` variables.

mhmerrill:
> Right, not strictly a bug. I meant to say that this would likely behave better as a feature request to make `private config` variables not be ambiguous with `public config` variables.

we made an issue https://github.com/chapel-lang/chapel/issues/14622

bradcray:
@BryantLam: Mike beat me to it, but since you tagged me:

* There's no ambiguity in the source code itself and things could keep the same names there.  
* Bill's right that fully-qualifying the config names on the command-line could be used to disambiguate them while they have the same name.
* Mike wanted to keep the command-line simpler and focused on the public config which led to this PR.
* On the issue he tagged, it's been proposed that for the sake of command-line disambiguation, a single public config would "win out" over one or more private configs with the same name.  This would be a new / additional interpretation of private (and one specific to configs), but a change that could be convenient in cases like this, and which seems to be getting support from the team so far.

PR #232, opened 2019-12-13
Optimize sorting negative values

ronawho:
Sort negatives directly instead of reordering them after sorting.
Previously, the radix sort code would result in negatives being reversed
and on the right hand side of the array. The old code would then perform
a final permutation to get them in the right order, but this added a lot
of extra comm. Here we use some bit tricks so negatives sort correctly.

For ints, we just invert the signbit so negatives get sorted correctly.
Because of the other getBitWidth optimization, we have to track if we're
sorting the last digit and if there was a signbit to flip (were there
any negatives.) Without the getBitWidth trick, this would be something
like:

```chpl
key ^= (1<<63);
```

For reals, we invert the signbit for positive numbers, and invert all
the bits for negative numbers. Without the getBitWidth trick this would
be something like:

```chpl
if signbit(key) {
  key = ~key;
} else {
  key ^= (1<<63);
}
```

There should not longer be any performance penalty for sorting negatives.

```
chpl UnitTestSort.chpl --fast -M ../src/ -sperfOnlyCompile -sperfValRange="int(16)"
./UnitTestSort --mode=testMode.performance -nl 16
```

Used to get ~125 MB/s per node and now gets ~600 MB/s per node, which is on par
with sorting `-sperfValRange="uint(16)"`

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -16,7 +16,6 @@ module RadixSortLSD
     use BlockDist;
     use BitOps;
     use AryUtil;
-    use UnorderedCopy;
     use CommAggregation;
     use IO;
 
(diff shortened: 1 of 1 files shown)


ronawho:
Easiest to review commit-by-commit.


Originally motivated by:
 - https://github.com/chapel-lang/chapel/blob/91b536840fd795d4b8d321dedddc474f9b96949c/modules/packages/Sort.chpl#L1449-L1454
 - https://github.com/chapel-lang/chapel/blob/91b536840fd795d4b8d321dedddc474f9b96949c/modules/packages/Sort.chpl#L2968-L2987

Issue #230, opened 2019-12-12
fix pip3 install arkouda for installs with zmq and hdf5 already installed in non-standard places


mhmerrill:
fix pip3 install arkouda for installs with zmq and hdf5 already installed in non-standard places.

I am having trouble with `pip3 install -e .` on our IB cluster where we have libraries installed not installed by `pip3`

Currently, we need to pip re-install when we rev the version number which we should also fix and detangle these issues.

@reuster986 @Spartee you guys should look at this.

reuster986:
@mhmerrill this is a pretty thorny issue, and I'm not sure exactly what to do. I propose that we support two official ways of installing arkouda:

* User mode
  * For non-developers who just want server and client to install with one command
  * Syntax: `pip install arkouda`
  * Status: almost done
    * Works on mac
    * On linux, fails to build `arkouda_server` because it can't find zmq or hdf5, even with `LD_LIBRARY_PATH` set. I propose we modify `installers.py` to copy `LD_LIBRARY_PATH` to `Makefile.paths` and check that it can build a trivial Chapel program linked against zmq and hdf5. I am willing to do this, if there are no objections.
* Developer mode
  * For developers or users who need to decouple the building of `arkouda_server` from the installation of the `arkouda` python package.
  * Current method: `make`, followed by `pip install -e /path/to/arkouda/repo`
  * Status: needs work
    * @mhmerrill what specific error is this command giving with respect to zmq and hdf5? I can't reproduce on sherlock because I can't re-install over your copy.
    * @mhmerrill note that even if there is a custom build of zmq for linking into the `arkouda_server`, the `arkouda` python package still needs the `zmq` python package, so `pip` still has to install `zmq`.
    * As @mhmerrill points out, the current code gets the `__version__` number from when the python package was built, i.e. when `pip install -e` was run. This is fine for user mode but wrong for development mode, because the `__version__` should be set at import-time. I think this can be done by reading the `VERSION` file on import, as long as we can force setuptools to always include this file in the distribution, so that it also works with user mode. @Spartee does it already do this, or do you know how to make it do so? I can't test it, because I can't complete a normal `pip install arkouda` to look at the result.

mhmerrill:
@ronawho proposed a fix for me which works for me.
```bash
make -j install-deps
pip3 install -e . --user
make
```

reuster986:
I believe #233 fixes this.

PR #233, opened 2019-12-13
Dependency check

reuster986:
This is an attempt to address #230 and improve both user and developer experience with three main changes:
* `Makefile` now adds contents of `LD_LIBRARY_PATH` and `LD_RUN_PATH` to compile line, so linux users can `pip install arkouda` if they have ZMQ and HDF5 in these paths. Until now, these libs had to be in `Makefile.paths`, but there is no opportunity to modify that file when doing a `pip install`, hence the need for reading the environment variables.
* Before building `arkouda_server`, `make` now builds and runs two tiny Chapel programs that link ZMQ and HDF5, just to test their availability. This adds an extra ~10 seconds to the build time, but now the build fails immediately if it can't find the libs, instead of failing 10 minutes in. I have been burned enough times by this that I figured it would be worth the trade-off.
* The arkouda python package now sets `__version__` by reading the `VERSION` file at import time, rather than using the value set at build time. This frees developers from having to re-install with `pip` every time they rebuild `arkouda_server` to avoid false warnings about version mismatches. Now you should just be able to `pip install -e /path/to/arkouda` once and for all.


mhmerrill:
@ronawho the travis build is failing on this I think due to the new "canary" programs which are compiled first. How do we fix this?

ronawho:
Blegh, that's fallout from https://github.com/mhmerrill/arkouda/pull/208, which added `--stop-after-pass codegen` to `CHPL_FLAGS` for travis testing. A lame workaround would be to add `--stop-after-pass makeBinary` for compiling these checks. 

I'll think about a cleaner solution, but that's the best I have at the moment.

Spartee:
@reuster986 The changes to the pip installers file look good, one thing I noticed in the prior implementation was that it would only work with the prefix of a virtual environment and some brew installed distributions of python, but that should be fixed with your changes. 

In reference to @mhmerrill questions in #230. The version issue should be addressed by these changes. We could incorporate a development build process which used to be a part of the ``setup.py`` install process(``python setup.py develop``), but that behavior is currently in the process of being deprecated and is only used as a fallback when the pip install fails. However, this could still be useful for devs.

PR #235, opened 2019-12-13
Convert most uses of unordredCopy to aggregation

ronawho:
Fine grained copies only perform well under Aries/ugni. To get bulk comm
use aggregation like we have previously done for sorting and indexingmsg

Part of https://github.com/mhmerrill/arkouda/issues/222

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -13,8 +13,7 @@ module ArgSortMsg
     
     use PrivateDist;
 
-    use UnorderedCopy;
-    use UnorderedAtomics;
+    use CommAggregation;
 
     use AryUtil;
     use PerLocaleHelper;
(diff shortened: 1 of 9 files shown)


ronawho:
Last commit is the one to look at, will rebase after https://github.com/mhmerrill/arkouda/pull/232. 

This is not ready to merge yet. There are a few place marked TODO that weren't trivial to convert. I may need help with some of those. 

I will likely also need help with testing

ronawho:
@mhmerrill and/or @reuster986 

I feel reasonably confident about the changes in this PR, but there are a couple of remaining calls to unorderedCopy I need help converting. I think they are all labeled with `TODO`. Feel free to merge this and convert them later or leave inline comments and I'll update them.

I'm also not really sure what tests to run beyond `make check` (checks.py), suggestions are welcome.

ronawho:
I can add a new aggregator that handles when either side might be remote. The only reason there's currently distinct Src/Dst aggregators was to limit memory footprint when you only need to aggregate in one direction. If there's cases where we need to aggregate in both directions, it should be easy to add a general `CopyAggregator` to handle that. 

I'd been hoping to add a `CopyAggregator` anyways, to make the aggregators easier to use -- that's captured in https://github.com/mhmerrill/arkouda/issues/237

PR #238, opened 2019-12-17
Migrate from Travis-CI to Github actions

ronawho:
Travis has lower RAM limits and no swap, which has led to OOMs for the
compiler. Switch over to github actions, which has 7GB of RAM and 7GB of
swap.

I pieced this together with trial and error and using:
 - https://help.github.com/en/actions/automating-your-workflow-with-github-actions
 - https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions

Personally, I think this is cleaner than what we had before, and it's
using a native github feature instead of something external.


ronawho:
There's a clone of this at https://github.com/ronawho/arkouda/pull/5 where the testing has actually triggered. I _think_ this should just enabled for future PRs after this is merged, but I'm not positive. It might be the case that somebody with write permission to the repo needs to create this PR.

mhmerrill:
@ronawho is there anything I have to do to the repo to get this to work?

ronawho:
It should just work. Apparently the 'push' command doesn't trigger events for PRs from a forked repo. I've added the `pull_request` event in https://github.com/mhmerrill/arkouda/pull/240, which seems to work.

PR #248, opened 2019-12-21
[Review] Allow Arkouda to Return Numba CUDA Array for GPU Processing

awthomp:
This is a work in progress PR and forum for discussion on how to add GPU awareness to the Arkouda project in the best and most reasonable way. I've added a new function in the `pdarrayclass` to pass a `pdarray` type to CUDA using the popular Numba Python package. This also adds Numba and the CUDA Toolkit (most easily installed via Conda) as dependencies. Numba's [`__cuda_array_interface__`](https://numba.pydata.org/numba-doc/dev/cuda/cuda_array_interface.html) standard with `deviceNDArray` sets standards for how memory is organized on GPU. If a library accepts the `__cuda_array_interface__`, you can zero-copy on GPU between compatible frameworks.

This PR allows someone to load in large datasets into Arkouda and interact with it as expected, take a subset of relevant data, and then push it to GPU for speciality processing. By leveraging Numba, this allows us to open the aperture to NVIDIA's RAPIDS libraries for data science and signal processing, CuPy for GPU accelerated NumPy, and PyTorch for DL.

I've currently only tested single GPU usage, but I imagine this PR would be helped by Numba adopting CUDA UVM support (Issue here: https://github.com/numba/numba/issues/4362) 

Sample usage:
```python
import arkouda as ak
ak.connect('adamt-nvidia', 5555)

N = 10**6
arr = ak.arange(1, N+1, 1)
reduce_arr_ak = ak.sum(arr)

# CUDA - cuDF (RAPIDS DataFrame)
import cudf
cuda_arr = arr.to_cuda()
gdf_reduce = cudf.Series(cuda_arr).sum()

# CUDA - CuPy
import cupy as cp
# look at cuda_arr addresses
cuda_arr.__cuda_array_interface__['data']
# Examine cupy context (this is zero copy)
cp.asarray(cuda_arr).__cuda_array_interface__['data']
# Perform reduction
gpu_reduce_cupy = cp.sum(cp.asarray(cuda_arr))

# Can also do all the other cool CuPy functions
gpu_fft = cp.fft.fft(cp.asarray(cuda_arr))

# With PyTorch 1.4 (currently in nightly), you can go to PyTorch without DLPack
import torch
gpu_tensor = torch.as_tensor(cuda_arr)

# Finally, we can go back to Arkouda from the Numba deviceNDArray type via
ak.array(cuda_arr)
```

Looking forward to the discussion and making this ultimately better and beyond a single GPU. This could also use the eyes of more NVIDIA folks who better understand the multi-GPU Numba use-cases.

You can install the latest RAPIDS packages with:
`conda install -c rapidsai-nightly -c nvidia -c conda-forge \
    -c defaults rapids=0.12 python=3.7 cudatoolkit=10.1`

The latest version of PyTorch can be installed with:
`conda install -c pytorch-nightly pytorch`

I'd like to again stress that this is a 'quick and dirty' _hack_ for enabling GPU support within the Arkouda project.

Happy Holidays,
Adam Thompson

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -522,11 +522,11 @@ class pdarray:
         Examples
         --------
         >>> a = ak.arange(0, 5, 1)
-        >>> a.to_ndarray()
+        >>> a.to_cuda()
         array([0, 1, 2, 3, 4])
 
-        >>> type(a.to_ndarray())
-        numpy.ndarray
+        >>> type(a.to_cuda())
+        numpy.devicendarray
         """
         # Total number of bytes in the array data
         arraybytes = self.size * self.dtype.itemsize
@@ -540,8 +540,8 @@ class pdarray:
             raise RuntimeError("Expected {} bytes but received {}".format(self.size*self.dtype.itemsize, len(rep_msg)))
         # Use struct to interpret bytes as a big-endian numeric array
         fmt = '>{:n}{}'.format(self.size, structDtypeCodes[self.dtype.name])
-        # Return a numpy ndarray
-        return cuda.to_device(np.array(struct.unpack(fmt, rep_msg)))
+        # Return a numba devicendarray
+        return cuda.to_device(struct.unpack(fmt, rep_msg))
 
     def save(self, prefix_path, dataset='array', mode='truncate'):
         """
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -1,10 +1,5 @@
 import json, struct
 import numpy as np
-try:
-    from numba import cuda
-except ModuleNotFoundError as error:
-    print(error)
-    print('CUDA is not enabled or installed. Consult documentation.')
 
 from arkouda.client import generic_msg, verbose, maxTransferBytes, pdarrayIterThresh
 from arkouda.dtypes import *
@@ -532,6 +527,12 @@ class pdarray:
         >>> type(a.to_cuda())
         numpy.devicendarray
         """
+        try:
+            from numba import cuda
+        except ModuleNotFoundError as error:
+            print(error)
+            print('CUDA is not enabled or installed. Consult documentation.')
+        
         # Total number of bytes in the array data
         arraybytes = self.size * self.dtype.itemsize
         # Guard against overflowing client memory
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -1,5 +1,6 @@
 import json, struct
 import numpy as np
+from numba import cuda
 
 from arkouda.client import generic_msg, verbose, maxTransferBytes, pdarrayIterThresh
 from arkouda.dtypes import *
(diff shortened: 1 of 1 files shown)


awthomp:
Also -- I want to be explicit that I don't recommend merging this request at the time (especially since this PR has hard dependencies on Numba and the CUDA toolkit; it would break an install without these). Again, I wanted to provide a quick demonstration of one way someone could connect Arkouda to some of the robust GPU based data science and deep learning libraries. Some immediate issues to solve/discuss:

1. Is there a better way besides Numba to handle an array transfer from Arkouda, particularly for multi-node/multi-GPU?
2. Do we want to make CUDA and Numba a hard dependency on Arkouda? This should probably be done/configured in the build process and is less clunky than checking if a GPU exists with something like Numba
3. Configure error handling for size pdarray > available GPU memory and if a user tries to invoke a to_cuda() command if a GPU isn't available

Just a few top of mind issues to consider. 

As some of ya'll may know, I'm on paternity leave through Jan 6th, so my bandwidth here is when my 2 month old is sleeping or eating! 😄

reuster986:
Welcome, @awthomp , and thanks for opening the discussion! Also, congratulations on the birth of your child!

We should all talk through the kinds of workflows we would want to enable with GPU support in arkouda. I can think of two categories:

* The arkouda server is running on a cluster/MPP, and the python client is running on a node with a GPU (this is the use case this PR targets). This setup would be useful for, e.g., loading a giant edge list on the MPP for communication-intensive sorting and gather operations to construct a graph that fits on the client node, then running something computationally intensive like spectral clustering on the GPU. We see this pattern -- boil the ocean, then pull results back to python for further analysis -- a lot with our users, so this approach should be fruitful. It also has the advantage that the handoff between arkouda and CUDA would be fairly light-touch, as this PR demonstrates.

* The arkouda server is running on a cluster/MPP with GPUs on the compute nodes. These systems are becoming more common, and enabling arkouda (and Chapel) to target those GPUs would open up a lot of possibilities. I know the Chapel team has been thinking through GPU support for a while now, so I will defer to them on long-term strategy. For the short term, would it make sense to target onboard GPUs from Chapel for no- or low-communication operations by creating a `__cuda_array_interface__` around each local slice of an array? I am wondering if scans would be a good test case for this.

Regarding questions 2 and 3, I agree that it probably makes sense to do a build-time check for GPU, rather than making CUDA and Numba hard dependencies. For the python code, you can move the `from numba import cuda` line inside the `to_cuda()` function, and it won't try the import until a user calls `to_cuda()`. This way, users can `import arkouda` and use most functionality without having Numba, but will get a `ModuleNotFoundError` if they try to call `to_cuda()`, which seems like the appropriate behavior. 

```python
def to_cuda(self):
    ...
    # Raises ImportError on function call, not arkouda import
    from numba import cuda
    ...
    return cuda.to_device(struct.unpack(fmt, rep_msg))
```

My availability will also be fairly low the next few weeks, but I will be following the discussion.

awthomp:
Hi @reuster986! Thanks for the kind welcome and nice words on our first child. It's been a whirlwind two months, but a complete joy.

Let's call a two week break and pick up the CUDA integration discussion in the new year. I like your thinking on both approaches, and I'm sure we'd be happy to meet with the Chapel developers to consider GPU integration for your 2nd point. Also, I'm not a data structures expert, but from what I've seen, implementation of the `__cuda_array_interface__` in other libraries seems to be relatively straight forward. In [CuPy](https://github.com/cupy/cupy), it's just a [decorator on the ndarray class.]( https://github.com/cupy/cupy/blob/bdd8e6b9015e4287b9b331a141b99b4d929fc984/cupy/core/core.pyx#L148)

Lastly, I submitted another commit to the PR to throw an exception if Numba isn't found. I liked your idea to place the import in the to_cuda function, but thought that might negatively affect performance if a user has the CUDA libs installed and runs that call multiple times. We still might want to more 'elegantly' handle the error catching, but now we have a try/except on numba.cuda. If a user tries to invoke the `to_cuda` function anyway, they'll get a `NameError: name 'cuda' not defined`.

Happy holidays! Excited to contribute 😄

awthomp:
Spoke to @mhmerrill recently about this PR, and I want to move this from WIP to Review. There's a try/except on importing Numba, so if a user invokes the proposed `to_cuda` function, a NameError will be thrown that Numba/CUDA isn't supported. To reiterate, this functionality will only currently support one GPU, but it will open up that resulting device array to the `__cuda_array_interface__` ecosystem.

mhmerrill:
@awthomp I'd like @reuster986 to review this PR, if he is ok with it then I'll be ok with it.

mhmerrill:
@awthomp I did take a quick look. I think I'd like the `pad.to_cuda()` to raise an exception if the import failed or potentially not even be defined.

awthomp:
I'll research how to throw this function-wise exception in a reasonable way @mhmerrill. I know @reuster986 had suggested putting the `import numba` into `pdarray.to_cuda()`, but there's got to be a more elegant way of handling optional dependencies within Python!

awthomp:
@reuster986 @mhmerrill -- I've moved the Numba error handling inside `to_cuda`, so it should now be isolated. Further, I noticed that you could successfully import cuda from numba even if you aren't on a machine with a GPU. To prevent this, I'm catching for `cuda.is_available()`.

Let me know if you have any suggestions on a better way to handle these optional dependencies.

Issue #180, opened 2019-11-05
allow negative index to pdarray


mhmerrill:
Allow `pda[-1]` indexing backwards from the end of a pdarray

mhmerrill:
I believe this can be closed now, after #249

PR #249, opened 2019-12-26
Added negative int indexing and tests for same

reuster986:
I can't believe it took me this long to do this, lol. Now you can use a negative integer index with a `pdarray` in the Pythonic fashion, where `a[-i]` means "get me the `i`th value from the end".


reuster986:
Resolves #180

Issue #250, opened 2019-12-26
Aggregation with loop-private variables?


reuster986:
@ronawho I'm not sure if this is a bug to fix or just a pitfall that I need to learn how to avaoid, but aggregators don't seem to play well with variables created inside a `forall`. For example, this code does not correctly copy `A` to `B`:

```chapel
var D: domain(1) dmapped Block(boundingBox={0..#N}) = {0..#N};
var A: [D] int = [i in D] i;
var B: [D] int;
forall (b, i) in zip(B, D) with (var agg = new SrcAggregator(int)) {
    var a = A[i];
    agg.copy(b, a);
}
```

Each task's chunk of `B` ends up filled with a constant value (the last value in that task's chunk of `A`). 

The `SegString` gather code stumbles over this same behavior [here](https://github.com/mhmerrill/arkouda/blob/d9ec5191ce93cade222de8ee3ee438f6bc4ecfb4/src/SegmentedArray.chpl#L129-L137). I'll be submitting a PR with a workaround shortly, but I wanted to ask you if there's a better way to handle this.

ronawho:
The `SrcAggregator` aggregates the source and destination addresses. In this case you're aggregating the address of the variable `a`, not the address of `A[i]`. When the aggregator is flushed it copies the current value of `a` to all the destination addresses.

I think what you'd want here is `agg.copy(b, A[i]);` or `ref a = A[i]; agg.copy(b, a);`

Not that `var a = A[i]` is effectively `var a = GET(A[i]);` so it's not actually aggregating the comm anyways.

reuster986:
I'm going to close this because there is nothing to be done except learn from the experience.

Issue #226, opened 2019-12-11
Implement != for strings


reuster986:
I just forgot to do this in the big PR. Actually, there is an opportunity for code consolidation here with FindSegmentsMsg.chpl:114 or thereabouts.

PR #251, opened 2019-12-26
String update and fix

reuster986:
This PR originally set out to address #226 (which it does). While testing, I found a bug in the code for gathering from a string array with integers. I traced the root of the bug and filed #250 while creating a workaround that fixes things for now.

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -124,17 +124,20 @@ module SegmentedArray {
       var t1 = getCurrentTime();
       ref oa = offsets.a;
       const low = offsets.aD.low, high = offsets.aD.high;
-      // Lengths of segments including null bytes
-      var gatheredLengths: [D] int;
-      forall (gl, idx) in zip(gatheredLengths, iv) with (var agg = new SrcAggregator(int)) {
-        var l: int;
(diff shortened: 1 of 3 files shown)


src/SegmentedArray.chpl
```diff
@@ -124,17 +124,20 @@ module SegmentedArray {
...
-      forall (gl, idx) in zip(gatheredLengths, iv) with (var agg = new SrcAggregator(int)) {
-        var l: int;
+      // Gather the right and left boundaries of the indexed strings
+      // NOTE: cannot compute lengths inside forall because agg.copy will
+      // experience race condition with loop-private variable
+      var right: [D] int, left: [D] int;
+      forall (r, l, idx) in zip(right, left, iv) with (var agg = new SrcAggregator(int)) {
         if (idx == high) {
-          l = values.size - oa[high];
+          agg.copy(r, values.size);
```
ronawho: I don't think you want to aggregate here. This will end up as something like `const tmp = values.size; agg.copy(r, tmp);` so everything is local. 

What was the original code you wanted to write here?

mhmerrill: I think this is just for the last value.

PR #260, opened 2020-01-17
Make the server compatible with Chapel master

e-kayrakli:
This PR updates Arkouda server to be able to use Chapel master while keeping it
compatible with 1.20. This is mainly motivated by UTF-8 validation on Chapel
strings that was added after 1.20. With that change, messages with arbitrary data
can no longer use Chapel `string`s and should use `bytes` instead.

Summary of changes in the order of importance:
- Updates `arkouda_server` to use `bytes` type when the message can include
  arbitrary data. This is currently limited to `array` and `tondarray` commands.

- Updates some of the helper functions to handle `bytes` messages.

- Adds `src/Chapel120.chpl` as a small compatibility layer. This file includes
  some ZMQ and IO overloads that converts `bytes` to `strings` in order to be
  able to use Chapel 1.20, which had some unimplemented features for `bytes`

- Adds `CHPL_VERSION_120` and `CHPL_COMPAT_FLAGS` to Makefile. These are set
  appropriately depending on what `chpl --version` returns.

- Adds few `use` statements. Because `use` is by default private now on Chapel
  master, some Arkouda source files need to explicitly `use` Chapel and/or
  Arkouda modules

Note: Some (all?) unit tests in the `test` dir may also need some work to bring
them up to speed with Chapel master. This PR does not do that as of now.
Those changes can be added to this PR or can be a follow up. But it depends on
whether we move forward with the changes in this PR.

Tested with `check.py` and `groupby_test.py` which normally break with UTF-8
validated strings on Chapel:
- [x] They pass with Chapel 1.20
- [x] They pass with Chapel Master (fee8582)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -146,11 +146,18 @@ endif
 # Version needs to be escape-quoted for chpl to interpret as string
 CHPL_FLAGS_WITH_VERSION = $(CHPL_FLAGS)
 CHPL_FLAGS_WITH_VERSION += -sarkoudaVersion="\"$(VERSION)\""
+
+ifeq ("$(shell chpl --version | sed -n "s/chpl version 1\.\([0-9]*\).*/\1/p")", "20")
+	CHPL_COMPAT_FLAGS := -sversion120=true --no-overload-sets-checks
+else
+	CHPL_COMPAT_FLAGS := -sversion120=false
+endif
+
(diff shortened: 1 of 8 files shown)


mhmerrill:
I am going to go ahead and merge this.

PR #265, opened 2020-01-23
Make the server compatible with Chapel master

e-kayrakli:
Updates Arkouda server so that it is compatible with Chapel 1.20 and master.

- Adds explicit types to map constructor, and changes the order to match
  master's order.

- Adds few uses since `use` is private by default on master now.

Test:
- [x] check.py and few other scripts pass

### DIFF
--- a/src/RandArray.chpl
+++ b/src/RandArray.chpl
@@ -3,6 +3,7 @@ module RandArray {
   use SegmentedArray;
   use ServerErrorStrings;
   use Map;
+  use SipHash;
 
   proc fillInt(a:[] ?t, const aMin, const aMax) where isIntType(t) {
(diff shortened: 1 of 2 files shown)


e-kayrakli:
I am not quite sure whether `ArgumentError` being defined in `src/SipHash.chpl` is the right thing. But regardless I added a relevant `use` statement to be able to use `ArgumentError`. Also there is [IllegalArgumentError](https://chapel-lang.org/docs/builtins/ChapelError.html?highlight=illegalargumenterror#ChapelError.IllegalArgumentError) that you can use if it makes sense. But this PR doesn't make that change and is limited to getting things working on Chapel master (and 1.20)

PR #269, opened 2020-01-28
String sort hash improvement

reuster986:
Improves performance of sorting, gather, and in1d for strings.

This PR incorporates @ronawho 's early changes to `src/SipHash.chpl`, as a stopgap untiil the work is complete.

### DIFF
--- a/src/SipHash.chpl
+++ b/src/SipHash.chpl
@@ -1,5 +1,5 @@
 module SipHash {
-
+  
   param cROUNDS = 2;
   param dROUNDS = 4;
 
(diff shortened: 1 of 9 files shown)


src/SipHash.chpl
```diff
@@ -49,32 +49,28 @@ module SipHash {
...
   }
 
-  proc sipHash128(msg: [] uint(8), k: [?kD] uint(8)): 2*uint(64) throws {
-    return computeSipHash(msg, k, 16);
+  proc sipHash128(msg: [] uint(8), D): 2*uint(64) throws {
+    return computeSipHash(msg, D, 16);
   }
   
-  private proc computeSipHash(msg: [?D] uint(8), k: [?kD] uint(8), outlen: int) throws {
+  private proc computeSipHash(msg: [] uint(8), D, param outlen: int) throws {
```
ronawho: I didn't mean to commit making outlen a `param` (though it was something I was playing with since it would eliminate some runtime branches.) I might leave that change out for now and let me experiment with that some more.

mhmerrill: @reuster986 do you want to make this change?

reuster986: @ronawho sorry for stealing your work-in-progress code, I just had to run with something quickly. I don't mind leaving outlen as a `param` in this PR, since it works and should be faster (right?). If in the course of your experimentation you decide to change it back, we can just apply your PR on top of this one.

The other change that's in this line is that I took out the hash key, `k`, as an argument and instead hard-coded a value for it inside the function. The `k` argument was incurring some extra communication in multi-locale setting that really messed with the scaling, and we don't really need a keyed hash here anyway.

ronawho: > sorry for stealing your work-in-progress code

Not a problem, glad that was enough of an improvement to be helpful

>  I don't mind leaving outlen as a param in this PR...

Sounds good. I tend to be pretty methodical (and usually pretty slow) with performance changes just so I can keep track of what changes improved what and by how much, but you're right that it should be strictly better.

Issue #272, opened 2020-02-01
Split strategy for string gather


reuster986:
Testing on Cray XC, the code prior to [this commit](https://github.com/mhmerrill/arkouda/commit/84914416ea7a5417396ca63e0edd892683d2c50a) is much faster on ugni than the current version. However, the current version is best for non-ugni because it uses aggregation. I want to make this code select the correct path based on `CHPL_COMM`.

PR #275, opened 2020-02-04
Split string gather

reuster986:
Addresses #272 by conditioning the strategy for gathering strings on `CHPL_COMM`. On single-locale or ugni, it uses unorderedCopy. On other networks, it now pre-computes a vector of local source indices  usable with `newSrcAggregator`. This costs a lot of memory, but avoids "death by small messages" on infiniband. This PR passes `tests/string_test.py` on single-locale linux.

Included is a new performance test for string gather. Some runs with `test-bin/StringGatherSpeedTest --N=1_000_000 --MEANLEN=100`:

| Platform | CHPL_COMM | Locales | strategy | Rate |
| ----------- | ----------- | --------- | ------ | ------ |
| laptop | none | 1 |  unordered | 172 MB/s |
| laptop | gasnet | 2 | aggregation | 4.6 MB/s |
| laptop | gasnet | 4 | aggregation | 3.1 MB/s |
| IB cluster | gasnet | 4 | aggregation | 10.3 MB/s |
| IB cluster | gasnet | 4 | unordered | 0.7 MB/s |

I think my aggregation strategy could still use work -- there is probably a more efficient way to do this, but this is an improvement for now.

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -186,6 +186,7 @@ module SegmentedArray {
            it is the difference between the src offset of the current segment ("left")
            and the src index of the last byte in the previous segment (right - 1).
         */
+        if DEBUG { writeln("Using aggregation at the expense of memory"); stdout.flush(); }
         var srcIdx = makeDistArray(retBytes, int);
         srcIdx = 1;
         var diffs: [D] int;
@@ -201,7 +202,8 @@ module SegmentedArray {
         forall (v, si) in zip(gatheredVals, srcIdx) with (var agg = newSrcAggregator(uint(8))) {
           agg.copy(v, va[si]);
         }
-      } else {      
+      } else {
+        if DEBUG { writeln("Using unorderedCopy"); stdout.flush(); }
         ref va = values.a;
         // Copy string data to gathered result
         forall (go, gl, idx) in zip(gatheredOffsets, gatheredLengths, iv) {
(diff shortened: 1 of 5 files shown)


src/SegmentedArray.chpl
```diff
@@ -176,39 +176,44 @@ module SegmentedArray {
...
+        }
+      } else {
+        if DEBUG { writeln("Using unorderedCopy"); stdout.flush(); }
+        ref va = values.a;
+        // Copy string data to gathered result
+        forall (go, gl, idx) in zip(gatheredOffsets, gatheredLengths, iv) {
+          for pos in 0..#gl {
+            // Note: do not replace this unorderedCopy with aggregation
+            use UnorderedCopy;
+            unorderedCopy(gatheredVals[go+pos], va[oa[idx]+pos]);
```
ronawho: Is the reason for the more complicated aggregation above instead just using this code with aggregation because the remoteness of Src/Dst isn't consistent? i.e. if there was a general purpose CopyAggregator, would that allow you to use this simpler loop?

reuster986: Yes, this is a situation where either, both, or neither Src/Dst could be remote. If there were a general purpose `CopyAggregator` that figured out what to do based on the remoteness of Src and Dst, I would definitely use it here for a much simpler loop. The performance would probably improve, and the memory usage might even go down (the existing code can use up to 8x the input data in scratch memory).

PR #285, opened 2020-02-12
Remove throws from SipHash

ronawho:
It's no longer needed now that outlen is a param. This eliminates a conditional at the callsite, which has a slight performance improvement (189 -> 198 MB/s) for the following serial run:

    ./SipHashSpeedTest -nl 1 --NINPUTS=1_000_000 --INPUTSIZE=8 --dataParTasksPerLocale=1

See https://github.com/chapel-lang/chapel/pull/9380 for more information on error-handling overhead.  Note that error-handling overhead is generally negligible, but the hash kernel is called millions of times and is generally very fast, so minor overheads add up. For `--INPUTSIZE=1024` there is no noticeable performance difference.

### DIFF
--- a/src/SipHash.chpl
+++ b/src/SipHash.chpl
@@ -49,18 +49,18 @@ module SipHash {
     return c;
   }
   
-  proc sipHash64(msg: [] uint(8), D): uint(64) throws {
+  proc sipHash64(msg: [] uint(8), D): uint(64) {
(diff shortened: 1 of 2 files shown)


ronawho:
This peels a few minor commits off a different branch, which I should have a PR for soon.

ronawho:
Yeah, that would be my preference.

Issue #101, opened 2019-09-06
add reattach() function to python


reuster986:
Add a python function `pda = ak.reattach(name)` that will recreate the pdarray instance in python for an existing array in the server. This is useful for when the Jupyter kernel dies but the server stays running, in which case you want to re-populate your python namespace without redoing the chapel computations.

mhmerrill:
@reuster986 I would like your thoughts on this...
 1. What if we would like to share the same arrays between Jupyter notebooks or python scripts?
 1. Is this needed for intermediate results of mostly to save on reading arrays again from the filesystem?
 1. Should we make the array immutable when it is registered? Some users have stated this desire to keep themselves from overwriting the array and having to load data again.
 1. How do we manage big memory leaks potentially created by this capability?
 1. Should we get rid of the server defined name for the array when it is registered?
 1. Should we provide a cleanup function to cleanup all arrays which are not registered?

My current thinking is an interface which looks like this:
```python3
pda1 = ak.ones(100)
# register a user defined name for the pdarray in the arkouda_server
# return a new pdarray with the "user_defined_name" in the pda.name field
# on the server side we could link the new name to the array
# and bump the shared reference count
pda2 = ak.register(pda1, "user_defined_name")
# list registered names from the server
name_list = ak.get_registered_names()
# attach the array under the user defined name to a new python pdarray object
pda3 = ak.attach("user_defined_name")
# attach all arrays registered in the server and put in a dictionary
# using the user defined names as the keys
pda_dict = ak.attach_all()
# remove the registration of the name
# should unregister reset the name back to a server assigned name
pda4 = ak.unregister("user_defined_name")
```

reuster986:
@mhmerrill I have been thinking about this proposal, and I am somewhat ambivalent about taking this path. On the one hand, there is clear value in allowing users to save potentially lots of load time by re-using other people's data. On the other hand, this capability would break the "one server, one user" model, which is the cornerstone of our access control right now, i.e. the client's privileges are set to the server's privileges, because it is assumed that the client and server processes are run by the same user. Adding this feature to arkouda would, at the very least, require us to add a mechanism for the server to authenticate the client and for the server administrator to control access by various users. 

[...]

PR #290, opened 2020-02-13
Pda register feature

mhmerrill:
this provides initial support for `ak.register_pda` functionality
addresses #101


mhmerrill:
@reuster986 You are correct I did not mean to include both files. `src/Registration.chpl` has been removed.

> @mhmerrill sorry for the delay. It looks like `src/Registration.chpl` and `src/RegistrationMsg.chpl` are redundant (although only the latter is used in `src/MsgProcessing.chpl`). Did you mean to include both?

PR #291, opened 2020-02-13
Turn on memory tracking by default in the arkouda_server

mhmerrill:
added -smemTrack=true to Makefile


reuster986:
I can't merge this yet because the Mac CI job failed, and GitHub won't allow it.

PR #298, opened 2020-02-25
URGENT bug fix in SipHash

reuster986:
After the switch to using C pointers inside `computeSipHash`, there was an indexing error that caused the hashing function to ignore all but the first block and the partial block at the end (if input length was not a multiple of 8). This PR fixes the indexing and updates `test/UnitTestSipHash.chpl` to run with the new syntax. Before fixing the indexing bug, the updated `UnitTestSipHash` was failing, but now it passes.

### DIFF
--- a/src/SipHash.chpl
+++ b/src/SipHash.chpl
@@ -38,15 +38,15 @@ module SipHash {
             (p[D.low+7]: uint(64) << 56));
   }
 
-  private inline proc U8TO64_LE(p: c_ptr(uint(8)), D): uint(64) {
-    return ((p[D.low]: uint(64)) |
-            (p[D.low+1]: uint(64) << 8) |
-            (p[D.low+2]: uint(64) << 16) |
-            (p[D.low+3]: uint(64) << 24) |
-            (p[D.low+4]: uint(64) << 32) |
-            (p[D.low+5]: uint(64) << 40) |
-            (p[D.low+6]: uint(64) << 48) |
-            (p[D.low+7]: uint(64) << 56));
+  private inline proc U8TO64_LE(p: c_ptr(uint(8))): uint(64) {
+    return ((p[0]: uint(64)) |
+            (p[1]: uint(64) << 8) |
+            (p[2]: uint(64) << 16) |
+            (p[3]: uint(64) << 24) |
+            (p[4]: uint(64) << 32) |
+            (p[5]: uint(64) << 40) |
+            (p[6]: uint(64) << 48) |
+            (p[7]: uint(64) << 56));
   }
 
 
@@ -64,39 +64,19 @@ module SipHash {
   }
   
   proc sipHash64(msg: [] uint(8), D): uint(64) {
-    if contiguousIndices(msg) {
-      ref start = msg[D.low];
-      if D.high < D.low {
-        var res = computeSipHash(c_ptrTo(start), 0..#0, 8);
-        return res[1];
-      }
-      ref end = msg[D.high];
-      const startLocale = start.locale.id;
-      const endLocale = end.locale.id;
-      const hereLocale = here.id;
-      const l = D.size;
-      if startLocale == endLocale {
-        if startLocale == hereLocale {
-          var res = computeSipHash(c_ptrTo(start), 0..#l, 8);
-          return res[1];
-        } else {
-          var a = c_malloc(msg.eltType, l);
-          GET(a, startLocale, getAddr(start), l);
-          var h = computeSipHash(a, 0..#l, 8);
-          c_free(a);
-          return h[1];
-        }
-      }
-    }
-    var res = computeSipHash(msg, D, 8);
-    return res[1];
+    var h = computeSipHashLocalize(msg, D, 8);
+    return h[1];
   }
 
   proc sipHash128(msg: [] uint(8), D): 2*uint(64) {
+    return computeSipHashLocalize(msg, D, 16);
+  }
+  
+  proc computeSipHashLocalize(msg: [] uint(8), D, param outlen: int): 2*uint(64) {
     if contiguousIndices(msg) {
       ref start = msg[D.low];
       if D.high < D.low {
-        return computeSipHash(c_ptrTo(start), 0..#0, 16);
+        return computeSipHash(c_ptrTo(start), 0..#0, outlen);
       }
       ref end = msg[D.high];
       const startLocale = start.locale.id;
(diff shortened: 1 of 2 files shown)


reuster986:
@ronawho I just want to make sure I didn't do anything in this PR that will nerf performance/scaling or do dangerous things with the raw C arrays. Would you mind taking a look?

src/SipHash.chpl
```diff
@@ -99,11 +126,11 @@ module SipHash {
...
     var v2 = 0x6c7967656e657261: uint(64);
     var v3 = 0x7465646279746573: uint(64);
-    const k0 = 0xf0e1d2c3b4a59687: uint(64);
-    const k1 = 0x79695a4b3c2d1e0f: uint(64);
+    const k0 = 0x0706050403020100: uint(64);
+    const k1 = 0x0f0e0d0c0b0a0908: uint(64);
     var m: uint(64);
     var i: int;
-    const lastPos = D.low + D.size - (D.size % 8);
+    const lastPos = D.size - (D.size % 8); // C index, 0-up
```
ronawho: I think you only want to 0-up when msg is a c_ptr. When a segment spans 2 locales, msg will still be a block distributed array.

src/SipHash.chpl
```diff
@@ -64,13 +64,40 @@ module SipHash {
...
+    }
     var res = computeSipHash(msg, D, 8);
     return res[1];
   }
 
   proc sipHash128(msg: [] uint(8), D): 2*uint(64) {
     if contiguousIndices(msg) {
       ref start = msg[D.low];
+      if D.high < D.low {
+        return computeSipHash(c_ptrTo(start), 0..#0, 16);
```
ronawho: How can this happen?

ronawho:
Ah, shoot -- sorry about that. I worry a little bit about the overhead more complicated indexing in U8TO64_LE, because you really want this to generate specific assembly. I'd recommend something like:

```diff
arkouda (master) $ git diff src/SipHash.chpl
diff --git a/src/SipHash.chpl b/src/SipHash.chpl
index 4cd8c22..5968bc9 100644
--- a/src/SipHash.chpl
+++ b/src/SipHash.chpl
@@ -38,7 +38,7 @@ module SipHash {
             (p[D.low+7]: uint(64) << 56));
   }

-  private inline proc U8TO64_LE(p: c_ptr(uint(8)), D): uint(64) {
+  private inline proc U8TO64_LE(p: c_ptr(uint(8))): uint(64) {
     return ((p[0]: uint(64)) |
             (p[1]: uint(64) << 8) |
             (p[2]: uint(64) << 16) |
@@ -99,11 +99,12 @@ module SipHash {
     var v1 = 0x646f72616e646f6d: uint(64);
     var v2 = 0x6c7967656e657261: uint(64);
     var v3 = 0x7465646279746573: uint(64);
-    const k0 = 0xf0e1d2c3b4a59687: uint(64);
-    const k1 = 0x79695a4b3c2d1e0f: uint(64);
+    const k0 = 0x0706050403020100: uint(64);
+    const k1 = 0x0f0e0d0c0b0a0908: uint(64);
     var m: uint(64);
     var i: int;
-    const lastPos = D.low + D.size - (D.size % 8);
+    const lastPos = if isSubtype(msg.type, c_ptr) then D.size - (D.size % 8)
+                                                  else D.low + D.size - (D.size % 8);
     // const uint8_t *end = in + inlen - (inlen % sizeof(uint64_t));
     const left: int = D.size & 7;
     // const int left = inlen & 7;
@@ -146,7 +147,11 @@ module SipHash {
     }

     for pos in D.low..lastPos-1 by 8 {
-        m = U8TO64_LE(msg, pos..#8);
+        if isSubtype(msg.type, c_ptr) {
+          m = U8TO64_LE(msg + pos);
+        } else {
+          m = U8TO64_LE(msg, pos..#8);
+        }
         v3 ^= m;
         TRACE();
         for i in 0..#cROUNDS {
```

ronawho:
(It would be great to run unit tests as part of CI testing -- I think that's captured between https://github.com/mhmerrill/arkouda/issues/214 and https://github.com/mhmerrill/arkouda/issues/217 though)

src/SipHash.chpl
ronawho: I think you can do that with something like:

```chpl
const lastPos = if isSubtype(msg.type, c_ptr) then D.size - (D.size % 8)
                                              else D.low + D.size - (D.size % 8);
```

ronawho:
The part about only 0-upping for c_ptr is the most important thing. I'm happy to do some quick perf tests on this then.

src/SipHash.chpl
```diff
@@ -64,13 +64,40 @@ module SipHash {
...
+          return res[1];
+        } else {
+          var a = c_malloc(msg.eltType, l);
+          GET(a, startLocale, getAddr(start), l);
+          var h = computeSipHash(a, 0..#l, 8);
+          c_free(a);
+          return h[1];
+        }
+      }
+    }
```
ronawho: Nit, but instead of duplicating the logic, I'd probably add a function like:

```chpl
private proc computeSipHashLocalize(msg, D, param outlen: int)
```

or whatever that handles the c_ptr/localization logic and just call that from `sipHash64`/`sipHash128`

src/SipHash.chpl
ronawho: Oh, actually -- for the c_ptr case, `D` is `0..#l`, so I think the 0-up is happening at the callsite.

src/SipHash.chpl
ronawho: Ah, I see when you're passing `0..#0` to it.

ronawho:
@reuster986 https://github.com/ronawho/arkouda/tree/sip-hash-bug-fix includes my suggestions. Feel free to grab that if it looks ok to you. I have verified there are no performance regressions on 32 node IB.

src/SipHash.chpl
reuster986: ah, thanks, i didn't recognize that.

src/SipHash.chpl
reuster986: good idea!

reuster986:
I like your suggestions. I incorporated them and tested for correctness (`UnitTestSipHash` passes) and just pushed.

src/SipHash.chpl
```diff
@@ -99,11 +106,12 @@ module SipHash {
...
     var v3 = 0x7465646279746573: uint(64);
-    const k0 = 0xf0e1d2c3b4a59687: uint(64);
-    const k1 = 0x79695a4b3c2d1e0f: uint(64);
+    const k0 = 0x0706050403020100: uint(64);
+    const k1 = 0x0f0e0d0c0b0a0908: uint(64);
     var m: uint(64);
     var i: int;
-    const lastPos = D.low + D.size - (D.size % 8);
+    const lastPos = if isSubtype(msg.type, c_ptr) then D.size - (D.size % 8)
+      else D.low + D.size - (D.size % 8); // if C index, 0-up
```
ronawho: I think you can drop this change since we 0-up at the call site by passing in `D=0..#l`

reuster986: ah, cool. I will just pull in your changes, i think, since you have already tested them.

PR #310, opened 2020-03-03
Remove mpi dependency for chapel-master

ronawho:
The default hdf5 module no longer brings in mpi as of chapel-lang/chapel#15018


ronawho:
MPI was never required for 1.20. It was only required on master since https://github.com/chapel-lang/chapel/pull/14463

PR #312, opened 2020-03-03
Start on automated unit testing for Chapel tests

ronawho:
Set up infrastructure to run Chapel unit tests with `start_test`.
start_test is the mainline Chapel testing infrastructure. It's pretty
sophisticated, but I think the use-case here will be pretty simple.

The basic usage is that you need a test.chpl file and a test.good file,
where the .good file contains the expected output. I've set up a PREDIFF
(before diffing .good files, run this script) that will create an empty
.good file if one doesn't exist so that if most tests have no output you
don't have bother creating the .good file.

For now I've set up CI testing to only run UnitTestSipHash.chpl, but
over time we can get more tests working and run them all.

Part of https://github.com/mhmerrill/arkouda/issues/217


ronawho:
You can see the runs here:
 - comm=none: https://github.com/mhmerrill/arkouda/pull/312/checks?check_run_id=483586915#step:5:1
 - comm=gasnet: https://github.com/mhmerrill/arkouda/pull/312/checks?check_run_id=483586922#step:5:1


There's way more information about the Chapel testing infrastructure [here](https://github.com/chapel-lang/chapel/blob/master/doc/rst/developer/bestPractices/TestSystem.rst), but I believe the use-case here is pretty simple.

BryantLam:
This is fine, though I'd prefer an approach where we could eventually move everything to Mason. E.g., what features are provided in `start_test` or this PR that couldn't be implemented using Mason? I'd like those gaps remedied.

ronawho:
I'm not very familiar with mason, but I took a stab at the mason approach and the current shortcomings (which @ben-albrecht has also requested I document so we can address them) are:
 - Not sure how to default to using multiple locales for comm!=none configuration
 - Not sure how to get the compopts to find zmq/hdf5 dependencies and other arkouda files
 - `mason test` is best for mason projects, which arkouda currently isn't

My hope is that over time we can convert arkouda to a mason project, use `mason external` to install hdf5/zmq dependencies, and use `mason test` for Chapel unit tests.

ben-albrecht:
> Not sure how to default to using multiple locales for comm!=none configuration

```chpl
UnitTest.minLocales(N);
// or
UnitTest.addNumLocales(N);

// +

UnitTest.skipIf(CHPL_COMM=='none', "This test only supports multilocale config");
```


> Not sure how to get the compopts to find zmq/hdf5 dependencies and other arkouda files

This unfortunately requires the project already follow a mason project format. We will need to extend the non-mason-project support in the future (maybe arkouda can be a motivating use-case).

ronawho:
> UnitTest.minLocales(N);
// or
UnitTest.addNumLocales(N);
// +
UnitTest.skipIf(CHPL_COMM=='none', "This test only supports multilocale config");

Right, but that's not a default, you have to specify that for every unit test.

ronawho:
Assuming @mhmerrill and @reuster986 are on board with this, I'd probably aim for merging as-is and leaving you guys to update unit tests to get them to compile and settle on the right expected output.

ben-albrecht:
> Right, but that's not a default, you have to specify that for every unit test.

@ronawho - understood.

I have opened these issues to track these feature requests in the meantime:

- https://github.com/chapel-lang/chapel/issues/15077
- https://github.com/chapel-lang/chapel/issues/15078

BryantLam:
@ben-albrecht -- it would make sense to move Arkouda to use Mason as soon as possible. chapel-lang/chapel#15078 seems unnecessary for that goal unless I'm missing something.

Also, does it make sense to have a feature request for `Mason.toml` to e.g., allow arbitrary specification of more tests to run that might not be Chapel programs? I'm thinking something along the lines of:
```terminal
# Mason.toml
[[test]]
py_check = run_python.bash check.py
```
(but don't use that since I'm glossing over many assumptions to show an example.)

ben-albrecht:
> it would make sense to move Arkouda to use Mason as soon as possible. chapel-lang/chapel#15078 seems unnecessary for that goal unless I'm missing something.

Noted, thanks. I will leave it open for now since other users have requested this feature as well.

> Also, does it make sense to have a feature request for Mason.toml to e.g., allow arbitrary specification of more tests to run that might not be Chapel programs?

I am not sure. Allowing arbitrary commands for mason tests could be incredibly powerful, but could also be a massive undertaking depending on how much we leave on the user. I would want to understand the example you had in mind more to help determine if it makes sense to include in mason. 

I encourage you to open an issue on it so we can discuss further (or we can continue here). Do you have a specific motivating use-case?

PR #317, opened 2020-03-06
Start on automated unit testing for Arkouda python tests

ronawho:
Set up infrastructure to run Arkouda python unit tests with the
`runClient` wrapper. Success/failure is currently solely determined by
the exit code. For now I'm only running operator_tests and string_test,
but eventually I expect that to  be `./util/test/runClient tests/*.py`

I'm not familiar with python unit testing best practices, but I'm not
sure if they are applicable here since you need to start a server first.
You might be able to get away with normal python unit tests that have
`setUp` and `tearDown` routines to start/stop the server, but for now
this was relatively straightforward.

Part of https://github.com/mhmerrill/arkouda/issues/217


ronawho:
Example run here: https://github.com/mhmerrill/arkouda/pull/317/checks?check_run_id=490881118#step:7:1

mhmerrill:
@ronawho @reuster986 I think we should run all of the python package tests from `make check`
My thinking on this is that `make` builds the arkouda server so `make check` should run all the tests against the server. looks like you might add `tests/check.py` to the `runClient` thing and change the `makefile` to use `runClient` or something like that... what do you guys think???

ronawho:
I tend to see `make check` as a quick sanity check, and `make test` or something for running more comprehensive tests, but I'm happy to organize it however you'd like.

mhmerrill:
`make test` currently runs the chapel unit tests from the `arkouda/test/` directory. While `make check`  (correstness) runs the python client/server tests in the `arkouda/tests/` directory. I do think this is confusing but we were aiming at one point to integrate with Mason which uses the `test/` directory by default, I think, perhaps we should rename `arkouda/test/` to `arkouda/unittests/`. @ronawho your opinion is valid here but I think starting building and starting a server is a bigger chunk of the time than running the python tests against the server, unless this hits against the multi-locale testing.

Issue #324, opened 2020-03-13
look into interactionand order of application of python operator overloading


mhmerrill:
we are currently not understanding the order of application of the python operator overloading scheme. a key to this understanding is when to return python's `NotImplemented`
* python odering of which operator overloads are used 
* python seems to call an objects iterator in some cases even when there a `__radd__` method defined
* __new__: seems to only happen when using NumPy scalars as the first operand
```python
a = ak.ones(100)
b = np.int64(5) + a
# python seems to call a's iterator instead of the __radd__ method
# b should be a pdarray but is not
```
* we will have to disable the pdarray iterator to fix this... I think

mhmerrill:
@hokiegeek2 (hope I got your username correct) could you look at this issue?

hokiegeek2:
> @hokiegeek2 (hope I got your username correct) could you look at this issue?

Will do!

mhmerrill:
@hokiegeek2 here are some pointers to the relevant code in the arkouda python package.
https://github.com/mhmerrill/arkouda/blob/55ef6f7cec9b763e75c47a17679ce3eb603004a8/arkouda/pdarrayclass.py#L33
https://github.com/mhmerrill/arkouda/blob/55ef6f7cec9b763e75c47a17679ce3eb603004a8/arkouda/pdarrayclass.py#L107-L145

mhmerrill:
here is a stack overflow question which borders on our need i think
https://stackoverflow.com/questions/22289093/can-you-impose-object-precedence-for-overloaded-operators-in-python

hokiegeek2:
@mhmerrill 🤘

mhmerrill:
@hokiegeek2 New discovery: this problem occurs when a NumPy scalar is used. NumPy appears to call the `pdarray.__iter__()` method. Bill and I currently think we should remove the iterator and force the user to use the `pdarray.to_ndarray()` method

hokiegeek2:
@mhmerrill Yup, running this through debugger shows that when adding a simple Python int, pdarray.__radd__ is called while adding a numpy.int64 results in the pdarray.__iter__ being called.

![image](https://user-images.githubusercontent.com/10785153/77323852-44c40480-6cec-11ea-8e0d-33ed9b204ff2.png)

![image](https://user-images.githubusercontent.com/10785153/77323923-5f967900-6cec-11ea-8fbc-50d02630c031.png)

[...]

PR #325, opened 2020-03-23
fixed issue with non-commutative addition operator  behavior #324

hokiegeek2:
Contains fix for non-commutative behavior contained inherent in issue #324 and added unit tests to confirm fix.


tests/pdarray_test.py
```diff
@@ -0,0 +1,108 @@
+import unittest
+import arkouda as ak
+import numpy as np
+import subprocess, os
+
+class PdArrayTest(unittest.TestCase):
+    
+    @classmethod
+    def setUpClass(cls):
+        try:
```
ronawho: There's some functionality in util/test/util.py for starting/stopping the server and getting host/port information that might be useful here. I think you could do something like:

```python
# probably not the right way to do this..
sys.path.append(os.path.join(os.path.dirname(__file__), '..', 'util', 'test'))
from util import *

...

    def setUpClass(cls):
        start_arkouda_server(get_arkouda_numlocales())

    def setUp(self):
        server_info = get_server_info()
        ak.client.connect(server=server_info.host, port=server_info.port)

...

    @classmethod
    def tearDownClass(cls):
        stop_arkouda_server()
```

That would allow this unit test to work for multi-locale (multi node) configurations. 

There's also a `runClient` wrapper that starts/stops the server and runs client programs that was also meant to be used for unit testing (https://github.com/mhmerrill/arkouda/pull/317), but I'm really not  familiar with python unit testing best practices to know if that was a good approach. I'd be interested in your thoughts on the right way to write/run Arkouda unit tests. The one consideration is that ideally we only want want to start the server once and then run all the unit tests (since queueing times on large clusters can be slow)

hokiegeek2: @ronawho Excellent points, I can integrate in the util methods you mentioned in the setUpClass and tearDownClass methods, which starts up/shuts down the arkouda server for each test suite. Basically each current test file would correspond to one test suite.

mhmerrill:
@hokiegeek2 delete these lines to remove the pdarray iterator https://github.com/mhmerrill/arkouda/blob/55ef6f7cec9b763e75c47a17679ce3eb603004a8/arkouda/pdarrayclass.py#L370-L374

tests/pdarray_test.py
```diff
@@ -0,0 +1,108 @@
...
+        except KeyError:
+            raise EnvironmentError(('The ARKOUDA_SERVER_PATH env variable output from the ' + 
+                                   '"which arkouda_server" command must be set'))
+        
+        PdArrayTest.ak_server = subprocess.Popen([arkouda_path, '--ServerPort=5566', '--quiet'])
+    
+    def setUp(self):
+        ak.client.connect(port=5566)
+      
+    def testPdArrayAddInt(self):
```
mhmerrill: @hokiegeek2 please add test cases to `tests/operator_tests.py` to check for the numpy scalar cases

hokiegeek2: @mhmerrill sure thing! on it now.

tests/pdarray_test.py
hokiegeek2: @ronawho Thanks! the util.test.util.get_arkouda_server works like a champ -> thanks for the tip!

PR #330, opened 2020-03-30
Optimize coargsort for numeric arrays

ronawho:
Optimize coargsort for int/real arrays when the combined elements require 32-bytes or less (e.g. 8 int arrays with values < 2**32). This results in a ~2x improvement for sorting 2 int arrays.

The previous coargsort would call argsort on each array, and then gather and merge the resulting index vectors. Here we merge all the arrays into a single array and call argsort once. This eliminates gathering the index vector, but does increase the peak memory footprint and the total number of bytes being communicated. Because of that, we only merge when the arrays will need less than 32-bytes of data. With more than that, the memory overhead is quite high and the performance is about on par.  The performance advantage of the merged sort starts to suffer because we exchange our KEY as part of sorting, and our KEY is now larger, which increases the total size of comm. We can probably optimize that, but we don't currently. Note that because sorting strings uses 16-bytes for the hash, this does not currently trigger for strings.

Performance results (256 node XC, Chapel 1.20):
---

```
./arkouda_server --logging=false --v=false -nl 256
./benchmarks/argsort.py --size=$((2**28))
./benchmarks/coargsort.py --size=$((2**28))
```

Results for argsort vs coargsort on a single array with 4 bytes values:

| config | argsort    | coargsort  |
| before | 48.6 GiB/s | 30.7 GiB/s |
| after  | 48.6 GiB/s | 46.2 GiB/s |

Here we can see that coargsort on a single array now has minimal overhead compared to argsort.

Results for coargsort on 2, 4, 8, and 16 arrays with 4 byte values:

| config | 2 arrays   | 4 arrays   | 8 arrays   | 16 arrays  |
| ------ | ---------: | ---------: | ---------: | ---------: |
| before | 24.3 GiB/s | 18.6 GiB/s | 12.6 GiB/s |  9.9 GiB/s |
| after  | 42.4 GiB/s | 28.0 GiB/s | 13.7 GiB/s |  9.9 GiB/s |

Here performance for sorting 2/4 arrays is significantly improved. 8 arrays has a minimal perf advantage because of the increased transfers sizes. 16 arrays of 4-byte values is over the current size limit for the optimization so that falls back to the old strategy so there is no perf change.

Implementation notes:
---

The basic algorithm is that we go through all the arrays to coargsort and compute how many bits/digits are needed using `getBitWidth()` (where each digit is 16-bits.) We then iterate through each array and use `getDigit()` to pluck out each digit and put it into the merged array.  This wastes some space (e.g. if you're merging 7 and 9 bit arrays we end up using 32 bits instead of 16) but this makes merging simpler and getDigit will modify the bits so we correctly sort negatives. The merged array is a tuple of digits (tuple of `uint(16)`). Since we have to know the size of tuples at compile time we stamp out versions of merging/sorting for 4, 8 and 16 byte merged arrays. Longer term we might want to use a dynamic array-of-arrays of segmented arrays, but to start the tuple is simplest to implement. Note that tuple[1] is the most significant digit (tuple[0] in 1.22 -- the lowest index is the most significant digit)

Part of https://github.com/mhmerrill/arkouda/issues/296

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -13,6 +13,9 @@ module RadixSortLSD
     private param numBuckets = 1 << bitsPerDigit; // these need to be const for comms/performance reasons
     private param maskDigit = numBuckets-1; // these need to be const for comms/performance reasons
 
+    // Hack to determine tuple lower bound across 1.20-1.22
+    const RSLSD_tupleLow = [1..1].domain.low;
+
     use BlockDist;
     use BitOps;
     use AryUtil;
@@ -52,9 +55,9 @@ module RadixSortLSD
 
     inline proc getBitWidth(a: [?aD] ?t): (int, bool)
         where isHomogeneousTuple(t) && t == t.size*uint(bitsPerDigit) {
-      for digit in 1..t.size {
-        const m = max reduce [ai in a] ai[digit];
-        if m > 0 then return ((t.size-digit+1) * bitsPerDigit, false);
+      for digit in 0..t.size-1 {
+        const m = max reduce [ai in a] ai[RSLSD_tupleLow+digit];
+        if m > 0 then return ((t.size-digit) * bitsPerDigit, false);
       }
       return (t.size * bitsPerDigit, false);
     }
@@ -99,11 +102,11 @@ module RadixSortLSD
       }
     }
 
-    inline proc getDigit(key: _tuple, rshift: int, last: bool, negs: bool): int 
+    inline proc getDigit(key: _tuple, rshift: int, last: bool, negs: bool): int
         where isHomogeneousTuple(key) && key.type == key.size*uint(bitsPerDigit) {
-      return key[key.size - rshift/bitsPerDigit];
+      const keyHigh = key.size - (1-RSLSD_tupleLow);
+      return key[keyHigh - rshift/bitsPerDigit]:int;
     }
-
 
     // calculate sub-domain for task
     inline proc calcBlock(task: int, low: int, high: int) {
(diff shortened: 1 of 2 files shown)


src/RadixSortLSD.chpl
```diff
@@ -85,6 +98,12 @@ module RadixSortLSD
       }
     }
 
+    inline proc getDigit(key: _tuple, rshift: int, last: bool, negs: bool): int 
+        where isHomogeneousTuple(key) && key.type == key.size*uint(bitsPerDigit) {
+      return key[key.size - rshift/bitsPerDigit];
```
reuster986, line 103: This assumes 1-up indexing, right? How could we make this agnostic to indexing?

ronawho: Yeah, this and 2 other places (`getBitWidth`, and `var curDigit = 1 + ...`) assume 1-up and will have to be adjusted.

mhmerrill: @bradcray made changes like these in some other PRs like PR #351 and #352. Do we need to change this for 1.21 and 1.22 compatibility?

ronawho: Yeah, we'll need to do something to make it compatible with 1.22. I was leaning towards something like https://github.com/mhmerrill/arkouda/pull/352. 

I'll rebase and make that change tomorrow.

ronawho: Rebased and made 1.20-1.22 compatible. I regathered performance numbers with the new coargsort.py benchmark and included them in the description.

PR #331, opened 2020-03-31
Remove Chapel 1.20 compatibility

e-kayrakli:
This is a draft PR that can be merged once Arkouda stops supporting Chapel 1.20.

- Removes `src/Chapel120.chpl` and its uses from other modules
- Removes the logic in Makefile to detect the version and use appropriate flags.

I expect some CI tests to fail.

Test:
- [x] builds on Chapel master today and `check.py` passes


reuster986:
We just merged #372 dropping support for chapel 1.20 and older, so it might be worth pulling master on this PR and running the checks again.

mhmerrill:
@reuster986 are we really ready for this yet?

reuster986:
@mhmerrill I was convinced by @ronawho 's reasoning on #372 that we should require >= 1.21, mostly for performance, and I thought it would probably be a good idea to make that change before CHIUW so that people checking it out after the workshop wouldn't get caught in the middle. The test infrastructure that @hokiegeek2 and @ronawho have built has given me a lot more confidence about a change like that. 

Regarding this PR, I think seems like a logical cleanup after #372 , but I did want the CI checks to run after pulling master, and it looks like one of them is failing with an error in the chapel code:

```
/__w/arkouda/arkouda//src/arkouda_server.chpl:61: error: Type "bytes" is not serializable by ZMQ
make: *** [Makefile:157: arkouda_server] Error 1
```

So that should definitely be resolved first, of course.

e-kayrakli:
I rebased the branch and I think it is ready for a review.

PR #334, opened 2020-04-02
Make arkouda neutral to 0- vs. 1-based tuple indexing

bradcray:
This PR changes Arkouda to avoid relying on 0- or 1-based indexing for tuples.  The motivation for this is that Chapel 1.22 will switch from 1-based tuple indexing to 0-based tuple indexing.  This PR uses Chapel's de-tupling syntax to avoid baking either 0- or 1-based assumptions about indexing into the code, making it compatible with Chapel 1.20, 1.21, or 1.22.

I didn't always have good intuition for what the meaning of the variables or tuples I was working with were, so this PR could almost certainly benefit from naming improvements.  I'd be happy to implement other name changes if someone wants to provide guidance.

[I should also note that I haven't done any performance testing of this branch; I'd like to think these changes will not impact performance significantly in either direction, but just wanted to make clear that I haven't checked]

### DIFF
--- a/src/ReductionMsg.chpl
+++ b/src/ReductionMsg.chpl
@@ -667,7 +667,7 @@ module ReductionMsg
           vi = values.domain.high;
         }
         if (vi >= low) {
-          r = cummin[vi][1];
+          (_,r) = cummin[vi];
         }
       }
       return res;
@@ -704,7 +704,7 @@ module ReductionMsg
           vi = values.domain.high;
         }
         if (vi >= low) {
-          r = cummax[vi][1];
+          (_,r) = cummax[vi];
         }
       }
       return res;
@@ -742,8 +742,8 @@ module ReductionMsg
           vi = values.domain.high;
         }
         if (vi >= low) {
-          v = cummin[vi][0][1];
-          l = cummin[vi][1];
+          ((_,v),_) = cummin[vi];
+          (_    ,l) = cummin[vi];
         }
       }
       return (vals, locs);
@@ -784,8 +784,8 @@ module ReductionMsg
           vi = values.domain.high;
         }
         if (vi >= low) {
-          v = cummax[vi][0][1];
-          l = cummax[vi][1];
+          ((_,v),_) = cummax[vi];
+          (_,    l) = cummax[vi];
         }
       }
       return (vals, locs);
@@ -972,15 +972,18 @@ module ReductionMsg
        agg.copy(IVi, firstIV[idx]);
       }
       var sortedKV: [kD] (int, int);
-      forall (kvi, idx) in zip(sortedKV, IV) with (var agg = newSrcAggregator(int)) {
-        agg.copy(kvi[0], keys[idx]);
-        agg.copy(kvi[1], values[idx]);
+      forall ((kvi0,kvi1), idx) in zip(sortedKV, IV) with (var agg = newSrcAggregator(int)) {
+        agg.copy(kvi0, keys[idx]);
+        agg.copy(kvi1, values[idx]);
       }
       if v {writeln("sort time = ", Time.getCurrentTime() - t1); try! stdout.flush();}
       if v {writeln("Finding unique (key, value) pairs..."); try! stdout.flush();}
       var truth: [kD] bool;
       // true where new (k, v) pair appears
-      [(t, kv, i) in zip(truth, sortedKV, kD)] if i > kD.low { t = (sortedKV[i-1][1] != kv[1]); }
+      [(t, (_,val), i) in zip(truth, sortedKV, kD)] if i > kD.low {
+        const (_,sortedVal) = sortedKV[i-1];
+        t = (sortedKV[i-1][1] != val);
+      }
       // first value of every segment is automatically new
       [s in segments] truth[s] = true;
       // count cumulative new values and take diffs at segment boundaries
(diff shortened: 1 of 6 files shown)


bradcray:
Hi @mhmerrill / @reuster986 : I suspect you guys might be slammed these days, but I wanted to bring this PR to your attention.  If you're OK with this approach, it would be nice to merge so that Arkouda will keep working in our testing against Chapel master (and eventually the 1.22 release).  If you like the general approach but would like to see modifications, including renamings before merging, let me know and I'll make the adjustments.

mhmerrill:
@bradcray I am ok with the general approach. I didn't know about the syntax you used ;-).
I would like for @reuster986 to check the code because he wrote most of it.
@bradcray you can test most of the functions you changed by running a couple of tests in the `tests/` directory namely
https://github.com/mhmerrill/arkouda/blob/master/tests/groupby_test.py
and
https://github.com/mhmerrill/arkouda/blob/master/tests/string_test.py

bradcray:
> you can test most of the functions you changed by running a couple of tests in the tests/ directory

Thanks... I had run `checks.py` and then hoped that the automated testing here would do everything else.  `./string_test.py` seems to pass.  However, `groupby_test.py` gets 1 error.  If I'm understanding the output correctly, I think this is the failing cases:

```
Doing aggregate(int64, nunique)
Different values (abs diff = 16081)
```

If you have tips which of the diffs might be involved that may save me some time.

mhmerrill:
@bradcray I think that test got through this code path...
https://github.com/mhmerrill/arkouda/blob/c41b4026a4d57267b4c414a75573573816a95c21/src/ReductionMsg.chpl#L310-L313
https://github.com/mhmerrill/arkouda/blob/c41b4026a4d57267b4c414a75573573816a95c21/src/ReductionMsg.chpl#L953-L1035

src/ReductionMsg.chpl
```diff
@@ -972,15 +972,18 @@ module ReductionMsg
        agg.copy(IVi, firstIV[idx]);
       }
       var sortedKV: [kD] (int, int);
-      forall (kvi, idx) in zip(sortedKV, IV) with (var agg = newSrcAggregator(int)) {
-        agg.copy(kvi[1], keys[idx]); 
-        agg.copy(kvi[2], values[idx]);
+      forall ((kvi0,kvi1), idx) in zip(sortedKV, IV) with (var agg = newSrcAggregator(int)) {
```
mhmerrill, line 975: one of these two changes to this code is the current problem

bradcray:
OK, thanks.  Those were some of my more convoluted changes, so I suppose I shouldn't be surprised that I messed something up.  Taking a look.

src/ReductionMsg.chpl
```diff
@@ -972,15 +972,18 @@ module ReductionMsg
...
+        agg.copy(kvi0, keys[idx]);
+        agg.copy(kvi1, values[idx]);
       }
       if v {writeln("sort time = ", Time.getCurrentTime() - t1); try! stdout.flush();}
       if v {writeln("Finding unique (key, value) pairs..."); try! stdout.flush();}
       var truth: [kD] bool;
       // true where new (k, v) pair appears
-      [(t, kv, i) in zip(truth, sortedKV, kD)] if i > kD.low { t = (sortedKV[i-1][2] != kv[2]); }
+      [(t, (_,val), i) in zip(truth, sortedKV, kD)] if i > kD.low {
```
mhmerrill, line 983: here is the other change which might have caused the error

bradcray: Aha, with another look, it looks like I made a transcription error here (note that sortedVal isn't used).  Testing a fix now.

bradcray: Got a fix for 1.20, but still don't have it passing with 1.22.  Timing out for tonight, will try again tomorrow.

bradcray: Latest info: It seems to be a bug of omission rather than a problem with any of the changes that were made.  This hypothesis is based on stepping back to the first commit on this branch in which I merely changed from 1-based to 0-based indexing for Chapel master, and still see the same failures.  So now to find the tuple access I overlooked...

bradcray: OK, I think I've got it now.  I had overlooked one tuple indexing case unfortunately.

bradcray:
@mhmerrill: https://github.com/mhmerrill/arkouda/pull/334/commits/5915068db44a30c9592fa57209ce583ee3f3c840 fixed the `groupby_test.py` failures I was seeing, but doing some more stress testing, I'm seeing `string_test.py` failing about 1/8 times, so I'm going to look more into that.  Failure modes seem to be either:

```
Generation and concatenate passed
  File "./string_test.py", line 46, in <module>
    assert(strings[N//3] == test_strings[N//3])
```

or an apparent hang after:

```
endswith passed
```

bradcray:
Ah, @ronawho points out that he has seen similar instability for the strings test with 1.20 and master:  https://github.com/mhmerrill/arkouda/pull/320#issuecomment-596625401  So maybe what I'm seeing is independent of this PR.

bradcray:
Update: I'm seeing both failure modes for string_test.py with Chapel 1.20 and Arkouda master as well, so I think it's independent of this PR.

PR #343, opened 2020-04-07
Add a noop benchmark to track client<->server latency

ronawho:
It feels a little weird to add a no-op to the server/client, but I
wanted some way to measure round-trip latency and this was the best I
came up with. This is primarily motivated by wanting to see latency
between the server and client in nightly performance testing, but is
also motivated by wanting a way to track the latency of ssh tunneling
instead of running directly on login nodes.


ronawho:
If you don't like adding a noop, I think we could also use `getConfig()` it'll include more than just round-trip latency, but it's probably more representative of the "minimum message size" for arkouda operations.

Issue #214, opened 2019-12-10
some chapel unit tests need to be fixed to use exception handling


mhmerrill:
some chapel unit tests no longer compile

PR #351, opened 2020-04-10
Make Arkouda's use of split() calls index-neutral

bradcray:
By way of introduction:  This started out as a branch in which I got Arkouda working with next week's Chapel 1.22 release (whose theme is switching from 1- to 0-based indexing) where the main stumbling block is with the `var fields = reqMsg.split(...);` pattern.  Getting a 1.22-compatible version of Arkouda working was fairly simple.  Then I decided to push harder on the index-neutral theme that I started in #334 to see if I could create something similar for split() that would work with Chapel 1.20, 1.21, or 1.22.  This turned out to be a fair amount more challenging, though it could've been less challenging to give up on 1.20 and just focus on 1.21 and 1.22 (or, simplest of all, to just focus on 1.22).

So, I'd suggest approaching this PR with the mindset of "Is the style taken here one that we like and would like to adopt in Arkouda?" (in which case it _may_ be worth merging) or "This is all well and good, but let's just change over to using 0-based indexing and stick with the previous style" (in which case we should close this PR and I'll open one for one of the more straightforward branches).

The approach I took here was to create variations of `split()` which would return a tuple so that I could use Chapel's de-tupling syntax to grab the fields out of them.  Thus, code that used to look like this:

```chapel
var fields = reqMsg.split(...);
var a = fields[1];
var b = fields[2];
var c = fields[3];
```

now looks like this:

```chapel
var (a, b, c) = reqMsg.splitMsgToTuple(...);
```

(where the name `splitMsgToTuple()` is completely open to negotiation).  Cases that require casting to different types I've written as:

```chapel
var (aStr, bStr, cStr) = reqMsg.splitMsgToTuple(...);
var a = aStr: int;
var b = bStr: real;
var c = cStr: bool;
```

And cases that are varargs-like I've written as:

```chapel
var (a, b, rest) = reqMsg.splitMsgToTuple(...);
var fields = rest.split();
...play some tricks with fields.domain.low...
```

The routines are defined at the bottom of ServerConfig.chpl though there may be a more appropriate place for them, and I'd be happy to move them.  It took me a few tries to get them right where the main challenge related to getting the behavior right for splits that had too many or two few fields.

Have I broken anything?  I've tested this by running `tests/check.py`, `tests/groupby_test.py`, and `tests/string.py`, but the last doesn't always work as Elliot has pointed out [elsewhere](https://github.com/mhmerrill/arkouda/pull/320#issuecomment-596625401), so while I have seen it work under this change, I'm not fully confident that I haven't broken it in new ways.  I also worry about code paths that aren't covered by the above tests.

I started on a similar effort to update the unit test directory, before realizing that many of them currently don't work on Arkouda master either (alluded to by issue #214).  So this PR does not contain similar updates for them, though I have a start on it in my local directory if Arkouda did want to continue in this vein.

### DIFF
--- a/pydoc/contributing.rst
+++ b/pydoc/contributing.rst
@@ -136,9 +136,8 @@ Then, define your argument parsing and function logic in ``src/FooMsg.chpl`` in 
        */
        proc FooMsg(reqMsg: string, st: borrowed SymTab): string {
            var repMsg: string; // response message
-           var fields = reqMsg.split(); // split request into fields
-           var cmd = fields[1];
-           var name = fields[2];
+           // split request into fields
+           var (cmd, name) = reqMsg.splitMsgToTuple(2);
            // get next symbol name
            var rname = st.nextName();
         
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -434,12 +434,12 @@ module ArgSortMsg
     proc coargsortMsg(reqMsg: string, st: borrowed SymTab) throws {
       param pn = Reflection.getRoutineName();
       var repMsg: string;
-      var (cmd, nstr, _) = reqMsg.splitMsgToTuple(3);
+      var (cmd, nstr, rest) = reqMsg.splitMsgToTuple(3);
       var n = nstr:int; // number of arrays to sort
-      var fields = reqMsg.split();
+      var fields = rest.split();
       // Check that fields contains the stated number of arrays
-      if (fields.size != (2*n + 2)) { return try! incompatibleArgumentsError(pn, "Expected %i arrays but got %i".format(n, fields.size/2 - 1)); }
-      const low = fields.domain.low + 2;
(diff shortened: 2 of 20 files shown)


bradcray:
Side notes:  I believe that the "not all strings" case ought to be able to be written as:

```chapel
var (aStr, bStr, cStr) = reqMsg.splitMsgToTuple(...): (int, real, bool);
```

but this seemed to trip over a bug that didn't seem familiar to me in Chapel.  I considered fixing it to make use of this idiom before realizing that it wouldn't help with the 1.20/1.21 compatibility theme of this PR...

Also, there are some slightly unfortunate choices made in the implementation of my split routines to keep them working with 1.20, and they could be written much more cleanly once we turn our backs on 1.20 and focus on 1.21 and later.

bradcray:
@mhmerrill, @reuster986: For your consideration.

PR #355, opened 2020-04-15
Make MyDmap an enum

ronawho:
Use an enum for MyDmap instead of hard-coding integer literals.


ronawho:
Minor cleanup in preparation for defaulting to using DefaultRectangular (local array) for comm=none.

PR #358, opened 2020-04-21
Improve comm=none build and execution speed

ronawho:
This is primarily motivated by faster build/compile times to make development easier, but there are some nice execution performance improvements too. When building locally with comm=none and setting `ARKOUDA_QUICK_COMPILE=true` my build times drop from 9 minutes to 3 minutes. This makes a few changes: 

 - Default to DefaultRectangular (local array) for comm=none. BlockDist has some overheads that are unnecessary for comm=none builds (block dist has locality checks, loops over the locales array, and other things that aren't needed for comm=none.) We're looking to optimize this in the future, but in the meantime this will improve compilation and execution performance for single node builds.
 - Add `ARKOUDA_SKIP_CHECK_DEPS` to skip zmq/hdf5 dependency checks. This saves a little time during a debug/develop cycle.
 - Add `ARKOUDA_QUICK_COMPILE` to try and compile as quickly as possible (disable slow Chapel optimizations, disable back-end optimizations, disable runtime checks.) This will hurt execution performance, but it makes development and debugging significantly faster.

Performance on the 24-core server that nightly single-node perf testing runs on is below.

Build time improvements:
---

| config                                                      | build time |
| ----------------------------------------------------------- | ---------- |
| -sMyDmap=Dmap.blockDist                                     | 580s       |
| -sMyDmap=Dmap.defaultRectangular                            | 425s       |
| -sMyDmap=Dmap.defaultRectangular ARKOUDA_QUICK_COMPILE=true | 190s       |

Execution performance improvements:
---

| benchmark | BlockDist    | DefaultRect  |
| --------- | -----------: | -----------: |
| argsort   |   0.34 GiB/s |   0.38 GiB/s |
| gather    |   6.40 GiB/s |   7.55 GiB/s |
| reduce    | 104.20 GiB/s | 105.30 GiB/s |
| scan      |   5.05 GiB/s |   6.90 GiB/s |
| scatter   |   4.35 GiB/s |  11.45 GiB/s |
| stream    |   14.0 GiB/s |  21.30 GiB/s |

This also helps superdome performance, though releasing those numbers is harder. The impact isn't as significant since the extra parallelism hides the overheads, but there's 10% improvement for gather/scatter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,10 @@ CHPL_FLAGS += $(patsubst %,-L%,$(strip $(subst :, ,$(LD_LIBRARY_PATH))))
 endif
 
 .PHONY: check-deps
-check-deps: check-zmq check-hdf5
+ifndef ARKOUDA_SKIP_CHECK_DEPS
+CHECK_DEPS = check-zmq check-hdf5
+endif
+check-deps: $(CHECK_DEPS)
 
 ZMQ_CHECK = $(DEP_INSTALL_DIR)/checkZMQ.chpl
 check-zmq: $(ZMQ_CHECK)
--- a/Makefile
+++ b/Makefile
@@ -18,10 +18,12 @@ ifeq ("$(shell chpl --version | sed -n "s/chpl version 1\.\([0-9]*\).*/\1/p")", 
 endif
 
 CHPL_DEBUG_FLAGS += --print-passes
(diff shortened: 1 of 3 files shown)


ronawho:
(I think this is good to merge, I'm just leaving it as a draft for a few days so we can get nightly perf results with the merged coargsort optimization.)

PR #360, opened 2020-04-27
327 create python unit test harness

hokiegeek2:


hokiegeek2:
Can anyone tell why the make check fails:

image

I had to change permissions of the check.py file to 755 and then make check worked. Otherwise, not sure what the issue is.

ronawho:
The run_client wrappers (and by extension `util/test/runClient` and `util/test/checkInstall`) just run the specified command in a subprocess so they expect the command to be executable. e.g.

    ./util/test/runClient ./tests/check.py

Turns into the command line equivalent of:

    ./arkouda_server &>/dev/null &
    ./tests/check.py
    ./util/test/shutdown.py
    wait

So the client command needs to be executable for that to work. We could change the wrappers to instead do `python3 <command>` with:

```diff
diff --git a/util/test/util.py b/util/test/util.py
index 8a34009..b272816 100644
--- a/util/test/util.py
+++ b/util/test/util.py
@@ -189,7 +189,7 @@ def run_client(client, client_args=None, timeout=get_client_timeout()):
     This is a thin wrapper over subprocess.check_output.
     """
     server_info = get_server_info()
-    cmd = [client] + [server_info.host, str(server_info.port)]
+    cmd = ['python3'] + [client] + [server_info.host, str(server_info.port)]
     if client_args:
         cmd += client_args
     logging.info('Running client "{}"'.format(cmd))
@@ -203,7 +203,7 @@ def run_client_live(client, client_args=None, timeout=get_client_timeout()):
     shim that returns the returncode instead of raising an exception.
     """
     server_info = get_server_info()
-    cmd = [client] + [server_info.host, str(server_info.port)]
+    cmd = ['python3'] + [client] + [server_info.host, str(server_info.port)]
     if client_args:
         cmd += client_args
     logging.info('Running client "{}"'.format(cmd))
```

hokiegeek2:
@ronawho Good call! Updated and that fixed the make test permission error, pushed to git and CI now runnning

hokiegeek2:
Yup, that looks better: 
image

mhmerrill:
@hokiegeek2 dies this fundamentally change how I as a developer run the python tests... say I want to run just one of the tests from the commend line?

hokiegeek2:
@mhmerrill hey, for the existing tests, I left the main method there, so each one of those can be run as before. For the ones I added, they are written in unittest.TestCase format and they, along with the existing test files can be run individually via the pytest command such as follows:

image

hokiegeek2:
The existing string_test.py via pytest:

image

Also, if pytest is run with the -s CLI argument, all print statements will be sent to the console:

image

Makefile
```diff
@@ -316,6 +316,9 @@ test/%: test/%.chpl
 print-%:
 	@echo "$($*)"
 
+test-python: 
+	pytest -c pytest.ini
```
ronawho, line 320: We have a number of systems where python2 is the default. Maybe `python3 -m pytest -c pytest.ini` to ensure 3 is being used?

README.md
```diff
@@ -100,11 +113,22 @@ export CHPL_RT_OVERSUBSCRIBED=yes
 cd $CHPL_HOME
 make
 ```
+# If you've built Chapel instead using brew, add the Chapel executable (chpl) to PATH:
+export PATH=$CHPL_HOME/bin/linux64-x86_64/:$PATH
+```
```
ronawho, line 118: This and the above anaconda install look linux specific, but are listed under the macos instructions. The header formatting with `###` also seems to have thrown some other formatting off -- https://github.com/mhmerrill/arkouda/blob/55f8a612664417c1d3795cded37079b47bb88538/README.md

tests/groupby_test.py
```diff
@@ -1,12 +1,13 @@
-#!/usr/bin/env python3
-
-import arkouda as ak
 import numpy as np
```
ronawho, line 1: groupby_test.py seems to fail sporadically for me (1/5 times)

```
======================================================= warnings summary =======================================================
tests/groupby_test.py::GroupByTest::test_groupby_on_one_level
  /Users/eronagha/Desktop/arkouda/tests/groupby_test.py:49: RuntimeWarning: invalid value encountered in subtract
    print(f"Different values (abs diff = {np.abs(pdvals - akvals).sum()})")

-- Docs: https://docs.pytest.org/en/latest/warnings.html
=================================================== short test summary info ====================================================
FAILED tests/groupby_test.py::GroupByTest::test_groupby_on_one_level - AssertionError: 0 != 1
============================================ 1 failed, 1 passed, 1 warning in 5.28s ============================================
```

tests/groupby_test.py
```diff
@@ -41,9 +51,17 @@ def compare_keys(pdkeys, akkeys, levels, pdvals, akvals):
...
+    The run_test method enables execution of ak.GroupBy and ak.GroupBy.Reductions
+    on a randomized set of arrays on the specified number of levels. 
+    
+    Note: the current set of valid levels is {1,2}
+    :return: 
+    '''
     d = make_arrays()
     df = pd.DataFrame(d)
     akdf = {k:ak.array(v) for k, v in d.items()}
+    print(akdf)
```
ronawho, line 64: Leftover debug print?

tests/README.md
```diff
@@ -1,5 +1,79 @@
...
+
+The arkouda project has a [pytest](https://docs.pytest.org/en/latest/)-based Python test harness that can be executed 
+via the command-line interface (cli) or within the make file. 
+
+## pytest environment preparation
+
+There are two python libraries that must be installed to execute the arkouda test harness, pytest and pytest-env,
+both of which can be installed via pip:
+```
+pip install pytest, pytest-env
```
ronawho, line 16: Add these to arkouda test dependencies ? 

```diff
diff --git a/setup.py b/setup.py
index 7a814a0..313ff93 100644
--- a/setup.py
+++ b/setup.py
@@ -149,7 +149,7 @@ setup(
     extras_require={  # Optional
         'doc': ['Sphinx', 'sphinx-argparse'],
         'dev': ['h5py'],
-        'test': ['pandas', 'pexpect'],
+        'test': ['pandas', 'pexpect', 'pytest', 'pytest-env'],
     },
     # replace orginal install command with version that also builds
     # chapel and the arkouda server.
```

tests/README.md
```diff
@@ -1,5 +1,79 @@
...
+  server=ARKOUDA\_SERVER\_HOST and port= ARKOUDA\_SERVER\_PORT. If False, the test harness runs in client mode and 
+  consequently an arkouda\_server is not started up.
+
+
+# Running the arkouda Python test harness
+
+To execute all tests in the arkouda test harness via the command line, execute the following command:
+
+```
+pytest -c pytest.ini
```
ronawho, line 63: If you're on board with the `python3 -m pytest` suggestion, probably change these docs too.

tests/README.md
```diff
@@ -1,5 +1,79 @@
...
+    ARKOUDA_VERBOSE=False
+```
+* testpaths: shows the paths to all test files. For the time-being, the arkouda unit tests to be executed are 
+specified on a per-file basis, but pytest can also support directory-level configuration. 
+* norecursedirs: indicates which directories contain files that should be ignored, including tests/deprecated
+* python\_functions: the naming pattern for all test functions. Ih the case of arkouda, all python method (functions)
+  with names starting with "test" are executed by pytest
+* env: the pytest env variables needed to execute the arkouda Python test harness 
+
+## arkouda pytest environmental variables
```
ronawho, line 50: The server supports a `ARKOUDA_SERVER_CONNECTION_INFO` option that will write `host port` out to a file. I wonder if it would make sense to reuse that instead of having the host/port env vars.

tests/string_test.py
```diff
@@ -1,9 +1,7 @@
-#!/usr/bin/env python3
-
```
ronawho, line 2: string_tests also sporadically fails for me (~1/10 times):

```
=================================================== short test summary info ====================================================
FAILED tests/string_test.py::StringTest::test_index - assert '\\"\\\\@PSTZ 4' == '"\\@PSTZ 4'
================================================= 1 failed, 8 passed in 2.69s ==================================================
```

pytest.ini
```diff
@@ -0,0 +1,18 @@
...
+testpaths = 
+    tests/client_test.py
+    tests/compare_test.py
+    tests/dtypes_tests.py
+    tests/groupby_test.py
+    tests/join_test.py
+    tests/operator_tests.py
+    tests/setops_test.py
+    tests/string_test.py
+    tests/where_test.py
```
ronawho, line 11: Just curious -- is the expectation that there will always be an explicit list, or do you expect this to eventually run everything in the `tests` dir?

hokiegeek2: @ronawho good question and yeah, would recommend getting to a point that eventually we run everything in tests directory

tests/string_test.py
hokiegeek2: Yeah, noticed that also and I believe that's an unrelated bug that I am gonna research.

tests/README.md
hokiegeek2: sure, good point!

tests/groupby_test.py
hokiegeek2: I think so

Makefile
hokiegeek2: Sure, I can make that change

tests/string_test.py
ronawho: Is there a way to ignore failures until we can figure that out? I'm thinking about automated testing, where false-positives are a problem.

tests/join_test.py
```diff
@@ -1,88 +1,65 @@
-#!/usr/bin/env python3                                                         
-
 import importlib
```
ronawho, line 1: Several of the join tests fail for me when running with multiple nodes/locales. Running with `arkouda_server -nl 2`:

```
=================================================== short test summary info ====================================================
FAILED tests/join_test.py::JoinTest::test_join_on_eq_with_abs_dt - AssertionError: 1000 != 500
FAILED tests/join_test.py::JoinTest::test_join_on_eq_with_pos_dt - AssertionError: 1000 != 500
FAILED tests/join_test.py::JoinTest::test_join_on_eq_with_true_dt - AssertionError: 1000 != 500
================================================= 3 failed, 3 passed in 12.65s =================================================
```

tests/README.md
hokiegeek2: @ronawho possibly, but what if you want to vary the server you are working with?

tests/join_test.py
hokiegeek2: Odd, does that happen on the master branch also?

tests/README.md
ronawho: Yeah, that's a good point.

tests/string_test.py
hokiegeek2: Yup, can actually set ignore test for now, will add that.

tests/README.md
hokiegeek2: Agreed, will do

tests/join_test.py
ronawho: Yeah, appears to fail on master too

ronawho: cc'ing @mhmerrill for join_test.py failures

pytest.ini
ronawho: Here's the tests that I think are not currently being run:

 - check.py
 - coargsort.py
 - read_all_tests.py
 - read_write_tests.py


check.py might be a little bit different since it's used as the quick sanity check too. 

coargsort was recently added, but it'd be great to convert that too. (I'd be interested in doing that to familiarize myself with pytest a little more)

The other ones look like hdf5 tests that require a file, so are maybe also weird?

ronawho:
I do see (I think expected) warnings from operator_tests.py. Is there a way to mark warnings as expected or suppress them?

ronawho:
I left some feedback, but this looks really good to me.

I tried running this on an InfiniBand cluster and a Cray-XC. Generally seems to work there, though I did need to switch to some of the start/stop server utilities to launch the server. I have a branch that incorporates that if you're interested in bringing that in: https://github.com/mhmerrill/arkouda/compare/master...ronawho:pytest-test

If we can bring that in, quiet the multi-locale join failures, and quiet the sporadic failing tests I think we're good to add this to CI (and I can add it to the local-server/InfiniBand/XC testing we run.)

hokiegeek2:
@ronawho Thanks a bunch for reviewing and for all the good, detailed feedback! Much appreciated. I am working your suggestions in now. I am going to have to look closer at the start/stop server utilities and will do that next.

pytest.ini
hokiegeek2: @ronawho Cool, go for it on coargsort! I agree that should be supported. Also, the read tests will be updated in the near future with an hdf5 file generation utility.

tests/string_test.py
hokiegeek2: @ronawho I submitted a [ticket](https://github.com/mhmerrill/arkouda/issues/364) for the periodic string_test fail and added a unittest ignore annotation for now.

pytest.ini
ronawho: I included updating coargsort in https://github.com/mhmerrill/arkouda/compare/master...ronawho:pytest-test

ronawho:
> I do see (I think expected) warnings from operator_tests.py. Is there a way to mark warnings as expected or suppress them?

I'm not very familiar with this, but following the link suggested in the warnings (https://docs.pytest.org/en/latest/warnings.html) -- something like this seems to work:

```diff
diff --git a/tests/operator_tests.py b/tests/operator_tests.py
index 2064f83..4128f75 100755
--- a/tests/operator_tests.py
+++ b/tests/operator_tests.py
@@ -227,7 +227,13 @@ class OperatorsTest(ArkoudaTest):
         self.assertEqual(np.float64(5), dArray[0])

     def testAllOperators(self):
-        run_tests(verbose)
+        import sys
+        if "pytest" in sys.modules:
+            import pytest
+            with pytest.warns(RuntimeWarning) as record:
+                run_tests(verbose)
+        else:
+            run_tests(verbose)

 if __name__ == '__main__':
     '''
```

---

```python
    @pytest.mark.filterwarnings("ignore:divide by zero")
    @pytest.mark.filterwarnings("ignore:invalid value encountered in true_divide")
    def testAllOperators(self):
        run_tests(verbose)
```

Also worked, and seems a little cleaner, but I wasn't sure if that meant pytest was required and I think you're trying to support modes without pytest still.

hokiegeek2:
> > I do see (I think expected) warnings from operator_tests.py. Is there a way to mark warnings as expected or suppress them?
> 
> I'm not very familiar with this, but following the link suggested in the warnings (https://docs.pytest.org/en/latest/warnings.html) -- something like this seems to work:
> 
> ```diff
> diff --git a/tests/operator_tests.py b/tests/operator_tests.py
> index 2064f83..4128f75 100755
> --- a/tests/operator_tests.py
> +++ b/tests/operator_tests.py
> @@ -227,7 +227,13 @@ class OperatorsTest(ArkoudaTest):
>          self.assertEqual(np.float64(5), dArray[0])
> 
>      def testAllOperators(self):
> -        run_tests(verbose)
> +        import sys
> +        if "pytest" in sys.modules:
> +            import pytest
> +            with pytest.warns(RuntimeWarning) as record:
> +                run_tests(verbose)
> +        else:
> +            run_tests(verbose)
> 
>  if __name__ == '__main__':
>      '''
> ```
> 
> ```python
>     @pytest.mark.filterwarnings("ignore:divide by zero")
>     @pytest.mark.filterwarnings("ignore:invalid value encountered in true_divide")
>     def testAllOperators(self):
>         run_tests(verbose)
> ```
> 
> Also worked, and seems a little cleaner, but I wasn't sure if that meant pytest was required and I think you're trying to support modes without pytest still.

 I just added the CLI argument to suppress warnings in general, but either way is fine with me.

hokiegeek2:
@ronawho thanks again for all the feedback! I incorporated the majority of your suggestions, still need to test a couple of 'em out.

ronawho:
> @ronawho thanks again for all the feedback! I incorporated the majority of your suggestions, still need to test a couple of 'em out.

Sure thing. Taking another pass through now.

Makefile
ronawho: Still game for this change?

tests/README.md
```diff
@@ -1,5 +1,93 @@
...
+
+To execute all tests in the arkouda test harness via make, execute the following command:
+
+```
+make test-python
+```
+pytest also enables a subset of 1..n unit tests to be executed. An example is shown below:
+
+```
+pytest tests/client_test.py
```
ronawho, line 80: `python3 -m pytest`

ronawho:
> I just added the CLI argument to suppress warnings in general, but either way is fine with me.

FWIW I still get the warnings with `--disable-warnings`, though they are a little less visible since it doesn't print out the warning text:

Screen Shot 2020-05-05 at 3 23 04 PM

Might be nice to completely suppress the warnings (I think the diff I sent does, but I'd have to double check) so that new users/testers aren't concerned about them. We can always change that later though, so I'm fine putting that off for now.

PR #369, opened 2020-05-09
Fix join_test.py

ronawho:
test_join_on_eq_with_true_dt_with_result_limit was failing for 16
locales (what our nightly testing uses.) Adjust the limit so that it
results in the correct size when accounting for integer division by
numLocales and BlockDist chunking.

### DIFF
--- a/tests/join_test.py
+++ b/tests/join_test.py
@@ -30,10 +30,11 @@ class JoinTest(ArkoudaTest):
                
     def test_join_on_eq_with_true_dt_with_result_limit(self):
         nl = ak.get_config()['numLocales']
-        lim = self.N//nl * self.N//nl
+        lim = (self.N + nl) * self.N
+        res_size = self.N * self.N
         I,J = ak.join_on_eq_with_dt(self.a1,self.a1,self.a1,self.a1,self.dt,"true_dt",result_limit=lim)
-        self.assertEqual(lim, I.size)
-        self.assertEqual(lim, J.size)
+        self.assertEqual(res_size, I.size)
+        self.assertEqual(res_size, J.size)
 
     def test_join_on_eq_with_abs_dt(self):
         I,J = ak.join_on_eq_with_dt(self.a2,self.a1,self.t1,self.t2,self.dt,"abs_dt")


mhmerrill:
@ronawho the join result limit is divided by the number of locales and each locale gets a result buffer of that size. if all of the results come from one locale then you may need to raise the limit by multiplying it by the number of locales.

tests/join_test.py
```diff
@@ -30,7 +30,7 @@ def test_join_on_eq_with_true_dt(self):
                
     def test_join_on_eq_with_true_dt_with_result_limit(self):
         nl = ak.get_config()['numLocales']
-        lim = self.N//nl * self.N//nl
+        lim = (self.N//nl*nl)**2
```
mhmerrill, line 33: this should be since all locales will contribute to the answer
```python
lim = self.N**2 + nl # the + nl takes care of the integer divide
```

ronawho: With `lim = self.N**2 + nl` for 16 locales this fails with:

```
>       self.assertEqual(lim, I.size)
E       AssertionError: 1000016 != 996008
```

mhmerrill: Well, it's closer.
something else is wrong...
I will look into this because I think the test code/case has changed since I originally wrote it.

mhmerrill: @ronawho ok this test should return 1000**2 answers so something is wrong the 996008 should be 1000000.

mhmerrill: This should be split into two values, the result limit and the actual results expected.

mhmerrill: @ronawho maybe there is an interplay between the `BlockDist` and the limit. This should definitely work...
```python
expected_results = self.N**2
lim = expected_results + nl**2 # nl**2 is to take care of the way BlockDist lays out indices
I,J = ak.join_on_eq_with_dt(self.a1,self.a1,self.a1,self.a1,self.dt,"true_dt",result_limit=lim)
self.assertEqual(expected_results, I.size)
self.assertEqual(expected_results, J.size)
```

ronawho:
```
=========================================================== FAILURES ============================================================
____________________________________ JoinTest.test_join_on_eq_with_true_dt_with_result_limit ____________________________________

self = <join_test.JoinTest testMethod=test_join_on_eq_with_true_dt_with_result_limit>

    def test_join_on_eq_with_true_dt_with_result_limit(self):
        nl = ak.get_config()['numLocales']
        expected_results = self.N**2
        lim = expected_results + nl**2 # nl**2 is to take care of the way BlockDist lays out indices
        I,J = ak.join_on_eq_with_dt(self.a1,self.a1,self.a1,self.a1,self.dt,"true_dt",result_limit=lim)
>       self.assertEqual(expected_results, I.size)
E       AssertionError: 1000000 != 996128

tests/join_test.py:36: AssertionError
==================================================== short test summary info ====================================================
FAILED tests/join_test.py::JoinTest::test_join_on_eq_with_true_dt_with_result_limit - AssertionError: 1000000 != 996128
================================================= 1 failed, 5 passed in 10.43s ==================================================
```

mhmerrill: oops now that I think about it I am trying to set the limit too tight.
we might need `nl**3` which would get `nl**2` per locale.
notice we got more results last time.

last try the I give up:
```python
lim = expected_results + nl**3 # nl**3 is to take care of nl**2 extra results per locale
```

ronawho: `$ ARKOUDA_NUMLOCALES=16 python3 -m pytest tests/join_test.py`

```
=========================================================== FAILURES ============================================================
____________________________________ JoinTest.test_join_on_eq_with_true_dt_with_result_limit ____________________________________

self = <join_test.JoinTest testMethod=test_join_on_eq_with_true_dt_with_result_limit>

    def test_join_on_eq_with_true_dt_with_result_limit(self):
        nl = ak.get_config()['numLocales']
        expected_results = self.N**2
        lim = expected_results + nl**3 # nl**3 is to take care of nl**2 extra results per locale
        I,J = ak.join_on_eq_with_dt(self.a1,self.a1,self.a1,self.a1,self.dt,"true_dt",result_limit=lim)
>       self.assertEqual(expected_results, I.size)
E       AssertionError: 1000000 != 998048

tests/join_test.py:36: AssertionError
----------------------------------------------------- Captured stdout call ------------------------------------------------------
--------------------------------------------------- Captured stdout teardown ----------------------------------------------------
disconnected from arkouda server tcp://*:5555
Stopping arkouda_server in full stack test mode
==================================================== short test summary info ====================================================
FAILED tests/join_test.py::JoinTest::test_join_on_eq_with_true_dt_with_result_limit - AssertionError: 1000000 != 998048
================================================== 1 failed, 5 passed in 9.35s ==================================================
```

ronawho: `lim = expected_results + nl**4` seems to work..

mhmerrill: ok, there is an interplay with the BlockDist allocation of indices and the `N**2` nature of the test

mhmerrill: this is good for now, thanks.

ronawho: What's in this PR seems to pass for 1-32 locales, but I'm still no closer to understanding why :)

mhmerrill: @ronawho explanation of the test and the result_limit:

1. This join test uses an array of `1`'s for both inputs so everything is equal thus giving `N**2` possible results because it is a cartesian product of the two inputs
1. This join test uses the `"true_dt"` predicate meaning don't filter anything (all predicates are true) so pass all results to the final result
1. In the Chapel code for join `result_limit` is divided by the number of locales https://github.com/mhmerrill/arkouda/blob/8411b1ed092f0239f668f00c7c1c6b476b9ecbbe/src/JoinEqWithDTMsg.chpl#L285
1. A buffer is allocated for the results on each locale `result_limit/numLocales` entries for `I` and `J`https://github.com/mhmerrill/arkouda/blob/8411b1ed092f0239f668f00c7c1c6b476b9ecbbe/src/JoinEqWithDTMsg.chpl#L90-L92
1. The buffer not being big enough per locale was the problem and I was trying to set the limit too tight so I wasn't allocating enough space per locale. Since there was a N**2 thing going on I am having trouble reasoning about the `result_limit`.
1. After going through this analysis I believe that the `result_limit` to capture all the results from this test needs to be something like `lim = (N+nl)*N` since the result represents a local index and a cartesian product with all global indices.
1. at the end of the join the per locale result buffers are combined into one array and put into the symbol table.

mhmerrill: I convinced myself that this is wrong! so a good estimate is `lim = 2 * N**2`
the right one is `lim = (N+nl)*N`

ronawho: Updated (and passed with 1-32 locales)

PR #372, opened 2020-05-15
Require Chapel 1.21/1.22

ronawho:
Chapel 1.21/1.22 have some significant performance improvements that
benefit Arkouda, especially at scale. Additionally, there are a number
of code cleanups that can be made if we can use 1.21/1.22 so we want to
require this instead of just suggesting it.

I've updated the documentation to say 1.22 is required, and added an
error if 1.20 or older is used. The motivation for allowing 1.21 but
documenting 1.22 is to encourage new users to use 1.22, but allow users
with other Chapel code that aren't ready for 1.22 to build Arkouda
still.

Closes https://github.com/mhmerrill/arkouda/issues/363

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,7 +4,6 @@ on: [push, pull_request]
 
 env:
   ARKOUDA_QUICK_COMPILE: true
-  CHPL_RT_OVERSUBSCRIBED: yes # limit contention on limited CI hardware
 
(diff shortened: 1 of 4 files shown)


ronawho:
Drafting until https://github.com/mhmerrill/arkouda/pull/373 is merged.

ronawho:
Should be ready to go now

PR #374, opened 2020-05-16
updates to re-enable pytest test harness

hokiegeek2:


tests/base_test.py
```diff
@@ -13,26 +14,24 @@
...
         :return: None
         :raise: RuntimeError if there is an error in configuring or starting arkouda_server
         '''
         if ArkoudaTest.full_stack_mode:
             print('starting in full stack mode')
             try: 
-                nl = get_arkouda_numlocales()
-                (ArkoudaTest.server, ArkoudaTest.port, ArkoudaTest.ak_server) = start_arkouda_server(nl)
+                nl = int(get_arkouda_numlocales())
+                start_arkouda_server(numlocales=nl, host=ArkoudaTest.server, port=ArkoudaTest.port)
```
ronawho, line 34: FYI this breaks our nightly XC and CS testing. The original motivation for start_arkouda_server is to launch the server, and once the server is up and running have it report the hostname and port back via a file, since on clusters/supercomputers you don't know the host ahead of time. e.g. if I'm running on cori, you'll only know the server hostname once you've made it through the slurm queue and been given your nodes. 

I think you could just stop passing the server to start_arkouda_server -- I'm not sure when you would need that in full stack mode.

tests/base_test.py
```diff
@@ -13,26 +14,24 @@
...
         
         :return: None
         :raise: RuntimeError if there is an error in configuring or starting arkouda_server
         '''
         if ArkoudaTest.full_stack_mode:
             print('starting in full stack mode')
             try: 
-                nl = get_arkouda_numlocales()
-                (ArkoudaTest.server, ArkoudaTest.port, ArkoudaTest.ak_server) = start_arkouda_server(nl)
+                nl = int(get_arkouda_numlocales())
```
ronawho, line 33: Nit but `get_arkouda_numlocales()` should already return an int

tests/base_test.py
```diff
@@ -1,7 +1,8 @@
 import unittest
 import subprocess, os, signal
 from context import arkouda as ak
-from util.test.util import start_arkouda_server, stop_arkouda_server, get_arkouda_numlocales
+from util.test.util import get_arkouda_server, get_arkouda_numlocales,\
```
ronawho, line 4: Unused `get_arkouda_server` import

tests/base_test.py
hokiegeek2: @ronawho oh wow, okay, I can take this out. I put it in there because I am pretty sure that specifying the server was being added to Chapel and I wanted to enable configuration of the host (hostname or IP address) for situations where the are multiple NICs for the host.

tests/base_test.py
hokiegeek2: Good point, will clean this up.

tests/base_test.py
hokiegeek2: @ronawho okay, made it so the host can be none. Is passing in the ServerPort param okay:

    `cmd = [get_arkouda_server(),
           '--logging={}'.format('true' if log else 'false'),
           '--v={}'.format('true' if verbose else 'false'),
           '--serverConnectionInfo={}'.format(connection_file),
           '-nl {}'.format(numlocales), '--ServerPort={}'.format(port)]`

Please confirm and I'll wrap up these updates and submit a pull request.

pytest.ini
```diff
@@ -12,3 +12,9 @@ testpaths =
     tests/where_test.py
 norecursedirs = .git dist build *egg* tests/deprecated/*
 python_functions = test*
+env =
+    ARKOUDA_SERVER_HOST=localhost
+    ARKOUDA_SERVER_PORT=5555
+    ARKOUDA_FULL_STACK_TEST=True
+    ARKOUDA_NUMLOCALES=2
```
ronawho, line 19: In our nightly XC/CS testing we `export ARKOUDA_NUMLOCALES=16` -- does the env var take priority, or does the value in the .ini file? (Based on a quick test it seems like the value in the ini file)

ronawho: Deferring to you for pytest expertise, but maybe we can prefix these with `D:` to not override command line settings?

tests/base_test.py
ronawho: Think I'm missing something -- looks like that's the same code that's on master

pytest.ini
hokiegeek2: @ronawho You are correct and it sounds like it's dorking up your nightly test so I will take that out of pytest.ini.

ronawho: I'm find leaving it in if the `.ini` style is what python programmers will be used to. I think it'd just need to be `D:ARKOUDA_NUMLOCALES=2` so that command line settings won't be overwritten

tests/base_test.py
```diff
@@ -13,26 +14,24 @@
...
         :raise: RuntimeError if there is an error in configuring or starting arkouda_server
         '''
         if ArkoudaTest.full_stack_mode:
             print('starting in full stack mode')
             try: 
-                nl = get_arkouda_numlocales()
-                (ArkoudaTest.server, ArkoudaTest.port, ArkoudaTest.ak_server) = start_arkouda_server(nl)
+                nl = int(get_arkouda_numlocales())
+                start_arkouda_server(numlocales=nl, host=ArkoudaTest.server, port=ArkoudaTest.port)
                 print('Started arkouda_server in full stack test mode')
```
ronawho, line 35: If you're making changes, would you mind adding the number of locales to this print statement just so it's easy to see how many are being used?

tests/base_test.py
hokiegeek2: @ronawho yes, it was merged into the master yesterday.

tests/base_test.py
hokiegeek2: @ronawho sure thing!

tests/base_test.py
ronawho: ooh, sorry -- I misread, I thought you were asking me if this new code snippet looked ok, but you were just asking if the code on master that passes the port is ok. 

Yeah, we don't do anything with the port, that change is good.

pytest.ini
hokiegeek2: @ronawho gotcha, okay

PR #376, opened 2020-05-19
371 reverse some recent pytest based test harness changes

hokiegeek2:


tests/base_test.py
```diff
@@ -30,9 +28,9 @@ def setUpClass(cls):
         if ArkoudaTest.full_stack_mode:
             print('starting in full stack mode')
             try: 
-                nl = int(get_arkouda_numlocales())
-                start_arkouda_server(numlocales=nl, host=ArkoudaTest.server, port=ArkoudaTest.port)
-                print('Started arkouda_server in full stack test mode')
+                nl = get_arkouda_numlocales()
+                start_arkouda_server(numlocales=nl, port=ArkoudaTest.port)
```
ronawho, line 32: Not setting `ArkoudaTest.server`, maybe:

```python
ArkoudaTest.server, _, _ = start_arkouda_server(numlocales=nl, port=ArkoudaTest.port)
```

pytest.ini
```diff
@@ -16,5 +16,5 @@ env =
     ARKOUDA_SERVER_HOST=localhost
     ARKOUDA_SERVER_PORT=5555
     ARKOUDA_FULL_STACK_TEST=True
-    ARKOUDA_NUMLOCALES=2
+    D:ARKOUDA_NUMLOCALES=2
     ARKOUDA_VERBOSE=False
```
ronawho, line 20: Would you be up for making all these env vars `D:` prefixed? For better or worse (probably worse) HPC people tend to be env var oriented, so if I was going to do a client session, I'd probably do so with the env vars instead of modifying the `.ini` file

hokiegeek2: Indeed! Just made the change and pushed to git

ronawho:
Thanks! Looks good other than not setting `ArkoudaTest.server`

hokiegeek2:
> Thanks! Looks good other than not setting `ArkoudaTest.server`

Cool, that is being set and will be used in the client mode and I kept it in the start_arkouada_server param list for potential future use. Ready for it to be merged to trunk or do you want to test in your environments?

ronawho:
>> Thanks! Looks good other than not setting ArkoudaTest.server

>Cool, that is being set and will be used in the client mode and I kept it in the start_arkouda_server param list for potential future use. Ready for it to be merged to trunk or do you want to test in your environments?

It's hanging under CS/XC still because the server name isn't being communicated back to the test infra yet, so the tests are trying to connect to `localhost` instead of `nid0002` (XC example) or `prod-0001` (CS example) or whatever node the server ended up running on.

Something like the following works for me:

```diff
diff --git a/tests/base_test.py b/tests/base_test.py
index 71d2975..261a90a 100644
--- a/tests/base_test.py
+++ b/tests/base_test.py
@@ -29,7 +29,7 @@ class ArkoudaTest(unittest.TestCase):
             print('starting in full stack mode')
             try:
                 nl = get_arkouda_numlocales()
-                start_arkouda_server(numlocales=nl, port=ArkoudaTest.port)
+                ArkoudaTest.server, _, _ = start_arkouda_server(numlocales=nl, port=ArkoudaTest.port)
                 print('Started arkouda_server in full stack test mode with {} locales'.format(nl))
             except Exception as e:
                 raise RuntimeError('in configuring or starting the arkouda_server: {}, check ' +
```

hokiegeek2:
@ronawho gotcha, okay, changed code to your suggestion, passed testing here, pushed change to git and tests are running now.

PR #391, opened 2020-06-15
Update keyPart in SegStringSort.StringIntComparator

mppf:
Following Chapel PR https://github.com/chapel-lang/chapel/pull/15802
the keyPart indices are numbered from 0. This PR adjusts StringIntComparator
to avoid the issue by simply running the defaultComparator keyPart
on the string argument to be sorted.

I verified that this strategy works across Chapel versions by checking the following program sorts the array in the same order with master (after PR 15802) and the 1.20 release:

``` chapel
use Sort;

record StringIntComparator {
  proc keyPart((a0,_): (string, int), in i: int) {
    // Just run the default comparator on the string
    return Sort.defaultComparator.keyPart(a0, i);
  }
}

proc main() {
  var A = [("ab", 1), ("a", 2), ("aa", 3)];
  sort(A, new StringIntComparator());
  writeln(A);
}
```

### DIFF
--- a/src/SegStringSort.chpl
+++ b/src/SegStringSort.chpl
@@ -17,16 +17,8 @@ module SegStringSort {
 
   record StringIntComparator {
     proc keyPart((a0,_): (string, int), in i: int) {
-      var len = a0.numBytes;
-      // This uses .indices to make this code work with Chapel 1.21 or 1.22;
-      // If/when we want to hard-code to 1.22+, we can remove the following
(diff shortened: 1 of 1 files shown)


mhmerrill:
@bradcray @mppf is this change make us compatible with Chapel 1.22 only?

bradcray:
@mppf's OP above stated that he verified with the 1.20 release.  Looking at the code, it appeared to me that he made it completely index neutral by calling back into one of the standard keyPart() implementations that was available with either release.

Issue #179, opened 2019-11-05
Translate setops into chapel server side operations


mhmerrill:
Currently pdarray set operations like `ak.union1d()` consist of 5-12 seperate server operations.  Making the various set operations into single server side operations would provide significant speed improvement.

mhmerrill:
> I am currently working on this issue, let me know if you have any concerns about this

@mcdobe100 I don't have any concerns. I do have a few ideas about this if you want to chat about it otherwise "have at it".

mcdobe100:
@mhmerrill Yes, I would definitely be interested in chatting about this, when are you available? I currently have a branch `translate-setops` on my fork that has a proposed implementation for `union1d()` moved onto the server as `newUnion1d()` if you wanted to take a look at that.

mhmerrill:
> @mcdobe100 asked: For getting the `overMemLimit` into my functions, do you have any recommendations for calculating out what they should be? Thanks.

I'll try to describe here what my process is for calulating a memory estimate.

1. remember, we want a high water mark for memory
1. look through the code and see what arrays get allocated
1. also look for extra data structures and calulate their memory foot print
1. remember all allocation in the symbol table get tracted and` overMemoryLimit` gets called for simple operations
1. more complex operations have to call `overMemoryLimit` with an estimate for the memory high watermark that will be hit during the computation

Lets take `argSort` for example:
https://github.com/mhmerrill/arkouda/blob/6bf11b62fbd0eec895e407559d752b86fd5da779/src/ArgSortMsg.chpl#L591-L597

1. the formula here says there are 4 intermediate arrays and the result allocated during the computation for the `radixSortLSD_ranks`
1. there are also the counts/bucket arrays allocated which depend on the task parallelism available and the number of locales.
1. Since we keep `itemsize` and `size` in the symboltable entry for the array passed in and we only sort` int64` and `float64` data I took a shortcut for the formula (my bad).

mcdobe100:
```
overMemLimit(((6 + 1) * gEnt.size * gEnt.itemsize * gEnt2.size * gEnt2.itemsize));
```
 @mhmerrill Would this then be an acceptable `overMemLimit` for `union1d` then because of the 6 intermediate arrays allocated or would it instead be something like this
```
overMemLimit(((6 + 1) * gEnt.size * gEnt.itemsize * gEnt2.size * gEnt2.itemsize)
       + (6 * here.maxTaskPar * numLocales * 2**16 * 8));
```
[...]

PR #392, opened 2020-06-15
Translate pdarray setops to server-side operations

mcdobe100:
Move the pdarray set operations to single server side operations instead of the several server calls that the operations required previously. Include`union1d()`, `setdiff1d()`, `setxor1d()`, and `intersect1d()`
Outlined in issue #179 

Set operations performance on 16 nodes of an XC (GiB/s):
| test        | before | after |
| union1d     | 1.36 | 2.00 |
| intersect1d   | 1.16 |2.06|
| setxor1d    | 1.13 | 1.99 |
| setdiff1d |0.48| 0.60 |

Set operations performance on single node (GiB/s):
| test        | before | after |
| union1d     | 0.12 |0.22 |
| intersect1d   | 0.15 | 0.21 |
| setxor1d    | 0.15 | 0.20 |
| setdiff1d | 0.11 | 0.13 |


mhmerrill:
@mcdobe100 I converted this to a Draft PR because it's not ready to be merged, in my opinion.

src/ArraySetops.chpl
```diff
@@ -0,0 +1,180 @@
...
+    use In1d;
+
+    use CommAggregation;
+
+    proc intersect1d(a1: [?aD] int, b1: [aD] int, assume_unique: string) {
+      if assume_unique == "False" {
+        var (a, _)  = uniqueSort(a1);
+        var (b, _)  = uniqueSort(b1);
+        var aux2 = concatset(a,b);
+        var aux_sort_indices = radixSortLSD_ranks(aux2);
```
mhmerrill, line 20: you could call the `radixSortLSD_keys` procedure to avoid the permutation on the next line. `radixSortLSD_keys` returns the sorted array if you don't need the permutation.

src/ArraySetops.chpl
```diff
@@ -0,0 +1,180 @@
...
+
+        var ret = boolIndexer(aux, mask);
+
+        return ret;
+      } else {
+        ref a  = a1;
+        ref b  = b1;
+
+        var aux2 = concatset(a,b);
+        var aux_sort_indices = radixSortLSD_ranks(aux2);
```
mhmerrill, line 76: same comment as above use the `radixSort_keys` procedure.

src/ArraySetops.chpl
```diff
@@ -0,0 +1,180 @@
...
+    use RadixSortLSD;
+    use Unique;
+    use Indexing;
+    use In1d;
+
+    use CommAggregation;
+
+    proc intersect1d(a1: [?aD] int, b1: [aD] int, assume_unique: string) {
+      if assume_unique == "False" {
+        var (a, _)  = uniqueSort(a1);
```
mhmerrill, line 17: we should probably generate a unique routine which does not return counts.

mhmerrill:
@mcdobe100 this generally looks good. I would like to see a few more comments in the Chapel code but good so far.

mcdobe100:
@mhmerrill Alright, thanks for the review, I will get those changes in that you mentioned and then put it up to ready for review when I get that done.

mcdobe100:
I think that this code is ready to be merged (unless you see anything that you would like to be changed), but please wait until PR #394 has been merged for a couple of days before merging this PR, so we can gain some benchmark data for the current set operations and have some data to compare these changes to.

src/ArraySetops.chpl
```diff
@@ -0,0 +1,137 @@
...
+ includes intersection, union, xor, and diff
+
+ currently, only performs operations with integer arrays 
+ */
+
+module ArraySetops
+{
+    use ServerConfig;
+
+    use BlockDist;
```
ronawho, line 11: I don't think BlockDist or CommAggregation are used in this module.

src/ArraySetops.chpl
```diff
@@ -0,0 +1,137 @@
...
+
+    /*
+    Medium bound const. Per locale associative domain in1d implementation recommended.
+    */
+    private config const mBound = 2**25; 
+
+    // returns intersection of 2 arrays
+    proc intersect1d(a: [?aD] int, b: [aD] int, assume_unique: string) {
+      //if not unique, unique sort arrays then perform operation
+      if (assume_unique == "False") {
```
ronawho, line 34: I'm not sure what other places in the code do, but it seems like it'd be cleaner to pass in a `bool` here, and at the callsite check that assume_unique is either "true" or "false" and convert to the appropriate bool (just thinking about a case where the message is improperly formatted, but we're not checking for that.

src/ArraySetops.chpl
mcdobe100: You are right, thanks, fixed that

mhmerrill:
@mcdobe100 @ronawho @bradcray I have merged in PR #394 . Since you want me to hold off on merging this PR... Let me know when you want to merge in this PR.

mcdobe100:
@mhmerrill @ronawho  I think that this is ready to go in if it looks good to you guys

mhmerrill:
@mcdobe100 @ronawho do we have an understanding of the performance increase between the old multi-message version and the new single message version of the set operations?

ronawho:
> do we have an understanding of the performance increase between the old multi-message version and the new single-message version of the set operations?

Yeah, see https://github.com/mhmerrill/arkouda/pull/392#issuecomment-651369249 for single node and 16 node XC numbers. @mcdobe100 could you add that table to the PR description?

mhmerrill:
ok, I thought these numbers were fromthe last optimizations made to the code.
So these numbers ARE from mulit-message vs single-message implementations... for some reason I would have expected more of an increase but it is still good.

> > do we have an understanding of the performance increase between the old multi-message version and the new single-message version of the set operations?
> 
> Yeah, see [#392 (comment)](https://github.com/mhmerrill/arkouda/pull/392#issuecomment-651369249) for single node and 16 node XC numbers. @mcdobe100 could you add that table to the PR description?

src/ArraySetops.chpl
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    // Get intersect of 2 arrays
+    // first concatenates the 2 arrays, then
+    // sorts arrays and removes all values that
+    // only occur once
+    proc intersect1dHelper(a: [] ?t, b: [] t) {
+      var aux = radixSortLSD_keys(concatset(a,b));
+
+      var head = sliceEnd(aux);
```
mhmerrill, line 47: maybe you should rename `sliceEnd()` to `sliceHead()`
and `tail()` to `sliceTail()`
it took me a minute to figure this out

src/ArraySetops.chpl
```diff
@@ -0,0 +1,135 @@
...
+      var mask = head == tail(aux);
```
mhmerrill, line 48: It seems that we are making a copy of the two sections of the array instead of doing the simple comaprasion between array elements which would avoid the copy all together.
Could we have the head/tail function just compute the slice and return it, or maybe just do something like...
```chapel
var mask = aux[aux.domain.low()..(aux.domain.high()-1)] == aux[(aux.domain.low()+1)..aux.domain.high()];
```
@bradcray what do you think?

ronawho:
> for some reason I would have expected more of an increase but it is still good.
> Both sets of measurements are benefiting from the optimizations in chapel 1.22 also before 1.22 the measurements probably would have been greater, cause for the all the array creation and destruction.

Yeah, I suspect the improvements would have been bigger before 1.22 or at higher scale. 

> it seems like the code works. I might still approve unless you guys want to do some more optimizations.

We were planning to look at further optimizations over the next 2 weeks, but I think what's here is a nice enough improvement that it's worth getting in to establish a new baseline for future optimizations.

src/ArraySetops.chpl
mcdobe100: Thanks for the suggestion, I struggled with the naming of those, so glad to settle on something.

src/ArraySetops.chpl
bradcray: Yeah, I agree that this feels pretty heavyweight since `head` and `tail` represent all of the array elements but one.  I suspect/hope we could get better performance by doing something like:

```chapel
const maskInds = aux.domain#(aux.size-1);  // maskInds describes all indices but the last
var mask = [i in maskInds] aux[i] == aux[i+1];
```

The main overheads of this implementation would be:

(a) creating a new distributed domain, `maskInds`, but this should be strictly less expensive than the current slicing-based approach, since it'll create distributed domains and arrays.
(b) the overheads for accessing potentially-remote data in the `aux[i+1]` expression
(c) and potentially similar overheads for the `aux[i]` expression unless @e-kayrakli's ongoing work on optimizing aligned accesses could handle it automatically... assuming not, `aux.localAccess[i]` would take care of it, so is probably where we should start).

We could work around (b) with more effort if it turned out to be overly expensive (for example, we could have per-locale loops that iterated over the first local_n-1 iterations using `.localAccess`, stripping off the final iteration that is guaranteed to be remote).  But my guess is that even without those changes, this version would still be a net win over creating the two distributed n-1 element arrays.

e-kayrakli: > and potentially similar overheads for the aux[i] expression unless @e-kayrakli's ongoing work on optimizing aligned accesses could handle it automatically... assuming not, aux.localAccess[i] would take care of it, so is probably where we should start)

I expect that optimization to not fire for this pattern today. But I'll get to it soon. I am confident that it'll fire in the next release.

PR #394, opened 2020-06-19
Add a benchmark for set operations

mcdobe100:
Add a benchmark that will test the speed of ``setdiff1d``, ``union1d``, ``setxor1d``, and ``intersect1d``


benchmarks/graph_infra/arkouda.graph
```diff
@@ -45,3 +45,9 @@ graphkeys: Noop ops/s
 files: noop.dat
 graphtitle: Noop Performance
 ylabel: Performance (ops/s)
+
+perfkeys: intersect Average rate =, union Average rate =, xor Average rate=, diff Average rate=
```
ronawho, line 49: Could you move this above the noop graph? (The graphs show up in the order they're listed here and I think this one is more interesting/important than the noop)

PR #397, opened 2020-06-22
Data frame

rolandqli:
General framework for pandas DataFrame-like class. Essentially, an organized collection of pdarrays where we can perform DataFrame-like operations by calling methods on the pdarrays within. The object itself exists on the client side and communication with the server is done through the pdarrays of the object. Still many methods to test and design options to consider.


mhmerrill:
@rolandqli Thank you for this PR. I have converted this PR to a draft. Since there are others thinking about and looking at creating/implementing data-frame-like functionality I would like to foster a discussion about the implementation you have started. I would like a few people to weigh-in @reuster986 @hokiegeek2 @mcdobe100 @timothyneumann1 @bradcray @ben-albrecht... We may have mostly the same thoughts on implementation but since I am not as much of an expert in python as these others I will defer to them until you guys come up with a recommendation.
Thanks, Mike.

mcdobe100:
@mhmerrill I am similar in the sense that I have very limited Python experience, so I am not the best to provide insight on the actual implementation, but I think that @rolandqli is going to be the main user of this data frame for the summer and because of that I think that the plan was to go with this one because I think it is working for him. Also, Roland and I have been working to add a functionality to skip over `NaN` values for a few of the GroupBy functions and I will open up a PR for that when I am feeling confident in it. Certainly I think it is important to have others weigh in, but I think that the plan was to go with this data frame and I will be working on something else for the summer.

mhmerrill:
@rolandqli @mcdobe100 @reuster986 @bradcray @hokiegeek2 I have some reservations about merging a completely python side package into the Arkouda python package.

- Is it possible to keep this as a separate python package?
- I am willing to make changes to Arkouda to facilitate this package, if need be.
- Part of this is a desire to potentially have multiple implementations of functionality/packages which depend on Arkouda.
- Just like Pandas is separate from NumPy.
- I am not sure about the some of the design choices

I can be convinced either way and I hope this makes sense to you guys.

hokiegeek2:
@mhmerrill Definitely some excellent points, and the main one that resonates with me is the Pandas-Numpy analogy. I think this is a really good effort that leverages Arkouda but I recommend that it becomes a separate project with a dependency on Arkouda, similar to Pandas and Numpy. Again, this is a really nice effort by @rolandqli!

ben-albrecht:
I may not have enough context on the motivation for merging this into Arkouda vs. separating, but I'll add:

> Is it possible to keep this as a separate python package?

I think this is an attractive option. It would help keep issues, design discussions, and PR reviews of dataframes vs. Arkouda separate. Also, it puts the responsibility of maintenance onto the package author.

> Just like Pandas is separate from NumPy.

👍 

> I am not sure about the some of the design choice

Even if this is split off into a separate repository / package, I would hope that many of the Arkouda developers could continue to weigh in on the design decisions so we end up with something everyone is happy with.

@rolandqli -- For reviewing the interface in general, it could be helpful to have a summary of where and how this interface deviates from pandas in the PR description. Also, I usually start by looking at usage of the interface in the tests - a few tests could be helpful here. Thanks!

reuster986:
First off, this is a good start @rolandqli ! I agree with the general sentiment above that an arkouda-based dataframe effort should probably be organized as a separate python package and github project (with its own issues, PRs, etc). Of course, there should be a lot of collaboration and cross-talk between arkouda and any arkouda-based dataframe effort, and I would definitely be willing to participate in design discussions and work out new arkouda functionality needed to support dataframes.

I see two fundamental design questions related to this PR and arkouda-based dataframes in general:

## 1. How Pandas-like?

Pandas is obviously the inspiration for all this, but it is not a perfect API, and we have a chance to improve it. Where should an arkouda dataframe deviate from the pandas API?

For example, I am not a huge fan of the pandas `df.loc[]` idiom. I know the pandas devs probably have a good reason for it, but I would prefer straight-up bracket indexing that understands forms like `df[row(s), column(s)]`. In that vein, I really like the direction @rolandqli is going in this PR (see, for example, the docstring of `DFrame.__getitem__`). My vote is that we should use the following principles:

* A 1-D index always implies a column or column slice, e.g. `df[col]` or `df[col1:col2]`
* To select rows, user must use a 2-D index, e.g. `df[row, :]` for all columns in a single row or `df[rows, col]` for a subset of rows in a single column
* Rows support logical indexing, e.g. `subdf = df[condition, :]` where condition is a boolean vector whose length equals the number of rows in the dataframe
* Row indices should be interepreted in the index domain, as opposed to assuming a 0-up range index. We could maybe provide a `df.iloc[i]` method to give the option of specifying 0-up row indices independent of the dataframe's actual index domain, like pandas does.

These are just suggestions, and I'm open to counterarguments.

## 2. Pure Python or Chapel?

Should dataframes be built entirely atop the arkouda python client, abstracted from Chapel, or is there a compelling reason to implement a dataframe class in Chapel? My intuition is that we should try to build the entire dataframe class on top of arkouda, because all the dataframe operations I can think of amout to bookkeeping around existing arkouda functionality, as opposed to new patterns of distributed computation. If we decide on this approach, I believe it would render the `DFrameServer` class unnecessary.

buddha314:
I agree that the `df.loc[]` convention is strange and I don't know if it exists anywhere else. Thus, I don't think it's worth preserving and Arkouda should respect the `R` inspiration or even more SQL-like structures.

I would like to suggest a user story around this:  

* Noelle is a data scientist building a large scale agent-based simulation.  Each agent can be updated according 1) the agent's current state and 2) some statistical properties of all agents (like center of mass, average speed, etc).  Thus, a population-level calculation is performed followed by an agent-level calculation. 
* She wants to collect the new agent state into a dataframe to analyze the evolution of the system.  She will need to record at least the following into the table
  * The epoch (or iteration) id
  * The agent id
  * The current (new) agent state
* To complete the simulation, she will need to order the dataframe by agent, epoch.

bradcray:
@mhmerrill, @reuster986, and @hokiegeek2:  @rolandqli and I were just discussing next steps for this Arkouda DataFrame effort today, and we wanted to get your thoughts on a few things:

1) It sounds like nobody (including us) disagrees with the concept of supporting this dataframe as a separate package rather than new functionality within the arkouda package, so I think we've got our work cut out for us there.

2) Any thoughts on the naming of the new package?  Should we call it something descriptive like `ArkoudaDataframe` or `akDataframe` or the like, or should we come up with a new Greek animal? (preferably one that sounds like a pair of other programming languages... :D )

3) I feel like we got slightly mixed signals on whether to host this in this same repo or move it off into a new one.  Obvious advantages to keeping it here would be that it would make it more likely to get noticed/used by existing Arkouda users, it could be tested simultaneously with Arkouda such that if an Arkouda interface changed, the breakage would be caught more quickly, etc.  Obvious downsides are that it blends two distinct (but related) capabilities into a single repository, and isn't scalable if we end up with dozens of such packages (which would be a nice problem to have).  With either choice, we could obviously reverse it in the future (split it into its own repo; or fold its repo into Arkouda's).  So I think the question here is what your preference is for the short-term: get this into a mergeable state or move it elsewhere?

4) If it were to remain in this repo, what do you view as minimal changes required to merge it (vs. improvements that can be done later), where I think point 1 above is an obvious precondition.  Some form of testing/validation?  (automated or manual?)  Any of the above suggestions/modifications?

Thanks!

mhmerrill:
my comments are included interleaved below:
> @mhmerrill, @reuster986, and @hokiegeek2: @rolandqli and I were just discussing next steps for this Arkouda DataFrame effort today, and we wanted to get your thoughts on a few things:
> 
> 1. It sounds like nobody (including us) disagrees with the concept of supporting this dataframe as a separate package rather than new functionality within the arkouda package, so I think we've got our work cut out for us there.

agreed, I don't think it's all that much work if you put it all in one package file.

> 2. Any thoughts on the naming of the new package?  Should we call it something descriptive like `ArkoudaDataframe` or `akDataframe` or the like, or should we come up with a new Greek animal? (preferably one that sounds like a pair of other programming languages... :D )

Suggested names:
@bradcray : Kallisto (greek mythology)
@mhmerrill : baloo (Kipling's Jungle Book), baahloo (hindi for bear)
I am also ok with akDF or something simple and straight forward

> 3. I feel like we got slightly mixed signals on whether to host this in this same repo or move it off into a new one.  Obvious advantages to keeping it here would be that it would make it more likely to get noticed/used by existing Arkouda users, it could be tested simultaneously with Arkouda such that if an Arkouda interface changed, the breakage would be caught more quickly, etc.  Obvious downsides are that it blends two distinct (but related) capabilities into a single repository, and isn't scalable if we end up with dozens of such packages (which would be a nice problem to have).  With either choice, we could obviously reverse it in the future (split it into its own repo; or fold its repo into Arkouda's).  So I think the question here is what your preference is for the short-term: get this into a mergeable state or move it elsewhere?

I would like it in the same repo for now, I'd suggest making a top level directory for the python package call it what you want to name the package. if it becomes cumbersome we can move things around in the repo.

> 4. If it were to remain in this repo, what do you view as minimal changes required to merge it (vs. improvements that can be done later), where I think point 1 above is an obvious precondition.  Some form of testing/validation?  (automated or manual?)  Any of the above suggestions/modifications?

I would start off with everything in a package file like `Kallisto.py` or `akDF.py` put that into a directory called `./Kallisto` along with an example, a test, and a readme. there doesn't seem to be that much code currently to make this unwieldy.

> 
> Thanks!

mhmerrill:
@rolandqli @bradcray @reuster986 what do you think of the suggestions? I'd like to get this wrapped up soon.

rolandqli:
@mhmerrill That sounds great. I will start working towards that. Thanks!

mhmerrill:
i haven't seen any further work here for two months so I am closing this PR.

PR #400, opened 2020-06-26
Adjust benchmarks to run all supported types in correctness mode

ronawho:
Previously the CI testing used to assume all benchmarks would work with
int/float and manually specified some that should work with bools.
However, the setops benchmark only works with ints and future benchmarks
may have varied types, so instead of hardcoding this, just let the
benchmarks maintain the list of types they support and run them all in
correctness mode.

Also ensure that graph generation works during CI testing. It can be
tricky for to get the graph keys/names right, so check it during CI.

Resolves https://github.com/mhmerrill/arkouda/issues/395

### DIFF
--- a/benchmarks/reduce.py
+++ b/benchmarks/reduce.py
@@ -5,6 +5,7 @@ import numpy as np
 import arkouda as ak
 
 OPS = ('sum', 'prod', 'min', 'max')
+TYPES = ('int64', 'float64')
 
 def time_ak_reduce(N_per_locale, trials, dtype, random):
     print(">>> arkouda reduce")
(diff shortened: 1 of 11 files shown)


benchmarks/reduce.py
```diff
@@ -5,6 +5,7 @@
 import arkouda as ak
 
 OPS = ('sum', 'prod', 'min', 'max')
+TYPES = ('int64', 'float64')
```
mhmerrill, line 8: `bool` reductions are also supported with `sum`, `prod`, `min`, and `max`

benchmarks/scan.py
```diff
@@ -5,6 +5,7 @@
 import arkouda as ak
 
 OPS = ('cumsum', 'cumprod')
+TYPES = ('int64', 'float64')
```
mhmerrill, line 8: `bool` also supports `cumsum` and `cumprod`

benchmarks/reduce.py
```diff
@@ -5,6 +5,7 @@
 import arkouda as ak
 
 OPS = ('sum', 'prod', 'min', 'max')
```
mhmerrill, line 7: suggestion: other reductions and types are supported but we might not really need them for benchmarking maybe just for testing correctness.
`int64`, `float64`, `bool` also support `any` and `all` reductions
`argmin`, `argmax`, and `is_sorted` are also supported for `int64` and `float64`

mhmerrill:
@ronawho @mcdobe100 I left a few comments, let me know if you want to make these changes otherwise I'll just accept and merge later today.

benchmarks/reduce.py
ronawho: As-is the benchmark doesn't work with bools (there's no case to handle creating the arrays of bools). I can add that here too if you want, or I can do that in a follow up PR.

benchmarks/reduce.py
ronawho: Interesting, that's up to you -- If they have performance characteristics it might be worth adding them here, but if performance won't be much different, I'd probably leave them out of here and just rely on the other correctness tests for these ops

ronawho:
I think lets merge this as-is and if you'd like to see the other types/ops added, let's create an issue and assign it to me.

mhmerrill:
ok, i'll merge as is... @ronawho can you create the issue?

ronawho:
> i'll merge as is... @ronawho can you create the issue?

Missed this comment earlier, but yeah -- I created https://github.com/mhmerrill/arkouda/issues/458

PR #401, opened 2020-06-26
Add capability for skipping nan values for some GroupBy functions

mcdobe100:
Can skip nan values by passing in a new argument to a GroupBy object
when calling `mean`, `max`, `min`, and `sum` on floats. This can be
separate if you prefer, as it is not something that I understand would be 
useful to you guys, but it does not change existing functionality (new 
argument defaults to false which will just use the old code), and would 
help our internal uses. Also, I can add additional testing if you would like
to ensure this code is working properly.


mhmerrill:
@mcdobe100 we either need to add or modify the groupie python test to comprehend this new functionality.

mcdobe100:
@mhmerrill Alright, I will convert this to a draft, update the testing and then put it up for review when it's ready. Thanks

mhmerrill:
@mcdobe100 I think things look good, just wanted the groupby test to comprehend this feature

mcdobe100:
@mhmerrill Alright, fixed a few things and added some testing, so let me know what you think of it now, thanks!

mhmerrill:
@mcdobe100 I have assigned @reuster986 to review this because he wrote the GroupBy and all the segmented reductions. Hopefully he will get to it this week.

src/ReductionMsg.chpl
```diff
@@ -611,11 +619,40 @@ module ReductionMsg
...
+    proc segMean(values:[] ?t, segments:[?D] int, skipNan=false): [D] real {
       var res: [D] real;
       if (D.size == 0) { return res; }
-      var sums = segSum(values, segments);
-      var counts = segCount(segments, values.size);
+      var sums;
+      var counts;
+      if (isFloatType(t) && skipNan) {
+        var arrCopy = makeDistArray(values.size, real);
+        var nancounts = makeDistArray(segments.size, int);
```
reuster986, line 629: I worry that the logic inside the `forall` here will be slow in multilocale. I might approach this with a more vectorized pattern, e.g.
```chapel
// count cumulative nans over all values
var cumnans = + scan isnan(values); // assuming isnan is promotable
// find cumulative nans at segment boundaries
var segnans: [segments.domain] int;
forall si in segments.domain {
    if si == segments.domain.high {
        segnans[si] = cumnans[cumnans.domain.high];
    } else {
        segnans[si] = cumnans[segments[si+1]-1];
    }
}
// take diffs of adjacent segments to find nan count in each segment
var nancounts = segnans[segnans.domain.interior(-(segnans.domain.size-1))] - segnans[segnans.domain.interior(segnans.domain.size-1)];
```

I have not checked the logic or syntax here (in particular, I always forget how `.interior` works), so treat this as pseudocode.

mcdobe100:
@reuster986 I think that the name change is a good idea. 

In regards to the `skipna=True`, I could try to gather some performance numbers on the affects of skipping nans so we can have more of a feel on what the impact of that would be on the project if that sounds good to you and others.

mcdobe100:
@reuster986 Here are the performance numbers that I gathered on the difference between using the `skipna=True` implementation and `skipna=False`. It appears that the new algorithm performs worse on a single locale, but surprisingly the performance actually improves when using multiple locales. I have changed the default to `skipna=True` like you suggested, but let me know if you think it should be reverted. Thanks.

| Environment | skipna=False | skipna=True |
| single locale | 0.49 | 1.27 |
| XC 16-locale   | 0.75 | 0.48 |
| CS 16-locale | 0.78 | 0.37 |

src/ReductionMsg.chpl
mcdobe100: Thanks for the suggestion, I think that your proposed implementation is much better, I added that in with some modifications.

src/ReductionMsg.chpl
```diff
@@ -611,11 +619,39 @@ module ReductionMsg
...
+          if si == segments.domain.high {
+              segnans[si] = cumnans[cumnans.domain.high];
+          } else {
+              segnans[si] = cumnans[segments[si+1]-1];
+          }
+        }
+        
+        // take diffs of adjacent segments to find nan count in each segment
+        var nancounts: [segments.domain] int;
+        nancounts[0] = segnans[0];
```
reuster986, line 644: In case the default domain ever changes, we should probably use implicit indexing here, i.e. `nancounts[segments.domain.low] = segnans[segments.domain.low];`

src/ReductionMsg.chpl
```diff
@@ -611,11 +619,39 @@ module ReductionMsg
...
+        nancounts[1..(segnans.domain.size-1)] = segnans[segnans.domain.interior(segnans.domain.size-1)] - segnans[segnans.domain.interior(-(segnans.domain.size-1))];
```
reuster986, line 645: Is there a reason for using the explicit range in the LHS expression? I'm worried that it will trigger some bad access patterns, although @bradcray or @ronawho would know better. I think if we use another interior expression, i.e. `nancounts[nancounts.domain.interior(-(nancounts.domain.size-1))] = ...`, the domains should do the right thing with respect to locality.

bradcray: I'm embarrassed at how quickly my memory has become rusty w.r.t. slicing, but I'm pretty sure that slicing a distributed array with a range will do the correct thing w.r.t. locality (and will continue to in the future).  Recall that other cases where we've tried to move away from slicing with a range were (I believe) ones in which the range described a completely/mostly local set of indices, such that using a default / non-distributed domain rather than a range resulted in performance improvements (?).

A more straightforward way to write this (if I'm reading it correctly) would be to inherit the bounds using:

```chapel
nancounts[1..] = segnans[..segnans.domain.high-1];
```

Seeing Bill's other comment, we may want to make this more index-neutral, though, and use:

```chapel
nancounts[segnans.domain.low+1..] = segnans[..segnans.domain.high-1];
```

(did I translate that correctly?)

That said, I also feel that `segments.domain` and `segnans.domain` seem long-winded.  If I'm reading the code correctly, I think that these domain expressions are the same as `D`?  (which is arguably a bit terse, but...).  If so, and it didn't have any negative performance impacts (I don't think it should), I might use:

```chapel
nancounts[D.low+1..] = segnans[..D.high-1];
```


The following are some other ways I wrote it before arriving at the above:

```chapel
const shiftSize = D.size - 1;
nancounts[D#shiftSize] = nancounts[D#(-shiftSize)];   // copy last n-1 elements into the first n-1
```

Note that `#n` says "give me the first n things" if `n` is positive and "...the final n things" if it's negative.

Thinking about it a moment more, we also ought to be able to write this as:

```chapel
const shiftSize = D.size - 1;
nancounts#shiftSize = nancounts#(-shiftSize);   // copy last n-1 elements into the first n-1
```

though I have a lot less experience applying `#` to arrays (and maybe none to distributed arrays), so am less confident in it (though it _ought_ to do the right thing if implemented correctly).

reuster986: Ah, that makes sense. Based on Brad's comments, I would go with the following: 

```chapel
nancounts[D.low+1..] = segnans[D.low+1..] - segnans[..D.high-1];
```

By the way, that `D#shiftSize` pattern is super cool. I did not know about that.

mcdobe100: Alright, think I think that I have these changes in now. Thanks for the suggestions guys, what you have suggested obviously looks much cleaner than what I have before.

mcdobe100:
Whoops, accidentally force pushed here not realizing you had already done the merge, will try to get that figured out how to undo that

mhmerrill:
@mcdobe100 I would like you to resolve the merge conflicts which could be as simple as doing `git pull origin master` into your branch.

mcdobe100:
@mhmerrill Sounds good, will do. One other thing I wanted to mention is that I found an issue in my performance test earlier that was skewing the numbers and making the skip nan functionality appear faster than the regular implementation, but I no longer believe this to be the case and have collected these numbers that (I believe) to be more accurate than the old. I am not sure if this at all changes your perception of whether or not `skipna` should default to True, but it is something to be aware of.

Execution time of operations ran with 10,000,000,000 elements on 16 locales shown below in seconds (lower is better)
| Operation | skipna=False | skipna=True |
| Mean | 0.39 | 2.13 |
| Min | 3.01 | 3.17 |
| Max | 3.04 | 3.04 |
| Sum | 0.36 | 0.53 |

So for `min`, `max`, and `sum`, the results are very comparable, but there is a significant decrease in performance for `mean`.

reuster986:
Ah, those timings make more sense than the earlier ones. I think it's fine to keep `skipna=True` as the default.

bradcray:
It seems reasonable to me that skipping nans would only hurt performance since it's presumably adding more special cases.  Maybe put another way, if it had been faster, it probably suggests that there's an optimization missing for the normal case that doesn't specialize them.

I missed this until today, but why does the Python interface use `skipna` rather than `skipnan`?  Is this naming inherited from DataFrames?  (and is "na" meant to be read as "N/A"? or the first "Na" in "NaN" or something else?)

I think my main surprise in the table above is that mean and sum execute at such different rates than min and max in the traditional case.  Any guesses what it is about those operations makes them inherently slower?  (My guess: maybe the reliance on a conditional to implement them where sum and mean are just straightline arithmetic code?)

I'm also surprised that the slowdown for mean is so significant (or maybe: so much worse than sum), since I think of mean as being similar to sum in general (though I'm admittedly not certain what it means in the context of GroupBy).

I think the PR's capability is still valuable and worth merging, and am only bringing up these questions out of curiosity, or maybe to suggest additional areas of investigation.

mhmerrill:
@bradcray `skipna` is inherited from Pandas, it is for skipping _missing data_ and I think missing floating point is represented with `NaN`. This is what I gather and interpreted from the Pandas documentation... maybe misinterpreted ;-)

reuster986:
@bradcray yes, `skipna` is from pandas and is meant to be read as "skip N/A", where N/A includes NaN as well as other null values and missing data (as Mike noted). Personally, I would like to implement `skipna` logic for integer arrays as well, but that is a different project (and a more involved one, as there is no standard sentinel value).

I have always wondered why min and max is slower than sum. I don't know whether it is because a `min scan` or `max scan` is inherently slower than a `+ scan` in Chapel, or whether it is the fact that the segmented min/max function in arkouda scans over an array of tuples, whereas sum scans an array of values. Either way, improving min/max performance would be valuable, as it is the slow step in a couple of the workflows I have seen.

To me, the slowdown for mean makes sense because the logic for counting nans within each segment is rather heavy and only gets invoked when `skipna=True`.

mcdobe100:
Alright, think I finally got that merge figured out, much harder than it should have been, but at least I got my first real merge out of the way (even though it was a small one). Sorry if the commits are a little messy, I can try to clean that up with a rebase if you would like. 

I would be happy to look into the performance differences of min and max against sum if you guys would like and could try to get a better idea of where the time is going there.

I also echo what Bill had said about the performance from `mean` being so much more significant than the others, there was a lot added to that function, that is why I was so surprised initially when I was seeing those performance numbers (which turned out to be a mistake on my end).

mhmerrill:
@mcdobe100 are you finished committing to this? is there any other changes I need to review before I do my final pass review and merge.

mcdobe100:
@mhmerrill I think that I am done committing to this for now, unless you come across anything else you would like changed before it goes in

Issue #407, opened 2020-07-03
io_test fails in CI/CD pipeline


hokiegeek2:
Need to figure out the installation process needed to fix this error raised in CI/CD pipeline. Disabled io_test.py until I fix this

![image](https://user-images.githubusercontent.com/10785153/86465283-55c6dd80-bcff-11ea-8d8e-3faa1a9dd544.png)

hokiegeek2:
The reason I did not see this in my testing is that Anaconda automatically install h5ls:

![image](https://user-images.githubusercontent.com/10785153/86591361-6f666000-bf5f-11ea-85da-6e71922ffa71.png)

![image](https://user-images.githubusercontent.com/10785153/86591412-8c9b2e80-bf5f-11ea-9876-b3d6e5faaacb.png)

The CI/CD environment does not use anaconda to manage python dependencies, so this is why the error is currently occurring.

hokiegeek2:
Hmmm...actually, as it turns out, h5ls is indeed installed with hdf5 in arkouda:

![image](https://user-images.githubusercontent.com/10785153/86593376-16003000-bf63-11ea-910b-add66eeb3725.png)

hokiegeek2:
Adding $ARKOUDA_HOME/dep/hdf5-install/bin to the $PATH appears to make the io_test.py run correctly as h5ls is found:

![image](https://user-images.githubusercontent.com/10785153/86602619-e7895180-bf70-11ea-87f4-74d11a0f4265.png)

hokiegeek2:
Still getting an error that h5ls cannot be found:

![image](https://user-images.githubusercontent.com/10785153/86603209-b5c4ba80-bf71-11ea-88c4-167ecddc0829.png)

hokiegeek2:
Okay, now I see why:  the installl-hdf5 goal is not being called in the CI/CD workflow:

![image](https://user-images.githubusercontent.com/10785153/86605087-3f758780-bf74-11ea-9fa3-8489b96eced9.png)

hokiegeek2:
Okay, adding the hdfs-tools install fixes the h5ls missing lib issue:

![image](https://user-images.githubusercontent.com/10785153/86609334-d3961d80-bf79-11ea-8403-f0ebb246ac88.png)

hokiegeek2:
[...]

PR #411, opened 2020-07-08
Added hdf5-tools and setup.py changes needed to support execution of …

hokiegeek2:
…io_test.py in CI/CD workflow, enabled io_test.py #407

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -25,7 +25,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install dependencies
       run: |
-        apt-get update && apt-get install -y libhdf5-dev libzmq3-dev python3-pip
+        apt-get update && apt-get install -y libhdf5-dev libzmq3-dev python3-pip hdf5-tools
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install dependencies
       run: |
-        apt-get update && apt-get install -y libhdf5-dev libzmq3-dev python3-pip
+        apt-get update && apt-get install -y libhdf5-dev libzmq3-dev python3-pip hdf5-tools
```
ronawho, line 28: nit, but I might put `hdf5-tools` next to `libhdf5-dev` since they're logically related

hokiegeek2:
@ronawho That's a good nit! I will make that change.

.github/workflows/CI.yml
hokiegeek2: agree, will make change

.github/workflows/CI.yml
```diff
@@ -53,7 +53,7 @@ jobs:
...
```
ronawho, line 56: Move this one too?

hokiegeek2: Argh, missed that, fixed now

PR #406, opened 2020-07-02
60 add basic authentication

hokiegeek2:
This pull request addresses the following issues:

#60 
#345 
#337 (latest suggestions by @bradcray) 
#338 (in part)

PR #415, opened 2020-07-09
Re-add mac CI testing

ronawho:
It was removed in #300 because the infrastructure seemed unstable, but
this might have caught portability issues in #406, so re-enable it to see
how stable things are.

I couldn't find an easy way to install a specific version with homebrew,
and so to pin to chapel 1.22.0 I followed this article:
https://itnext.io/how-to-install-an-older-brew-package-add141e58d32


ronawho:
Example run: https://github.com/mhmerrill/arkouda/pull/415/checks?check_run_id=855620192

PR #417, opened 2020-07-13
60 add basic authentication final

hokiegeek2:
Final version of branch with latest version of upstream master merged in


arkouda/client.py
```diff
@@ -65,25 +79,39 @@ def connect(server = "localhost", port = 5555, timeout = 0):
...
+    
     # "protocol://server:port"
     pspStr = "tcp://{}:{}".format(server,port)
+    
+    # check to see if tunnelled connection is desired. If so, start tunnel
     tunnel_server = os.getenv('ARKOUDA_TUNNEL_SERVER')
     if tunnel_server:
-        from zmq import ssh
-        (pspStr, _) = ssh.tunnel.open_tunnel(pspStr, tunnel_server)
+        _start_tunnel(addr=pspStr, tunnel_server=tunnel_server)
```
ronawho, line 99: @hokiegeek2 This broke tunneling support (which we use for nightly XC testing where we build/run from a cray elogin node, tunneling to the login node)

Here's a diff that seems to resolve it:

```diff
diff --git a/arkouda/client.py b/arkouda/client.py
index 3c3e8a7..9a969bb 100644
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -96,7 +96,7 @@ def connect(server : str="localhost", port : int=5555, timeout : int=0,
     # check to see if tunnelled connection is desired. If so, start tunnel
     tunnel_server = os.getenv('ARKOUDA_TUNNEL_SERVER')
     if tunnel_server:
-        _start_tunnel(addr=pspStr, tunnel_server=tunnel_server)
+        (pspStr, _) = _start_tunnel(addr=pspStr, tunnel_server=tunnel_server)

     if verbose: print("psp = ",pspStr);

@@ -213,7 +213,7 @@ def _start_tunnel(addr : str, tunnel_server : str) -> None:
     Starts ssh tunnel

     :param str tunnel_server: the ssh url
-    :return: None
+    :return: The url that has been forwarded, and the tunnel object
     :raise: ConnectionError if the ssh tunnel could not be created given the
             tunnel_server url and credentials (either password or key file)
     """
@@ -229,7 +229,7 @@ def _start_tunnel(addr : str, tunnel_server : str) -> None:
         kwargs['password'] = password

     try:
-        ssh.tunnel.open_tunnel(**kwargs)
+        return ssh.tunnel.open_tunnel(**kwargs)
     except Exception as e:
         raise ConnectionError(e)
```

(Basically `ssh.tunnel.open_tunnel` opens the tunnel and returns the new url, but that's currently being dropped on the floor.)

I don't quite grok this change though. In particular it's not obvious to me what `ARKOUDA_KEY_FILE` and `ARKOUDA_PASSWORD` are for in `_start_tunnel`. I haven't followed most of this change overall though so this confusion is likely just ignorance on my part.

hokiegeek2: @ronawho on it now

hokiegeek2: @ronawho I added config of a password or key to integrate authentication into tunneling, which would integrate into zmq authentication eventually and token-based authentication in the near term.

ronawho: > I added config of a password or key to integrate authentication into tunneling, which would integrate into zmq authentication eventually and token-based authentication in the near term.

Ah, gotcha -- thanks

hokiegeek2: @ronawho Sorry for the delay, but I am still trying to figure out how my changes broke tunneling.

hokiegeek2: @ronawho this is what you added:

image

And here's what I replaced it with:

image

And here's the tunnel start code:

image

Still not seeing how my code changes broke anything, still looking at it.

ronawho: `ssh.tunnel.open_tunnel` opens the tunnel and returns the new url, but the value of `ssh.tunnel.open_tunnel` in your patch is dropped on the floor. My diff above just returns the value and sets pspStr again, and that seems to work, I just didn't understand the other changes around this so I wasn't confident that wouldn't break the authentication code.

ronawho: e.g. if I'm running on a cray-xc elogin node doing `python3 tests/check.py nid00076 5555` and `export ARKOUDA_TUNNEL_SERVER=$EPROXY_LOGIN`:

The untunneled psp is `tcp://nid00076:5555`, and the return value of `ssh.tunnel.open_tunnel` is `tcp://127.0.0.1:56793` -- so `ssh.tunnel.open_tunnel` tunnels through the login node, and returns an address/port to access that tunnel to connect to, but since that value is getting dropped on the floor, the elogin node is trying to connect to `tcp://nid00076:5555`, but the elogin node does not have direct/non-tunneled access to that address.

hokiegeek2: Oh, now I see what happened. Right below the tunnel method you added there's an openssh variant:

image

I figured it would be best to switch to that 'cause it's a standard lib and does not alter the URL 'cause it's an external tunneling server. But...I forgot to switch out the methods. That's why it seems I am dropping something on the floor which is rather important. The other odd thing is that I did not see a ConnectionError being thrown, so this failed silently in my case. Odd. Okay, putting in a patch ASAP.

hokiegeek2: @ronawho Yup, understood, I updated the client, pulled upstream master, building now and will test and submit PR as soon as I can.

ronawho: Fixed in https://github.com/mhmerrill/arkouda/pull/444

PR #419, opened 2020-07-13
Enhance Set Operations memory usage and performance

mcdobe100:
Includes improvements to the changes that came in #392 and adds a speed test to test the performance of the set operations without having to remake Arkouda.
Improves speed of `intersect1d` by moving to an explicit `coforall` loop. Also, cuts down on memory usage in intersect1d, union1d, and setxor1d by accessing indices or arrays rather than creating temporary arrays to hold those copies and changes `Indexing.chpl` to reflect these changes as well.

### DIFF
--- a/src/ArraySetops.chpl
+++ b/src/ArraySetops.chpl
@@ -44,11 +44,21 @@ module ArraySetops
     proc intersect1dHelper(a: [] ?t, b: [] t) {
       var aux = radixSortLSD_keys(concatset(a,b));
 
-      var head = sliceHead(aux);
-      var mask = head == sliceTail(aux);
-      
-      var int1d = boolIndexer(head, mask);
+      var mask: [aux.domain] bool;
 
+      coforall loc in Locales do
+        on loc {
+          const localDom = aux.localSubdomain();
+          const maskIndices = localDom#(localDom.size-1);
+          forall i in maskIndices {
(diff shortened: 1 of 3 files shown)


mcdobe100:
@mhmerrill found a bug with `setxor1d()` that `make test-python` was not catching, not sure why, seems like the `setops_test.py` isn't catching a lot of correctness errors, I encountered this a few times here, hoping to figure out why the test isn't catching them. Converting this to a draft for now until I get that bug fixed.

mhmerrill:
> @mhmerrill found a bug with `setxor1d()` that `make test-python` was not catching, not sure why, seems like the `setops_test.py` isn't catching a lot of correctness errors, I encountered this a few times here, hoping to figure out why the test isn't catching them. Converting this to a draft for now until I get that bug fixed.

@mcdobe100 I admit I didn't spend a lot of time crafting test cases for the `setops_test.py` if there are other cases to add please either add them to the python or comvey them in an issue.

mhmerrill:
@mcdobe100 you don't have to submit all of the optimization to the setops as one big PR. I'd actucally prefer it if you did each one  seprately. Unless this causes too much heartburn for you.

mcdobe100:
@mhmerrill no problem! I will close this one and then open them up separately

PR #425, opened 2020-07-14
Improve setxor1d memory usage and performance

mcdobe100:
Added some changes to the slicing used in `setxor1d` and added an extra scope to improve performance and memory usage and also moved `var` to `const` where possible.

Memory: significant improvements to the high mark of memory
`Memory.memoryUsed()` with 10,000 elements on a single locale 
| time of measure | master | improvement |
| memory created in function | 793763 | 447371 |

Performance: slight speed increase
Execution time of 1,000,000,000 elements on 16 locales in seconds (lower is better)
|  | master | improvement |
| execution time (s) | 2.61 | 2.38 |

### DIFF
--- a/src/ArraySetops.chpl
+++ b/src/ArraySetops.chpl
@@ -68,16 +68,18 @@ module ArraySetops
     // sorts and removes all values that occur
     // more than once
     proc setxor1dHelper(a: [] ?t, b: [] t) {
-      var aux = radixSortLSD_keys(concatset(a,b));
+      const aux = radixSortLSD_keys(concatset(a,b));
+      const D = aux.domain;
 
-      var sliceComp = sliceTail(aux) != sliceHead(aux);
+      const sliceComp = aux[..D.high-1] != aux[D.low+1..];
       
       // Concatenate a `true` onto each end of the array
       var flag = makeDistArray((sliceComp.size + 2), bool);
+      const fD = flag.domain;
       
-      flag[0] = true;
-      flag[{1..#(sliceComp.size)}] = sliceComp;
-      flag[sliceComp.size + 1] = true;
+      flag[fD.low] = true;
+      flag[fD.low+1..#fD.high-1] = sliceComp;
+      flag[fD.high] = true;
 
       var mask = sliceTail(flag) & sliceHead(flag);
 
--- a/src/ArraySetops.chpl
+++ b/src/ArraySetops.chpl
@@ -69,16 +69,14 @@ module ArraySetops
     // more than once
     proc setxor1dHelper(a: [] ?t, b: [] t) {
       const aux = radixSortLSD_keys(concatset(a,b));
-      const D = aux.domain;
+      const ref D = aux.domain;
 
-      const sliceComp = aux[..D.high-1] != aux[D.low+1..];
-      
       // Concatenate a `true` onto each end of the array
-      var flag = makeDistArray((sliceComp.size + 2), bool);
-      const fD = flag.domain;
+      var flag = makeDistArray(aux.size+1, bool);
+      const ref fD = flag.domain;
       
       flag[fD.low] = true;
-      flag[fD.low+1..#fD.high-1] = sliceComp;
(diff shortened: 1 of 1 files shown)


src/ArraySetops.chpl
```diff
@@ -68,18 +68,23 @@ module ArraySetops
     // sorts and removes all values that occur
     // more than once
     proc setxor1dHelper(a: [] ?t, b: [] t) {
-      var aux = radixSortLSD_keys(concatset(a,b));
+      const aux = radixSortLSD_keys(concatset(a,b));
+      const D = aux.domain;
 
-      var sliceComp = sliceTail(aux) != sliceHead(aux);
+      const sliceComp = aux[..D.high-1] != aux[D.low+1..];
```
mhmerrill, line 74: @bradcray do the ranges in this line of code have performance implications? I seem to remember that they did when we wrote the initial version of the radix sort.

src/ArraySetops.chpl
```diff
@@ -68,18 +68,23 @@ module ArraySetops
...
       
       // Concatenate a `true` onto each end of the array
       var flag = makeDistArray((sliceComp.size + 2), bool);
+      const fD = flag.domain;
       
-      flag[0] = true;
-      flag[{1..#(sliceComp.size)}] = sliceComp;
-      flag[sliceComp.size + 1] = true;
+      flag[fD.low] = true;
+      flag[fD.low+1..#fD.high-1] = sliceComp;
```
mhmerrill, line 81: seems like the computation to make `sliceComp` should go here unless the const keeps the copy from being made...???

src/ArraySetops.chpl
bradcray: My recollection / expectation is that for a slice of a distributed array `aux` that spans most of the array, this is what we'd want for locality / preserving the distributed nature of the computation.  I think in the previous cases where we expressed slices as anonymous domains (e.g., `A[{lo..hi}]`), we were focused on the case where the slice was purely or largely local to a single locale, where the current range-based slicing implementation results in too much overhead.

src/ArraySetops.chpl
bradcray: Mike: I think you're saying we could write this as:

```chapel
flag[fD.low+1..#fD.high-1] = aux[..D.high-1] != aux[D.low+1..];
```

and eliminate `sliceComp` altogether?  If so, I agree that this seems preferable (should remove an explicit array temp, saving memory without any real negative impact on performance that I can anticipate).  We'd also need to replace the `sliceComp.size + 2` expression, but it looks like this should just be `aux.size + 1` if I'm keeping the logic correct?

Typing that up, I'm noticing that the `#fD.high-1` expression feels a little awkward because I think it presumes a relationship between fD's high bound and its size, which locks in a specific indexing scheme (?).  So I think better would be something like:

```chapel
flag[fD.low+1..fD.high-1] = aux[..D.high-1] != aux[D.low+1..];
```

if I've interpreted and translated that properly.

While looking at that and wondering about the relationship between `fD` and `D`, I think we could also turn those from `const` declarations (which will create a new distributed domain) to `const ref` declarations (which will refer to the existing distributed domain).  I suspect it would have a positive (although possible modest) performance benefit without any real downside.

mhmerrill: @mcdobe100 I'd like you to change the code using @bradcray suggestions.

mcdobe100: Alright, just put those in, thanks for the suggestions, makes sense to me

bradcray: :+1: 

If you haven't, I'd suggest re-running your perf tests to make sure none of our changes have undermined your original motivations for this PR (hopefully they'll only have improved them).

mcdobe100: Reran and updated the description with the new numbers. Saw slight improvements in both categories (speed and memory).

mhmerrill: @mcdobe100 @bradcray Good Work!, seems like the sorts are dominating the running time now. I'll merge at this point.

PR #426, opened 2020-07-16
408 use GitHub pages to host arkouda docs

hokiegeek2:


README.md
```diff
@@ -200,17 +189,29 @@ make
...
+```
+
 ## Testing Arkouda
 
-There are two unit test suites for Arkouda, one for Python and one for Chapel. As mentioned above, the Arkouda Python 
-test harness leverages the [pytest](https://docs.pytest.org/en/latest/) and [pytest-env](https://pypi.org/project/pytest-env/) 
-libraries, whereas the Chapel test harness does not require any external librares.
+There are two unit test suites for Arkouda, one for Python and one for Chapel. As mentioned above, the Arkouda  
+Python test harness leverages multiple libraries such as [pytest](https://docs.pytest.org/en/latest/) and 
+[pytest-env](https://pypi.org/project/pytest-env/) that must be installed via `pip3 install -e .[test]`,
```
ronawho, line 213: `[test]` -> `[dev]`?

hokiegeek2: @ronawho yes, we folded doc and test dependencies into dev

ronawho: Right, I was wondering if this one needs to be updated to reflect that too.

PR #430, opened 2020-07-16
Update setops python test to test against numpy

mcdobe100:
The old setops test had some strange behavior, so I moved it over to a more similar convention used as tests such as the `groupby_test.py` and `nan_test.py`. I removed the support for running the test standalone vs through `make test-python`, but let me know if that is still a feature that is in use and you would prefer to have that added back in.

### DIFF
--- a/tests/setops_test.py
+++ b/tests/setops_test.py
@@ -4,76 +4,69 @@ import pandas as pd
 from context import arkouda as ak
 from base_test import ArkoudaTest
 
-if __name__ == "__main__":
-    import argparse, sys
-    parser = argparse.ArgumentParser(description="Runs and times reductions over arrays in both arkouda and numpy.")
-    parser.add_argument('hostname', help='Hostname of arkouda server')
-    parser.add_argument('port', type=int, help='Port of arkouda server')
-    parser.add_argument('-n', '--size', type=int, default=10**6, help='Problem size: length of array to argsort')
+SIZE = 1000
+OPS = frozenset(['intersect1d', 'union1d', 'setxor1d', 'setdiff1d'])
 
-    args = parser.parse_args()
-    ak.verbose = False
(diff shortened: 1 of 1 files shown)


mcdobe100:
@mhmerrill what do you think about this test change? Do you have any concerns/anything that you would like modified? Thanks.

mhmerrill:
@mcdobe100 I was looking over it yesterday and there were a couple of things to change which I will put comments into the code for... I might get to it today or tomorrow.
Also when you put in a PR you should tag a couple of reviewers under the Reviewers section on the right hand side of the page.

mcdobe100:
@mhmerrill Sounds good, thanks for the feedback, I will make sure to start tagging a few reviewers from now on

tests/setops_test.py
```diff
@@ -4,76 +4,69 @@
...
+    '''
+    akvals = akvals.to_ndarray()
     
-    set_intersection = ak.intersect1d(a,b)
-    print("intersect1d = ", set_intersection.size,set_intersection)
-    # elements in a and elements in b (elements in both a and b)
-    passed = ak.all(ak.in1d(set_intersection,a) & ak.in1d(set_intersection,b))
-    print("intersect1d passed test: ",passed)
+    if not np.array_equal(akvals,npvals):
+        print(f"Different values (abs diff = {np.abs(akvals - npvals).sum()}")
```
mhmerrill, line 27: this computation is not right for this kind of test. you should print the mismatches between results. we used in1d to test the differences in the original code maybe think about how to apply that

mcdobe100: Changed that to hopefully more what you were talking about, where it prints the values that are in the np array, but not the ak array and also the other way around. Let me know if you had something else in mind. Also, it looks like I am not able to tag reviewers myself. After some research, I think I have to be marked as a "collaborator" to be able to request a review

PR #434, opened 2020-07-17
Update groupbyclass.py

### LINKED ISSUE #432: GroupBy on strings fails with unkown symbol error
There is a python-side bug that is causing `GroupBy` on strings to return an uknown symbol error. I believe the bug is in this block:

https://github.com/mhmerrill/arkouda/blob/b75031c3c2e219d550588dd6fe9bb76fbaef1623/arkouda/groupbyclass.py#L104-L125

I think what is happening is `h1` and `h1` are being garbage collected by python before the `generic_msg` call, so the arkouda server deletes the arrays before they are used in `findSegments`. A possible fix would be to add `h1` and `h2` directly to the keynames array instead of just their string names. That might tell python to keep the objects in scope.

For now, there is a workaround: use `ak.GroupBy(..., hash_strings=False)`. That keyword arg bypasses `h1` and `h2`.

mhmerrill:
fixes #432 
take care of out of scope object refs (h1 and h2) from the Strings case which were not alive and deleted before the server operation started
original code is here https://github.com/mhmerrill/arkouda/blob/b75031c3c2e219d550588dd6fe9bb76fbaef1623/arkouda/groupbyclass.py#L104-L125

### DIFF
--- a/tests/groupby_test.py
+++ b/tests/groupby_test.py
@@ -20,8 +20,8 @@ def groupby_to_arrays(df : pd.DataFrame, kname, vname, op, levels):
 def make_arrays():
     keys = np.random.randint(0, GROUPS, SIZE)
     keys2 = np.random.randint(0, GROUPS, SIZE)
-    i = np.random.randint(0, SIZE//GROUPS, SIZE)
-    f = np.random.randn(SIZE)
+    i = np.random.randint(0, SIZE//GROUPS, SIZE)+1
+    f = np.random.randn(SIZE) # normally dist random numbers
     b = (i % 2) == 0
     d = {'keys':keys, 'keys2':keys2, 'int64':i, 'float64':f, 'bool':b}
 
--- a/tests/groupby_test.py
+++ b/tests/groupby_test.py
@@ -11,6 +11,10 @@ verbose = True
 def groupby_to_arrays(df : pd.DataFrame, kname, vname, op, levels):
     g = df.groupby(kname)[vname]
     agg = g.aggregate(op.replace('arg', 'idx'))
+    if op == 'prod':
+        # There appears to be a bug in pandas where it sometimes
+        # reports the product of a segment as NaN when it should be 0
+        agg[agg.isna()] = 0
     if levels==1:
         keys = agg.index.values
     else:
@@ -20,7 +24,7 @@ def groupby_to_arrays(df : pd.DataFrame, kname, vname, op, levels):
 def make_arrays():
     keys = np.random.randint(0, GROUPS, SIZE)
     keys2 = np.random.randint(0, GROUPS, SIZE)
-    i = np.random.randint(0, SIZE//GROUPS, SIZE)+1
+    i = np.random.randint(0, SIZE//GROUPS, SIZE)
     f = np.random.randn(SIZE) # normally dist random numbers
     b = (i % 2) == 0
     d = {'keys':keys, 'keys2':keys2, 'int64':i, 'float64':f, 'bool':b}
--- a/tests/groupby_test.py
+++ b/tests/groupby_test.py
@@ -118,7 +118,7 @@ which enables integration into a pytest test harness.
 class GroupByTest(ArkoudaTest): 
 
     # https://github.com/mhmerrill/arkouda/issues/365
-    @unittest.skip
+    #@unittest.skip
     def test_groupby_on_one_level(self):
         '''
         Executes run_test with levels=1 and asserts whether there are any errors
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -105,7 +105,7 @@ class GroupBy:
             if isinstance(k, Strings):
                 if self.hash_strings:
                     h1, h2 = k.hash()
-                    keyobjs.extend[[h1,h2])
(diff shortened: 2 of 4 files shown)


mhmerrill:
I flipped the default value for `skipna` on `GroupBy.aggregate` from `False` to `True` I believe that this now matches the pandas default behavior which fixes the test failure we were seeing. That said, @reuster986 can `np.random.randn(N)` generate either very large, very small, NaN, or INF values? could this have contributed to the test failure because we didn't have `skipna` before and after we got it until today it was set to `False`.

tests/groupby_test.py
```diff
@@ -118,7 +118,7 @@ def run_test(levels, verbose=False):
 class GroupByTest(ArkoudaTest): 
 
     # https://github.com/mhmerrill/arkouda/issues/365
-    @unittest.skip
+    #@unittest.skip
```
ronawho, line 121: Still fails for me ~1/20 trials. There's probably a one-liner for this, but I'm just testing with:

```bash
#!/usr/bin/env bash

for run in {1..100}; do
  python3 -m pytest tests/groupby_test.py
  if [ $? -ne 0 ]; then
    exit 1
  fi
done
```

mhmerrill: @ronawho can you look for a message on the failure which looks like:
`Different values (abs diff = nan)`
this is what is printed when the answers don't match between pandas and arkouda.
which reduction is failing? Is it always the same one?
for me it was:
```
Doing aggregate(int64, prod)
Different values (abs diff = nan)
```
I ran your script and failed after 15 successes

ronawho: It's not clear to me where the actual failure occurs, but I do see:

```
Doing aggregate(int64, prod)
Different values (abs diff = nan)
```

I updated my script to just print failures to a file:

```bash
for run in {1..100}; do
  python3 -m pytest tests/groupby_test.py &> out.txt
  if [ $? -ne 0 ]; then
    cat out.txt >> failures.txt
    echo "Failure"
  fi
done
```

And here's the log for 4 failures: [failures.txt](https://github.com/mhmerrill/arkouda/files/4940891/failures.txt)

mhmerrill: @reuster986 @mcdobe100 it seems we did not implement `skipna` for the segmented product reduction. Or for segmented `argmin` or `argmax`

https://github.com/mhmerrill/arkouda/blob/f3e75a28ea47e29cb91c7ecc53e85da0161443e8/src/ReductionMsg.chpl#L315-L325
https://github.com/mhmerrill/arkouda/blob/f3e75a28ea47e29cb91c7ecc53e85da0161443e8/src/ReductionMsg.chpl#L588-L607

mhmerrill: @reuster986 it occurred to me that the log(0) is undefined, what if zero is sneaking into the computation. That would produce `NaN` wouldn't it. Maybe the algorithm here takes care of this but I keep seeing NaN from the check routine in the test.

mhmerrill: @reuster986 @ronawho @mcdobe100 so on further investigation it is the `int64` code path where the failures occur, if I add `+1` to the integer data the test passed 100 time in a row.

mhmerrill: ok now it passed 200 times in a row.

reuster986: Ah, so we probably just need logic that checks for zeros in the segment and short-circuits them. I will work on this.

reuster986: @mhmerrill regarding your latest commit, I think we should keep the zeros in the int64 test data. The `prod` reduction needs to be able to handle zeros -- I'm writing logic for this now, and I would like the test data to cover that case.

mhmerrill: @reuster986 I was going to put an issue in to cover this and decouple the logic fix from this PR so we would not delay these changes. Unless you are going to get the logic change done and tested... today or tomorrow.

mcdobe100: @mhmerrill Yes, that is correct, we only implemented `skipna` for `min`, `max`, `sum`, and `mean`. Would you like it to be extended to those operations as well? The reason we only went with those 4 is that the original intent for `skipna` was motivated by Roland's use, which only required the use of those 4, but I think it makes sense to have it implemented for all of them if you want that done as well.

reuster986: @mcdobe100 The code I'm testing right now includes `skipna` for `prod`. Since I was writing code for handling zeros, it was relatively easy to do the same for NaNs at the same time.

reuster986: I think I discovered a bug in pandas. When doing `groupby` followed by `.aggregate('prod')`, pandas will sometimes compute a `nan` when the product should be zero. I discovered this by manually generating a dataframe with int64 columns, doing a groupby-aggregate with `prod`, and finding a group with `nan` product. Then, I manually sliced out the group from the original dataframe and computed the true product, which was 0. 

I believe the pandas bug is the root cause of the intermittent test failures. In the latest commit, I modified `groupby_test.py` to replace `nan` with 0 in the pandas `prod` result, and I ran the 100-fold test with no failures.

reuster986:
Ok, I think this PR is ready to go.

Issue #437, opened 2020-07-20
io_test failures on distributed systems


ronawho:
I/O tests like io_test.py are writing to /tmp, but /tmp is usually a filesystem that's local to each machine. For clusters and other cases where the client and and server aren't running on the same machine, this creates problems. 

I'd probably just use a subdir of the arkouda directory to ensure the directory is shared between the login and computes notes.

ronawho:
FYI @hokiegeek2 we see this for our nightly Cray CS/XC testing

hokiegeek2:
@ronawho gotcha, okay, the easiest thing to do would be to add an env variable to make the io test directory configurable. I can do that, or if you want to, that's cool also.

ronawho:
If you have time, that'd be great. It'll be a while until I have a chance.

hokiegeek2:
@ronawho cool, I'm on it

ronawho:
Fixed by https://github.com/mhmerrill/arkouda/pull/439 -- thanks @hokiegeek2!

PR #439, opened 2020-07-21
replaced /tmp dir with cwd for distributed testing, set one security_…

hokiegeek2:
…test to skip until I can figure out why shutil in script will not delete the .arkouda hidden directory #437


mhmerrill:
is this failing in CI or something... is it urgent?

hokiegeek2:
@mhmerrill It would be good to get 'em in if possible 'cause they are causing tests on the Crays to fail.

hokiegeek2:
Whoops, okay, there in there, thanks

PR #445, opened 2020-07-22
Improve filename extension detection in write1DDistArray

ronawho:
Previously, the code was just splitting on `.` and assuming anything to
the right of that delimiter was an extension, but that's not always
true. In particular in some of our nightly testing we have directories
with `.` in them, so update the check to not consider something an
extension if it has a path separator in it.

For example our nightly testing path to arkouda is something like
`performance-test-perf.chapcs.arkouda/test/studies/arkouda/arkouda`,
where the `.` in `performance-test-perf.chapcs.arkouda` was skewing this
check previously.

### DIFF
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -1,6 +1,7 @@
 module GenSymIO {
   use HDF5;
   use IO;
+  use Path;
   use MultiTypeSymbolTable;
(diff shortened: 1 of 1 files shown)


mhmerrill:
@ronawho I've added @reuster986 as a reviewer because he may have an opinion about parsing filenames and what should be done there.

PR #448, opened 2020-07-29
Add mink and maxk reductions

mcdobe100:
Able to get the maximum or minimum `k` values in an array by calling `ak.mink(pda, kval)`, `pda.mink(kval)`, `ak.maxk(pda,kval)`, `pda.maxk(kval)`.

Implemented with a user defined reduction where each task has it's own heap-like data structure (kextreme) and then merges them together, keeping the extreme (min or max) values. Data structure acts like a heap until it needs to combine, then sorts the values, giving up its heap property, to perform the merge.

Timings collected at 16 nodes of an XC with 10,000,000,000 elements
Execution time in seconds (similar for both `maxk` and `mink`):
| k val | maxk (s)| 
| ----------- | -----: | 
| 10,000     | 0.21 | 
| 100,000   | 0.55 |
| 1,000,000    | 2.84 |
| 10,000,000 | 23.19 |


arkouda/extrema.py
```diff
@@ -0,0 +1,83 @@
+from arkouda.client import generic_msg, verbose
```
mhmerrill, line 1: @mcdobe100 do we need this file since you added these to the` pdarrayclass.py` file?

arkouda/reductions.py
```diff
@@ -0,0 +1,83 @@
+from arkouda.client import generic_msg, verbose
```
mhmerrill, line 1: this is the third definition of `mink` `maxk`

arkouda/extrema.py
mcdobe100: Looks like you are right, we do not need this file! I guess that I had a misunderstanding of how the Python side worked, sorry about that. Pushing up the removal of these files now

arkouda/reductions.py
mcdobe100: Oops! Thanks for catching that, I think this was a residual file from a name change, I need to look over these branches a little closer before opening the PR's!

mhmerrill:
@mcdobe100 things are looking better. My vision for this was to be able to skim off the top-K scores from a float64 array (something which you computer from some scoring function) and effectively return the indices (like argmin/argmax)... I don't know how hard this would be to extend what you have to do this?

mcdobe100:
@mhmerrill OK, I see what you are saying, definitely makes sense. I don't think that it should be too difficult to implement from what I have, hopefully I will have some time to at least get started on that this week, and if not, then I will definitely be able to take a look next week. Thanks for the feedback.

mcdobe100:
@mhmerrill After diving into this issue a little bit more, what I'e been thinking is that there would be a `mink()` and a `minkIndices()` where you can get either the values or the indices back, since the value portion has already been implemented. How does that sound to you? Also, do you have any preferences for the naming of the `minkIndices()` function? (minkIndices might be a little long...)

mcdobe100:
Now that I have thought about it a little more... maybe it makes more sense to take a third argument like `isInd` or something like that which would determine whether or not the user wanted the indices

mhmerrill:
We have reductions named like pdarray.argmin() or argmax() which return the index of the min/max value. Why not something like argmink().

mcdobe100:
@mhmerrill Alright, I have changed it to have `argmink()` and `argmaxk()` for the indices and `mink()` and `maxk()` for the values. Changing to also hold the indices made things quite a bit slower, and I think there are some potential optimizations that could be done to improve performance a bit, but I think that this has all of the functionality we were going for (I think at least). Let me know what you think about this and whether we should do the optimizations (if any) before merging or maybe merge this first and then do some optimizations as a separate PR. Thanks.

mcdobe100:
@mhmerrill While doing some optimizations for this, the question came up of whether or not you would like this to return an Arkouda or a NumPy array. Do you have a preference for how we should handle this?

pytest.ini
```diff
@@ -1,20 +1,6 @@
 [pytest]
 testpaths =
-    tests/client_test.py
```
mhmerrill, line 3: @mcdobe100 you need to add these tests back in so CI can run them

mhmerrill:
> @mhmerrill While doing some optimizations for this, the question came up of whether or not you would like this to return an Arkouda or a NumPy array. Do you have a preference for how we should handle this?

these should always return arkouda pdarrays.

mcdobe100:
Some more notes on this:
- The mink/maxk/argmink/argmaxk reduction significantly outperforms sorting the entire array for small (<1,000,000) `k` values
- As `k` gets higher (observed about 5 million on a CS and about 3 million on an XC), it will get to a point where sorting the array (also dependent on the size of the array) can be faster
- Adding more locales with a large `k` value will hurt performance 

This is happening for a few reasons:
1. The current Chapel user defined reductions perform their combine step on a single locale, so each task will be sorting on the same locale and using a lock on that locale, creating a bit of a bottleneck there
2. As `k` increase, `heapifyDown()` will not be utilizing the cache efficiently, resulting in lots of random access

With where Chapel intends to move this user defined reduction framework, this will hopefully see performance improvements in the future and there may be more ways to tune this in the future, but I think it is hard to improve it much more from where it is now currently.

arkouda/pdarrayclass.py
```diff
@@ -6,7 +6,7 @@
 from arkouda.dtypes import structDtypeCodes, NUMBER_FORMAT_STRINGS
 
 __all__ = ["pdarray", "info", "any", "all", "is_sorted", "sum", "prod", "min", "max",
-           "argmin", "argmax", "mean", "var", "std"]
+           "argmin", "argmax", "mean", "var", "std", "mink", "maxk", "argmink", "argmaxk"]
```
mhmerrill, line 9: @reuster986 we have created some new entry points here are you ok with the names we chose? `ak.argmink` to find the indices of the minimum k values in an array and so on.

arkouda/pdarrayclass.py
```diff
@@ -809,3 +837,189 @@ def std(pda, ddof=0):
...
+    --------
+    >>> A = ak.array([10,5,1,3,7,2,9,0])
+    >>> ak.mink(A, 3)
+    array([0, 1, 2])
+    """
+    if isinstance(pda, pdarray):
+        if k == 0:
+            return []
+        if pda.size == 0:
+            raise TypeError("must be a non-empty pdarray {} of type int".format(pda))
```
mhmerrill, line 881: message says "...of type int" I thought we were supporting int64 and float64 dtypes?

arkouda/pdarrayclass.py
```diff
@@ -809,3 +837,189 @@ def std(pda, ddof=0):
...
+    --------
+    >>> A = ak.array([10,5,1,3,7,2,9,0])
+    >>> ak.maxk(A, 3)
+    array([7, 9, 10])
+    """
+    if isinstance(pda, pdarray):
+        if k == 0:
+            return []
+        if pda.size == 0:
+            raise TypeError("must be a non-empty pdarray {} of type int".format(pda))
```
mhmerrill, line 928: same comment as before... change the message text.

arkouda/pdarrayclass.py
```diff
@@ -809,3 +837,189 @@ def std(pda, ddof=0):
...
+    --------
+    >>> A = ak.array([10,5,1,3,7,2,9,0])
+    >>> ak.argmink(A, 3)
+    array([7, 2, 5])
+    """
+    if isinstance(pda, pdarray):
+        if k == 0:
+            return []
+        if pda.size == 0:
+            raise TypeError("must be a non-empty pdarray {} of type int".format(pda))
```
mhmerrill, line 974: same comment as before... change the message text

arkouda/pdarrayclass.py
```diff
@@ -809,3 +837,189 @@ def std(pda, ddof=0):
...
+    --------
+    >>> A = ak.array([10,5,1,3,7,2,9,0])
+    >>> ak.argmaxk(A, 3)
+    array([4, 6, 0])
+    """
+    if isinstance(pda, pdarray):
+        if k == 0:
+            return []
+        if pda.size == 0:
+            raise TypeError("must be a non-empty pdarray {} of type int".format(pda))
```
mhmerrill, line 1021: same comment as before... change the message text

src/KExtremeMsg.chpl
```diff
@@ -0,0 +1,140 @@
...
+    use ArraySetopsMsg;
+
+    /*
+    Parse, execute, and respond to a intersect1d message
+    :arg reqMsg: request containing (cmd,name,name2,assume_unique)
+    :type reqMsg: string
+    :arg st: SymTab to act on
+    :type st: borrowed SymTab
+    :returns: (string) response message
+    */
```
mhmerrill, line 30: make sure this reads correctly. it currently talks about intersect1d.

src/KExtreme.chpl
```diff
@@ -0,0 +1,110 @@
...
+ * to a heap data structure, but can be
+ * sorted and merged to extract the
+ * extreme values from multiple kextreme
+ * objects.
+ */
+
+module KExtreme {
+  use SymArrayDmap;
+  use Sort;
+  record kextreme {
```
mhmerrill, line 14: i'd like a little more documentation describing the record and maybe how it gets used with the chapel reduction framework, also since it's a definition I think Id like it to follow the rest of the code and use uppercase camel case, I know that would make it collide with the module name... maybe rename the module.

mhmerrill: it would also be good to make the doc chpldoc/sphinx formatted.

src/KExtremeMsg.chpl
```diff
@@ -0,0 +1,140 @@
...
+
+        select(gEnt.dtype) {
+          when (DType.Int64) {
+             var e = toSymEntry(gEnt,int);
+
+             var aV;
+             if !stringtobool(returnIndices) {
+               aV = computeExtrema(e.a, k:int, false);
+             } else {
+               aV = computeInds(e.a, k:int, false);
```
mhmerrill, line 103: `computeInds` is too generic of a name for me maybe use `computeExtremaInds`

src/KExtremeMsg.chpl
```diff
@@ -0,0 +1,140 @@
...
+             return s;
+            }
+           }
+           otherwise {
+             return notImplementedError("mink",gEnt.dtype);
+           }
+        }
+    }
+
+    proc maxkMsg(cmd: string, payload: bytes, st: borrowed SymTab): string throws {
```
mhmerrill, line 85: doc comments?

src/KReduce.chpl
```diff
@@ -0,0 +1,110 @@
...
+ * operation is equivalent to performing an `argsort()`
+ * on an array, and then using a slice of these values
+ * to retrieve the first or last `k` indices from that
+ * sorted indice array.
+ */
+
+module KReduce {
+  use KExtreme;
+
+  class kreduce : ReduceScanOp {
```
mhmerrill, line 26: doc comments

mhmerrill: also a couple of comments on how this works with the chapel reduction framework.

mhmerrill: or a pointer to the chapel documentation which describes it.

mhmerrill:
@mcdobe100 I left a bunch of comments mostly about comments and documentation, otherwise I think it looks pretty good.

arkouda/pdarrayclass.py
reuster986: I'm good with these names. Pandas calls these `nlargest` and `nsmallest`, but I think `mink` and `maxk` are easier to remember. I also like the addition of `argmink` and `argmaxk`, which pandas does not have but which are very useful, in my opinion.

PR #449, opened 2020-07-30
LisExpr.chpl updates for 1.22, and in general

bradcray:
This PR updates LisExpr.chpl to work with Chapel 1.22 while also making some other minor changes that @mhmerrill mentioned today.  I'm sorry to say that this was more painful than I'd expected, which made it a really good exercise for me, and resulted in my filing a lot of new issues (see the source code for URLs).

To get things working with 1.22 again, the primary changes were:
  - adapting to additional rigor within the compiler w.r.t. lists of non-nilable owned classes (see discussion below)
  - pop_front() is no longer supported on arrays, so I switched to a list of strings (which is what Mike wanted anyway)
  - switching from 1- to 0-based indexing when evaluating the parsetrees
  - adding a now-necessary import of `IO` to get access to `stdout`

In addition, I switched a lone remaining case of `list(GenListValue)` to `GenList`.  Mike, as we discussed, this was happily just a matter of moving the `type` definition up in the code to ensure that it was evaluated before code used it.

Getting the lists of non-nilable owned to work with 1.22 turned out to be nontrivial.  Challenges included:

In several places the code did `listOfOwned1 = listOfOwned2` in either an initialization or assignment context, which the compiler now (rightfully) complains about.  The reason for this is that it's not clear what should happen to the RHS class instances.  Ownership usually transfers with assignment which means that the RHS list would become useless (i.e., we wouldn't really be copying it as intended, but demolishing its elements as we went).  In addition, since it's a list of non-nilable owned, it's not clear what the list element's values should be set to after the ownership is transferred (the general reason a non-nilable owned can't be a RHS for an assignment).

Since we know we want a deep copy of the list, the approach I took was to create a helper function that would make a new list by calling `copy()` on all of the original list's elements.  This turned out to be a fair amount of work, though:

* I wrote the `copyOwnedList()` routines themselves
* I had to implement a `copy()` method on the classes stored in lists that didn't already have one
* I ended up adding `isListType()` queries because I couldn't seem to get an `==` looking for the specific type working correctly (I suspect I could with more effort, but this seemed easier in the end)

The lack of assignment between lists also meant that I had to complicate the `ListValue.init()` method to handle the list case separately and after the initialization had completed since it isn't a simple assignment-based initialization (I could've handled both list and non-list cases after initialization had been completed, but still would've had to insert a conditional)

I also had to work around some ergonomic issues with lists, such as:

* they can't be initialized using an iterator expression, so I had to loop over it explicitly, adding the elements to the list
* I couldn't chain one call that returned a list to another that wanted to take it by reference without inserting a temp

I think the new version is reasonable, but I think it could be a lot more reasonable/usable if we were to address the issues that I opened as a result of this exercise.

### DIFF
--- a/toys/LisExpr.chpl
+++ b/toys/LisExpr.chpl
@@ -12,6 +12,9 @@ module LisExpr
     
     type Symbol = string;
 
+    /* type: list of genric list values */
+    type GenList = list(owned GenListValue);
+
     /* generic list value */
     class GenListValue
     {
@@ -32,8 +35,9 @@ module LisExpr
 
         proc copy(): owned GenListValue throws {
           select (this.lvt) {
-            when (LVT.Sym) {
-              var copyList = copyOwnedList(this.toListValue(list(owned GenListValue)).lv);
+            when (LVT.Lst) {
+              //              var copyList = copyOwnedList(this.toListValue(list(owned GenListValue)).lv);
+              var copyList = copyOwnedList(this.toListValue(GenList).lv);
               return new owned ListValue(copyList);
             }
             when (LVT.Sym) {
@@ -103,9 +107,6 @@ module LisExpr
     */
         
     
-    /* type: list of genric list values */
-    type GenList = list(owned GenListValue);
-
     // allowed value types int and real
     enum VT {I, R};
 
@@ -329,21 +330,38 @@ module LisExpr
       parse, check, and validate code and all symbols in the tokenized prog
     */ 
     proc parse(line: string): owned GenListValue throws {
-        return read_from_tokens(tokenize(line));
+                                                         var l: list(string) = tokenize(line);
+                                                         
+        return read_from_tokens(l);
     }
     
     /*
       parse throught the list of tokens generating the parse tree / AST
       as a list of atoms and lists
     */
-    proc read_from_tokens(in tokens: list(string)): owned GenListValue throws {
+    proc read_from_tokens(ref tokens: list(string)): owned GenListValue throws {
+                                                                                //writeln("In read_from_tokens(", tokens, ")");
         if (tokens.size == 0) then
             throw new owned ErrorWithMsg("SyntaxError: unexpected EOF");
+        //                                                                                                                                                                    writeln("pre-pop: ", tokens);
+                                                                                                                                                                    
         var token = tokens.pop(0);
+        //                                                                                                                                                                    writeln("post-pop: ", tokens);
+                                                                                                                                                                    
         if (token == "(") {
+          //          writeln("inside conditional: ", tokens);
+          
             var L: GenList;
+            //            writeln("before while loop: ", tokens);
+            
             while (tokens.first() != ")") {
+              //              writeln("inside while loop: ", tokens);
+              
+              //              writeln("calling recursively with: ", tokens);
+              
                 L.append(read_from_tokens(tokens));
+                //                writeln("returned from read_from_tokens()");
+                
                 if (tokens.size == 0) then
                     throw new owned ErrorWithMsg("SyntaxError: unexpected EOF");
             }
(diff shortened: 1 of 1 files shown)


bradcray:
Beyond @mhmerrill, I wanted to tag @dlongnecke-cray and @mppf on this PR because of the interactions with lists and non-nilable and owned and return intents.  I think it's worth reading through the diff and related issues, and thinking about what could improve quality of life for this case.

toys/LisExpr.chpl
```diff
@@ -261,32 +325,50 @@ module LisExpr
...
         if (token == "(") {
             var L: GenList;
-            while (tokens[D.low] != ")") {
+            while (tokens.first() != ")") {
                 L.append(read_from_tokens(tokens));
                 if (tokens.size == 0) then
                     throw new owned ErrorWithMsg("SyntaxError: unexpected EOF");
             }
-            tokens.pop_front(); // pop off ")"
+            tokens.pop(0); // pop off ")"
```
bradcray, line 371: With all of Chapel's index-neutral features, it seems a shame that the only way to remove an element from the front of a list is by using an explicit index value...  This might be worth another issue as well.

toys/LisExpr.chpl
```diff
@@ -349,36 +431,36 @@ module LisExpr
...
             when (LVT.Lst) {
-                var lst = ast.toListValue(GenList).lv;
+                ref lst = ast.toListValue(GenList).lv;
                 // no empty lists allowed
                 checkGEqLstSize(lst,1);
                 // currently first list element must be a symbol of operator
-                checkSymbol(lst[1]);
-                var op = lst[1].toListValue(Symbol).lv;
+                checkSymbol(lst[0]);
+                var op = lst[0].toListValue(Symbol).lv;
```
bradcray, line 439: These could arguably be lst.first() to avoid the explicit indices...

toys/LisExpr.chpl
```diff
@@ -349,36 +431,36 @@ module LisExpr
...
-                    when "<"  {checkEqLstSize(lst,3); return eval(lst[2], env) < eval(lst[3], env);}
-                    when ">"  {checkEqLstSize(lst,3); return eval(lst[2], env) > eval(lst[3], env);}
-                    when "<="  {checkEqLstSize(lst,3); return eval(lst[2], env) <= eval(lst[3], env);}
-                    when ">="  {checkEqLstSize(lst,3); return eval(lst[2], env) >= eval(lst[3], env);}
-                    when "==" {checkEqLstSize(lst,3); return eval(lst[2], env) == eval(lst[3], env);}
-                    when "!=" {checkEqLstSize(lst,3); return eval(lst[2], env) != eval(lst[3], env);}
-                    when "or" {checkEqLstSize(lst,3); return or(eval(lst[2], env), eval(lst[3], env));}
-                    when "and" {checkEqLstSize(lst,3); return and(eval(lst[2], env), eval(lst[3], env));}
-                    when "not" {checkEqLstSize(lst,2); return not(eval(lst[2], env));}
+                    when "+"  {checkEqLstSize(lst,3); return eval(lst[1], env) + eval(lst[2], env);}
```
bradcray, line 441: These would need to have some sort of helper, similar to the ones we built for Arkouda, to pull out elements in an index-neutral manner.  Or something like https://github.com/chapel-lang/chapel/issues/15445.

toys/LisExpr.chpl
dlongnecke-cray: We could totally add a method called `popFront` to list.

dlongnecke-cray:
List changes look OK to me. Glad to see there are several questions about `list` ergonomics being raised.

mhmerrill:
@bradcray all the copying of the ast/list during creation really bothers me. every time we recurse during the parse we make a complete deep copy when we really just want a reference I think.

mppf:
> In several places the code did listOfOwned1 = listOfOwned2 in either an initialization or assignment context, which the compiler now (rightfully) complains about. The reason for this is that it's not clear what should happen to the RHS class instances. Ownership usually transfers with assignment which means that the RHS list would become useless (i.e., we wouldn't really be copying it as intended, but demolishing its elements as we went). In addition, since it's a list of non-nilable owned, it's not clear what the list element's values should be set to after the ownership is transferred (the general reason a non-nilable owned can't be a RHS for an assignment).

Another way to solve this would be with getting the swap operator for list to swap elements (without copying anything).

bradcray:
> all the copying of the ast/list during creation really bothers me. every time we recurse during the parse we make a complete deep copy when we really just want a reference I think.

@mhmerrill: I think that's a reasonable concern, though I don't believe my PR has introduced any deep copies that weren't there previously, does it?  Specifically, the `read_from_tokens()` routine takes the list by reference, so modifies it in place (and in fact, I had a bug at first before I added this intent).  And I think the other places that I've added a `copyOwnedList()` routine were already copying the list, the compiler was just (arguably incorrectly) supporting those copies via `=` in Chapel 1.20.  But maybe I missed something?

toys/LisExpr.chpl
mppf: `popFirst` ?

mhmerrill:
> > all the copying of the ast/list during creation really bothers me. every time we recurse during the parse we make a complete deep copy when we really just want a reference I think.
> 
> @mhmerrill: I think that's a reasonable concern, though I don't believe my PR has introduced any deep copies that weren't there previously, does it? Specifically, the `read_from_tokens()` routine takes the list by reference, so modifies it in place (and in fact, I had a bug at first before I added this intent). And I think the other places that I've added a `copyOwnedList()` routine were already copying the list, the compiler was just (arguably incorrectly) supporting those copies via `=` in Chapel 1.20. But maybe I missed something?

@bradcray I think I am worried about the line linked below because we are returning back up the recursion and I worry that we are getting O(n**2) behavior from all the copying... maybe I am not understanding correctly, please help explain.
https://github.com/bradcray/arkouda/blob/d626a60f18ef86439c586d1efaef897104dab3aa/toys/LisExpr.chpl#L372

toys/LisExpr.chpl
```diff
@@ -261,32 +325,50 @@ module LisExpr
...
             var L: GenList;
-            while (tokens[D.low] != ")") {
+            while (tokens.first() != ")") {
                 L.append(read_from_tokens(tokens));
                 if (tokens.size == 0) then
                     throw new owned ErrorWithMsg("SyntaxError: unexpected EOF");
             }
-            tokens.pop_front(); // pop off ")"
+            tokens.pop(0); // pop off ")"
             return new owned ListValue(L);
```
bradcray, line 372: @mhmerrill: This is the line you're asking about right?  Note that, in this diff, I did not modify it.  That's what makes me say that although it's a potential problem, it's a pre-existing one, not anything specific to what I did in this PR.  Again, unless I'm missing something subtle.

mhmerrill: Yes, you didn't change this line but you drasticly changed the `ListValue.init` and the `GenListValue.init` making the deep copy. I also think I am gaining an understanding of why you made the change. My question is how to do this and not make the deep copy... The problem is that I create an `owned GenList L` then add a bunch of stuff to it then use it to initialize a `ListValue(L)` that is where all the copying happens. Maybe we should create/initialize the `new owned ListValue(L)` when the list is still empty before we populate it in line 367, and then line 367 might look like `LV.L.append(read_from_tokens(tokens));`
Am I making sense? Also, could we change this to a `shared` to get around this copying problem`

toys/LisExpr.chpl
mhmerrill: does `pop` equal `popLast`?

bradcray: Yeah, at present.  I personally like popFirst and popLast better than popHead and popTail because they seem less ambiguous w.r.t. order (and, depending on the implementation, expense).  I believe lists support .first and .last queries already, so this would also be more uniform.

mhmerrill: so is `last` the first thing you appended? and `first` the last thing you appended? or the other way around?

bradcray: For me, the key is to think of lists as being neither stack nor queue.  Do I know what a list is, in real life?  (Yeah).  Do I know what the first thing is?  (Yeah).  Do I know what the last thing is?  (Yeah).  What does appending something to the list do?  (Add a new entry to the end / after the last thing).  Pop is admittedly weird given all this and wouldn't have been my choice for a list operation.  I don't recall what the rationale there was, though I expect it related to following in some other language's footsteps and/or "What is the cheapest thing to pop off of a list [for our current implementation]".  That said, we're discussing having multiple implementation options under the hood, which I think would make simple `pop()` go away and require `popFirst` vs. `popLast` vs. `pop(i)`.

bradcray: I captured some of my takeaways from this conversation here:  https://github.com/chapel-lang/chapel/issues/16194

PR #451, opened 2020-08-03
Capture stderr in lshdfMsg

ronawho:
Previously, lshdfMsg was doing something like `h5ls in_file > out_file`
so stderr was lost, which made it hard to identify failures in our
nightly testing. Update it so that stderr is written to the output file
as well and then if there are errors we'll report them back to the user.

In our nightly testing we used to just see `Error:` and now we get
`Error: /bin/sh: h5ls: command not found` (which helps us figure out
that the error is on our end with h5ls not being in our PATH)


ronawho:
FYI @hokiegeek2 since I think Brad pinged you about the errors we were seeing in some I/O tests.

PR #453, opened 2020-08-12
Adjust + reduce on atomics for Chapel POI change

vasslitvinov:
Arkouda runs `+ reduce` on arrays of atomic ints. For that, it redefines `+` and `+=` on atomics in user code to read() from each atomic int. Starting with https://github.com/chapel-lang/chapel/pull/16158 this does not work any longer.

Instead, implement this reduction using a forall loop with a reduce intent, reading the atomics within the loop. This works with both Chapel 1.22 release and Chapel master.

Testing: `make test-python test-chapel` with 1.22 homebrew and with Chapel master under gasnet.

Minor performance improvement left as future work: specialize the pattern:
```chpl
var hist = makeDistArray(bins,int);
hist = lHist;
return hist;
```
for the case when `MyDmap` == `Dmap.defaultRectangular`, by returning `lHist` (or `gHist`) directly.

### DIFF
--- a/src/Histogram.chpl
+++ b/src/Histogram.chpl
@@ -7,15 +7,6 @@ module Histogram
 
     use PrivateDist;
     use SymArrayDmap;
-
-    // need + and += defined to support reduction on atomic int 
-    proc +(x: atomic int, y: atomic int) {
-        return x.read() + y.read();
-    }
-    // need + and += defined to support reduction on atomic int     
-    proc +=(X: [?D] int, Y: [D] atomic int) {
-        [i in D] {X[i] += Y[i].read();}
-    }
     
     /*
     Takes the data in array a, creates an atomic histogram in parallel, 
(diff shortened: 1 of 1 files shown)


bradcray:
Hi @mhmerrill — Arkouda has been breaking in testing against our master branch this week due to changes to how point of instantiation is handled for generic functions to prevent surprises and hijacking.  This change will fix it, so would be nice to merge if you're open to it.  The reason POI comes into play is that the code implementing the reduction doesn't necessarily have access to the scope where the Arkouda-overloaded + operations are defined.  This is a question we're continuing to wrestle with ("How should visibility of operators be handled?") and relates to a question I asked you a month or so ago about whether you had preferences between expressing operator overloads as standalone functions or methods.  Bottom line: I don't think the design questions here are fully resolved yet, but it would be helpful to us to merge this PR to keep our testing working and on-track if you're OK with it.

bradcray:
In discussing this, Mike asked "Should the language just have support for `+` and `+ reduce` on atomics?" which led me to file https://github.com/chapel-lang/chapel/issues/16239 on the Chapel repo

vasslitvinov:
Mike, thank you for reviewing my two PRs.

PR #468, opened 2020-08-31
cosine distance example

mhmerrill:
example of using arkouda to compute cosine distance between two arkouda pdarrays.

### DIFF
--- a/examples/cos_dist.py
+++ b/examples/cos_dist.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python3
+
+import arkouda as ak
+import math
+
+# dot product or two arkouda pdarrays
+def ak_dot(u, v):
+    return ak.sum(u*v)
+
+# magnitude/L_2-norm of arkouda pdarray
+def ak_mag2(u):
+    return math.sqrt(ak_dot(u,u))
+
+# cosine distance of two arkouda pdarrays
+# should function similarly to scipy.spatial.distance.cosine
+def ak_cos_dist(u, v):
+    return (1.0 - ak_dot(u,v)/(ak_mag2(u)*ak_mag2(v)))
+
+if __name__ == "__main__":
+    import argparse, sys, gc, time
+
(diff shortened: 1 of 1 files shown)


examples/cos_dist.py
```diff
@@ -0,0 +1,48 @@
...
+# cosine distance of two arkouda pdarrays
+# should function similarly to scipy.spatial.distance.cosine
+def ak_cos_dist(u, v):
+    return (1.0 - ak_dot(u,v)/(ak_mag2(u)*ak_mag2(v)))
+
+if __name__ == "__main__":
+    import argparse, sys, gc, time
+
+    parser = argparse.ArgumentParser(description="Example of cosine distance/similarity in arkouda")
+    parser.add_argument('hostname', help='Hostname of arkouda server')
```
hokiegeek2, line 23: I recommend re-naming the hostname arg to server to match the Arkouda Python API

examples/cos_dist.py
```diff
@@ -0,0 +1,48 @@
...
+    parser.add_argument('port', type=int, help='Port of arkouda server')
+    args = parser.parse_args()
+
+    ak.v = False
+    ak.connect(server=args.hostname, port=args.port)
+
+    u1 = [1, 0, 0]
+    v1 = [0, 1, 0]
+    d1 = ak_cos_dist(ak.array(u1), ak.array(v1))
+    print("d1 = ", d1)
```
hokiegeek2, line 33: May make sense to put assert statements in here in addition to or in lieu of print statements

examples/cos_dist.py
mhmerrill: I assume you wanted me to call this argument `'server'`

hokiegeek2: Indeed! :)

examples/cos_dist.py
mhmerrill: added asserts and compare against `spicy.spatial.distance.cosine` result

Issue #465, opened 2020-08-25
save with mode=append to non-existent file throws non-intuitive error


hokiegeek2:
If a user mistakenly attempts to append a non-existent file, the following error is thrown:

![image](https://user-images.githubusercontent.com/10785153/91173481-e8567e00-e6ab-11ea-983d-0d6a2ce87d8c.png)

There's two paths:
1. Throw error stating something like "Cannot append a non-existent file"
2. Create the file, write the pdarray(s), and return a warning or a notification message that a new file was created

hokiegeek2:
Development complete, PR merged.

PR #476, opened 2020-09-03
implemented new Error class and catch block to provide intuitive erro…

hokiegeek2:
…r message #465


mhmerrill (approved):
change looks good

PR #478, opened 2020-09-16
Fix brew install of Chapel 1.22 in mac CI testing

ronawho:
There's no direct way to install a specific version of software with
brew that I can find, but we want to pin to Chapel 1.22.0 in order to
test with the Chapel release Arkouda officially supports. Brew used to
be able to install the 1.22.0 formula directly from a URL, but that is
longer supported for security reasons, so instead `curl` the file
locally and do a local install.


ronawho:
Originally reported in https://gitter.im/ArkoudaProject/community?at=5f6221d3a9c2c8511e9bbb2a

mhmerrill (approved):
looks ok to me

PR #485, opened 2020-09-22
Optimize DstAggregator by caching remote allocations

ronawho:
Previously, the DstAggregator would allocate a remote buffer every time
the aggregator was flushed. Here we cache allocations in order to
eliminate an on-stmt and a PUT in the common case, which can
significantly decrease the amount of communication.

The logic to do this caching is encapsulated in a remoteBuffer wrapper
so that we can reuse it for optimizing the SrcAggreator too. Early
implementations had non-trivial overhead, especially for the small
problem size at 256 nodes so this is careful to avoid any additional
communication from the wrapper and as a result I did pretty extensive
benchmarking to check for issues. I expect to be able to clean up the
implementation over time, but wanted to keep the initial implementation
more explicit.

Caching allocations results in some non-trivial performance
improvements, especially for high core count systems. We see an 80%
improvement for scatter on 32 nodes of a 128 core HDR IB system.

Performance Summary:
---

Benchmark results for scatter and argsort were collected for the default
problem size (3/4 GiB arrays) and a large problem size (8GiB arrays) on
3 different systems:
 - 32 node 56Gb/s FDR IB with 36 cores per node
 - 32 node 200Gb/s HDR IB with 128 cores per node
 - 256 node Aries with 36 cores per node

The following tables summarize the results in terms of % speedup for the
small and large problem sizes:

#### Scatter:

| System         | small | large |
| -------------- | ----- | ----- |
| 32 node FDR    |  19%  | 33%   |
| 32 node HDR    |  25%  | 82%   |
| 32 node Aries  |   9%  | 15%   |
| 256 node Aries |   3%  | 13%   |

#### Argsort:

| System         | small | large |
| -------------- | ----- | ----- |
| 32 node FDR    |  13%  | 16%   |
| 32 node HDR    |  27%  | 40%   |
| 32 node Aries  |   7%  | 10%   |
| 256 node Aries |   1%  |  3%   |

Scalability graphs:
---

#### FDR IB with large problem size:

Stable improvements, nothing too interesting.

<details>

ak-scatter-perf-cs-fdr-lg
ak-argsort-perf-cs-fdr-lg

</details>

#### HDR IB with small problem size:

Argsort performance is interesting (bad) at 32 nodes, but this is
because of known scalability issues with the SrcAggregator. I expect
that to improve significantly with some upcoming optimizations to the
SrcAggregator.

<details>

ak-scatter-perf-cs-hdr
ak-argsort-perf-cs-hdr

</details>

#### HDR IB with large problem size:

Nothing too interesting and the sort performance is good because the
problem size is large enough that the SrcAggregator overhead is a
smaller fraction of the total time.

<details>

ak-scatter-perf-cs-hdr-lg
ak-argsort-perf-cs-hdr-lg

</details>

#### Aries with large problem size:

Stable improvements, nothing too interesting.

<details>

ak-scatter-perf-xc-lg
ak-argsort-perf-xc-lg

</details>

Communication counts:
---

Here's a summary of the comm count difference for sorting. GETs are high
because we're using unorderedCopy for the SrcAggreator under ugni.  For
the size=10 case the comm counts are the same as before, which is good
because there's really no flushing going on, so this just shows there's
no extra base communication introduced. The small problem size has a
non-trivial reduction on PUTs/ONs, and difference is more substantial
for the large problem size.

```
CHPL_FLAGS="-sperfOnlyCompile -sperfValRange='uint(32)'" make test-bin/UnitTestSort
./test-bin/UnitTestSort --elemsPerLocale=$((<size>)) -nl 256 --printDiagsSum

10**1
-----
before : ( 1.61s) -- GETS: 1_218_279_871, PUTS:   9_545_670, ONS:   9_604_575
after  : ( 1.58s) -- GETS: 1_218_279_646, PUTS:   9_545_662, ONS:   9_604_567

10**8
-----
before : ( 4.97s) -- GETS: 1_218_715_234, PUTS:  37_736_430, ONS:  37_795_335
after  : ( 4.74s) -- GETS: 1_218_715_232, PUTS:  28_336_110, ONS:  28_395_015

2**30
-----
before : (34.16s) -- GETS: 1_218_715_228, PUTS: 282_144_750, ONS: 282_203_655
after  : (32.88s) -- GETS: 1_218_715_230, PUTS: 150_540_269, ONS: 150_599_174
```

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -40,8 +40,15 @@ module CommAggregation {
     const bufferSize = dstBuffSize;
     const myLocaleSpace = LocaleSpace;
     var itersSinceYield: int;
-    var lBuffers: [myLocaleSpace][0..#bufferSize] aggType;
+    var lBuffers: [myLocaleSpace] [0..#bufferSize] aggType;
+    var rBuffers: [myLocaleSpace] remoteBuffer(aggType);
     var bufferIdxs: [myLocaleSpace] int;
+
+    proc postinit() {
+      for loc in myLocaleSpace {
+        rBuffers[loc] = new remoteBuffer(aggType, bufferSize, loc);
+      }
+    }
 
     proc deinit() {
       flush();
@@ -49,7 +56,7 @@ module CommAggregation {
 
     proc flush() {
       for loc in myLocaleSpace {
-        _flushBuffer(loc, bufferIdxs[loc]);
+        _flushBuffer(loc, bufferIdxs[loc], freeData=true);
       }
     }
 
@@ -69,7 +76,7 @@ module CommAggregation {
       // other tasks run, yield so that we're not blocking remote tasks from
       // flushing their buffers.
       if bufferIdx == bufferSize {
-        _flushBuffer(loc, bufferIdx);
+        _flushBuffer(loc, bufferIdx, freeData=false);
       } else if itersSinceYield % maxItersBeforeYield == 0 {
         chpl_task_yield();
         itersSinceYield = 0;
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
These changes look good.

PR #488, opened 2020-09-23
Improve behavior of extrema_test.py when there's a runtime error

ronawho:
Previously, it would report the error, but then try to use undefined
vars in compare_results.


mhmerrill (approved):
looks good.

Issue #480, opened 2020-09-18
"Too many open files" error with many strings


reuster986:
When loading more than ~ 1000 HDF5 files with strings arrays (does not seem to happen with int64 or float64), the C HDF5 library throws an error with the message "too many open files". I suspect this is related to the way in which `get_dataset_info` tests for a string: by trying to open the name as a dataset, detecting error, then trying to open the name as a group. This might be preventing the file from closing.

https://github.com/mhmerrill/arkouda/blob/a1f219e82b2a2807b7eb96e262961d23c769a3d1/src/GenSymIO.chpl#L531-L552

https://github.com/mhmerrill/arkouda/blob/a1f219e82b2a2807b7eb96e262961d23c769a3d1/src/GenSymIO.chpl#L556-L560

If this is indeed the root of the problem, then we should find the appropriate way of using the HDF5 API to determine whether the name is a dataset or a group, so we don't have to fail the `H5Dopen`.

hokiegeek2:
@reuster986 do you see the "The dataset is neither at the root of the HDF5 file nor within a group" in the stack trace?

hokiegeek2:
@reuster986 Not sure, but these lines either could be the culprit now or in the future, because errors are thrown prior to the file being closed on line 552

https://github.com/mhmerrill/arkouda/blob/a1f219e82b2a2807b7eb96e262961d23c769a3d1/src/GenSymIO.chpl#L519-L525

https://github.com/mhmerrill/arkouda/blob/a1f219e82b2a2807b7eb96e262961d23c769a3d1/src/GenSymIO.chpl#L552

hokiegeek2:
Just submitted a [PR](https://github.com/mhmerrill/arkouda/pull/489) that hopefully fixes this.

mhmerrill:
@hokiegeek2 should we close this one also?

hokiegeek2:
@mhmerrill Yes, I believe so.

PR #489, opened 2020-09-24
refactored code to prevent error, put file close in defer block, adde…

hokiegeek2:
…d file lose statements to two other catch blocks to ensure the file is closed irrespective of function invocation result #480


reuster986 (approved):
looks good!

PR #492, opened 2020-09-24
Fix some bugs in the mink/maxk implementation

ronawho:
The mink/maxk code maintains a heap while accumulating reduction values
and then as an optimization at the end it gets rid of the heap and sorts
to improve how quickly different tasks can merge their mink/maxk data.
However, it's possible for some accumulation to happen after combining, so we
need to handle that case. For arkouda's use case it's only ever doing final
accumulations with the identity value, so we can just ignore them.

See https://github.com/chapel-lang/chapel/issues/16459 for more context

### DIFF
--- a/src/Extremas.chpl
+++ b/src/Extremas.chpl
@@ -47,30 +47,26 @@ to increase the efficiency of the merge step.
     // it is an encountered extreme
     proc push(val: (eltType, int)) {
 
-      // Common case where _data is a heap
-      if !isSorted {
-        const shouldAdd = if isMinReduction then val(0)<_data[0](0) else val(0)>_data[0](0);
-        const shouldAddEmpty = (numEmpty>0) && if isMinReduction then
-          _data[0](0)==max(eltType) else _data[0](0)==min(eltType);
-
-        if shouldAddEmpty {
-          _data[numEmpty+1] = val;
-          numEmpty-=1;
-        } else if shouldAdd {
-          _data[0] = val;
-          heapifyDown();
-        }
-      } else {
-        // Uncommon for an accumulate to be called after combining, so does not
-        // need to be very fast/optimized.
-        if isMinReduction && val < _data[dom.high] {
-          _data[dom.high] = val;
-          sort(_data);
-        } else if !isMinReduction && val > _data[dom.low] {
-          _data[dom.low] = val;
-          sort(_data);
-        }
+      // Accumulate/push can be called after combine/merge. This routine
+      // expects the data to be a heap and not sorted, so it would fail if we
+      // tried to insert now. The way Arkouda uses this, val will be the
+      // identity values, so we can ignore it.
+      if isSorted {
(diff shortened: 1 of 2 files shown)


ronawho:
Easiest to review ignoring whitespace

src/Extremas.chpl
```diff
@@ -46,16 +46,31 @@ to increase the efficiency of the merge step.
...
+        if shouldAddEmpty {
+          _data[numEmpty+1] = val;
+          numEmpty-=1;
+        } else if shouldAdd {
+          _data[0] = val;
+          heapifyDown();
+        }
+      } else {
+        // Uncommon for an accumulate to be called after combining, so does not
+        // need to be very fast/optimized.
```
ronawho, line 65: For Arkouda's use of this reduction, the accumulates after combine will always be with an identity value so nothing will actually be done. This could be `ifSorted then assert(val == if isMinReduction then (max(eltType), -1) else (min(eltType), -1))`, but it wasn't much harder to do this.

src/KReduce.chpl
```diff
@@ -67,6 +67,7 @@ module KReduce {
     }
 
     proc generate() {
+      v.doSort();
```
bradcray, line 70: Is this necessary in the common case?  It seems like it will add overhead.  I'd be inclined to move the .doSort() into any "accumulate-after-combine" branches since we expect that to be rare and so aren't as worried about the performance there (or else, have that branch set a flag that a doSort() is still required in order to do it here).

ronawho: This is only needed when doing serial reductions. Arkouda doesn't do this, but I ran into this for some standalone reproducers. `doSort` is a no-op if things are already sorted, so there's no real overhead.

ronawho: Oh, actually that's a lie, there is no check in `doSort` the `isSorted` check is in the merge code.

mhmerrill: Is this ready to be merged?

mhmerrill: do we have a test case that exhibits the previous bug?

mhmerrill:
@ronawho probably should have left the comment her instead s0 is this ready to be merged and do we have a test case that tests for the fixed bug?

src/Extremas.chpl
ronawho: In our perf meeting we decided that for now it's probably best to just turn this into an assert so that if it's ever hit you'd know and can address it. Currently it'll work, but will be really slow.

ronawho:
> @ronawho probably should have left the comment her instead s0 is this ready to be merged and do we have a test case that tests for the fixed bug?

Not ready to be merged, this is failing in our 16 node testing that uses chapel nightly.

ronawho:
Ok, think it's ready for review

mhmerrill (approved):
looks good.

Issue #498, opened 2020-09-29
HDF5 errors (non-fatal) on non-string data


reuster986:
@hokiegeek2 after you fixed the strings issue with not closing the HDF5 file, now I'm getting a ton of HDF5 errors to stderr when I read in an int64 or float64 dataset from HDF5. Errors look like:

```
HDF5-DIAG: Error detected in HDF5 (1.10.2) thread ...:
  #000: H5G.c line 472 in H5Gopen2(): unable to open group
    major: Symbol table
    minor: Inappropriate type
  #001: H5Gint.c line 294 in H5G__open_name(): not a group
    major: Symbol table
    minor: Inappropriate type
```

There must be a way in HDF5 to tell whether an object is a group or a dataset before trying to open it, right? I'm concerned these errors will leave open files and cause that same "too many files" error with numeric data instead of strings.

hokiegeek2:
@reuster986 whoa, on it now.

hokiegeek2:
@reuster986 Okay, I've fixed the issue of throwing the error that was resulting in the error you posted:

![image](https://user-images.githubusercontent.com/10785153/94679061-7a762580-02ed-11eb-9105-4d6ae79517fc.png)

It's still throwing an error that I am now catching withtin the isDatasetWithinGroup function, so I am working out the syntax to get the object info so I can differentiate between group and dataset and therefore precent the SystemError being thrown.

hokiegeek2:
Added exists check to see if the strings  <dsetName>/values object exists in the hdf5 file within a method called isStringsDataset. If true, then /values and /segments datasets are processed. If false, then just dsetName is processed.

Also added logic to quiet the HDF5 error message which is printed when the exists check is executed. Consequently, Arkouda prints/handles an HDF5-related error message only when an actual error is thrown. 

![image](https://user-images.githubusercontent.com/10785153/94804848-7e6f7980-03b9-11eb-9893-2bbcd83df71e.png)

The following snapshots show the logic is working correctly and that no actual errors are thrown:

![image](https://user-images.githubusercontent.com/10785153/94804752-5bdd6080-03b9-11eb-8360-f48bd44b9b19.png)

hokiegeek2:
Developed and tested fix, documented here, PR merged, issue resolved.

PR #501, opened 2020-09-30
Initial fix with added error handling to prevent open file descriptor…

hokiegeek2:
… buildup #498

this is the initial version to fix immediate issue, will follow-up with version that checks for dataset vs group to prevent the HDF5 error.


reuster986 (approved):
looks good, thanks!

PR #505, opened 2020-10-01
450 add dtype field to strings array like and pdarray

hokiegeek2:


arkouda/strings.py
```diff
@@ -89,7 +91,8 @@ def __init__(self, offset_attrib : Union[pdarray,np.ndarray],
             self.ndim = self.offsets.ndim
             self.shape = self.offsets.shape
         except Exception as e:
-            raise ValueError(e)
+            raise ValueError(e)   
+        self.dtype = self.bytes.dtype
```
mhmerrill, line 95: doesn't this need to change to `self.dtype = np.str_` or something

hokiegeek2: LOL, yeah, git fail on the part of your favorite contractor.

arkouda/strings.py
```diff
@@ -32,6 +32,8 @@ class Strings:
         The rank of the array (currently only rank 1 arrays supported)
     shape : tuple
         The sizes of each dimension of the array
+    dtype : dtype
+        The dtype is np.str_
```
mhmerrill, line 36: missed this should be `np.str`

hokiegeek2: @mhmerrill good catch, just fixed it

reuster986 (approved):
Looks good to me!

PR #506, opened 2020-10-02
adding new marker version of picture

mhmerrill:
adding new marker version of picture for more contrast and better text placement


reuster986 (approved):
very cool!

PR #507, opened 2020-10-02
Enable SrcAggregator under ugni

### LINKED ISSUE #318: Enable SrcAggregation under ugni
See https://github.com/mhmerrill/arkouda/issues/301#issuecomment-595522280 for some motivation.

In Chapel 1.20, ugni yields while waiting for comm to complete in order to enable comm/compute overlap. Unfortunately, this can artificially increase the lifetime of tasks. When aggregating at scale (256-512 nodes) this can result in enough concurrent tasks to run out of memory. In https://github.com/mhmerrill/arkouda/pull/209 we worked around that for the DstAggregator, but not the SrcAggregator.

The performance of the SrcAggregator is not far ahead of unorderedCopy for 8-byte items, but it should be much faster for 1-byte copies.

Switch to source aggregation for ugni, ensure there's no scalability issues, and collect gather.py benchmark results for int and bool dtypes.

ronawho:
Previously, the SrcAggregator was not used under ugni because of memory
pressure at scale. Those issues were resolved in #503.

At 256 nodes for the default problem size (3/4 GiB arrays) this improves
the performance for gathers of int/real by ~1.5x, and gathers of bools
by ~25x. Previously, we were just using unorderedCopy which offers good
performance on XC, but aggregation is better now. For the bool case,
even though we were calling unorderedCopy the Aries NIC can't do RDMA on
things that aren't 4-byte aligned, so we had to fall back to a slow path
that used ordered bounce-buffers. Because of this aggregation offers a
much more significant improvement for bools.

There may still be unaligned comm for the last flush of the bool
aggregation buffer, but the common case will be aligned and it's a bulk
transfer anyway so the performance hit isn't very noticeable. For that
reason I just removed some TODOs to align the flush, instead of rounding
the transfer up.

Perf results for gather on 256 nodes using the default problem size (3/4
GiB arrays):

| type  | before      | after       |
| ----- | ----------: | ----------: |
| bool  |   2.1 GiB/s |  55.3 GiB/s |
| int64 | 271.7 GiB/s | 398.8 GiB/s |

Note that the time for bool/int gather is the same now, the bool rate is
just lower since we're transferring 1/8 as much data.

Scalability graphs for int64 gather:
---

<details> 

ak-gather-perf-xc
ak-gather-perf-xc-lg

</details>

Resolves #318

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -22,8 +22,7 @@ module CommAggregation {
 
   /* Creates a new source aggregator (src/rhs will be remote). */
   proc newSrcAggregator(type elemType, param useUnorderedCopy=false) {
-    // SrcAggregator is not currently optimized for ugni
-    if CHPL_COMM == "none" || CHPL_COMM =="ugni" || useUnorderedCopy {
+    if CHPL_COMM == "none" || useUnorderedCopy {
       return new SrcUnorderedAggregator(elemType);
     } else {
       return new SrcAggregator(elemType);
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
looks good

PR #509, opened 2020-10-05
Optimize aggregators by yielding more often

ronawho:
Yield every 1024 operations instead of 4096. We yield occasionally so
we're not blocking other tasks from flushing their buffers. Chapel will
yield while doing on-statements, so this is just an optimization to help
remote tasks make progress sooner. The old value was tuned a while back
and some recent aggregation optimizations mean yielding more often
benefits performance.

This improves gather and scatter benchmarks by ~5% or so. Longer term we
should probably autotune the buffer sizes and yield frequency, but for
now this is a better default.

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -2,11 +2,10 @@ module CommAggregation {
   use SysCTypes;
   use CPtr;
   use UnorderedCopy;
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
This looks good.

PR #512, opened 2020-10-05
348 replace python print statements with logging logger

hokiegeek2:


mhmerrill:
@hokiegeek2 you are going to make the changes we talk about here.

- [ ] factory to make logger objects to bury stuff like verbose flag and getting env vars
- [ ] removing redundant code from different files

mhmerrill:
@hokiegeek2 can i still use `ak.verbose` to turn on/off debug messages in the code?
like the idiom:
```python
ak.verbose = True
ak.ones(10)
ak.verbose = False
```

hokiegeek2:
> @hokiegeek2 can i still use `ak.verbose` to turn on/off debug messages in the code?
> like the idiom:
> 
> ```python
> ak.verbose = True
> ak.ones(10)
> ak.verbose = False
> ```
> 
> maybe we could call it `ak.debug` instead?

@mhmerrill No, you would change the logging level to DEBUG, which would enable all debug messages to be written out. But, I can add helper methods named something like "enableVerbose/disableVerbose" and use that to set the logger handlers to DEBUG.

mhmerrill:
> > @hokiegeek2 can i still use `ak.verbose` to turn on/off debug messages in the code?
> > like the idiom:
> > ```python
> > ak.verbose = True
> > ak.ones(10)
> > ak.verbose = False
> > ```
> > 
> > 
> > maybe we could call it `ak.debug` instead?
> 
> @mhmerrill No, you would change the logging level to DEBUG, which would enable all debug messages to be written out. But, I can add helper methods named something like "enableVerbose/disableVerbose" and use that to set the logger handlers to DEBUG. Please lemme know if that's cool.

it's cool

hokiegeek2:
> > > @hokiegeek2 can i still use `ak.verbose` to turn on/off debug messages in the code?
> > > like the idiom:
> > > ```python
> > > ak.verbose = True
> > > ak.ones(10)
> > > ak.verbose = False
> > > ```
> > > 
> > > 
> > > maybe we could call it `ak.debug` instead?
> > 
> > 
> > @mhmerrill No, you would change the logging level to DEBUG, which would enable all debug messages to be written out. But, I can add helper methods named something like "enableVerbose/disableVerbose" and use that to set the logger handlers to DEBUG. Please lemme know if that's cool.
> 
> it's cool

Cool, I will...

image

hokiegeek2:
@mhmerrill please take another look when you get a chance

mhmerrill (approved):
looks good

Issue #316, opened 2020-03-05
Include error class name in error messages


reuster986:
When the main server loop catches an error that is not a sublcass of `ErrorWithMsg`, it sends a generic error message over the socket, which is not very helpful to the user. Is there a way to print the class name of the error and send a response like "Error: <class UndefinedSymbolError>" or something? Not sure if this works:
```chpl
try {
    // main server select statement
} catch e {
    repMsg = "Error: %s".format(e.type:string);
}
```

Alternatively, we could go through the arkouda source code and make all custom error classes subclass `ErrorWithMsg` and set their own name as part of the message.

mhmerrill:
@bradcray Maybe someone on the chapel team could point us to a solution? Is there a way to use the Reflection module to do this???

bradcray:
I'm not sure reflection can help here because it tends to provide access to information known at compile-time ("What fields are in this record?"  "If I were to make this call, can the compiler resolve it?"), whereas in this case, I'm guessing it's the dynamic type of the class that you want to reason about.

In Chapel 1.21 and on master, all errors now store strings, but I'm not familiar enough with that effort to know what to suggest here (e.g., maybe the default value of that message will help?).  Tagging @dlongnecke-cray, @ben-albrecht, and @mppf who probably have a better grasp on the current status than I do.

mhmerrill:
I think I am going to just fix the exceptions in the arkouda server code making them all subclass the `ErrorWithMsg` class

dlongnecke-cray:
That seems like the best approach to me, for the time being. I can't think of a particular way to circumvent it.

ben-albrecht:
> I'm guessing it's the dynamic type of the class that you want to reason about.

I wonder if this is a reasonable feature request for the long-term.

bradcray:
> I wonder if this is a reasonable feature request for the long-term.

I'm not sure what's being requested here.  It can't be for the compiler to reason about the class's dynamic type (because the compiler can only reason about static information).

ben-albrecht:
I was thinking a function to return the dynamic type of a class at runtime (as a string?). 

That said, embedding the error type in the default message is another reasonable feature request - then you could query off of `e.message()`.

mppf:
The internal function `chpl_error_type_name` uses some compiler primitive magic to get the class name. It seems that we could make a similar function available in Reflection - it's mainly a matter of settling on function names & API description.

mhmerrill:
[...]

PR #513, opened 2020-10-07
446 enable rethrowing of errors and passing in strings to errors

hokiegeek2:


reuster986:
@hokiegeek2 I like what you've done here! I like the `ErrorWithContext` class, and I think that will be really helpful and could be a "one-stop shop" for arkouda errors thrown all the way up to the main server loop.

I see now what you mean about the difficulty of getting the class name. Given that, I think it is fine to leave out the class name, as long as we discipline ourselves to include a good error message whenever we throw one. For example, in `GenSymIO.chpl`, instead of throwing a `DatasetNotFoundError`, we could throw a `ErrorWithContext` with the message "dataset {name} not found in file {filename}", which I think would be more helpful than knowing the class name anyway. What do you think?

hokiegeek2:
@reuster986 Cool, I think this will help us throw errors that are easier to investigate. I think throwing errors as ErrorWithContext with a detailed message is a good idea. The only caveat would be situations where we want to vary what we do in the error catch blocks depending upon the error conditions; in cases like these we may need child classes of ErrorWithContext.

mhmerrill:
@hokiegeek2 I like this also.
We need to figure out how to incorporate this with the ServerErrorStrings.chpl.
I'd be willing to incorporate this but I think in light of this we need to rework all the other exceptions.
maybe we should do this in several increments? so we don't change ~27 files all at once?

hokiegeek2:
@mhmerrill Yeah, increments is a good idea. Perhaps I can scope this to GenSymIO.chpl, end-to-end from Error through to message sent back to client. Once you and @reuster986 are good with that, we can roll out the design and implementation to the rest of Arkouda.

mhmerrill:
@hokiegeek2 sounds good.

hokiegeek2:
@mhmerrill @reuster986 Please take another look at this PR when y'all get a chance. I have pretty much all of the revamped error handling logic in place for GenSymIO (need to do one more check) and this PR now covers #316.

mhmerrill:
> @mhmerrill @reuster986 Please take another look at this PR when y'all get a chance. I have pretty much all of the revamped error handling logic in place for GenSymIO (need to do one more check) and this PR now covers #316.

@hokiegeek2 are you ready to take this out of the draft state?

mhmerrill (commented):
I think the changes look good, will wait on Bill's review also. Should we move this PR out of the draft state?

hokiegeek2:
> I think the changes look good, will wait on Bill's review also. Should we move this PR out of the draft state?

Cool, yes, I recommend moving this PR out of draft state.

reuster986 (approved):
looks good!

hokiegeek2:
@reuster986 @mhmerrill Thanks a bunch for reviewing!

Issue #348, opened 2020-04-08
Replace Python print statements with logging.Logger


hokiegeek2:
Moving from Python print statements to leveraging the logging.Logger API will enable differentiating error and warning-level logging from debug/info, control over where the log messages go (stdout, file, export to external system such as Elasticsearch or Fluentd, set global logging level, etc...

hokiegeek2:
[PR](https://github.com/mhmerrill/arkouda/pull/512) submitted

hokiegeek2:
development and testing complete, merged PR, issue resolved.

hokiegeek2:
development complete, merged into master, issue resolved.

PR #514, opened 2020-10-08
Added ak.enableVerbose and ak.disableVerbose methods to fully emulate…

hokiegeek2:
… verbose global variable, added unit test case, cleaned up file write/remove for logging_test #348


mhmerrill (approved):
looks better

PR #518, opened 2020-10-14
initial implementation of the ak.clear() function and a test for it

### LINKED ISSUE #499: Add a convenience function to clear specific or all unregistered data from server symbol table.
After working with Arkouda from a notebook I realized it would be very nice to have a `ak.clear()` function to clear out a specific `pdarray` or all unregistered `pdarray`s from the Arkouda server symbol table.

Maybe the interface looks like:
```python
import arkouda as ak
#...
# to clear out a specific pdarray
ak.clear(pda)
# or
ak.clear('id_123')
# or 
# to clear out all unregistered pdarrays
ak.clear(ak.AllSymbols)
```

mhmerrill:
Fixes #499 
This PR implements the `ak.clear()` function which clears the server symbol table of all unregistered data.

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -146,6 +146,14 @@ module MultiTypeSymbolTable
                 if (v) {writeln("deleteEntry: unkown symbol ",name);try! stdout.flush();}
             }
         }
+
+        /*
+        Clears all unregistered entries from the symTable
+        */
+        proc clear() {
+            for n in tab { deleteEntry(n); }
+        }
+
         
         /*
         Returns the sym entry associated with the provided name, if the sym entry exists
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -6,7 +6,7 @@ from arkouda.dtypes import *
 from arkouda.dtypes import structDtypeCodes, NUMBER_FORMAT_STRINGS
 from arkouda.logger import getArkoudaLogger
 
-__all__ = ["pdarray", "info", "any", "all", "is_sorted", "sum", "prod", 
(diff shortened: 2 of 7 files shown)


hokiegeek2 (approved):

hokiegeek2 (approved):

src/MultiTypeSymbolTable.chpl
```diff
@@ -146,6 +146,14 @@ module MultiTypeSymbolTable
                 if (v) {writeln("deleteEntry: unkown symbol ",name);try! stdout.flush();}
             }
         }
+
+        /*
+        Clears all unregistered entries from the symTable
+        */
+        proc clear() {
+            for n in tab { deleteEntry(n); }
```
ronawho, line 154: I don't think this idiom is legal since you're removing from a map (which could shrink it) while iterating over it.

I think `for n in tab.keysToArray() { deleteEntry(n); }` would work though. 

@daviditen and/or @dlongnecke-cray would know best though.

mhmerrill: @bradcray @daviditen @ronawho is this a bug? really?

ronawho: Most languages don't allow this. e.g. you can't remove from a python dictionary while iterating over it:

```python
mydict = {'one': 1, 'two': 2, 'three': 3}
for k in mydict.keys():
   if k == 'two':
       del mydict[k]
```

```
RuntimeError: dictionary changed size during iteration
```

But you can do `for k in list(mydict.keys()):`

daviditen: I agree that removing from a map while iterating over it is problematic. It's iterating over a fixed range while potentially shrinking the storage the range represents.

mhmerrill: ok, i'll make the change.

Issue #520, opened 2020-10-15
io_test failures at 16 locales after #517


ronawho:
I'm seeing some failures for io_test.py at 16 locales after https://github.com/mhmerrill/arkouda/pull/517.

```
ARKOUDA_NUMLOCALES=16 python3 -m pytest tests/io_test.py
```

```
====================================================== short test summary info ======================================================
FAILED tests/io_test.py::IOTest::testLoad - AssertionError: False is not true
FAILED tests/io_test.py::IOTest::testReadAllWithGlob - AssertionError: False is not true
FAILED tests/io_test.py::IOTest::testSaveAllLoadAllWithDict - AssertionError: False is not true
FAILED tests/io_test.py::IOTest::testSaveAllLoadAllWithList - AssertionError: False is not true
============================================= 4 failed, 10 passed in 257.59s (0:04:17) ==============================================
```

Full log:

<details>

```
$ ARKOUDA_NUMLOCALES=16 python3 -m pytest tests/io_test.py
======================================================== test session starts ========================================================
platform darwin -- Python 3.6.9, pytest-5.4.1, py-1.8.1, pluggy-0.13.1
rootdir: /Users/eronagha/Desktop/arkouda, inifile: pytest.ini
plugins: env-0.6.2
collected 14 items

tests/io_test.py ...F..F..FF...                                                                                               [100%]

============================================================= FAILURES ==============================================================
__________________________________________________________ IOTest.testLoad __________________________________________________________

self = <io_test.IOTest testMethod=testLoad>

    def testLoad(self):
        '''
        Creates 1..n files depending upon the number of arkouda_server locales with three columns
        AKA datasets, loads each corresponding dataset and confirms each corresponding pdarray
        equals the input pdarray.

        :return: None
        :raise: AssertionError if the input and returned datasets (pdarrays) don't match
        '''
        self._create_file(columns=self.dict_columns,
                          prefix_path='{}/iotest_dict_columns'.format(IOTest.io_test_dir))
        result_array_tens = ak.load(path_prefix='{}/iotest_dict_columns'.format(IOTest.io_test_dir),
                                    dataset='int_tens_pdarray')
        result_array_hundreds = ak.load(path_prefix='{}/iotest_dict_columns'.format(IOTest.io_test_dir),
                                        dataset='int_hundreds_pdarray')
        result_array_float = ak.load(path_prefix='{}/iotest_dict_columns'.format(IOTest.io_test_dir),
                                     dataset='float_pdarray')
[...]

PR #523, opened 2020-10-15
Added ndarray conversion and ndarray.sort to fix pytest errors at 11+…

hokiegeek2:
… locales #520

### DIFF
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -18,12 +18,21 @@ class IOTest(ArkoudaTest):
 
     def setUp(self):
         ArkoudaTest.setUp(self)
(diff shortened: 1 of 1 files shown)


ronawho (approved):
Passes testing for me. 

Just FYI since the io_tests seem to be somewhat prone to issues at different locale counts I was thinking about adding an additional 16-node CI config that just ran the I/O tests. Unfortunately it added 20 minutes to the CI to run `ARKOUDA_NUMLOCALES=16 python3 -m pytest tests/io_test.py`, and that amount of time for every CI run doesn't seem worthwhile to me.

PR #530, opened 2020-10-19
Recommend Chapel 1.23 and use it for CI testing

ronawho:
Chapel 1.23 has a number of performance optimizations (see #516) and
build times should be 60-80% faster.

Change the documentation to require 1.23, but only recommend it in the
Makefile instead of making it an error. Arkouda should continue to work
with Chapel 1.22 so existing users don't necessarily have to upgrade,
but I would recommend it for the performance gains.


mhmerrill (approved):
looks good

PR #532, opened 2020-10-19
Improve the performance of string gathers under ugni

ronawho:
This partially reverts #275. In #275 the string gather implementation was
split to optimize for ugni vs. non-ugni configurations. At the time pure
unorderedCopy was faster than localizing and using aggregation but
that's no longer true.

Back then, `newSrcAggregator` actually mapped to `unorderedCopy` under
ugni because of scalabiltiy isues with the SrcAggregator under ugni. In
#507 that was changed so that `newSrcAggregator` actually maps to
`SrcAggregator` now. That provided significant speedups for gathering
remote things, especially single-byte items like the `uint(8)` used to
implemented segmented strings. There have also been performance
improvements to the DstAggregator since then. All that to say that while
in the #275 timeframe unorderedCopy was faster for string gathers, that
is no longer the case, so use aggregation.

Here's some performance results from the machine we run the nightly
16-node-xc testing on (the old implementation took too long for the
large problem size):

```
make test-bin/StringGatherSpeedTest
./test-bin/StringGatherSpeedTest --N=1_000_000   --MEANLEN=100 -nl 16
./test-bin/StringGatherSpeedTest --N=100_000_000 --MEANLEN=100 -nl 16
```

| --N         | before    | now         |
| ----------: | --------: | ----------: |
|   1_000_000 | 25.5 MB/s |  827.5 MB/s |
| 100_000_000 | N/A       | 1855.5 MB/s |

Part of https://github.com/mhmerrill/arkouda/issues/527

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -218,9 +218,9 @@ module SegmentedArray {
         t1 = getCurrentTime();
       }
       var gatheredVals = makeDistArray(retBytes, uint(8));
-      // For comm layer with poor small-message performance, use aggregation
-      // at the expense of memory. Otherwise, unorderedCopy is faster and smaller.
-      if !(CHPL_COMM == 'none' || CHPL_COMM == 'ugni') {
+      // Multi-locale requires some extra localization work that is not needed
+      // in CHPL_COMM=none
+      if CHPL_COMM != 'none' {
         // Compute the src index for each byte in gatheredVals
         /* For performance, we will do this with a scan, so first we need an array
            with the difference in index between the current and previous byte. For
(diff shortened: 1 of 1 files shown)


src/SegmentedArray.chpl
```diff
@@ -218,17 +218,16 @@ module SegmentedArray {
         t1 = getCurrentTime();
       }
       var gatheredVals = makeDistArray(retBytes, uint(8));
-      // For comm layer with poor small-message performance, use aggregation
-      // at the expense of memory. Otherwise, unorderedCopy is faster and smaller.
-      if !(CHPL_COMM == 'none' || CHPL_COMM == 'ugni') {
+      // Multi-locale requires some extra localization work that is not needed
+      // in CHPL_COMM=none
+      if CHPL_COMM != 'none' {
```
ronawho, line 223: If you want to reduce the number of implementations and test this code for `comm=none` as well, we could just get rid of this branch. 

This will seriously hurt `comm=none` string gather performance, but I'm not sure if that specialization is worthwhile to you or not. On the machine that runs the nightly `chapcs` testing I see performance go from ~2150 MB/s to ~350 MB/s -- https://github.com/mhmerrill/arkouda/issues/527#issuecomment-712400645

mhmerrill: I still think we need the `comm=none` case.

mhmerrill (approved):
Looks good

Issue #446, opened 2020-07-23
Enable rethrowing of errors and passing in strings to errors


hokiegeek2:
There are sections of code within Arkouda that throw new Errors in error catch blocks such as this:

![image](https://user-images.githubusercontent.com/10785153/88279434-0cd8c800-ccb2-11ea-8006-2de68d437831.png)

It could help in debugging if we enable rethrowing errors by passing the root error in to the init method of the Error being rethrown. It should be pretty straightforward, basically need to add an init method that accepts an Error parameter:

![image](https://user-images.githubusercontent.com/10785153/88279322-dbf89300-ccb1-11ea-908d-78532576f7ff.png)

Also, would recommend adding init methods that accept strings which should also help in debugging.

hokiegeek2:
Leveraged the getLineNumber(), getRoutineName(), and getModuleName() functions @mhmerrill mentioned, which works well:  

![image](https://user-images.githubusercontent.com/10785153/95231392-182b9200-07d1-11eb-95bc-51c92793bff1.png)

![image](https://user-images.githubusercontent.com/10785153/95235707-eb2cae80-07d3-11eb-8ce5-606294670b1b.png)

Probably a good idea to implement a factory to generate the derived classes of ErrorWithContext.

hokiegeek2:
@mhmerrill @reuster986 I updated the error message generation logic to deliver the following:

1. Two types of messages: one that includes the entire context, which is accessible via ErrorWithContext.message(), which is inherited from the core Chapel Error class, and a second that only contains the string that should be sent back to the calling object and Python client, which is accessible via the ErrorWithContext.publish() method. An example is shown below:

![image](https://user-images.githubusercontent.com/10785153/96107239-3288fe00-0eaa-11eb-8017-b002058b33a2.png)

This enables easier integration into try...catch blocks by replacing custom error message generation with invocations of the ErrorWithContext.message() and ErrorWithContext.publish() functions:

![image](https://user-images.githubusercontent.com/10785153/96108638-d2935700-0eab-11eb-8e75-7d7da6ff1ae1.png)


2. Capture of Error class name and inclusion within ErrorWithContext.message() output

3. Extension of the Errors.getErrorWithContext factory function to enable integration of standard Chapel Errors into the ErrorsWithContext message generation paradigm, which ensures Arkouda can generate consistent error messages with the same contextual information. The message is generated in the following helper function:

[...]

Issue #329, opened 2020-03-28
write error condition reporting/handling python unit tests


hokiegeek2:
Write Python unit tests that test error condition reporting and handling in the Arkouda Python API

hokiegeek2:
completed development and testing, merged PR into master, issue resolved.

PR #535, opened 2020-10-20
446 enable rethrowing of errors and passing in strings to errors

hokiegeek2:
Includes addition of unit tests for error handling to ensure added error handling logic is working correctly, covers completion of #446 and start of #329

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -26,6 +26,8 @@ module ArgSortMsg
 
     use RadixSortLSD;
     use SegmentedArray;
+    use Reflection;
+    use Errors;
     
     // thresholds for different sized sorts
     var lgSmall = 10;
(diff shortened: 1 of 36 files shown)


src/ArgSortMsg.chpl
```diff
@@ -598,17 +653,26 @@ module ArgSortMsg
...
+                    st.addEntry(ivname, new shared SymEntry(iv));
+                }
+                when (DType.Float64) {
+                    var e = toSymEntry(gEnt, real);
+                    var iv = argsortDefault(e.a);
+                    st.addEntry(ivname, new shared SymEntry(iv));
+                }
+                otherwise {
+                    var errorMsg = notImplementedError(pn,gEnt.dtype);
+                    try! writeln(generateErrorContext(
```
reuster986, line 668: Can we remove the `try!` from this and similar lines? The parent function `throws`, so it should be fine without it. I would eventually like to remove all `try!` from arkouda to prevent server halts, but in the meantime we should avoid introducing new ones.

reuster986 (changes requested):
I like the new behavior. My only request is that we remove all `try!` and avoid that pattern for the future (see inline comment).

src/ArgSortMsg.chpl
hokiegeek2: @reuster986 Excellent point! I removed all try! clauses I added and remove try! clauses for a few of the other try! writeln clauses.

reuster986 (approved):
great, thanks!

src/ArgSortMsg.chpl
reuster986: excellent, thank you!

PR #538, opened 2020-10-21
Registration feature

mhmerrill:
Here is an initial stab at registration for `pdarray` and `Strings`. I wanted to get comments on this code from @hokiegeek2 and @reuster986


arkouda/pdarrayclass.py
```diff
@@ -908,8 +909,132 @@ def save(self, prefix_path, dataset='array', mode='truncate') -> str:
...
+        >>> # potentially disconnect from server and reconnect to server
+        >>> b = ak.pdarray.attach("my_zeros")
+        >>> # ...other work...
+        >>> b.unregister()
+        """
+        unregister_pdarray(self)
+        
+    # class method self is not passed in
+    # invoke with ak.pdarray.attach('user_defined_name')
+    def attach(user_defined_name : str) -> 'pdarray':
```
mhmerrill, line 995: is this a class method or static method? should I use a decorator here?

arkouda/strings.py
```diff
@@ -754,3 +754,16 @@ def save(self, prefix_path : str, dataset : str='strings_array',
         """       
         self.bytes.save(prefix_path=prefix_path, 
                                     dataset='{}/values'.format(dataset), mode=mode)
+
+    def register(self, user_defined_name : str) -> 'Strings':
+        return Strings(self.offsets.register(user_defined_name+'_offsets'),
```
mhmerrill, line 759: this needs `self` and the class construction... `@classmethod`  doesn't seem to fit the bill, suggestions?

arkouda/strings.py
```diff
@@ -754,3 +754,16 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+                       self.bytes.register(user_defined_name+'_bytes'))
+
+    def unregister(self) -> None:
+        self.offsets.unregister()
+        self.bytes.unregister()
+
+    def attach(user_defined_name : str) -> 'Strings':
```
mhmerrill, line 766: this is probably a `@classmethod`

src/MultiTypeSymbolTable.chpl
```diff
@@ -66,7 +66,7 @@ module MultiTypeSymbolTable
                                    moduleName=getModuleName(),
                                    errorClass="ErrorWithContext");
             }
-            tab.remove(name); // clear out entry for name
+            //tab.remove(name); // clear out entry for name
```
mhmerrill, line 69: should unregister remove the `name` from the symbol table or just the `registry`

arkouda/strings.py
hokiegeek2: @mhmerrill this interesting. I always thought a class method had to be decorated with @classmethod. But, this does indeed work. However, you cannot call self from inside this method (analogous method below):

image


image

arkouda/strings.py
hokiegeek2: @mhmerrill sure, you could do that.

arkouda/pdarrayclass.py
mhmerrill: I think classmethod

arkouda/strings.py
```diff
@@ -760,3 +760,21 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+
+    def register(self, user_defined_name : str) -> 'Strings':
+        return self.register_helper(self.offsets.register(user_defined_name+'_offsets'),
+                               self.bytes.register(user_defined_name+'_bytes'))
+
+    def unregister(self) -> None:
+        self.offsets.unregister()
+        self.bytes.unregister()
+
+    @staticmethod
```
hokiegeek2, line 776: recommend adding typechecked annotation here

arkouda/strings.py
```diff
@@ -760,3 +760,21 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+    def attach(user_defined_name : str) -> 'Strings':
```
hokiegeek2, line 777: Recommend replacing 'Strings' with Strings

arkouda/strings.py
```diff
@@ -760,3 +760,21 @@ def save(self, prefix_path : str, dataset : str='strings_array',
         """       
         self.bytes.save(prefix_path=prefix_path, 
                                     dataset='{}/values'.format(dataset), mode=mode)
+
+    @classmethod
+    def register_helper(cls, offsets, bytes):
+        return cls(offsets, bytes)
+
+    def register(self, user_defined_name : str) -> 'Strings':
```
hokiegeek2, line 768: Recommend adding typechecked annotation and replace 'Strings' with Strings

arkouda/pdarrayclass.py
```diff
@@ -1560,3 +1690,156 @@ def argmaxk(pda : pdarray, k : int) -> pdarray:
...
+    >>> # potentially disconnect from server and reconnect to server
+    >>> b = ak.attach_pdarray("my_zeros")
+    >>> # ...other work...
+    >>> ak.unregister_pdarray(b)
+    """
+    if isinstance(pda, pdarray):
+        repMsg = generic_msg("unregister {}".format(pda.name))
+    elif isinstance(pda, str):
+        repMsg = generic_msg("unregister {}".format(pda))
+    else:
```
hokiegeek2, line 1844: the raise TypeError block is no longer needed now that you have the typechecked annotation

arkouda/pdarrayclass.py
```diff
@@ -1560,3 +1690,156 @@ def argmaxk(pda : pdarray, k : int) -> pdarray:
...
+    Examples
+    --------
+    >>> a = zeros(100)
+    >>> r_pda = ak.register_pdarray(a, "my_zeros")
+    >>> # potentially disconnect from server and reconnect to server
+    >>> b = ak.attach_pdarray("my_zeros")
+    >>> # ...other work...
+    >>> ak.unregister_pdarray(b)
+    """
+    if not isinstance(user_defined_name, str):
```
hokiegeek2, line 1795: The raise TypeError clause no longer needed with typechecked annotation

arkouda/pdarrayclass.py
```diff
@@ -1560,3 +1690,156 @@ def argmaxk(pda : pdarray, k : int) -> pdarray:
...
+    ret = None
+    if isinstance(pda, pdarray):
+        repMsg = generic_msg("register {} {}".\
+                             format(pda.name, user_defined_name))
+        ret = create_pdarray(repMsg)
+    elif isinstance(pda, str):
+        repMsg = generic_msg("register {} {}".\
+                             format(pda, user_defined_name))        
+        ret = create_pdarray(repMsg)
+    else:
```
hokiegeek2, line 1750: This raise TypeError clause no longer needed with typechecked annotation.

arkouda/pdarrayclass.py
```diff
@@ -920,10 +921,139 @@ def save(self, prefix_path, dataset='array', mode='truncate') -> str:
...
+        >>> # potentially disconnect from server and reconnect to server
+        >>> b = ak.pdarray.attach("my_zeros")
+        >>> # ...other work...
+        >>> b.unregister()
+        """
+        unregister_pdarray(self)
+        
+    # class method self is not passed in
+    # invoke with ak.pdarray.attach('user_defined_name')
+    @staticmethod
```
hokiegeek2, line 1008: recommend adding typechecked annotation

hokiegeek2 (approved):

arkouda/pdarrayclass.py
mhmerrill: this brings up a question i had since the method being called is `@typechecked`

do you need it because there are many places where other functions which are `@typechecked` are immediately called

arkouda/pdarrayclass.py
mhmerrill: i was hoping you could help me with the type annotations in the function arguments. I think it should be something like
```python
@typechecked
 def register_pdarray(pda : Union[pdarray,str], user_defined_name : str) -> pdarray:
```

mhmerrill:
> i recommend adding an if..else statement and removing the raise RuntimeError clause on 1749

what if I don't trust the type checker ;-)

PR #544, opened 2020-10-26
Re-enable logging of command message from client to server

reuster986:
Some months ago, `arkouda_server.chpl` stopped logging command strings with args (the lines starting with `>>> `). However, the command args are very useful for analysis of the logs, so this PR adds them back in.

### DIFF
--- a/src/arkouda_server.chpl
+++ b/src/arkouda_server.chpl
@@ -173,6 +173,19 @@ proc main() {
                 authenticateUser(token);
             }
 
+            if (logging) {
+              try {
+                if (cmd != "array") {
+                  writeln(">>> %s %s".format(cmd, payload.decode(decodePolicy.replace)));
+                } else {
+                  writeln(">>> %s [binary data]".format(cmd));
+                }
+                stdout.flush();
+              } catch {
+                // No action on error
+              }
+            }
+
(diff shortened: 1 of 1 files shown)


src/arkouda_server.chpl
```diff
@@ -173,6 +173,19 @@ proc main() {
...
+            if (logging) {
+              try {
+                if (cmd != "array") {
+                  writeln(">>> %s %s".format(cmd, payload.decode(decodePolicy.replace)));
+                } else {
+                  writeln(">>> %s [binary data]".format(cmd));
+                }
+                stdout.flush();
+              } catch {
+                // No action on error
```
hokiegeek2, line 185: Is there any action that should be taken if there is an error in the payload.decode invocation? If so, I recommend two catch blocks, one for the type of Error thrown in the payload.decode invocation

hokiegeek2 (approved):

mhmerrill (approved):
this looks ok

src/arkouda_server.chpl
reuster986: I thought about that, but if there's a payload decode error, it should be caught by the message function specific to the command, with presumably a more specialized error message, so I figured I should let it fall through. Thoughts?

PR #545, opened 2020-10-26
More mem checks

reuster986:
**Note: Merge the command logging PR before this one**

Another user and I have been experiencing server crashes lately due to running out of memory. This PR adds some `overMemLimit` checks to functions that I suspect might be the culprit. Many of these checks might be redundant with the check in `addEntry`, but this is sort of a belt-and-suspenders approach. This PR includes the code to re-enable command logging, so it should be viewed after the other PR is merged.


mhmerrill (approved):
looks good

PR #546, opened 2020-10-26
Remove per locale operations

reuster986:
Per-locale sort, groupby, and reductions have been unsupported and unused for a long time, so this PR removes them from the client and server. Original source code remains in `src/deprecated/`, in case we ever want to resurrect this functionality.

This change shrinks the effective codebase by about 1,000 lines and reduces single-node build time by about 15 seconds (hopefully more for multi-locale).


mhmerrill (changes requested):
The server code and python tests look good but the Chapel unit tests fail the CI. maybe we should move them to a deprecated directory there as well.

mhmerrill:
CI failure on string test. Was this test supposed to be skipped? I don't understand the error message.

reuster986:
Are you seeing a CI failure? I'm showing all checks passed...

mhmerrill:
> Are you seeing a CI failure? I'm showing all checks passed...

I re-ran the CI to see if it would succeed and it did.


> Are you seeing a CI failure? I'm showing all checks passed...

@reuster986 I re-ran the CI to see if it would succeed and it did. but it did fail once before this. I don't think it was because of code you changed so I thought I would accept this PR and put in an issue about the string ends-with test which seemed to be failing.

mhmerrill (approved):

ronawho:
Just FYI this shaved 15s off our nightly [single-locale build](https://chapel-lang.org/perf/arkouda/chapcs/?startdate=2020/10/22&enddate=2020/10/28&graphs=buildtime) like you expected, and 50s off the [multi-locale XC build](https://chapel-lang.org/perf/arkouda/16-node-xc/?startdate=2020/10/21&enddate=2020/10/28&graphs=buildtime).

Issue #562, opened 2020-11-04
setting ak.pdarrayIterThresh has no effect on limiting the values printed


mhmerrill:
This is probably a Python scoping/global variable thing. when you set `ak.pdarrayIterThresh` it does not have an effect only the original default values is used for some reason.

hokiegeek2:
Yep, confirmed with a code example:

![image](https://user-images.githubusercontent.com/10785153/98537681-9f987500-2257-11eb-8cb7-c92766f316dd.png)

![image](https://user-images.githubusercontent.com/10785153/98537714-ac1ccd80-2257-11eb-8999-68941cfd426d.png)

hokiegeek2:
The logic in set_defaults() looks correct to me per Python docs, investigating.

hokiegeek2:
Ah, I see what's going on. Each module has it's own namespace (a good explanation [here](https://stackoverflow.com/questions/423379/using-global-variables-in-a-function)). Explicitly accessing pdarrayIterThresh within the client module yields the expected behavior:

![image](https://user-images.githubusercontent.com/10785153/98540491-0586fb80-225c-11eb-8aec-17e8b5f12255.png)

The recommended pattern is to define a config module where all global variables are stored. Looking to see if we can retain the pattern of accessing global variables via ak.[variable name]

hokiegeek2:
For what I described above, the client.set_defaults() will not work as expected in that the ak.pdarrayIterThreshold is scoped to the current module. 

![image](https://user-images.githubusercontent.com/10785153/98547954-ecd01300-2266-11eb-8237-058445e55d31.png)

The bottom line is that global values are scoped to modules, not across modules.

hokiegeek2:
To ensure the client-based variables behave as expected, any updates to them must be done as follows:

1. Setting individual variables: client.[variable name] = value
2. Resetting client variables to defaults: client.set_defaultsI()

mhmerrill:
@hokiegeek2 what if we created a singleton object with these values in it and initialized it in the `__init_.py` ? instead of accessing the values directly we would have accessors... ?

hokiegeek2:
@mhmerrill we could create a config module, and put the variables in there. Unfortunately with Python, there's no concept of privately-scoped variables that are only accessible via getters.

hokiegeek2:
@mhmerrill 

[...]

PR #564, opened 2020-11-09
updated logic so that client config updates are accessible to calling…

hokiegeek2:
… methods #562


arkouda/strings.py
```diff
@@ -8,9 +8,6 @@
 import numpy as np
 import json
 
-global verbose
-global pdarrayIterThresh
```
mhmerrill, line 12: there is a reference to this in the Strings class. lone 105 or so please fix

hokiegeek2: fixed

mhmerrill (approved):
looks good

Issue #571, opened 2020-11-12
fix dtypes-future annotations incompatibility breaking build


hokiegeek2:

hokiegeek2:
Fixed root cause, merged PR into master, issue resolved.

PR #572, opened 2020-11-12
Reversed futures annotation for now to fix build #571

hokiegeek2:


mhmerrill (approved):
looks like the string test is fixed

PR #574, opened 2020-11-16
fixed import that was missed #562

hokiegeek2:
Missed one import update needed to ensure updated pdarrayIterThresh global var updates are pulled


mhmerrill (approved):
Looks good

PR #582, opened 2020-11-18
Add a couple of use statements to maintain access to string secondary methods

lydia-duncan:
With the change on Chapel's master to return to respecting the privacy and limitations
of use and import statements, these modules started encountering errors when
trying to access string.format (defined in IO) and string.splitMsgToTuple
(defined in ServerConfig).  For the former, add a use of IO to the main body
of the module as it is used in multiple functions.  For the latter, add a use
within the body of castMsg, as it is the only function in the module that uses
the method.

This allowed `make` to run to completion locally

----

### DIFF
--- a/src/CastMsg.chpl
+++ b/src/CastMsg.chpl
@@ -9,6 +9,7 @@ module CastMsg {
   use CommAggregation;
 
   proc castMsg(cmd: string, payload: bytes, st: borrowed SymTab): string throws {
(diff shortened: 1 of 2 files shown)


bradcray:
@mhmerrill: For context, this will keep Arkouda working on Chapel's master branch (and arguably make module dependencies that it's had more explicit, where it was arguably getting away with implicit references to them previously).

mhmerrill (approved):
looks ok

PR #584, opened 2020-11-19
580 change tohdf output filenames to preserve array order

hokiegeek2:


src/GenSymIO.chpl
```diff
@@ -1559,15 +1559,33 @@ module GenSymIO {
      * Generates a list of filenames to be written to based upon a file prefix,
      * extension, and number of locales.
      */
-    proc generateFilenames(prefix : string, extension : string, A) : [] string { 
+    proc generateFilenames(prefix : string, extension : string, A) : [] string throws { 
         // Generate the filenames based upon the number of targetLocales.
         var filenames: [0..#A.targetLocales().size] string;
         for i in 0..#A.targetLocales().size {
-            filenames[i] = try! "%s_LOCALE%s%s".format(prefix, i:string, extension);
+            filenames[i] = generateFilename(prefix, extension, i);
```
reuster986, line 1566: Looking at the Chapel docs on [format strings](https://chapel-lang.org/docs/modules/standard/IO/FormattedIO.html?highlight=formattedio#integral-conversions), I think we can simplify this by doing

```chpl
filenames[i] = "%s_LOCALE%04i%s".format(prefix, i, extension);
```

reuster986 (changes requested):
Logic looks good, just one code simplification to consider.

reuster986 (approved):
Oops, my review crossed with your commit. Everything looks good!

src/GenSymIO.chpl
hokiegeek2: @reuster986 excellent point, @mhmerrill told me about the same thing and I just pushed a change with that change. I retained the generateFilename function in case things get more involved filename generation-wise in the future

Issue #575, opened 2020-11-17
bug: strings access with singleton logical index


reuster986:
When using a logical index with only one True value to index a `ak.Strings`, either the wrong value is returned or an error is raised. Reproducer:

```python
>>> s = ak.array(['a', 'b', 'c', 'd'])
>>> idx = ak.array([False, True, False, False])
>>> s[idx]
[bad value or error]
```

With more than one True, it works fine.

hokiegeek2:
@reuster986 For one locale, I cannot duplicate the errors you describe above:

![image](https://user-images.githubusercontent.com/10785153/99701290-418f3d00-2a62-11eb-8d64-8b0be1eee58c.png)

When I retest with 3 locales, I can duplicate the error:

![image](https://user-images.githubusercontent.com/10785153/99702537-e8c0a400-2a63-11eb-8de6-faa0fc4f8c27.png)

hokiegeek2:
For one logical index, the values are as follows:

![image](https://user-images.githubusercontent.com/10785153/99713917-e74aa800-2a72-11eb-878f-aaf0722e4a1a.png)

which results in a null char being selected:

![image](https://user-images.githubusercontent.com/10785153/99714090-24af3580-2a73-11eb-90c5-f7c2c2e7fc89.png)

For logical indices of two:

![image](https://user-images.githubusercontent.com/10785153/99714200-50322000-2a73-11eb-8538-9fc6d3d4139b.png)

which results in two non-null chars being selected:

![image](https://user-images.githubusercontent.com/10785153/99714359-81125500-2a73-11eb-89b1-ff86149de7f9.png)

hokiegeek2:
I can replicate the error by setting the logical index to the first of strings object:

![image](https://user-images.githubusercontent.com/10785153/99717857-f6802480-2a77-11eb-9308-7eb2e2deb17c.png)

![image](https://user-images.githubusercontent.com/10785153/99717914-0bf54e80-2a78-11eb-9f17-a9b015f6e87a.png)

What's odd is that the newSegs and newVals matches the result above where the seg points to the wrong vals char but throws an error instead of returning the null char.

reuster986:
Yes, this is very strange... Additionally, it looks like newValsEntry is [0, 97] (i.e. [NULL, 'a']), regardless of whether the first or third character was indexed. Am I reading that right?

hokiegeek2:
@reuster986 Yup, you are correct.

reuster986:
[...]

PR #586, opened 2020-11-19
String bug fix

reuster986:
Fixes two string bugs:
* Singleton logical indexing #575 
* Indexing a single string with a trailing double-quote (this was the cause of intermittent `string_test.py` failure a few weeks ago)

Added unit tests to cover both cases.


hokiegeek2 (approved):
Looks good!

PR #587, opened 2020-11-23
500 fix ak.array to accept string columns from a pandas dataframe directly

hokiegeek2:


mhmerrill (commented):
this looks ok to me but I am not sure about requiring `pandas`, I would like Bill to weigh in on this before I approve and merge.

reuster986 (changes requested):
This looks good. The only change needed is to require `pandas` as a dependency in `setup.py`. It was already part of the dev dependencies, but this change brings it into the main package.

hokiegeek2:
@reuster986 excellent point, updated accordingly.

reuster986 (approved):

PR #588, opened 2020-11-24
Accept variable precision in HDF5 dtypes for read

reuster986:
Data is always messy, and we have datasets spread across multiple files where, for example, the dtype is `uint32` in half the files and `uint64` in the other half. Currently, attempting to `ak.read_all()` these files raises an error because of inconsistent dtypes. This PR introduces the optional `strictTypes` keyword arg for `ak.read_hdf` and `ak.read_all`. The default (`strictTypes=True`) preserves current behavior, but the user can set `strictTypes=False` and arkouda will read the dataset as long as all the types can be cast losslessly to an arkouda dtype (e.g. `int64`). There will still be an error if the dtypes are of different kinds, e.g. `int` and `float` are still incompatible.

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -27,7 +27,8 @@ def ls_hdf(filename : str) -> str:
     return generic_msg("lshdf {}".format(json.dumps([filename])))
 
 @typechecked
-def read_hdf(dsetName : str, filenames : Union[str,List[str]]) \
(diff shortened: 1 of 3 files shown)


mhmerrill (commented):
@reuster986 this looks like it's just for integer types? does the io_test comprehend this new functionality?

hokiegeek2 (approved):
Looks really good.

reuster986:
@mhmerrill It works for integer and float types. There is a new test in `io_test.py` for both types.

PR #589, opened 2020-11-24
576 enable seeded rng

hokiegeek2:


mhmerrill:
We are going with PR #590 over this one.

PR #595, opened 2020-12-01
Check all supported python versions in CI testing

ronawho:
Historically, we've just run with whatever the CI configuration happens
to use but recently there were some portability issues with 3.9 that
Mac CI happened to catch. To feel more confident in portability across
python versions add explicit testing for all versions that are currently
supported (3.7, 3.8, 3.9)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,6 +42,31 @@ jobs:
       run: |
         make doc
 
+  arkouda_python_portability:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
+    container:
+      image: chapel/chapel:1.23.0
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        apt-get update && apt-get install -y libhdf5-dev hdf5-tools libzmq3-dev python3-pip
+        echo "\$(eval \$(call add-path,/usr/lib/x86_64-linux-gnu/hdf5/serial/))" > Makefile.paths
+    - name: Build/Install Arkouda
+      run: |
+        make
+        python3 -m pip install -e .[dev]
+    - name: Arkouda make check
+      run: |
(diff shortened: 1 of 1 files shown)


ronawho:
@hokiegeek2 do I have the supported python versions right? I'm pretty sure 3.7 is the current oldest because of typeguard or some of the type-checking changes, and I think 3.9 is the newest release.

hokiegeek2:
> @hokiegeek2 do I have the supported python versions right? I'm pretty sure 3.7 is the current oldest because of typeguard or some of the type-checking changes, and I think 3.9 is the newest release.

Yes and yes. 👍

.github/workflows/CI.yml
```diff
@@ -42,6 +42,31 @@ jobs:
...
+      run: |
+        apt-get update && apt-get install -y libhdf5-dev hdf5-tools libzmq3-dev python3-pip
+        echo "\$(eval \$(call add-path,/usr/lib/x86_64-linux-gnu/hdf5/serial/))" > Makefile.paths
+    - name: Build/Install Arkouda
+      run: |
+        make
+        python3 -m pip install -e .[dev]
+    - name: Arkouda make check
+      run: |
+        make check
```
mhmerrill, line 68: @ronawho I'd use `make test` here also, it is much more complete and I think it's where we caught the 3.9 discrepancy.

.github/workflows/CI.yml
```diff
@@ -42,6 +42,31 @@ jobs:
       run: |
         make doc
 
+  arkouda_python_portability:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8, 3.9]
```
ronawho, line 49: Discussed on chat, but I'll see if there's an alias for the latest python and maybe an easier way to maintain this list going forward instead of having to manually add new versions.

.github/workflows/CI.yml
ronawho: Done

.github/workflows/CI.yml
ronawho: Added `'3.x'` as an alias for latest. 

Looks like there no way to say "test all minor versions >= 3.7", so I think this list will have to be manually maintained.

mhmerrill (approved):
looks ok

PR #605, opened 2020-12-14
599 develop detailed logging framework

hokiegeek2:
The goal of this issue is to deliver detailed logging messages that provide a detailed context for each log message in the form of module name, routine name, and line number. An example is shown below:

image

The current implementation sets the log level to DEBUG if the --v flag is passed in at server startup.

the logger.error function is designed to be used to log any messages associated with any thrown errors.

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -29,10 +29,12 @@ module ArgSortMsg
     use Errors;
     use Logging;
 
+    const asLogger = new Logger();
+
     if v {
-        logger.level = LogLevel.DEBUG;
+        asLogger.level = LogLevel.DEBUG;
     } else {
-        logger.level = LogLevel.INFO;
+        asLogger.level = LogLevel.INFO;
     }
     
     // thresholds for different sized sorts
@@ -454,7 +456,7 @@ module ArgSortMsg
       var (nstr, rest) = payload.decode().splitMsgToTuple(2);
       var n = nstr:int; // number of arrays to sort
       var fields = rest.split();
-      logger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+      asLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
                                   "number of arrays: %i fields: %t".format(n,fields));
       // Check that fields contains the stated number of arrays
       if (fields.size != 2*n) { 
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -68,7 +68,8 @@ module ArgSortMsg
 
     /*     // how many bins in histogram */
     /*     var bins = aMax-aMin+1; */
-    /*     if v {try! writeln("bins = %t".format(bins));} */
+    /*     asLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                                    "bins = %t".format(bins)); */
 
     /*     // histogram domain size should be equal to a_nvals */
     /*     var hD = makeDistDom(bins); */
@@ -119,7 +120,8 @@ module ArgSortMsg
 
     /*     // how many bins in histogram */
     /*     var bins = aMax-aMin+1; */
-    /*     if v {try! writeln("bins = %t".format(bins));} */
+    /*     asLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                                   "bins = %t".format(bins)); */
 
     /*     // create a global count array to scan */
     /*     var globalCounts = makeDistArray(bins * numLocales, int); */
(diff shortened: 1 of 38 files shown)


mhmerrill (approved):
looks ok to me.

src/ArgSortMsg.chpl
```diff
@@ -27,6 +27,15 @@ module ArgSortMsg
...
     use Errors;
+    use Logging;
+
+    const asLogger = new Logger();
+
+    if v {
+        asLogger.level = LogLevel.DEBUG;
+    } else {
+        asLogger.level = LogLevel.INFO;
+    }
```
ronawho, line 38: I wonder if instead of setting the level after creation, if this should just pass the level to the Logger. Something like:

```chpl
const asLogger = new Logger(level=if v then LogLevel.DEBUG else LogLevel.INFO);
```

Or if this logic is the same for all loggers, it could be moved into the Logger type instead of being at all the call sites.

I was also wondering if there needs to be a logger per module, or if you could just create a single global logger in Logging.chpl and have these modules use it. e.g. instead of creating `asLogger` here, just use a global `logger`

src/MsgProcessing.chpl
```diff
@@ -60,10 +69,16 @@ module MsgProcessing
...
-        if v {try! writeln("%s %s %i : %s".format(cmd,dtype2str(dtype),size,rname)); try! stdout.flush();}
+        mpLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+            "cmd: %s dtype: %s size: %i new pdarray name: %s".format(
+                                                     cmd,dtype2str(dtype),size,rname));
         // create and add entry to symbol table
         st.addEntry(rname, size, dtype);
-        // response message
+        // if verbose print result
+        mpLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                                    "created the pdarray %t".format(st.lookup(rname)));
```
ronawho, line 79: Noted in https://github.com/mhmerrill/arkouda/issues/614, but I think this is resulting in trying to convert the entire symbol entry and the array it contains into a string.

src/Logging.chpl
```diff
@@ -0,0 +1,88 @@
...
+     * The LogLevel enum is used to provide a strongly-typed means of
+     * configuring the logging level of a Logger object
+     */
+    enum LogLevel {DEBUG,INFO,WARN,ERROR,CRITICAL};
+    
+    /*
+     * The Logger class provides structured log messages at various levels
+     * of logging sensitivity analogous to other languages such as Python.
+     */
+    class Logger {
```
ronawho, line 16: I think this could be a record instead of a class. That avoids a heap allocation, but shouldn't have a major perf impact.

src/Logging.chpl
```diff
@@ -0,0 +1,88 @@
...
+        var warnLevels = new set(LogLevel,[LogLevel.WARN, LogLevel.INFO,
+                           LogLevel.DEBUG]);
+        
+        var criticalLevels =  new set(LogLevel, [LogLevel.CRITICAL, LogLevel.WARN,
+                                         LogLevel.INFO,LogLevel.DEBUG]);
+                                         
+        var errorLevels =  new set(LogLevel,[LogLevel.ERROR, LogLevel.CRITICAL,
+                           LogLevel.WARN, LogLevel.INFO, LogLevel.DEBUG]);
+                             
+        var infoLevels = new set(LogLevel,[LogLevel.INFO,LogLevel.DEBUG]);
```
ronawho, line 28: Micro-optimization, but it would be better to declare these sets with `parSafe=false` to avoid locking when accessing them. I also think they could just be created once at the module level instead of once per instance. I think they could also be `const`

src/Logging.chpl
```diff
@@ -0,0 +1,88 @@
...
+        proc debug(moduleName, routineName, lineNumber, msg: string) throws {
+            if level == LogLevel.DEBUG  {
+                writeln(generateLogMessage(moduleName, routineName, lineNumber, 
+                                            msg, "DEBUG"));
+                stdout.flush();
+            }
+        }
+        
+        proc info(moduleName, routineName, lineNumber, msg: string) throws {
+            if infoLevels.contains(level) {
```
ronawho, line 41: Another micro-optimization, but these `contains` calls aren't free, and it should be faster to just check against all the enums instead of putting them in a set. Or if the enums are ordered I think you can just compare them. Something like:

```chpl
if level >= LogLevel.info {
  ...
}
```

src/Logging.chpl
```diff
@@ -0,0 +1,88 @@
...
+                 generateDateTimeString(), moduleName,routineName,lineNumber, 
+                                     level,msg);
+             } else {
+                 return "[%s] %s Line %i %s [Chapel] %s".format(moduleName, 
+                 routineName,lineNumber,level,msg);            
+             }
+        }
+         
+        proc generateDateTimeString() throws {
+            var dts = datetime.today():string;
```
ronawho, line 81: Another minor thing, but it's not free to get the current time. This was historically a system call, though with modern `vDSO` (virtual dynamic shared object) it's all in user-space and only takes ~15 ns (at least based on my last timings for this in https://github.com/chapel-lang/chapel/pull/12244)

I don't think it's worth changing anything here, and the overhead of the getting the time compared to the time to do large array operations should be negligible. I was just looking for sources of overhead while looking at https://github.com/mhmerrill/arkouda/issues/614 so wanted to note it before I forgot.

Issue #602, opened 2020-12-08
connect message is no longer printed from ak.connect, should this be the default?


mhmerrill:
connect message is no longer printed from `ak.connect`, should this be the default?

mhmerrill:
resolved in #608

PR #608, opened 2020-12-18
Re-enabled connect message, fixed logic error in disconnect that caus…

hokiegeek2:
…es method to hang, cleaned up logging config logic, added unit test for disconnect -> disconnect workflow #602


mhmerrill (approved):
looks good

PR #615, opened 2021-01-06
Avoid converting GenSymEntry's to strings

### LINKED ISSUE #614: Execution slowdown from logging framework in 605
Something about the logging framework added in https://github.com/mhmerrill/arkouda/pull/605 is slowing down Arkouda. The actual benchmark timings haven't changed, but our nightly configurations are timing out from slow non-kernel operations.

With https://github.com/mhmerrill/arkouda/commit/27c293b (before #605):

```
time -p ./benchmarks/stream.py localhost 5555 --trials=1

> ...
> Average time = 0.2336 sec
> real 1.65
> ...
```

With https://github.com/mhmerrill/arkouda/commit/c28e953 (after #605):

```
time -p ./benchmarks/stream.py localhost 5555 --trials=1

> ...
> Average time = 0.2202 sec
> real 169.61
> ...
```

i.e. the timed kernel of stream is 0.2 seconds, but total time jumps from ~1.7s -> ~170s

ronawho:
There were some places that were converting a `GenSymEntry` into a
string. This resulted in converting the array contained in the
symbol entry into a string, which will be be very slow.

Change to using the symbol entry's attributes, which just includes
some useful metadata, but not the array itself.

This also adjusts `GenSymEntry.writeThis` to only write a portion of
large arrays instead of the whole thing. That way if they are
accidentally printed in the future there won't be such a large
performance hit.

Resolves #614

### DIFF
--- a/src/ConcatenateMsg.chpl
+++ b/src/ConcatenateMsg.chpl
@@ -115,7 +115,7 @@ module ConcatenateMsg
                 }
                 var repMsg = "created " + st.attrib(segName) + "+created " + st.attrib(valName);
                 cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                  "created concatenated pdarray %t".format(st.lookup(valName)));
+                                  "created concatenated pdarray %s".format(valName));
                 return repMsg;
             }
             when "pdarray" {
@@ -140,7 +140,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                         "created concatenated pdarray: %t".format(st.lookup(rname)));
+                                         "created concatenated pdarray: %s".format(rname));
                     }
                     when DType.Float64 {
                         // create array to copy into
@@ -159,7 +159,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                         "created concatenated pdarray: %t".format(st.lookup(rname)));
+                                         "created concatenated pdarray: %s".format(rname));
                     }
                     when DType.Bool {
                         // create array to copy into
@@ -178,7 +178,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                           "created concatenated pdarray: %t".format(st.lookup(rname)));
+                                           "created concatenated pdarray: %s".format(rname));
                     }
                     otherwise {
                         var errorMsg = notImplementedError("concatenate",dtype);
--- a/src/ConcatenateMsg.chpl
+++ b/src/ConcatenateMsg.chpl
@@ -115,7 +115,7 @@ module ConcatenateMsg
                 }
                 var repMsg = "created " + st.attrib(segName) + "+created " + st.attrib(valName);
                 cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                  "created concatenated pdarray %s".format(valName));
+                                  "created concatenated pdarray %s".format(st.attrib(valName)));
                 return repMsg;
             }
             when "pdarray" {
@@ -140,7 +140,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                         "created concatenated pdarray: %s".format(rname));
+                                         "created concatenated pdarray: %s".format(st.attrib(rname)));
                     }
                     when DType.Float64 {
                         // create array to copy into
@@ -159,7 +159,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                         "created concatenated pdarray: %s".format(rname));
+                                         "created concatenated pdarray: %s".format(st.attrib(rname)));
                     }
                     when DType.Bool {
                         // create array to copy into
@@ -178,7 +178,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                           "created concatenated pdarray: %s".format(rname));
+                                           "created concatenated pdarray: %s".format(st.attrib(rname)));
                     }
                     otherwise {
                         var errorMsg = notImplementedError("concatenate",dtype);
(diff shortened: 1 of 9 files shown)


hokiegeek2 (approved):

src/MultiTypeSymEntry.chpl
```diff
@@ -110,6 +114,9 @@ module MultiTypeSymEntry
             if v {writeln("deinit SymEntry");try! stdout.flush();}
         }
         
+        override proc writeThis(f) throws {
+          halt("SymEntry cannot be stringified");
+        }
```
ronawho, line 119: These halts helped me find a few other cases. 

If you want to avoid halts, we could make this a warning (though I worry we won't see the warning), or we could just define how to print symbol entries in a way that doesn't print the whole array.

hokiegeek2: @ronawho First of all, excellent work on figuring this out. I see that I changed a few st.lookup to st.attrib, just didn't catch 'em all. Since this one returns a string, it should be good to replace the name with st.attrib. 

From my grepping about the only one left is InqdMsg.chpl

src/ConcatenateMsg.chpl
```diff
@@ -115,7 +115,7 @@ module ConcatenateMsg
                 }
                 var repMsg = "created " + st.attrib(segName) + "+created " + st.attrib(valName);
                 cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                  "created concatenated pdarray %t".format(st.lookup(valName)));
+                                  "created concatenated pdarray %s".format(valName));
```
hokiegeek2, line 118: could we change this to st.attrib(valName)? That will provide the info I think would be good for debugging purposes.

src/ConcatenateMsg.chpl
```diff
@@ -140,7 +140,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                         "created concatenated pdarray: %t".format(st.lookup(rname)));
+                                         "created concatenated pdarray: %s".format(rname));
```
hokiegeek2, line 143: could we change this to st.attrib(valName)? That will provide the info I think would be good for debugging purposes.

src/ConcatenateMsg.chpl
```diff
@@ -159,7 +159,7 @@ module ConcatenateMsg
...
```
hokiegeek2, line 162: same as above

src/IndexingMsg.chpl
```diff
@@ -87,8 +87,8 @@ module IndexingMsg
         var gEnt: borrowed GenSymEntry = st.lookup(name);
         
         imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-            "cmd: %s pdarray to slice: %t start: %i stop: %i stride: %i slice: %t new name: %s".format(
-                       cmd, gEnt, start, stop, stride, slice, rname));
+            "cmd: %s pdarray to slice: %s start: %i stop: %i stride: %i slice: %t new name: %s".format(
+                       cmd, name, start, stop, stride, slice, rname));
```
hokiegeek2, line 91: same as above

src/IndexingMsg.chpl
```diff
@@ -318,7 +318,7 @@ module IndexingMsg
         var gIV: borrowed GenSymEntry = st.lookup(iname);
         
         imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                              "cmd: %s gX: %t gIV: %t value: %s".format(cmd,gX,gIV,value));
+                              "cmd: %s gX: %s gIV: %s value: %s".format(cmd,name,iname,value));
```
hokiegeek2, line 321: same as above

src/MsgProcessing.chpl
```diff
@@ -76,7 +76,7 @@ module MsgProcessing
         st.addEntry(rname, size, dtype);
         // if verbose print result
         mpLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
-                                    "created the pdarray %t".format(st.lookup(rname)));
+                                    "created the pdarray %s".format(rname));
```
hokiegeek2, line 79: ditto

src/MsgProcessing.chpl
```diff
@@ -342,7 +342,7 @@ module MsgProcessing
         var gEnt: borrowed GenSymEntry = st.lookup(name);
 
         mpLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                      "cmd: %s value: %s in pdarray %t".format(cmd,name,gEnt));
+                                      "cmd: %s value: %s in pdarray %s".format(cmd,name,name));
```
hokiegeek2, line 345: ditto

src/ConcatenateMsg.chpl
```diff
@@ -178,7 +178,7 @@ module ConcatenateMsg
                             start += o.size;
                         }
                         cmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                           "created concatenated pdarray: %t".format(st.lookup(rname)));
+                                           "created concatenated pdarray: %s".format(rname));
```
hokiegeek2, line 181: same as above

src/EfuncMsg.chpl
```diff
@@ -49,7 +49,7 @@ module EfuncMsg
         var gEnt: borrowed GenSymEntry = st.lookup(name);
         
         eLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                           "cmd: %s efunc: %s pdarray: %t".format(cmd,efunc,gEnt));
+                           "cmd: %s efunc: %s pdarray: %s".format(cmd,efunc,name));
```
hokiegeek2, line 52: same as above

hokiegeek2 (dismissed):

src/ConcatenateMsg.chpl
ronawho: Yup, sounds good. I figured just printing `name` wasn't the right thing to do, but it was a quick fix. Happy to change to `st.attrib()`

hokiegeek2: @ronawho Cool, yeah, like I wrote, I changed some of the st.lookup to st.attrib once I'd worked out some debugging stuff for another issue, but I missed the ones you found. Again, great work figuring this out!

src/MultiTypeSymEntry.chpl
ronawho: Swapped out the halts for printing things, but limiting how much of the array is printed (this was used as part of `SymTab.pretty()`, which is used in a few tests. 

I also updated the instance in In1dMsg.chpl

src/ConcatenateMsg.chpl
ronawho: Changed

mhmerrill (approved):
this looks ok

PR #616, opened 2021-01-07
Benchmark improvements

reuster986:
This PR enables seeded random number generation for all benchmarks and introduces a new benchmark for IO. The IO benchmark requires a path in order to target a particular storage device, so it is only for manual use and does not run in CI. The correctness of IO functionality is already measured by the unit tests.

### DIFF
--- a/benchmarks/IO.py
+++ b/benchmarks/IO.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3                                                         
+
+import time, argparse
+import arkouda as ak
+import os
+from glob import glob
+
+TYPES = ('int64', 'float64')
+
+def time_ak_write_read(N_per_locale, trials, dtype, path):
+    print(">>> arkouda write/read")
+    cfg = ak.get_config()
+    N = N_per_locale * cfg["numLocales"]
+    print("numLocales = {}, N = {:,}".format(cfg["numLocales"], N))
+    if dtype == 'int64':
+        a = ak.randint(0, 2**32, N)
+    elif dtype == 'float64':
+        a = ak.randint(0, 1, N, dtype=ak.float64)
+     
+    writetimes = []
+    readtimes = []
+    for i in range(trials):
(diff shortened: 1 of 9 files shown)


reuster986:
@ronawho once the PR is merged, would it make sense to add the IO benchmark to the nightly performance runs? Do those platforms have a parallel storage appliance that could be targeted?

ronawho:
We run on lustre, so we should be able to run it nightly. The machines we run on don't have very provisioned lustre systems so performance would be lower than on a real machine, but it should still let us establish a baseline and notice if there's any major regressions or improvements.

hokiegeek2 (approved):

ronawho:
FYI here's some preliminary data from the 16-node-cs-hdr machine we run some nightly testing on (rough specs listed in https://chapel-lang.org/perf/arkouda/). This is for the default problem size running on lustre backed by hard drives (not flash/nvme):

| locales | read       | write       |
| ------- | ---------: | ----------: |
|  1      | 3.82 GiB/s |  0.74 GiB/s |
|  2      | 1.92 GiB/s |  1.56 GiB/s |
|  4      | 2.52 GiB/s |  3.57 GiB/s |
|  8      | 1.85 GiB/s |  5.10 GiB/s |
| 16      | 1.96 GiB/s | 11.10 GiB/s |
| 32      | 1.85 GiB/s | 23.07 GiB/s |

I have no idea what this system is supposed to be capable of in terms of read/write speeds. Write scaling looks good (performance roughly doubles when we double the number of locales) but something about reads seems to be serialized.

I'll start looking at this in the background. I also plan on adding this as a nightly performance test and gathering scalability numbers on some other platforms.

ronawho:
>  once the PR is merged, would it make sense to add the IO benchmark to the nightly performance runs?

Adding nightly testing in https://github.com/mhmerrill/arkouda/pull/647

PR #618, opened 2021-01-08
Add interleave mode for concatenation

reuster986:
Concatenation is a very communication-intensive operation. In many cases, the user does not care about the order in which elements are concatenated (e.g. if the concatenated array is just going to be sorted immediately). In these cases, performing the concatenation in a block-interleaved fashion can minimize communication and improve scaling. This PR enables interleaved concatenation and uses it in several of the pdarray setops. Testing is accomplished by `test/UnitTestConcatenate.chpl` and by the existing `tests/setops_test.py`.

The user entry point for this functionality is `ak.concatenate(arrays, ordered=False)`


src/ConcatenateMsg.chpl
```diff
@@ -130,14 +184,24 @@ module ConcatenateMsg
...
-                            ref ea = e.a;
-                            // copy array into concatenation array
-                            forall (i, v) in zip(o.aD, o.a) with (var agg = newDstAggregator(int)) {
-                              agg.copy(ea[start+i], v);
+                            const o = toSymEntry(st.lookup(name), int);
+                            if mode == "interleave" {
+                              coforall loc in Locales {
+                                on loc {
+                                  const size = o.aD.localSubdomain().size;
+                                  e.a[blockstarts[here.id]..#size] = o.a.localSlice[o.aD.localSubdomain()];
```
reuster986, line 192: @ronawho @bradcray I'd love your opinion on this line, from a comms perspective. The source is local, and the destination is either entirely local or it might have one non-local element on the end(s), depending on how the BlockDist boundaries line up. Is a range slice on the destination the best pattern to use, or should I be using a domain slice? Or something else?

bradcray: Tagging @e-kayrakli as well, who's been looking at slicing recently and may have better data than I do.

While I'm thinking about it, just to check on terminology, when you say the destination is local, do you mean "non-distributed" or "located on the same locale as the source" (modulo those end conditions you mention?).

reuster986: Good question -- I mean located on the same locale as the source (i.e. `loc` from the `on` block), modulo end conditions.

e-kayrakli: If there's a chance that the slice can expand more than one locale, then keep in mind that you cannot just slice the array by `{blockstarts[here.id]..#size}` as the slice won't inherit the distribution of the domain. So, you'll first need to slice the domain of that array and then pass that domain slice (e.g. `myArr[myArr.domain[start..end]]` and not `myArr[{start..end}]`)

I can't think of any comm-related gotchas between the two. However, if you do domain-based slicing, I think what will happen is that we'll recognize the newly created domain as constant, but we may not recognize that it is a "single-use" domain. So, its creation will involve local and non-contentious locking (to count its arrays). Whereas, if you do range-based slicing, the array code will create a domain internally, and it'll be aware that it is a single-use one. So it will avoid that lock. Obviously this matters only if your data movement is small.

reuster986: @e-kayrakli Thanks for the info! Just to confirm, the code as-written (using a range-based slice) will create a single-use, internal domain that inherits the distribution of the destination array's domain... do I have that right?

bradcray: Engin got back here before me, but I'm not sure I agree with what he says.  :)  I believe it's legal to slice a distributed array with a non-distributed domain and that the differences relate more to what the domain of the resulting expression will be rather than whether one is legal or not (e.g., if I said `var A = myDistArr[lo..hi];`, A would be a distributed array itself, even if it's only distributed to a single locale.  Whereas if I said `var A = myDistArr[{lo..hi}];`, then A would be a local array, definitely on this locale.  It shouldn't matter whether the indices `lo..hi` are local to the current locale for the original array or not.

Though I felt confident in the previous statement, I wasn't actually certain which would be cheaper for this pattern, so did some quick experiments and was going to advise writing the LHS as `e.a[{blockstarts[here.id]..#size}]` because I think it will be cheaper.  Specifically, slicing a distributed array with a range currently creates a new distributed domain, which is a fairly expensive operation involving all locales; whereas slicing it with a domain does not.

This is regrettable since the slice-by-range syntax is cleaner, and we're looking into ways to reduce those overheads.  Using the `-schpl_serializeSlices=true` flag at compile-time helps (for either idiom), but I recall we've talked about using this before with Arkouda, but aren't in the Makefile today, so maybe there were other downsides that I'm not remembering.

Here's the test I was using in making these characterizations:

```chapel
use BlockDist, CommDiagnostics;

var A = newBlockArr(1..1000, real),
    B = newBlockArr(1..1000, real);

resetCommDiagnostics();
startCommDiagnostics();
A[1..10] = B.localSlice[11..20];
stopCommDiagnostics();
printCommDiagnosticsTable();

resetCommDiagnostics();
startCommDiagnostics();
A[{1..10}] = B.localSlice[11..20];
stopCommDiagnostics();
printCommDiagnosticsTable();


resetCommDiagnostics();
startCommDiagnostics();
A[{492..501}] = B.localSlice[11..20];
stopCommDiagnostics();
printCommDiagnosticsTable();


resetCommDiagnostics();
startCommDiagnostics();
A[492..501] = B.localSlice[11..20];
stopCommDiagnostics();
printCommDiagnosticsTable();
```

e-kayrakli: I stand corrected. I wasn't implying that it would be illegal to slice it with a local domain, but I was expecting that it would incur some fine grained communication (though reading what I wrote, I haven't made it clear). But looking a bit more, it looks like it is not the case and I agree with Brad that you should slice with a domain.

@reuster986 said

> Just to confirm, the code as-written (using a range-based slice) will create a single-use, internal domain that inherits the distribution of the destination array's domain... do I have that right?

That's right, yes. It'll in fact slice the array's domain to create the new one. However, it'll do some more internal adjustments to make that domain a bit lighter-weight compared to a regular user-level domain.

reuster986:
I'm at a loss as to why the CI is failing. The specific test that's failing is arkouda_python_portability (3.x), which tests with python 3.9.1. I have seen the 3.7 and 3.8 tests succeed in runs where they complete before 3.x, so I believe the issue is isolated to python 3.9. However, I cannot reproduce the error on my laptop with a matching configuration (linux, python 3.9.1, pytest 6.2.1, etc.). Can anyone else reproduce this error?

Looking at the CI logs, the relevant lines for this error appear to be:
```
arkouda/groupbyclass.py:796: in broadcast
    diffs = concatenate((array([values[0]]), values[1:] - values[:-1]))
arkouda/pdarrayclass.py:261: in __sub__
    return self._binop(other, "-")
arkouda/pdarrayclass.py:200: in _binop
    repMsg = generic_msg(msg)
arkouda/client.py:504: in generic_msg
    return cast(str, _send_string_message(message=cast(str,message),

message = 'home:None:binopvv - id_23 id_24', recv_bytes = False

...
RuntimeError: Error: ErrorWithContext Line 210 In MultiTypeSymbolTable.lookup: Error: unknown symbol: id_23: unexpected error
```

In other words, `values[1:] - values[:-1]` is the offending statement, with the `-` triggering the error, and the specific error is that `id_23` is not in the symbol table. To me, this suggests that the python garbage collector jumped the gun on the temporary result of `values[1:]`, triggering the `delete` message before the `binopvv`. However, I don't have the trace from the `arkouda_server`, and I can't reproduce it locally, so I cannot verify this hypothesis. Even if I could, I'm not sure how to fix it.

@hokiegeek2 @ronawho any ideas?

src/ConcatenateMsg.chpl
reuster986: Thank you, both! Sorry, I spent all of yesterday dealing with a strange python 3.9 issue and did not get around to incorporating your feedback until today, but I just switched to domain-based slicing. I'll start a separate issue to discuss using  `-schpl_serializeSlices=true`.

reuster986:
Redux: in an attempt to get around what I think is overzealous garbage collection in python 3.9, I broke out the offending statement and assigned temporary results to variables, (see [here](https://github.com/mhmerrill/arkouda/pull/618/files/5964331ffac5ce66768acff15e7d2185492583e7..f421094cf63d0c13214b9fa63505935c008f181a#diff-b425b3200b45149e68202c74d7cbc5e04ae20ce20ec3b12c134d6e1b5ebdf35dR796-R800)), but that merely moved the error to a different line. I can't remember which line, but the error message continued to imply that one of the slices (e.g. `values[1:]`) was getting deleted before being used.

After that, I gave up on tracking down the root problem. Because the error was occurring inside `GroupBy.broadcast()`, which I had been wanting to make a server-side function anyway, I simply rewrote it in Chapel. Taking it out of python seems to have fixed the CI (for now).

Anyway, the PR should be ready to review and merge now.

mhmerrill (approved):
looks good now

Issue #617, opened 2021-01-08
test_random_strings_uniform_with_seed fails at 16 locales


ronawho:
Added in https://github.com/mhmerrill/arkouda/pull/611, `test_random_strings_uniform_with_seed` is failing in our nightly multi-locale testing.

```
ARKOUDA_NUMLOCALES=16 python3 -m pytest tests/pdarray_creation_test.py
```

```
================================================================================== FAILURES ===================================================================================
__________________________________________________________ PdarrayCreationTest.test_random_strings_uniform_with_seed __________________________________________________________

self = <pdarray_creation_test.PdarrayCreationTest testMethod=test_random_strings_uniform_with_seed>

    def test_random_strings_uniform_with_seed(self):
        pda = ak.random_strings_uniform(minlen=1, maxlen=5, seed=1, size=10)

>       self.assertTrue((ak.array(['TVKJ', 'EWAB', 'CO', 'HFMD', 'U', 'MMGT',
                        'N', 'WOQN', 'HZ', 'VSX']) == pda).any())
E       AssertionError: False is not true

tests/pdarray_creation_test.py:329: AssertionError
---------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------
[Arkouda Client] Line 101 DEBUG: ZMQ version: 4.3.3
[Arkouda Client] Line 118 DEBUG: psp = tcp://localhost:5555
[Arkouda Client] Line 141 DEBUG: [Python] Sending request: connect
[Arkouda Client] Line 146 DEBUG: [Python] Received response: connected to arkouda server tcp://*:5555
[Arkouda Client] Line 499 DEBUG: [Python] Sending request: getconfig
connected to arkouda server tcp://*:5555
[Arkouda Client] Line 499 DEBUG: [Python] Sending request: randomStrings 10 uniform uppercase 1 5 1
[pdarray] Line 1111 DEBUG: id_77 int64 10 1 [10] 8
[pdarray] Line 1111 DEBUG: id_78 uint8 34 1 [34] 1
[pdarray] Line 1111 DEBUG: id_79 int64 10 1 [10] 8
[pdarray] Line 1111 DEBUG: id_80 uint8 39 1 [39] 1
[Arkouda Client] Line 499 DEBUG: [Python] Sending request: segmentedBinopvv == str id_79 id_80 str id_77 id_78
[pdarray] Line 1111 DEBUG: id_81 bool 10 1 [10] 1
[...]

PR #619, opened 2021-01-11
Fixed unit test failures that occur at ARKOUDA_NUMLOCALES > 10 #617

hokiegeek2:


tests/pdarray_creation_test.py
```diff
@@ -323,15 +324,20 @@ def test_random_strings_uniform(self):
         self.assertEqual('type of argument "size" must be int; got str instead', 
                          cm.exception.args[0])              
 
-    def test_random_strings_uniform_with_seed(self):
-        pda = ak.random_strings_uniform(minlen=1, maxlen=5, seed=1, size=10)
- 
-        self.assertTrue((ak.array(['TVKJ', 'EWAB', 'CO', 'HFMD', 'U', 'MMGT', 
+    def test_random_strings_uniform_with_seed(self):        
+        try:
+            locales = int(os.environ['ARKOUDA_NUMLOCALES'])
```
ronawho, line 329: You should be able to get this with `locales = ak.get_config()['numLocales']`, which is used in some other tests and would avoid needing the env var and the try/except block here.

ronawho:
Is this a workaround to get clean nightly testing, or the long term solution? Seems slightly unfortunate to just skip a test in nightly testing, but I'm ok with it if Mike is.

hokiegeek2:
> Is this a workaround to get clean nightly testing, or the long term solution? Seems slightly unfortunate to just skip a test in nightly testing, but I'm ok with it if Mike is.

Yes. Running this test where the number of locales > 10 appears to result in non-deterministic behavior. For 10 or less, these tests consistently work. I am definitely open to other options if you have 'em.

reuster986:
Looking at `src/RandArray.chpl`, in multilocale execution, each locale gets its own seed (the user-specified seed plus `here.id`). This is to guard against each locale ending up with an identical sub-array; however, it means that the contents of the global array will depend on `numLocales`. Thus, I am actually quite surprised that this tests works for `1 < numLocales < 10` as @ronawho mentioned earlier.

For the test, I would argue that we should generate test vectors for the number of locales used in CI and nightly testing (is that 1, 2, and 16?), but skip the test for other configurations. That will at least test the functionality in both single- and multi-locale mode but won't trip up users who want to test a different setup.

ronawho:
> Thus, I am actually quite surprised that this tests works for 1 < numLocales < 10 as @ronawho mentioned earlier.

Huh, yeah that is surprising. I just ran again and 1-10 locales passes, but I couldn't explain why without digging some more. 

> For the test, I would argue that we should generate test vectors for the number of locales used in CI and nightly testing (is that 1, 2, and 16?), but skip the test for other configurations. That will at least test the functionality in both single- and multi-locale mode but won't trip up users who want to test a different setup.

This sounds good to me. 1, 2, and 16 locales are the ones we run under CI or nightly testing.

hokiegeek2:
@reuster986 so create the arkouda arrays, concatenate, and then run the .any() test, correct?

reuster986:
Oh, that's why the test is succeeding for `1 < numLocales < 10`: because it's using `any()` instead of `all()`. We should be using `all()`, right?

Here's what I would do:
1. Build arkouda in multi-locale mode
2. Run `arkouda_server -nl 1`
3. Record output of `ak.random_strings_uniform(..., seed=5)` as a test vector
4. Repeat 2-3 with `-nl 2` and `-nl 16`
5. Hard-code those test vectors in the python test
6. At runtime, select a test vector based on `ak.get_config()['numLocales']` and check equality with `.all()`

hokiegeek2:
@reuster986 IIRC using any() is the way I got the test to work cross-platform.

ronawho:
> Oh, that's why the test is succeeding for 1 < numLocales < 10: because it's using any() instead of all(). We should be using all(), right?

Ah, that makes a lot of sense. I was driving myself crazy trying to figure out why the test was passing for 1-10 locales currently. For many of those cases it's just passing because the first element is right.

@reuster986 @hokiegeek2 I'm working on a patch that should make the fillInt/fillUint stable regardless of the number of locales. I'll create a PR for that as a potential alternative approach.

reuster986:
@hokiegeek2 I merged in the fix from @ronawho , so forget about the steps I listed above. I think if you pull master and regenerate the test vector, the result should be locale-invariant. Some of the tests probably still need to be changed from `.any()` to `.all()` though.

reuster986 (approved):
looks good

tests/pdarray_creation_test.py
```diff
@@ -1,3 +1,4 @@
+import os
```
ronawho, line 1: nit, but os import isn't needed anymore

ronawho (approved):

PR #621, opened 2021-01-11
Make fill* random routines stable with varying numLocales

ronawho:
Previously `fillInt`, `fillUint`, `fillReal`, and `fillBool` would do a
coforall across the locales and use `fillRandom` on the  local array
with a seed that was based on `here.id`. Basing the seed on the locale
ID made the results unstable as the number of locales varies.

Instead of doing that, just use `fillRandom` on the distributed array.
This routine does similar operations under the covers on distributed
arrays as the manual coforall+local-fill, but we take care to make sure
the results are stable no matter how many locales are used.

This fixes the instability in `test_random_strings_uniform_with_seed()`
as the number of locales changes. I checked that test for 1-32 locales.

Resolves https://github.com/mhmerrill/arkouda/issues/617

Co-developed with @mppf (our RNG expert)

### DIFF
--- a/src/RandArray.chpl
+++ b/src/RandArray.chpl
@@ -20,71 +20,49 @@ module RandArray {
   } 
 
   proc fillInt(a:[] ?t, const aMin, const aMax, const seedStr:string="None") throws where isIntType(t) {
-    coforall loc in Locales {
-      on loc {
-        ref myA = a.localSlice[a.localSubdomain()];
-        if (seedStr.toLower() == "none") {
-          fillRandom(myA);
-        } else {
-          var seed = (seedStr:int) + here.id;
-          fillRandom(myA, seed);
-        }
-        [ai in myA] if (ai < 0) { ai = -ai; }
-        if (aMax > aMin) {
-          const modulus = aMax - aMin;
-          [x in myA] x = ((x % modulus) + aMin):t;
-          //myA = (myA % modulus) + aMin:t;
-        }
+      if (seedStr.toLower() == "none") {
+        fillRandom(a);
+      } else {
+        var seed = (seedStr:int) + here.id;
+        fillRandom(a, seed);
       }
-    }
+      [ai in a] if (ai < 0) { ai = -ai; }
+      if (aMax > aMin) {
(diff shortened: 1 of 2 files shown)


ronawho:
Probably easiest to review ignoring whitespace -- https://github.com/mhmerrill/arkouda/pull/621/files?w=1. It basically just turns:

```chpl
coforall loc in Locales do on loc {
  ref myA = a.localSlice[a.localSubdomain()];
  fillRandom(myA);
}
```

into:

```chpl
fillRandom(a);

```

reuster986 (approved):
Excellent! Thanks @ronawho and @mppf ! I didn't realize `fillRandom(a, seed)` "just works" multi-locale, which is cool. I'm sure you guys have looked at performance, but is there any additional communication overhead to coordinate seeding across locales?

ronawho:
There are some extra on-stmts, but the cost is negligible. I created a temporary test/rand.chpl to experiment with and was running on 16-node-xc:

```chpl
use TestBase, RandArray;

config const size = 10**9;
var A = makeDistArray(size*numLocales, int);

var d: Diags; d.start();
fillInt(A, 2, 11, "1");
d.stop("fillInt");
```

```
make test-bin/rand
./test-bin/rand -nl 16 --printDiagsSum
> fillInt: 0.71s
> fillInt: (GETS = 0, PUTS = 0, ONS = 15)

wget https://github.com/mhmerrill/arkouda/pull/621.diff
git apply 621.diff
make test-bin/rand
./test-bin/rand -nl 16 --printDiagsSum
> fillInt: 0.71s
> fillInt: (GETS = 0, PUTS = 0, ONS = 45)

```

So there's 3x the on-stmts because the `fillRandom()`, `[ai in a] if (ai < 0) { ai = -ai; }`, and `if (aMax > aMin)...` will each do a `coforall loc in Locales do on loc` under the covers, but the coforall+on is so fast you'll never notice the overhead. 

Note that there's no GETs/PUTs for coordinating seeds or anything because those values get sent as part of the on-stmts used under the covers.

reuster986:
@ronawho that is very cool :)

PR #623, opened 2021-01-12
Add an array creation benchmark

ronawho:
Add a benchmark that tracks performance for `ak.zeros`, `ak.ones`, and
`ak.randint`. This is primarly motivated by wanting to track how long
array creation takes on the backend with `ak.zeros`, but the other modes
are interesting too.


mhmerrill (approved):
looks good

Issue #620, opened 2021-01-11
Multi-locale prod reduce performance regression 


ronawho:
There was a performance regression for prod reduce in nightly multi-locale perf testing:

![image](https://user-images.githubusercontent.com/1588337/104229513-5fea3700-541a-11eb-9b14-d390ee1ae211.png)

Interactive graph [link](https://chapel-lang.org/perf/arkouda/16-node-cs/?startdate=2020/11/14&enddate=2021/01/11&graphs=reduceperformance). The regression started some time after the 18th, but testing was timing out for a while because of https://github.com/mhmerrill/arkouda/issues/614 so it's not immediately obvious what PR introduced the regression.

ronawho:
While the GiB/s regression looks large, the amount of time taken is quite small still. On 16-node-cs:

```
Before regression : 0.0069s
After regression  : 0.0130s
```

ronawho:
This is from https://github.com/mhmerrill/arkouda/pull/616, specifically changing the range of values in the array from `0-N` to `1-N` -- https://github.com/mhmerrill/arkouda/pull/616/files#diff-fbdcd69de0ebdfd79b3745a619024ec30e41f8a265a252cbe32f5e3b7a868bc2R21

The prod reduce code actually has 2 reductions. One that checks for 0's and one that does the prod reduce:

https://github.com/mhmerrill/arkouda/blob/06924fd21e10064f4fcba28c6325a3552ad9968d/src/ReductionMsg.chpl#L65-L72

When the array had values 0-N we'd only do the first && reduce, but now we're doing both, so it makes sense that things are ~2x slower. 

Note that the && reduce was added in f596b7398a3167df79c8383c93c46a1647248ca0

@reuster986 and @mhmerrill it's not obvious to me if that check-for-zeros is required for correctness or was a performance fast-path. If it's for performance, I think we can just remove it. If it's for correctness and we care about getting the performance back I think we can use a manual reduction that does the * reduce and checks for 0's at the same time.

reuster986:
Ah, that makes sense -- thanks for tracking this down. I actually can't remember why we added that check, but if it does affect correctness, the `--correctness-only` test should catch it if we remove it.

ronawho:
Looking a little more, I think a case where it impacts correctness is if you have an array that would otherwise prod-reduce to INF, and you add a 0 to the array. Do you want that to return 0, or NaN, or INF?

e.g. 

```chpl
var A = [INFINITY, 0.0];
[...]

PR #625, opened 2021-01-12
Optimize prod reductions

ronawho:
We want prod-reduce to return 0 if any element of the array is 0. This requires
special casing and checking for 0's because otherwise you can get into a
situation where you get a running `inf` and then `inf*0` will be `nan` instead
of 0.

Previously, for int64 this was achieved by doing an && reduction first to check
for 0's but this adds an extra reduction. Instead, use a forall with reduce
intents to do the prod-reduce and check for 0's with 1 pass over the array.

Use this for the float64 case as well, which was previously not checking for
0's.

Part of https://github.com/mhmerrill/arkouda/issues/620

TODOs: 
- [ ] Add a test
- [ ] Decide on return type for int64 prod-reduce

### DIFF
--- a/src/ReductionMsg.chpl
+++ b/src/ReductionMsg.chpl
@@ -64,14 +64,19 @@ module ReductionMsg
                     }
                     when "prod" {
                       ref ea = e.a;
-                      // If any element is zero, skip the computation and return 0.0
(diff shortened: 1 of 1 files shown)


ronawho:
Closing, see discussion in #620

PR #626, opened 2021-01-13
Fix compilation error with Chapel git

ronawho:
Bring in `ServerConfig` to make `splitMsgToTuple` available in
`BroadcastMsg`. In 1.23 this was brought in through some lax visibility
rules on methods, but those have been tightened up on the dev branch.


reuster986 (approved):
Good catch!

PR #627, opened 2021-01-13
using string-suffix-array-functionality branch instead of master

zhihuidu:
Introduces Suffix Arrays into Arkouda:
Contains native Chapel implementation of the suffix array construction algorithm contained in [Simple Linear Work Suffix Array Construction](https://www.cs.helsinki.fi/u/tpkarkka/publications/jacm05-revised.pdf)

A potential use case of suffix arrays is the [Burrows–Wheeler transform](https://en.wikipedia.org/wiki/Burrows%E2%80%93Wheeler_transform), a reversible preprocessing step for improving lossless data compression


mhmerrill:
@zhihuidu i am puzzled as to why you would close this PR. I am sorry we have not gotten this PR done. We should have a discussion about this work. @reuster986 what do you think about this?

zhihuidu:
@mhmerrill I will be glad to discuss with this PR more! I  did not close the PR and it seems that some member of Arkouda closed it when all errors have been solved.
I found that your arkouda version has updated and I am trying to solve the new conflicts.

mhmerrill:
@zhihuidu oh, I see what happened, I think you removed your branch and it automatically closed the PR. Just reopen the PR when you get things in order.

> @mhmerrill I will be glad to discuss with this PR more! I did not close the PR and it seems that some member of Arkouda closed it when all errors have been solved.
> I found that your arkouda version has updated and I am trying to solve the new conflicts.

zhihuidu:
@mhmerrill I have solved the conflicts and updated the branch.

zhihuidu:
@mhmerrill After merge with your updated version, I do not know why string_test.py can not pass the test.

pierce314159:
I made the minimum changes required to pass the current tests. I believe this should pass the CI once Issue #839 is resolved.  The recent commits:
- Update logging statements to use logLevel instead of v (PR #760)
- Update `check` to `check_table` (PR #792)
- Remove outdated `Strings.attach` function which used `_` for subcomponents (new convention is `.` per PR #774)
- Make `unregister_strings_by_name` a staticmethod
- Manually apply recent changes to `SegmentedArray.chpl` because automatic merge was thrown off by addition of `class SegSArray`, including:
  -PR #666 string test ends_with failure
  -PR #781 off by one in SegmentedArray peel method

zhihuidu:
> I made the minimum changes required to pass the current tests. I believe this should pass the CI once Issue #839 is resolved. The recent commits:
> 
> * Update logging statements to use logLevel instead of v (PR #760)
> * Update `check` to `check_table` (PR #792)
> * Remove outdated `Strings.attach` function which used `_` for subcomponents (new convention is `.` per PR #774)
> * Make `unregister_strings_by_name` a staticmethod
> * Manually apply recent changes to `SegmentedArray.chpl` because automatic merge was thrown off by addition of `class SegSArray`, including:
>   
>   * PR #666 string test ends_with failure
>   * PR #781 off by one in SegmentedArray peel method

@pierce314159 thanks for solving the CI problem.

arkouda/pdarraycreation.py
```diff
@@ -821,3 +820,137 @@ def random_strings_lognormal(logmean : numeric_scalars, logstd : numeric_scalars
...
+        RuntimeError
+            Raised if there is a server-side error in executing group request or
+            creating the pdarray encapsulating the return message
+        """
+        cmd=  "segmentedSuffixAry"
+        args = "{} {} {}".format( strings.objtype,
+                                                        strings.offsets.name,
+                                                        strings.bytes.name) 
+        repMsg = generic_msg(cmd=cmd,args=args)
+        return SArrays(*(cast(str,repMsg).split('+')))
```
pierce314159, line 872: While not the biggest deal, there are a handful of place where spacing/formatting differs from the rest of the arkouda codebase. This is a bit of nitpick from me but I think having consistency is important for readability/maintainability

I would likely update this to look something like this
```python
cmd = "segmentedSuffixAry"
args = "{} {} {}".format(strings.objtype, strings.offsets.name, strings.bytes.name)
repMsg = generic_msg(cmd=cmd, args=args)
return SArrays(*(cast(str, repMsg).split("+")))

```

pierce314159 (changes requested):
I have left comments on a few examples but the gist of my requested changes are:
- Make  `Suffix Array` it's own class separate from `Strings`
- Import the 3rd party C code ([libdivsufsort](https://github.com/y-256/libdivsufsort)) as a dependency to avoid having our own copies to maintain
- Change print statements to logging/errors where appropriate
- Update various formatting/spacing nitpicks for more consistent formatting with the current codebase

Although it's passing the all the tests, I am a little worried that some things from master might've been lost in the merging. My opinion is the safest bet is to apply this functionality onto a fresh copy of master.

I think it would be reasonable to make the suggested changes while applying this functionality onto a fresh copy of master. I would be happy to take point on this if @mhmerrill and @zhihuidu are okay with it

arkouda/strings.py
```diff
@@ -1043,3 +1045,332 @@ def unregister_strings_by_name(user_defined_name : str) -> None:
         """
         unregister_pdarray_by_name(f"{user_defined_name}.bytes")
         unregister_pdarray_by_name(f"{user_defined_name}.offsets")
+
+class SArrays:
+    """
+    Represents an array of (suffix) arrays whose data resides on the arkouda server.
```
pierce314159, line 1051: the `SArrays` class currently resides in the `strings.py` file. In my opinion it should be separated into it's own file

arkouda/pdarraycreation.py
pierce314159, line 872: While not the biggest deal, there are a handful of place where spacing/formatting differs from the rest of the arkouda codebase. This is a bit of nitpick from me but I think having consistency is important for readability/maintainability

I would likely update this to look something like this
```python
cmd = "segmentedSuffixAry"
args = "{} {} {}".format(strings.objtype, strings.offsets.name, strings.bytes.name)
repMsg = generic_msg(cmd=cmd, args=args)
return SArrays(*(cast(str, repMsg).split("+")))

```

benchmarks/sa.py
```diff
@@ -0,0 +1,128 @@
...
+    c=ak.suffix_array(v)
+#    print("size of suffix array={}".format(c.bytes.size))    
+#    print("offset/number of suffix array={}".format(c.offsets.size))    
+#    print("itemsize of suffix array={}".format(c.offsets.itemsize))    
+    print("All the random strings are as follows")
+    for k in range(vsize):
+       print("the {} th random tring ={}".format(k,v[k]))    
+       print("the {} th suffix array ={}".format(k,c[k]))    
+       print("")
+    timings = []
```
pierce314159, line 30: There are a couple places in this PR where print statements are being used. I believe these should be converted into logging statements or raised errors (in the case of `Wrong data type`)

thirdparty/SA/libdivsufsort/include/config.h
```diff
@@ -0,0 +1,81 @@
+/*
+ * config.h for libdivsufsort
+ * Copyright (c) 2003-2008 Yuta Mori All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person
+ * obtaining a copy of this software and associated documentation
+ * files (the "Software"), to deal in the Software without
```
pierce314159, line 7: All the `thirdparty/SA/` code seems to be copied directly from [libdivsufsort](https://github.com/y-256/libdivsufsort). Unless there is a good reason not to, I think it would be best to add this to our list of dependencies and import it. This avoids us having our own copies to maintain.

pierce314159 (changes requested):
I have left comments on a few examples but the gist of my requested changes are:
- Make  `Suffix Array` it's own class separate from `Strings`
- Import the 3rd party C code ([libdivsufsort](https://github.com/y-256/libdivsufsort)) as a dependency to avoid having our own copies to maintain
- Change print statements to logging/errors where appropriate
- Update various formatting/spacing nitpicks for more consistent formatting with the current codebase

Although it's passing the all the tests, I am a little worried that some things from master might've been lost in the merging. My opinion is the safest bet is to apply this functionality onto a fresh copy of master.

I think it would be reasonable to make the suggested changes while applying this functionality onto a fresh copy of master. I would be happy to take point on this if @mhmerrill and @zhihuidu are okay with it

edit: Reposted review with no changes other than shrinking areas being commented on

pierce314159:
I believe a larger discussion about the best way for us to benefit from suffix arrays/data compression using the Burrows-Wheeler algorithm would be very worthwhile. If you don't mind @zhihuidu maybe you could make an issue to explain how you envisioned this functionality being used

benchmarks/sa.py
glitch: The `benchmarks` directory and code might be one of those special cases where prints are ok.  I'm not sure how the benchmarking is run but I think there is some type of aggregation of std-out taking place.  We should probably make a larger/longer-term issue to convert the benchmark output to use a logging framework or module, but for now I think they can get a pass.

Any print statements in the _library_ code should be a no-go.

arkouda/strings.py
zhihuidu: Yes, I agree!

thirdparty/SA/libdivsufsort/include/config.h
zhihuidu: This is a good suggestion. I am not familar with the dependency mechanism and I will learn how to do it.

zhihuidu:
> I believe a larger discussion about the best way for us to benefit from suffix arrays/data compression using the Burrows-Wheeler algorithm would be very worthwhile. If you don't mind @zhihuidu maybe you could make an issue to explain how you envisioned this functionality being used

Yes, Burrows-Wheeler algorithm is a major scenario of suffary array. We can discuss more about it.

src/SACA.chpl
```diff
@@ -0,0 +1,230 @@
...
+
+//Algorithm 2
+
+// The Chapel version of suffix array construction algorithm using skew algorithm
+// Rewrite the algorithm and codes in paper
+// "Simple Linear Work Suffix Array Construction" by Juha Karkkainen and Peter Sanders (2003)
+// Dec.7, 2020
+
+inline proc leq(a1 :int, a2:int, b1:int, b2:int) // lexicographic order
+{  return(a1 < b1 || a1 == b1 && a2 <= b2);
```
pierce314159, line 33: @zhihuidu, I noticed you have an external C implementation of suffix array construction (Algorithm 1) and a native implementation (Algorithm 2). Have you done performance testing to compare the native implementation with the C implementation? If the native implementation is fast, there's no need for us to force the `libdivsufsort` dependency on Arkouda users

zhihuidu: @pierce314159  yes, you are right! Currently the performance of libdivsufsort is better. However it has two problems. (1) it cannot use multiple locales. (2)  we have to make some change on the code to interate it into Arkouda. We hope to develop a new multilocal algorithm to handle very large strings in the next step. For small strings, the performance difference in several seconds or miliseconds is not important. The major reason we use libdivsufsort is that we want to check how to use external code in Arkouda.

pierce314159:
@zhihuidu, I opened a new PR (#865) which replays these changes over a fresh copy of master, address the PR feedback, and makes an effort to minimize diff with master to make reviewing easier. If you are okay with the changes, I'll invite the other team members to review and hopefully we can get your functionality in master

I made a new PR to avoid force pushing and overwriting the history of your `string-suffix-array-functionality` branch. You are welcome checkout my branch (`pierce314159/suffix-array`) and cherry-pick the commits over to your branch, but the merge conflicts are non-trivial

zhihuidu:
> @zhihuidu, I opened a new PR (#865) which replays these changes over a fresh copy of master, address the PR feedback, and makes an effort to minimize diff with master to make reviewing easier. If you are okay with the changes, I'll invite the other team members to review and hopefully we can get your functionality in master
> 
> I made a new PR to avoid force pushing and overwriting the history of your `string-suffix-array-functionality` branch. You are welcome checkout my branch (`pierce314159/suffix-array`) and cherry-pick the commits over to your branch, but the merge conflicts are non-trivial

@pierce314159 Yes, no problem and thanks！

src/SACA.chpl
pierce314159: That makes sense! For more info on how to use external code, you can checkout @glitch's [Investigating Arrow wiki page](https://github.com/glitch/arkouda/wiki/Investigating-Arrow). This is how I was planning to import the library instead of having a copy in our codebase.

In general, I think we want to keep the number of dependencies low. And native chapel implementations are almost always preferred because they can better leverage HPC resources

pierce314159:
Since this PR is fairly out of date and the functionality is captured in #865, I am going to go ahead and close it. If anyone has issue with this feel free to reopen (or ask me to)

Issue #388, opened 2020-06-10
Speed up CI testing


ronawho:
gasnet CI testing takes ~45 minutes, which is a bit long for my taste especially since the comm=none testing only takes 6 minutes. The comm=none config is faster in both build and testing times. I don't think there's much we can do about the build times, but we should be able to improve testing times. I'm hoping we can get gasnet CI below 30 minutes. Things to try:
 - [x] use gasnet-smp instead of gasnet-udp (will use shared memory for comm instead of going through the kernel with udp)
 - [ ] try to speed up long running tests 
   - I suspect the sorting tests are the bottleneck. Even though we're sorting a small number of elements, the sort code has `2**16` buckets, which results in a lot fo communication regardless of how many elements are being sorted. We may just want to run fewer sorting tests as part of CI and rely on nightly testing or something for more exhaustive tests.

ronawho:
comm=none `make python` takes 25s and gasnet-smp takes 19 minutes. Here's all tests under gasnet-smp that take over 5 seconds:

```
523.45s call     tests/groupby_test.py::GroupByTest::test_groupby_one_two_levels
123.86s call     tests/coargsort_test.py::CoargsortTest::test_int
 75.38s call     tests/operator_tests.py::OperatorsTest::testAllOperators
 67.09s call     tests/join_test.py::JoinTest::test_join_on_eq_with_true_dt_with_result_limit
 61.59s call     tests/join_test.py::JoinTest::test_join_on_eq_with_abs_dt_outside_window
 59.67s call     tests/coargsort_test.py::CoargsortTest::test_large
 39.85s call     tests/coargsort_test.py::CoargsortTest::test_float
 23.84s call     tests/setops_test.py::SetOpsTest::testOneDimensionalSetXor
 22.47s call     tests/coargsort_test.py::CoargsortTest::test_int_float
 11.09s call     tests/compare_test.py::CompareTest::test_compare_argsort
 11.04s call     tests/setops_test.py::SetOpsTest::testOneDimensionalSetUnion
 10.60s call     tests/setops_test.py::SetOpsTest::testOneDimensionalSetIntersection
  9.14s call     tests/setops_test.py::SetOpsTest::testOneDimensionalSetDifference
  9.03s call     tests/compare_test.py::CompareTest::test_compare_sort
```

ronawho:
Current results where gasnet-smp took 48 minutes total (I'm just adding `--duration=0` to the pytest command to get this and running in CI):

```
[...]

PR #629, opened 2021-01-13
Print out arkouda unit test times

ronawho:
Print out test duration for any tests that take longer than 5 seconds
for standard comm=none and comm=gasnet-smp arkouda unit test runs.

Motivated by #388. I've been doing this manually from time to time, but
seems like it could be valuable to have the history in the CI logs.


reuster986 (approved):
I agree this would be good information to have.

PR #631, opened 2021-01-15
Groupby benchmark

reuster986:
Adds a benchmark for `ak.GroupBy`. By default, it mimics the `coargsort` benchmark and runs over `(1, 2, 4, 8, 16)` `int64` arrays. Users can also request to run over `str` or both dtypes, but this is extremely slow on my laptop, and I'm not quite sure why.


benchmarks/groupby.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+
+import time, argparse
+import numpy as np
```
ronawho, line 4: Is there a numpy equivalent to groupby, or is this more in the pandas realm?

Definitely not required, but it can be nice to have a reference benchmark to compare serial performance to to get a baseline.

ronawho (approved):
👍 Looks good. I can add the nightly graphing infrastructure code in a followup PR.

benchmarks/groupby.py
mhmerrill: groupby is pandas functionality,

reuster986: Also, I believe groupby is a lazy operation in pandas, so to get a true comparison we would need to pair it with a groupwise reduction. This might be a good idea, actually, because some of that functionality is not exercised by existing benchmarks. But even though we can't compare groupby alone vs pandas, it would still be nice to benchmark arkouda's groupby separately from the reductions. I guess I just punted on the design questions and made this benchmark as bare as possible, but I would be open to adding more complexity.

PR #633, opened 2021-01-20
Flatten feature

reuster986:
This PR is in answer to a request from a couple of our users, who wanted the ability to regularize some semi-structured string data in arkouda. The `Strings.flatten(delimiter)` method treats each string in the array as a variable-length, delimited list of values and unpacks it into a flat array, with one value per element. 

In addition to the code itself, this PR adds a unit test and documentation for `flatten` and updates the docs for some related functions that have changed.


hokiegeek2 (commented):
Looks good to me

mhmerrill (approved):
looks ok to me

PR #635, opened 2021-01-20
Correct Strings.__init__ typing

mhmerrill:
correct the typing of the `Strings.__init__` to accept `pdarray` and `str` types instead of `np.ndarray`


reuster986 (approved):
good catch

PR #637, opened 2021-01-21
Resolve a SipHash regression from 605

### LINKED ISSUE #636: Hash performance regression
For some reason, the SipHash performance has tanked. On my laptop, hash rates measured with `SipHashSpeedTest` are down about a factor of 1000 from what they were after the work in #266 . I'm not sure when the regression happened because we don't have a benchmark, which is my fault for losing track of the request in #356 . 

Sample runs on my laptop:

```
$ ./SipHashSpeedTest -nl 1 --NINPUTS=100_000
Hashed 100000 blocks (6.0 MB) in 52.89 seconds (0.11 MB/s)

$ ./SipHashSpeedTest -nl 4 --NINPUTS=100_000
Hashed 100000 blocks (6.0 MB) in 35.32 seconds (0.17 MB/s)
```

I think we should be seeing in the range of 100 MB/s, based on the discussion in #266 .

@ronawho do you have time to look at this soon-ish? I will also try to figure out what's going on, but there's a good chance you will see something that I miss.

ronawho:
In #605 a param flag around the `TRACE` function was removed, which led to a
severe performance degradation for SipHash. What was previously a no-op,
became a `logger.debug()` call. And even though that logger.debug() call was
itself a no-op, there was still a lot of string creation and concatenation at
the call site. Add the param flag back to resolve the regression.

Performance on the nightly chapcs machine with comm=none:

Before:

```
make test-bin/SipHashSpeedTest
./test-bin/SipHashSpeedTest --NINPUTS=100_000
> Hashed 100000 blocks (6.0 MB) in 63.91 seconds (0.09 MB/s)
```

Now:

```
make test-bin/SipHashSpeedTest
./test-bin/SipHashSpeedTest --NINPUTS=100_000
> Hashed 100000 blocks (6.0 MB) in 0.00 seconds (17391.30 MB/s)
./test-bin/SipHashSpeedTest --NINPUTS=100_000_000
> Hashed 100000000 blocks (6152.0 MB) in 0.31 seconds (20112.27 MB/s)
```

Resolves #636

### DIFF
--- a/src/SipHash.chpl
+++ b/src/SipHash.chpl
@@ -10,6 +10,8 @@ module SipHash {
   param cROUNDS = 2;
   param dROUNDS = 4;
 
+  private config param DEBUG = false;
+
   const defaultSipHashKey: [0..#16] uint(8) = for i in 0..#16 do i: uint(8);
 
   const shLogger = new Logger();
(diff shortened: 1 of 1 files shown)


reuster986 (approved):
Thanks! I would have had a lot of trouble finding this...

PR #638, opened 2021-01-21
634 typeguard too strict with scalars

hokiegeek2:


reuster986 (approved):
Thanks @hokiegeek2 !

reuster986:
@hokiegeek2 looks like there's still one merge conflict around `_as_dtype(dt)`. I had removed the `@typechecked` decorator and the type hint from `dt` because I could not get it to accept a `Union[np.dtype, str]` like I wanted. If you want to give it a try, you're welcome to, or we could just leave the typing off that function.

hokiegeek2:
> @hokiegeek2 looks like there's still one merge conflict around `_as_dtype(dt)`. I had removed the `@typechecked` decorator and the type hint from `dt` because I could not get it to accept a `Union[np.dtype, str]` like I wanted. If you want to give it a try, you're welcome to, or we could just leave the typing off that function.

Gotcha, well, it's an internal method, so I'd say it's cool with leaving typeguard off

Issue #356, opened 2020-04-17
Add a string sort / SipHash benchmark


ronawho:
Add a benchmark for sorting strings and/or for hashing strings. 

Hashing was optimized in https://github.com/mhmerrill/arkouda/issues/266 and it'd be great to have a benchmark we can run nightly to track future improvements and to ensure we don't let any regressions slip in.

ronawho:
No rush on this, and I'm happy to help with the benchmarking infrastructure I just wasn't sure what the arkouda code itself would look like.

ronawho:
May also be worth having a string gather benchmark since https://github.com/mhmerrill/arkouda/pull/275 special cased for ugni vs non-ugni. (Not sure if that needs a separate benchmark, or if there will be a gather included in string sorting.)

ronawho:
See https://github.com/mhmerrill/arkouda/issues/527 for initial string gather benchmark.

PR #640, opened 2021-01-22
String benchmarks

reuster986:
In answer to #356 , this PR adds `str` dtype to the groupby and argsort benchmarks and includes them in the nightly performance testing. Running groupby on `str` exercises the siphash functionality, so we should see any future regressions. There is more string functionality that we could benchmark, but I think these ones cover the most important cases.

@ronawho would you mind checking to make sure I handled all the benchmark graphing infrastructure correctly? It seems to work when I do `benchmarks/run_benchmarks.py --gen-graphs`, but I can't tell for sure.


benchmarks/argsort.py
```diff
@@ -13,8 +13,13 @@ def time_ak_argsort(N_per_locale, trials, dtype, seed):
     print("numLocales = {}, N = {:,}".format(cfg["numLocales"], N))
     if dtype == 'int64':
         a = ak.randint(0, 2**32, N, seed=seed)
+        nbytes = a.size * a.itemsize
     elif dtype == 'float64':
         a = ak.randint(0, 1, N, dtype=ak.float64, seed=seed)
+        nbytes = a.size * a.itemsize
+    elif dtype == 'str':
+        a = ak.random_strings_uniform(1, 8, N, seed=seed)
+        nbytes = (a.offsets.size * a.offsets.itemsize) + (a.bytes.size * a.bytes.itemsize)
```
ronawho, line 22: Should offsets be left off here? It's technically part of the underlying implementation, but based on the discussion in https://github.com/mhmerrill/arkouda/issues/531#issuecomment-712370644 I wonder if it makes more sense to just track given "X GiB of string data from a user, how fast does that sort" and ignore how we represent/manipulate it internally

benchmarks/graph_infra/arkouda.graph
```diff
@@ -4,9 +4,15 @@ files: stream.dat
 graphtitle: Stream Performance
 ylabel: Performance (GiB/s)
 
-perfkeys: Average rate =
-graphkeys: Argsort GiB/s
-files: argsort.dat
+perfkeys: 1-array (int64) Average rate =, 2-array (int64) Average rate =, 8-array (int64) Average rate =, 16-array (int64) Average rate =, 1-array (str) Average rate =, 2-array (str) Average rate =, 8-array (str) Average rate =, 16-array (str) Average rate =, 2-array (mixed) Average rate =, 8-array (mixed) Average rate =, 16-array (mixed) Average rate =
+graphkeys: 1-array (int64) GroupBy GiB/s, 2-array (int64) GroupBy GiB/s, 8-array (int64) GroupBy GiB/s, 16-array (int64) GroupBy GiB/s, 1-array (str) GroupBy GiB/s, 2-array (str) GroupBy GiB/s, 8-array (str) GroupBy GiB/s, 16-array (str) GroupBy GiB/s, 2-array (mixed) GroupBy GiB/s, 8-array (mixed) GroupBy GiB/s, 16-array (mixed) GroupBy GiB/s
+files: groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat, groupby.dat
```
ronawho, line 9: This graph will likely have too many lines on it to be readable.

ronawho:
> @ronawho would you mind checking to make sure I handled all the benchmark graphing infrastructure correctly? It seems to work when I do benchmarks/run_benchmarks.py --gen-graphs, but I can't tell for sure.

https://github.com/mhmerrill/arkouda/blob/master/benchmarks/graph_infra/README.md has instructions for how to view the graphs locally. Should be something like:

```
./benchmarks/run_benchmarks.py --gen-graphs
cd benchmarks/datdir/html
python3 -m http.server 8000
open http://localhost:8000/ (or navigate to localhost:8000 in your browser)
```

---

When I've been thinking about having string performance graphs, I've mostly been thinking about having separate graphs (to avoid having overly busy graphs and because I expect the performance of say int sort vs str sort to be different enough that having them on the same graph would make it hard to see regressions at a glance). 

Do you mind if we sit on this over the weekend, and I'll try to throw together what I was thinking in terms of graphing infrastructure?

benchmarks/argsort.py
reuster986: Good point. I removed offsets.

reuster986:
I see what you mean about plotting int and str separately. I'm fine with waiting on this PR for a while to think about how we want it.

reuster986:
For groupby, argsort, and gather, I split the plots into numeric vs. string benchmarks. @ronawho let me know what you think.

ronawho:
> For groupby, argsort, and gather, I split the plots into numeric vs. string benchmarks. @ronawho let me know what you think.

Yeah, I like this better. 

Popping up a level -- there's 3 main ways I use the benchmarks:
 - Nightly graphs and triaging said graphs. For these I think we want separate numeric and string variants to make triaging easier. I expect the performance delta to be large enough that it'd be hard to tell at a glance if there were regressions in the string case. You addressed that point with the latest commit
 - Tuning a specific benchmark. e.g. if I'm working on tuning one benchmark, I'll just do `./benchmarks/run_benchmarks.py <benchmark>`. With the current PR, I'd either have to specify the type or would end up running all types. This is a minor inconvenience, but I do find it slightly easier for each benchmark to have a default. I also wonder as we extend this to other benchmarks if it's worth running/graphing all types. e.g. is it worth running and graphing float64 gathers? I think this one is tricky, because there's a tension between wanting to graph what's interesting, but not wanting to have too many graphs that it's hard to triage. On the Chapel side of things we have created so many graphs over time that it's not practical to run them manually and we ignore a large fraction of the graphs we have for triaging. On the other hand when we make specific optimizations it's sometimes nice to have graphs that we don't ordinarily look it.
 - Generating scalability plots. For this I'm typically just running all the benchmarks with the default and a large problem size (`--size=$((2**30))`). Historically this worked well because the large problem size scaled pretty consistently across benchmarks. However with the string benchmarks we'll probably want to lower the problem size. I think you do that in your PR (` // 100`), but it does make me think about just adding `--size=large` or something variants where the benchmark defines what the default/large problem sizes are.

And one important (and annoying) limitation to be aware of with our graphing infrastructure is that the .dat files we generate are pretty rigid in the sense that it's not easy to add or remove keys from existing .dat files. If you run `./benchmarks/run_benchmarks.py --gen-graphs` before and after this change you'll probably get failures. I can correct those manually for the nightly graphs, but it's something to be aware of.

I put together an alternative proposal for adding string benchmarks in https://github.com/mhmerrill/arkouda/pull/642 (Only adds string argsort to start.) The big difference is that instead of having argsort run all supported variants, it creates a `str-argsort` version. The main reason I prefer this is because it makes it trivial to run benchmarks for specific turning efforts and it will not require changes to the .dat files we generate. It also makes it a little harder to run all variants, which can be a plus or a negative. The downside to this approach is that it kinda arbitrarily adds an extra file for benchmarking and there could be other tradeoffs.

Anyways, let me know what you think, I'm happy to make either direction work. (And just FYI while I'm trying to avoid some of the mistakes me made with the Chapel performance graphs, it's also possible I'm way overthinking the solution for running/graphing string tests here.)

reuster986:
Closing in favor of #652

Issue #644, opened 2021-01-27
ensure string_test delimiter is not in gremlins array


hokiegeek2:
Today I ran into an instance where the string_test self.delim was in the gremlins array (in this case, ' '). Recommend adding recursive helper method to ensure gremlin delimiters are avoided

reuster986:
@hokiegeek2 I would expect `' '` (space) to be a common delimiter. Shouldn't we support it?

reuster986:
Same with `'"'` (double quote), which should be a valid delimiter. But we should throw an error on `''` (empty string), since it makes no sense to split on it or ask whether a string ends with it.

hokiegeek2:
@reuster986 Good points! Yeah, I agree that we should support space and double quote as delimiters. That's something I'll be happy to work on at your's and Mike's direction.

hokiegeek2:
Merged in code changes, issue resolved.

PR #645, opened 2021-01-27
added recursive _get_delimiter method to avoid gremlin delimiters #644

hokiegeek2:


tests/string_test.py
```diff
@@ -295,25 +294,31 @@ def setUp(self):
...
-        self.gremlins_base_words = base_words = ak.concatenate((base_words1, base_words2, gremlins))
-        self.gremlins_strings = ak.concatenate((base_words[choices], gremlins))
+        self.gremlins_base_words = ak.concatenate((self.base_words, self.gremlins))
+        self.gremlins_strings = ak.concatenate((self.base_words[choices], self.gremlins))
         self.gremlins_test_strings = self.gremlins_strings.to_ndarray()
         self.gremlins_cat = ak.Categorical(self.gremlins_strings)
 
+    def _get_delimiter(self,x : Tuple, w : Tuple, gremlins : np.ndarray) -> str:
+        delim = np.random.choice(x, p=(np.array(w)/sum(w)))
+        if delim in gremlins:
```
reuster986, line 315: Rather than requiring `delim` not to be in `gremlins`, we should just make sure it's not the empty string. The other two gremlins (`"` and ` `) are valid delimiters.

reuster986 (changes requested):
Looks mostly good, but see inline comment for one requested change.

tests/string_test.py
hokiegeek2: @reuster986 I made the change you requested and test_ends_with is failing when the delimiter = " " (one space):

image

hokiegeek2: @reuster986 I see what you mean about invalid delimiters. Both the double quote and space work correctly as long as the last string in the Strings object does not end with the delimiter:

image

And, if the delimiter is an empty string, it fails for all strings in the Strings object that are an empty string.

@reuster986 @mhmerrill In this PR I am fixing the manifestation, not the root problem, where the root problem is that the ends_with logic fails when the last string in the Strings object ends with the delimiter. All I am attempting to do with this PR is to ensure the CI workflow succeeds. Please confirm how you'd like to proceed with this PR.

reuster986 (approved):
I approve this as a workaround, with intent to address root cause later.

Issue #632, opened 2021-01-19
Reading hdf5 files with ak.load does not scale 


ronawho:
Something about reading HDF5 files via `ak.load` appears to be serialized and does not scale as the number of locales increases. Writing on the other hand does appear to scale.

Here's some preliminary data from the 16-node-cs-hdr machine we run some nightly testing on (rough specs listed in https://chapel-lang.org/perf/arkouda/). This is for the default problem size (3/4 GiB per locale) running on lustre backed by hard drives (not flash/nvme):

| locales | read       | write       |
| ------- | ---------: | ----------: |
|  1      | 3.82 GiB/s |  0.74 GiB/s |
|  2      | 1.92 GiB/s |  1.56 GiB/s |
|  4      | 2.52 GiB/s |  3.57 GiB/s |
|  8      | 1.85 GiB/s |  5.10 GiB/s |
| 16      | 1.96 GiB/s | 11.10 GiB/s |
| 32      | 1.85 GiB/s | 23.07 GiB/s |

I have no idea what this system is supposed to be capable of in terms of read/write speeds. Write scaling looks good (performance roughly doubles when we double the number of locales) but something about reads seems to be serialized.

We see this for a bigger Cray XC as well. It's also on lustre, though I'm not sure how provisioned it is. Regardless the scaling behavior is similar, though writes starts to fall off at 64 nodes. I'm not sure yet if we're hitting a hardware limit with this particular system or if this is a bottleneck on the software/hardware side. 

| locales | read       | write       |
| ------- | ---------: | ----------: |
|   1     | 2.59 GiB/s |  1.00 GiB/s |
|   2     | 1.82 GiB/s |  1.67 GiB/s |
|   4     | 2.02 GiB/s |  3.27 GiB/s |
|   8     | 2.52 GiB/s |  6.46 GiB/s |
|  16     | 2.51 GiB/s | 12.00 GiB/s |
|  32     | 2.69 GiB/s | 21.15 GiB/s |
|  64     | 2.68 GiB/s | 18.10 GiB/s |
| 128     | 2.82 GiB/s | 22.49 GiB/s |
| 256     | 3.33 GiB/s | 27.65 GiB/s |

ronawho:
FYI @mhmerrill @reuster986. I expect to start looking into this more in the near-ish term, but wanted to share preliminary results. 

[...]

PR #647, opened 2021-02-02
Run the IO benchmark as part of regular benchmarking

ronawho:
Add the new IO benchmark to the benchmarks suite and run it as part of
regular/nightly testing.

Also add some standalone IO speed tests.  One that calls
`write1DDistArray()`/`read_files_into_distributed_array()` directly, and
another that calls the msg wrappers `tohdfMsg()`/`readAllHdfMsg()`.
These are being used to investigate #632 and seem like they could be
useful in the longer term.

### DIFF
--- a/benchmarks/IO.py
+++ b/benchmarks/IO.py
@@ -33,12 +33,12 @@ def time_ak_write_read(N_per_locale, trials, dtype, path, seed):
     avgwrite = sum(writetimes) / trials
     avgread = sum(readtimes) / trials
 
-    print("Write times: min = {:.4f} sec, max = {:.4f} sec, avg = {:.4f} sec".format(min(writetimes), max(writetimes), avgwrite))
-    print("Read times : min = {:.4f} sec, max = {:.4f} sec, avg = {:.4f} sec".format(min(readtimes), max(readtimes), avgread))
+    print("write Average time = {:.4f} sec".format(avgwrite))
+    print("read Average time = {:.4f} sec".format(avgread))
 
     nb = a.size * a.itemsize
-    print("Write rates: min = {:.4f} GiB/sec, max = {:.4f} GiB/sec, avg = {:.4f} GiB/sec".format(nb/2**30/max(writetimes), nb/2**30/min(writetimes), nb/2**30/avgwrite))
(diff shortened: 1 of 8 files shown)


ronawho:
Not sure what's up with the mypy failures -- https://github.com/mhmerrill/arkouda/pull/647/checks?check_run_id=1811459415#step:5:1:


```
arkouda/dtypes.py:22: error: Module has no attribute "bool"; maybe "bool_" or "bool8"?
arkouda/dtypes.py:27: error: Module has no attribute "str"; maybe "str_" or "str0"?
arkouda/dtypes.py:91: error: Name 'np.bool' is not defined
arkouda/dtypes.py:99: error: Name 'np.float' is not defined
arkouda/dtypes.py:101: error: Module has no attribute "str"; maybe "str_" or "str0"?
arkouda/dtypes.py:105: error: "dtype[Any]" has no attribute "dtype"; maybe "type" or "subdtype"?
...
```

I don't think I'm changing any client code, so maybe an upstream change to mypy? @hokiegeek2 any ideas?

benchmarks/IO.py
```diff
@@ -33,12 +33,12 @@ def time_ak_write_read(N_per_locale, trials, dtype, path, seed):
     avgwrite = sum(writetimes) / trials
     avgread = sum(readtimes) / trials
 
-    print("Write times: min = {:.4f} sec, max = {:.4f} sec, avg = {:.4f} sec".format(min(writetimes), max(writetimes), avgwrite))
-    print("Read times : min = {:.4f} sec, max = {:.4f} sec, avg = {:.4f} sec".format(min(readtimes), max(readtimes), avgread))
+    print("write Average time = {:.4f} sec".format(avgwrite))
```
reuster986, line 36: I had included the min and max time because I/O rates are notoriously affected by buffering, but on second thought, the best solution for that is probably just to restrict it to a single trial.

reuster986 (approved):
I think this looks good. I think it might be good to reduce the default number of trials to 1 for the python benchmark (see inline comment), but I'll leave that decision up to you.

benchmarks/IO.py
ronawho: Yeah, makes sense -- lowering to 1 trial.

reuster986:
@ronawho , #653 disabled mypy, so you should be able to pull master and pass the CI.

reuster986 (approved):

PR #652, opened 2021-02-03
Add string benchmarks

ronawho:
Add string versions of argsort, coargsort, groupby, and gather to run as
part of nightly performance testing.

The approach taken here is to create separate `str-<bench>` benchmark
files that just call into the main `<bench>` files with adjusted problem
sizes and number of trials. This makes it easier to fit into the
graphing infrastructure without running all the type variants. See
discussion in https://github.com/mhmerrill/arkouda/pull/640

Resolves https://github.com/mhmerrill/arkouda/issues/356
Resolves https://github.com/mhmerrill/arkouda/issues/534


ronawho:
Probably easiest to review commit-by-commit. Sample graphs for 16-node-cs-hdr are available [here](https://ronawho.github.io/ak-652/?startdate=2021/02/01&enddate=2021/02/04&graphs=argsortperformance,coargsortperformance,groupbyperformance,gatherperformance,stringargsortperformance,stringcoargsortperformance,stringgroupbyperformance,stringgatherperformance)

benchmarks/IO.py
```diff
@@ -8,7 +8,7 @@
 TYPES = ('int64', 'float64')
 
 def time_ak_write_read(N_per_locale, trials, dtype, path, seed):
-    print(">>> arkouda write/read")
+    print(">>> numpy {} write/read".format(dtype))
```
reuster986, line 11: This looks like it should be arkouda, rather than numpy...

reuster986 (approved):
This looks good to me, aside from a minor fix in one of the print statements.

benchmarks/IO.py
ronawho: Whoops, good catch -- must have screwed up a `sed` command. Fixed now

reuster986:
ditto here, pulling master should fix CI

reuster986 (approved):

reuster986:
@ronawho small merge conflict with the string benchmark PR. Looks like it's just unioning the list of benchmarks to run.

ronawho:
Rebased

PR #657, opened 2021-02-04
656 specific pandas and numpy versions needed for datetime functionality

hokiegeek2:


reuster986 (approved):
Confirmed that our internal environment supports this configuration.

PR #664, opened 2021-02-08
Replace "operator" variable with "op"

lydia-duncan:
`operator` will be a keyword in Chapel 1.24, so rename this local variable in
anticipation of that change.

----


lydia-duncan:
Without this change, the arkouda Chapel smoke test breaks.  It seems to be working with it (python tests passed locally)

bradcray:
@mhmerrill / @reuster986 : For additional context, we're in the process of revamping how operators in Chapel are expressed and associated with types / `import`ed, which has resulted in reserving `operator` as a keyword.  Unfortunately, Arkouda was already using it as a variable name, necessitating this change.  Sorry for the hassle.

mhmerrill (approved):
this look ok to me

PR #666, opened 2021-02-11
646 string test test ends with periodically fails

hokiegeek2:
This PR contains the following:

1. Fix Strings.endswith where a match is missed if the user-supplied delimiter matches the last character in the last string within the Strings object.
2. Added corresponding test case for Strings.endswith fix
3. Fixed broken gremlins-related tests in test_ends_with, test_peel, test_stick

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -494,7 +494,7 @@ module SegmentedArray {
         // Position where substr aligns with end of segment must be a hit
         // -1 for null byte
         hits[oD.interior(-(oD.size-1))] = truth[oa[oD.interior(oD.size-1)] - substr.numBytes - 1];
-        hits[oD.high] = truth[D.high];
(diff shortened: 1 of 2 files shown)


src/SegmentedArray.chpl
```diff
@@ -494,7 +494,7 @@ module SegmentedArray {
         // Position where substr aligns with end of segment must be a hit
         // -1 for null byte
         hits[oD.interior(-(oD.size-1))] = truth[oa[oD.interior(oD.size-1)] - substr.numBytes - 1];
-        hits[oD.high] = truth[D.high];
+        hits[oD.high] = truth[D.high-1];
```
reuster986, line 497: @hokiegeek2 Great catch!

reuster986 (approved):
Good work! How appropriate that the PR that finally fixes this diabolical bug is number 666.

PR #667, opened 2021-02-11
Fix connect() call in registration_test.py

ronawho:
`test_attach()` was doing a plain `connect()`, which tries to connect to
localhost instead of the server/port the testing systems knows. This
would cause hang in our nightly distributed testing.

Similar to https://github.com/mhmerrill/arkouda/pull/613

### DIFF
--- a/tests/registration_test.py
+++ b/tests/registration_test.py
@@ -64,7 +64,7 @@ class RegistrationTest(ArkoudaTest):
                                       aar_array.to_ndarray()).all())
         
         ak.disconnect()
-        ak.connect()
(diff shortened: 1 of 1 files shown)


tests/registration_test.py
```diff
@@ -64,7 +64,7 @@ def test_attach(self):
                                       aar_array.to_ndarray()).all())
         
         ak.disconnect()
-        ak.connect()
+        ak.connect(server=ArkoudaTest.server, port=ArkoudaTest.port)
```
hokiegeek2, line 67: whoa, good catch, sorry that no-args connect() method call got in there.

hokiegeek2 (approved):

Issue #663, opened 2021-02-08
Use aggregation in segmented reductions


reuster986:
Groupby aggregations (e.g. `ak.GroupBy(x).sum(y)`) use the segmented reductions in `src/SegmentedMsg.chpl`. These aggregations can be pretty slow. I always subconsciously blamed it on scan performance, but when I looked into it just now, scan is much faster than these aggregations (which we do not have benchmarks for yet). I'm beginning to suspect that the problem is that the segmented reductions have not been updated to use comm aggregation. For example, segmented sum does the following:

https://github.com/mhmerrill/arkouda/blob/4a3855fd940160355880a5194736500fb896d982/src/ReductionMsg.chpl#L403-L415

Which performs two potentially non-local gets per segment without aggregation, so lots of potential slowdown there.

Additionally, `segAny` and `segAll` use a non-vectorized pattern that is slower than `segSum`. Implementing these in terms of `segSum` (with aggregation) would speed them up as well.

To do:
- [x] Update the code to use aggregation
- [x] Fix `segAny` and `segAll` to use `segSum`
- [x] Create a benchmark for groupby aggregations

reuster986:
closed by #669

PR #669, opened 2021-02-12
663 Update reductions

reuster986:
Addresses #663 by using comm aggregation in segmented reductions and adding benchmarks.

I also added 3 new reductions for bitwise OR, AND, and XOR. The first two perform rather poorly, and I might ask @ronawho to take a look at the logic if you have time:

https://github.com/mhmerrill/arkouda/blob/4f4d71e34a743c00036657725e455ced00de7114/src/ReductionMsg.chpl#L638-L667


reuster986:
Hmm, one of the tests is failing after pulling master... I'll have to look into that later.

ronawho:
>  I might ask @ronawho to take a look at the logic if you have time:

No problem. I'm heading out soon for today, but I should have time to start looking Tuesday or Wednesday.

> Hmm, one of the tests is failing after pulling master... I'll have to look into that later.

When running the benchmark I'm also seeing:

```
RuntimeError: message: skipNan must be of type bool
```

benchmarks/graph_infra/arkouda.graph
```diff
@@ -22,6 +22,24 @@ files: groupby.dat, groupby.dat, groupby.dat, groupby.dat
 graphtitle: Groupby Performance
 ylabel: Performance (GiB/S)
 
+perfkeys: Aggregate sum Average rate =, Aggregate xor Average rate =, Aggregate mean Average rate =
+graphkeys: sum (GiB/s), xor (GiB/s), mean (GiB/s)
+files: aggregate.dat, aggregate.dat, aggregate.dat
+graphtitle: Grouped Aggregation Performance (Tier 1)
```
ronawho, line 28: Do the tiers indicates which ones you care most about, or is it somewhat arbitrary just to avoid putting everything on one graph?

mhmerrill:
@reuster986 @ronawho another way to compute segmented reduction is to compute the scan/parallel-prefix operation and pullout/manipulate values at the segment boundaries. This strategy would require less memory and potentially be faster.

benchmarks/graph_infra/arkouda.graph
reuster986: The tiers are by speed, so that slow ops won't get lost in the noise floor when graphed alongside fast ops. I put them into tiers based on the rates I observed on a single locale, so these tiers might be inappropriate for multi-locale.

reuster986:
@mhmerrill good suggestion - most of the reductions actually do use this strategy already. The exceptions (prior to this PR) were `any` and `all`, but this PR updates them to use the scan pattern (by calling `sum` under the hood). However, I don't think the bitwise `and` and `or` reductions added in this PR are able to use the scan pattern, because manipulating the boundary values requires that the reduction operator have an inverse.

But now that I say it out loud, I realize that `min` and `max` don't have an inverse either, yet I somehow managed to use the scan pattern with them (I wrote that code more than a year ago, so I don't recall the trick right now). I'll have to go back and see if I can apply a similar principle to `and` and `or`.

reuster986:
@bradcray is there a place in the Chapel docs that describes the methods a class needs to implement before it can be used in `myclass reduce x` or `myclass scan x` expressions? I seem to remember seeing an example of a user-defined class like that in the docs a while ago, but I can't find it now.

ronawho:
https://github.com/chapel-lang/chapel/blob/master/doc/rst/technotes/reduceIntents.rst has an example. There's also a use in arkouda for the KReduce:

https://github.com/mhmerrill/arkouda/blob/146cfd101cabf871d85e286c1143794968976e1b/src/KReduce.chpl#L31-L77

bradcray:
@reuster986: I'm sorry to say that there are not, and that it's intentional—because we aren't satisfied with the current interface and have plans to improve it, we don't want to imply more commitment to it than we have.  In practice, people (myself included) usually look at an existing reduction in the module or test code and copy what's there.  For example,

```chapel
  class MaxReduceScanOp: ReduceScanOp {
    type eltType;
    var value = min(eltType);

    proc identity return min(eltType);
    proc accumulate(x) {
      value = max(x, value);
    }
    proc accumulateOntoState(ref state, x) {
      state = max(state, x);
    }
    proc combine(x) {
      value = max(value, x.value);
    }
    proc generate() return value;
    proc clone() return new unmanaged MaxReduceScanOp(eltType=eltType);
  }
```

or:

```chapel
  class BitwiseAndReduceScanOp: ReduceScanOp {
    type eltType;
    var value = _band_id(eltType);

    proc identity return _band_id(eltType);
    proc accumulate(x) {
      value &= x;
    }
    proc accumulateOntoState(ref state, x) {
      state &= x;
    }
    proc combine(x) {
      value &= x.value;
    }
    proc generate() return value;
    proc clone() return new unmanaged BitwiseAndReduceScanOp(eltType=eltType);
  }
```

reuster986:
@bradcray @ronawho thank you both!

src/ReductionMsg.chpl
```diff
@@ -625,30 +621,100 @@ module ReductionMsg
...
+        // Some results will have to be sent to remote locales
+        ref start = values[s];
+        const startLocale = start.locale.id;
+        const hereLocale = here.id;
+        if startLocale == hereLocale {
+          r = | reduce values[s..e];
+        } else {
+          on startLocale {
+            r = | reduce values[s..e];
+          }
```
bradcray, line 663: I just happened to notice this from the excerpt in the OP:  I wouldn't expect this type of specialization for locality in the event that the on-clause refers to the current locale to be necessary.  The Chapel compiler should already be doing this kind of optimization automatically for every on-clause.  If you find that this kind of manual specialization results in an improvement, that's something we'd definitely want to know about.

reuster986:
I added a performance test for bitwise reductions (`test/SpeedTestSegReduce.chpl`). On my laptop (both single and multi-locale), the custom `ReduceScanOp` child class outperforms the per-segment reduction that I had been using, so I put it in the source code. However, the custom class throws a compiler warning about scans being serialized...

@ronawho would you mind running the performance test on XC and CS when you get a chance, just to make sure "Method 2 (custom scan class)" is outperforming "Method 1 (on startLocale)"?

Also, the compiler warnings appear to be making the Chapel unit tests fail for some reason. Does anyone know how to ignore compiler warnings in the tests? Or do I have to put the warnings in `*.good`?

ronawho:
> I added a performance test for bitwise reductions (test/SpeedTestSegReduce.chpl). On my laptop (both single and multi-locale), the custom ReduceScanOp child class outperforms the per-segment reduction that I had been using, so I put it in the source code. However, the custom class throws a compiler warning about scans being serialized...

> Also, the compiler warnings appear to be making the Chapel unit tests fail for some reason. Does anyone know how to ignore compiler warnings in the tests? Or do I have to put the warnings in *.good?

You could add `--no-warnings` to the COMPOPTS file (but that's a big hammer) or you could put the warning in the .good file. However, in this case, the warning indicates a serious performance penalty from not parallelizing the scan. Parallel scans currently require the input and output of the scan to have the same type. I'm not sure if that's an inherent limitation or if that was just something done in the first draft of parallel scans. @bradcray should know more.

The following patch should get rid of the warning by making the input/output type the same. This allows the scan to be parallelized (and distributed), which should further improve performance.

```diff
diff --git a/test/SpeedTestSegReduce.chpl b/test/SpeedTestSegReduce.chpl
index f49afe3..4297b96 100644
--- a/test/SpeedTestSegReduce.chpl
+++ b/test/SpeedTestSegReduce.chpl
@@ -73,9 +73,9 @@ proc segOr2(values:[?vD] int, segments:[?D] int): [D] int {
   }
   const scanresult = ResettingOrScanOp scan flagvalues;
   forall (r, s) in zip(res[..D.high-1], segments[D.low+1..]) with (var agg = newSrcAggregator(int)) {
-    agg.copy(r, scanresult[s-1]);
+    agg.copy(r, scanresult[s-1](1));
   }
-  res[D.high] = scanresult[vD.high];
+  res[D.high] = scanresult[vD.high](1);
   return res;
 }

@@ -104,11 +104,10 @@ class ResettingOrScanOp: ReduceScanOp {
   }

   proc generate() {
-    const (reset, val) = value;
-    return val;
+    return value;
   }

   proc clone() {
-    return new unmanaged ResettingOrScanOp(eltType);
+    return new unmanaged ResettingOrScanOp(eltType=eltType);
   }
-}
\ No newline at end of file
+}
```

ronawho:
> @ronawho would you mind running the performance test on XC and CS when you get a chance, just to make sure "Method 2 (custom scan class)" is outperforming "Method 1 (on startLocale)"?

Yup, even serialized the "custom scan class" is way outperforming the "on startLocale" version. The normal XC and CS configurations I run on are busy at the moment, but here are some results from the 16-node-cs-hdr machine using a different processor (48-core cascadelake instead of the usual 128-core rome.) The trends should be the same, but the exact number will be different:

```
make test-bin/SpeedTestSegReduce
./test-bin-opt/SpeedTestSegReduce -nl 16
Method 1 (on startLocale): : 3.05s
Method 2 (custom scan class): : 0.07s
```

And doing some quick experiments with parallelizing using the above patch ("on startLocale" removed since it's too slow at larger sizes):

```
make test-bin/SpeedTestSegReduce
./test-bin-opt/SpeedTestSegReduce -nl 16 --NINPUTS=$((10**6))
Method 1 (custom scan class ser): : 0.30s
Method 2 (custom scan class par): : 0.07s

./test-bin/SpeedTestSegReduce -nl 16 --NINPUTS=$((10**8))
Method 1 (custom scan class ser): : 26.72s
Method 2 (custom scan class par): : 0.11s
```

And just the parallel version:

```
make test-bin/SpeedTestSegReduce
./test-bin/SpeedTestSegReduce -nl 16 --NINPUTS=$((10**8))
Method 2 (custom scan class par): : 0.11s

./test-bin/SpeedTestSegReduce -nl 16 --NINPUTS=$((10**9)
Method 2 (custom scan class par): : 0.23s

./test-bin/SpeedTestSegReduce -nl 16 --NINPUTS=$((10**10)
Method 2 (custom scan class par): : 1.47s
```

bradcray:
@reuster986 / @ronawho : Looks like you've worked around it, but I'll refresh my memory as to the reason that input and output types needed to be the same for parallel scan soon, hopefully tomorrow.  I don't remember offhand, and need to poke into the scan code this week for other reasons anyway.  Sorry for the delay.

reuster986:
@bradcray no rush! 

As you've noticed, we worked around the input/output types issue and got the scan parallelized, but now the bitwise OR reduction is failing the correctness test in the CI. I was just wondering if there are any semantics that I'm missing about `ReduceScanOp`, like does the operation have to be commutative or associative? I would think that, for a scan, Chapel would need to guarantee that all operations happen in array order, but maybe my intuition is wrong. 

If nothing comes readily to mind, please don't feel like you have to spend a lot of time on this -- it's likely that I made some other mistake unrelated to the `ReduceScanOp` implementation, so I don't want you to get stuck down a rabbit hole of my making. I won't be able to work on it until Tuesday, anyway.

bradcray:
> I was just wondering if there are any semantics that I'm missing about ReduceScanOp, like does the operation have to be commutative or associative?

Yeah, it needs to be associative for scans.  The reason being that the way you get parallelism for a parallel prefix operation is to apply the operator to the array elements in chunks in parallel (e.g., you scan your half of the array, I'll scan mine), then exchange boundary values (e.g., "what was the result of scanning your half of the array, Bill?"), and then update your segments with the result from your neighbor (e.g., "now I'll update my sub-array results with Bill's result.").  So effectively, I'm applying the operator to my first two elements first, and then combining your result afterwards, which is effectively like doing `(boundaryResult op (first op second))` rather than `((boundaryResult op first) op second)`, which is what array order would produce.

I'm not sure whether scans require the operator to be commutative as well, though reductions do (in that we don't guarantee the order in which we accumulate results globally).

We've discussed ways of creating scan/reduce operations that are not commutative or associative, where the impact is that the parallelism would be more restricted or potentially lost altogether, but haven't pursued that direction yet.  Think of this as one of the potential future changes to the interface.

All that said, is your operator non-commutative / non-associative?

reuster986:
@bradcray thank you for that explanation. I definitely have a better idea of what's going on now, and I will have to think about how that applies to this particular case. The result I'm trying to achieve is a "segmented bitwise reduction", i.e. for each segment of a larger array, reduce all the values in the segment, but don't carry any reductions across segment boundaries. (When I say "segment", I'm using it in the sense of a user-defined partitioning of the array, not in the sense of locality and parallelism that I think you're using it above). Doing this is trivial with an invertible operator like `+`, where you can simply scan and take the difference between the end and the start of each segment as the segment's reduced value. However, with an operator that has no inverse, like bitwise OR, I need to find another way.

My idea in this PR was to create a ternary operator with the following serial behavior:
```chpl
proc resetOR(x, y, reset) {
    return if reset then y else (x | y);
}
```
and then turn it into a `ReduceScanOp`. The `reset` flag basically says "ignore all values before here", so this op is definitely not commutative, because the "ignore" is directional. I'll have to think more about whether it is associative.

Scenario: say I am scanning an array with 10 elements using 2 tasks (assume single-locale for now). Task 1 is scanning with instance A of our `ReduceScanOp` and task 2 is scanning with instance B. If I understand correctly, in step 1, A will scan elements 0..4 and B will scan elements 5..9 independently of each other. Then, in step 2, A will report to B the last value (at index 4) of its chunk, and B will apply that value to all its elements. If this is the correct picture, then my question is, which method of `ReduceScanOp` gets called to update B's partial answers with A's last value?

reuster986:
Based on some quick tests, I have a guess:

Step 1, in parallel:

```chpl
result[0..4] = for x in values[0..4] do A.accumulate(x);
result[5..9] = for x in values[5..9] do B.accumulate(x);
```

Step 2:

```chpl
// If A had a predecessor, I'm guessing there would be a A.combine(preA) here
const Bboundary = A.generate();
for x in result[5..9] do B.accumulateOntoState(x, boundary);
```

Working with this mental model, I'm able to get the scan working correctly on a toy problem with this implementation:
```chpl
class ResettingOrScanOp: ReduceScanOp {
  type eltType;
  /* value is a tuple comprising a flag and the actual result of 
     segmented bitwise OR. 

     The meaning of the flag depends on
     whether it belongs to an array element yet to be scanned or 
     to an element that has already been scanned (or the state of
     an instance doing the scanning). For elements yet to be scanned,
     the flag means "reset to the identity here". For elements that
     have already been scanned, or for internal state, the flag means 
     "there has already been a reset in the computation of this value".
     */
  var value = (false, 0);

  proc identity return (false, 0);

  proc accumulate(x) {
    // Assume x is an element that has not yet been scanned, and
    // that it comes after the current state.
    const (reset, other) = x;
    const (hasReset, v) = value;
    // x's reset flag controls whether value gets replaced or combined
    // also update this instance's "hasReset" flag with x's reset flag
    value = (hasReset | reset, if reset then other else (v | other));
  }

  proc accumulateOntoState(ref state, x) {
    // Assume state is an element that has already been scanned,
    // and x is an update from a previous boundary.
    const (_, other) = x;
    const (hasReset, v) = state;
    // x's hasReset flag does not matter
    // If state has already encountered a reset, then it should
    // ignore x's value
    state = (hasReset, if hasReset then v else (v | other));
  }

  proc combine(x) {
    // Assume x is an instance that scanned a prior chunk.
    const (xHasReset, other) = x.value;
    const (hasReset, v) = value;
    // Since current instance is absorbing x's history,
    // xHasReset flag should be ORed in.
    // But if current instance has already encountered a reset,
    // then it should ignore x's value.
    value = (hasReset | xHasReset, if hasReset then v else (v | other));
  }

  proc generate() {
    return value;
  }

  proc clone() {
    return new unmanaged ResettingOrScanOp(eltType=eltType);
  }
}
```

bradcray:
@reuster986 : With the caveat that I'm not quite awake / "at work" yet, but I think you're on the right track and the one i was going to suggest in response to your previous comment.  Namely, it seemed to me as though you ought to be able to do your segmented computation in a single pass by simply resetting the state whenever you reach a new segment boundary (where a tricky case to test against would be the one in which a segment boundary falls right at the boundary between two tasks' chunks).

Taking a quick look at the scan code, I believe you're right that the first pass will use accumulate() and that the second will use accumulateOntoState().  The one part that didn't look quite right to me was the call to generate() between the two passes, but checking the code, I think you're actually right, and that this might relate back to how the current implementation requires similar input/output/state types.  I still need to refresh my memory as to why that was (where my guess was that it was either a simplifying assumption to get the common cases up and running, or that I was concerned about the amount of memory that would be required if we had to allocate an entire array of state type and output type (?).  That IOU is still in my court.

reuster986:
@bradcray cool, sounds like we are mostly on the same page, with some details left to resolve. The CI appears to be passing locally on my laptop, so I think the functionality in this PR is at least correct.

Regarding the `ReduceScanOp` interface in Chapel, I'm actually quite pleased with how it was able to handle this admittedly strange use case, once my mental model caught up with what was going on. I can't think of any changes to recommend for the interface itself. One suggestion for if/when you do stabilize on the interface: it would be handy to have documentation with a pseudocode-level description of how scans and reductions are implemented, including which `ReduceScanOp` methods get called when, and what the argument types are. For the basic operators, this probably isn't necessary, but it would really help with use cases like this.

bradcray:
> I can't think of any changes to recommend for the interface itself.

Just as a preview, one of the main issues with the current interface is that it isn't really set up well for doing partial reductions (which is the term we use for reducing a subset of an n-dimensional array — say, computing a per-row sum).  So it's less that the current interface is ridiculous as not quite sufficient yet.  That said, I think we should be able to reduce some of the repetition in the current interface, like the need for two accumulate-style calls.  And then there are questions like whether there should be a way to write non-commutative / non-associative operators as well.

> it would be handy to have documentation with a pseudocode-level description of how scans and reductions are implemented

Yeah, I agree.  Especially since I have to go look at the code every time a question like yours comes up.

reuster986:
@ronawho would you mind re-reviewing this one? I think it's ready to go.

ronawho:
> @ronawho would you mind re-reviewing this one? I think it's ready to go.

Yeah, will do -- I should have time on Monday to review

bradcray:
@reuster986: I finally found a little time to dig into scans before the weekend.  First, as a minor correction, I don't _think_ it's the case that the input and output types have to match, but rather that the "state" and output types must match (where the state type is the one that the operator uses to track its state as it goes).  (@ronawho: Do you think you mis-characterized above, or is there a more implicit constraint that I missed?).

It looks to me like that requirement for the parallel implementation was just to optimize for a common/simple case and to avoid having to allocate two temporary arrays:  One of the state type, one of the result type.  I think if this became an annoyance, we could relax it by taking different branches depending on whether the state types were the same or not (and ultimately, definitely should make them parallel regardless of this constraint; it's more a question of how much effort to put in now vs. later).

ronawho:
> @ronawho: Do you think you mis-characterized above, or is there a more implicit constraint that I missed?

Nope you're right, I just mis-characterized

benchmarks/aggregate.py
```diff
@@ -0,0 +1,81 @@
...
+            res = g.aggregate((ones == 1), op)[1]
+        else:
+            res = g.aggregate(ones, op)[1]
+        assert (res.size == g.unique_keys.size)
+
+def create_parser():
+    parser = argparse.ArgumentParser(description="Measure performance of aggregations on grouped arrays.")
+    parser.add_argument('hostname', help='Hostname of arkouda server')
+    parser.add_argument('port', type=int, help='Port of arkouda server')
+    parser.add_argument('-n', '--size', type=int, default=10**7, help='Problem size: total length of all arrays to group')
```
ronawho, line 61: I think we can comfortably bump this up to `10**8`. nunique is the slowest one, and it's only ~2x slower than sorting. Could also probably bump trials up, but I'm fine leaving it at 1.

test/SpeedTestSegReduce.chpl
```diff
@@ -0,0 +1,184 @@
+use TestBase;
```
ronawho, line 1: Just curious -- do you want to keep this test as-is for future tuning, or do you think it's served it's purpose with the initial experimentation?

ronawho:
Here's 16-node-cs-hdr performance results with the problem size bumped to `10**8`: https://ronawho.github.io/ak-669/?startdate=2021/02/20&enddate=2021/02/23&graphs=groupedaggregationperformancetier1,groupedaggregationperformancetier2,groupedaggregationperformancetier3

My thought would be to bump `or`/`and` to tier 1 or 2, and drop `any`/`all` to tier 3. 

`nunique` is the only slower one anymore. It's ~2-3x slower than sorting, which is still pretty fast. I don't have any ideas off the top of my head for speeding up unique.

benchmarks/aggregate.py
reuster986: I agree.

test/SpeedTestSegReduce.chpl
reuster986: I think it's probably served its purpose. I'll move it to untested so it doesn't slow down the CI.

reuster986:
Nice! I'll put `or`/`and` in Tier 1 and `any`/`all` in Tier 3.

`nunique` is essentially a coargsort of a group ID with the values to be uniqued, so it makes sense that it would be 2-3x slower than a sort. The only thing I can think of would be to modify it to take advantage of the integer bit-packing that you implemented for coargsort, but I don't think it's urgent.

benchmarks/graph_infra/arkouda.graph
```diff
@@ -22,6 +22,24 @@ files: groupby.dat, groupby.dat, groupby.dat, groupby.dat
 graphtitle: Groupby Performance
 ylabel: Performance (GiB/S)
```
ronawho, line 23: Not from this PR, but just noticing this is `GiB/S` instead of `GiB/s` (uppercase S)

benchmarks/graph_infra/arkouda.graph
ronawho: Gotcha. I like the idea of separating them like this, but wonder if there's a term that makes it a little more obvious that it's just speed tiers? In some of the chapel perf graphs we have 2 tiers labeled `(faster idioms)` and `(slower idioms)`. Maybe fastest/faster/slower here? I don't feel very strongly about this, so I'm fine leaving it as-is too.

ronawho (approved):
👍

PR #671, opened 2021-02-15
Optimize string argsort

ronawho:
Skip checking if strings are already sorted and use more aggregation to
optimize string argsort

`SegString.isSorted()` involves a lot of fine-grained remote array
accesses, which doesn't perform well. Skip that check since it's faster
to just sort directly. We can probably optimize this check later, but
for now it was easier to just skip it.

Use aggregation to gather values in string sorting. Elements from the
`values` array can be remote, so using aggregation avoids lots of
fine-grained remote accesses. To do that replace `unorderedCopy` for the
first `copyDigit` with assignment for local operations and aggregation
for the potentially remote `values`. Increase aggregation usage for the
second `copyDigit` by aggregating the remote `values` as well. Note that
this requires calculating and storing the bucket position prior to
gathering the new values/digits.

Perf results for str-argsort on 16-node-xc and 16-node-cs-hdr. First 3
use the current default problem size (10**6), but with this we can start
running larger problem sizes as shown in the last 2 rows:

| config            | 16-node-xc   | 16-node-cs-hdr |
| ----------------- | ------------ | -------------- |
| master            | 0.0086 GiB/s | 0.0027 GiB/s   |
| no checkSorted    | 0.0414 GiB/s | 0.0035 GiB/s   |
| copyDigit agg     | 0.1137 GiB/s | 0.0118 GiB/s   |
| --size=$((10**8)) | 0.8214 GiB/s | 0.7413 GiB/s   |
| --size=$((2**29)) | 0.8543 GiB/s | 1.6285 GiB/s   |

Part of https://github.com/mhmerrill/arkouda/issues/659

### DIFF
--- a/src/SegStringSort.chpl
+++ b/src/SegStringSort.chpl
@@ -184,49 +184,46 @@ module SegStringSort {
   proc radixSortLSD_raw(const ref offsets: [?aD] int, const ref lengths: [aD] int, const ref values: [] uint(8), const ref inds: [aD] int, const pivot: int): [aD] int throws {
     const numBuckets = 2**16;
     type state = (uint(8), uint(8), int, int, int);
-    inline proc copyDigit(ref k: state, const off: int, const len: int, const rank: int, const right: int) {
-      // TODO can we only use the aggregated version?
-      use UnorderedCopy;
+    inline proc copyDigit(ref k: state, const off: int, const len: int, const rank: int, const right: int, ref agg) {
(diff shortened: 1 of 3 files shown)


reuster986 (approved):
This is great! I must have written this code before I understood how aggregators work, because I'm looking at the old code and wondering what I was thinking. Anyway, this looks a lot better!

PR #672, opened 2021-02-16
Remove some old compatibility shims

ronawho:
Arkouda requires Chapel 1.22 or newer nowadays, so remove some
compatibility shims for older versions of Chapel.

Remove `unorderedCopyWrapper`, which was a compatibility shim to have
code work with 1.20 and 1.21.

Remove some tuple indexing tricks that were only needed when supporting
1.21 and 1.22 concurrently. This reverts 88d1f44.


mhmerrill (approved):
These changes look ok to me.

PR #673, opened 2021-02-16
Remove some dead code that uses unorderedCopy

ronawho:
In looking to get rid of more uses of `unorderedCopy` I noticed
`getHeads()` is unused, so remove it.

This also removes commented out counting sort variants that used
`unorderedCopy()`.  Some of these were always experimental, and the
ones that weren't used were replaced with RadixSortLSD in 7823d4a.

Related to https://github.com/mhmerrill/arkouda/issues/222


mhmerrill (approved):
These changes look ok to me.

Issue #675, opened 2021-02-16
Casting to str is slow


ronawho:
From the gather.py benchmark, I'm seeing the casting a `numeric` array to a `str` array is slow (much slower than the gather itself):

From a variant of 16-node-cs-hdr with cast timings added to gather.py:

| size  | cast       | str-gather  |
| ----- | ---------: | ----------: |
| 10**5 |   0.36 sec | 0.32 sec    |
| 10**6 |   2.59 sec | 0.38 sec    |
| 10**7 |  28.68 sec | 0.90 sec    |
| 10**8 | 366.05 sec | 5.90 sec    |

ronawho:
https://github.com/mhmerrill/arkouda/pull/676 changed the str-gather benchmark to default to `--randomize=true` to work around this, but it is still an issue. Not sure how often casts to `str` get used in practice though.

ronawho:
I believe the slow routine is `castGenSymEntryToString`:

https://github.com/Bears-R-Us/arkouda/blob/49832db6b97b013d9515adac3647f3c9649b9841/src/CastMsg.chpl#L145-L186


Creating the array-of-strings will result in a lot of concurrent allocations and casting the int to a string probably does as well. We've seen high overhead from lots of short-lived allocations in other work. For those cases we were able to optimize out the allocations, but that is harder to do here. We use `jemalloc` to allocate memory and it's highly optimized for concurrent allocation, so I kept being surprised by the magnitude of this performance hit until I realized it's `--memTrack` that is serializing. https://github.com/Bears-R-Us/arkouda/issues/929 captures this. 

Here's performance for a cast microbenchmark w/ and w/o `--memTrack` on 1 node with 128-core Rome:

```python
#!/usr/bin/env python3

import arkouda as ak
import time, sys

ak.verbose = False
ak.connect(server=sys.argv[1], port=sys.argv[2])
N = int(sys.argv[3]) if len(sys.argv) > 3 else 10**8

A = ak.arange(N)
start = time.time()
v = ak.cast(A, 'str')
print('cast {:.1f}'.format(time.time() - start))
```

| config       | Time   |
| ------------ | -----: |
| w/o memTrack | 311.7s |
| w/ memTrack  |   2.4s |

So resolving https://github.com/Bears-R-Us/arkouda/issues/929 will significantly improve the performance of casting to a str, and I suspect that will be fast enough to close this issue.

ronawho:
[...]

Issue #659, opened 2021-02-05
String sort performance seems low


ronawho:
There are new string benchmarks for groupby, gather, argsort, and coargsort. groupy and gather performance isn't terrible, but argsort looks pretty bad. From today's [16-node-cs-hdr](https://chapel-lang.org/perf/arkouda/16-node-cs-hdr/?startdate=2021/02/04&enddate=2021/02/06&graphs=argsortperformance,coargsortperformance,groupbyperformance,gatherperformance,stringargsortperformance,stringcoargsortperformance,stringgroupbyperformance,stringgatherperformance) runs:

| benchmark       | numeric     | string      |
| --------------- | ----------: | ----------: |
| 2-arr goupby    |  1.49 GiB/s | 0.122 GiB/s |
| gather          | 43.85 GiB/s | 4.310 GiB/s |
| argsort         |  4.69 GiB/s | 0.002 GiB/s |
| 2-arr coargsort |  2.35 GiB/s | 0.005 GiB/s |

So string goupby/gather are ~10x slower, but string sorting is ~2000x slower.

hokiegeek2:
@ronawho is this with the latest from master?

ronawho:
Yeah, this is from the nightly runs.

reuster986:
Yeah, that does seem fishy. I would expect string sorting to be slower than string groupby (because the former operates on the actual, variable-length strings while the latter uses fixed-length hashes), but a net 200x is quite striking.

Also, it is a bit surprising that strings are 10x slower than ints for groupby/gather. I would have expected about 3-4x, but I don't have an easy way to test on infiniband.

Alas, nothing comes readily to mind as a culprit for these issues. I mean, other than me :)

ronawho:
> Yeah, that does seem fishy. I would expect string sorting to be slower than string groupby (because the former operates on the actual, variable-length strings while the latter uses fixed-length hashes), but a net 200x is quite striking.

Oh, that's interesting. For some reason I thought string sorting was operating on the hashes too, but I've never looked at this code in detail. I plan on digging into this over the next week or 2. 

This is probably the code I'm remembering (from when I was optimizing numeric coargsort), but it might not be what's actually called:

https://github.com/mhmerrill/arkouda/blob/4a3855fd940160355880a5194736500fb896d982/src/ArgSortMsg.chpl#L432-L445

[...]

PR #676, opened 2021-02-16
Increase the problem size for string benchmarks

ronawho:
Bump the problem size for string benchmarks to 10**8, which matches the
numeric benchmarks. Previously, we used smaller problem sizes for string
benchmarks because they were slower. Regular argsort was optimized in
#671 and coargsort wasn't actually that slow so we can bump sort problem
size for them. str-gather itself is fast, but the cast was slow (see
#675) so switch to using random data to avoid that issue. The old goupby
problem size wasn't bad, but bump it to be consistent with other codes.
This also makes the string size consistent across benchmarks by
increasing the max length to 16 for argsort/coargsort.

Here's the rates with the old and new problem sizes on 16-node-cs-hdr:

| benchmark       | before      | after       |
| --------------- | ----------: | ----------: |
| 2-arr goupby    |  0.12 GiB/s |  0.64 GiB/s |
| gather          |  5.42 GiB/s | 12.54 GiB/s |
| argsort         |  0.01 GiB/s |  0.86 GiB/s |
| 2-arr coargsort |  0.01 GiB/s |  0.51 GiB/s |

For some of these benchmarks (especially sorting) there is a non-trivial
fixed overhead regardless of problem size. So the actual time to run
some of these hasn't changed drastically because most of the time was
spent in that initial set up, not doing the core operation. Here's the
times to demonstrate that:

| benchmark       | before      | after       |
| --------------- | ----------: | ----------: |
| 2-arr goupby    | 20.96 sec   | 39.13 sec   |
| gather          |  1.37 sec   |  5.94 sec   |
| argsort         |  6.28 sec   | 15.52 sec   |
| 2-arr coargsort | 13.18 sec   | 26.14 sec   |

Part of #659


ronawho:
@reuster986 I'm also wondering if we should unify the string lengths across these benchmarks. argsort/coargsort use `random_strings_uniform(1, 8..)` and gather/groupby use `random_strings_uniform(1, 16...)`

reuster986 (approved):

reuster986:
This will be great to bring the string benchmarks in line with the numeric ones. 

Regarding unifying string lengths, I'm thinking we should go with `random_strings_uniform(1, 16, ...)` across the board, so that average string length will be 8 and will be about the size of an equivalent numeric array (or two numeric arrays, if you consider the offsets).

reuster986:
@ronawho I just merged @hokiegeek2 's fix for mypy, so a rebase should do the trick.

ronawho:
Rebased and updated argsort/coargsort to use `random_strings_uniform(1, 16, ...)` 

Rates/times in PR description updated to reflect the new length (the rate is slightly higher)

PR #681, opened 2021-02-19
Skip long string partitioning in string argsort

### LINKED ISSUE #680: Single node str-argsort is slow
The chapcs configuration (single node `CHPL_COMM=none`) is timing out on str-argsort with the recent problem size increase. It looks like we optimized multi-locale string sorting, but single locale is still slow. Looking at the breakdown, I see that almost all the time is spent gathering and sorting long strings:

```
make test-bin/StringArgsortSpeedTest
./test-bin/StringArgsortSpeedTest --maxLen=16 --size=$((10**6*2)) --v

> twoPhaseStringSort Line 73 DEBUG [Chapel] Gathered long strings in 4.771183e+00 seconds
> twoPhaseStringSort Line 80 DEBUG [Chapel] Sorted long strings in 2.226543e+01 seconds
> ...
> twoPhaseStringSort Line 93 DEBUG [Chapel] Sorted ranks in 2.614360e-01 seconds
> Sorted 2000000 strings (17.0 MB) in 28.25 seconds (0.60 MB/s)
```

@reuster986 I'm not very familiar with this code yet, but it looks like it takes "long" strings and converts them to chapel strings and then calls the regular chapel sort on them. Do you remember the context/motivation for this?

ronawho:
String argsort first partitions long strings and sorts them separately.
The core string radix sort then has an iteration for every digit of the
longest short string, so the idea is that if you just have a few long
strings you sort them separately and you can minimize the total number
of iterations in the core radix sort. This is still a good idea, but the
current implementation has pretty high overhead, especially for single
node, so skip it by default. Note that prior to optimizing the core
radix sort in #671 the separate sorting was still a win for performance,
but after that the core radix sort is much faster and we'll have to do
more tuning for the separate long string sorting.

I expect to revisit this in the future since the idea is still good, but
I think we need to improve the heuristics for what to consider a long
string and look at optimizing the sort that's used for long strings.

This improves single node str-argsort. There's no difference for the
multi-locale str-argsort since it happened to set up the pivot such that
everything was considered short anyways. Here's some chapcs comm=none
results before and after this change:

```
make test-bin/StringArgsortSpeedTest
```

| size  | before    | after      |
| ----- | --------: | ---------: |
| 10**6 | 0.6 MB/s  |  81.0 MB/s |
| 10**8 | timed out | 128.9 MB/s |

Resolves #680

### DIFF
--- a/src/SegStringSort.chpl
+++ b/src/SegStringSort.chpl
@@ -18,7 +18,9 @@ module SegStringSort {
   private const MINBYTES = SSS_MINBYTES;
   private config const SSS_MEMFACTOR = 5;
   private const MEMFACTOR = SSS_MEMFACTOR;
-  
+  private config const SSS_PARTITION_LONG_STRING = false;
+  private const PARTITION_LONG_STRING = SSS_PARTITION_LONG_STRING;
+ 
   
   const ssLogger = new Logger();
   if v {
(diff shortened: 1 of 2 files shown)


ronawho:
Probably easiest to review ignoring whitespace.

reuster986 (approved):
I agree with the reasoning here, and the performance is looking great!

PR #684, opened 2021-02-23
340_enable_runtime_error-checking_with_enums#340

hokiegeek2:
This draft PR demonstrates how to add more Python enums for Arkouda constants mgmt.

### DIFF
--- a/src/arkouda_server.chpl
+++ b/src/arkouda_server.chpl
@@ -46,6 +46,31 @@ proc initArkoudaDirectory() {
 }
 
 proc main() {
+
+    proc printServerSpashMessage() throws {
+        var arkDirectory = initArkoudaDirectory();
+        var version = "arkouda server version = %s".format(arkoudaVersion);
+        var directory = "initialized the .arkouda directory %s".format(arkDirectory);
+        var memLimit =  "getMemLimit() %i".format(getMemLimit());
+        var memUsed = "bytes of memoryUsed() = %i".format(memoryUsed());
+        var serverMessage: string;
+        var serverToken: string;
+    
+        if authenticate {
+            serverToken = getArkoudaToken('%s%s%s'.format(arkDirectory, pathSep, 'tokens.txt'));
+            serverMessage = ">>>>>>>>>>>>>>> server listening on tcp://%s:%t?token=%s " +
+                        "<<<<<<<<<<<<<<<".format(serverHostname, ServerPort, serverToken);
+        } else {
+            serverMessage = ">>>>>>>>>>>>>>> server listening on tcp://%s:%t <<<<<<<<<<<<<<<".format(
+                                        serverHostname, ServerPort);
+        }
+        
+        var smPadding = 50 - serverMessage.size;
+        writeln('***************************************************');
+        
+        writeln('***************************************************');
+    }
+
     asLogger.info(getModuleName(), getRoutineName(), getLineNumber(),
                                                "arkouda server version = %s".format(arkoudaVersion));
     asLogger.info(getModuleName(), getRoutineName(), getLineNumber(),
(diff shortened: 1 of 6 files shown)


reuster986 (commented):
This is interesting -- would this change anything for the user, or is this for developers? For example, can the user still say `mydtype = ak.int64`, or do they have to do something like `mydtype = ak.DTypeObject.INT64.value`?

hokiegeek2:
@reuster986 This is primarily for developers, kinda like what I did for MessageType and MessageFormat. The user would still use ak.int64. 

But, this reminds, me I need to add a str() method override like I did for MessageType/MessageFormat:

image

This way it gets passed in by the developer as the enum ak.DTypeObject.INT64 and the desired str value gets generated with the request string, analogous to the code sample below:

image

mhmerrill (approved):
this looks ok to me.

src/arkouda_server.chpl
```diff
@@ -81,13 +106,23 @@ proc main() {
     }
 
     socket.bind("tcp://*:%t".format(ServerPort));
-
-    const boundary = "**************************************************************************" +
-                   "**************************";
-
-    asLogger.info(getModuleName(), getRoutineName(), getLineNumber(), boundary);
-    asLogger.info(getModuleName(), getRoutineName(), getLineNumber(), serverMessage);
```
ronawho, line 89: @hokiegeek2 I think with this change the above "server listening on..." message is no longer printed. (Noticed while trying to do some manual server runs and not seeing the host/port printed anymore)

src/arkouda_server.chpl
```diff
@@ -46,6 +46,31 @@ proc initArkoudaDirectory() {
 }
 
 proc main() {
+
+    proc printServerSpashMessage() throws {
```
ronawho, line 50: Maybe related to https://github.com/mhmerrill/arkouda/pull/684/files#r600548315, but I don't think this is ever called right now.

src/arkouda_server.chpl
hokiegeek2: @ronawho not sure what you mean; this is from a fresh build of the main branch:

image

The "server listening on" message is there.

src/arkouda_server.chpl
hokiegeek2: @ronawho Yes, that fuhction is not called now, it is a placeholder for a PR I am submitting later today

src/arkouda_server.chpl
ronawho: Huh, ok. So I see what you see on my local machine, but on a new machine I'm running on I just see:

```
$ ./arkouda_server -nl 1
2021-03-24:15:10:08 [arkouda_server] main Line 64 INFO [Chapel] arkouda server version = 2020.07.07
2021-03-24:15:10:08 [arkouda_server] main Line 66 INFO [Chapel] memory tracking = true
2021-03-24:15:10:08 [arkouda_server] main Line 69 INFO [Chapel] initialized the .arkouda directory /home/eronaghan/arkouda-sort-subtimers/.arkouda
2021-03-24:15:10:08 [arkouda_server] main Line 73 INFO [Chapel] getMemLimit() 1947559340851
2021-03-24:15:10:08 [arkouda_server] main Line 75 INFO [Chapel] bytes of memoryUsed() = 39472
```

And then I don't get anything else until I connect a client to it. It looks like the `socket.bind` call is blocking until a client connects on this system. Ok, so not a new problem and the behavior didn't change with this PR. Sorry for the noise.

reuster986: Actually, some of our internal users are seeing the same thing with recent builds of arkouda, whereas a build from 2/26 works fine on the same system. It's a pretty big problem on a multi-node system, because the user doesn't know the hostname to connect to. If the problem is a blocking `socket.bind`, can we just print out the connect string before calling `bind`?

hokiegeek2: @reuster986 Okay, I will put a logging message w/ connect info back in right before bind and will do so in #715

hokiegeek2: @reuster986 @ronawho put print message including connect string before socket.bind. @ronawho please pull down my latest [PR](https://github.com/mhmerrill/arkouda/pull/716) and see if the print message is seen on startup. Sorry, I cannot replicate this on my mac, so I did not see the issue you guys are reporting.

hokiegeek2: @reuster986 @ronawho Also added stdout.flush() at @mhmerrill suggestion. I have that in the logging framework I wrote but not in the printServerSplashMessage. From what I understand from my discussion with Mike, that might explain what y'all are seeing on the Crays

Issue #683, opened 2021-02-22
Type checking too restrictive on Strings ops


reuster986:
```python
s = ak.array(['hello', 'world', 'I', 'am', 'a', 'string'])
s == 'world'
s != 'world'
```

The last two lines raise a `TypeError` saying "type of argument "other" must be one of (Strings, str_); got str instead".  These operators should accept any python or numpy string type, whic I believe the code supports, but the type checking needs to be loosened or removed.

Since this has happened a couple of times, I propose that we stop type checking all non-arkouda types, including scalars like int, float, and str. We want to rely on ducktyping for types that are outside our control, so that we can interoperate with other packages that might overload the basic types (as numpy does). We can keep the type checking for pdarrays and `ak.Strings`, etc., because we control those and know how the should (not) be used.

hokiegeek2:
@reuster986 gotcha, I can update the type checking as you suggested. For the Strings __eq__ method I added the str dtype and now the use case you detailed above works:

![image](https://user-images.githubusercontent.com/10785153/108917195-5d742300-75fd-11eb-8fb0-dbdcc5b09449.png)

reuster986:
@hokiegeek2 I know we can just add `str`, but what about `np.str`? Or any other numpy string-related types? Or some other package's custom string type? Rather than try to guess them all and enumerate them in the type-checking (and have users complain when they don't work), let's just not do any type checking on scalars. If the user passes something that doesn't behave like a string, it'll throw an error anyway.

Am I missing something? Is there any advantage to type-checking scalars?

hokiegeek2:
@reuster986 Good points. The type checking enforced by typeguard was previously being done within several methods in various forms. I basically centralized and standardized the type checking within typeguard. 

The advantages of type checking w/ typeguard are that we'll (1) catch the error on the client instead of sending invalid requests to the server (2) provide the user with consistent error messages instead of a large stack trace that includes reply messages from the Arkouda server (3) enforce type checking with the ```@typeguard``` annotation instead of custom code and (4) tie in type checking with typeguard into the type hints we already have.

Having written all of that, if users are getting frustrated, that supersedes the four advantages I listed above.

hokiegeek2:
@reuster986 just added support for str, bytes, and np.str_ for the Strings methods

hokiegeek2:
@reuster986 are we good to close this issue?

mhmerrill:
@reuster986 I am going to close this, we can reopen if you so choose.

PR #686, opened 2021-03-01
683 type checking too restrictive on strings ops

hokiegeek2:
Addresses #683


mhmerrill (approved):
This looks ok to me. I'd like @reuster986 to look at it also because he put in the original issue.

reuster986 (approved):
This looks like it addresses the original issue, and I think we should merge the PR so we can move forward. I'll open a discussion on another thread about our longer term strategy for type checking.

PR #688, opened 2021-03-03
548 error handling in msg procedures and supporting classes should be consistent

hokiegeek2:
This PR covers the following:

1. Each *Msg class returns a MsgTuple, which consists of the msg to be returned and a message type, which is an enum with possible values of NORMAL, WARN, and ERROR. arkouda_server determines whether to handle message as normal or an error depending upon the message type, not by parsing the message string.
2. Updated client logic that handles the return message via the message type as opposed to parsing returned message. Note: this is for messages in string format, not binary format.
3. Replaced error string messages from supporting classes such as MultiTypeSymbolTable
4. Updated chapel and python unit tests.


mhmerrill (approved):
it looks ok to me... cross your fingers... fire-in-the-hole ;-)

PR #695, opened 2021-03-09
694 ci fails for python 3.8.8

hokiegeek2:
This PR (1) cleans up client-side binary message debug logging and (2) re-enabled client-side debug verbosity for test-python.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -42,33 +42,33 @@ jobs:
       run: |
         make doc
 
-  arkouda_python_portability:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.x']
-    container:
-      image: chapel/chapel:1.23.0
-    steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v2
-      with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        apt-get update && apt-get install -y libhdf5-dev hdf5-tools libzmq3-dev python3-pip
-        echo "\$(eval \$(call add-path,/usr/lib/x86_64-linux-gnu/hdf5/serial/))" > Makefile.paths
-    - name: Build/Install Arkouda
-      run: |
-        make
-        python3 -m pip install -e .[dev]
-    - name: Arkouda make check
-      run: |
-        make check
-    - name: Arkouda unit tests
-      run: |
-        make test-python
+  #arkouda_python_portability:
+  #  runs-on: ubuntu-latest
+  #  strategy:
+  #    matrix:
+  #      python-version: ['3.7', '3.8', '3.9', '3.x']
+  #  container:
+  #    image: chapel/chapel:1.23.0
+  #  steps:
+  #  - uses: actions/checkout@v2
+  #  - name: Set up Python ${{ matrix.python-version }}
+  #    uses: actions/setup-python@v2
+  #    with:
+  #      python-version: ${{ matrix.python-version }}
+  #  - name: Install dependencies
+  #    run: |
+  #      apt-get update && apt-get install -y libhdf5-dev hdf5-tools libzmq3-dev python3-pip
+  #      echo "\$(eval \$(call add-path,/usr/lib/x86_64-linux-gnu/hdf5/serial/))" > Makefile.paths
+  #  - name: Build/Install Arkouda
+  #    run: |
+  #      make
+  #      python3 -m pip install -e .[dev]
+  #  - name: Arkouda make check
+  #    run: |
+  #      make check
+  #  - name: Arkouda unit tests
+  #    run: |
+  #      make test-python
 
   arkouda_tests_linux:
     runs-on: ubuntu-latest
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,7 @@ print-%:
 	@echo "$($*)"
 
 test-python: 
-	python3 -m pytest $(ARKOUDA_PYTEST_OPTIONS) -c pytest.ini
+	python3 -m pytest $(ARKOUDA_PYTEST_OPTIONS) -c pytest.ini -s
 
 CLEAN_TARGETS += test-clean
 .PHONY: test-clean
--- a/Makefile
+++ b/Makefile
@@ -351,7 +351,7 @@ print-%:
 	@echo "$($*)"
 
 test-python: 
-	python3 -m pytest $(ARKOUDA_PYTEST_OPTIONS) -c pytest.ini -s
+	python3 -m pytest $(ARKOUDA_PYTEST_OPTIONS) -c pytest.ini
 
 CLEAN_TARGETS += test-clean
 .PHONY: test-clean
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -404,8 +404,6 @@ def _send_binary_message(cmd : str, payload : bytes, recv_bytes : bool=False,
     send_message = RequestMessage(user=username, token=token, cmd=cmd, 
                                 format=MessageFormat.BINARY, args=cast(str,args))
 
-    logger.debug('sending message {}'.format(send_message))
-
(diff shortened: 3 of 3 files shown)


glitch:
How did this fix the CI failures for python 3.8.8 ?  (I'm not seeing it)

hokiegeek2:
> How did this fix the CI failures for python 3.8.8 ? (I'm not seeing it)

@glitch I don't know, to be honest. As I noted in the [issue](https://github.com/mhmerrill/arkouda/issues/694), I cannot replicate this error. The only place it shows up is in the CI workflow, so it *could* be a race condition of some sort as @mhmerrill discussed on Friday. The code updates are cleanup and better debug logging by default which are good things that also happen to address the Python 3.8 CI workflow-specific error.

glitch:
RE: Test failures - Oh good, I thought I was going nuts

mhmerrill:
@hokiegeek2 does it still pass if you remove all the client debug logging?

glitch:
I've been doing a little debugging on this locally (mainly to learn) and I haven't been able to get it to fail.  Just to make sure we're on the same page my understanding is this:
 - pytest runs in a serial fashion, to run tests in parallel you have to install extensions (we haven't)
 - The tests run in sequential order (this is also further defined in the pytest.ini file)
 - The test `tests/setops_tests.py::SetOpsTest::testSetxor1d` periodically fails  on the first assertion [here](https://github.com/mhmerrill/arkouda/blob/master/tests/setops_test.py#L80)
 - The failure has only been observed in the CI python portability tests when running Python 3.8.8
 - The failure _appears_ to be due to an unknown symbol/id in the MultiTypeSymbolTable.lookup table

I spun up a local environment with Python 3.8.8 etc. and limited the running tests just to `tests/setops_test.py`.  I then added a test method to run `testSetxor1d` N number of times and have _not_ been able to get it to break locally after running the test 150 times in succession.

I agree with @hokiegeek2 that it _feels_ like it's some type of race condition but we're running single threaded right, so how could that be?  Is it possible that ZMQ is interleaving requests since it is asynchronous?  However, since we only have a single client operating at any one time (allowing for stand/up tear down between tests classes) I don't really see how the requests could get out of order when creating pdarray and performing cleanup (granted I'm not yet familiar enough with how the pdarray class and ak client operate regarding zmq and whether or not Chapel is doing some type of lazy initialization).  This is interesting.

hokiegeek2:
> I've been doing a little debugging on this locally (mainly to learn) and I haven't been able to get it to fail. Just to make sure we're on the same page my understanding is this:
> 
> * pytest runs in a serial fashion, to run tests in parallel you have to install extensions (we haven't)
> * The tests run in sequential order (this is also further defined in the pytest.ini file)
> * The test `tests/setops_tests.py::SetOpsTest::testSetxor1d` periodically fails  on the first assertion [here](https://github.com/mhmerrill/arkouda/blob/master/tests/setops_test.py#L80)
> * The failure has only been observed in the CI python portability tests when running Python 3.8.8
> * The failure _appears_ to be due to an unknown symbol/id in the MultiTypeSymbolTable.lookup table
> 
> I spun up a local environment with Python 3.8.8 etc. and limited the running tests just to `tests/setops_test.py`. I then added a test method to run `testSetxor1d` N number of times and have _not_ been able to get it to break locally after running the test 150 times in succession.
> 
> I agree with @hokiegeek2 that it _feels_ like it's some type of race condition but we're running single threaded right, so how could that be? Is it possible that ZMQ is interleaving requests since it is asynchronous? However, since we only have a single client operating at any one time (allowing for stand/up tear down between tests classes) I don't really see how the requests could get out of order when creating pdarray and performing cleanup (granted I'm not yet familiar enough with how the pdarray class and ak client operate regarding zmq and whether or not Chapel is doing some type of lazy initialization). This is interesting.

@glitch Actually, pretty sure Arkouda by default starts up n threads where n is the number of VCPUs on the host system, so I don't think Arkouda is single-threaded. That's why it's *possible* there is some type of race condition.

glitch:
@hokiegeek2 (Sorry, still in learning mode over here ;) When you say Arkouda in this case are you talking about the Server or the Client or both?  I know the server backend (chapel) can function in a multi-threaded fashion, but the client is only single threaded and operating in a serial fashion.  Is that correct?  Or does `client.py` and/or `pdarrayclass.py` have some type of threading or request interleaving baked in?  Would ZMQ be doing that?

Based on the error, it seems like something is populating/allocating a symbol name in `MultiTypeSymbolTable.chpl` and it's either
- not created before being used by the client (which shouldn't be possible since the client won't know its name until the server returns it right?  Or is there some type of a _future_ / _promise_ type thing going on here?)
- something is deleting it before the client is done using it (shouldn't be possible in this case since we have only 1 client who is running single threaded)

I'm still trying to understand the isolation of operations.

hokiegeek2:
> @hokiegeek2 (Sorry, still in learning mode over here ;) When you say Arkouda in this case are you talking about the Server or the Client or both? 

Sorry, I did not specify that I am referring to the Arkouda server (Chapel process(es)). Each Chapel locale by default has n threads where n = number of vCPUs.

I know the server backend (chapel) can function in a multi-threaded fashion, but the client is only single threaded and operating in a serial fashion. Is that correct? Or does `client.py` and/or `pdarrayclass.py` have some type of threading or request interleaving baked in? Would ZMQ be doing that?

Both are single-threaded and neither is using asyncio (for now, although I've looked into that for another feature request). Also, we're using the REQ/REP pattern in ZMQ, so the client will block until the Arkouda server (again, the Chapel process(es)) returns the reply.

> 
> Based on the error, it seems like something is populating/allocating a symbol name in `MultiTypeSymbolTable.chpl` and it's either
> 
> * not created before being used by the client (which shouldn't be possible since the client won't know its name until the server returns it right?  Or is there some type of a _future_ / _promise_ type thing going on here?)
> * something is deleting it before the client is done using it (shouldn't be possible in this case since we have only 1 client who is running single threaded)
> 
Good thoughts! Because, as I mentioned above, the client blocks until the reply is received since we're not doing asyncio, the future option is not there (again, good idea to consider that). 

I don't _think_ it's an issue of the server-side objects (arrays) failing to be created, because, as you pointed out the name (id) would not have been returned to the client.

The third scenario (deleting array before client is done with it) is possible because the ids are returned to the client and then concatenate is called:

image

> I'm still trying to understand the isolation of operations.
You're doing a great job! There's lots to learn! 💯 And...this is a really ah, "interesting" issue that we're all trying to wrap our collective heads around.

mhmerrill:
@hokiegeek2 @glitch @ronawho I was looking at the error... and the name `id_27` seems awful low unless we are restarting the server for every test, is that the case. Another reason is we are rolling over the name counter for some reason...

hokiegeek2:
> @hokiegeek2 @glitch @ronawho I was looking at the error... and the name `id_27` seems awful low unless we are restarting the server for every test, is that the case. Another reason is we are rolling over the name counter for some reason... or maybe we are not interpreting the return from the `created id_27 ...` correctly.

@mhmerrill Good catch -> a new Arkouda server instance starts up with each test class, in this case, setops_test. The server startup and shutdown invocations occur in the setUpClass and tearDownClass methods, respectively:

image

image

The setUp and tearDown methods are called at the start and end of each test case within the test class, and that is where the server connection to the test class-scoped server is opened and closed:

image

So the server-side Arkouda objects build up over the course of all the test cases within a class, and then they are destroyed when the server is shutdown after all test cases within a test class have been executed.

glitch:
On a related note, I was looking into PyTest fixtures to compare it to UnitTest.  For reference it looks like `pytest` is being used as the test runner, but `unittest` is being used for fixture creation and test assertions.  Pytest is pretty good interoperability with `unittest` on a number of things.

Regarding fixture scopes `pytest` has 5 levels (session, package, module, class, function) and `unittest` has 3 (module, class, function).  Our tests aren't organized into modules or packages so were only using class and function isolation.  If we wanted to change the scope (i.e. is an arkouda-server instance used between classes or not) the good news is that both frameworks are capable of doing it, I think it would be more a matter of organizing the test files.

mhmerrill (approved):
I am making the decision to remove the Python compatibility tests until we resolve the issue.
I am approving this PR, Mike.

PR #697, opened 2021-03-10
690 add seg string factory methods

hokiegeek2:
This PR contains code changes that move throwing functions from the SegString init method to obviate the need to include a try! invocation; this refactoring removes the possibility of crashing the Arkouda server during instantiation of SegString objects.


mhmerrill (approved):
this looks god to me.

Issue #694, opened 2021-03-05
CI fails for Python 3.8.8


mhmerrill:
CI Failed in the same place 3 times in a row using Python 3.8.8
Python 3.7 and 3.9 pass, weird...
Here is a link to the failed CI run.
https://github.com/mhmerrill/arkouda/runs/2042148592?check_suite_focus=true

hokiegeek2:
Tested within the chapel 1.23-ubuntu docker image with a Python 3.8.5 virtualenv and the setops_test cases are passing:

![image](https://user-images.githubusercontent.com/10785153/110330157-73191d80-7feb-11eb-9381-b37d33beedde.png)

hokiegeek2:
@mhmerrill it appears there's some kind of race condition as the 3.8 test is now failing again once I take the verbose pytest logging out:

![image](https://user-images.githubusercontent.com/10785153/110619153-759f8280-8165-11eb-8594-cb08cb7e0f01.png)

Please let me know how you'd like for me to proceed.

hokiegeek2:
Switching to INFO-level logging with pytest -s also enables all compat tests to pass:

![image](https://user-images.githubusercontent.com/10785153/110678382-40feeb80-81a4-11eb-8f5a-6a4ee2b002c7.png)

hokiegeek2:
Turning off client logging by running pytest (removing -s argument) and adding calls to repr on both input pdarrays also enabled the 3. 8 tests to pass:

![image](https://user-images.githubusercontent.com/10785153/110681104-6e996400-81a7-11eb-8e0b-63bc5c808ca0.png)

![image](https://user-images.githubusercontent.com/10785153/110681228-9688c780-81a7-11eb-905f-ccbeb1448f92.png)

hokiegeek2:
@mhmerrill This change to setops_test addresses what appears to be a race condition specific to the 3.8 compatibility test as demonstrated in the [action](https://github.com/mhmerrill/arkouda/actions/runs/640468305) triggered in the action.

![image](https://user-images.githubusercontent.com/10785153/110689423-a9ec6080-81b0-11eb-9e57-ede285aa7a02.png)

![image](https://user-images.githubusercontent.com/10785153/110689583-d902d200-81b0-11eb-9c53-387017f0898f.png)

mhmerrill:
[...]

PR #700, opened 2021-03-12
691 restore support for categorical == str

hokiegeek2:
This PR does the following:

1. enables Categorical == and != via the Categorical._binops, __eq__, and __ne__ methods 
2. enables for both scalars and Categoricals
3. enables == and != operations between different Categoricals objects
4. adds corresponding unit tests 
5. expands categorical_test coverage
6. includes fix for #694


mhmerrill (approved):
looks good to me

PR #701, opened 2021-03-12
699 categorical.concatenate fails

hokiegeek2:


tests/categorical_test.py
```diff
@@ -106,4 +106,17 @@ def testBinop(self):
...
         with self.assertRaises(NotImplementedError):
             cat._binop('string 1', '===')
-        
\ No newline at end of file
+            
+    def testConcatenate(self):
+        catOne = self._getCategorical()
+        catTwo = self._getRandomizedCategorical()
+        
+        resultCat = catOne.concatenate([catTwo])
```
reuster986, line 114: I'd like to make sure we also test the preferred idiom of `resultCat = ak.concatenate([catOne, catTwo])`, either by changing this line or adding a separate test (even better).

I *think* `ak.concatenate(..., ordered=False)` should also work with Categoricals, but we should explicitly test that one too.

arkouda/categorical.py
```diff
@@ -477,7 +477,7 @@ def concatenate(self, others : List[Categorical], ordered : bool=True) -> Catego
             idxsizes = np.array([self.categories.size] + \
                                 [o.categories.size for o in others])
             idxoffsets = np.cumsum(idxsizes) - idxsizes
-            oldvals = concatenate([c.codes + off for c, off in \
+            oldvals = concatenate([c + off for c, off in \
```
reuster986, line 480: nice catch!

reuster986 (changes requested):
Thanks for fixing this! Just one request for an additional test (see inline comment)

tests/categorical_test.py
hokiegeek2: @reuster986 Done! Just pushed the commit including the two tests you requested.

reuster986 (commented):
Perfect! Thanks!

tests/categorical_test.py
hokiegeek2: @reuster986 Please review when you get a chance and lemme know if you have any other changes you'd like to see.

reuster986 (approved):

reuster986:
Sorry, I thought I had already approved but must not have clicked the right thing!

PR #705, opened 2021-03-16
703 define arkouda scalars to simplify method signatures

hokiegeek2:


mhmerrill (approved):
looks ok to me but I want @reuster986 to weigh in with his opinion.

reuster986 (changes requested):
Yay! This is a huge improvement for both clarity and maintainability, so thank you!

Just one request: I and other users use `ak.bool` a fair amount, so can we please leave that as an alias to `np.bool` for backwards compatibility?

hokiegeek2:
> Yay! This is a huge improvement for both clarity and maintainability, so thank you!
> 
> Just one request: I and other users use `ak.bool` a fair amount, so can we please leave that as an alias to `np.bool` for backwards compatibility?

Okay, I will make that change

hokiegeek2:
@reuster986 requested changes made, please review
@mhmerrill I added a couple of dtypes unit tests

reuster986 (approved):
looks great!

PR #711, opened 2021-03-18
Closes #665 : Updates GroupBy example documentation.

### LINKED ISSUE #665: Clarify GroupBy example in documentation
A user pointed out to me that [this example](https://arkouda.readthedocs.io/en/latest/usage/groupby.html) in the docs could be greatly improved by indicating that `dayOfWeek` is not a column name, but the actual column itself:

```python
byDayOfWeek = ak.GroupBy(data['dayOfWeek'])
```

glitch:
Updating Documentation for GroupBy example.  I also added a small note since I didn't want people thinking they could pass the contents of a pandas dataframe column or a list straight out of a python dictionary etc.

### DIFF
--- a/pydoc/usage/groupby.rst
+++ b/pydoc/usage/groupby.rst
@@ -10,7 +10,9 @@ For example, imagine a dataset with two columns, ``userID`` and ``dayOfWeek``. T
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
this looks ok.

Issue #674, opened 2021-02-16
Check benchmark rate calculation


ronawho:
For argsort we try to report the sort rate (results per second) and not how much data is being transferred under the covers. See https://github.com/mhmerrill/arkouda/issues/531 for discussion on that.

In some other benchmarks I'm wondering if we're mistakenly reporting the amount being transferred instead.

- Numeric gather/scatter:
  - We report `3 * iv.size * iv.itemsize`. Why * 3?
- String gather:
  - Rate includes offsets too, which I think is wrong?
- String groupby:
  - Rate also includes offsets
- At a glance the other benchmarks look right to me

ronawho:
@mhmerrill @reuster986 I feel somewhat confident we should remove offsets from the string gather/groupby calculations. I'm less sure about why gather/scatter has the ` * 3` multiplier. I would not be surprised if that's right, but I couldn't explain why myself.

mhmerrill:
@ronawho I think the gather metric is ok because we just use a constant. The thought behind the sort is we wanted to be algorithm agnostic, so different algorithms could be used and we could still compare them base on something like keys per sec.

reuster986:
I concur with @mhmerrill . I vaguely remember talking with him about a  year ago about two classes of benchmarks: a lower level designed to illuminate hardware performance and a higher level designed to measure algorithm performance from a user perspective. I think at the time we put (gather, scatter, scan, reduce, stream) in the lower level and argsort in the higher level. So I think that's why the lower level benchmarks measure rates in amount of actual data transferred while argsort measures rate in terms of elements visible to the user. Most benchmarks we've added since then are upper-level (coargsort, groupby, aggregate).

So based on that principle, I would say we should keep the `3 *` in gather/scatter and include offsets in string gather, but we should ignore offsets in string groupby. I realize I was the one who included offsets in string groupby -- I wasn't really thinking about this distinction when I wrote it, but this discussion is helping me clarify my thoughts.

ronawho:
> So based on that principle, I would say we should keep the 3 * in gather/scatter and include offsets in string gather, but we should ignore offsets in string groupby. I realize I was the one who included offsets in string groupby -- I wasn't really thinking about this distinction when I wrote it, but this discussion is helping me clarify my thoughts.

[...]

PR #712, opened 2021-03-19
fix groupby benchmark bug

reuster986:
Indentation error was causing extra output with incorrect reported rates when more than one trial was requested.

### DIFF
--- a/benchmarks/groupby.py
+++ b/benchmarks/groupby.py
@@ -39,10 +39,10 @@ def time_ak_groupby(N_per_locale, trials, dtype, seed):
             g = ak.GroupBy(arrays)
             end = time.time()
             timings.append(end - start)
-            tavg = sum(timings) / trials
-            print("{}-array Average time = {:.4f} sec".format(numArrays, tavg))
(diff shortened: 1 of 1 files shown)


ronawho (approved):
Whoops, good catch. The graphing infrastructure picks out the first reported value, so this may slightly improve or at least stabilize rates since we're actually grabbing the average now instead of the first value.

Only somewhat related, but since this may impact reported rates I wonder if it makes sense to stop including the str offsets while you're making changes (https://github.com/mhmerrill/arkouda/issues/674)

ronawho:
> Whoops, good catch. The graphing infrastructure picks out the first reported value, so this may slightly improve or at least stabilize rates since we're actually grabbing the average now instead of the first value.

Oh, actually nevermind this -- I forgot we're only doing 1 trial by default for groupby so this won't change nightly rates.

reuster986:
> Only somewhat related, but since this may impact reported rates I wonder if it makes sense to stop including the str offsets while you're making changes (#674)

Good call. Fixed.

Yeah, I went through the same thought progression about the plots :)

PR #713, opened 2021-03-20
Adjust uses of functions in the Memory module for 1.24

dlongnecke-cray:
The contents of the `Memory` module were deprecated in 1.24, and
copies were made in a submodule called `Memory.Diagnostics`.

Add some wrapper functions for `locale.physicalMemory()` and
`memoryUsed()` to the `ServerConfig` module. Due to how import and
use statements are currently resolved, these wrappers use a
compatibility module with contents that differ depending on whether
Chapel's version is < or >= 1.24.

Adjust the Makefile to include the correct version of the module
`MemDiagnostics` based on the Chapel minor version. Do the same
for the `COMPOPTS` in `$ARKOUDA_HOME/test`.

To remove the compatibility module, Arkouda's minimum required
version should be bumped to 1.24 -or- the Chapel issue at
https://github.com/chapel-lang/chapel/issues/17438 should be
resolved.

TESTING:

- [x] Passed all tests using `make test` on 1.24
- [x] Compiled server on 1.23

Unfortunately I'm having trouble getting `start_test` to work on a
1.23 release, but I'm hoping that the CI checks are proof enough
that things seem to be working for 1.23.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,6 @@ check-deps: $(CHECK_DEPS)
 CHPL_MINOR := $(shell $(CHPL) --version | sed -n "s/chpl version 1\.\([0-9]*\).*/\1/p")
 CHPL_VERSION_OK := $(shell test $(CHPL_MINOR) -ge 22 && echo yes)
 CHPL_VERSION_WARN := $(shell test $(CHPL_MINOR) -le 22 && echo yes)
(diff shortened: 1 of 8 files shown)


mhmerrill (approved):
Looks ok to me.

Issue #714, opened 2021-03-24
Authenticate flag throws exception/error


glitch:
When running the following, an exception/error is thrown, because the string concatenation & formatting is done in the wrong order.

Steps to reproduce, run
```
./arkouda_server -nl 1 --authenticate

uncaught SystemError: Invalid argument: Too many arguments for format string (in string.format)
  /home/foo/projects/arkouda//src/arkouda_server.chpl:92: thrown here
  /home/foo/projects/arkouda//src/arkouda_server.chpl:92: uncaught here
```

The offending line is [here](https://github.com/mhmerrill/arkouda/blob/591d26ea71ee54f3006673c8f5cfcd045b7e0c02/src/arkouda_server.chpl#L91)
```
        serverMessage = ">>>>>>>>>>>>>>> server listening on tcp://%s:%t?token=%s " +
                        "<<<<<<<<<<<<<<<".format(serverHostname, ServerPort, serverToken);
```
The formatting is done on the second string which has no substitution args; those live in the first string.  It looks like it was a simple formatting mistake from commit:300c142417d74e85e95178a9a5bbc3867e39ee47 which added some extra output to that string.

Note: It's a pretty trivial patch, but I want to add a unit test for it.

hokiegeek2:
@glitch Yeesh, good catch. That's my fault. Actually, if you don't mind, please hold off. I updated the startup message to fix the formatting and I'll fold a fix for this into my PR

glitch:
@hokiegeek2 Cool, no problem.

glitch:
This was what I was going to do:
```
serverMessage = ">>>>>>>>>>>>>>> " +
                            "server listening on tcp://%s:%t?token=%s ".format(serverHostname, ServerPort, serverToken) +
                            "<<<<<<<<<<<<<<<";
```
;)

hokiegeek2:
@glitch Yup, that will fix it. What won't get fixed is the crappy formatting I added, LOL. My PR will fix that. 🥇

glitch:
@hokiegeek2 While you're in there, `printServerSpashMessage` is missing an `l`, it should be "Splash".

hokiegeek2:
@glitch Actually, go for it. You found it, you should fix it. It may be my formatting changes will be frowned upon, LOL

glitch:
[...]

Issue #715, opened 2021-03-24
enhanced client load message and server startup message


hokiegeek2:
I updated the client and server to show the following on load (client) and startup (server):

client: 

![image](https://user-images.githubusercontent.com/10785153/112336699-af0edc80-8c93-11eb-9903-ca12bf942cce.png)

server (no authentication):

![image](https://user-images.githubusercontent.com/10785153/112336898-d82f6d00-8c93-11eb-97b0-5287a8b21bf3.png)

server (authentication):

![image](https://user-images.githubusercontent.com/10785153/112337001-ec736a00-8c93-11eb-9a45-49531f9cbfec.png)

hokiegeek2:
[PR](https://github.com/mhmerrill/arkouda/pull/716), issue resolved

PR #717, opened 2021-03-24
Issue #714: Fixes error with out-of-order string concat & format.

glitch:
This fixes Issue #714  where launching arkouda_server with the --authenticate flag was throwing an error due to an order of operations issue with concatenate and format.

Alternatively this can be solved via Issue #715 (with [PR#716](https://github.com/mhmerrill/arkouda/pull/716)

Why no test?
-------------------
In starting an integration/unit test for this over in python it became apparent that we don't include the token information in the server info file that gets written out (there isn't really a way to get it over to the client other than screen scraping).  Since I didn't have access to the token in the test, I couldn't shut the server down.  Rather than try and solve that here, I'm going to address it in a separate ticket because it's a bit of a bigger change than this simple bugfix.

### DIFF
--- a/src/arkouda_server.chpl
+++ b/src/arkouda_server.chpl
@@ -37,7 +37,7 @@ proc initArkoudaDirectory() {
 
 proc main() {
 
-    proc printServerSpashMessage() throws {
+    proc printServerSplashMessage() throws {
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
Looks good to me.

Issue #710, opened 2021-03-18
Categorical.concatenate fails with certain segments array lengths


hokiegeek2:
As noted in [706](https://github.com/mhmerrill/arkouda/issues/706), concatenate appears to fail if the pdarray length < numLocales.

hokiegeek2:
Example:

![image](https://user-images.githubusercontent.com/10785153/111630370-ae280780-87c8-11eb-9565-bae005ab9c2f.png)

![image](https://user-images.githubusercontent.com/10785153/111628203-56889c80-87c6-11eb-8d6a-556280120ba5.png)

hokiegeek2:
Interesting -> when I concatenate dupes of the two arrays listed above, the error does not occur:

![image](https://user-images.githubusercontent.com/10785153/111634947-6061ce00-87cd-11eb-8648-d3f4b3c111a1.png)

So it appears that the difference in the bytes pdarray lengths in the Strings object is at least part of the issue.

hokiegeek2:
It is possible the issue is due to differences in segments arrays (?):

![image](https://user-images.githubusercontent.com/10785153/111635878-568c9a80-87ce-11eb-85fa-e54e339cff0f.png)

hokiegeek2:
Yes, while I don't get the exact same error, it appears that different length Categorical.segments pdarrays causes concatenate to fail:

![image](https://user-images.githubusercontent.com/10785153/111636895-49bc7680-87cf-11eb-8840-c2e0ace965f6.png)

![image](https://user-images.githubusercontent.com/10785153/111636780-2e516b80-87cf-11eb-86ff-0c8319fbe085.png)

hokiegeek2:
And it does appear that, if the Categorical.segments array length is < numlocales, concatenate fails:

![image](https://user-images.githubusercontent.com/10785153/111641049-27c4f300-87d3-11eb-929b-4d9b9662e71c.png)

![image](https://user-images.githubusercontent.com/10785153/111641206-4dea9300-87d3-11eb-8fca-03818fdfda1e.png)

hokiegeek2:
Found another scenario where concatenate fails:
1 locale
one Categorical with 10 strings and segments, one with 5 strings and 5 segments:

![image](https://user-images.githubusercontent.com/10785153/111677339-80a58300-87f5-11eb-8da8-655d55950c71.png)

hokiegeek2:
Fixed the array index out of bounds error but now getting the return message error noted above:

[...]

PR #718, opened 2021-03-24
added logic to handle situations where (1) the size of 1..n Categoric…

hokiegeek2:
This PR addresses #710 as follows:

1. Fixes bug where 1..n Categorical objects have a segments pdarray with a length < numLocales
2. Fixes bug where the lengths of 1..n Categorical objects have segments arrays with differing lengths 
3. Added unit tests for items (1) and (2)


mhmerrill (approved):
Looks ok to me.

hokiegeek2:
@ronawho so the Categorical.concatenate works up to 12 locales, correct?

Issue #222, opened 2019-12-11
Convert remaining uses of unorderedCopy to aggregation


ronawho:
Aggregation has significantly better performance on non-cray networks. We should remove the remaining uses of unorderedCopy (and look for plain assignments that could be aggregated.)

mhmerrill:
@ronawho these occurrences on ArgSortMsg.chpl should be replaced.
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/ArgSortMsg.chpl#L376
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/ArgSortMsg.chpl#L384
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/ArgSortMsg.chpl#L392
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/ArgSortMsg.chpl#L400-L401
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/ArgSortMsg.chpl#L407

mhmerrill:
these occurrences in FindSegments.chpl should be converted to aggregation.
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/FindSegmentsMsg.chpl#L96
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/FindSegmentsMsg.chpl#L142
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/FindSegmentsMsg.chpl#L148

mhmerrill:
there is a question I have about FindSegments.chpl, would unorderedCopy/Aggregation help with the two commented-out lines? Scattering a `true` into a block distributed bool array.
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/FindSegmentsMsg.chpl#L195
https://github.com/mhmerrill/arkouda/blob/9712f3608ba20b7d69f7f1ab51927d3413f01d3b/src/FindSegmentsMsg.chpl#L211

[...]

PR #720, opened 2021-03-26
Convert unordredCopy to aggregation in findSegmentsMsg

reuster986:
`findSegmentsMsg` is the second half of `ak.GroupBy` and currently takes on the same order of time as the (co)argsort. The whole function could probably use a refactor, but one clear issue was a lingering `unorderedCopy` near the end, which this PR replaces with aggregation. This should hopefully benefit `GroupBy` on all platforms, but especially IB.

### DIFF
--- a/src/FindSegmentsMsg.chpl
+++ b/src/FindSegmentsMsg.chpl
@@ -195,10 +195,8 @@ module FindSegmentsMsg
          Segment boundaries are in terms of permuted arrays, so invert the permutation to get 
          back to the original index
          */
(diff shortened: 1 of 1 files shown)


reuster986:
Addresses #222

mhmerrill (approved):
This looks good to me.

ronawho (commented):
Looks good, heads up that IB testing is down until at least midway through this week (system upgrade) but I'll let you know if I see any impact when that's back up.

PR #722, opened 2021-03-29
Replace some more uses for memoryUsed

ronawho:
Replace more uses of `memoryUsed` with the `getMemUsed` wrapper that
works with 1.23 and 1.24. This is a follow up to #713.


mhmerrill (approved):
this looks ok to me.

PR #732, opened 2021-04-05
Increase the aggregation buffer sizes for non-ugni configurations

ronawho:
Increase the aggregation buffer sizes from 4K to 8K for non-ugni
configurations. This is motivated by improving perforrmace on InfiniBand
systems, but the larger buffer sizes also benefit Ethernet networks as
well. Note that longer term we want to auto-tune these buffer sizes, but
for now this is a better default. On ugni increasing the size does not
change performance much so I didn't want to increase the memory overhead
for no benefit.

Here are some results for the nightly 16-node-cs-hdr configuration using
the default problem size:

| benchmark | before       | after        |
| --------: | -----------: | -----------: |
| gather    |  83.47 GiB/s | 108.17 GiB/s |
| scatter   | 106.99 GiB/s | 128.14 GiB/s |
| argsort   |   8.65 GiB/s |  10.08 GiB/s |

And some scaling results using 8 GiB / node:

ak-gather-perf-lg
ak-scatter-perf-lg
ak-argsort-perf-lg

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -5,9 +5,10 @@ module CommAggregation {
   use CommPrimitives;
 
   // TODO should tune these values at startup
+  param defaultBuffSize = if CHPL_COMM == "ugni" then 4096 else 8096;
   private config const yieldFrequency = getEnvInt("ARKOUDA_SERVER_AGGREGATION_YIELD_FREQUENCY", 1024);
-  private config const dstBuffSize = getEnvInt("ARKOUDA_SERVER_AGGREGATION_DST_BUFF_SIZE", 4096);
-  private config const srcBuffSize = getEnvInt("ARKOUDA_SERVER_AGGREGATION_SRC_BUFF_SIZE", 4096);
+  private config const dstBuffSize = getEnvInt("ARKOUDA_SERVER_AGGREGATION_DST_BUFF_SIZE", defaultBuffSize);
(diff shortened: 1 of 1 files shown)


mhmerrill:
@ronawho so at 128cores/node and 16 nodes this would be 256MB in buffers (512MB at 32 nodes)
cores*8192*(8addr+8data)*nodes = 128*8192*16*16 = 2**(7+13+4+4) =  2**28 bytes = 256MB per aggregator

mhmerrill:
We should set an upper limit and divide it by the number of cores time the number of locales or something like that so we don't blow the bank at higher node counts...

ronawho:
> `256MB` per aggregator

I think that's the per node memory usage, not per aggregator. I agree the memory usage can be a problem, but that's been a problem since the introduction of aggregators and this just doubles the value for some platforms. My general assumption is that as numCores increases so does memory per node, so I don't think 128 core systems changes what percentage of memory we're using too much.

Longer term I expect to use multi-hop aggregators to reduce memory overheads.

> We should set an upper limit and divide it by the number of cores time the number of locales or something like that so we don't blow the bank at higher node counts

Note that even at ~256 nodes I was seeing perf improvements from increasing the buffer size so doing that would limit performance at scale. 256 nodes with 128 cores each would be 32 MiB per aggregator, 4GiB per node which isn't great but I don't this is bad.

mhmerrill (approved):
this look ok to me.

PR #733, opened 2021-04-06
730 categorical in1d bug

hokiegeek2:


arkouda/categorical.py
```diff
@@ -366,10 +366,59 @@ def endswith(self, substr : str) -> pdarray:
...
+        >>> cat = ak.Categorical(strings)
+        >>> cat.in1d(strings)
+        array([True, True, True, True, True])
+        >>> strings = ak.array(['String {}'.format(i) for i in range(5,9)])
+        >>> catTwo = ak.Categorical(strings)
+        >>> cat.in1d(catTwo)
+        array([False, False, False, False, False])
+        """
+        if isinstance(test,Categorical):
+            return in1d(self.categories,test.categories)
```
reuster986, line 418: This is not the correct logic, because this will only tell you whether the categories of a are in the categories of b, not whether the *elements* of a are in the *elements* of b.  I would write it like this:

```python
if isinstance(test, Categorical):
    categoriesisin = in1d(self.categories, test.categories)
else:
    categoriesisin = in1d(self.categories, test)
return categoriesisin[self.codes]
```

mhmerrill (approved):
looks ok to me, as long as @reuster986 concurs.

tests/categorical_test.py
```diff
@@ -112,7 +112,26 @@ def testBinop(self):
         self.assertEqual(('type of argument "other" must be one of (Categorical, str, str_);' +
                           ' got int instead'), 
                          cm.exception.args[0])
-            
+    
+    def testIn1d(self):
+        vals = [i for i in range(0, 5)]
```
reuster986, line 117: The test vectors should have some duplicates, in order to truly exercise the `Categorical` functionality. For example,

```python
vals = [i % 3 for i in range(10)]
valsTwo = [i % 2 for i in range(10)]
# Stringify and Categoricalify
...
answer = ak.array([x < 2 for x in vals])
self.assertTrue((answer == ak.in1d(catOne, catTwo)).all())
```

Side note: the unit tests should reflect the preferred idiom of `ak.in1d(catOne, catTwo)`, as opposed to the method call `catOne.in1d(catTwo)`. When I originally wrote the method, I probably should have prefixed it with an underscore, because I did not intend for users to call it directly.

arkouda/categorical.py
```diff
@@ -366,10 +366,59 @@ def endswith(self, substr : str) -> pdarray:
...
         return categoriesendswith[self.codes]
 
-    def in1d(self, test):
-        #__doc__ = in1d.__doc__
-        categoriesisin = in1d(self.categories, test)
-        return categoriesisin[self.codes]
+    @typechecked
+    def in1d(self, test : Union[Strings,Categorical]) -> pdarray:
+        """
+        Test whether each element of the Categorical.categories object is
```
reuster986, line 372: This is not quite correct -- it's not testing whether each element of `Categorical.categories` is in the test set, but each element of the `Categorical` itself. The returned array should be the same size as the Categorical, not necessarily the same size as `Categorical.categorie`s.

The docstring should not mention `.categories`, because that is an implementation detail abstracted from the user. From the user's perspective, the `Categorical` is an array of strings (not categories), and this function is testing membership of each of those strings in a test set.

reuster986 (changes requested):
A few changes needed to get the right semantics.

arkouda/categorical.py
hokiegeek2: @reuster986 ready for review

tests/categorical_test.py
hokiegeek2: @reuster986 ready for review

arkouda/categorical.py
hokiegeek2: @reuster986 ready for review

mhmerrill:
@reuster986 and @hokiegeek2 when you guys resolve a change pleas click the "resolve conversation" button ;-) it helps me keep track of what is agreed upon

reuster986 (approved):
Looks good!

Issue #729, opened 2021-04-02
Registration should update `pdarray.name` in place


reuster986:
Currently, to safely use a registered array, a user must reassign the array:

```python
a = a.register('myarray')
```

In my experience, users typically expect registration to be an in-place operation:

```python
a.register('myarray')
```

but after this operation, `a.name` is still `id_123`, not `myarray`. This tends to cause confusion, e.g. after calling `ak.clear()`. 

Having the registration function update `a.name` in-place would provide more intuitive usage and would also make it easier to register complex objects without having to reassign every attribute.

mhmerrill:
@reuster986 this is ok with me. I am not sure if you recall but we discussed this when I did the implementation.

reuster986:
@mhmerrill Ah, I forgot about that -- I apologize if I started us down the wrong path. I think after seeing it used in a few real applications, it's easier to see what's important. I'm trying to recall why we settled on an out-of-place approach, but I can't remember...

Just to clarify what I'm thinking, something like:

```python
# For base pdarray
class pdarray:
    ...
    def register(self, name):
        # Send the server message and check for success
        ...
        self.name = name
        # Keep backwards compatibility for code that does `a = a.register(name)`
        return self

# And for complex objects:
class GroupBy:
    ...
    def register(self, name):
        # Register each attribute array with its own suffix
        for attrib in ('segments', 'permutation', 'unique_keys'):
            array = getattr(self, attrib)
            array.register('_'.join([name, attrib]))
        # Backwards compatibility
        return self
```

In this paradigm, each client-side class would have its own `.register` method. Does that seem like a viable approach?

mhmerrill:
@reuster986 i think this is ok, we will need to make some changes in the server also.

mhmerrill:
@reuster986 I took a look at what needs to change to support this change in the server, client code will have to change also.

server side changes:
- change to `RegistrationMsg.chpl`, change the returned `"created ..."` message in `registerMsg()` to just return `"success"`, we will no longer need to call `st.attrib()` since we will not be creating a new object on the client side
- change to `MultiTypeSymbolTable.chpl`, the functionality of `regName()` will have to change to add the new name to the symbol table, point new name at the symbol table entry, and delete the old name from the `tab`
- there may be more changes but I don't see them at this time.

mhmerrill:
@reuster986
On the client side:

[...]

PR #740, opened 2021-04-08
Issue #729: In-line modification for registration

glitch:
This PR (Issue #729) changes arkouda-python object registration to an in-place modification for

- `pdarray`
- `Strings`

It includes server-side changes to the Registration message as well as the actions applied to the SymbolTable.
It includes client-side changes to the register methods.

**Change in behavior**: register is now an in-place operation versus returning a new python object.
**Change in API**: The static method `register_pdarray` has been removed since you can no longer pass it a simple string.  `register(user_defined_name)` are the instance methods which must be used for registration of an object.

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -15,7 +15,7 @@ import builtins
 
 __all__ = ["pdarray", "info", "clear", "any", "all", "is_sorted", "sum", "prod", 
            "min", "max", "argmin", "argmax", "mean", "var", "std", "mink", 
-           "maxk", "argmink", "argmaxk", "register_pdarray", "attach_pdarray", 
+           "maxk", "argmink", "argmaxk", "attach_pdarray",
            "unregister_pdarray"]
 
 logger = getArkoudaLogger(name='pdarrayclass')    
@@ -952,7 +952,7 @@ class pdarray:
 
     def register(self, user_defined_name : str) -> pdarray:
         """
-        Return a pdarray with a user defined name in the arkouda server 
+        Register this pdarray with a user defined name in the arkouda server
         so it can be attached to later using pdarray.attach()
         
         Parameters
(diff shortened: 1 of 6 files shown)


arkouda/pdarrayclass.py
```diff
@@ -15,7 +15,7 @@
 
 __all__ = ["pdarray", "info", "clear", "any", "all", "is_sorted", "sum", "prod", 
            "min", "max", "argmin", "argmax", "mean", "var", "std", "mink", 
-           "maxk", "argmink", "argmaxk", "register_pdarray", "attach_pdarray", 
+           "maxk", "argmink", "argmaxk", "attach_pdarray",
            "unregister_pdarray"]
```
mhmerrill, line 19: did you want to remove `unregister_pdarray` also?

src/MultiTypeSymbolTable.chpl
```diff
@@ -62,10 +62,15 @@ module MultiTypeSymbolTable
                                      "Registering symbol: %s ".format(userDefinedName));            
             }
             
+            // RE: Issue#729 we no longer support multiple name registration of the same object
+            if (registry.contains(name)) {
+                registry -= name;
+            }
+
```
glitch, line 69: Dev Note: With in-place modification on the client side, it was necessary to remove the "old name" to handle the case where a user does something like this:
```
pda.register("ok_name")
pda.register("better_name")
```

arkouda/pdarrayclass.py
glitch: I thought you mentioned there were times when you needed to manually remove something by name because the python object got mis-placed so I left it in.  I can take this out if you want.

mhmerrill: you're right I think we will leave the interface i there, for now. Thanks!

src/MultiTypeSymbolTable.chpl
mhmerrill: nice catch!

mhmerrill (approved):
looks good to me.

PR #741, opened 2021-04-08
ak.from_series support for object dtype to enable generation of Strings

hokiegeek2:
This PR fixes inconsistency in the ak.from_series method signature. Specifically, the method signature does not require a dtype, deferring to the Pandas Series type inference; this type inference does not work for Pandas Series containing strings, for which the dtype is object.

This PR adds logic to check for dtype of object and sets the dtype to str for the numpy conversion. Also added corresponding unit test and pydoc comments.


mhmerrill (approved):
looks good

reuster986 (commented):
looks good!

Issue #734, opened 2021-04-06
Create Registration List Function 


pierce314159:
Related to issue #729 

Create a function to list all objects currently registered. This will be useful for testing the registration method and will enable users to easily see which objects are available to them after reconnecting to the server. I'll be working this issue with @glitch

glitch:
Note: @pierce314159 did find the `SymbolTable.info("__AllSymbols__")` special arg, but the info returned doesn't contain any registration information.  We talked about the trade-offs of adding that registry info here or making a dedicated command / msg to do it.  In the end we feel adding a first-class command is probably better for the following reasons:

- Better visibility from a client perspective since it will be part of the client API
- Adding parsing/filtering code on the client side to split the registered items from everything else would be a little messy to maintain since there really isn't a formalized format to the info strings message.
- Less bytes on the wire and less iterations across a table.  The number of registry items should always be less than or equal to what is in the lookup table.

If folks have a strong feeling opposed to a new message command, please let us know and we could change the approach.

mhmerrill:
@glitch @pierce314159 I want it added to the ak.info call

pierce314159:
@mhmerrill sounds good! I also just noticed that `ak.info(ak.AllSymbols)` throws a `ValueError` if no objects have been created yet. Would it be fine for me to check if the symbol table is empty and return a message saying the table is empty or is the `ValueError` being used by something?

mhmerrill:
> @mhmerrill sounds good! I also just noticed that `ak.info(ak.AllSymbols)` throws a `ValueError` if no objects have been created yet. Would it be fine for me to check if the symbol table is empty and return a message saying the table is empty or is the `ValueError` being used by something?

@pierce314159 
We should be catching this and returning this from the server... i hope the server does not crash ;-o
If the ST is empty maybe that is the right call to return an message which says the ST is empty.

pierce314159:
The server isn't crashing! Just a error on the client side. I'll do a check and return a message if the ST is empty

PR #743, opened 2021-04-09
Issue mhmerrill#734: Create Registration List Function

pierce314159:
This pull request (Issue #734) adds functionality to list all objects in the registry

- Adds `registered` field to the output of `ak.info()`
- Adds special `__RegisteredSymbols__` symbol which prints all registered objects when passed to `ak.info`  (`ak.info(ak.RegisteredSymbols)`).  This behaves similarly to the `__AllSymbols__` symbol
- Adds checks to return a message when info is called with an empty symbol table or `RegisteredSymbols` is used with an empty registry
- Updates `registration_test.py` to test new info() functionality

Change in behavior:  A `registered` field has been added to `ak.info()` . This may effect any logic which relies on the number or order of fields returned by `ak.info()`

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -310,8 +310,9 @@ module MultiTypeSymbolTable
         /*
         Returns verbose attributes of the sym entry at the given string, if the string maps to an entry.
         Pass __AllSymbols__ to process all sym entries in the sym table.
+        Pass __RegisteredSymbols__ to process all registered sym entries.
 
-        Returns: name, dtype, size, ndim, shape, and item size
+        Returns: name, dtype, size, ndim, shape, item size, and registration status
 
         :arg name: name of entry to be processed
         :type name: string
@@ -321,35 +322,26 @@ module MultiTypeSymbolTable
         proc info(name:string): string throws {
             var s: string;
             if tab.size == 0 {
-                s = "the symbol table is empty";
+                s = "__EMPTY_SYMBOLTABLE__";
             }
             else if name == "__AllSymbols__" {
                 for n in tab {
-                    s += "name:%t dtype:%t size:%t ndim:%t shape:%t itemsize:%t registered:%t\n".format(n, 
-                              dtype2str(tab.getBorrowed(n).dtype), tab.getBorrowed(n).size, 
-                              tab.getBorrowed(n).ndim, tab.getBorrowed(n).shape, 
-                              tab.getBorrowed(n).itemsize, registry.contains(n));
+                    s += formatEntry(n, tab.getBorrowed(n));
                 }
             } 
             else if name == "__RegisteredSymbols__" {
                 if registry.size == 0 {
-                    s = "the registry is empty";
+                    s = "__EMPTY_REGISTRY__";
                 }
                 else {
                     for n in registry {
-                        s += "name:%t dtype:%t size:%t ndim:%t shape:%t itemsize:%t registered:%t\n".format(n, 
-                                  dtype2str(tab.getBorrowed(n).dtype), tab.getBorrowed(n).size, 
-                                  tab.getBorrowed(n).ndim, tab.getBorrowed(n).shape, 
-                                  tab.getBorrowed(n).itemsize, registry.contains(n));
+                        s += formatEntry(n, tab.getBorrowed(n));
                     }
                 }
             }
             else {
                 if (tab.contains(name)) {
-                    s = "name:%t dtype:%t size:%t ndim:%t shape:%t itemsize:%t registered:%t\n".format(name, 
-                              dtype2str(tab.getBorrowed(name).dtype), tab.getBorrowed(name).size, 
-                              tab.getBorrowed(name).ndim, tab.getBorrowed(name).shape, 
-                              tab.getBorrowed(name).itemsize, registry.contains(name));
+                    s += formatEntry(name, tab.getBorrowed(name));
                 }
                 else {
                     throw getErrorWithContext(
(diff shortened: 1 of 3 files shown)


src/MultiTypeSymbolTable.chpl
```diff
@@ -320,19 +320,36 @@ module MultiTypeSymbolTable
...
+            else if name == "__RegisteredSymbols__" {
+                if registry.size == 0 {
+                    s = "the registry is empty";
+                }
+                else {
+                    for n in registry {
+                        s += "name:%t dtype:%t size:%t ndim:%t shape:%t itemsize:%t registered:%t\n".format(n, 
+                                  dtype2str(tab.getBorrowed(n).dtype), tab.getBorrowed(n).size, 
+                                  tab.getBorrowed(n).ndim, tab.getBorrowed(n).shape, 
+                                  tab.getBorrowed(n).itemsize, registry.contains(n));
```
glitch, line 343: I think you did the right thing following suit from the previous code.  Two future optimizations we could probably make here are:
 1. Pull the item out of the map (`tab`) once vs the multiple `tab.getBorrowed(n)` calls.
 2. It looks like we're generating/formatting the same underlying string for AllSymbols, RegisteredSymbols, and single items.  If that's the case it might be worth it to consolidate that into its own function and just call it so it ends up looking something like:
 ```
for n in (registry|tab|etc.) {
  s += formatEntry(n, tab.getBorrowed(n));
}
...
proc formatEntry(name:string, item:GenSymEntry): string {
  return "name:%t dtype:%t size:%t ...".format(name, dtype2str(item.dtype), item.size ...);
}
```

If the strings are not in fact the same, forget I said anything :)
(No change requested from me.)

File: tests/registration_test.py [hunk 2]
```diff
@@ -125,3 +125,71 @@ def test_clear(self):
...
+
+        1. info() with an empty symbol table returns 'the symbol table is empty' regardless of arguments
+        2. info(ak.RegisteredSymbols) when no objects are registered returns 'the registry is empty'
+        3. The registered field is set to false for objects that have not been registered
+        4. The registered field is set to true for objects that have been registered
+        5. info(ak.AllSymbols) contains both registered and non-registered objects
+        6. info(ak.RegisteredSymbols) only contains registered objects
+        '''
+        # Cleanup symbol table from previous tests
+        cleanup()
```
glitch, line 141: RE: cleanup() -> Nice catch here, the previous tests leave behind some state since the server & SymTab are shared across the tests.

src/MultiTypeSymbolTable.chpl
```diff
@@ -320,19 +320,36 @@ module MultiTypeSymbolTable
         */
         proc info(name:string): string throws {
             var s: string;
-            if name == "__AllSymbols__" {
+            if tab.size == 0 {
+                s = "the symbol table is empty";
```
mhmerrill, line 324: I think we may want to return a different string here maybe something like `"__EMPTY_SYMBOLTABLE__"`
what do you think? I also worry about the interaction between and empty symboltable/registry and your `cleanup()` function in the test flow.

src/MultiTypeSymbolTable.chpl
```diff
@@ -320,19 +320,36 @@ module MultiTypeSymbolTable
...
                               dtype2str(tab.getBorrowed(n).dtype), tab.getBorrowed(n).size, 
                               tab.getBorrowed(n).ndim, tab.getBorrowed(n).shape, 
-                              tab.getBorrowed(n).itemsize);
+                              tab.getBorrowed(n).itemsize, registry.contains(n));
                 }
-            } else {
+            } 
+            else if name == "__RegisteredSymbols__" {
+                if registry.size == 0 {
+                    s = "the registry is empty";
```
mhmerrill, line 336: same comment as above maybe return `"__EMPTY_REGISRTY__"`

tests/registration_test.py
```diff
@@ -125,3 +125,71 @@ def test_clear(self):
...
+        self.assertEqual(ak.info("keep_me"), 'the symbol table is empty',
+                         msg='info(keep_me) empty symbol message failed')
+        self.assertEqual(ak.info(ak.AllSymbols), 'the symbol table is empty',
+                         msg='info(AllSymbols) empty symbol table message failed')
+        self.assertEqual(ak.info(ak.RegisteredSymbols), 'the symbol table is empty',
+                         msg='info(RegisteredSymbols) empty symbol table message failed')
+
+def cleanup():
+    ak.clear()
+    for registered_object in ak.info(ak.AllSymbols).split('\n')[:-1]:
```
mhmerrill, line 192: what happens when this gets called on an empty symboltable/registry?

mhmerrill:
at some point we may want to convert the returned value from `ak.info()` to a JSON formatted thing so it can easily be a `dict` in python

src/MultiTypeSymbolTable.chpl
hokiegeek2: @mhmerrill agreed, and perhaps an enum instead of a string?

tests/registration_test.py
pierce314159: I double checked this and `ak.info(ak.AllSymbols).split('\n')[:-1]` is an empty list when the symbol table is empty. So a `cleanup()` call with an empty registry should be the same as calling `ak.clear()`

mhmerrill: ok, thanks

mhmerrill (approved):
I think this looks good.

mhmerrill:
@pierce314159 I think this looks good.
Are there any commits outstanding from you? 
I also marked this ready for review.

pierce314159:
In response to feedback above, I've

- [added a helper function](https://github.com/mhmerrill/arkouda/pull/743#discussion_r610754774) , `formatEntry(name:string, item:GenSymEntry)`, for the `ak.info` calls
- updated the `ak.info` doc string
- [changed the empty symbol table message](https://github.com/mhmerrill/arkouda/pull/743#discussion_r611727165)
- [changed the empty registry message](https://github.com/mhmerrill/arkouda/pull/743#discussion_r611727965)
- updated the `cleanup()` function in `registration_test.py` to be more clear and resilient to future changes

Future work for the `info` function will probably include moving towards json like @mhmerrill mentioned

pierce314159:
> 
> 
> @pierce314159 I think this looks good.
> Are there any commits outstanding from you?
> I also marked this ready for review.

@mhmerrill, no outstanding commits! I was just doing a short write up of changes. This should be ready for review,

mhmerrill:
thanks! @pierce314159 i will merge when the CI finished cleanly ;-)

PR #744, opened 2021-04-09
Add module compat flags to makefile targets that build chapel tests

ronawho:
Follow on to #713, we need to add the compatibility modules when
compiling standalone chapel tests using `make test-bin` or 
`make test-chapel`

While here, remove an unneeded `test` target. I added that in 3d0739e,
but I didn't realize that it duplicated `test-bin` functionality.


mhmerrill (approved):
looks good.

Issue #747, opened 2021-04-12
revert fixes for 710


hokiegeek2:

mhmerrill:
closing this because reopening #710

PR #748, opened 2021-04-12
Revert changes for #710 to support important use cases, remove test …

hokiegeek2:
…case that will now fail #747


mhmerrill (approved):
looks ok. I know that the mac CI test failed due to a software install problem in the mac CI flow.

PR #749, opened 2021-04-12
Disable mac CI testing again

ronawho:
It's been pretty flaky recently with infrastructure issues outside our
control. Most issues seem to be networking issues with being unable to
brew install or curl but there have been other variants.

We have gone back and forth with disabling/re-enabling in the past in
#241, #300, and #415 due to similar issue. Given the current instability
just disable it again.


mhmerrill (approved):
looks ok

PR #751, opened 2021-04-12
Issue #729: Registering two objects with the same name will now result in an error

glitch:
Issue #729 Follow up:  Registering two objects with the same name will now result in an error

Given the case:
```
a = ak.ones(3, dtype=ak.int64)
b = ak.ones(3, dtype=ak.int64)
b.fill(2)
a.register("foo")
b.register("foo")
```

This will now result in an error message from the server and present as a RegistrationError to the user.
Previously, this would have silently updated `a` to point to the object `b`.

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -53,18 +53,16 @@ module MultiTypeSymbolTable
                                    errorClass="UnknownSymbolError");
             }
 
-            // check to see if userDefinedName is defined
+            // check to see if userDefinedName is already defined, with in-place modification, this will be an error
             if (registry.contains(userDefinedName)) {
                 mtLogger.error(getModuleName(),getRoutineName(),getLineNumber(),
                                      "regName: requested symbol `%s` is already in use".format(userDefinedName));
-                // We should return an error here
                 throw getErrorWithContext(
                                     msg=incompatibleArgumentsError("regName", name),
                                     lineNumber=getLineNumber(),
                                     routineName=getRoutineName(),
                                     moduleName=getModuleName(),
                                     errorClass="ArgumentError");
-                // throw new owned ArgumentError();
             } else {
                 mtLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                      "Registering symbol: %s ".format(userDefinedName));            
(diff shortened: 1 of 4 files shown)


src/MultiTypeSymbolTable.chpl
```diff
@@ -55,8 +55,16 @@ module MultiTypeSymbolTable
...
+                mtLogger.error(getModuleName(),getRoutineName(),getLineNumber(),
+                                     "regName: requested symbol `%s` is already in use".format(userDefinedName));
+                // We should return an error here
+                throw getErrorWithContext(
+                                    msg=incompatibleArgumentsError("regName", name),
+                                    lineNumber=getLineNumber(),
+                                    routineName=getRoutineName(),
+                                    moduleName=getModuleName(),
+                                    errorClass="ArgumentError");
+                // throw new owned ArgumentError();
```
mhmerrill, line 67: remove this comment?

mhmerrill: oh, i get the comment now it brackets the throw statement.

mhmerrill:
The semantics of `pda.register("UserDefinedName")` is to replace the old name in the symbol table and if it is already registered then also replace the name in the registry... this has the effect of only registering the object under the most recently used name. Please update the the doc strings.

src/MultiTypeSymbolTable.chpl
glitch: Doh, you were right, I goofed it, the `throw new owned ArgumentError();` was supposed to come out (in fact I should probably remove the preceding comment as well or else make it more descriptive).

Actually this brings up a question:   I originally had it as throw new owned ArgumentError(), but saw that other places were using the getErrorWithContext proc... However, there was no entry in the `ServerErrorStrings.chpl` for `argumentError`, only `incompatibleArgumentsError`.    Is it ok to use that here or should I just throw the straight ArgumentError ?

mhmerrill: I think it's ok. unless @hokiegeek2 has something to say about this?

glitch:
Docs updated and those 2 spurious comments removed.

mhmerrill (approved):
these changes look good, thanks!

Issue #745, opened 2021-04-09
request message arguments missing from server command trace


mhmerrill:
When running the server I noticed that the server command arguments are not getting printed in the command trace on my mac. maybe the `format` on line 270 is broken.

```
2021-04-09:14:41:14 [arkouda_server] main Line 270 INFO [Chapel] >>> "delete" ""
2021-04-09:14:41:14 [arkouda_server] sendRepMsg Line 173 INFO [Chapel] repMsg: {"msg":"deleted id_7", "msgType":"NORMAL", "msgFormat":"STRING", "user":"valton"}
2021-04-09:14:41:14 [arkouda_server] main Line 418 INFO [Chapel] <<< delete took 0.00019300000030852971 sec
2021-04-09:14:41:14 [arkouda_server] main Line 422 INFO [Chapel] bytes of memory used after command 42302
```
https://github.com/mhmerrill/arkouda/blob/d6ae98690c9211161ed76b7a5a6f8363b1de295a/src/arkouda_server.chpl#L270-L272

hokiegeek2:
Fixed! 👍 

![image](https://user-images.githubusercontent.com/10785153/114427326-fac4ef80-9b88-11eb-963e-e727fad01b42.png)

mhmerrill:
fix looks good, thanks. closing issue.

PR #752, opened 2021-04-12
fixed missing args string in trace message #745

hokiegeek2:
This PR fixes the trace logging statement bug where the cmd string failed to get passed to the string.format method.


mhmerrill (approved):
looks good

Issue #753, opened 2021-04-12
add doc strings and type checking for Strings registration


mhmerrill:
here is the relevant code to be added to:
https://github.com/mhmerrill/arkouda/blob/08c7cc1645579a853e4dd67161199ec4f5a99a08/arkouda/strings.py#L831-L844

PR #756, opened 2021-04-13
Issue #753: Add doc strings and type checking for Strings registration

pierce314159:
This pull request (Issue #753) updates doc strings and type checking for registration functions in `strings.py` and `pdarrayclass.py` 

- Adds doc strings to the `register`, `unregister`, and `attach` functions in `strings.py`
- Updates doc strings to the `register`, `unregister`, and `attach` functions in `pdarrayclass.py`
- Adds `@typechecked`  check to ensure `user_defined_name`  is of type `str` for `register` and `attach` functions in both `strings.py` and `pdarrayclass.py`
- Adds `test_registration_type_check` to test the type checking of the `register` and `attach` functions

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -828,17 +828,114 @@ class Strings:
         self.bytes.save(prefix_path=prefix_path, 
                                     dataset='{}/values'.format(dataset), mode=mode)
 
-    def register(self, user_defined_name : str) -> Strings:
+    def register(self, user_defined_name: str) -> Strings:
+        """
+        Register this Strings object with a user defined name in the arkouda server
+        so it can be attached to later using Strings.attach()
+        This is an in-place operation, registering a Strings object more than once will
+        update the name in the registry and remove the previously registered name.
+        A name can only be registered to one object at a time.
+
+        Parameters
+        ----------
+        user_defined_name : str
+            user defined name which the Strings object is to be registered under
+
+        Returns
+        -------
+        Strings
+            The same Strings object which is now registered with the arkouda server and has an updated name.
+            This is an in-place modification, the original is returned to support a fluid programming style.
+            Please note you cannot register two different objects with the same name.
+
+        Raises
+        ------
+        TypeError
+            Raised if pda is neither a pdarray nor a str or if
+            user_defined_name is not a str
+        RegistrationError
+            If the server was unable to register the Strings object with the user_defined_name
+            If the user is attempting to register more than one object with the same name, the former should be
+            unregistered first to free up the registration name.
+
+        See also
+        --------
+        attach, unregister
+
+        Notes
+        -----
+        Registered names/Strings objects in the server are immune to deletion
+        until they are unregistered.
+        """
+        if not isinstance(user_defined_name, str):
+            raise TypeError(f"user_defined_name must be of type str, was {type(user_defined_name)}")
+
         self.offsets.register(user_defined_name+'_offsets')
         self.bytes.register(user_defined_name+'_bytes')
         self.name = user_defined_name
         return self
 
     def unregister(self) -> None:
+        """
+        Unregister a Strings object in the arkouda server which was previously
+        registered using register() and/or attached to using attach()
+
+        Parameters
+        ----------
+
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        TypeError
+            Raised if pda is neither a pdarray nor a str
+
+        See also
+        --------
+        register, unregister
+
+        Notes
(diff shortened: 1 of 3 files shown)


arkouda/strings.py
```diff
@@ -828,17 +828,114 @@ def save(self, prefix_path : str, dataset : str='strings_array',
         self.bytes.save(prefix_path=prefix_path, 
                                     dataset='{}/values'.format(dataset), mode=mode)
 
-    def register(self, user_defined_name : str) -> Strings:
+    def register(self, user_defined_name: str) -> Strings:
```
mhmerrill, line 831: add `@typechecked` to these methods and you don't have to manually check the types of the arguments.

arkouda/strings.py
```diff
@@ -828,17 +828,114 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+        See also
+        --------
+        attach, unregister
+
+        Notes
+        -----
+        Registered names/Strings objects in the server are immune to deletion
+        until they are unregistered.
+        """
+        if not isinstance(user_defined_name, str):
```
mhmerrill, line 870: you don't need this if you use `@typechecked` to decorate the function

arkouda/strings.py
```diff
@@ -828,17 +828,114 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+        -----
+        Registered names/Strings objects in the server are immune to deletion until
+        they are unregistered.
+        """
         self.offsets.unregister()
         self.bytes.unregister()
 
     @staticmethod
     def attach(user_defined_name : str) -> Strings:
+        """
```
mhmerrill, line 909: add `@typechecked` here also

arkouda/strings.py
```diff
@@ -828,17 +828,114 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+        See also
+        --------
+        register, unregister
+
+        Notes
+        -----
+        Registered names/Strings objects in the server are immune to deletion
+        until they are unregistered.
+        """
+        if not isinstance(user_defined_name, str):
```
mhmerrill, line 937: not needed if we use `@typechecked` on the function

arkouda/strings.py
```diff
@@ -828,17 +828,114 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+        -------
+        Strings
+            The same Strings object which is now registered with the arkouda server and has an updated name.
+            This is an in-place modification, the original is returned to support a fluid programming style.
+            Please note you cannot register two different objects with the same name.
+
+        Raises
+        ------
+        TypeError
+            Raised if pda is neither a pdarray nor a str or if
```
glitch, line 854: I think we can remove the pda & pdarray line here... (both here and in the corresponding pdarray class method since we are only taking in a string argument, so the user_defined_name can stay).

pierce314159: Okay, I can go ahead and remove it for `strings`. Should I also go ahead and remove it from `pdarrayclass` or open a separate issue for changing those `pdarrayclass` docstrings?

arkouda/strings.py
```diff
@@ -828,17 +828,114 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+        ----------
+
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        TypeError
+            Raised if pda is neither a pdarray nor a str
```
glitch, line 893: I'm not sure the original docs on pdarrayclass.unregister() were correct to begin with.  I don't think a TypeError is raised here, only a RuntimeError.

I believe this piece should be replaced with:
```
RuntimeError
    Raised if the server could not find the internal name/symbol to remove.
```

arkouda/strings.py
glitch: My opinion is that it's ok to fix the documention in pdarrayclass as well.

If @mhmerrill agrees it's ok to modify the code for the `typechecked` annotations he mentioned above, then I say this ticket should be updated to add the type-checking in both `Strings` and `pdarrayclass` along with the documentation updates.

mhmerrill: yes, I think that is good.

mhmerrill (approved):
look ok to me.

pierce314159:
I forgot to update the doc_strings of the `attach_pdarray` and `unregister_pdarray` functions at the bottom of the `pdarrayclass.py` file. These docstrings are essentially the same as the class level functions `attach` and `unregister`. I think this should be my last commit to satisfy the current requested changes.

PR #758, opened 2021-04-14
Fix categorical bug

reuster986:
This PR attempts to fix a bug when concatenating two Categoricals with different categories that scrambles the mapping of codes to categories and results in completely wrong output. I believe I traced the bug back to a wrong flag in the `Categorical.concatenate()` method. I added a test case that addresses the scenario in question.

Marked as draft until CI passes.

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -520,8 +520,7 @@ class Categorical:
                                        [o.categories for o in others],
                                        ordered=True))
             newidx = g.unique_keys
-            wherediditgo = zeros(newidx.size, dtype=akint64)
-            wherediditgo[g.permutation] = arange(newidx.size)
+            wherediditgo = g.broadcast(arange(newidx.size), permute=True)
             idxsizes = np.array([self.categories.size] + \
                                 [o.categories.size for o in others])
             idxoffsets = np.cumsum(idxsizes) - idxsizes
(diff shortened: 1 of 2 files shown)


hokiegeek2:
@reuster986 was this tested with an Arkouda version with [710](https://github.com/mhmerrill/arkouda/issues/710) changes?

reuster986:
@hokiegeek2 I'm not sure, but this branch was created from master about an hour ago, if that helps.

hokiegeek2:
> @hokiegeek2 I'm not sure, but this branch was created from master about an hour ago, if that helps.

Ah, gotcha, just checking to see if the testing you did there was with a version of Arkouda that had 710 in it (reverted those changes on Monday).

hokiegeek2 (approved):

reuster986:
@hokiegeek2 ah, ok. In that case, I think the answer is "no". I can merge this PR and then you could try re-introducing 710 on a different branch, if you have a way to test it?

hokiegeek2:
> @hokiegeek2 ah, ok. In that case, I think the answer is "no". I can merge this PR and then you could try re-introducing 710 on a different branch, if you have a way to test it?

Yup, will test it out and then if you would, please pull the corresponding branch in to test in the office.

Issue #735, opened 2021-04-07
Create isRegistered function for pdarrayclass/Strings objects


glitch:
Related to issues #729 and #734 it would be nice to have an `isRegistered` function on the pdarrayclass objects.  More generally all we really need is a function on the client `isRegistered(<name:str>)` which can be used by pdarrayclass, GroupBy, etc.

Server side code to check the internal registry in MultiTypeSymbolTable.SymbolTable  and return simple True/False messages will need to be added as well as all of the plumbing for a new command etc.

Note: We can submit this change with the #729 PR, but as a separate consolidated commit in case we need to unwind/backport/etc. it later.

mhmerrill:
I am ok with this approach.

mhmerrill:
@pierce314159 did we add this functionality? I don't remember seeing it in the PRs.

pierce314159:
@mhmerrill not yet, this is my goal for today. Do we want to stick with adding a new command on the server side approach or would it be better to handle this all on the client side using the list registry functionality of the info command?

mhmerrill:
i think just using the pulling the registered value from `ak.info()` seems like the right way to go since the information is there.

mhmerrill:
@pierce314159 could you also pull the notebook from the notebook repo and add some stuff to it...

https://github.com/mhmerrill/ArkoudaNotebooks/blob/master/RegistrationTest.ipynb

pierce314159:
> 
> 
> @pierce314159 could you also pull the notebook from the notebook repo and add some stuff to it...
> 
> https://github.com/mhmerrill/ArkoudaNotebooks/blob/master/RegistrationTest.ipynb

Can do!

pierce314159:
I'm about to put a pull request for this issue. I added a generic `list_registry` function in addition to class level `is_registered` functions for `Strings` and `pdarrayclass` all of this was done on the client side using the `ak.info("__RegisteredSymbols__")` functionality from issue #734

I did not include `groupby` or `categorical` objects since they don't have `register` functions yet

PR #759, opened 2021-04-14
Issue #735: Create isRegistered function for pdarrayclass/Strings objects

pierce314159:
This pull request (Issue #735 ) adds a generic `ak.list_registry()` method and class level `is_registered` methods for `Strings` and `pdarrayclass`. All of this was done on the client side using the `ak.info("__RegisteredSymbols__")` functionality from issue #734

- Adds `ak.list_registry()` method and a test for this functionality
- Adds `Strings.is_registered()` and `pdarray.is_registered()` methods and associated tests
- Slight change to `Strings.attach` method to re-declare `Strings.name` asked for by @glitch  

Changes resulting from PR feedback:
- Updates return type of `list_registry` method to `List[str]`
- Updates `list_registry` so function has single exit point 
- Adds `ak.EmptyRegistry` and `ak.EmptySymbolTable` and updates references to the string literals `'__EMPTY_REGISTRY__'` and `'__EMPTY_SYMBOLTABLE__'`
- Moves `test_string_registration` and `test_string_is_registered` from `string_test.py` to `registration_test.py`

This PR not include `groupby.is_registered()` or `categorical.is_registered()` methods since they don't have `register` methods yet


arkouda/strings.py
```diff
@@ -840,5 +860,7 @@ def unregister(self) -> None:
 
     @staticmethod
     def attach(user_defined_name : str) -> Strings:
-        return Strings(pdarray.attach(user_defined_name+'_offsets'),
+        s = Strings(pdarray.attach(user_defined_name+'_offsets'),
                        pdarray.attach(user_defined_name+'_bytes'))
+        s.name = user_defined_name
+        return s
```
pierce314159, line 866: This change was requested by @glitch. The rationale is if the client is disconnected from the server, then `Strings.name` would've been wiped out and should be re-declared when we attach

arkouda/pdarrayclass.py
```diff
@@ -1171,6 +1191,38 @@ def info(pda : Union[pdarray, str]) -> str:
         raise TypeError("info: must be pdarray or string".format(pda))
         return generic_msg(cmd="info", args="{}".format(pda))
 
+def list_registry() -> list:
```
hokiegeek2, line 1194: I suggest replacing this with List[str]

arkouda/strings.py
```diff
@@ -828,6 +828,26 @@ def save(self, prefix_path : str, dataset : str='strings_array',
...
+        """
+        Return True iff the object is contained in the registry
+
+        Parameters
+        ----------
+        None
+
+        Returns
+        -------
+        bool
```
hokiegeek2, line 841: Not sure why you are returning np.bool_ vs a Python bool?

arkouda/strings.py
hokiegeek2: 👍 good

arkouda/strings.py
pierce314159: I chose to do it out of consistency with the other code. Like `def is_sorted(pda : pdarray) -> np.bool_:` in `pdarrayclass.py`. I wasn't exactly sure why `np.bool_` was chosen but figured I should probably follow suit. I didn't see a good reason to use it in `is_sorted` but not here, but I can certainly change it if there is

hokiegeek2: ah gotcha, cool, makes sense, was curious, that's all, no need to change.

pierce314159: That's totally reasonable! If @mhmerrill wanted to weigh in on the original design choice (and if it makes sense to use `np.bool_` here), I'd be interested to hear it

arkouda/pdarrayclass.py
pierce314159: @hokiegeek2, I think this should be updated now. I changed the return type to `list[str]`. I'm not sure if the capitalization was important here but mypy complained less with this way

arkouda/pdarrayclass.py
```diff
@@ -1168,6 +1188,38 @@ def info(pda : Union[pdarray, str]) -> str:
...
+
+    Raises
+    ------
+    RuntimeError
+        Raised if there's a server-side error thrown
+    """
+    registered_list: list[str] = []
+
+    if info('__RegisteredSymbols__') == '__EMPTY_REGISTRY__':
+        return registered_list
```
glitch, line 1212: No change requested, just stylistic:
You could wrap the for loop below in this if condition (flipping to != comparison) and reduce it to a single exit point.
i.e.

````
registered_list: List[str] = []
if info(...) != '__EMPTY_REGISTRY__':
  for (...):
    # update registered_list as necessary
    registered-list.append(...)

return registered_list
```

arkouda/pdarrayclass.py
```diff
@@ -1168,6 +1188,38 @@ def info(pda : Union[pdarray, str]) -> str:
...
+
+    registered_object: str
+    for registered_object in filter(None, info('__RegisteredSymbols__').split('\n')):
+        if registered_object is not None:
+            name = registered_object.split()[0].split(':')[1].replace('"', '')
```
glitch, line 1217: No change required,  stylistic comment:
It may seem silly for a one-liner, but moving this object parsing to its own function makes it more test-able from a unit testing perspective and easier to swap out later... although it would add a bit more complexity (additional method etc.) to there is a trade-off.

I know we've been discussing moving the return structure to some type of json so it's probably best to wait for that to happen.

arkouda/pdarrayclass.py
```diff
@@ -1168,6 +1188,38 @@ def info(pda : Union[pdarray, str]) -> str:
...
+    RuntimeError
+        Raised if there's a server-side error thrown
+    """
+    registered_list: list[str] = []
+
+    if info('__RegisteredSymbols__') == '__EMPTY_REGISTRY__':
+        return registered_list
+
+    registered_object: str
+    for registered_object in filter(None, info('__RegisteredSymbols__').split('\n')):
```
glitch, line 1215: I noticed `__RegisteredSymbols__` is used twice here; does it make sense to import it directly from `arkouda.client` and use the defined form vs the string literal?

arkouda/pdarrayclass.py
```diff
@@ -1168,6 +1188,38 @@ def info(pda : Union[pdarray, str]) -> str:
...
+    -------
+    list
+        List of all object names in the registry
+
+    Raises
+    ------
+    RuntimeError
+        Raised if there's a server-side error thrown
+    """
+    registered_list: list[str] = []
```
glitch, line 1209: I think the type string here should be `List[str]` vs `list[str]`
The python `typing` module uses the capitalized for for annotating typing info.

arkouda/pdarrayclass.py
```diff
@@ -1168,6 +1188,38 @@ def info(pda : Union[pdarray, str]) -> str:
         raise TypeError("info: must be pdarray or string".format(pda))
         return generic_msg(cmd="info", args="{}".format(pda))
 
+def list_registry() -> list[str]:
```
glitch, line 1191: I think the type string here should be List[str] vs list[str]
The python typing module uses the capitalized for for annotating typing info.

glitch (commented):
Looks pretty good 👍 

- I think the `List[str]` vs `list[str]` piece should be made to be consistent with python's typing module.
- I left a _stylistic_ recommendations, but you don't need to change them unless you want to.

hokiegeek2:
> Looks pretty good 👍
> 
> * I think the `List[str]` vs `list[str]` piece should be made to be consistent with python's typing module.
> * I left a few _stylistic_ recommendations, but you don't need to change them unless you want to.

@glitch Agreed, but list[str] is not supported until 3.8, so we're stuck with List[str] for now, (ugh)

arkouda/pdarrayclass.py
hokiegeek2: @glitch yup, but I believe we can do list[str] (better option, IMHO) when we officially upgrade to 3.8 as minimum Python version (pretty sure it's 3.8, will test to confirm).

arkouda/pdarrayclass.py
hokiegeek2: @pierce314159 Gothca, yeah, unfortunately, I don't believe list[str] works for 3.7:

image

So pretty sure you need to switch to List[str]

glitch:
> 
> 
> > Looks pretty good 👍
> > 
> > * I think the `List[str]` vs `list[str]` piece should be made to be consistent with python's typing module.
> > * I left a few _stylistic_ recommendations, but you don't need to change them unless you want to.
> 
> @glitch Agreed, but list[str] is not supported until 3.8, so I think we're stuck with List[str] for now, (ugh)
> 

@hokiegeek2 I was agreeing with your comment from yesterday, I think the capital version is what we need (`List[str]`).  @pierce314159 does mypy complain about the capitalized version?  (Note, my base install is with python 3.7)

hokiegeek2:
> > > Looks pretty good 👍
> > > 
> > > * I think the `List[str]` vs `list[str]` piece should be made to be consistent with python's typing module.
> > > * I left a few _stylistic_ recommendations, but you don't need to change them unless you want to.
> > 
> > 
> > @glitch Agreed, but list[str] is not supported until 3.8, so I think we're stuck with List[str] for now, (ugh)
> 
> @hokiegeek2 I was agreeing with your comment from yesterday, I think the capital version is what we need (`List[str]`). Sorry looking at my comment that probably wasn't clear. @pierce314159 does mypy complain about the capitalized version? (Note, my base install is with python 3.7)

Gotcha, yeah, was just ensuring that Pierce was aware that list[str] is not in 3.7, that's all. 👍

arkouda/pdarrayclass.py
mhmerrill: I would like you to use ak.RegisteredSymbols instead of the string literal.

tests/string_test.py
```diff
@@ -505,3 +505,17 @@ def test_string_registration(self):
         ak.clear()
         with self.assertRaises(RuntimeError, msg="keep was unregistered and should be cleared"):
             str(keep) # should cause RuntimeError
+
+    def test_string_is_registered(self):
```
mhmerrill, line 509: I think this should be included in the registration tests instead of here but I am open to a counter argument.

arkouda/pdarrayclass.py
mhmerrill: I think I'd rather have defined symbols like `ak.EmptyRegistry` and `ak.EmptySymbolTable` instead of string literals showing up in the code.

tests/string_test.py
hokiegeek2: @mhmerrill ah, good point, I agree.

glitch: @mhmerrill @hokiegeek2 , if @pierce314159 moves this new test `test_string_is_registered` should the existing `test_string_registration` also go with it?

pierce314159:
> 
> 
> > > > Looks pretty good 👍
> > > > 
> > > > * I think the `List[str]` vs `list[str]` piece should be made to be consistent with python's typing module.
> > > > * I left a few _stylistic_ recommendations, but you don't need to change them unless you want to.
> > > 
> > > 
> > > @glitch Agreed, but list[str] is not supported until 3.8, so I think we're stuck with List[str] for now, (ugh)
> > 
> > 
> > @hokiegeek2 I was agreeing with your comment from yesterday, I think the capital version is what we need (`List[str]`). Sorry looking at my comment that probably wasn't clear. @pierce314159 does mypy complain about the capitalized version? (Note, my base install is with python 3.7)
> 
> Gotcha, yeah, was just ensuring that Pierce was aware that list[str] is not in 3.7, that's all. 👍

@glitch helped me find my issue here. I was missing a `from Typing import List` 🙄
I'll push up a new commit for this and the other feedback soon

arkouda/pdarrayclass.py
pierce314159: @glitch and @mhmerrill, I agree completely. I'll change this in my next commit

arkouda/pdarrayclass.py
pierce314159: Sounds good! I can make that change. I'll also update references to this in registration tests

tests/string_test.py
pierce314159: yeah my reasoning for putting `test_string_is_registered` in `string_test.py` was that the `test_string_registration` method was already there. And that they are both methods of the `Strings` class. It makes just as much sense to me to put them in the registration tests. Should I move both or just `test_string_is_registered`?

mhmerrill: yes, move them

mhmerrill (approved):
this looks good now.

mhmerrill:
@pierce314159 @glitch @hokiegeek2 Anybody want to pile on or am I good to merge now?

glitch (approved):

Issue #757, opened 2021-04-14
replace --v flag with --loglevel


hokiegeek2:
Logging in Arkouda can be set either at the debug or info level, which can make it difficult to debug specific modules as well as lead to unnecessarily verbose logging overall.  I recommend replacing the --v boolean flag with a --loglevel flag, the values of which can be DEBUG, INFO, WARN, ERROR, or CRITICAL.

hokiegeek2:
got this working directly w/ the LogLevel enum (thanks @mhmerrill!):

![image](https://user-images.githubusercontent.com/10785153/114760432-4b754d80-9d2d-11eb-9b1c-ca822e2a4e5c.png)

hokiegeek2:
This is a good example of using the new logging config options to enable focus on debugging concatenate:

![image](https://user-images.githubusercontent.com/10785153/115009625-7a4d0a00-9e7a-11eb-88cb-3c6b4178c61a.png)

Issue #728, opened 2021-04-02
Make server-side loggers individually configurable.


glitch:
Currently the loggers on the arkouda_server are managed with a global configuration variable `v` defined in [ServerConfig](https://github.com/mhmerrill/arkouda/blob/8e4400fd1abb307aacef47f56404d59f3a05f6ff/src/ServerConfig.chpl#L24)

Chapel allows `config const` values to be set via cmd line args such as  `--FooClass.v=true` and we can take advantage of this feature to allow more customized logging configuration.

It may also be worth looking into a common log configuration format to allow groups of modules/classes to be turned on and off at once (i.e. something similar to how logback or log4j work... not the dynamic part, but the config file part).  If we look at the (optional) config file the format should be something like yaml or json, and the order of override should be config file, overridden by command-line args.

glitch:
Associating with Issue #757

hokiegeek2:
Got this working. The default logging level is INFO. The first example is with arkouda_server (notice arkouda_directory debug message):

![image](https://user-images.githubusercontent.com/10785153/114772394-bd549380-9d3b-11eb-9926-411efd8801c6.png)

![image](https://user-images.githubusercontent.com/10785153/114772342-ac0b8700-9d3b-11eb-8a5d-7d5fdabff151.png)

Second one is MsgProcessing debug:

![image](https://user-images.githubusercontent.com/10785153/114772749-2e944680-9d3c-11eb-9806-ffab33111d49.png)

![image](https://user-images.githubusercontent.com/10785153/114772570-f5f46d00-9d3b-11eb-9976-6cc7bc5dbe71.png)

And the third one is MsgProcessing warn:

![image](https://user-images.githubusercontent.com/10785153/114772714-20462a80-9d3c-11eb-8908-afdabb62d848.png)

PR #760, opened 2021-04-14
replaced --v flag with LogLevel enum config and implemented module-level logging #728 #757

hokiegeek2:
This draft PR enhances logging configuration as follows:

1. Replaces --v flag with --logLevel flag to enable a global DEBUG, INFO. WARN, ERROR, or CRITICAL logging level
2. Enables module-level logging level setting (implemented in a subset of the modules)

If the design of (2) is cool, will rollout to rest of modules, remove --v flag, and remove no-arg Logger.init() function

### DIFF
--- a/src/Logging.chpl
+++ b/src/Logging.chpl
@@ -2,6 +2,8 @@ module Logging {
     use Set;
     use IO;
     use DateTime;
+    use Reflection;
+    use Errors;
 
     /*
      * The LogLevel enum is used to provide a strongly-typed means of
@@ -14,7 +16,7 @@ module Logging {
      * of logging sensitivity analogous to other languages such as Python.
      */
     class Logger {
-        var level = LogLevel.WARN;
+        var level: LogLevel = LogLevel.INFO;
         
         var warnLevels = new set(LogLevel,[LogLevel.WARN, LogLevel.INFO,
                            LogLevel.DEBUG]);
@@ -28,6 +30,12 @@ module Logging {
         var infoLevels = new set(LogLevel,[LogLevel.INFO,LogLevel.DEBUG]);  
         
         var printDate: bool = true;
+        
+        proc init(logLevel : LogLevel) {
+            level = logLevel;
+        }
+        
+        proc init() {}
         
         proc debug(moduleName, routineName, lineNumber, msg: string) throws {
             if level == LogLevel.DEBUG  {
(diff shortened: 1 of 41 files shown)


glitch:
Finished initial review at commit baa5017.  Looks good!

pierce314159:
This looks good to me!

mhmerrill (commented):
I think this is looking good so far. I think we should add some examples of use to the PR.

hokiegeek2:
> I think this is looking good so far. I think we should add some examples of use to the PR.

@mhmerrill Gotcha, cool, good point, I will add a few use cases and corresponding screenshots

hokiegeek2:
setting logLevel to INFO via the --logLevel=LogLevel.INFO leads to just INFO,WARN,ERROR,CRITICAL messages to be printed from all modules:

image

setting the global logging level at INFO and MsgProcessing to DEBUG results in targeted debug messages:

image

logging is strongly quieted if the global logging level is set to WARN, enabling focus on the MsgProcessing module:

image

mhmerrill (commented):
this looks good to me. I'd like to get @bradcray or @ronawho to weigh in on this since it's a newer chapel feature we haven't really leaned into before.

src/Logging.chpl
```diff
@@ -85,4 +93,26 @@ module Logging {
             return "%s:%s".format(cd,rawCms(0));        
         }
     }
+    
+    /*
+     * Returns the LogLevel enum corresponding to the supplied log
+     * level string.
+     */
+    proc getLogLevel(level : string) : LogLevel throws {
```
ronawho, line 101: I'm not seeing this used anywhere. I also think that casting strings to enums is a built-in feature so you might not need this anyways:

```chpl
enum LogLevel {INFO, DEBUG}
config const level = "INFO";
if level:LogLevel == LogLevel.INFO then writeln("level=info");
```

ronawho (approved):

src/Logging.chpl
hokiegeek2: @ronawho Ah gotcha. I added this with the thought that eventually we may have a admin feature where a user could pass in a log level string like "INFO" and then have a *Msg function set the log level for the corresponding module. But yeah, good idea, since we're passing in a CLI string like LogLevel.INFO that is getting converted to the enum.

hokiegeek2: @ronawho removed the unneeded Logging module code.

mhmerrill (approved):
looks good to me

PR #765, opened 2021-04-19
Closes #764: Adds guard condition to list_registry for empty symbol table

### LINKED ISSUE #764: list_registry() fails when SymbolTable is empty
When the SymbolTable is empty a call to `pdarrayclass.list_registry()` raises an IndexError when trying to parse the result.  This is due to a missing guard condition for `"__EMPTY_SYMBOLTABLE__"`

glitch:
Closes #764: Adds guard condition to list_registry for empty symbol table
along with updated unit test.


mhmerrill (approved):
good catch

PR #766, opened 2021-04-19
fix failing concatenate and add guard condition

hokiegeek2:
This PR does the following:
1. Addresses #721 by disabling the testConcatenate test case that fails due to #710 
2. Adds client-side guard condition that prevents manifestation of #710 by checking to see if any Categorical or Strings object to be concatenated has a len() < numLocales. If so, a ValueError is raised.
3. Add test_concatenate to string_test

### DIFF
--- a/arkouda/pdarraysetops.py
+++ b/arkouda/pdarraysetops.py
@@ -222,7 +222,7 @@ def concatenate(arrays : Sequence[Union[pdarray,Strings,'Categorical']], #type: 
         Check if any Strings or Categorical objeeccts have length < numLocales per 
         #710 and #721. TODO: remove once #710 is resolved.
         '''
-        numLocales = get_config()['numLocales']
+        numLocales = int(get_config()['numLocales'])
         for arr in arrays:
             if len(arr) < numLocales:
                 raise ValueError('Lengths of Strings or Categoricals must be >= numLocales')        
--- a/arkouda/pdarraysetops.py
+++ b/arkouda/pdarraysetops.py
@@ -217,7 +217,7 @@ def concatenate(arrays : Sequence[Union[pdarray,Strings,'Categorical']], #type: 
     else:
         mode = 'interleave'
 
-    if mode == 'interleave' and isinstance(arrays[0],(type(Strings),type(Categorical))):
+    if mode == 'interleave' and isinstance(arrays[0],(Strings,Categorical)):
         '''
         Check if any Strings or Categorical objects have length < numLocales per 
         #710 and #721 (Strings and Categorical concatenate in interleave mode fails with
--- a/arkouda/pdarraysetops.py
+++ b/arkouda/pdarraysetops.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from typing import cast, Optional, Sequence, Tuple, Union, ForwardRef
 from typeguard import typechecked
-from arkouda.client import generic_msg
+from arkouda.client import generic_msg, get_config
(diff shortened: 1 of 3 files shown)


reuster986:
@hokiegeek2 I don't see the guard condition you added -- the only file I see changed is `string_test.py`. Am I missing something?

More to the point, I don't think the guard condition should raise a `ValueError`, given that concatenating small arrays is a perfectly valid operation. Better to have the guard condition call `.to_ndarray()`, do the concatenation on the client, then transfer the result back over to the server. 

Finally, I just want to make sure we are treating this PR as a temporary workaround while we continue working to get concatenation correct across all cases.

hokiegeek2:
> @hokiegeek2 I don't see the guard condition you added -- the only file I see changed is `string_test.py`. Am I missing something?
> 
> More to the point, I don't think the guard condition should raise a `ValueError`, given that concatenating small arrays is a perfectly valid operation. Better to have the guard condition call `.to_ndarray()`, do the concatenation on the client, then transfer the result back over to the server.
> 
> Finally, I just want to make sure we are treating this PR as a temporary workaround while we continue working to get concatenation correct across all cases.

Argh, sorry, git commit fail on my part, fixed. 

The guard condition is there because concatenation of Strings or Categorical objects where 1..n of the objects have a length < numLocales fails. I thought I had fixed this in #710 but the tests you added proved that my change caused other errors. The guard condition does not apply to pdarrays.

And yes, as I discussed with @mhmerrill, this is intended to be a temp workaround.

reuster986 (approved):
Looks like users can still concatenate small arrays using `ordered=True`, so I'm fine with this for now, although maybe the error message should tell the user to retry with `ordered=True`.

hokiegeek2:
> Looks like users can still concatenate small arrays using `ordered=True`, so I'm fine with this for now, although maybe the error message should tell the user to retry with `ordered=True`.

yes, I apologize, I should have made that clearer. The guard condition only applles to Strings or Categoricals in interleave mode AKA ordered=False. I will update the error message to reflect that

reuster986:
> The guard condition is there because concatenation of Strings or Categorical objects where 1..n of the objects have a length < numLocales fails. I thought I had fixed this in #710 but the tests you added proved that my change caused other errors. The guard condition does not apply to pdarrays.

I was talking about Categoricals, not pdarrays. Side note: when I say "array", I mean any of the following: pdarray, Strings, Categorical, Timedelta, Datetime -- basically anything that can be a column of data. I wanted to make sure the user has a way to concatenate small Categorical arrays, but it looks like they can still do that with `ordered=True` without needing to bring back to the client -- is that correct?

hokiegeek2:
> > The guard condition is there because concatenation of Strings or Categorical objects where 1..n of the objects have a length < numLocales fails. I thought I had fixed this in #710 but the tests you added proved that my change caused other errors. The guard condition does not apply to pdarrays.
> 
> I was talking about Categoricals, not pdarrays. Side note: when I say "array", I mean any of the following: pdarray, Strings, Categorical, Timedelta, Datetime -- basically anything that can be a column of data. I wanted to make sure the user has a way to concatenate small Categorical arrays, but it looks like they can still do that with `ordered=True` without needing to bring back to the client -- is that correct?

👍 Indeed!

reuster986:
@hokiegeek2 looks like you just answered my question -- thanks!

hokiegeek2:
@reuster986 new error message:

image

reuster986:
@mhmerrill would you like to review, or should we go ahead and merge?

PR #768, opened 2021-04-20
Closes #761: Adding registration suite of functions for Categorical

### LINKED ISSUE #761: Add registration functionality to Categorical
Part of Issue #729 involves adding `register` support to GroupBy.  GroupBy internal components may be of type `Categorical` which currently does not have a `register` implementation.

UPDATED 2021-04-22: The scope of this ticket has changed after discussion on registration, in-place changes, and scopes.  The goal of this ticket is now:

- Use dot character to delineate sub-component names/extensions for Categorical & Strings subcomponents (i.e. for Strings object registered as `foo`, it will register components `foo.offsets` and `foo.bytes`)
- Static unregister functions will be added / remain in place for Categorical, Strings, and pdarray with the following changes
    - consistent naming following pattern `unregister_<class>_by_name`
    - Only support string arg case.  Object case should use the instance `unregister()` methods
- Add static set of component names for Categorical so we can iterate through it where it makes sense. (NOTE: We were going to do the same for Strings, but there are only two components so it may not make sense, may need to revisit.  Similarly type sensitive methods will have to do manual / hard-coded removal i.e. `attach`, `unregister_<>_by_name`)
- is_registered() will be updated to handle cases where not all of the required parts are registered.
- Regarding Issue #729-GroupBy sub-section: We are going to skip GroupBy registration functions in favor of a Complex object implementation approach on the server.

glitch:
(Issue #761)  Adding registration suite of functions for Categorical
  This adds the following methods/functions to Categorical
    - register
    - is_registered
    - attach (static)
    - unregister
    - unregister_categorical_by_name (static)
  Unit tests were added to test and document various usage.
  Changes to other classes were also made to support Categorical registration.

This is required to support registration changes for GroupBy (part of Issue #729 )

NOTE: The static unregister_*_by_name methods have a warning denoting the fact that object deletion messages will be sent to the server because the objects will go out of scope.  This is _not_ a change introduced in this PR, rather it was a side-effect introduced by the switch to in-place object registration.  See the included unit test in `registration_test.py:test_attach_weak_binding` which illustrates the current behavior in `pdarrayclass`.

Feedback welcomed ;)

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1,15 +1,16 @@
 from __future__ import annotations
-from typing import cast, List, Union
+from typing import cast, Dict, List, Union, Tuple, Type
 import numpy as np # type: ignore
 from typeguard import typechecked
 from arkouda.strings import Strings
-from arkouda.pdarrayclass import pdarray
+from arkouda.pdarrayclass import pdarray, RegistrationError
 from arkouda.groupbyclass import GroupBy
 from arkouda.pdarraycreation import zeros, zeros_like, arange
 from arkouda.dtypes import resolve_scalar_dtype, str_scalars
 from arkouda.dtypes import int64 as akint64
 from arkouda.sorting import argsort
 from arkouda.pdarraysetops import concatenate, in1d
+from arkouda.logger import getArkoudaLogger
 
 __all__ = ['Categorical']
 
@@ -46,11 +47,14 @@ class Categorical:
 
     """
     BinOps = frozenset(["==", "!="])
+    RegisterableComponents : Dict[str, Type[Union[pdarray,Strings]]] =\
+        {"categories":Strings, "codes":pdarray, "permutation":pdarray, "segments":pdarray}
     objtype = "category"
     permutation = None
     segments = None
     
     def __init__(self, values, **kwargs) -> None:
+        self.logger = getArkoudaLogger(name=__class__.__name__)  # type: ignore
         if 'codes' in kwargs and 'categories' in kwargs:
             # This initialization is called by Categorical.from_codes()
             # The values arg is ignored
@@ -77,6 +81,7 @@ class Categorical:
         self.nlevels = self.categories.size
         self.ndim = self.codes.ndim
         self.shape = self.codes.shape
+        self.name : Union[str,None] = None
 
     @classmethod
     @typechecked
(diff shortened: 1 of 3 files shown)


arkouda/categorical.py
```diff
@@ -529,3 +534,197 @@ def concatenate(self, others : List[Categorical], ordered : bool=True) -> Catego
...
+        for k,v in Categorical.RegisterableComponents.items():
+            parts[k] = v.attach(f"{user_defined_name}.{k}")
+        # Now call constructor with kwargs and hand it the unpacked dictionary object
+        c = Categorical(None, **parts)
+        c.name = user_defined_name # Update our name
+        return c
+
+    @staticmethod
+    @typechecked
+    def unregister_categorical_by_name(user_defined_name:str) -> None:
```
mhmerrill, line 700: @glitch should we change the `pdarray` interface to match this?

glitch: This is a good question,  right now `pdarrayclass.unregister_pdarray(...) ` takes either a String name or a `pdarray` object itself as an argument and currently the instance method `pda_obj.unregister()` delegates to the static method, so there would be some wiring changes.

I didn't want to run too far afield in this PR but I'm happy to make the change.  If you're only thinking of the naming for consistency I could do that instead.  Let me know what you think.

arkouda/categorical.py
```diff
@@ -529,3 +534,197 @@ def concatenate(self, others : List[Categorical], ordered : bool=True) -> Catego
...
+
+        Notes
+        -----
+        Objects registered with the server are immune to deletion until
+        they are unregistered.
+        """
+        components = self._get_vars_with_register()
+        if components:
+            for k, v in components:
+                self.logger.debug(f"Registering attribute {k} with name {user_defined_name}_{k}")
```
mhmerrill, line 588: i noticed you use a `.` on the next line vs a `_` to separate the field name, we should be consistent. Strings uses a `_` I think

arkouda/categorical.py
mhmerrill: I think we should maintain consistency in the interface.

arkouda/categorical.py
glitch: We chatted and decided to use the dot character for all sub-components (so changes to Categorical & Strings).

Note: I updated the #761 ticket with the items we discussed.

arkouda/categorical.py
```diff
@@ -529,3 +534,197 @@ def concatenate(self, others : List[Categorical], ordered : bool=True) -> Catego
...
+        -----
+        Objects registered with the server are immune to deletion until
+        they are unregistered.
+        """
+        reg:bool
+        components = self._get_vars_with_register()
+        if components:
+            registered = [k for k, v in components if v.is_registered()]
+            if registered and len(registered) == len(components):  # everything is registered
+                reg = True
```
pierce314159, line 654: no change requested but I was wondering if something like this would work here
```python
registered = [v.is_registered() for k, v in components.items()]
if registered and np.all(registered):
    reg = True
```
It doesn't change the functionality but using `np.all` seems more straightforward and pythonic to me than the length comparison.

pierce314159 (approved):

arkouda/categorical.py
glitch: I see what you are saying and I think you are correct.  However, this piece of code is being pitched.

hokiegeek2: 👍

arkouda/categorical.py
```diff
@@ -529,3 +534,197 @@ def concatenate(self, others : List[Categorical], ordered : bool=True) -> Catego
...
+        c.name = user_defined_name # Update our name
+        return c
+
+    @staticmethod
+    @typechecked
+    def unregister_categorical_by_name(user_defined_name:str) -> None:
+        """
+        Function to unregister Categorical object by name which was registered
+        with the arkouda server via register()
+        WARNING: This will cause the underlying object to go out-of-scope issuing a deletion
```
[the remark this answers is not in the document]
glitch: This is a weird issue because I'm doing an attach which makes a new object  It's bound to the arkouda_server with a string name.  It invokes unregister (make it deletable on the arkouda_server) and then goes out of scope when the function returns.  That triggers a __del__() operation on this specific instance and it tells the arkouda_server to delete the server object associated with the name.  This means any other objects bound to it are now pointing to something which no longer exists on the server.

@mhmerrill  and I had a discussion on what to do about this PR  and I'm in the process of refactoring it, so the scope issue is now moot.

arkouda/categorical.py
```diff
@@ -529,3 +534,197 @@ def concatenate(self, others : List[Categorical], ordered : bool=True) -> Catego
...
+        See also
+        --------
+        register, attach, unregister_categorical_by_name, is_registered
+
+        Notes
+        -----
+        Objects registered with the server are immune to deletion until
+        they are unregistered.
+        """
+        if not self.name:
```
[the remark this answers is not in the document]
glitch: Based on the semantics of pdarray, you're only supposed to get a name through invoking register on an object.  I'm really just trying to guard against someone calling unregister on an object that is not currently registered

glitch:
Note for all reviewers, @mhmerrill and I had a long chat about what to do here and the scope has changed a bit.  I've updated ticket #761 with the requested changes.

arkouda/categorical.py
```diff
@@ -77,6 +81,7 @@ def __init__(self, values, **kwargs) -> None:
         self.nlevels = self.categories.size
         self.ndim = self.codes.ndim
         self.shape = self.codes.shape
+        self.name : Union[str,None] = None
```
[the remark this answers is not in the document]
glitch: Oh... that's nice thanks!

arkouda/categorical.py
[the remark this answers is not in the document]
hokiegeek2: 👍

arkouda/categorical.py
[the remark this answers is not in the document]
hokiegeek2: Ah, I see, cool 👍

glitch:
Closing / Declining PR for redesign.

Issue #767, opened 2021-04-20
Strings.save() to 1 locale fails for multi-locale Arkouda


hokiegeek2:
If Arkouda is built to support multi locales, Strings.save() to 1 locale fails:

![image](https://user-images.githubusercontent.com/10785153/115386576-949f2480-a1a7-11eb-85b1-b33fa9943797.png)

Note: if Arkouda is built in single locale mode, Strings.save works correctly:

![image](https://user-images.githubusercontent.com/10785153/115386714-bd271e80-a1a7-11eb-999f-754c34dc82cb.png)

hokiegeek2:
Fixed:

![image](https://user-images.githubusercontent.com/10785153/115390771-a3d4a100-a1ac-11eb-9423-3802f8e9f723.png)

![image](https://user-images.githubusercontent.com/10785153/115390999-ee561d80-a1ac-11eb-8433-cbe5e849b606.png)

hokiegeek2:
[PR](#769) submitted

hokiegeek2:
PR accepted and merged, issue resolved

PR #769, opened 2021-04-20
Fixes Strings.save fails for multi-locale Arkouda where nl=1

hokiegeek2:
This PR adds logic to confirm the numLocales is > 1 prior to applying shuffle logic for final locale #767


mhmerrill (approved):
looks ok to me.

PR #772, opened 2021-04-21
Closes #771: Update deleteMsg for registered objects

### LINKED ISSUE #771: Update deleteMsg for registered objects
The response message for `deleteMsg` is misleading for registered objects. After `deleteEntry` is called with a registered object, the object is skipped for deletion, but `repMsg` is still `deleted name`.

I think `deleteEntry` should return a boolean indicating successful deletion or failure due to the object being registered. Then this boolean will be used to determine the appropriate `repMsg`.

Also, would it make sense to throw an `UnknownSymbolError` when `deleteEntry` is called with a name that isn't in the symbol table?

https://github.com/mhmerrill/arkouda/blob/f9edd8e2b423ea2f4def7f044bff2edb84567c97/src/MsgProcessing.chpl#L100-L103

https://github.com/mhmerrill/arkouda/blob/f9edd8e2b423ea2f4def7f044bff2edb84567c97/src/MultiTypeSymbolTable.chpl#L199-L218

pierce314159:
This PR (Issue #771 )
- Updates `repMsg` when `deleteMsg` is called on a registered object
- Adds bool return for `deleteEntry` indicating successful or failed deletion
- Raises `UnknownSymbolError` when `deleteEntry` is called on an object not found in the Symbol Table
- Adds a test for this new behavior in `registration_test.py`
- Adds `ak.list_symbol_table`  to functions available to user

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -3,7 +3,7 @@ from typing import cast, Sequence, Union, List
 from typeguard import typechecked
 import json, struct
 import numpy as np # type: ignore
-from arkouda.client import generic_msg, RegisteredSymbols, EmptyRegistry, EmptySymbolTable
+from arkouda.client import generic_msg, RegisteredSymbols, AllSymbols, EmptyRegistry, EmptySymbolTable
 from arkouda.dtypes import dtype, DTypes, resolve_scalar_dtype, \
      structDtypeCodes, translate_np_dtype, NUMBER_FORMAT_STRINGS, \
      int_scalars, numeric_scalars, numpy_scalars
@@ -13,8 +13,8 @@ from arkouda.dtypes import bool as npbool
 from arkouda.logger import getArkoudaLogger
 import builtins
 
-__all__ = ["pdarray", "info", "clear", "any", "all", "is_sorted", "list_registry", "sum", "prod",
-           "min", "max", "argmin", "argmax", "mean", "var", "std", "mink", 
+__all__ = ["pdarray", "info", "clear", "any", "all", "is_sorted", "list_registry", "list_symbol_table",
+           "sum", "prod", "min", "max", "argmin", "argmax", "mean", "var", "std", "mink",
            "maxk", "argmink", "argmaxk", "attach_pdarray",
            "unregister_pdarray", "RegistrationError"]
 
@@ -1216,6 +1216,35 @@ def list_registry() -> List[str]:
                 registered_list.append(name)
 
     return registered_list
+
+def list_symbol_table() -> List[str]:
+    """
+    Return a list containing the names of all objects in the symbol table
+
+    Parameters
+    ----------
+    None
+
+    Returns
+    -------
+    list
+        List of all object names in the symbol table
+
+    Raises
+    ------
+    RuntimeError
+        Raised if there's a server-side error thrown
+    """
+    symbol_list: List[str] = []
+
+    if info(AllSymbols) not in [EmptyRegistry, EmptySymbolTable]:
+        symbol: str
+        for symbol in filter(None, info(AllSymbols).split('\n')):
+            if symbol is not None:
+                name = symbol.split()[0].split(':')[1].replace('"', '')
+                symbol_list.append(name)
+
+    return symbol_list
 
 def clear() -> None:
     """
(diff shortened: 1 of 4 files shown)


src/MsgProcessing.chpl
```diff
@@ -98,9 +98,12 @@ module MsgProcessing
...
                                      "cmd: %s array: %s".format(cmd,st.attrib(name)));
         // delete entry from symbol table
-        st.deleteEntry(name);
-
-        repMsg =  "deleted %s".format(name);      
+        if st.deleteEntry(name) {
+            repMsg = "deleted %s".format(name);
+        }
+        else {
+            repMsg = "failed to delete registered symbol %s".format(name);
```
hokiegeek2, line 105: Just a nit...I dunno about the repMsg -> the delete request doesn't fail in the case; it's not executed per the registration biz rules.

src/MultiTypeSymbolTable.chpl
```diff
@@ -195,16 +194,20 @@ module MultiTypeSymbolTable
 
         :arg name: name of the array
         :type name: string
+
+        :returns: bool indicating whether deletion was successful
```
hokiegeek2, line 198: As above, I recommend changing the wording to "bool indicating whether the deletion occurred"

hokiegeek2 (approved):

src/MsgProcessing.chpl
pierce314159: Ah, that makes sense!

pierce314159:
I added the function `list_symbol_table` as a helper function for `test_delete_registered` . It seems like it would be a nice function to have function as a user too. I can add it as `ak.list_symbol_table` if people think that's a good idea

tests/registration_test.py
```diff
@@ -306,6 +306,44 @@ def test_string_is_registered(self):
...
+
+        # RuntimeError when calling delete on an object not in the symbol table
+        with self.assertRaises(RuntimeError):
+            ak.client.generic_msg(cmd='delete', args='not_in_table')
+
+def list_symbol_table() -> List[str]:
+    symbol_list: List[str] = []
+
+    if ak.info(ak.AllSymbols) not in [ak.EmptyRegistry, ak.EmptySymbolTable]:
+        bject: str
```
pierce314159, line 340: this is a typo, should be `object: str`. I'm a bit surprised this didn't cause issues. I'll fix this along with @hokiegeek2s comments

pierce314159:
> I added the function `list_symbol_table` as a helper function for `test_delete_registered` . It seems like it would be a nice function to have function as a user too. I can add it as `ak.list_symbol_table` if people think that's a good idea

I went ahead and made this change. My logic is the information is already available in `ak.info(ak.AllSymbols)`, so it would just be convenient for the user if we parse the return string for them

arkouda/pdarrayclass.py
```diff
@@ -1217,6 +1217,35 @@ def list_registry() -> List[str]:
...
+        List of all object names in the symbol table
+
+    Raises
+    ------
+    RuntimeError
+        Raised if there's a server-side error thrown
+    """
+    symbol_list: List[str] = []
+
+    if info(AllSymbols) not in [EmptyRegistry, EmptySymbolTable]:
```
glitch, line 1240: Since it's a network call, It's probably a good idea to store the result of `info(AllSymbols)` so you don't need to call it again in the for loop below.

arkouda/pdarrayclass.py
```diff
@@ -1217,6 +1217,35 @@ def list_registry() -> List[str]:
...
+        symbol: str
+        for symbol in filter(None, info(AllSymbols).split('\n')):
+            if symbol is not None:
```
glitch, line 1243: You shouldn't need this guard condition since you are filtering out `None` in the loop generator.

glitch (commented):
Looks pretty good, two minor improvement suggestions.

arkouda/pdarrayclass.py
pierce314159: we chatted offline. I think I originally had just the filter and mypy was not happy. But there's no point in double checking, so I'll remove the filter and leave the guard condition

glitch (approved):

mhmerrill (approved):
looks ok to me also

Issue #736, opened 2021-04-08
Add `ak.isna()` function


reuster986:
In the course of putting together an Arkouda tutorial, I realized we have no way to check whether a `float64` is `nan`. I believe it will require a new server message to call a Chapel function for detecting `nan`s.

mhmerrill:
@reuster986 do you want this to behave like `numpy.isnan()`?

mhmerrill:
@reuster986 do we need `numpy.isnat()` also?

reuster986:
I was thinking like `pandas.isna()`, which encompasses both. But I am not super familiar with the differences between numpy and pandas in this area.

mhmerrill:
@reuster986 so from this I am assuming that the requirement for this is to mirror the pandas functionality.

here is the link to the pandas doc about `pandas.isna()`
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.isna.html

mhmerrill:
@reuster986 do we also need a function to check for `inf` values?

reuster986:
Yes, let's mirror the pandas functionality. I'd say `isinf` is probably helpful, but less of a priority than `isna`.

mhmerrill:
Ok, we should add this functionality to the `efunc` message flow in the server since it tests each element and returns a `bool` `pdarray`.

glitch:
My understanding is that numpy only supports NaN in float-based arrays as they use the IEEE 754 standard; [numpy_doc](https://numpy.org/doc/stable/reference/constants.html#numpy.NAN) for reference.

Due to this, It looks like if you create a Pandas series from a list containing a None value (i.e. `pd.Series([1, 2, None, 3, 4])`) it gets coerced into a float64.

Given the pandas and numpy no-NaN-in-integers thing, my assumption is that you can't make an int-based array in Arkouda/Chapel with a missing value in it.  Is that reasoning correct?

If this is true, it seems like an `isna `/`isnan` function for Arkouda can only be used on a float64 array.  (We could add it for int64, but it would always return an array of false values since there is no way you can put a NaN value in an integer based array).

Does this sound right?

Note:
Reading up on the Pandas documentation it looks like they are [working around it](https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html) by making their own _Int64_ type (notice the capital I) capable of holding a pd.NAN, but we still wouldn't have a way of constructing an int64 arkouda pdarray with a missing value in it.

PR #776, opened 2021-04-26
Issue #736: Adding isna functionality for float based arrays

glitch:
Issue #736:  Adding isna functionality for float based arrays

### Changes:
Adding isna to efunc on server side and wiring isna capability on client side.  isna(pdarray) returns T/F pdarrary for
NaN detection on **_float64_** arrays.  Int-based is not supported at this time.

### Limitations
My understanding is that numpy only supports NaN in float-based arrays as they use the IEEE 754 standard; [numpy_doc](https://numpy.org/doc/stable/reference/constants.html#numpy.NAN) for reference.

If you create a Pandas series from a list containing a None value (i.e. `pd.Series([1, 2, None, 3, 4])`) it gets coerced into a float64.  (Note: the Pandas team does have a custom implementation of an _Int64_ array (capital I) where they have added support for their own type of NaN for integers but I think they are using object wrappers around ints instead of primitives, so something similar may need to b implemented for Arkouda/Chapel to support something like this?)

It doesn't look like you can make an integer based array in Arkouda/Chapel containing missing values (you also can't assign a `nil` value to an int in Chapel).  Given these limitations I only added the `isna` function for float based arrays.  While we could add it to the API for int64 it won't ever return true because NaN values would likely have been coerced to a default (0, -1?) or the array would have been coerced to a Float, so currently I let it traverse the error route if you call isna on an integer based array in Arkouda.

### Draft
I'm marking this PR as a draft until we can have a discussion or confirm my assumptions about the float vs. int thing in Chapel regarding NaN / missing values.

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -13,7 +13,7 @@ from arkouda.strings import Strings
 Categorical = ForwardRef('Categorical')
 
 __all__ = ["cast", "abs", "log", "exp", "cumsum", "cumprod", "sin", "cos", 
-           "where", "histogram", "value_counts"]    
+           "where", "histogram", "value_counts", "isna"]
 
 @typechecked
 def cast(pda : Union[pdarray, Strings], dt: Union[np.dtype,str]) -> Union[pdarray, Strings]:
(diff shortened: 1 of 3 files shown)


arkouda/numeric.py
```diff
@@ -508,3 +508,28 @@ def value_counts(pda : pdarray) -> Union[Categorical, # type: ignore
     (array([0, 2, 4]), array([3, 2, 1]))
     """
     return unique(pda, return_counts=True)
+
+
+@typechecked
+def isna(pda : pdarray) -> pdarray:
```
mhmerrill, line 514: @glitch the more I think about this the more I think that the first thing is to use the numpy interface for `np.isnan()` for `float64`, then later we can build up the pandas interface on top of these functions.

glitch: @mhmerrill RE: alter `int64` and `strings` yes, I was thinking along similar lines.  We'd almost have to encapsulate int64 in our own object which would be something like `class Int64(value:int64, missing:bool)`.  I haven't looked at their implementation but I think this is roughly what Pandas ended up doing.

Regarding float64 do you want me to change the `isna` to `isnan`?

mhmerrill: yes change `isna` to `isnan`

mhmerrill (approved):
this looks good

PR #779, opened 2021-04-29
763 improve performance of tohdf msg with strings v1

hokiegeek2:
This PR does the following:

1. improves the performance of Strings tohdfMsg by refactoring the generateValuesMetadata by using the Strings segments array to generate the needed bytes array metadata, obviating the need for costly list generation. 
2. refactors the Strings.save method to send both the bytes and segments array names to enable (1)

Testing a 10 million length Strings save on the corp AWS instance with a 3-node Arkouda cluster in smp mode demonstrates that (1) decreases the generateValuesMetadata runtime from ~ 250 seconds to ~ 1 second and overall runtime down to ~ 36 seconds.

note: a second version of this work that should further improve tohdfMsg performance is currently under development.

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -94,7 +94,7 @@ class Strings:
             self.shape = self.offsets.shape
         except Exception as e:
             raise ValueError(e)   
-        self.offsets.register('{}_offsets'.format(self.bytes.name))
+
         self.dtype = npstr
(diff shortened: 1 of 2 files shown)


reuster986 (approved):
I don't have a great grasp of the logic, but I can certainly see how this is faster, and I don't see any red flags.

PR #781, opened 2021-04-29
Closes #762: Fixes off by one error in Segemented array peel method.

### LINKED ISSUE #762: Off-by-one error in Strings.peel
The inner for loops in the following code should iterate over `0..#(len-1)`, not `0..#len`:

https://github.com/mhmerrill/arkouda/blob/5c4d04a15ca6fb150223a789bc43a70e393ac34c/src/SegmentedArray.chpl#L619-L630

Background: the `peel` function splits strings into left and right portions based on a delimiter. After the logic of finding the split point is all complete, this code copies the bytes of the left and right portions into their respective destination arrays. The `leftLengths` and `rightLengths` arrays are supplying the full lengths of the destination buffers, *including the null byte*, i.e. one greater than the actual length of the substring to be copied. Thus, when the code copies `#len` bytes from the source buffer, it is grabbing an extra byte from the source string and overwriting the destination's null terminator with it. This results in strings that are not null-terminated and misbehave in weird and hard-to-find ways.

As part of this issue, we should consider adding validation on the `ak.Strings` class that checks for null terminators, and/or a CI test that checks for null-terminators after a `peel` operation. I am hopeful that resolving this issue will also address the sporadic CI failures for peel/stick and allow us to move this from experimental to fully supported functionality.

glitch:
Closes #762
  - Fixes off by one error in Segemented array peel method.
  -  Updates Strings.peel unit test to check that bytes pdarrays are appropriately structured with respect to null bytes.

### DIFF
--- a/tests/string_test.py
+++ b/tests/string_test.py
@@ -457,14 +457,14 @@ class StringTest(ArkoudaTest):
         # This mimics what should be stored server-side in the strings.bytes pdarray
         expected_series_dec = convert_to_ord(series.to_list())
         actual_dec = pda.bytes.to_ndarray().tolist()
-        self.assertEqual(expected_series_dec, actual_dec)
+        self.assertListEqual(expected_series_dec, actual_dec)
 
         # Now perform the peel and verify
         a, b = pda.peel(":")
         expected_a = convert_to_ord(["k1", "k2", "k3", ""])
         expected_b = convert_to_ord(["v1", "v2", "v3", "no_colon"])
-        self.assertEqual(expected_a, a.bytes.to_ndarray().tolist())
-        self.assertEqual(expected_b, b.bytes.to_ndarray().tolist())
+        self.assertListEqual(expected_a, a.bytes.to_ndarray().tolist())
+        self.assertListEqual(expected_b, b.bytes.to_ndarray().tolist())
 
     def test_stick(self):
         run_test_stick(self.strings, self.test_strings, self.base_words, 
--- a/tests/string_test.py
+++ b/tests/string_test.py
@@ -1,5 +1,6 @@
-from typing import Tuple
+from typing import List, Tuple
 import numpy as np
+import pandas as pd
 from collections import Counter
 from context import arkouda as ak
 from base_test import ArkoudaTest
@@ -9,7 +10,12 @@ UNIQUE = N//4
 
 def compare_strings(a, b):
     return all(x == y for x, y in zip(a, b))
-  
+
+
+def convert_to_ord(s:List[str]) -> List[int]:
+    return [ord(i) for i in "\x00".join(s)] + [ord("\x00")]
+
+
 errors = False
 
 def run_test_argsort(strings, test_strings, cat):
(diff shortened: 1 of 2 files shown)


tests/string_test.py
```diff
@@ -441,7 +447,24 @@ def test_peel(self):
...
+
+        # Run a test with a specific set of strings to verify strings.bytes matches expected output
+        series = pd.Series(["k1:v1", "k2:v2", "k3:v3", "no_colon"])
+        pda = ak.from_series(series, "string")
+
+        # Convert Pandas series of strings into a byte array where each string is terminated by a null byte.
+        # This mimics what should be stored server-side in the strings.bytes pdarray
+        expected_series_dec = convert_to_ord(series.to_list())
+        actual_dec = pda.bytes.to_ndarray().tolist()
+        self.assertEqual(expected_series_dec, actual_dec)
```
reuster986, line 460: I'm not familiar enough with pytest to know this, but does `self.assertEqual(list1, list2)` actually do the element-wise comparison and check that they are all `True`? To me, I would be wary that it is just calling `assert list1 == list2` under the hood, which would produce the wrong result (or an error).

Elsewhere, I have been using `self.assertTrue((arr1 == arr2).all())` with either numpy arrays or arkouda arrays to ensure all the element-wise comparisons are happening, but if `assertEqual` does the right thing with python lists, I would be interested to know.

glitch: That's a good question, I'm fairly certain python's unittest.TestCase (where the assertion methods are coming from) really default down to the operator methods, so assertEquals(a, b) delegates to a == b which is an equivalence test vs `a is b` which is same ref.

I ran the following in a python console to double check
[Gist here](https://gist.github.com/glitch/20db958b8ae5d357ed90cc0586fb09a6)
```
x = ["a", "b", "c"]
y = ["a", "b", "c"]
z = ["c", "b", "a"]
x == y
> True
x is y
> False
x == z
> False

import unittest
t = unittest.TestCase()
t.assertEqual(x, y)  # passes
t.assertTrue(x is y) # fails
t.assertEqual(x, z)  # fails

# There is also a assertListEqual(a, b) which may be a bit more explicit
t.assertListEqual(x, y)  # passes
t.assertListEqual(x, z)  # fails
```

So while it's testing the correct thing, I think I'll push an update to improve clarity with `assertListEquals`

glitch: Note: I don't have a link to the python src handy, but for 3.7.9 unittest / case.py _baseAssertEqual the code is this
```python
    def _baseAssertEqual(self, first, second, msg=None):
        """The default assertEqual implementation, not type specific."""
        if not first == second:
            standardMsg = '%s != %s' % _common_shorten_repr(first, second)
            msg = self._formatMessage(msg, standardMsg)
            raise self.failureException(msg)
```
so it is basically python's `==` function, we should be good.

reuster986: Ah, ok. I guess I had confused the behavior of python lists with numpy arrays and had forgotten that `a == b` will work with the former. 

Side note: if `a` and `b` are numpy arrays (or arkouda arrays), `a == b` will return a bool array the same length as `a` and `b` with the result of each element-wise comparison, instead of a single bool value. So calling `assertEqual(a, b)` will raise a ValueError error saying "the truth value of an array is ambiguous" or something like that. Not a problem for python lists, but something to watch out for when writing unit tests on numpy/arkouda arrays.

reuster986 (approved):

PR #785, opened 2021-05-04
Closes #777: Re-enable GroupBy(Strings).aggregate()

### LINKED ISSUE #777: Re-enable GroupBy(Strings).aggregate()
We used to be able to do the following:
```python
s = ak.array(['a', 'b', 'a', 'b', 'c'])
i = ak.arange(s.size)
bys = ak.GroupBy(s)
us, ui = bys.nunique(i)
```

But at some point, the last line started throwing a `TypeError`, seemingly because type checking is too restrictive and does not allow `ak.Strings` to be returned from `GroupBy.aggregate` (which is called by `nunique`).

pierce314159:
This PR (Issue #777) :
- Adds `Strings` to aggregate function return types
- Updates example in the `Groupby.nunique` doc string
- Adds `test_aggregate_strings` to `groupby_test` suite

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -226,8 +226,8 @@ class GroupBy:
         return self.unique_keys, create_pdarray(repMsg)
     
     @typechecked
-    def aggregate(self, values : pdarray, operator : str, skipna : bool=True) \
-                    -> Tuple[Union[pdarray,List[Union[pdarray,Strings]]],pdarray]:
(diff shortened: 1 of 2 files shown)


pierce314159:
I updated the example in the `nunique` doc string based on an example that @glitch showed me because it made `nunique` click a bit better for me. If anyone has suggestions on how to make the example better or want me to change it back, please let me know

reuster986 (approved):
looks great!

PR #790, opened 2021-05-06
Closes #788: Reducing duplicative code for datastr and datarepr in MultiTypeSymbolTable

### LINKED ISSUE #788: Remove duplicative code for datastr/datarepr in MultiTypeSymbolTable
MultiTypeSymbolTable's `datastr` and `datarepr` both contain repetitive code both internally and between the two procs.  We should be able to remove some of the duplicative code and make this section a bit more concise.

(Note: I'm going to work on this as a way to get more experience working in Chapel.)

glitch:
Closes #788: Reducing duplicative code for datastr and datarepr in MultiTypeSymbolTable.
Added unit tests for __str__() and __repr__().

This PR adds a `__str__(...)` method to `GenSymEntry` & ` SymEntry` which can be used by both the `datastr` and `datarepr` methods.  It reduces the amount of code/logic which was previously duplicated between those two functions.

One possible down-side with this approach (mostly philosophical) is that it pushes behavior into the `SymEntry` class... although arguably a `toString` style method is within bounds of a Class's responsibility.

*This PR does _not_ change behavior only the underlying implementation.

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -51,6 +51,30 @@ module MultiTypeSymEntry
         inline proc toSymEntry(type etype) {
             return try! this :borrowed SymEntry(etype);
         }
+
+        /* 
+        Formats and returns data in this entry up to the specified threshold. 
+        Arrays of size less than threshold will be printed in their entirety. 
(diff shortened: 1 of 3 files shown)


mhmerrill (approved):
fun!, fixing some of the first Chapel code I wrote ;-)

PR #792, opened 2021-05-06
Closes #791: Remove duplicated check table in MultiTypeSymbolTable

### LINKED ISSUE #791: Remove duplicated check login in MultiTypeSymbolTable
The `check` function in `MultiTypeSymbolTable.chpl` checks if a symbol is in `tab` and throws a `UnknownSymbolError` if not.
https://github.com/mhmerrill/arkouda/blob/62611aec7552038276209124b83ab15e7cc50b9f/src/MultiTypeSymbolTable.chpl#L267-L280

This logic is duplicated in other functions in `MultiTypeSymbolTable.chpl` . A good example is in `lookup`
https://github.com/mhmerrill/arkouda/blob/62611aec7552038276209124b83ab15e7cc50b9f/src/MultiTypeSymbolTable.chpl#L249-L262

There are also places where very similar logic is being use to check if a symbol is in the registry which may be able to be handled by the `check` function

I'll update the functions of `MultiTypeSymbolTable` to call the existing `check` function and extend `check` to handle the registry as well

pierce314159:
This PR (Closes #791)
- Changes name of `check` in `MultiTypeSymbolTable` to `checkTable`
- Updates functions in `MultiTypeSymbolTable.chpl` to use the existing `checkTable` function to check if a symbol is in the symbol table
- Replaces references to `check` in `SegmentedMsg.chpl` with `checkTable`

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -36,18 +36,7 @@ module MultiTypeSymbolTable
         }
 
         proc regName(name: string, userDefinedName: string) throws {
-
-            // check to see if name is defined
-            if (!tab.contains(name)) {
-                mtLogger.error(getModuleName(),getRoutineName(),getLineNumber(),
-                                     "regName: undefined symbol: %s".format(name));
-                throw getErrorWithContext(
-                                   msg=unknownSymbolError("regName", name),
-                                   lineNumber=getLineNumber(),
-                                   routineName=getRoutineName(),
-                                   moduleName=getModuleName(),
-                                   errorClass="UnknownSymbolError");
-            }
+            checkTable(name, "regName");
 
             // check to see if userDefinedName is already defined, with in-place modification, this will be an error
             if (registry.contains(userDefinedName)) {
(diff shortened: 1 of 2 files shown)


src/MultiTypeSymbolTable.chpl
```diff
@@ -199,32 +176,17 @@ module MultiTypeSymbolTable
...
+            if !registry.contains(name) {
+                mtLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                                       "Deleting unregistered entry: %s".format(name)); 
+                tab.remove(name);
+                return true;
             } else {
-                if registry.contains(name) {
-                    mtLogger.error(getModuleName(),getRoutineName(),getLineNumber(),
-                                     "Registered entry is not in SymTab: %s".format(name));
-                } else {
```
pierce314159, line 217: This condition in `deleteEntry` is one area that I was unsure about. This case of "registered object that's not in the symbol table" is no longer captured and logged since the generic "in symbol table check" is done before checking in registry. I could switch the order and add the logging back in if people want. My thinking was this shouldn't happen and the code is cleaner without multiple cases

glitch: Just to make sure I understand this correctly:  In both implementations if the item is not in the SymTab we're still going to throw an error.  So the difference here is whether we log additional information about the missing item (whether it was registered or not).

So my (rhetorical) question is what would we do with that additional information.  Is it going to help us determine what happened / what led to the item being missing?  I'm not sure, I think I'm leaning towards no it's not really helpful, but I'm still a bit of a newb here.

If we decided that extra logging information is useful, I'd probably wrap the `checkTable` call in a try/catch, log the registry info and then re-throw the error.

glitch (approved):
Overall I think this looks pretty good and cleans up the code quite a bit.

I think others may need to weigh in on whether we should keep the additional logging in the deleteEntry section, but if it isn't _actionable_ information then I lean towards no, we don't need it.  (See comments in that section.)

PR #796, opened 2021-05-11
Sort-based in1d for Strings

### LINKED ISSUE #731: Improve performance of `in1d` with large string arrays
The current implementation of `segIn1d` is slow when the second array is large. Currently, it constructs a local associative domain from the second array and tests each element of the first array for membership in the domain. Constructing the local domain is probably what is scaling poorly. The analogous `in1d` function for integers switches to a sort-based algorithm for large arrays, so I suspect a similar strategy would be useful here.

reuster986:
Closes #731 

This PR adds a case to in1d with Strings that uses a sort-based strategy when the second "test" array is large. This should improve scaling on distributed systems. 

The threshold for when this case is activated can be controlled at the command line with `--SegmentedArray.in1dAssocSortThreshold`. The current default is `10**6`, which was chosen by a cursory exploration with a multi-locale build on a laptop. In the future, we should probably benchmark this operation.

The test for correctness of this functionality is in `test/UnitTestIn1d.chpl`, which forces the computation via both the associative and sort-based strategies and then ensures the results agree.

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -870,11 +870,11 @@ module SegmentedArray {
     return truth;
   }
 
-  private config const in1dAssocSortThreshold = 10**4;
+  private config const in1dAssocSortThreshold = 10**6;
 
   /* Test array of strings for membership in another array (set) of strings. Returns
      a boolean vector the same size as the first array. */
-  proc in1d(mainStr: SegString, testStr: SegString, invert=false) throws where useHash {
+  proc in1d(mainStr: SegString, testStr: SegString, invert=false, forceSort=false) throws where useHash {
     var truth: [mainStr.offsets.aD] bool;
     // Early exit for zero-length result
     if (mainStr.size == 0) {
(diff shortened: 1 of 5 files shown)


reuster986:
Looks like the docs/readthedocs.org check is hung, but I didn't change anything about the docs. Anyone know how to rerun that check?

glitch:
> Looks like the docs/readthedocs.org check is hung, but I didn't change anything about the docs. Anyone know how to rerun that check?

Oh, this looks like it broke with the Bears-R-Us repo move.  Since we're generally moving away from read-the-docs anyway (in favor of github pages), we should probably disable that check.

reuster986:
@hokiegeek2 please review when you get a chance, thanks!

mhmerrill (commented):
@reuster986 Everything look ok to me. Do we need a python unit test for this?

reuster986:
> @reuster986 Everything look ok to me. Do we need a python unit test for this?

I thought about this, but in order to trigger the sort-based strategy, the test vector would have to be pretty large, and I worried that would take too much time in the CI, especially the chapel-gasnet-smp run. I haven't actually measured the time increase, but a while back I seem to recall having to scale back a 10**5 element groupby or something because it was taking too long, and this would probably take longer.

PR #797, opened 2021-05-11
Categorical equality bug fix

### LINKED ISSUE #789: `Categorical == Categorical` incorrect when categories differ
There is a bug in the code that implements comparisons between Categoricals, e.g. `a == b`. It works correctly when `a.categories` and `b.categories` are identical, but fails to check for this condition or handle the case when the categories differ.

https://github.com/mhmerrill/arkouda/blob/d46c8e5ae0eac1b0f677d10f6c25dc206c16bf68/arkouda/categorical.py#L207-L208

This should be special-cased: if the categories are the same, call the existing logic; otherwise, remap both codes to the union of the categories and compare.

reuster986:
Closes #789 

This PR fixes a bug in `Categorical == Categorical` which would give incorrect output when the categories differed. In this PR, that case is correctly handled, and the unit tests have been extended to cover it.

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -205,7 +205,32 @@ class Categorical:
             raise ValueError("Categorical {}: size mismatch {} {}".\
                              format(op, self.size, cast(Categorical,other).size))
         if isinstance(other, Categorical):
-            return self.codes._binop(other.codes, op)
+            if (self.categories.size == other.categories.size) and (self.categories == other.categories).all():
+                # Because categories are identical, codes can be compared directly
+                return self.codes._binop(other.codes, op)
+            else:
+                # Remap both codes to the union of categories
+                union = ak.union1d(self.categories, other.categories)
+                newinds = ak.arange(union.size)
+                # Inds of self.categories in unioned categories
+                selfnewinds = newinds(ak.in1d(union, self.categories))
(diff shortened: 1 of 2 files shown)


hokiegeek2 (approved):
Looks good to me

arkouda/categorical.py
```diff
@@ -205,7 +206,32 @@ def _binop(self, other : Union[Categorical,str_scalars], op : str_scalars) -> pd
             raise ValueError("Categorical {}: size mismatch {} {}".\
                              format(op, self.size, cast(Categorical,other).size))
         if isinstance(other, Categorical):
-            return self.codes._binop(other.codes, op)
+            if (self.categories.size == other.categories.size) and (self.categories == other.categories).all():
```
glitch, line 209: Is it worth it to add `self.categories.objtype == other.categories.objtype` as an additional clause either in first or second position (short circuit) to weed out `str` vs `pdarray` cases?  (Not sure if that's possible or not but categories looks like it could be either Strings or pdarray.)

hokiegeek2: @glitch good thought, but Categoricals can only be composed of Strings:

image

glitch: ah, gotcha, I see how it's wound together now.

glitch (approved):

Issue #763, opened 2021-04-16
Improve performance of `tohdfMsg` with strings


reuster986:
Writing `ak.Strings` to disk seems quite slow. In a recent session working with 65 million-element arrays on a single locale, writing out a `Strings` array to Lustre was taking 10-12 minutes, whereas an equal-length integer array was taking about 1 second. Since this is single locale, it's probably nothing to do with data movement or communication, and should be pretty easy to test. Probably the first thing to do is write a benchmark in Chapel and then start adding timing statements.

hokiegeek2:
Testing this now and I am seeing an array index out of bounds error with a single locale, opening up a new issue:

![image](https://user-images.githubusercontent.com/10785153/115380549-acbf7580-a1a0-11eb-9537-de8f464b763b.png)

The error occurs here:

![image](https://user-images.githubusercontent.com/10785153/115380758-e98b6c80-a1a0-11eb-9337-f217789a1c3d.png)

hokiegeek2:
Opened new [issue](https://github.com/mhmerrill/arkouda/issues/767) for bug

hokiegeek2:
Comparing the pdarrray vs Strings creation and tohdf times, it appears that, for the ranges I studied, the Strings creation times started off over two orders of magnitude longer than for int pdarrays, with the gap increasing to three orders of magnitude. For example, for 1 million ones and 1 million strings, the corresponding Strings object takes 352% longer than the pdarray, and the gap widens to over 1200% for 10 million length Strings vs pdarrays.

Similarly, the Strings save times start off over two orders of magnitude longer than pdarrays, with the gap widening to 3 orders of magnigute . For 1 million long pdarray and Strings objects, the Strings object takes roughly 800% longer, while for 10 million long pdarrays and Strings the gap increases to approximately 1800%.

![image](https://user-images.githubusercontent.com/10785153/115433203-486ad900-a1d5-11eb-86a1-d219b6da7837.png)

hokiegeek2:
Added bypass logic to the  [write1DDistStrings](https://github.com/hokiegeek2/arkouda/blob/fa67284791b6342bf84ad5babc6f7a95b1d992b9/src/GenSymIO.chpl#L1237) function if the numLocales == 1

![image](https://user-images.githubusercontent.com/10785153/115613101-fea7ee80-a2b9-11eb-9ea8-5684fe785dd7.png)

[...]

PR #798, opened 2021-05-12
763 improve performance of tohdf msg with strings v2

hokiegeek2:
This PR addresses #763, containing a second round of Strings.save() optimizations. Specifically, this PR contains changes that do the following:
1. Obviates the need for several list creation and iteration operations by leveraging locale-scoped array slicing
2. Builds up one char list per locale with the corresponding array slices
3. Generates a one segments list per locale once the final locale char list is generated
4. Provides detailed debug messages to detail the slice and shuffle ops needed to prepare for writing to disk
5. Updated and expanded chapel doc comments 

In testing, this set of changes, together with the [first round](https://github.com/Bears-R-Us/arkouda/pull/779), decreases the save runtime of a 10-million length Strings object on a 3-locale smp (one server) Arkouda cluster from 400 seconds down to ~ 10 seconds.

### DIFF
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -1148,13 +1148,13 @@ module GenSymIO {
          * The shuffleLeftIndices object, which is a globally-scoped PrivateSpace, 
          * contains indices for each locale that (1) specify the chars that can be 
          * shuffled left to complete the last string in the previous locale and (2)
-         * are used to remove the corresponding chars from the donor locale.  
+         * are used to remove the corresponding chars from the current, donor locale.  
          *
          * The shuffleRightIndices PrivateSpace is used in the special case 
          * where the majority of a large string spanning two locales is the sole
          * string on a locale; in this case, each index specifies the chars that 
          * can be shuffled right to start the string completed in the next locale
-         * and remove the corresponding chars from the donor locale 
+         * and remove the corresponding chars from the current, donor locale 
          *
          * The isSingleString PrivateSpace indicates whether each locale contains
          * chars corresponding to one string/string segment; this occurs if 
@@ -1164,6 +1164,9 @@ module GenSymIO {
          * The endsWithCompleteString PrivateSpace indicates whether the values
          * array for each locale ends with complete string, meaning that the last
          * character in the local slice is a null uint(8) char.
+         *
+         * The charArraySize PrivateSpace contains the size of char local slice
+         * corresponding to each locale.
          */
         var shuffleLeftIndices: [PrivateSpace] int;    
         var shuffleRightIndices: [PrivateSpace] int;
@@ -1181,8 +1184,8 @@ module GenSymIO {
         t1.start();
 
         coforall (loc, idx) in zip(A.targetLocales(), filenames.domain) 
-                with (ref shuffleLeftIndices, ref shuffleRightIndices, 
-                     ref isSingleString, ref endsWithCompleteString, ref charArraySize) do on loc {
+             with (ref shuffleLeftIndices, ref shuffleRightIndices, 
+                   ref isSingleString, ref endsWithCompleteString, ref charArraySize) do on loc {
              generateStringsMetadata(idx,shuffleLeftIndices, shuffleRightIndices, 
                           isSingleString, endsWithCompleteString, charArraySize, A, SA);
         }
@@ -1325,7 +1328,7 @@ module GenSymIO {
 
                 /* 
                  * To prepare for writing the charList to hdf5, do the following, if applicable:
-                 * 1. Remove the characters shuffled left to previous locale
+                 * 1. Remove the characters shuffled left to the previous locale
                  * 2. Remove the characters shuffled right to the next locale
                  * 3. If (2) does not apply, add null uint(8) char to end of the charList
                  */
@@ -1408,11 +1411,10 @@ module GenSymIO {
                      charList = new list(A.localSlice(locDom));
 
                      /*
-                      * If (1) this locale (idx) contains one string/string segment and (2) ends 
-                      * with the null uint(8) char, check to see if the shuffleRightIndex from 
-                      * the previous locale (idx-1) is > -1. If so, the chars following the 
-                      * shuffleRightIndex from the previous locale (idx-1) complete the one 
-                      * string/string segment within the current locale (idx). 
+                      * If this locale (idx) ends with the null uint(8) char, check to see if 
+                      * the shuffleRightIndex from the previous locale (idx-1) is > -1. If so, 
+                      * the chars following the shuffleRightIndex from the previous locale complete 
+                      * the one string/string segment within the current locale. 
                       */
                      if isSingleString[idx] && idx > 0 {
                          /*
@@ -1468,9 +1470,10 @@ module GenSymIO {
                      writeStringsToHdf(myFileID, idx, group, charList, segmentsList);
                   } else {
                       /*
-                       * Check to see if previous locale (idx-1) ends with a complete string.
+                       * Check to see if previous locale (idx-1) ends with a null character.
                        * If not, then the left shuffle slice of this locale was used to complete
-                       * the last string in the previous locale, so slice those chars.
+                       * the last string in the previous locale, so slice those chars from 
+                       * this locale and create a new, corresponding charList.
                        */
                       if !endsWithCompleteString(idx-1) {
                           var localStart = locDom.first;
@@ -1935,7 +1938,7 @@ module GenSymIO {
 
             /*
              * Normalize the first and last seg elements (make them zero-based) by
-             * subtracting the bytes domain first element. 
+             * subtracting the char domain first index element. 
 '            */
             var normalize = 0;
             if idx > 0 {
@@ -1967,7 +1970,7 @@ module GenSymIO {
             } else {
                 /*
                  * Since there is neither a shuffleLeftIndex nor a shuffleRightIndex for 
-                 * this locale, this local contains a single complete string.
+                 * this locale, this local contains a single, complete string.
                  */
                 isSingleString[idx] = true;
             }
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -1371,6 +1371,8 @@ module GenSymIO {
                                         idx,sliceList,leadingSlice, here.id));                         
                         charList.extend(sliceList); 
                         charArrayList.extend(leadingSlice);  
+                        gsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),"charList AFTER APPEND %t for locale %i".format(charList,idx));
+                        gsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),"charArrayList AFTER APPEND %t for locale %i".format(charArrayList,idx));
                         t1.stop();  
                         var elapsed = t1.elapsed();
                         gsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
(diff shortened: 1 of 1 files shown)


reuster986 (approved):
I can't say I fully follow all the logic, but the CI passes, and the speedup is great!

hokiegeek2:
> I can't say I fully follow all the logic, but the CI passes, and the speedup is great!

Yeah, there is a lot going on there, but the high-level logical flow is based upon the following (with some additional terminology translations for locales, domains, and coforall loops for Kyle and Pierce 👍)

1. Generate char array metadata for each locale indicating if the local slice (array chunk assigned to a locale) (1) composes one string, (2) ends with a null character (and therefore ends with a complete string) as well as (3) an index from which to shuffle characters left to the preceding locale, (4) an index from which to shuffle chars to the right to the succeeding locale, and (5) the length of the char array slice on each locale.
2. Generate complete strings at the end of each locale, either by a left shuffle of chars **from** the succeeding locale or a right shuffle of the chars from the current locale **to** the succeeding locale, using array slices configured by metadata from (1) to generate a char list on each locale.
3. Adjust each "donor" locale char list following shuffle left or shuffle right operations.
4. Generate a final segments list following all shuffle operations
5. Write the final char and segments lists to hdf5

A coforall loop is used (since all ops are done independently per locale, these operations can be done in parallel via a coforall loop) to start at locale 0 and iterate through all of the locales, generating the char and sequence lists from (1) the local char slices assigned to each locale and (2) the char array metadata described above to shuffle chars left/right and adjust to those shuffle ops so that each locale contains a char list that starts with and ends with a character sequence corresponding to one, complete string.

reuster986:
@hokiegeek2 thanks for the explanation!

glitch:
Ok, time for a really newb question.  Conceptually I think I understand what's going on here, it's basically trying to align the boundaries with the start and ending of complete strings so that we don't have strings splitting across locales and hence across files.

Now for the newbie part:  I thought the purpose of segmenting the strings was to evenly distribute the bytes across locales, why not just write the bytes directly to files (and the offsets to their own file).  Wouldn't that preserve the uniformity?  Or is the purpose to write out whole strings so they can be used by something else?  (i.e. another program/application)

Do we ever just write the bytes out uniformly?

src/GenSymIO.chpl
```diff
@@ -1908,16 +1929,16 @@ module GenSymIO {
...
                 }
                 lastSeg = segsArray.back();
             }
 
             /*
-             * Normalize the first and last seg elements (make then zero-based) by
-             * subtracting the bytes domain first element. 
+             * Normalize the first and last seg elements (make them zero-based) by
+             * subtracting the char domain first index element. 
 '            */
```
glitch, line 1942: Note: There is a dangling single quote here on line 1942 (it's in a comment so it doesn't really affect anything other than IDE code plugins).  Since there is some minor cleanup going on it's probably worth it to get that one to.

hokiegeek2:
> Ok, time for a really newb question. Conceptually I think I understand what's going on here, it's basically trying to align the boundaries with the start and ending of complete strings so that we don't have strings splitting across locales and hence across files.
> 
> Now for the newbie part: I thought the purpose of segmenting the strings was to evenly distribute the bytes across locales, why not just write the bytes directly to files (and the offsets to their own file). Wouldn't that preserve the uniformity? Or is the purpose to write out whole strings so they can be used by something else? (i.e. another program/application)
> 
> Do we ever just write the bytes out uniformly?

Excellent questions! The purpose of segmenting the bytes (chars) array is to know where each char sequence that equals a string starts and ends. We need to write out whole strings (no strings split between two locales) to ensure all read as well as load operations from files work correctly. For example, say a user is running a 10 locale arkouda cluster across 10 physical nodes, he/she may wish to load the strings from 2 files via the following command 

```
ak.read_hdf(filenames=['strings_LOCALE0000','strings_LOCALE0001'], dsetName='strings_array')
```

If we don't ensure strings are written in their respective entirety to a single locale, there will be 1..n logical strings that are broken up across multiple actual strings (e.g. ['Arkoud','a is awe','some'] vs ['Arkouda is awesome']. There may be other errors that would occur from read or load operations involving Strings if there 1..n logical strings spanning multiple locales, but this is the one scenario I am most familiar with.

As you correctly point out, because we need to ensure logical strings are written to one locale, there the bytes array is not _totally_ uniform across locale files, but relatively speaking, I believe we are okay.

Again, excellent questions/thoughts

src/GenSymIO.chpl
```diff
@@ -1908,17 +1929,17 @@ module GenSymIO {
                 firstSeg = 0;
                 lastSeg = segsArray.front();
```
glitch, line 1930: I know this line wasn't changed in the PR, but I have a question. about the `lastSeg` in the first locale.  Isn't the first "segment" (aka offset) in the `segsArray` going to be 0 in the first locale?  i.e. The first string starts at offset 0.  So segsArray.front() on local-0 should have the value of 0?  NOTE: When I construct a sample Strings array and inspect the `Strings.offsets` in the client, it shows the first element being zero.
```
>>> s = ak.array(["abc", "def", "xyz"])
>>> str(s.offsets)
'[0 4 8]'
```

If the goal is to find the end of the first segment then shouldn't it be segsArray[1] in the first local?

Moreover, isn't this the wrong goal anyway?  Isn't the real thing we want to find "the start position of the last segment" which would be `lastSeg = segsArray.back()`, the same as in the else clause?

src/GenSymIO.chpl
```diff
@@ -1928,146 +1949,101 @@ module GenSymIO {
...
-             * values array to be written to a locale.
-             */ 
-            if value == NULL_STRINGS_VALUE && i < rawChars.size-1 {
-                indices.append(i+1);
-            }
-        }
-
-        return (charList, indices);
-    }
+    private proc adjustCharArrayForLeadingSlice(sliceIndex, charArray, last, idx) throws {
```
glitch, line 2007: Note: both here and in `adjustForLeftShuffle`, the `idx` parameter is never used and can probably be removed.

hokiegeek2: @glitch It was used for logging, will remove it

src/GenSymIO.chpl
```diff
@@ -1928,146 +1949,101 @@ module GenSymIO {
...
-             */
-            if value == NULL_STRINGS_VALUE && i < segmentsBound {
-                indices.append(i+1);
-            }
-            i+=1;
-        }
-
-        return (valuesList,indices);
+    private proc adjustForLeftShuffle(shuffleLeftIndex: int, charList, idx: int) throws {
+        return charList[shuffleLeftIndex..charList.size-1];
```
glitch, line 2017: One minor thing I noticed, the following are all basically doing the same thing, slicing an array/list at a given start/end points.

- adjustCharArrayForLeadingSlice
- adjustForLeftShuffle
- adjustForRightShuffle

In theory they could be reduced to a single genericSlicing function i.e.
```
private proc genericSlice(thingToSlice, startIndex, endIndex) throws {
    return thingToSlice[startIndex..endIndex];
}

// and the corresponding calls would be changed
// ~1354
- charList = new list(adjustForLeftShuffle(shuffleLeftIndex, charList, idx);
+ charList = new list(genericSlice(charList, shuffleLeftIndex, charList.size -1));

// ~1372
- charList = new list(adjustForRightShuffle(shuffleRightIndex, charList, idx));
+ charList = new list(genericSlice(charList, 0, shuffleRightIndex));

// ~1481
- var leadingCharArray = adjustCharArrayForLeadingSlice(localLeadingSliceIndex, A.localSlice(locDom), locDom.last, idx);
+ var leadingCharArray = genericSlice(A.localSlice(locDom), localLeadingSliceIndex, locDom.last);
```
However, you could easily make an argument that having separately named procs improves the clarity of what is going on, so it's the old _6 vs a half-dozen_ idiom.

hokiegeek2: Good call, agree a common function is better, will make that change

src/GenSymIO.chpl
hokiegeek2: Ah, yeah, will take a closer look at this later today

src/GenSymIO.chpl
```diff
@@ -1421,48 +1451,51 @@ module GenSymIO {
...
+                                  'no left shuffle from locale %i to locale %i'.format(
+                                         idx,idx-1));
+                      } 
+                      
+                      // Generate the segments list now that the char list is finalized
+                      var segmentsList = generateFinalSegmentsList(charList,idx);
 
                       // Write the finalized valuesList and segmentsList to the hdf5 group
-                      writeStringsToHdf(myFileID, group, valuesList, segmentsList);
+                      writeStringsToHdf(myFileID, idx, group, charList, segmentsList);
```
glitch, line 1498: I noticed these lines are now common to both if/else clauses (which is great!):
```
// Generate the segments list now that the char list is finalized
var segmentsList = generateFinalSegmentsList(charList,idx);

// Write the finalized valuesList and segmentsList to the hdf5 group
writeStringsToHdf(myFileID, idx, group, charList, segmentsList);
```
If you want to do further clean-up they could be moved outside the if/else and consolidated.

glitch: Hmm... it's also happening in the parent if block starting at line 1238 (executed around line 1385).  In theory if you moved the declartion of `charList` outside of that block you could then move the `generateFinalSegmentsList` and `writeStringsToHdf` outside of those parent if/else blocks, but it might be a bit dicey and beyond the scope of this PR.  Sorry just thinking out-loud here as I work through this section of the code ;)

src/GenSymIO.chpl
```diff
@@ -1928,146 +1949,101 @@ module GenSymIO {
...
+             */ 
+            if value == NULL_STRINGS_VALUE && i < charList.size-1 {
+                segments.append(i+1);
             }
-            i+=1;
         }
-
-        return (valuesList,indices);
+        
+        return segments;
```
glitch, line 2046: Man this stinks that we have to scan over the bytes themselves looking for Nulls to generate the segments when in theory we already know where all the nulls are based on the offsets/segments pdarray... but I get it, there is a lot of moving around going on and those segments are being localized / re-indexed etc.  So this entire PR is a great  step in streamlining & making things faster.

glitch (approved):
Wow, if feel like I learned a ton walking though this PR :+1: 
I may not fully understand everything going on, but I feel like I have the gist of it and it helped me dig down into some of the Chapel constructs I wasn't very familiar with.  The only thing I'm really confused over is the firstSeg/lastSeg thing happening on Locale-zero but that wasn't modified in this PR so it can be a separate/future discussion.

Review completed :)

src/GenSymIO.chpl
hokiegeek2: Agreed, good call, will make that change

pierce314159 (approved):

src/GenSymIO.chpl
hokiegeek2: @glitch this is an excellent catch as you are correct in your thoughts. This bug was never caught because the biz rules are such that the majority of shuffles are shuffle lefts as shown in this example:

image

Accordingly, chars are shuffled left from locale 1 to locale 0 most of the time and therefore the segments and values are spread evenly amongst the locales. An example is below, with the current logic in the master and feature branches listed first, and the corrected version listed second:

image

image

Again, this is an excellent catch -> thank you!

hokiegeek2: And here's the corrected logic with logging statements:

image

image

hokiegeek2: [Committed](https://github.com/hokiegeek2/arkouda/commit/bbd874e4a5e4532b5e222df2e779f6f3abaa65b2) corresponding code change after extensive testing, all set.

src/GenSymIO.chpl
hokiegeek2: Fixed, checked in

src/GenSymIO.chpl
hokiegeek2: Yes, as soon as chars get shuffled around, the segments have to change.

reuster986:
@hokiegeek2 good work, and @glitch thanks for the thorough review! Merging now.

PR #802, opened 2021-05-14
Closes #738: Support ak.Timedelta.std()

### LINKED ISSUE #738: Support `ak.Timedelta.std()`
Currently, attempting to compute `td.std()` where `td` is an `ak.Timedelta` raises an error because `ak.Timedelta ** int` is not supported. I think the fix is a simple matter of adding `**` to one of the class attributes that stores the supported operators. There should also be a unit test for `td.std()`.

pierce314159:
This PR (Closes #738)
- Adds `.std` method to the `Timedelta` class, which calls `pdarrayclass.std` on the internal `_data` pdarray and converts result back to a `Timedelta` object
- Adds a test for this functionality which compares the result to the pandas implementation of standard deviation on time delta objects

### DIFF
--- a/arkouda/timeclass.py
+++ b/arkouda/timeclass.py
@@ -466,8 +466,11 @@ class Timedelta(_AbstractBaseTime):
         '''
         return to_timedelta(self.to_ndarray())
 
-    def std(self, ddof : Union[int, np.int64] = 0):
-        return Timedelta(ak_array([np.int64(self._data.std(ddof=ddof))]))
+    def std(self, ddof: Union[int, np.int64] = 0):
+        '''
+        Returns the standard deviation as a pd.Timedelta object
+        '''
+        return self._scalar_callback(self._data.std(ddof=ddof))
 
     def sum(self):
         # Sum as a pd.Timedelta
--- a/arkouda/timeclass.py
+++ b/arkouda/timeclass.py
@@ -5,6 +5,7 @@ from arkouda.pdarraycreation import from_series, array as ak_array
 from arkouda.numeric import cast, abs as akabs
 import numpy as np # type: ignore
 import datetime
+from typing import Union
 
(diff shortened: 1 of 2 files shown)


arkouda/timeclass.py
```diff
@@ -465,6 +466,9 @@ def to_pandas(self):
         '''
         return to_timedelta(self.to_ndarray())
 
+    def std(self, ddof : Union[int, np.int64] = 0):
+        return Timedelta(ak_array([np.int64(self._data.std(ddof=ddof))]))
```
glitch, line 470: I _think_ you can avoid the ak_array pdarray creation here by doing something like: 
```
return self._scalar_callback(self._data.std(ddof=ddof))
``` 
Since we're only expecting a scalar value (wrapped as a Timedelta) I think that's what you want.

I tested the change and it works with the unit test if you remove the correspnoding `to_pandas()` call.
i.e.
`self.assertEqual(ak_timedel_std, pd_timedel_std)`

Hopefully someone else can weigh in here on which is the right return type.

pierce314159: yeah that seems much cleaner. The return type is a pandas.Timedelta object not an arkouda.Timedelta in this case. That is what is being done is Timedelta.sum below so maybe I should do this for consistency. I can't think of a good reason for `std` and `sum` to have different return types

https://github.com/Bears-R-Us/arkouda/blob/eee8dd21d8f91d2861fdd14ded0d9216fdeb6a10/arkouda/timeclass.py#L468-L470

reuster986: I agree that the return type of `std` should be `pd.Timedelta`, like with `sum`.

glitch (approved):

reuster986 (approved):
looks good!

PR #806, opened 2021-05-17
Recommend Chapel 1.24.1 and use it for CI testing

ronawho:
Chapel 1.24.x contains a number of performance optimizations,
particularly on InfiniBand systems. For more information see
https://chapel-lang.org/releaseNotes/1.24/03a-perf-ib.pdf

Change the documentation to require 1.24.1, but only recommend it in the
Makefile instead of making it an error. Arkouda should continue to work
with Chapel 1.23 so existing users don't necessarily have to upgrade,
but I would recommend it for the performance gains.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     container:
-      image: chapel/chapel:1.23.0
+      image: chapel/chapel:1.24.1
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


glitch:
For what it's worth, I've been running locally with Chapel version 1.24.1 for a couple of weeks now and haven't run into any problems, granted I'm only running a single locale on a laptop.

mhmerrill (approved):
Maybe the README.md should say "We still compile with 1.23.0 but we recommend  1.24.1"
I still approve this.

Issue #805, opened 2021-05-17
CategoricalTest fails at 4+ nodes after #797


ronawho:
After https://github.com/Bears-R-Us/arkouda/pull/797 `CategoricalTest::testEquality` fails at 4+ nodes and `CategoricalTest::testBinop` fails at 6+ nodes. They both fail with "Strings or Categorical concatenate with ordered=False currently only supported with lengths >= numLocales. Please retry with ordered=True".

PR #808, opened 2021-05-18
Fixes bug in Categorical concatenate when length < numLocales

### LINKED ISSUE #710: Categorical.concatenate fails with certain segments array lengths

reuster986:
This PR should close #710 , #721 , and #805 by making `ak.concatenate(..., ordered=False)` safe with `Categorical` arrays when the length of an array is less than `numLocales`. I tested manually with `test/UnitTestConcatenate.chpl` in several such configurations, e.g. with `-nl 6 --N=3`. The `categorical_test.py` unit tests also pass with `ARKOUDA_NUMLOCALES=6`.

I also added a block to `categorical_test.py` that concatenates length-1 arrays, which should cover this case in the normal CI using 2 locales, and I re-enabled the test of unordered concatenation that had been disabled in response to #710 and #721.

@ronawho would you mind confirming that this change fixes the errors you are seeing on your systems?

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import cast, List, Optional, Union
+from typing import cast, List, Sequence, Optional, Union
 import numpy as np # type: ignore
 import itertools
 from typeguard import typechecked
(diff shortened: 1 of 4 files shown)


hokiegeek2 (approved):

ronawho (approved):
Yup, this resolves the failures I was seeing. Thanks!

PR #811, opened 2021-05-19
Closes #795: __iter__ for pdarray, Strings, and Categorical

### LINKED ISSUE #795: __iter__ for pdarray, Strings, and Categorical
I noticed that `pdarray` and `Strings` have not implemented `__iter__` methods. I wasn't sure if this is to discourage users from iterating over these presumably large arrays on client side or if this just hasn't been implemented yet

For example:
```python
>>> mystr = ak.array(['a', 'b', 'c'])
>>> for char in mystr:
        print(char)
  File "<input>", line 1, in <module>
  File "/home/arkouda/arkouda/strings.py", line 106, in __iter__
    raise NotImplementedError('Strings does not support iteration')
NotImplementedError: Strings does not support iteration
```
This can be worked around by doing either using `to_ndarray()` (which has a built-in size limit):
```python
>>> for char in mystr.to_ndarray():
        print(char)  
a
b
c
```
or by indexing the Strings/pdarray object directly:
```python
>>> for i in range(len(mystr)):
        print(mystr[i])
a
b
c
```
I was wondering if others thought it would be best to either:
- implement `__iter__` methods for these classes
- change the error message to say something like `__iter__` has not been implemented since Strings/pdarray objects can be quite large and `to_ndarray` should be called to iterate over them
- or leave it as is? 

My main concern is a user new to arkouda will probably default to indexing these objects directly before calling `to_ndarray` on them if not pointed in the right direction

pierce314159:
This PR (Closes #795):
- Updates `__iter__` error message to be consistent for `pdarrayclass`, `Strings`, and `Categorical` and refer user to `to_ndarray`
- Adds documentation for `Iteration` in each of these classes explaining why it is not supported

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -150,7 +150,7 @@ class Categorical:
         return idx[valcodes]
 
     def __iter__(self):
-        return iter(self.to_ndarray())
+        raise NotImplementedError('Categorical does not support iteration. To force data transfer from server, use to_ndarray')
         
     def __len__(self):
         return self.shape[0]
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -443,7 +443,7 @@ class pdarray:
         return self.opeq(other,"**=")
     
     def __iter__(self):
-        raise NotImplementedError('pdarray does not support iteration')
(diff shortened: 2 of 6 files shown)


pierce314159:
Suggestions on rewording of docs/error messages are more than welcome!

hokiegeek2 (approved):
Looks good to me

PR #812, opened 2021-05-19
Cacheline pad bufferIdxs for aggregators

ronawho:
Follow on to #783. In addition to cacheline aligning `bufferIdxs` also
pad it now to avoid any possibility of false-sharing.


mhmerrill (approved):
looks ok to me.

PR #816, opened 2021-05-20
810 fix strings.get lengths

### LINKED ISSUE #810: Fix Strings.get_lengths()
For an `ak.Strings` instance, the `.get_lengths()` method fails with the following error:

```python
RuntimeError: Unrecognized command: segmentedLengths str id_m id_n
```

I suspect the client-side code just needs to be updated to use the new message format, as it is still using a space-delimited string.

hokiegeek2:
This PR addresses the following:

1. Fixes #810 
2. Adds corresponding unit test for Strings.get_lengths()
3. Standardizes generic_msg invocation to generic_msg(cmd=cmd,args=args)
4. Updates client.generic_msg pydoc comments for cmd and args params
5. cleans up some import statements
6. Adds client.ruok() test case

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1,5 +1,5 @@
 from __future__ import annotations
-from typing import cast, List, Sequence, Optional, Union
+from typing import cast, List, Sequence, Union
 import numpy as np # type: ignore
 import itertools
 from typeguard import typechecked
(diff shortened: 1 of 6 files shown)


reuster986 (approved):

glitch (approved):

pierce314159 (approved):

arkouda/categorical.py
```diff
@@ -11,9 +11,8 @@
 from arkouda.dtypes import resolve_scalar_dtype, str_scalars
 from arkouda.dtypes import int64 as akint64
 from arkouda.sorting import argsort
-from arkouda.pdarraysetops import concatenate, in1d
 from arkouda.logger import getArkoudaLogger
-from arkouda.infoclass import information, pretty_print_information
```
mhmerrill, line 16: @hokiegeek2 why did you remove `pretty_print_information` ?

pierce314159: I was wondering the same thing at first but I realized that `Categorical` and `Strings` call `.pretty_print_info` on their components, so the only class that actually needs to import the `infoclass.pretty_print_information` is `pdarrayclass`.

I think I was calling `pretty_print_information` on all the subcomponents at one point before I switched to this delegating down approach and must've forgot to remove the import

hokiegeek2: @mhmerrill cause it's an unused import

Issue #786, opened 2021-05-04
Discussion about `pdarray` for strings in symbol table as well as possible other arrays


mhmerrill:
Maybe we should extend the base symbol table to include `pdarray`s of other types like `strings` as well as maybe `complex` or `timestamp `. This to me makes a lot of sense as a step toward making a symbol table of more complex objects.

hokiegeek2:
👍

mhmerrill:
We might want to make a stab at making `SegString` a subclass of `GenSymEntry` or something like that.

hokiegeek2:
@mhmerrill may not be a bad idea to capture in the  [wiki](https://github.com/mhmerrill/arkouda/wiki/String-pdarrays-in-Symbol-Table-Design-Discussion---Notes) all the functional and system requirements driving this issue.

One possible element of this could be a composite key that encapsulates all the metadata required to link elements of complex objects such as Strings (in this case, the chars and segments pdarray ids).

glitch:
Hi all, sorry for the commit spam.  I have a working implementation using a new `SegStringSymEntry:GenSymEntry` which is a composite entry of the previous offsets/bytes (aka segments/values) previously used by the Strings construct.  Check it out here: https://github.com/Bears-R-Us/arkouda/pull/819

Also, I'm putting together a wiki entry (https://github.com/Bears-R-Us/arkouda/wiki/SegStringSymEntry-to-encapsulate-Strings-as-single-entity) on some of the implementation details and common patterns & problems I was able to discover by tracing out the various usages in the code. 👍

PR #819, opened 2021-05-21
WIP / Draft 786 seg string entry

glitch:
### WIP (Work in Progress / Draft PR, please Do Not Merge)

Issue #786 
This is to get a feel for converting a Strings object into a single server-side composite object instead of having the client manage separate pdarrays.

This implementation is meant to be the start of an incremental, iterative approach instead of a complete redesign... meaning that it is likely to evolve as we start to understand its limitations or expand to a more comprehensive approach as we consider/include other complex objects.

### UPDATE:
This is now a fully-working implementation.  I am beginning to write up details of it on the wiki here: https://github.com/Bears-R-Us/arkouda/wiki/SegStringSymEntry-to-encapsulate-Strings-as-single-entity

Status: All aspects of the CI-workflow are passing.

### Future work:
Stream-line the message passing components.  I put _legacy_placeholder_ values for the various message args which previously represented the values/bytes SymEntry component of the Strings construct.  Also we need a better way of passing back meta-data about the new SegStringSymEntry from the server to the client.  It has an additional value that regular SymEntry doesn't have.

If we're going to follow this kind of pattern for other complex objects like Categoricals, we'll face the same problem, how to pass-back the additional meta-data values (i.e. we pass back something like 6 values now, but will need space for more; the SymTab.attrib proc doesn't really handle this).

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -153,7 +153,7 @@ class GroupBy:
         keytypes = []
         effectiveKeys = self.nkeys
         for k in mykeys:
-            if isinstance(k, Strings):
+            if isinstance(k, Strings):  # TODO fix Strings to perform as a single entity
                 if self.hash_strings:
                     h1, h2 = k.hash()
                     keyobjs.extend([h1,h2])
@@ -162,8 +162,8 @@ class GroupBy:
                     effectiveKeys += 1
                 else:
                     keyobjs.append(k)
-                    keynames.append('{}+{}'.format(k.offsets.name, 
-                                                   k.bytes.name))
+                    keynames.append('{}+{}'.format(k.entry.name,
+                                                   "legacy_placeholder"))
                     keytypes.append(k.objtype)
             # for Categorical
             elif hasattr(k, 'codes'):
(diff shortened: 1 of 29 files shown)


glitch:
I added in the messaging layer (server-side) changes, which basically plumb the code to run through the factory methods in `SegmentedArray.chpl` to construct and enter the  `Strings`/`SegString` object in the `SymbolTable` rather than the separate, direct allocations happening in the various messages.

Previously if you were making a `SegString` object you would allocate two names (one for offets/segments and the other for bytes/values), get the backing arrays separately and bind them individually in SymTab, then construct a SegString object to use in the rest of your procedure.

Now you pass the backing arrays together to the factory method in SegmentedArray which encapsulates the the pieces, constructs a single `SegStringSymEntry:GenSymEntry` and returns the `SegString` which is bound by its `name` in the `SymTab`.

Chapel side unit tests are passing here, but the client side is still broken (in fact they hang at the moment).

I left a `legacy_placeholder` note in place in the message parsing so I could change the _pipes_ without changing what's running through them.

glitch:
The basic implementation is done and all checks are passing; I consider it to be working from a functional stand-point.  (Future work would be to clean-up / formalize / refine some of the structures in the message passing layer using something like json.)

Issue #778, opened 2021-04-26
Type Aliasing


reuster986:
The goal of this proposal is to make arkouda more readable, maintainable, and stable by using type aliases. Many arkouda functions are generic, in the sense that they are designed to accept multiple types of arguments. Examples:
- Functions like `arange`, `randint`, and `zeros` take scalars from both python and numpy, possibly of both int and float type.
- `GroupBy` should accept `int64` `pdarray`, `Strings`, `Categorical`, or a list/iterable of any combination of these. 
- Almost every function that accepts `ak.Strings` should also accept `ak.Categorical`, because they are designed to be interchangeable.

The current type checking in arkouda relies on explicit lists of allowed types for each function argument. Using type aliases in place of explicit type lists would have the following advantages:
- Function signatures would be more readable by users (e.g. in Jupyter using Shift+Tab or '?'), with a descriptive type name like `Groupable`, rather than `Union[List[Union[pdarray, Strings, Categorical]], Union[pdarray, Strings, Categorical]]`
- The code would be more maintainable, because type updates would be localized to the alias definition, rather than requiring updates to dozens of source files
- Type aliasing would guard against incorrect type annotations, like overly restrictive type checking (see #777 ), where a type that should be allowed is mistakenly omitted from a list.

If we choose to move to type aliasing, probably the first thing we should do is look through the source code, compile a list of  type aliases that we should have, and agree on their definitions. Then someone can replace all explicit type lists with the aliases.

- [ ] Compile and define type aliases
- [ ] Replace explicit types with aliases

reuster986:
PR #845 implements type aliasing for `GroupBy`, which is probably the most complicated case, and the source of most of the type-related issues I've seen. Question for the group: should we close this issue, or go through and implement type aliasing more thoroughly?

hokiegeek2:
@reuster986 In general I vote yes for aliasing. We are already doing aliasing with scalars and can do the same in some case you mentioned above. For methods that accept scalars, Strings/Categoricals/pdarrays as well as Iterables of each, not sure.

glitch:
This may only be semi-related, but the `multidispatch` module (https://github.com/mrocklin/multipledispatch) may be helpful here.

It lets you annotate functions with the same name but with different arg types to handle dispatch to the proper func based on type.  We might be able to clean up a lot of the conditional logic going on when we have many types in the input args etc.

mhmerrill:
[...]

PR #826, opened 2021-05-26
Fixes #825: GroupBy.nunique() fails on Categorical.from_codes()

### LINKED ISSUE #825: GroupBy.nunique() fails on Categorical.from_codes()
I ran into the following bug while working Issue #692 
```python       
cat = ak.Categorical.from_codes(codes=ak.array([0, 1, 0, 1, 2]), categories=ak.array(['a', 'b', 'c']))
i = ak.arange(cat.size)

cat_group = ak.GroupBy(cat)
cat_labels, cat_values = cat_group.nunique(i)
```
calling nunique on a `Categorical.from_codes()` results in a `TypeError`

This only seems to impact `Categorical`s made using `.from_codes()` because the following runs fine
```python
cat = ak.Categorical(ak.array(['a', 'b', 'a', 'b', 'c']))
i = ak.arange(cat.size)

cat_group = ak.GroupBy(cat)
cat_labels, cat_values = cat_group.nunique(i)
```

This seems similar to Issue #777, so I expect this is the result of overly restrictive typing in `aggregate`

pierce314159:
This PR (#825):
- Adds test for `Groupby.nunique(Categorical.from_codes())`
- Turns off type checking for `aggregate` in a similar way to what is done for `argsort` in `sorting.py`

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -224,10 +224,9 @@ class GroupBy:
         repMsg = generic_msg(cmd=cmd, args=args)
         self.logger.debug(repMsg)
         return self.unique_keys, create_pdarray(repMsg)
-    
-    @typechecked
+
     def aggregate(self, values: pdarray, operator: str, skipna: bool=True) \
-                    -> Tuple[Union[pdarray, Strings, List[Union[pdarray, Strings]]], pdarray]:
+            -> Tuple[Union[pdarray, Strings, 'Categorical', List[Union[pdarray, Strings]]], pdarray]:  # type: ignore
         '''
         Using the permutation stored in the GroupBy instance, group another 
         array of values and apply a reduction to each group's values. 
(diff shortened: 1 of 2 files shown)


pierce314159:
I'm not convinced this is the right thing to do. I took away `@typechecked` over `aggregate` similar to what is done with `argsort`. I had to modify all the return signatures that called aggregate to make mypy happy.

I made this PR a draft to get input from y'all about if you think this a workable solution for now. I might keep digging into a bit because im still not sure why it works for `Categoricals` that aren't made using `from_codes`.

I'm also a little confused about this section: 
https://github.com/Bears-R-Us/arkouda/blob/2b0b33186ace57448afad330d7f48a4d4b9961d4/arkouda/groupbyclass.py#L3-L5
Is this to get around circular import issues?

I feel like taking away type checking probably isn't the best way to go about it. Would something like @reuster986 suggested in #778 fix the cirular imports and restrictive typing?

hokiegeek2:
> I'm not convinced this is the right thing to do. I took away `@typechecked` over `aggregate` similar to what is done with `argsort`. I had to modify all the return signatures that called aggregate to make mypy happy.
> 
> I made this PR a draft to get input from y'all about if you think this a workable solution for now. I might keep digging into it a bit because im still not sure why it works for `Categoricals` that aren't made using `from_codes`.
> 
> I'm also a little confused about this section:
> https://github.com/Bears-R-Us/arkouda/blob/2b0b33186ace57448afad330d7f48a4d4b9961d4/arkouda/groupbyclass.py#L3-L5
> 
> 
> Is this to get around circular import issues?

Yes, that's exactly what this is for

> I feel like taking away type checking probably isn't the best way to go about it. Would something like @reuster986 suggested in #778 fix the cirular imports and restrictive typing?

pierce314159:
This bug has been fixed by PR #827

PR #827, opened 2021-05-27
Fixed typeguard error by widening type scope to Any #825

### LINKED ISSUE #825: GroupBy.nunique() fails on Categorical.from_codes()

hokiegeek2:
This PR does the following:

1. Fixes #825 by widening the scope of nunique and aggregate return[0] element to Any
2. Adds aggregate unit tests to confirm bug fix

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import enum
-from typing import cast, List, Sequence, Tuple, Union, TYPE_CHECKING
+from typing import cast, List, Sequence, Tuple, Union, TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from arkouda.categorical import Categorical
 import numpy as np # type: ignore
@@ -227,7 +227,7 @@ class GroupBy:
     
     @typechecked
     def aggregate(self, values: pdarray, operator: str, skipna: bool=True) \
-                    -> Tuple[Union[pdarray, Strings, List[Union[pdarray, Strings]]], pdarray]:
+                    -> Tuple[Any, pdarray]:
         '''
         Using the permutation stored in the GroupBy instance, group another 
         array of values and apply a reduction to each group's values. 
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Looks great thanks @hokiegeek2!

glitch (approved):

arkouda/groupbyclass.py
```diff
@@ -639,8 +639,7 @@ def argmax(self, values : pdarray)\
             raise TypeError('argmax is only supported for pdarrays of dtype float64 and int64')
         return self.aggregate(values, "argmax")
     
-    def nunique(self, values : pdarray) \
-                    -> Tuple[Union[pdarray,List[Union[pdarray,Strings]]],pdarray]:
+    def nunique(self, values : pdarray) -> Tuple[Any,pdarray]:
```
reuster986, line 642: Should we apply this change to all the other reduction methods (`.sum()`, `.min()`, `.max()`, etc.)? I'm thinking they will encounter the same problem as `.nunique()` with GroupBy on Strings or Categorical.

reuster986 (commented):
I definitely like this change, I'm just wondering if it should be applied to all the `GroupBy` methods, not just `.nunique()`. Thoughts?

arkouda/groupbyclass.py
hokiegeek2: @reuster986 Good point and I vote yes

hokiegeek2: I recommend a separate one so we can get this merged into master for @pierce314159 work that is dependent upon this bug fix

PR #831, opened 2021-05-29
Stop importing arkouda in util/test/util.py

ronawho:
We don't need it since 0814c49


glitch (approved):
+1 Looking in my IDE its analyzer shows this as unused.

PR #832, opened 2021-05-29
Make the array values in groupby and aggregate benchmarks consistent

### LINKED ISSUE #793: GroupBy rate decreases for large problem sizes on IB
On 16 nodes of 128-core/2TB AMD with IB, the rate of the groupby benchmark peaks at a problem size of 10**9 elements/node for < 8 arrays and then falls off. All rates below are in GiB/s.

| elem/node | 1-array rate | 2-array rate | 8-array rate | 16-array rate |
| ------ | ---------- | ------- | ------ | -------- |
| 10**6 | 0.0953 | 0.0544 | 0.0163 | 0.0065 |
| 10**7 | 0.8931 | 0.6009 | 0.1716 | 0.0664 |
| 10**8 | 3.8745 | 3.1622 | 1.0653 | 0.5642 |
| 10**9 | 5.8647 | 5.0061 | 3.1355 | 2.7681 |
| 10**10 | 4.5569 | 3.525 | 3.1916 | 3.0995 |

This performance dropoff after 10**9 is not observed with coargsort, gather, or scan, which are the main kernels used in groupby.

ronawho:
Previously, the values in the array were based on the array size, which
meant that larger values were used as the array size grew. Some
algorithms in Arkouda like radixsort are based on the number of bits
used so when we transitioned from something like 32-bit to 33-bit values
an extra round of radixsort would be required.

This gave us the false impression that there was a regression at certain
problem sizes when in reality it was about the array values not the
array size. To avoid this confusion just use 32-bit values for all
problem sizes. This matches what we do in all the benchmarks.

Resolves #793

### DIFF
--- a/benchmarks/aggregate.py
+++ b/benchmarks/aggregate.py
@@ -8,7 +8,7 @@ BOOLOPS = ('any', 'all')
 def generate_arrays(N, seed):
     # Sort keys so that aggregations will not have to permute values
     # We just want to measure aggregation time, not gather
-    keys = ak.sort(ak.randint(0, N//8, N, seed=seed))
+    keys = ak.sort(ak.randint(0, 2**32, N, seed=seed))
     if seed is not None: seed += 1
     intvals = ak.randint(0, 2**16, N, seed=seed)
     boolvals = (intvals % 2) == 0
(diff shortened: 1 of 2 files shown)


reuster986 (approved):

reuster986:
@mhmerrill I wanted to make sure you saw this and had a chance to comment before I merge. This explains that "performance drop-off" (actually an artifact) in GroupBy above 1 billion elem/node that I was showing you.

mhmerrill:
@reuster986 this explains a lot. Thanks to @ronawho for discovering and fixing this!

PR #836, opened 2021-06-04
Closes #692: Add support for multi-level GroupBy with Categorical

### LINKED ISSUE #692: Add support for multi-level groupby with Categorical
Currently, `GroupBy` and `coargsort` support `Sequence[Union[pdarray, Strings]]`, but they should be expanded to support `Sequence[Union[pdarray, Strings, Categorical]]`. This is not just a matter of expanding the types; it requires new logic as well.

pierce314159:
This PR (Closes #692):
- Updates coargsort to sort Categoricals by their `Categorical.codes` component (this guarantees grouping but not sorting, the same standard for Strings)
- Adds test for GroupBy with multiple Categoricals and test for coargsort with multiple Categoricals

Note:
Because Categoricals are sorted according to their codes, two seemingly equivalent categoricals can be "sorted" by coargsort differently. I'm not convinced this is the functionality we want, so feel free to weigh in 

```python
>>> cat1 = ak.Categorical.from_codes(codes=ak.array([0, 1, 0, 1, 2]), categories=ak.array(['a', 'b', 'c']))
>>> cat2 = ak.Categorical.from_codes(codes=ak.array([1, 0, 1, 0, 2]), categories=ak.array(['b', 'a', 'c']))
>>> cat1
array(['a', 'b', 'a', 'b', 'c'])
>>> cat2
array(['a', 'b', 'a', 'b', 'c'])
>>> cat1[ak.coargsort([cat1])]
array(['a', 'a', 'b', 'b', 'c'])
>>> cat2[ak.coargsort([cat2])]
array(['b', 'b', 'a', 'a', 'c'])
```

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -56,7 +56,7 @@ class GroupBy:
 
     Parameters
     ----------
-    keys : (list of) pdarray, int64 or Strings
+    keys : (list of) pdarray, int64, Strings, or Categorical
         The array to group by value, or if list, the column arrays to group by row
     assume_sorted : bool
         If True, assume keys is already sorted (Default: False)
@@ -69,7 +69,7 @@ class GroupBy:
         The length of the array(s), i.e. number of rows
     permutation : pdarray
         The permutation that sorts the keys array(s) by value (row)
-    unique_keys : (list of) pdarray or Strings
+    unique_keys : (list of) pdarray, Strings, or Categorical
         The unique values of the keys array(s), in grouped order
     segments : pdarray
         The start index of each group in the grouped array(s)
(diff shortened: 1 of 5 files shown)


pierce314159:
I had a hard time writing the tests. I feel like they're just making sure no errors are thrown when you pass multiple categoricals to `coargsort`/`groupby` rather than actually checking all the functionality. Feedback on those would be very appreciated. 

In particular, if there's a good way to check that `coargsort` groups without explicitly comparing to another list. I kinda modeled this after what is done here in `string_test.py`
https://github.com/Bears-R-Us/arkouda/blob/042d1f5780d553072f6884085c35571d7c40f35d/tests/string_test.py#L23-L30

glitch (approved):
This took a lot less than I expected

mhmerrill:
@reuster986 I'd like to see your review on this before we merge, thanks.

reuster986 (approved):
@pierce314159 Sorry it's taken me so long to review. This PR looks really good. 

It is difficult to write tests for something like groupby. On the one hand, it would be nice to fully verify correctness via independent logic on random (non-toy) data, but on the other hand, doing so is quite expensive given our limited CI budget. I think the tests you wrote are at the right level: sufficient to confirm that the code is functioning without raising explicit errors. I am comfortable assuming correctness because a) the groupby/coargsort logic is really just operating on integers under the hood, the correctness of which is thoroughly tested elsewhere, and b) the integrity of the categories -> codes mapping, which is the likely place for errors to creep in, is also validated in other tests.

You are right to point out that coargsort only guarantees grouping, not sorting, for Strings and Categoricals. More and more, I am thinking we should refactor groupby to decouple it from coargsort and explicitly claim grouping semantics for the former and sorting semantics for the latter. But that is an issue for another time.

Finally, you raise the issue of seemingly equivalent Categoricals not sorting the same due to different ordering on the categories. I do believe this behavior is what we want, as it allows the user to specify an ordering on the categories by supplying them manually and then have the (co)sorting logic respect that ordering. Comparisons (`==` and `!=`) between two Categoricals with equivalent but permuted categories will still function as intended, because the comparison logic now normalizes the categories and codes between the two arrays. It might be worth adding a note to the Categorical documentation that if a user wants to create a Categorical with ordered categories (pandas supports this option), they should use the `.from_codes` constructor, and then sorting operations and code comparisons will respect that order.

PR #841, opened 2021-06-10
Update these operator declarations to use the operator keyword

lydia-duncan:
Chapel 1.24 added support for declaring operator overloads with a new "operator"
keyword.  This enables operator definitions to be more easily found.
Declarations without the keyword will be deprecated starting in 1.25 (which is a
few months down the road, but I figured it would be good to address avoid the
deprecation warning as soon as possible)

At least two of these could be operator methods.  I'm choosing to avoid doing that right
now for compatibility with 1.24 - there were some bugs with operator methods
remaining by the time of that release that should be resolved on master, or
shortly after a PR I have opened is merged.  In the future, we can choose to
use that functionality

`make default` and `make test` both worked

----

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,16 +98,17 @@ endif
 check-deps: $(CHECK_DEPS)
 
 CHPL_MINOR := $(shell $(CHPL) --version | sed -n "s/chpl version 1\.\([0-9]*\).*/\1/p")
-CHPL_VERSION_OK := $(shell test $(CHPL_MINOR) -ge 22 && echo yes)
-CHPL_VERSION_WARN := $(shell test $(CHPL_MINOR) -le 23 && echo yes)
+CHPL_VERSION_OK := $(shell test $(CHPL_MINOR) -ge 24 && echo yes)
+CHPL_VERSION_WARN := $(shell test $(CHPL_MINOR) -le 24 && echo yes)
 .PHONY: check-chpl
 check-chpl:
 ifneq ($(CHPL_VERSION_OK),yes)
-	$(error Chapel 1.22.0 or newer is required)
+	$(error Chapel 1.24.0 or newer is required)
 endif
-ifeq ($(CHPL_VERSION_WARN),yes)
-	$(warning Chapel 1.24.1 or newer is recommended)
-endif
+# Re-enable when support more than just one version again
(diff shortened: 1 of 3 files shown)


lydia-duncan:
I'll take a look at the arkouda_tests_linux failure, but I'm confused what I'm supposed to do about the mypy one or how my change could have caused it (since I did rebase to master beforehand)

glitch:
@lydia-duncan The `mypy` thing is not your fault.  I noticed this broke a few days ago and opened Issue#839  this morning along with a fix in PR#840 . I believe it is an upstream change which caused an issue.

lydia-duncan:
Okay, the only failure remaining is the mypy failure, so I think this is actually ready for a look!

glitch (approved):

ronawho:
This bumps the minimum Chapel version to 1.24.0, right? If so the version check in the [Makefile](https://github.com/Bears-R-Us/arkouda/blob/042d1f5780d553072f6884085c35571d7c40f35d/Makefile#L100-L110) should be updated

(Historically we have tried to keep things compatible with the last 2 major chapel releases, but I'm not sure if that's still important so I'll defer to Mike/Bill)

lydia-duncan:
> This bumps the minimum Chapel version to 1.24.0, right?

That's correct.  Running against Chapel master without this change will generate deprecation warnings for operators declared without the keyword, so it's a choice between supporting 1.23.x or having warning messages with HEAD.

If we want to ensure it works with the last two release, it might be good to have testing that locks that in - I thought the testing we run for arkouda only checked latest release and current HEAD

reuster986:
@lydia-duncan and @ronawho sorry it's taken me so long to review! I am ok with requiring Chapel 1.24 -- all our systems are already running it for the performance improvements anyway. I think let's go ahead and update the Makefile check.

reuster986 (changes requested):
Changes look good, just waiting on an update to the Makefile version check and possibly a rebase to pull in the fix for the mypy issue.

lydia-duncan:
Alright, the Makefile change is in, and I rebased so the mypy issue was fixed as well.  I think this is ready for another look!

reuster986 (approved):

Issue #834, opened 2021-06-03
Refactor GenSymIO.chpl module to allow for multiple file types


mhmerrill:
We need to refactor the `src/GensymIO.chpl` module so we can support HDF5 as well as Arrow/Parquet memory and file formats.

The vision for this is to refactor the GenSymIO module and make it auto detect which file format is being used and call the right library to read the data.
You could in theory also pass in a url or something pointing to an arrow server or something like that...

glitch:
I was able to locate the _magic numbers_ for the following:

- HDF5 (https://portal.hdfgroup.org/display/HDF5/File+Format+Specification)
    - Hex `89 48 44 46 0d 0a 1a 0a`
- Arrow (https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format) for IPC File Format
    - `ARROW1` padded out to 8 bytes
- Parquet (https://parquet.apache.org/documentation/latest/) see _File format_ heading
    - 4 byte header: `PAR1`

glitch:
Note/Update: An initial PR #835 was posted, reviewed, and merged, but it was just an initial refinement.  I still have some work left to do here in integrating the magic number detection and adding unit tests for it.

glitch:
@mhmerrill I put up another [PR#843](https://github.com/Bears-R-Us/arkouda/pull/843) which includes the file _magic_ implementation along with further clean up on `GenSymIO.chpl`

The next step here would probably be to move the specific HDF5 specific procs to their own module to start making the following structure:
```
GenSymIO
  |- Hdf5IO
  |- ParquetIO
  |- ArrowIO
  |-...
etc.
```
After that we could start to formalize the read/write messages and delegate down to the appropriate module, but since we don't currently have support for other types I didn't want to go to far down this road without checking in.  Thoughts?  Should I start moving the HDF5 stuff to its own file or is this a good stopping point?

PR #843, opened 2021-06-11
834 file magic

glitch:
Issue #834

This PR adds the following:

- FileIO.chpl : File Magic detection using published Magic Header information for
    - HDF5
    - Parquet
    - Arrow
- Chapel unit test for magic detection and boundary conditions
- GenSymIO.chpl : Minor improvements and code reduction
    - Replacing `try!` with error handling in `arrayMsg` proc
    - Removing duplicate code in `processFilenames` the two procs were identical save for a `prepareGroup` proc which is now wrapped in conditional logic.


reuster986 (approved):
yay! progress towards I/O abstraction!

reuster986:
@mhmerrill I haven't merged yet because doing so would close #834 and I wanted to give you a chance to weigh in before that point. We might want to start a new issue that tracks progress towards a full refactor.

mhmerrill:
@glitch and @reuster986 Since we are doing incremental PRs to address issue #834 maybe we should have some sub issues that link to that issue which we can address and close with the subsequent PRs… what do you think?

glitch:
> @glitch and @reuster986 Since we are doing incremental PRs to address issue #834 maybe we should have some sub issues that link to that issue which we can address and close with the subsequent PRs… what do you think?

I think that's a good idea.  I stopped refactoring at this point because it was about to get a lot more involved depending on the direction we want to go (see this [comment](https://github.com/Bears-R-Us/arkouda/issues/834#issuecomment-859683226))

PR #844, opened 2021-06-16
Change broadcast default and update docs

reuster986:
This PR does two things:
- Changes default for `permute` keyword argument of `GroupBy.broadcast` from `False` to `True`. In personal experience and discussions with users, I have only ever seen it called with `permute=True`. Making this the default would clean up user code and make the behavior more intuitive for new users.
- Updates documentation for `GroupBy.broadcast` and `ak.broadcast` to reflect new default and the fact that we can now broadcast booleans.

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -73,9 +73,7 @@ class Categorical:
                                  "Strings not yet supported"))
             g = GroupBy(values)
             self.categories = g.unique_keys
-            self.codes = zeros(values.size, dtype=np.int64)
-            self.codes[cast(pdarray, g.permutation)] = \
-                                g.broadcast(arange(self.categories.size))
+            self.codes = g.broadcast(arange(self.categories.size), permute=True)
             self.permutation = cast(pdarray, g.permutation)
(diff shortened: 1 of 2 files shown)


arkouda/categorical.py
```diff
@@ -73,9 +73,7 @@ def __init__(self, values, **kwargs) -> None:
                                  "Strings not yet supported"))
             g = GroupBy(values)
             self.categories = g.unique_keys
-            self.codes = zeros(values.size, dtype=np.int64)
-            self.codes[cast(pdarray, g.permutation)] = \
-                                g.broadcast(arange(self.categories.size))
+            self.codes = g.broadcast(arange(self.categories.size), permute=True)
```
pierce314159, line 76: We don't need the `permute=True` here since this is the default for broadcast now, right? 

It's probably good to be explicit in case the default does get changed back for whatever reason. No change requested

pierce314159 (approved):
Looks good to me!

hokiegeek2 (approved):
Looks good to me

Issue #780, opened 2021-04-29
.nunique aggregation slow on IB


reuster986:
Testing arkouda on an IB cluster, performance looks good overall, but one case where it gets hung up is in `GroupBy.nunique()`, which calls `src/ReductionMsg.chpl:segNumUnique` under the hood. This works fine on XC, so I'm thinking we just missed an opportunity for aggregation in this function somewhere. However, the function itself is fairly old and unnecessarily complex -- I should know because I wrote it :) -- so it is probably time for a replacement.

I propose two stages:
- [x] Write a python-side method that performs nunique via another groupby:
```python
class GroupBy:
...

    def nunique(self, vals):
        uksize = ... # get length of self.unique_keys
        ukidx = self.broadcast(ak.arange(uksize), permute=True)
        # Find unique pairs of (key, val)
        g = GroupBy([ukidx, vals])
        # Group pairs again by keyidx 
        g2 = GroupBy(g.unique_keys[0], assume_sorted=True)
        # Count number of unique vals per keyidx
        _, nuniq = g2.count()
        # Re-join unique counts with original keys (sorting guarantees same order)
       return self.unique_keys, nuniq

    def aggregate(...):
        # Insert special case to handle 'nunique' by calling above method instead of sending a server message
```

- [ ] Once the above is stable, port the logic over to a new Chapel function. This would greatly benefit from making some arkouda behaviors available as pure Chapel functions that can be called on Chapel arrays (or a custom array class), like:
  - broadcast
  - gather (`array[indices]` and `array[mask]`)
  - findSegments, or `ediff`, or whatever we want to call it
  - coargsort

reuster986:
@hokiegeek2 had you started working on this yet, or am I ok to take over?

hokiegeek2:
@reuster986 I have not so, feel free to take this on

reuster986:
PR #845 addresses the first half of this issue. I'm thinking of rolling the second half into a broader issue about refactoring groupby, findSegments, and unique.

PR #845, opened 2021-06-16
Client-side nunique for IB performance and more functionality

reuster986:
This PR migrates the logic for `GroupBy.nunique()` from the Chapel side to the Python side, in hopes of achieving two results:
- Improving performance, especially on IB, by re-using primitives that already benefit from CommAggregation. As described in #780 , there is some part of the current implementation that is missing out on aggregation and is prohibitively slow on IB.
- Expanding the allowed argument types from only `int64` pdarray to any groupable type.


pierce314159 (approved):
I think this is great! 

I like the movement towards type aliasing (#778) with groupable, I think this is cleaner and less error prone

Should this PR be linked to close #780 now that the second half is covered by #846?

reuster986:
@pierce314159 thanks! good call on linking #780 -- done.

Issue #847, opened 2021-06-17
Uncaught Chapel Logging Error


reuster986:
There is a user who is getting an uncaught BlockingIOError when calling `ak.read_all()` on a list of several thousand filenames. The error is thrown by a Chapel function and somehow not caught in the main server loop. Sanitized text of the terminal output:

```
uncaught BlockingIOError: Resource temporarily (in channel.flush with path "pipe:[<somenumber>]" offset <othernumber>)
$ARKOUDA_HOME/src/Logging.chpl:52: thrown here
$ARKOUDA_HOME/src/arkouda_server.chpl:168: uncaught here
```

Throwing function:
https://github.com/Bears-R-Us/arkouda/blob/e9053141904bbb462f29a6ada346272692d9d15e/src/Logging.chpl#L48-L54

Non-catching function:
https://github.com/Bears-R-Us/arkouda/blob/e9053141904bbb462f29a6ada346272692d9d15e/src/arkouda_server.chpl#L161-L173

I think we should do two things:
1. Add a `throws` to the signature of `sendRepMsg`, so that errors will be passed up to the big try/catch in the main server loop.
2. Add logic in all the `Logger` methods to gracefully handle logging errors that do not affect the actual computation, so that the user's workflow can proceed uninterrupted.

The first is a safety net to ensure that the server doesn't crash, but the second is necessary because logging failures should not really interrupt execution.

hokiegeek2:
@reuster986 I'm on it

hokiegeek2:
@reuster986 submitted [PR](#849)

hokiegeek2:
@pierce314159 I believe so, but that's Mike's or Bill's call

PR #849, opened 2021-06-17
847 Uncaught Chapel Logging Error

hokiegeek2:
This PR addresses #847 by the following:

1. Added throws clause to arkouda_server.sendRepMsg to push uncaught errors up to the top of the call stack in arkouda_server.main
2. Added try/catch block around Logging.generateLogMessage to catch errors in message generation and write out a lower-risk error message within the catch block


reuster986 (approved):
Looks good!

Issue #838, opened 2021-06-08
Peel segfaults when delimiter is longer than last string


reuster986:
Technically, it's when the delimiter is longer than the last null-terminated segment, i.e. when `length(delimiter) > length(last_string) + 1`.

glitch:
I can't get this to segfault with a basic, but something weird is definitely going on.

This works as expected:
```python
import arkouda as ak
import pandas as pd
ak.connect("localhost")
d = "-" * 25
print(f"delimiter:{d}")
series = pd.Series([f"abc{d}xyz", f"small{d}dog", f"blue{d}hat", "last"])
print(f"series:\n{series}")
pda = ak.from_series(series, "string")
a, b = pda.peel(d)
print(f"a from peel: {a}")
print(f"b from peel: {b}")
a, b = pda.rpeel(d)
print(f"a from rpeel: {a}")
print(f"b from rpeel: {b}")
a = pda.endswith(d)
print(f"a from endswith: {a}")
```
Outputs
```
delimiter:-------------------------
series:
0      abc-------------------------xyz
1    small-------------------------dog
2     blue-------------------------hat
3                                 last
dtype: object
a from peel: ['abc', 'small', 'blue', '']
b from peel: ['xyz', 'dog', 'hat', 'last']
a from rpeel: ['abc', 'small', 'blue', 'last']
b from rpeel: ['xyz', 'dog', 'hat', '']
a from endswith: [False False False False]
```

However, when I change the series to
```python
series = pd.Series([f"abc{d}xyz", f"small{d}dog", "last"])
```
(so basically one element shorter)
The peel completely misses the second element.  Output:
```
delimiter:-------------------------
series:
0      abc-------------------------xyz
1    small-------------------------dog
2                                 last
dtype: object
a from peel: ['abc', '', '']
b from peel: ['xyz', 'small-------------------------dog', 'last']
a from rpeel: ['abc', 'small-------------------------dog', 'last']
b from rpeel: ['xyz', '', '']
a from endswith: [False False False]
```
See "a from peel" and "b from peel" second elements, it looks like it missed the delimiter.  Notice that it did NOT miss it in the 4 element case.

glitch:
LoL, I put this in a python unit test and now the exact opposite is happening.  The array of length 3 works and length 4 misses. 🤯

glitch:
I added some DEBUG info for the case I can get to fail (a 3 string case)
```python
[...]

PR #850, opened 2021-06-21
Issue #838: Fixing peel segfault when long delimiter

glitch:
Issue #838: The peel operation may segFault when the delimiter is sufficiently longer than the last string.

This looked to be caused by an out of bounds index operation when determining if the string `hasEnough` to peel.  The fix was made by adjusting the bounds checking conditional statements.

A number of unit tests both server-side (in chapel) and client side (in python) were added to verify correctness.


reuster986 (approved):
looks good!

Issue #508, opened 2020-10-02
update version and make a new tag in repo


mhmerrill:
we've had a lot of new features added and bug fixes since we rev'd the version number.

mhmerrill:
@glitch @hokiegeek2 @pierce314159 @reuster986 we should come up with a methodology for updating the version and tag in the repo.

Currently
- I rev the VERSION file then submit a PR
- after PR is merged into the master, I tag the commit in the commit stream which generates a release

glitch:
Something we might want to consider is [versioneer](https://github.com/python-versioneer/python-versioneer).  The short summary is that it auto-generates the version string based on where you are in your repo.  If you create a tag in the repo the version string is based on that tag.  Depending on style (which you can configure) you can have it set as "distance" from previous tag for cases where you haven't cut a new release, you could also set it to date, etc.  The key thing is that it is generated, so there is no version file to maintain.  I'm not sure how it would splice into the chpl-side of things, but between python & make we could probably figure something out.

reuster986:
I think the need for versioning and tagging is becoming more pronounced as we are trying to automate the building and deployment of arkouda, so I'm labeling this issue "important". I like the versioneer idea for auto-generating the version string. Our devops guy has said that he can set it up so that arkouda auto-builds whenever a new tag gets merged into master, which I think would be great, but it would require us developing some guidelines and adopting new habits around tagging.

I believe adopting an auto-versioning scheme would also make it easier for us to deploy to pypi and truly support pip installation.

@glitch if @mhmerrill is on board, would you be willing to put together a draft PR showing how versioneer would work?

glitch:
@reuster986 You bet.  I should be able to roll something together and test it out in my fork to see/demonstrate how it works.

PR #851, opened 2021-06-22
Issue #508: Adding versioneer to project for version management.

glitch:
Issue #508 
This adds versioning through [versioneer](https://github.com/python-versioneer/python-versioneer) and removes the static `VERSION` file from the repo.  Version numbers will now by dynamically determined based on checking out a tag (for any new tags after merge of this PR) or based on the commit _distance_ from the last tag in the branch lineage.

I added the following to the README.md to show various examples

An example using a hypothetical tag `1.2.3.4`
```bash
git checkout 1.2.3.4
python -m arkouda |tail -n 2
>> Client Version: 1.2.3.4
>> 1.2.3.4

# If you were to make uncommitted changes and repeat the command you might see something like:
python -m arkouda|tail -n 2
>> Client Version: 1.2.3.4+0.g9dca4c8.dirty
>> 1.2.3.4+0.g9dca4c8.dirty

# If you commit those changes you would see something like
python -m arkouda|tail -n 2
>> Client Version: 1.2.3.4+1.g9dca4c8
>> 1.2.3.4+1.g9dca4c8
```
In the hypothetical cases above _Versioneer_ tells you the version and how far / how many commits beyond the tag your repo is.

When building the server-side code the same versioning information is included in the build.  If the server and client do not
match you will receive a warning.  For developers this is a useful reminder when you switch branches and forget to rebuild.

```bash
# Starting the arkouda when built from tag 1.2.3.4 shows the following in the startup banner 
arkouda server version = 1.2.3.4

# If you built from an arbitrary branch the version string is based on the derived coordinates from the "closest" tag
arkouda server version = v2019.12.10+1679.abc2f48a

# The .dirty extension denotes a build from uncommitted changes, or a "dirty branch" in git vernacular
arkouda server version = v2019.12.10+1679.abc2f48a.dirty
```

For maintainers, creating a new version is as simple as creating a tag in the repository; i.e.
```bash
git checkout master
git tag 1.2.3.4
python -m arkouda |tail -n 2
>> Client Version: 1.2.3.4
>> 1.2.3.4
git push --tags
```

### A note on version mis-match
The version number will be folded into the build of the arkouda_server binary (I hooked it in via Makefile) so you can easily see a mismatch of client / server versions in a warning message when connecting the client to the server.  This is useful when you built the `arkouda_server` binary on one branch and either switch branches / make changes on your branch and the version no longer matches up with the client.

If folks want, I can draft of a wiki page with more step-by-step examples and screen shots; but otherwise versioning is as simple as making a tag in git and running make.


Makefile
```diff
@@ -165,13 +165,13 @@ endef
 $(eval $(call create_help_target,arkouda-help,ARKOUDA_HELP_TEXT))
 
 # Set the arkouda server version from the VERSION file
-VERSIONFILE=./VERSION
+VERSION=$(shell python -m arkouda|tail -n 1)
```
reuster986, line 168: I see you added a `__main__.py` that prints the version on module load to ensure this Makefile line works, but I'm wondering if it would be more robust to do something like `VERSION=$(shell python -c "import arkouda; print(arkouda.__version__)")`?

reuster986 (changes requested):
I like the look of this! 

In addition to one suggestion about harvesting the version in the Makefile, I wanted to ask about the new `versioneer.py` file included in this PR. It looks a lot like source code from versioneer itself, and I was wondering if it was included by mistake...

reuster986:
@mhmerrill should definitely approve before we merge.

Makefile
glitch: RE: `versioneer.py` - The versioneer docs say this is necessary if you want `setup.py` to be able to compute the version (see the section [Theory of Operation](https://github.com/python-versioneer/python-versioneer#theory-of-operation) for details).

RE: the shell operation and `main.py`, Yes! I can take that out and streamline it a bit.

There are 2 basic options:
1. Directly from versioneer
```
VERSION=$(shell python -c "import versioneer; print(versioneer.get_versions()[\"version\"])")
```

2. via import; note: Pyfiglet prints out the ascii-art any time arkouda is imported so I still need `tail` to filter down to the last line, but here's that version:
```
VERSION=$(shell python -c "import arkouda; print(arkouda.__version__)"|tail -n 1)
```

I'll push a fix/change momentarily

reuster986 (approved):

Makefile
reuster986: Ok, that makes sense.

Makefile change looks good!

PR #854, opened 2021-06-23
Closes #842 adding empty string check for ls_hdf

### LINKED ISSUE #842: Fix server crash for ak.ls_hdf("")
Discovered server crashing issue due to an assertion failure in [GenSymIO.chpl](https://github.com/Bears-R-Us/arkouda/blob/36abc34050e430f393ff681e6095911046648b4f/src/GenSymIO.chpl#L206), line 206, when `ak.ls_hdf` is called for an empty string.  After a little digging, it appears that an assertion is being violated within the `glob` iterator in the `FileSystem` module of Chapel.

This was confirmed by compiling a single-line Chapel script containing only the following:

```
use FileSystem;

var tmp = glob("");
```

While this corner case can easily be handled in the client or server code, it may be best to raise this issue with the Chapel team.

glitch:
Closes #842 adding empty string check for ls_hdf

Check occurs on both client & server side code.
Unit test added for python tests.

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -23,7 +23,15 @@ def ls_hdf(filename : str) -> str:
     -------
     str
         The string output of `h5ls <filename>` from the server
+
+    Raises
+    ------
+    ValueError
+        Raised if filename is empty or contains only whitespace
     """
+    if not (filename and filename.strip()):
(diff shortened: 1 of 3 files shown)


mhmerrill (approved):
i think this looks ok.

Issue #794, opened 2021-05-10
Investigate performance of client-server array transfers


reuster986:
The functions `ak.array()` and `pdarray.to_ndarray()` are about 2x slower on a 128-core/2TB IB system than on Cray XC. Even on XC, the transfer rate for these functions is in the neighborhood of 1 MB/s, which seems quite slow. We should investigate how to improve this performance. The first step is to make a benchmark.

For cases where the client and server share the same filesystem, we could theoretically get a much faster transfer via disk, using `ak.save_all()` and `h5py` on the client side.

hokiegeek2:
@reuster986 @mhmerrill Chapel support for Apache Arrow/Parquet would also be awesome here

glitch:
I started looking into Arrow/Parquet to see how hard it would be to call from Chapel.  Turns out there is a C api and I was able to get a rudimentary example working from Chapel:
https://github.com/glitch/arkouda/wiki/Investigating-Arrow

In theory we could ship this over a socket and read it back out on the python side with `pyarrow`.  It's another set of dependencies, but it doesn't look too bad.

hokiegeek2:
@glitch that's really cool 💯

mhmerrill:
@glitch we have an intern starting on the Chapel team who is going to work on making a Chapel module to wrap [Arrow](http://arrow.apache.org/).
I will put in another issue to track this work.
It would be great if you could look into refactoring the [`src/GenSymIO.chpl`](https://github.com/Bears-R-Us/arkouda/blob/master/src/GenSymIO.chpl), I will also put in another issue to track the refactor.

mhmerrill:
@Bears-R-Us/arkouda-core-dev Since ZMQ doesn't seem to be optimized for data transport... I have asked @hokiegeek2 to look into what library we could use to move the data between the client and the server. Refactoring the client-server messaging into a control/command channel and a data channel. This could also simplify the control/command channel messages to get the binary data out of that path.

hokiegeek2:
👍

ronawho:
As far as speeding up `pdarray.to_ndarray()`itself I think there's probably 2 important aspects here:
 - How fast are we turning the distributed arrays into bytes
 - How quickly are those bytes being transferred across the network

[...]

PR #856, opened 2021-06-24
Improve the performance of client-server array transfers

ronawho:
Improve the performance of `ak.array()` and `pdarray.to_ndarray()`. This
improves how the arrays are translated to bytes. For pdarray.to_ndarray,
previously we would just read from the distributed pdarray directly, but
this results in lots of fine-grained reads from that distributed array.
Change to copying to a local array, which takes advantage of Chapel's
bulk transfer, and then doing the byte conversion locally. For ak.array
do something similar but write into a local array and then do bulk
transfer to the distributed array.

This also adds a benchmark to track the performance. Note that unlike
most tests that take a per-locale size, this just takes a fixed size
since it ultimately depends on the client to head-node connection.

| before |  3.9 MiB/s |  4.7 MiB/s |
| after  | 20.5 MiB/s | 23.5 MiB/s |

And for 16-node-cs-hdr:

| before |  2.4 MiB/s |  2.9 MiB/s |
| after  | 29.5 MiB/s | 25.9 MiB/s |

So roughly a ~5x improvement for XC and ~10x for CS.

Part of https://github.com/Bears-R-Us/arkouda/issues/794

### DIFF
--- a/benchmarks/array_transfer.py
+++ b/benchmarks/array_transfer.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+import time, argparse
+import numpy as np
+import arkouda as ak
+
+TYPES = ('int64', 'float64')
+
+def time_ak_array_transfer(N, trials, dtype, random, seed):
+    print(">>> arkouda {} array creation".format(dtype))
+    cfg = ak.get_config()
+    print("numLocales = {}, N = {:,}".format(cfg["numLocales"], N))
+
+    a = ak.randint(0, 2**32, N, dtype=dtype, seed=seed)
+    nb = a.size * a.itemsize
+    ak.client.maxTransferBytes = nb
+
+    to_ndarray_times = []
+    to_pdarray_times = []
+    for i in range(trials):
+        start = time.time()
+        npa = a.to_ndarray();
+        end = time.time()
+        to_ndarray_times.append(end - start)
+        start = time.time();
+        aka = ak.array(npa)
+        end = time.time()
+        to_pdarray_times.append(end - start)
+    avgnd = sum(to_ndarray_times) / trials
+    avgpd = sum(to_pdarray_times) / trials
+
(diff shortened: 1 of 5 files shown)


benchmarks/array_transfer.py
```diff
@@ -0,0 +1,76 @@
...
+    print("to_ndarray Average rate = {:.4f} GiB/sec".format(nb/2**30/avgnd))
+    print("ak.array Average rate = {:.4f} GiB/sec".format(nb/2**30/avgpd))
+
+def check_correctness(dtype, random, seed):
+    N = 10**4
+
+    a = ak.randint(0, 2**32, N, dtype=dtype, seed=seed)
+    npa = a.to_ndarray();
+    aka = ak.array(npa)
+    assert np.allclose(a, aka)
```
reuster986, line 44: No change requested, I'm just mildly surprised this works on two arkouda arrays. Very cool.

reuster986 (approved):
Awesome!

benchmarks/array_transfer.py
ronawho: Hmm, I am too actually. I wonder what this is doing under the covers. Is there a built in way to compare 2 arkouda arrays?

reuster986: Unfortunately, no. When I do these kinds of comparisons, I usually start with a numpy array, do the round trip, and then compare two numpy arrays. But it's very possible that `np.allclose(a, b)` is just using operators and methods that pdarrays support.

reuster986:
@ronawho I just merged @glitch 's PR that fixes the docs and mypy CI issues, so a rebase will hopefully make the checks pass.

benchmarks/array_transfer.py
ronawho: I'll switch to to start with the numpy array.

Issue #858, opened 2021-06-29
Typeguard error on multi-level GroupBy + .aggregate


reuster986:
Reproducer:

```python
keys = [ak.randint(0, 10, 100), ak.randint(0, 10, 100)]
g = ak.GroupBy(keys)
g.min(ak.randint(0, 10, 100))

[stack trace involving typeguard functions]

NameError: name 'Categorical' is not defined
```

This error only triggers when more than one array is passed to `GroupBy` and then an aggregation method is called. I'm not sure the unit tests cover that case.

How do we handle a forward reference like `Categorical` with typeguard? 

Is there a way to turn off runtime type checking by default for users? They should not have to deal with type "errors" on functions that actually work fine. The typeguard documentation says you can turn it off by enabling optimization in the python interpreter, but it's not clear how to do that in a jupyter notebook.

@hokiegeek2 @glitch @pierce314159 I know I wrote the code that introduced this bug, but could one of you take on this issue, since it is a major blocker with no workaround, and since I don't have any telework scheduled this week? Thanks in advance...

hokiegeek2:
@reuster986 we will rock this out ASAP. 
@pierce314159 I am gonna take a look as well right now.

reuster986:
Thanks, guys! I'll take a look at the PR now.

pierce314159:
> Is there a way to turn off runtime type checking by default for users? They should not have to deal with type "errors" on functions that actually work fine. The typeguard documentation says you can turn it off by enabling optimization in the python interpreter, but it's not clear how to do that in a jupyter notebook.

@reuster986 I did some looking into this and seems like users need to run `%env PYTHONOPTIMIZE=1` in their jupyter notebooks to disable typechecking. This disables all assert statements and code dependent on `__debug__`

I think you're referring to this Typeguard Issue: [Disable type checking on production](https://github.com/agronholm/typeguard/issues/54), which says 
> Type checks can be fairly expensive so it is recommended to run Python in "optimized" mode (python -O or setting the PYTHONOPTIMIZE environment variable) when running code containing type checks in production. The optimized mode will disable the type checks, by virtue of removing all assert statements and setting the __debug__ constant to False.

To set the `PYTHONOPTIMIZE` environment variable in a jupyter notebook [`%` magic commands](https://stackoverflow.com/questions/37890898/how-to-set-env-variable-in-jupyter-notebook/47326270) seem to work
[...]

PR #859, opened 2021-06-29
Hotfix for Issue #858- Removes @typechecked over aggregate

pierce314159:
Issue #858:
Certainly not a permanent fix but should be enough to remove the 'Categorical' typechecking block for users for the time being

This PR removes `typechecked` over aggregate and adds `# type: ignore` for all aggregate calls. This is intended to be a quick fix to avoid interrupting user workflows while we look into handling `Categorical` typechecking in `Groupby`.


reuster986 (approved):

hokiegeek2 (approved):
Fixes the immediate problem, good work

glitch (approved):

PR #863, opened 2021-06-29
Closes #862: Explict python3 for versioneer and adding 'make version' target

### LINKED ISSUE #862: Fix VERSION in Makefile to use python3
Fix the VERSION variable in the Makefile to use python3 explicitly.

Also add a `make version` target.

glitch:
Closes #862: Explict python3 for versioneer and adding 'make version' target


mhmerrill (approved):
looks good.

PR #864, opened 2021-06-30
Optimize byte<->array conversions in to_ndarray

ronawho:
Optimize the byte->nparray conversion by using `np.frombytes` instead of
`struct.unpack`. Unfortunately, we still have to make a copy because `bytes`
aren't mutable so we have to create a `bytearray` copy. There may be some way
to eliminate this with `memoryview`, but I'm not familiar with that yet.

Then optimize pdarray->byte conversations by directly copying the localized
array's buffer into a bytes type instead of writing to and then reading from a
memory mapped file. Here again there's an extra copy, but we could eliminate
that by using the `makeArrayFromPtr` interface.

Note that this breaks portability when the endianness of the server doesn't
match the client. Previously, everything forced big-endian and now we assume
client and server match. This can probably be addressed by sending the
endianness along with the message and interpreting differently, but I'm not
sure. Because of this breakage I've drafted this for now.

Here are some performance results for to_ndarray on 16-node-xc

| config        | Time  |  Rate     |
| ------------- | ----: | --------: |
| before        | 39.0s |  20 MiB/s |
| np.frombuffer | 23.8s |  32 MiB/s |
| direct-bytes  |  2.1s | 362 MiB/s |

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -856,10 +856,10 @@ class pdarray:
         if len(rep_msg) != self.size*self.dtype.itemsize:
             raise RuntimeError("Expected {} bytes but received {}".\
                                format(self.size*self.dtype.itemsize, len(rep_msg)))
-        # Use struct to interpret bytes as a big-endian numeric array
-        fmt = '>{:n}{}'.format(self.size, structDtypeCodes[self.dtype.name])
-        # Return a numpy ndarray
-        return np.array(struct.unpack(fmt, rep_msg)) # type: ignore
+        dt = np.dtype(self.dtype)
+        dt = dt.newbyteorder('>')
+        mut_rep_msg = bytearray(rep_msg)
+        return np.frombuffer(mut_rep_msg, dt)
 
     def to_cuda(self):
         """
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -857,7 +857,6 @@ class pdarray:
             raise RuntimeError("Expected {} bytes but received {}".\
                                format(self.size*self.dtype.itemsize, len(rep_msg)))
         dt = np.dtype(self.dtype)
-        dt = dt.newbyteorder('>')
         mut_rep_msg = bytearray(rep_msg)
         return np.frombuffer(mut_rep_msg, dt)
 
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -139,40 +139,31 @@ module GenSymIO {
         var arrayBytes: bytes;
         var entry = st.lookup(payload);
         overMemLimit(2*entry.size*entry.itemsize);
-        var tmpf: file; defer { ensureClose(tmpf); }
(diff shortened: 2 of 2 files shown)


ronawho:
@reuster986 this is a draft PR that pretty significantly improves `to_ndarray` performance more as part of https://github.com/Bears-R-Us/arkouda/issues/794. It breaks portability when the endianness of the server doesn't match the client, which is why I opened as a draft. Previously, everything forced big-endian, but now it just relies on native endianness.

I think we can do something like send the endianness as part of the message and convert if needed, but I'm not positive yet. I wanted to check if this is worth pursuing (or if things seem fast enough without this) and how much you care about things working when the client/server have different endianness. So far as I know the only practical case where that would matter is if you're running on a powerpc server and have your client running on an x86 laptop or something. I'm not sure how important that is, but it seems like a shame to break the functionality anyways.

reuster986:
@ronawho Cool! I think this is definitely worth pursuing, as there are a lot of workflows that pass data back and forth between arkouda and numpy/pandas. The 15x performance improvement is too good to pass up, and I think the potential memory savings you mentioned are also quite attractive. As server-client transfers become faster, users will probably start doing bigger and bigger transfers, which will start to put pressure on the memory of the head node and the client.

I don't know of anyone using arkouda where the client and server have different endianness, but it's bound to happen at some point, so it's good to plan for it now. Of course, it's difficult to find a platform to test on... I'm wondering if the best thing to do for now is 

1. Have `ServerConfig.chpl` store the endianness of the server so that it can be retrieved by the client with `ak.get_config()`
2. The client can detect its own endianness with `sys.byteorder`
3. If the two don't match, have the client fall back to using `struct.unpack` and plugging in the endianness of the server.

This would delegate the endianness handling to `struct`, at the expense of some performance, but the mixed-endian case is so rare that I think it's ok.

ronawho:
Sounds good. I have access to a powerpc system so I should be able to test the mixed endianness. Do you remember if there was a specific reason for changing to using big-endian in https://github.com/Bears-R-Us/arkouda/commit/a375b5933674ed777bb7d98b1aa4a861c67c0911? I suspect not, but want to make sure I'm not missing something that would make this approach obviously wrong.

For now I'll go with the plan you outlined and in followup PRs optimize to `ak.array()` case and look to minimize the number of copies that are needed. Do you have a sense for whether `ak.array()` or `to_nparray()` is more common or are the used about the same?

ronawho:
Ah, so it turns out the powerpc system I have access to is little-endian (and most ppc after power 7 are little-endian.) I can still test this by using the current file writing trick on the server, but I think this makes it very unlikely anybody will have mixed endian client/server in practice.

ronawho:
I'm going to close this in favor of multiple PRs so we can better track perf changes in nightly configs. I expect there to be PRs for:
 - Optimizing client conversations between bytes and numpy arrays 
 - Optimizing server conversations between bytes and pdarrays 
 - Reducing the number of memory copies needed in the client

Issue #821, opened 2021-05-21
Enable `ak.read_all()` to ignore/warn of bad files


reuster986:
Currently, `ak.read_all(filenames)` will raise an error if any file has any of the following issues:
* permission error
* bad HDF5 header
* missing dataset/column
* dataset with incorrect dtype

In some use cases, there are so many files that the likelihood of one of them having an issue is high, and we don't always want `ak.read_all` to hard fail in that condition. We should make a keyword arg, like `ak.read_all(filenames, errors='warn')` that will warn the user of problematic files but skip over them and read the remaining files anyway. The warning message should contain the number of bad files and the corresponding filenames, so the user can investigate.

hokiegeek2:
👍 agree, this would be a nice, time-saving enhancement

mhmerrill:
@glitch maybe you could tackle this in conjunction with #834

glitch:
One thing I noticed is that the client no longer calls the single `readhdf` -> `readhdfMsg` path; [see here](https://github.com/Bears-R-Us/arkouda/blob/91ba907e6c57798382ce324298f6661f4f5cad77/arkouda/pdarrayIO.py#L78).  It looks like that was changed in Feb2020, see commit e2324284d8204b032cea340ff6b9b71d68c90ee7

For now I'm going to mark the various code sections as DEPRECATED, but it looks like we can remove the server side `readhdfMsg` proc in GenSymIO along with the associated bits leading to it:

- arkouda_server.chpl -> cmd routing section
- GenSymIO.chpl -> readhdfMsg
- UnitTestStringType.chpl -> will need update to readAllHdf version
- io_test.py -> associated unit tests can be removed

If you think it should be removed in this round, let me know.

glitch:
In creating a client side unit test for this, I ran into an interesting problem.  The client `pdarrayIO.py::read_all` function makes a call to `pdarraIO.py::get_datasets` when the datasets option is `None`.  This in turn makes a server request `ls_hdf` on the first file.

`ls_hdf` throws a RuntimeError if it can't read the file, so it looks like we'll need to keep trying files until we can find a _good one_.

PR #866, opened 2021-07-02
Issue 821 ak.read_all option to warn on HDF5 file read errors

glitch:
Closes Issue #821 

### This PR adds an option to `ak.read_all` to allow file read errors instead of failing the call.

There are two major commits in this PR.

1. The first half adds the option to the cmd message on both the client side and the message processing component on the server side.  This boolean flag `allow_errors` signals to the server that it should not error out if one or more HDF5 file cannot be read due to conditions like file permissions, corruption, missing file, etc.
2. The second major commit changes the _response message_ sent back from the server into a json structure to allow a sample of the file error messages along with the total count of failed files.

*If we don't want to move to the json response message in this PR I can unwind that change, but it means we can't really return the error messages and total file error counts.*

### Additional components of interest

- The `ak.read_all` function also calls the `ls_hdf` component to retrieve the datasets, it was necessary to add another ls_hdf function to the client which was resilient in the face of file read errors.
- `hdfreadMsg` and corresponding functions.  This component was meant to read a single hdf5 file, however it is no longer in the chain of function calls.  The client now redirects to using the read_all component in the single file case.  The server side `readHdfMsg` proc can be removed, so I marked it as DEPRECATED.  (I can add that to this PR if desired or we can complete it in a separate Issue).

### JSON notes
The Chapel string formatting "%jt" construct doesn't really work for the `list` data structure so it was necessary to add a procedure and string constants to produce our own json.  I also tried various json structures and ended up with something I think is pretty succinct and clear.  It is documented in the `GenSymIO._buildReadAllHdfMsgJson` doc string with an example.

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -1,5 +1,5 @@
 from typeguard import typechecked
-import json, os
+import json, os, warnings
 from typing import cast, Dict, List, Mapping, Optional, Union
 from arkouda.client import generic_msg
 from arkouda.pdarrayclass import pdarray, create_pdarray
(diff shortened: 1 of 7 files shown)


mhmerrill (approved):
these changes look ok to me, I would like another concurring review from someone else.

pierce314159 (approved):

mhmerrill:
@glitch after merging a couple of other PRs there is now a conflict I'd like you to fix, thanks

glitch:
> @glitch after merging a couple of other PRs there is now a conflict I'd like you to fix, thanks

No problem, I'll have a look.

glitch:
Rebased with mainline to resolve conflicts.

PR #868, opened 2021-07-02
Closes #867: Remove dash character replacement in version stringf

### LINKED ISSUE #867: Version strings with '-' raise warnings
Version strings with a `-` character in them raise warnings when the client connects to the server.

This is due to version string manipulation happening [here](https://github.com/Bears-R-Us/arkouda/blob/master/src/ServerConfig.chpl#L93)

I removed the character replacement ( `-` to `.`) and all tests pass without issue (both chapel and python based tests).

I believe this character replacement for dash in the version string can be removed.

glitch:
Closes #867: Remove dash character replacement in version string from ServerConfig.


pierce314159 (approved):

mhmerrill (approved):
looks good

PR #870, opened 2021-07-02
Optimize client conversions for client-server array transfers

ronawho:
Improve the performance of `ak.array()` and `pdarray.to_ndarray()` by
optimizing how the client converts between bytes and numpy arrays. This
switches from `struct.{pack,unpack}` to `numpy.{tobytes,frombuffer}`
which has a non-trivial performance improvement. We have to be more
careful about endianness now, but the conversions are still pretty
straightforward. Note that the server currently always uses big-endian
for these conversions though a future optimization will look to use
native endianness.

Currently, the client code is still making some extra copies. In the
`to_ndarray()` case we need a copy because the underlying `bytes` buffer
is immutable, which makes the numpy array immutable so things that
modify the array in place like `numpy.array.sort()` would fail without
the copy. We should be able to optimize this by looking at the zmq frame
directly, but this will be future work. On the `ak.array()` side I think
`tobytes` copies, and I wonder if we can use `memoryview` here instead.
That too will be left for future work.

| before | 20.5 MiB/s | 23.5 MiB/s |
| after  | 32.5 MiB/s | 50.0 MiB/s |

While the rate increases don't look too large the time difference shows
the improvement a little better.

| before | 36.6s      | 31.2s      |
| after  | 23.6s      | 15.2s      |

Next steps will be to optimize the server side conversions more, which
should shave another 10-15 seconds off these operations which will give
more respectable rates.

Part of #794

### DIFF
--- a/arkouda/dtypes.py
+++ b/arkouda/dtypes.py
@@ -3,12 +3,13 @@ from enum import Enum
 import numpy as np # type: ignore
 from typeguard import typechecked
 import builtins
+import sys
 
 __all__ = ["DTypes", "DTypeObjects", "dtype", "bool", "int64", "float64", 
            "uint8", "str_", "check_np_dtype", "translate_np_dtype", 
            "resolve_scalar_dtype", "ARKOUDA_SUPPORTED_DTYPES", "bool_scalars",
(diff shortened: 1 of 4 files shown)


mhmerrill (approved):
these changes look ok to me.

PR #872, opened 2021-07-06
Fix MsgType.ERROR and MsgType.NORMAL MsgTuple prep/handing in IndexingMsg

### LINKED ISSUE #869: `a[mask] = b` fails silently when b and mask are incompatible
Reproducer:

```python
a = ak.zeros(10)
mask = ak.arange(10) < 5
b = ak.ones(10)
a[mask] = b  # Should raise ValueError but doesn't
a
array([0 0 0 0 0 0 0 0 0 0])
# Nothing set in a
```

The third line should raise a ValueError, because `mask` only has 5 `True`s and `b` has 10 values, which is illegal. However, instead of raising a ValueError, it just returns as if everything is ok, without changing `a` at all.

In the server log, the `repMsg` does start with `"Error: "`, so the client is not picking it up somehow.

hokiegeek2:
Fixes #869 by ensuring MsgTuple objects are created within helper message to correctly account for MsgType.NORMAL, MsgType.WARNING, and MsgType.ERROR

### DIFF
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -94,20 +94,20 @@ module IndexingMsg
             forall (elt,j) in zip(aa, slice) with (var agg = newSrcAggregator(t)) {
               agg.copy(elt,ea[j]);
             }
-            repMsg = "created " + st.attrib(rname);
+            var repMsg = "created " + st.attrib(rname);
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
I think  these changes look good.

Issue #871, opened 2021-07-02
Can `get_config()` be cached/memoized?


ronawho:
Currently `get_config()` always results in a server call, but the server config doesn't change so I think we could cache the results, at least for the current implementation. However, I'm not sure if this could create problems in the future so I wanted to open this issue to discuss.

glitch:
I looked through the client source code for `get_config()` and it's only really used in the following:

- `benchmarks` (13 times and in each case it is only once per file)
- `tests` (6 times, the `join_test.py` calls it 4 times which can probably be cleaned up).

Looking through the unit test usage we can clean up the 4 calls in `join_test`.

Looking at the benchmark testing, the get_config call is being invoked at the start of a function directly invoked by `__main__` where a fresh arkouda client is being stood up so I don't think caching is going to make a difference (a new python interpreter is being used each time).

Here's an example:  [benchmarks/IO.py](https://github.com/Bears-R-Us/arkouda/blob/master/benchmarks/IO.py)

main is called, client is created / connects to server, `time_ak_write_read` is called in main and calls `get_config` to retrieve the number of locales.  This is the pattern across all 13 of the benchmark cases.

ronawho:
Should have mentioned this, but I'm looking to add byteorder/endianness to get_config and I'll need to query endianness for each to_ndarray/ak.array() call and I was wondering if I should just cache endianness or if the entire server config should be cached.

e.g. what my current branch does is:

```python
@functools.lru_cache(maxsize=None)
def get_server_byteorder() -> str:
    from arkouda.client import get_config
    order = get_config()['byteorder']
    return cast('str', order)
```

and I was wondering if this caching should be moved up to the get_config level instead.

mhmerrill:
@ronawho We could cache a new copy each time we connect to the server.

glitch:
@ronawho  I put up [PR#873](https://github.com/Bears-R-Us/arkouda/pull/873) to implement the `@lru_cache` on get_config but also added in a default time expiration of 1 hour (you can see stack overflow post/answer https://stackoverflow.com/a/55900800 for the explanation).

This also returns the config object as part of the `ak.connect` call.  Let me know what you think.

ronawho:
Great, thanks! What's the motivation for the 1 hour expiration?

glitch:
> Great, thanks! What's the motivation for the 1 hour expiration?

[...]

PR #873, opened 2021-07-07
Issue #871: Adds caching of `get_config` response

glitch:
Issue #871 : Adds caching of `get_config` response by using lru_cache decorator with a ttl enhancement to add time expiration.
The client.connect() also updated to return the config object.

This uses the lru_cache from functools but adds one level of indirection to get_config so we can keep the client function call the same.

If the client gets disconnect and calls connect again, the cache is cleared so we can ensure up to date information from the server.

Note: The connect call now returns the `get_config` dictionary.


glitch:
Declining PR

PR #874, opened 2021-07-07
Memoize the server config on the client

### LINKED ISSUE #871: Can `get_config()` be cached/memoized?

ronawho:
Memoize the server's configuration on calls to `ak.connect()` so that
`ak.get_config()` can return the server configuration without any server
additional calls. The motivation for this is to be able to check the
server byteorder/endianness in an upcoming client-server array transfer
optimization without an additional server call.

This also changes the `Config/LocaleConfig` declarations on the server
side to be `const` to make it more obvious to future developers that
these values are not intended to change over the course of program
execution (and are therefor safe for the client to memoize.)

Closes #871
Related to #794

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -15,6 +15,7 @@ pspStr = ''
 context = zmq.Context()
 socket = context.socket(zmq.REQ)
 connected = False
+serverConfig = None
 # username and token for when basic authentication is enabled
 username = ''
 token = ''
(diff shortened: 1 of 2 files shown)


ronawho:
The server change are probably easiest to review by looking at the individual commits. The diff looks big-ish, but not much has changed.

glitch (approved):
👍

mhmerrill (approved):
these changes look good to me.

PR #880, opened 2021-07-12
Optimize server conversions for client-server array transfers

ronawho:
Improve the performance of `ak.array()` and `pdarray.to_ndarray()` by
optimizing how the server converts between bytes and pdarrays.
Previously, the server would write to and read from a big-endian memory
mapped file to convert between bytes and arrays but this is fairly
slow. Optimize this conversion by directly interpreting the underlying
memory as the type we're converting to. In the `ak.array()` case we
create the local array with `makeArrayFromPtr`, which will create an
array from the existing bytes without any copies. For the `to_ndarray()`
case `makeArrayFromPtr` is also used on some local memory, which is then
reinterpreted as bytes with `createBytesWithOwnedBuffer`.

| before |   33 MiB/s |   50 MiB/s |
| after  |  410 MiB/s |  175 MiB/s |

Where `ak.array()` is slower because there are more copies compared to
`to_ndarray()` on both the server and client side. Optimizing those
copies out is future work.

Part of #794

### DIFF
--- a/arkouda/dtypes.py
+++ b/arkouda/dtypes.py
@@ -8,14 +8,10 @@ import sys
 __all__ = ["DTypes", "DTypeObjects", "dtype", "bool", "int64", "float64", 
            "uint8", "str_", "check_np_dtype", "translate_np_dtype", 
            "resolve_scalar_dtype", "ARKOUDA_SUPPORTED_DTYPES", "bool_scalars",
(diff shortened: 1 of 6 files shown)


glitch (approved):
(caveat: For as much as I understand the chapel stuff) this looks good.

reuster986 (approved):
looks great!

PR #882, opened 2021-07-14
Rename Errors to ServerErrors

mppf:
Over on the main Chapel repository, we've moved a few modules around as part of module stabilization (see https://github.com/chapel-lang/chapel/pull/18028). As part of that we've added an included-by-default internal module called Errors. However, Arkouda was using a top-level module called Errors so that led to problems. This PR renames the Arkouda one to ServerErrors. This name was chosen to be symmetric with ServerErrorStrings.

(Longer term I think it would be reasonable for Arkouda to consist of an Arkouda module and then submodules beneath that, including in different files. The module system should support having a submodule named Errors along with the toplevel (internal) one - the naming conflict here is specific to top-level modules).

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -26,7 +26,7 @@ module ArgSortMsg
     use RadixSortLSD;
     use SegmentedArray;
     use Reflection;
-    use Errors;
+    use ServerErrors;
     use Logging;
     use Message;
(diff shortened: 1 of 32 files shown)


ronawho (approved):

mhmerrill (approved):
This looks￼ ok to me.

PR #885, opened 2021-07-22
Optimize the number of copies for client-server array transfers

ronawho:
Reduce the number of copies for `ak.array()` and `pdarray.to_ndarray()`.
On the client side use non-copying sends/recvs that take/return a
memoryview. For `ak.array` send the binary payload in a separate message
to avoid concatenating (which creates a copy) on the client side and
splitting (which also creates a copy ) on the server side. For
`pdarray.to_ndarray()` on the server side move the binaryRepMsg
declaration to avoid an extra copy that was happening.

Here's the performance impact for 16-node-xc:

| before |  410 MiB/s |  175 MiB/s |
| after  |  715 MiB/s |  705 MiB/s |

Part of #794

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -295,8 +295,8 @@ def _start_tunnel(addr : str, tunnel_server : str) -> Tuple[str,object]:
     except Exception as e:
         raise ConnectionError(e)
 
-def _send_string_message(cmd : str, recv_bytes : bool=False, 
-                                   args : str=None) -> Union[str, bytes]:
+def _send_string_message(cmd : str, recv_binary : bool=False,
+                         args : str=None) -> Union[str, bytes]:
     """
     Generates a RequestMessage encapsulating command and requesting
     user information, sends it to the Arkouda server, and returns 
@@ -306,16 +306,15 @@ def _send_string_message(cmd : str, recv_bytes : bool=False,
     ----------
     cmd : str
         The name of the command to be executed by the Arkouda server
-    recv_bytes : bool, defaults to False
-        A boolean indicating whether the return message will be in bytes
-        as opposed to a string
+    recv_binary : bool, defaults to False
+        Indicates if the return message will be a string or binary data
     args : str
         A delimited string containing 1..n command arguments
 
     Returns
     -------
     Union[str,bytes]
-        The response string or byte array sent back from the Arkouda server
+        The response string or binary data sent back from the Arkouda server
         
     Raises
     ------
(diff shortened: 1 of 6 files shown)


ronawho:
Probably easiest to review commit at a time. 

This optimizes out most unnecessary copies. There are some remaining in the Chapel ZMQ shim that we're looking at optimizing out, but on the Arkouda side the only "extra" copy is for to_ndarray when the memoryview returned to us is readonly (which is will be true for most pyzmq backends) at https://github.com/Bears-R-Us/arkouda/pull/885/files#diff-552f96acc4e4316d5d653916e1902f8f15ef77473e864e605c975aee228bbbb1R867. I plan on opening an issue against pyzmq to see if the view has to be readonly, but an alternative to this would be to make `to_ndarray()` return a readonly array and require users to create a copy if they want to modify it.

reuster986 (approved):
looks great!

arkouda/client.py
```diff
@@ -539,14 +537,14 @@ def generic_msg(cmd : str, args : Union[str,bytes]=None, send_bytes : bool=False
...
+            assert payload is not None
+            return _send_binary_message(cmd=cmd, payload=payload,
+                                        recv_binary=recv_binary, args=args)
         else:
-            return _send_string_message(cmd=cmd, args=cast(str,args), 
-                                            recv_bytes=recv_bytes)
-                
+            assert payload is None
+            return _send_string_message(cmd=cmd, args=args, recv_binary=recv_binary)
+
```
glitch, line 547: Minor: instead of the assertions, the payload conditions could be moved into the if/else conditional logic and a third case could be added to handle the assertion failures with warning messages.
i.e.
```
if send_binary and payload is not None:
  return _send_binary....
elif not send_binary and payload is None:
  return _send_string...
else:
  warnings.warn(..you forgot something...)
```

src/arkouda_server.chpl
```diff
@@ -285,17 +288,18 @@ proc main() {
...
 
             select cmd
             {
-                when "array"             {repTuple = arrayMsg(cmd, payload, st);}
-                when "tondarray"         {binaryRepMsg = tondarrayMsg(cmd, args, st);}
+                when "array"             {repTuple = arrayMsg(cmd, args, payload, st);}
+                when "tondarray"         {
+                  var binaryRepMsg = tondarrayMsg(cmd, args, st);
+                  sendRepMsg(binaryRepMsg);
+                }
```
glitch, line 302: Trivial/Stylistic: If you wanted to keep with the flow of the code here, it looks like you can make that a one-liner
```
sendRepMsg(tondarrayMsg(cmd, args, st));
```

arkouda/client.py
```diff
@@ -327,21 +326,19 @@ def _send_string_message(cmd : str, recv_bytes : bool=False,
...
-        elif return_message.startswith(b"Warning:"): 
-            warnings.warn(return_message.decode())
-        return return_message
+    if recv_binary:
+        frame = socket.recv(copy=False)
+        view = frame.buffer
+        # raise errors sent back from the server
+        if bytes(view[0:len(b"Error:")]) == b"Error:":
+            raise RuntimeError(frame.bytes.decode())
+        return view
```
glitch, line 341: I noticed handling of "Warning" was dropped here (as well as later on ~line 400-409).  I haven't really looked into it but were we never actually returning a Warning string so the handling was never reached?

glitch (approved):
Non-binding approval :) :+1:

arkouda/client.py
ronawho: Noted in https://github.com/Bears-R-Us/arkouda/pull/885/commits/32eba00ba480648ff251e4a48840a403e6f32aa1, but this code never really supported warnings. We could warn here, but then would later fail to convert the binary message to numpy array because the raw buffer would still have the warning. We'd need some way to separate the warning from the binary message and for now I decided to just remove the warning capability since it isn't used.

src/arkouda_server.chpl
ronawho: I have no strong preference here. I went with the current approach to make it stick out a little more since it is doing something different from every other control flow path, but I'd be fine switching to this if others prefer.

ronawho:
> I didn't know about the memoryview stuff, it looks pretty cool 

Agreed, this was the first I learned of it too. These are some of the links I stashed away that I found the most helpful:
 - https://docs.python.org/3/c-api/buffer.html
 - https://www.python.org/dev/peps/pep-3118/
 - https://julien.danjou.info/high-performance-in-python-with-zero-copy-and-the-buffer-protocol/
 - https://stackoverflow.com/questions/18655648/what-exactly-is-the-point-of-memoryview-in-python
 - https://pyzmq.readthedocs.io/en/latest/serialization.html

PR #890, opened 2021-07-28
Closes #848 Limit Length of Logging Lines:

### LINKED ISSUE #848: Limit Length of Logging Lines
Above alliteration is assuredly accidental.

This is related to the error in #847 . I speculate that the cause of that error was that the logger was trying to write several thousand filenames to the terminal and immediately flush stdout. While #847 aims to tighten up error handling for logging, I think we should also impose a limit on the size of messages that get logged, similar to how we do not log the binary data that gets transferred with `ak.array()` and `.to_ndarray()`. The code that logs incoming commands and their args is here: 

https://github.com/Bears-R-Us/arkouda/blob/e9053141904bbb462f29a6ada346272692d9d15e/src/arkouda_server.chpl#L262-L274

I propose we set a threshold for the length of `args`, and if it exceeds that threshold, only print out [first N characters `...` last N chars]. This will not only help avoid logging errors, but will also make the logs more readable.

pierce314159:
Closes #848 Limit Length of Logging Lines:
- Adds a threshold, `n`, that caps the length of `args` being logged to `2*n`. Whens `args.size > 2*n`, it will be logged as `[first n characters]...[last n characters]`

I had no idea what `n` should be. I made it 12 for no reason in particular other than ease of testing (and it made my example below look nice hahaha). I'm certainly open to suggestions for different values

```python
# client commands:
pd = ak.array([1,2,3])
pd.register('reallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallylongname')

# server logs before PR:
2021-07-28:13:41:35 [arkouda_server] main Line 268 INFO [Chapel] >>> "register" "id_1 reallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallylongname"
...
2021-07-28:13:41:35 [arkouda_server] main Line 268 INFO [Chapel] >>> "repr" "reallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallylongname 100"

# server logs after PR:
2021-07-28:15:30:02 [arkouda_server] main Line 271 INFO [Chapel] >>> "register" "id_1 reallyr...allylongname"
...
2021-07-28:15:30:02 [arkouda_server] main Line 271 INFO [Chapel] >>> "repr" "reallyreally...longname 100"
```

### DIFF
--- a/src/arkouda_server.chpl
+++ b/src/arkouda_server.chpl
@@ -265,8 +265,11 @@ proc main() {
             if (trace) {
               try {
                 if (cmd != "array") {
+                  var n: int = 12;
+                  // if args.size > 2*n, only first n characters and last n characters are logged
+                  var short_args: string = if args.size > 2*n then args[0..#n]+'...'+args[args.size-n..#n] else args;
                   asLogger.info(getModuleName(), getRoutineName(), getLineNumber(),
-                                                     ">>> %t %t".format(cmd, args));
+                                                     ">>> %t %t".format(cmd, short_args));
                 } else {
                   asLogger.info(getModuleName(), getRoutineName(), getLineNumber(),
                                                      ">>> %s [binary data]".format(cmd));


src/arkouda_server.chpl
```diff
@@ -265,13 +265,24 @@ proc main() {
             if (trace) {
               try {
                 if (cmd != "array") {
+                  var size: int = args.size;
+                  var n: int = 12;
+                  // if args.size > 2*n, only first n characters and last n characters are logged
+                  if (size > 2*n) {
+                    asLogger.info(getModuleName(), getRoutineName(), getLineNumber(),
+                                  ">>> %t %t...%t".format(cmd, args[0..#n], args[size-n..#n]));
+                  }
```
pierce314159, line 274: Note: I can change this to the following 
```python
if (size > 2*n) {
  asLogger.info(getModuleName(), getRoutineName(), getLineNumber(),
                         ">>> %t %t".format(cmd, args[0..#n]+'...'+args[size-n..#n]));
}
```
which results in the logging output being only 2 arguments instead of 3

```python
pd = ak.array([1,2,3])
pd.register('reallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallylongname')

2021-07-28:14:55:41 [arkouda_server] main Line 272 INFO [Chapel] >>> "register" "id_1 reallyr...allylongname"
...
2021-07-28:14:55:41 [arkouda_server] main Line 272 INFO [Chapel] >>> "repr" "reallyreally...longname 100"
```

pierce314159: I changed this section to have this functionality and minimize diff with master by using a ternary declaration of `short_args` and eliminating the need for an extra if-else

mhmerrill (commented):
I am unsure about limiting all logging lines because we parse these lines with a script sometimes. Maybe a intermediate solution is to limit the number of filenames from the I/O calls. I am open to discussion about this.

pierce314159:
> 
> 
> I am unsure about limiting all logging lines because we parse these lines with a script sometimes. Maybe a intermediate solution is to limit the number of filenames from the I/O calls. I am open to discussion about this.

Makes sense to me! I implemented it this way based on @reuster986's suggestion in #848. I've got no strong preferences, but it doesn't seem great to break the log parsing scripts. I wonder what @reuster986 thinks since he posted the issue for this

reuster986:
Sorry for the delayed reply; lost track of PRs while on vacation!

We do parse logs and harvest array names from command arguments to track input data size to commands, so I think I would like to keep those intact. When I wrote the issue, I mostly had in mind filenames for I/O commands, so I would be inclined to go the route of just limiting those. Sorry for not being more specific in the issue.

pierce314159:
Closing this PR and will open a new one to limit number of filenames that get logged

PR #891, opened 2021-07-28
Increase the number of trials for array_transfer

ronawho:
With recent optimizations this is fast enough to gather multiple trials.
gc.collect between trials to limit memory usage on the client side.

### DIFF
--- a/benchmarks/array_transfer.py
+++ b/benchmarks/array_transfer.py
@@ -3,6 +3,7 @@
 import time, argparse
 import numpy as np
 import arkouda as ak
+import gc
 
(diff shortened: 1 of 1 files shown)


ronawho:
Note that the first transfer is much slower because there's large allocations on the kernel side and in the ZMQ library, so it may also be beneficial to do a warmup trial first to capture the speed of transfers in the steady state. But none of the other benchmarks do a warmup trial so I wasn't sure if it was odd to do that here.

glitch (approved):

mhmerrill (approved):
Looks ok to me

PR #893, opened 2021-08-02
Further reduce the number of copies for client-server array transfers

ronawho:
Eliminate a copy for `ak.array()` by ensuring that copy elision fires
when recv'ing the binary payload.

Here's the performance impact for 16-node-xc:

| before |  715 MiB/s |  705 MiB/s |
| after  |  715 MiB/s |  750 MiB/s |

Part of #794


mhmerrill (approved):
looks ok to me

PR #897, opened 2021-08-10
Closes #855 Add mem checks for scans

### LINKED ISSUE #855: Add mem checks for scans
I was seeing a fair number of server crashes due to out-of-memory while performing scans, so I decided to investigate. A quick test with `benchmarks/scan.py` reveals that, on a system with 128 core, 2TB nodes, any value of `--size` above 48 billion (approx. 1/6 of available memory) will crash the server. Excluding the input and output arrays, this implies that the scan is using about 4x the input array size in working memory. @bradcray and @ronawho , does that sound right?

In any case, we should find all the scans in the source code and put in an `overMemLimit()` check, or bump the size of existing checks.

pierce314159:
Closes #855 Add mem checks for scans:
- Adds `overMemLimit` check before `scan` operations
- Adds `throws` to the function signature of methods with newly added `overMemLimit` checks and to function signatures of any methods which call them
- Rearranges `cumsum` and `cumprod` to avoid unnecessary copies

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -305,7 +305,7 @@ module ArgSortMsg
       return new MsgTuple(repMsg, MsgType.NORMAL);
     }
     
-    proc argsortDefault(A:[?D] ?t):[D] int {
+    proc argsortDefault(A:[?D] ?t):[D] int throws {
       var t1 = Time.getCurrentTime();
       //var AI = [(a, i) in zip(A, D)] (a, i);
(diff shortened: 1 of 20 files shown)


glitch (approved):

glitch:
Note: I resolved the merge conflict with master (due to merge of PR#901 / Issue#900).  It was a whitespace issue and was resolved using github's web ui.

reuster986 (approved):

PR #899, opened 2021-08-12
Closes #822 : Adds save to hdf5 functionality for Categorical

### LINKED ISSUE #822: Support writing `ak.Categorical` to HDF5
Add an `ak.Categorical.save()` method, similar to `ak.Strings.save()`. The method signature should be:

```python
def save(self, prefix_path : str, name : str='categorical_array', mode : str='truncate')
```

And the layout of the resulting HDF5 file should look like:

```
<name>               Group
    categories       Group
        segments     Dataset (int64)
        values       Dataset (uint8)
    codes            Dataset (int64)
```

I haven't tried yet, but I think this could be accomplished by creating a HDF5 group in Chapel with the user-specified name, then in Python calling 
```python
cat.categories.save(prefix_path+'/'+name, dataset='categories')
cat.codes.save(prefix_path+'/'+name, dataset='codes')
```

glitch:
Closes #822 Support writing `ak.Categorical` to HDF5

This PR implements a `save` function for `Categorical` in the same vein as `pdarray` and `Strings` save implementations.
The saved HDF5 file(s) can be read back in using the standard `ak.load_all` call.

UPDATE: The `ak.load_all` call will now detect & automatically create the Categorical object stored in HDF5.  Based on @reuster986 's feedback you can now _round trip_ a pseudo dataframe (see the unit test `categorical_test.py:testSaveAndLoadCategoricalMulti` for a full example)

i.e.
```python
df = {'pda1': ak.arange(5),
      'str1': ak.array(['one', 'two', 'three', 'two', 'one']),
      'cat1': ak.Categorical(ak.array(['a', 'b', 'c', 'b', 'a']))}
ak.save_all(df, path)
df2 = ak.load_all(path)
```
Note: I also took this opportunity to fix some white-space and documentation for functions/procs in the call chain.

Note2: This also handles multiple categorical objects stored in the same HDF5 file as well as separate pdarray and Strings objects.

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -876,7 +876,7 @@ class Categorical:
         """
         removal_names: List[str] = []
         groups: DefaultDict[str, List[str]] = defaultdict(list)
-        result_categoricals = {}
+        result_categoricals: Dict[str, Categorical] = {}
         for k in d.keys():  # build dict of str->list[components]
             if "." in k:
                 groups[k.split(".")[0]].append(k)
@@ -893,7 +893,7 @@ class Categorical:
                     if part.endswith(".categories"):
                         base_name = ".".join(part.split(".categories")[0:-1])
 
-                # Construct categorical add it to the return_categoricals under the parent name
+                # Construct categorical and add it to the return_categoricals under the parent name
                 result_categoricals[base_name] = Categorical.from_codes(**cat_parts)
 
         return removal_names, result_categoricals
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1,13 +1,15 @@
 from __future__ import annotations
-from typing import cast, List, Optional, Sequence, Union, Dict
-import numpy as np # type: ignore
+from typing import cast, List, Optional, Sequence, Union, Dict, Mapping, Tuple, DefaultDict
+import numpy as np  # type: ignore
 import itertools
+from collections import defaultdict
 from typeguard import typechecked
+from arkouda.client import generic_msg
 from arkouda.strings import Strings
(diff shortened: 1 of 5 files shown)


pierce314159 (approved):

glitch:
@reuster986 if you get a chance can you check this one over?  It's for saving categoricals to HDF5.  The only real change is the addition of the `save` function to `categorical.py`.  Everything else is just whitespace cleanup, import cleanup, doc strings, or unit tests.

reuster986 (changes requested):
@glitch the mechanics look good, but I would like the `load` API to auto-detect Categoricals, so that the user can round-trip DataFrames without having to explicitly call `ak.Categorical()`.

For example,

```python
df = {'int': ak.arange(5),
      'str': ak.array(['one', 'two', 'three', 'two', 'one']),
      'cat': ak.Categorical(ak.array(['a', 'b', 'c', 'b', 'a']))}
ak.save_all(df, path)
df2 = ak.load_all(path)
```

`df2` should be equivalent to `df` and should have column types `pdarray(int64)`, `Strings`, and `Categorical`. My thought in proposing the particular directory structure I did in #822 was that it would enable auto-detection of Categoricals vs. Strings via the filesystem. Alternatively, HDF5 group/dataset attributes could be used for this purpose, if that seems better.

glitch:
@reuster986 Sounds good 👍  Just a head's up I'm on vacation for a few days so it'll be a bit, but I'll ping you when the changes are ready.   Thanks for the feedback!

reuster986:
@glitch enjoy your vacation!

reuster986 (approved):
looks great!

PR #901, opened 2021-08-17
Closes #900 : bugfix on groupby with zero length array.

### LINKED ISSUE #900: Duplicate pdarray return name causing some "unknown symbol" errors
https://github.com/Bears-R-Us/arkouda/blob/6ecd185cb695a13c3b5e43e3d6e80285e7afd9c5/src/FindSegmentsMsg.chpl#L107

The `findSegments` command is supposed to return two pdarrays (segments and unique_key_indices), but in the case of zero-length input, the above code accidentally returns the attributes of the same array twice. In the client, these both get instantiated as pdarray objects with the same server-side name, but one of them goes out of scope fairly quickly and calls "delete" on the server side. Meanwhile, the other array gets saved as the `.segments` attribute of the `GroupBy`, and the next time it is referenced, the server complains about an unknown symbol.

Here's a reproducer:
```python
g = ak.GroupBy(ak.zeros(0, dtype=ak.int64))
g.segments
[unknown symbol error]
```

I suspect that our other elusive unknown symbol errors have a similar cause. Maybe we should create a DEBUG mode for `create_pdarray()` or `pdarray.__init__()` that checks whether the name is already used in the server? This check should probably not be turned on by default, since it would add a lot of overhead to array creation, but maybe just for the CI, or so that we can spot check.

glitch:
Closes #900 : bugfix on groupby with zero length array.
Also fixes indentation issues in findSegmentsMsg.

Note: The actual fix is a single character change on line 107 of FindSegmentsMsg
```
repMsg = "created " + st.attrib(n1) + " +created " + st.attrib(n2);
```

The rest of the changes are purely white-space to fix bad indentation in the proc.

### DIFF
--- a/src/FindSegmentsMsg.chpl
+++ b/src/FindSegmentsMsg.chpl
@@ -58,43 +58,43 @@ module FindSegmentsMsg
         }        
         // var keyEntries: [0..#nkeys] borrowed GenSymEntry;
         for (name, objtype, i) in zip(knames, ktypes, 1..) {
-      // var g: borrowed GenSymEntry;
-      var thisSize: int;
-      var thisType: DType;
-      select objtype {
-          when "pdarray" {
-              var g = st.lookup(name);
-              thisSize = g.size;
-              thisType = g.dtype;
+          // var g: borrowed GenSymEntry;
+          var thisSize: int;
+          var thisType: DType;
+          select objtype {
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):

reuster986 (approved):
Thanks for including a test!

PR #905, opened 2021-08-31
Closes #904 : Fixes issue with tmp dir where /tmp might not be mounted

### LINKED ISSUE #904: categorical_test.py failures on distributed systems
https://github.com/Bears-R-Us/arkouda/pull/899 added some tests that write files to `/tmp`, but `/tmp` is usually a filesystem that's local to each machine. For clusters, including the ones we do nightly testing on, this is causing some failures for those tests since the results written by the server aren't visible to the clients.

This is a similar problem to https://github.com/Bears-R-Us/arkouda/issues/437, which was resolved by https://github.com/Bears-R-Us/arkouda/pull/439 by writing the files to the current working directory.

The following patch gets testing working on these systems:

```diff
diff --git a/tests/categorical_test.py b/tests/categorical_test.py
index 65dacb1..01de0f5 100644
--- a/tests/categorical_test.py
+++ b/tests/categorical_test.py
@@ -2,6 +2,7 @@ import numpy as np
 import tempfile
 from context import arkouda as ak
 from base_test import ArkoudaTest
+import os


 class CategoricalTest(ArkoudaTest):
@@ -194,7 +195,7 @@ class CategoricalTest(ArkoudaTest):
         with self.assertRaises(ValueError):  # Expect error for mode not being append or truncate
             cat.save("foo", dataset="bar", mode="not_allowed")

-        with tempfile.TemporaryDirectory() as tmp_dirname:
+        with tempfile.TemporaryDirectory(dir=os.getcwd()) as tmp_dirname:
             dset_name = "categorical_array"  # name of categorical array

             # Test the save functionality & confirm via h5py
@@ -233,7 +234,7 @@ class CategoricalTest(ArkoudaTest):
         pda1 = ak.zeros(5)
         strings1 = ak.random_strings_uniform(9, 10, 5)

-        with tempfile.TemporaryDirectory() as tmp_dirname:
+        with tempfile.TemporaryDirectory(dir=os.getcwd()) as tmp_dirname:
             df = {
                 "cat1": c1,
                 "cat2": c2,
```

But I'm not sure if you want to go with something like that or if you want to mirror what was done in https://github.com/Bears-R-Us/arkouda/pull/439 .

glitch:
Closes #904 : Fixes issue with tmp dir on test systems where /tmp may not be mounted.

This solves the issue the same way as it was done in Issue #437 (tests/io_test.py :: IOTest class).

Note: I also confirmed that it cleans up after itself when I ran it locally.

### DIFF
--- a/tests/categorical_test.py
+++ b/tests/categorical_test.py
@@ -1,10 +1,20 @@
 import numpy as np
+import glob
+import shutil
+import os
 import tempfile
+from arkouda import io_util
 from context import arkouda as ak
 from base_test import ArkoudaTest
 
 
 class CategoricalTest(ArkoudaTest):
+
+    @classmethod
+    def setUpClass(cls):
+        super(CategoricalTest, cls).setUpClass()
+        CategoricalTest.cat_test_base_tmp = '{}/categorical_test'.format(os.getcwd())
+        io_util.get_directory(CategoricalTest.cat_test_base_tmp )
     
     def _getCategorical(self, prefix : str='string', size : int=11) -> ak.Categorical:
         return ak.Categorical(ak.array(['{} {}'.format(prefix,i) for i in range(1,size)]))
@@ -194,7 +204,7 @@ class CategoricalTest(ArkoudaTest):
         with self.assertRaises(ValueError):  # Expect error for mode not being append or truncate
             cat.save("foo", dataset="bar", mode="not_allowed")
 
-        with tempfile.TemporaryDirectory() as tmp_dirname:
+        with tempfile.TemporaryDirectory(dir=CategoricalTest.cat_test_base_tmp) as tmp_dirname:
             dset_name = "categorical_array"  # name of categorical array
 
             # Test the save functionality & confirm via h5py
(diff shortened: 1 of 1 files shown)


tests/categorical_test.py
```diff
@@ -1,9 +1,15 @@
 import numpy as np
+import os
 import tempfile
+import io_util
 from context import arkouda as ak
 from base_test import ArkoudaTest
 
 
+cat_test_base_tmp = '{}/categorical_test'.format(os.getcwd())
+io_util.get_directory(cat_test_base_tmp)
```
ronawho, line 10: I think the main thing `get_directory` does is create a path if it doesn't exist, and `TemporaryDirectory` should do that already so I'm not sure if this is needed.

glitch: It bombed on me when I didn't call `get_directory`.  It's really because the `TemporaryDirectory` wrapper is creating another directory under it.  However, now that I look at it, it doesn't clean up the sub dir like I expected it to.  I'm going to postpone this one for a bit until I can get a cleaner approach.

ronawho (approved):
👍  Passes testing on the machine that originally had the issue.

pierce314159 (approved):

PR #913, opened 2021-09-08
Use --interleave-memory if it's available

ronawho:
chapel-lang/chapel#18350 added an `--interleave-memory` feature to
reduce the performance hit incurred by a commit that reduced memory
fragmentation under configurations like gasnet-ibv. If the chapel
compiler/build supports the option, automatically use it for the Arkouda
build. This improves performance for gasnet-ibv segment large (the
default) by interleaving memory allocations to avoid getting
progressively worse NUMA affinity the longer the arkouda server has been
running.

See chapel-lang/chapel#18286 for more details

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,14 @@ ifdef ARKOUDA_HDF5_PATH
 $(eval $(call add-path,$(ARKOUDA_HDF5_PATH)))
 endif
 
+# For configs that use a fixed heap, but still have first-touch semantics
+# (gasnet-ibv-large) interleave large allocations to reduce the performance hit
+# from getting progressively worse NUMA affinity due to memory reuse.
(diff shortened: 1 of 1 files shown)


ronawho:
Note that this option is being tested in the nightly config manually and this is just enabling it automatically. 

https://chapel-lang.org/perf/arkouda/16-node-cs-hdr/?startdate=2021/08/03&enddate=2021/09/09&configs=release,nightly&suite=benchmarks has the performance graphs. On 08/27 is when the fragmentation reduction commit went in. That increased usable memory at the expense of performance, and the interleave option went in last night to limit the fragmentation performance impact. 

Generally speaking performance is behind the old peak that we had from other chapel improvements since 1.24.1, but most operations are still ahead of where they were with 1.24.1.

reuster986 (approved):

Issue #914, opened 2021-09-08
Categorical/Strings array with size smaller than num locales fails read/write to HDF5


glitch:
As identified in issue #904 we have an issue where both Strings & Categoricals appear to fail reading / writing to HDF5 when the size of the array is less than the number of locales.

This can be produced by modifying the unit test `tests/categorical_test.py::CategoricalTest::testSaveAndLoadCategorical` and setting the number of elements to something smaller than the number of locales.

Note: It is probably best to make an explicit unit test to cover this issue.
Reproducible via:
```python
# mkdir -p /tmp/junk
# ./arkouda_server -nl 2 --logLevel=LogLevel.DEBUG --GenSymIO.logLevel=LogLevel.DEBUG

import arkouda as ak
ak.connect("localhost")
c1 = ak.Categorical(ak.array(["abc"]))
c1.save("/tmp/junk/cat", dataset="c1_array")
x = ak.load_all("/tmp/junk/cat")
c2 = x["c1_array"]
```

NOTE: To enable more than one locale for local testing you need to turn on the following environment variables when compiling and running (see also chapel-lang docs for running multi-locale https://chapel-lang.org/docs/usingchapel/multilocale.html#readme-multilocale )
```
export CHPL_COMM=gasnet
export CHPL_COMM_SUBSTRATE=smp
export CHPL_RT_OVERSUBSCRIBED=yes

# It's a good idea to also turn on
export ARKOUDA_DEVELOPER=true
```

glitch:
What's interesting is that I can get the same operation to fail in different places depending on whether or not `ARKOUDA_DEVELOPER` is turned on.  I believe the issue is still the same underlying problem (elements < num_locales), but it fails trying to write out Categorical.codes (AK_DEV=true) and Categorical.categories (AK_DEV=false).  These are separate areas in the code but in general we are trying to roughly the same thing.

glitch:
Based on the current master branch these are the fundamental breaking conditions
```python
import arkouda as ak
ak.connect("localhost")
a1 = ak.array([1, 2])
a1.save("/tmp/junk/a1")
>> src/GenSymIO.chpl:1581: error: halt reached - array index out of bounds
```

```python
import arkouda as ak
ak.connect("localhost")
s1 = ak.array(["abc", "xyz"])
s1.save("/tmp/junk/s1")
>> src/GenSymIO.chpl:1856: error: halt reached - back called on an empty array
```

[...]

PR #915, opened 2021-09-08
Closes #904: Temporary fix for Categorical unit test 

### LINKED ISSUE #904: categorical_test.py failures on distributed systems

glitch:
Closes #904: Temporary fix for unit test
This fixes the boundary problem for the categorical unit test by increasing the total number of elements in the strings array, however it does not address the underlying problem; see issue #914

This temporary fix is so nightly testing can be resumed.

### DIFF
--- a/tests/categorical_test.py
+++ b/tests/categorical_test.py
@@ -200,7 +200,8 @@ class CategoricalTest(ArkoudaTest):
         """
         Test to save categorical to hdf5 and read it back successfully
         """
-        cat = self._getCategorical()
+        num_elems = 51  # _getCategorical starts counting at 1, so the size is really off by one
+        cat = self._getCategorical(size=num_elems)
         with self.assertRaises(ValueError):  # Expect error for mode not being append or truncate
             cat.save("foo", dataset="bar", mode="not_allowed")
 
@@ -222,8 +223,7 @@ class CategoricalTest(ArkoudaTest):
             self.assertTrue(dset_name in x)
             cat_from_hdf = x[dset_name]
 
-            expected_categories = ['string 1', 'string 2', 'string 3', 'string 4', 'string 5', 'string 6', 'string 7',
-                                   'string 8', 'string 9', 'string 10']
+            expected_categories = [f"string {i}" for i in range(1, num_elems)]
 
             # Note assertCountEqual asserts a and b have the same elements in the same amount regardless of order
             self.assertCountEqual(cat_from_hdf.categories.to_ndarray().tolist(), expected_categories)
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

tests/categorical_test.py
```diff
@@ -222,24 +223,24 @@ def testSaveAndLoadCategorical(self):
...
         """
         Test to build a pseudo dataframe with multiple categoricals, pdarrays, strings objects and successfully
         write/read it from HDF5
         """
-        c1 = ak.Categorical(ak.array(['abc', 'def', 'ghi', 'jkl']))
-        c2 = ak.Categorical(ak.array(['mno', 'pqr', 'stu', 'vwx', 'yz']))
+        c1 = self._getCategorical(prefix="c1", size=51)
+        c2 = self._getCategorical(prefix="c2", size=52)
         pda1 = ak.zeros(5)
         strings1 = ak.random_strings_uniform(9, 10, 5)
```
ronawho, line 245: This test fails for me at 6+ locales now instead of 5+. Guessing it's because `pda1`/`strings1` are of length 5. Bumping those to 52 (no idea if that makes sense for a size) seems to work at 16 nodes.

glitch: Yup, you are correct, I forgot to bump all of the things that go in the map; tunnel vision doh!

glitch: Ok, sorry for the delay, I pushed the additional changes to increase the number of elements in those two arrays.

ronawho (approved):

Issue #917, opened 2021-09-09
`substring_search` benchmark is timing out in nightly performance run


pierce314159:
@ronawho found that the `substring_search` benchmark is timing out in the nightly performance runs. He's offered to help track down the performance bottleneck later next week. He recommends switching the default problem size from `10**8` to `10**6` for the time being

Here are the times for `10**6` on 16-node-cs-hdr:

```
>>> arkouda substring search
numLocales = 16, N = 1,000,000
non-regex with literal substring Average time = 0.1250 sec
regex with literal substring Average time = 57.8762 sec
regex with pattern Average time = 57.1415 sec
non-regex with literal substring Average rate = 0.1415 GiB/sec
regex with literal substring Average rate = 0.0003 GiB/sec
regex with pattern Average rate = 0.0003 GiB/sec
```

ronawho's original comment:
https://github.com/Bears-R-Us/arkouda/issues/912#issuecomment-916030364

ronawho:
I haven't been able to dive too deep into this, but here's what I found so far. I ran substring_search on chapcs (single node, comm=none) and I don't see way way better performance there, which makes me suspect this is more about local overheads than inherent comm limitations (though there could be some comm overhead still.)

On chapcs here's the performance I see:

```
>>> arkouda substring search
numLocales = 1, N = 1,000,000
non-regex with literal substring Average time = 0.0463 sec
regex with literal substring Average time = 9.5469 sec
regex with pattern Average time = 9.4727 sec
non-regex with literal substring Average rate = 0.3823 GiB/sec
regex with literal substring Average rate = 0.0019 GiB/sec
regex with pattern Average rate = 0.0019 GiB/sec
```

Looking at the current code there are a fair number of memory allocations so I wonder if the overhead is mostly from lots of short-lived allocations similar to what we saw in https://github.com/Bears-R-Us/arkouda/issues/266#issuecomment-578420803. There's allocations to slice the segmented array, and to create the string from the slice.

The below patch eliminates some of the copies. It's terrible since it only works for single node, but it should be good enough to do some experiments:

<details>

```diff
diff --git a/src/CastMsg.chpl b/src/CastMsg.chpl
index 621f09f..54e40d8 100644
--- a/src/CastMsg.chpl
+++ b/src/CastMsg.chpl
@@ -203,7 +203,7 @@ module CastMsg {
               } else {
                    end = oa[i+1] - 1;
               }
-              e = interpretAsString(va[start..end]) : toType;
[...]

PR #921, opened 2021-09-14
Issue #917: Adds aggregators to regex substring search methods

pierce314159:
This PR (Issue #917):
- Adds aggregators to the regex substring search functions with the goal of improving performance

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -507,31 +507,31 @@ module SegmentedArray {
 
       select mode {
         when SearchMode.contains {
-          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern)) {
+          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern), var agg = newDstAggregator(bool)) {
             // regexp.search searches the receiving string for matches at any offset
-            h = myRegex.search(interpretAsString(va[o..#l])).matched;
+            agg.copy(h, myRegex.search(interpretAsString(va[o..#l])).matched);
           }
         }
         when SearchMode.startsWith {
-          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern)) {
+          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern), var agg = newDstAggregator(bool)) {
             // regexp.match only returns a match if the start of the string matches the pattern
-            h = myRegex.match(interpretAsString(va[o..#l])).matched;
+            agg.copy(h, myRegex.match(interpretAsString(va[o..#l])).matched);
           }
         }
         when SearchMode.endsWith {
-          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern)) {
+          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern), var agg = newDstAggregator(bool)) {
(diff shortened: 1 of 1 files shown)


src/SegmentedArray.chpl
```diff
@@ -507,31 +507,31 @@ module SegmentedArray {
 
       select mode {
         when SearchMode.contains {
-          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern)) {
+          forall (o, l, h) in zip(oa, lengths, hits) with (var myRegex = _unsafeCompileRegex(pattern), var agg = newDstAggregator(bool)) {
             // regexp.search searches the receiving string for matches at any offset
-            h = myRegex.search(interpretAsString(va[o..#l])).matched;
+            agg.copy(h, myRegex.search(interpretAsString(va[o..#l])).matched);
```
reuster986, line 512: The way this `forall` and `zip` is structured I believe already guarantees that `h` is local, so I don't think aggregation will help here, although @ronawho might know something I don't.

I'm guessing the biggest slowdown is happening inside of `interpretAsString`, which copies the slice `va[o..#l]` into a local buffer. Most of the time, this copy is unnecessary, so there's a lot of wasted work (this function was originally intended for getting a handful of strings to display back to the user, not for bulk operations, which is why it hasn't been optimized). Unfortunately, there's no clear way to use aggregation in that function either. We might have to do something like what @ronawho did here https://github.com/Bears-R-Us/arkouda/blob/36497dff3b55e21a14edfa16ecff1ea738423756/src/SipHash.chpl#L81-L105

When `startLocale == endLocale == hereLocale`, I think you can do
```chpl
myRegex.search(createStringWithBorrowedBuffer(c_ptrTo(va[o]), l-1, l))
```

with no extra copying. For the second case (string is all on one locale, but not `here`), you could localize with `GET` as above, then use the same `createStringWithBorrowedBuffer` call.

For the final case of a string spanning locales, you would still have to use `interpretAsString`, but there are at most `numLocales-1` of these, so it shouldn't hurt performance much.

reuster986 (commented):
See inline comment for strategic suggestion.

src/SegmentedArray.chpl
ronawho: I agree that most of the overhead is from the slicing and copy inside `interpretAsString`. https://github.com/Bears-R-Us/arkouda/issues/917 has some more discussion of this. I wasn't sure if `h` was local so I wasn't sure if this PR would have an impact and to start I was just looking at the local overheads.

I think we'll end up with something similar to that localizing code in SipHash, though I'm hoping to move that to a common helper that we can use in both locations.

pierce314159: Awesome! Thanks for the suggestion! I'll try that out

reuster986: @ronawho those are some encouraging results on #917, sorry I missed that discussion. 

That leaves two questions:
- What should we do about this PR and the use of aggregators? I think `h` is always local, but I could be wrong... should we merge this PR to see what effect it has on the benchmark?
- Who wants to write the common helper function?

ronawho: I'd be inclined to make the local optimizations in #917, and see what type of impact that has on multi-locale results. If multi-locale perf still looks poor then I'd explore aggregation or other comm optimizations, but I'd probably close this until then (looks like @pierce314159 has already done so)

The code is a little tricky and low-level so I'm happy to own writing the common helper function. Things are a bit busy here with the upcoming 1.25 release, so it'll probably be next week till I have time to get to that.

reuster986: @ronawho sounds good; totally understand the need to focus on 1.25 right now.

@pierce314159 does that sound ok to you?

pierce314159: sounds good to me!

PR #926, opened 2021-09-20
Issue #917: Speedup `substring_search` benchmark startup

pierce314159:
Issue #917:
- Switches to fixed `digit_substr` for `substring_search` benchmark to allow use of `ak.stick` and speed up `test_substring` construction time

This won't test the regex functionality as thoroughly (which is okay because this should be covered by the unit tests), but speeds up the startup considerably since we can use `ak.stick` instead of pulling the random string info back from the server to construct `test_substring`


ronawho (approved):
This makes startup much, much faster on single node and 16-node-cs-hdr. 

On 16-node-cs-hdr I see like 0.25s for `--size=$((10**8))` compared to ~120s for `--size=$((10**6))` before.

glitch (approved):

PR #927, opened 2021-09-23
Closes #911: Implement `find_locations` and `findall` on regexes for `Strings`

### LINKED ISSUE #911: Implement `find_locations` and `findall` on regexes for `Strings`
In support of #894 :
- Implement `Strings.find(pattern)` which returns the number of matches, start positions, and lengths of matches (return value similar to a list of [`reMatch` records ](https://chapel-lang.org/docs/modules/standard/Regexp.html?highlight=regexp#Regexp.reMatch))
- Implement `Strings.slice` which uses the return values of `find` and returns the slices of `Strings` that match `pattern`

pierce314159:
This PR (closes #911):
- Adds `find_locations`/`findall` to `Strings` and unit tests for this functionality. Also caches previous `Strings.find_locations` results
- `find_locations(pattern)` 
  - Finds pattern matches and returns pdarrays containing the number, start postitions, and lengths of matches
  - Results of `find_locations` are cached in `regex_dict`
- `findall(pattern, return_match_origins)`
  - Return all non-overlapping matches of pattern in Strings as a new Strings object. If `return_match_origins` is True, return a pdarray containing the index of the original string each pattern match is from
  - Uses the output of `find_locations`

Example:
```python
>>> strings = ak.array(['', '____', '_1_2', '3___4___', '5'])
>>> strings.find('_+')
(array([0 1 2 2 0]), array([1 6 8 12 16]), array([4 1 1 3 3]))

>>> strings.print_cached_regex_patterns()
dict_keys(['_+'])

>>> strings.slice('_+', return_match_origins=True)
(array(['____', '_', '_', '___', '___']), array([1 2 2 3 3]))

>>> strings.slice('\\d', return_match_origins=True)
(array(['1', '2', '3', '4', '5']), array([2 2 3 3 4]))

>>> strings.print_cached_regex_patterns()
dict_keys(['_+', '\\d'])

>>> strings.register('myStr')
array(['', '____', '_1_2', '3___4___', '5'])

>>> ak.clear()
>>> strings.cached_regex_patterns()
dict_keys([])

>>> strings.slice('\\d')
array(['1', '2', '3', '4', '5'])

>>> strings.print_cached_regex_patterns()
dict_keys(['\\d'])
```

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import itertools
-from typing import cast, Tuple, List, Optional, Union
+from typing import cast, Tuple, List, Optional, Union, Dict
 from typeguard import typechecked
 from arkouda.client import generic_msg
 from arkouda.pdarrayclass import pdarray, create_pdarray, parse_single_value, \
@@ -13,7 +13,7 @@ from arkouda.dtypes import NUMBER_FORMAT_STRINGS, resolve_scalar_dtype, \
      translate_np_dtype
 import json
 import re
-from arkouda.infoclass import information
+from arkouda.infoclass import information, list_symbol_table
 
 __all__ = ['Strings']
 
@@ -39,6 +39,10 @@ class Strings:
         The sizes of each dimension of the array
     dtype : dtype
         The dtype is ak.str
+    regex_dict: Dict[str, Tuple[pdarray, pdarray, pdarray]]
+        Dictionary storing information on matches (cache of Strings.find(pattern))
+        Keys - regex patterns
+        Values - tuples of pdarrays (numMatches, matchStarts, matchLens)
     logger : ArkoudaLogger
         Used for all logging operations
         
@@ -102,6 +106,7 @@ class Strings:
 
         self.dtype = npstr
         self.name:Optional[str] = None
+        self.regex_dict: Dict = dict()
         self.logger = getArkoudaLogger(name=__class__.__name__) # type: ignore
 
     def __iter__(self):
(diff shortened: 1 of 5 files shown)


reuster986 (commented):
Great work! I like the API here. Just a couple of questions:

1) I wonder if we should change the names of the `Strings.slice` method to `Strings.search`. I'm envisioning a future `slice` method that pulls out substrings based on arrays of start and end indices. To me, a function that takes a regex and returns substrings feels more like a `search` than a `slice`. What do you think?

2) For both @pierce314159 and @ronawho what is the relationship of this PR to #928 and #931? Does the code in this PR need to change based on those other PRs, or is it abstracted enough to merge now?

pierce314159:
> 
> 
> Great work! I like the API here. Just a couple of questions:
> 
>     1. I wonder if we should change the names of the `Strings.slice` method to `Strings.search`. I'm envisioning a future `slice` method that pulls out substrings based on arrays of start and end indices. To me, a function that takes a regex and returns substrings feels more like a `search` than a `slice`. What do you think?
> 
>     2. For both @pierce314159 and @ronawho what is the relationship of this PR to #928 and #931? Does the code in this PR need to change based on those other PRs, or is it abstracted enough to merge now?

Thanks! 
1. I agree `Strings.search` feels like a better fit and leaves the `slice` name available
2. I believe the only thing that should need to change is in `findMatches`
```C
var matches = myRegex.matches(interpretAsString(origVals[off..#len])); 
->
var matches = myRegex.matches(interpretAsBytes(origVals, off..#len, borrow=true));
```

arkouda/strings.py
```diff
@@ -255,6 +260,146 @@ def get_lengths(self) -> pdarray:
...
+        """
+        sym_tab = list_symbol_table()
+        self.regex_dict = {key: val for key, val in self.regex_dict.items() if all([pda.name in sym_tab for pda in val])}
+
+    def print_cached_regex_patterns(self):
+        """
+        Prints the regex patterns for which Strings.find(pattern) has been cached
+        """
+        self._refresh_regex_dict()
+        print(self.regex_dict.keys())
```
glitch, line 275: It looks like `_refresh_regex_dict` is only called by `print_cached_regex_patterns`, I'm wondering if a better approach would be to return a copy of the dictionary you made in `_refresh_regex_dict` if you're worried about a user modifying the state of it.  Then the user can choose to print it or not... having a function that prints feels weird to me.

pierce314159: ooo yeah i agree! that makes more sense

ronawho:
> For both @pierce314159 and @ronawho what is the relationship of this PR to #928 and #931? Does the code in this PR need to change based on those other PRs, or is it abstracted enough to merge now?

`interpretAsString` will need to be updated to pass `borrow=true` to borrow the segmented array data instead of allocating a new string. I can roll that into #931

arkouda/strings.py
```diff
@@ -255,6 +260,146 @@ def get_lengths(self) -> pdarray:
...
+                                        self.offsets.name,
+                                        self.bytes.name,
+                                        json.dumps([pattern]))
+            repMsg = cast(str, generic_msg(cmd=cmd, args=args))
+            arrays = repMsg.split('+', maxsplit=2)
+            self.regex_dict[pattern] = (create_pdarray(arrays[0]), create_pdarray(arrays[1]), create_pdarray(arrays[2]))
+        return self.regex_dict[pattern]
+
+    @typechecked
+    def slice(self, pattern: Union[bytes, str_scalars], return_match_origins: bool = False) -> Union[Strings, Tuple]:
```
glitch, line 343: I agree with @reuster986 on the naming here.  I'd try to make the naming match the `re2` naming of what's happening here if possible.  `slice` seems to imply a different context..something you apply to the structure of an array/list.  Looking at the ongoing conversation it sounds like you are already going to change it.

pierce314159:
@reuster986 and @glitch, looking at the python `re` api (which seems like what we should try to mimic to me, let me know if anyone thinks otherwise), it seems `slice` here is essentially [`findall`](https://docs.python.org/3/library/re.html#re.findall). So I think I'll rename `slice` to that. I think I'll also change `find` to `find_locations`

Let me know if there's any issue with this!

reuster986:
@pierce314159 `findall` and `find_locations` sound good to me.

glitch (approved):

reuster986 (approved):

PR #928, opened 2021-09-25
Move localizing code from SipHash to a general helper

ronawho:
In #287 SipHash was optimized by localizing segmented string data with a
bulk GET instead of collecting elements one at a time. This moves that
localization logic into a common `lowLevelLocalizingSlice` helper so
that we can use it in an upcoming PR to localize segmented string data
used in the regex implementation.

This also adds a seed argument to SipHashSpeedTest to get the same
arrays from run-to-run. I used this to verify there is no performance
difference from this change.

Part of #917

### DIFF
--- a/src/AryUtil.chpl
+++ b/src/AryUtil.chpl
@@ -1,5 +1,6 @@
 module AryUtil
 {
+    use CPtr;
     use Random;
     use Reflection;
     use Logging;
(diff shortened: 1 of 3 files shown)


reuster986 (approved):
cool!

reuster986:
I'll wait for @pierce314159 to review before merging, since he will be using this code directly.

pierce314159 (approved):
Looks good to me! Thanks @ronawho!

Issue #930, opened 2021-09-27
`flatten` benchmark is timing out in nightly performance testing


ronawho:
Similar to https://github.com/Bears-R-Us/arkouda/issues/917, the `flatten` benchmark is timing out in nightly perf testing. There are similar issues to https://github.com/Bears-R-Us/arkouda/issues/917, but there are some overheads beyond that too. 

The array creation is somewhat slow because casting to strings is slow -- https://github.com/Bears-R-Us/arkouda/issues/675. I'm looking into optimizing the str cast, but am curious if there's an alternate way to create the input. 

The significantly slower part is the `answer` creation with `answer = ak.array(['{}'.format(i) for i in range(N*3)])`. This will create a large array on the client (which is not practical when the problem size and number of server locales grows) and then transfers it via ZMQ over to the server, which even with recent optimizations will ultimately be limited by the TCP bandwidth between the client and server.

pierce314159:
yeah this makes a lot of sense. I think I'm going to change the answer creation to
```python
answer = ak.cast(ak.arange(N*3), 'str')
```
to keep everything server side and take advantage of the str cast speed up. Any alternate way to create the input will hopefully also apply to the answer creation

ronawho:
Sounds good. Most of the benchmarks don't do validation so if this still ends up being slow, I think we could skip the answer-checking. (That said, I like doing the validation, so hopefully it's fast enough)

PR #931, opened 2021-09-27
Optimize how segmented array slices are interpreted as strings/bytes

ronawho:
Optimize how slices of segmented arrays are interpreted as a Chapel
string/bytes. Interpreting is required to use Chapel's regex support,
which only operates on string/bytes types. Previously, the regex code
was localizing a region with a slice-assignment, and then making a new
string from it. This resulted in a large amount of short-lived
concurrent allocations, which was slow.

Update the regex code to use the new `lowLevelLocalizingSlice` routine
and instead of creating a new string, borrow memory from the segmented
array (or take ownership of the localized slice if the original data was
remote). This avoids allocation when the slice is local and minimizes
the amount of allocation and communication when the slice is remote.

This should significantly improve the performance of regex operations.

Part of #917
Part of #930
Part of #940

### DIFF
--- a/src/CastMsg.chpl
+++ b/src/CastMsg.chpl
@@ -203,7 +203,7 @@ module CastMsg {
               } else {
                    end = oa[i+1] - 1;
               }
-              e = interpretAsString(va[start..end]) : toType;
+              e = interpretAsString(va, start..end) : toType;
           }
       } catch e: IllegalArgumentError {
           var errorMsg = "bad value in cast from string to %s".format(toType:string);
--- a/src/Flatten.chpl
+++ b/src/Flatten.chpl
@@ -1,6 +1,7 @@
 module Flatten {
   use ServerConfig;
 
+  use AryUtil;
   use SegmentedArray;
   use ServerErrors;
   use SymArrayDmap;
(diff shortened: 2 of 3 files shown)


ronawho:
FYI @reuster986 and @pierce314159. I would like to get the concurrent allocation optimizations in first before merging this, so I've marked it as a draft for now but I don't expect the code to change.

There were some preliminary single node performance results in https://github.com/Bears-R-Us/arkouda/issues/917#issuecomment-919567770, but I'll gather some multi-locale comparisons after the alloc optimizations are in.

reuster986 (approved):
looks good! will wait to merge until dependency is merged.

pierce314159 (approved):

ronawho:
This is rebased and ready for review. I did not end up gathering concrete performance numbers since the exact benefit differs between chapel 1.25 and main as well as single node vs multi-node. I'm planning to just let nightly do the work of collecting numbers and reporting back. That said, I expect significant performance benefits from this for all configs.

glitch (approved):

pierce314159 (approved):

PR #934, opened 2021-09-29
Closes #914 by handling case where the num_elements < num_locales

### LINKED ISSUE #914: Categorical/Strings array with size smaller than num locales fails read/write to HDF5

glitch:
Closes #914 by handling case where the num_elements < num_locales
when saving to HDF5. When data is not present for a locale an
empty dataset is written to the file. These empty datasets get ignored
when reading back from disk.

Added unit tests for small numeric & strings arrays where we
expect the number of locales to be larger then the number of elements.

### Notes
1. Most of the changes for big blocks of code are white-space related because they were wrapped in if/else conditionals to handle boundary conditions.  It may be easier to view these sections in an ide or visual diff tool that can ignore whitespace changes (I like pycharm for this personally)
2. The fundamental change here is adding empty datasets to files on write when data does not exist on a locale and ignoring those on read.  This allows dataset / groups to be present in all files so no errors get thrown while reading back and the `alllow_errors` and `strictTypes` flags do not need to be toggled by the user (potentially masking other errors).

### DIFF
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -13,6 +13,7 @@ module GenSymIO {
     use NumPyDType;
     use List;
     use Map;
+    use Set;
     use PrivateDist;
     use Reflection;
     use ServerErrors;
@@ -467,17 +468,17 @@ module GenSymIO {
                                            "Verified all dtypes across files for dataset %s".format(dsetName));
             var subdoms: [filedom] domain(1);
             var segSubdoms: [filedom] domain(1);
+            var skips = new set(string);
             var len: int;
             var nSeg: int;
             try {
                 if isSegArray {
-                    // TODO wrap this in a conditional with option to ignore
                     if (!calcStringOffsets) {
-                        (segSubdoms, nSeg) = get_subdoms(filenames, dsetName + "/" + SEGARRAY_OFFSET_NAME);
+                        (segSubdoms, nSeg, skips) = get_subdoms(filenames, dsetName + "/" + SEGARRAY_OFFSET_NAME);
                     }
-                    (subdoms, len) = get_subdoms(filenames, dsetName + "/" + SEGARRAY_VALUE_NAME);
+                    (subdoms, len, skips) = get_subdoms(filenames, dsetName + "/" + SEGARRAY_VALUE_NAME);
                 } else {
-                    (subdoms, len) = get_subdoms(filenames, dsetName);
+                    (subdoms, len, skips) = get_subdoms(filenames, dsetName);
                 }
             } catch e: HDF5RankError {
                 var errorMsg = notImplementedError("readhdf", "Rank %i arrays".format(e.rank));
@@ -503,7 +504,7 @@ module GenSymIO {
 
                     // Load the strings bytes/values first
                     var entryVal = new shared SymEntry(len, uint(8));
-                    read_files_into_distributed_array(entryVal.a, subdoms, filenames, dsetName + "/" + SEGARRAY_VALUE_NAME);
+                    read_files_into_distributed_array(entryVal.a, subdoms, filenames, dsetName + "/" + SEGARRAY_VALUE_NAME, skips);
                     var valName = st.nextName();
                     st.addEntry(valName, entryVal);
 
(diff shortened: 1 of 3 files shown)


ronawho:
> may be easier to view these sections in an ide or visual diff tool that can ignore whitespace

FYI github supports this. In https://github.com/Bears-R-Us/arkouda/pull/934/files under the settings gear icon, or you can just do url hacking and add `?w=1`. e.g.  https://github.com/Bears-R-Us/arkouda/pull/934/files?w=1

src/GenSymIO.chpl
```diff
@@ -650,10 +651,17 @@ module GenSymIO {
...
-            b = sd.low; // Boundary is index of first segment in file
-            // Height increase of next region is number of bytes in current region
-            if (i < fD.high) {
-                diffs[i+1] = vd.size;
+            // if we encounter a malformed subdomain i.e. {1..0} that means we encountered a file
+            // that has no data for this SegString object, we can safely skip processing this file.
+            if sd.low > sd.high {
+                gsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                    "fD:%t segments subdom:%t is malformed signaling no segment data in file, skipping".format(i, sd));
+            } else {
```
hokiegeek2, line 659: more chpldoc comments would be good in the else code block

src/GenSymIO.chpl
```diff
@@ -930,6 +939,12 @@ module GenSymIO {
                 C_HDF5.HDF5_WAR.H5LTget_dataset_info_WAR(file_id, dName.c_str(), 
                                            c_ptrTo(dims), nil, nil);
                 lengths[i] = dims[0]: int;
+                if lengths[i] == 0 {
+                    skips.add(filename);
+                    try! gsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
```
hokiegeek2, line 944: recommend getting rid of try! as it can crash server

hokiegeek2 (approved):

reuster986 (changes requested):
General logic looks good, just a couple minor requests.

reuster986 (approved):

src/GenSymIO.chpl
```diff
@@ -646,14 +647,33 @@ module GenSymIO {
...
 
+    /**
+     * inline proc to validate the range for our domain.
+     * Valid domains must be increasing with the lower bound <= upper bound
+     * :arg r: 1D domain
+     * :type domain(1): one dimensional domain
+     *
+     * :returns: bool True iff the lower bound is less than or equal to upper bound
+     */
+    inline proc _isValidRange(r: domain(1)): bool {
```
ronawho, line 658: Note that this is the same as `!r.isEmpty()` if you wanted to just use a builtin. In terms of terminology we don't tend to think of this as valid vs malformed, but just empty vs non-empty.

PR #936, opened 2021-10-01
Fix get_mem_used test for multi-locale testing

ronawho:
#935 added a mem tracking threshold, which required updating the
get_mem_used test to allocate something over that threshold. I just
allocated a fixed amount, but forgot to scale it by the number of
locales so this was failing for 9+ locales when the size-per-locale
dropped under the threshold. Add the scaling by numLocales here.

### DIFF
--- a/tests/client_test.py
+++ b/tests/client_test.py
@@ -89,7 +89,8 @@ class ClientTest(ArkoudaTest):
                 expected or the call to ak.client.get_mem_used() fails 
         '''  
         try:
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

ronawho:
(Probably a trivial commit to request 3 reviewers on, but all of you reviewed the PR that introduced the bug so I figured I ping you here too)

reuster986 (approved):

glitch (approved):

PR #941, opened 2021-10-04
Fix sorting arrays with values on bitsPerDigit boundaries

### LINKED ISSUE #937: Sort fails to sort under bizarre condition
The `assert` below fails:

```python
L = -2**15
U = 2**16
a = ak.randint(L, U, 100)
assert ak.is_sorted(ak.sort(a))
```

Varying `L` and `U` leads to the following inferred conditions for triggering the bug:
 - `a.dtype == ak.int64` (The `assert` succeeds for `float64`)
 - `a` must contain both positive and negative values
 - `2**(n-1) <= a.max() < 2**n` where `n` is a multiple of 16. That is, the leading non-zero bit of the positive subset of `a` must be the 16th, 32nd, or 48th bit.
 - `-2**(n-1) <= a.min() < 0`. That is, the nth bit in the negative subset of `a` must not always be zero (in two's complement). 

I'm guessing that there is a bug in how negative integer values are handled in the LSD radix sort when the highest bit in the final digit is set for both positive and negative elements.

ronawho:
Fix sorting arrays that contain a positive value that requires a
multiple of `bitsPerDigit` digits and at least one negative value. As an
optimization, the sort code checks how many bits are used so that we can
reduce the number of rounds of radix sort. e.g. if an array only
contains unsigned 16-bit values (`0 - 2**16-1`) we only need 1 round of
radixsort. Similarly, if an array only contains signed 16-bit values
(`-2**15 - 2**15-1`) we still only need 1 round of siphash.

However, there was a bug in the code where we were independently
computing the number of bits needed for positive and negative values.
If you have an array `0 - 2**16-1` you only need 16 bits, but as soon as
we add a negative value the signbit needs to be accounted for, which we
weren't doing before. So for an array with `-2**15 - 2**16-1` values we
were saying only 16 bits were needed even though it actually needs 17.
Fix that here by computing the number of bits needed for positive and
negative values and then if there are any negative values add 1 for the
signbit.

Resolves #937

### DIFF
--- a/tests/sort_test.py
+++ b/tests/sort_test.py
@@ -12,7 +12,33 @@ class SortTest(ArkoudaTest):
         spda = ak.sort(pda)
         maxIndex = spda.argmax()
         self.assertTrue(maxIndex > 0)
-   
+
+    def testBitBoundaryHardcode(self):
+
+        # test hardcoded 16/17-bit boundaries with and without negative values
+        a = ak.array([1, -1, 32767]) # 16 bit
+        b = ak.array([1,  0, 32768]) # 16 bit
+        c = ak.array([1, -1, 32768]) # 17 bit
+        assert ak.is_sorted(ak.sort(a))
+        assert ak.is_sorted(ak.sort(b))
+        assert ak.is_sorted(ak.sort(c))
+
+        # test hardcoded 64-bit boundaries with and without negative values
+        d = ak.array([1, -1, 2**63-1])
+        e = ak.array([1,  0, 2**63-1])
+        f = ak.array([1, -2**63, 2**63-1])
+        assert ak.is_sorted(ak.sort(d))
+        assert ak.is_sorted(ak.sort(e))
+        assert ak.is_sorted(ak.sort(f))
+
+    def testBitBoundary(self):
+
+        # test 17-bit sort
+        L = -2**15
+        U = 2**16
+        a = ak.randint(L, U, 100)
+        assert ak.is_sorted(ak.sort(a))
+
     def testErrorHandling(self):
         
         # Test RuntimeError from bool NotImplementedError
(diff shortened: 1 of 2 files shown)


tests/sort_test.py
```diff
@@ -12,7 +12,26 @@ def testSort(self):
         spda = ak.sort(pda)
         maxIndex = spda.argmax()
         self.assertTrue(maxIndex > 0)
-   
+
+    def testBitBoundaryHardcode(self):
```
mppf, line 16: It'd be good to add a test along these lines that includes `max(int)` and `min(int)` values

mppf (approved):

tests/sort_test.py
mppf: (if there isn't already one -- I didn't look)

ronawho: added

tests/sort_test.py
```diff
@@ -12,7 +12,33 @@ def testSort(self):
...
+        a = ak.array([1, -1, 32767]) # 16 bit
+        b = ak.array([1,  0, 32768]) # 16 bit
+        c = ak.array([1, -1, 32768]) # 17 bit
+        assert ak.is_sorted(ak.sort(a))
+        assert ak.is_sorted(ak.sort(b))
+        assert ak.is_sorted(ak.sort(c))
+
+        # test hardcoded 64-bit boundaries with and without negative values
+        d = ak.array([1, -1, 2**63-1])
+        e = ak.array([1,  0, 2**63])
```
reuster986, line 28: Did you mean `2**63-1` here as well? I don't think `2**63` will transfer over correctly, since it is out of bounds for `int64`.

ronawho: Ah, I did mean `2**63` to get something that's an unsigned 64-bit value, but I was just thinking about this from the perspective of a bag-of-bits, forgetting that the actual types are signed 64-bits so this would overflow. I'll change it.

ronawho: Changed to `2**63-1`

reuster986 (approved):

PR #944, opened 2021-10-05
Issue #914 : Disabling testSmallStringArrayToHDF5 from unit tests.

glitch:
This is causing issues with nightly benchmarking and represents a
very rare corner case on a tiny amout of data. We already know this
section of code requires and overhaul so we feel comfortable in
disabling it for now.

### DIFF
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -2,7 +2,7 @@ import os, shutil, glob
 import tempfile
 
 import numpy as np
-from pytest import warns
+import pytest
 from typing import List, Mapping, Union
 from base_test import ArkoudaTest
 from context import arkouda as ak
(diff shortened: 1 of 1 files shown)


tests/io_test.py
```diff
@@ -579,13 +579,15 @@ def testSmallArrayToHDF5(self):
...
-    def testSmallStringArrayToHDF5(self):
-        a1 = ak.array(["ab", "cd"])
-        with tempfile.TemporaryDirectory(dir=IOTest.io_test_dir) as tmp_dirname:
-            a1.save(f"{tmp_dirname}/small_numeric", dataset="a1")
-            # Now load it back in
-            a2 = ak.load(f"{tmp_dirname}/small_numeric", dataset="a1")
-            self.assertEqual(str(a1), str(a2))
+    # This currently breaks on 4 or greater locales.  Since it is such a rare corner case
+    # we are going to remove it for now.
+    # def testSmallStringArrayToHDF5(self):
```
ronawho, line 584: At least historically, the way tests were skipped at other times was with `@pytest.mark.skip` (requires `import pytest`) instead of commenting out.

ronawho: (Not sure if there's an advantage one way or another, I'll defer to @hokiegeek2 for that)

glitch: Cool, I wasn't sure of the interplay between pytest marks and python's `unittest` module.  I'll test it out and update the PR.

ronawho (approved):

pierce314159 (approved):

PR #945, opened 2021-10-05
Add BitOps #943

### LINKED ISSUE #943: Add methods for bitops, e.g. `popcount`
At least three users (including me) have rolled our own implementation of popcount, so it's probably time to make it a true arkouda function. While we're there, might as well add the other [bitops](https://chapel-lang.org/docs/modules/standard/BitOps.html): parity, clz, ctz, rotl, rotr.

My plan is to slip these into `EfuncMsg.chpl`, because they fit the same signature and would reuse a lot of the boilerplate for message processing. On the client side, these will be methods of `pdarray`, and should raise `TypeError` if not `int64`.

reuster986:
This PR adds support for the operations in the Chapel `BitOps` package:
- `popcount`
- `parity`
- `clz`
- `ctz`
- `rotl`
- `rotr`

These operations are accessible as both module-level functions (e.g. `ak.popcount(x)`) and methods (e.g. `x.popcount()`). The rotations accept both array and scalar arguments, like other binary operators.

Closes #943

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -15,7 +15,8 @@ from arkouda.infoclass import list_registry, information, pretty_print_informati
 import builtins
 
 __all__ = ["pdarray", "clear", "any", "all", "is_sorted", "sum", "prod", "min", "max", "argmin",
-           "argmax", "mean", "var", "std", "mink", "maxk", "argmink", "argmaxk", "attach_pdarray",
+           "argmax", "mean", "var", "std", "mink", "maxk", "argmink", "argmaxk", "popcount",
+           "parity", "clz", "ctz", "attach_pdarray",
            "unregister_pdarray_by_name", "RegistrationError"]
 
 logger = getArkoudaLogger(name='pdarrayclass')    
@@ -802,6 +803,17 @@ class pdarray:
         """
         return argmaxk(self,k)
 
+    def popcount(self) -> pdarray:
+        return popcount(self)
+
+    def parity(self) -> pdarray:
+        return parity(self)
+
+    def clz(self) -> pdarray:
+        return clz(self)
+
+    def ctz(self) -> pdarray:
+        return ctz(self)
     
     def to_ndarray(self) -> np.ndarray:
         """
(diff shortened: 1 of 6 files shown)


src/EfuncMsg.chpl
```diff
@@ -101,6 +102,22 @@ module EfuncMsg
                         // code append second array's attrib
                         repMsg += "created " + st.attrib(rname2) + "+";
                     }
+                    when "popcount" {
+                        var a = st.addEntry(rname, e.size, int);
+                        a.a = popcount(e.a);
```
reuster986, line 107: @ronawho I tested this locally, but I wanted to double check: when promoting a scalar function like this over a block distributed array, will the compiler automatically use data parallelism?

ronawho: Correct, this compiler will transform this into something like:

```chpl
forall (_aa, _ea) in zip(a.a, e.a) {
  _aa = popcount(_ea);
}
```

https://chapel-lang.org/docs/users-guide/datapar/promotion.html and https://chapel-lang.org/docs/language/spec/data-parallelism.html#promotion have more descriptions of this behavior.

ronawho (approved):
Looks good to me.

At a high level I'll note that `--specialize` (implied by `--fast` when you have `CHPL_TARGET_CPU` is defined to something other than `none`/`unknown`) was originally added to make code generation for our bitops more optimized towards the target CPU. This just falls out on Cray XC, but on InfiniBand and other systems where you might be setting `CHPL_TARGET_CPU=none` you may not be achieving full performance with bitops.

I don't actually think this will be a matter for you because operations will still be really fast and scalable, they just might not be as fast as the CPU is capable of.

arkouda/pdarrayclass.py
```diff
@@ -1772,6 +1809,214 @@ def argmaxk(pda : pdarray, k : int_scalars) -> pdarray:
...
+            return x._binop(rot, "<<<")
+        else:
+            raise TypeError("Rotations only supported on integers")
+    elif isSupportedInt(x):
+        if isinstance(rot, pdarray) and rot.dtype == akint64:
+            return rot._r_binop(x, "<<<")
+        else:
+            raise TypeError("Rotations only supported on integers")
+    else:
+        raise TypeError("Rotations only supported on integers")
```
pierce314159, line 1976: This is just stylistic but I think restructuring the if/else statements here makes it more readable. Something similar can be done to `rotr`. No need to change it
```python
    if isinstance(x, pdarray) and x.dtype == akint64:
        if (isinstance(rot, pdarray) and rot.dtype == akint64) or isSupportedInt(rot):
            return x._binop(rot, "<<<")
        else:
            raise TypeError("Rotations only supported on integers")
    elif isSupportedInt(x) and isinstance(rot, pdarray) and rot.dtype == akint64:
        return rot._r_binop(x, "<<<")
    else:
        raise TypeError("Rotations only supported on integers")
```

src/EfuncMsg.chpl
reuster986: Cool!

pierce314159 (approved):

arkouda/pdarrayclass.py
reuster986: That's a good suggestion -- implemented!

reuster986:
@ronawho good to know. This question is mostly for my general knowledge, but is it even possible for us to specialize on our infiniband system? I don't see an option for "rome" on [this page](https://chapel-lang.org/docs/usingchapel/chplenv.html?highlight=chpl_target_cpu#chpl-target-cpu)...

ronawho:
> @ronawho good to know. This question is mostly for my general knowledge, but is it even possible for us to specialize on our infiniband system? I don't see an option for "rome" on this page...

We're having some internal discussions about this -- both about our lack of maintenance on that page and the underlying mapping, as well how this is changing with switching to llvm-by-default. I don't know that you can specify anything for Rome there today, but what you can do is `CHPL_TARGET_CPU=native`, which requires your login node to be Rome (or just in general the same as your compute) or you can build on the compute node. 

Note that if you have a system with say intel and amd chips, this could lead to a binary that only works on one set of cpus and not the other. It may be worth experimenting to see if target specialization has much of an impact for you for these operations, but for production use I might stick with `CHPL_TARGET_CPU=none` until we figure out our story here.

PR #956, opened 2021-10-25
Closes #452 by simulating h5ls using the HDF5 api H5Literate

### LINKED ISSUE #452: Avoid requiring h5ls
`lshdfMsg` shells out to `h5ls` in order to get the objects in the root group of an hdf5 file. This has created portability issues in our nightly testing where we've had to add `h5ls` to our `PATH` and even then we have some situations where `PATH` gets overridden (such as when running on a cray-xc elogin node.)

We can work around this in our testing, but it seems like this could be a pain point for users who are using `make install-deps` or `ARKOUDA_HDF5_PATH` when they don't have permissions to install a system version of hdf5 themselves. (this is true for most chapel devs on our corporate machines)

I'm proposing we just get rid of the h5ls dependency and use the C API to get the same information. I'm happy to take a stab at that, though I'm not very familiar with hdf5 yet.

An alternative (but IMO hacky) workaround would be to update which h5ls is called with something like the following:

<details> 

```diff
diff --git a/Makefile b/Makefile
index a039a16..9de311a 100644
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,9 @@ $(eval $(call add-path,$(ARKOUDA_ZMQ_PATH)))
 endif
 ifdef ARKOUDA_HDF5_PATH
 $(eval $(call add-path,$(ARKOUDA_HDF5_PATH)))
+ifneq (,$(wildcard $(ARKOUDA_HDF5_PATH)/bin/h5ls))
+H5LS_FLAGS += -sGenSymIO.h5ls="\"$(ARKOUDA_HDF5_PATH)/bin/h5ls\""
+endif
 endif

 .PHONY: install-deps
@@ -67,6 +70,10 @@ HDF5_NAME_VER := hdf5-$(HDF5_VER)
 HDF5_BUILD_DIR := $(DEP_BUILD_DIR)/$(HDF5_NAME_VER)
 HDF5_INSTALL_DIR := $(DEP_INSTALL_DIR)/hdf5-install
 HDF5_LINK :=  https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-$(HDF5_MAJ_MIN_VER)/$(HDF5_NAME_VER)/src/$(HDF5_NAME_VER).tar.gz
+ifneq (,$(wildcard $(HDF5_INSTALL_DIR)/bin/h5ls))
+H5LS_FLAGS += -sGenSymIO.h5ls="\"$(HDF5_INSTALL_DIR)/bin/h5ls\""
+endif
+
 install-hdf5:
        @echo "Installing HDF5"
        rm -rf $(HDF5_BUILD_DIR) $(HDF5_INSTALL_DIR)
@@ -173,7 +180,7 @@ ARKOUDA_SOURCES = $(shell find $(ARKOUDA_SOURCE_DIR)/ -type f -name '*.chpl')
 ARKOUDA_MAIN_SOURCE := $(ARKOUDA_SOURCE_DIR)/$(ARKOUDA_MAIN_MODULE).chpl

 $(ARKOUDA_MAIN_MODULE): check-deps $(ARKOUDA_SOURCES) $(ARKOUDA_MAKEFILES)
-       $(CHPL) $(CHPL_DEBUG_FLAGS) $(PRINT_PASSES_FLAGS) $(CHPL_FLAGS_WITH_VERSION) $(ARKOUDA_MAIN_SOURCE) -o $@
+       $(CHPL) $(CHPL_DEBUG_FLAGS) $(PRINT_PASSES_FLAGS) $(CHPL_FLAGS_WITH_VERSION) $(H5LS_FLAGS) $(ARKOUDA_MAIN_SOURCE) -o $@

 CLEAN_TARGETS += arkouda-clean
 .PHONY: arkouda-clean
diff --git a/src/GenSymIO.chpl b/src/GenSymIO.chpl
index 4e47078..2678127 100644
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -9,6 +9,7 @@ module GenSymIO {
   use Sort;
   use CommAggregation;

+  private config const h5ls = "h5ls";
   config const GenSymIO_DEBUG = false;
   config const SEGARRAY_OFFSET_NAME = "segments";
   config const SEGARRAY_VALUE_NAME = "values";
@@ -150,7 +151,7 @@ module GenSymIO {
       if exists(tmpfile) {
         remove(tmpfile);
       }
-      var cmd = try! "h5ls \"%s\" > \"%s\" 2>&1".format(filename, tmpfile);
+      var cmd = try! "%s \"%s\" > \"%s\" 2>&1".format(h5ls, filename, tmpfile);
       var sub = spawnshell(cmd);
       // sub.stdout.readstring(repMsg);
       sub.wait();
```

</details>

glitch:
Closes #452 by simulating h5ls using the HDF5 api H5Literate
to list datasets and groups in the top level of an HDF5 file.
This uses helper functions written in C to handle the opaque
c_void_ptr / void* used to pass data between HDF5 iterations.
This generates and returns a json formatted  list of the top
level dataset and group objects.

### DIFF
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -25,6 +25,8 @@ module GenSymIO {
     use Search;
     use IndexingMsg;
 
+    require "c_helpers/help_h5ls.h", "c_helpers/help_h5ls.c";
+
     private config const logLevel = ServerConfig.logLevel;
     const gsLogger = new Logger(logLevel);
 
@@ -210,14 +212,12 @@ module GenSymIO {
     }
 
     /*
-     * Spawns a separate Chapel process that executes and returns the 
-     * result of the h5ls command
+     * Simulates the output of h5ls for top level datasets or groups
+     * :returns: string formatted as json list
+     * i.e. ["_arkouda_metadata", "pda1", "s1"]
      */
-    proc lshdfMsg(cmd: string, payload: string,
-                                st: borrowed SymTab): MsgTuple throws {
+    proc lshdfMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
         // reqMsg: "lshdf [<json_filename>]"
-        use Spawn;
-        const tmpfile = "/tmp/arkouda.lshdf.output";
         var repMsg: string;
         var (jsonfile) = payload.splitMsgToTuple(1);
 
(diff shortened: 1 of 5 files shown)


src/GenSymIO.chpl
```diff
@@ -258,48 +258,105 @@ module GenSymIO {
...
+        var idx_p:C_HDF5.hsize_t; // This is the H5Literate index counter
+        
+        // First iteration to get the item count so we can ballpark the char* allocation
+        var nfields:c_int = 0:c_int;
+        C_HDF5.H5Literate(fid, C_HDF5.H5_INDEX_NAME, C_HDF5.H5_INDEX_NAME, idx_p, c_ptrTo(_get_item_count), c_ptrTo(nfields));
+        
+        // Allocate space for array of strings
+        var c_field_names = c_calloc(c_char, 255 * nfields);
+        idx_p = 0:C_HDF5.hsize_t; // reset our iteration counter
+        C_HDF5.H5Literate(fid, C_HDF5.H5_INDEX_NAME, C_HDF5.H5_INDEX_NAME, idx_p, c_ptrTo(_simulate_h5ls), c_field_names);
```
bradcray, line 355: @reuster986 : I know that @glitch isn't actively developing anymore, but I noticed today that the third argument to each of the H5Literate() calls above isn't of the correct type (which the C compiler will complain about in certain modes).  Specifically, it's expecting an argument of type `H5_iter_order_t`, but we're sending it an argument of `H5_index_t`.  Chapel doesn't catch this because of the white lies that the HDF5 module uses to describe external enums, and C's permissiveness typically lets it slide as well if not compiling with various warnings turned on.  I think the argument should be switched to one of the following:

```
    // H5_iter_order_t enum                                                     
    extern type H5_iter_order_t = c_int;
    extern const H5_ITER_UNKNOWN :H5_iter_order_t;
    extern const H5_ITER_INC :H5_iter_order_t;
    extern const H5_ITER_DEC :H5_iter_order_t;
    extern const H5_ITER_NATIVE :H5_iter_order_t;
    extern const H5_ITER_N :H5_iter_order_t;
```

bradcray: I ended up opening #1189 about this in case it got lost once I closed this tab.

PR #965, opened 2021-11-03
Closes #963: Performance drop in substring search methods

### LINKED ISSUE #963: performance drop in substring search methods
https://chapel-lang.org/perf/arkouda/16-node-xc/?startdate=2021/10/20&enddate=2021/11/03&configs=release,nightly&graphs=substringsearchperformance

The performance of the substring search benchmark dropped significantly after switching to use `Matcher(pattern).get_match(MatchType).matched()` to calculate `contains`, `startswith`, and `endswith` in PR #958 

The benefit of routing through `Matcher` is this information is cached so subsquent calls to these methods (or any `Match` methods) require no calculation, but this requires a lot of overhead to calculate the locations for every match type when the substring search methods are called before a `Matcher` object has been created

Strategy:
- check if the `Matcher` has already been chached and if so, return the appropriate `matched` pdarray
- if not fall back to more performant `contains`, `startswith`, `endswith`

This way the substring search methods take advantage of the `Matcher` object if it's been created but don't take the performance hit if someone just wants the boolean array

pierce314159:
Closes #963:
- Use cached `Matcher` if available and default to more performant substring search methods if not
- Still benefits from caching but saves overhead of calculating all the `Matcher` info if only a boolean array for matches is required

I did some basic timing on my laptop to get some idea of how the caching is coming into play
<details>
<summary> Timing Examples:</summary>

```python
# before PR contains
>>> import timeit
>>> strings = ak.array([f'{i%10} string {i%10}' for i in range(10**7)])
>>> pattern = 'ing \\d'
>>> timeit.timeit('strings.contains(pattern, regex=True)', globals=globals(), number=1)
4.9482425000005605

>>> strings.search(pattern)
<ak.Match object: matched=True, span=(5, 10); matched=True, span=(5, 10); matched=True, span=(5, 10); ... ; matched=True, span=(5, 10); matched=True, span=(5, 10); matched=True, span=(5, 10)>

>>> timeit.timeit('strings.contains(pattern, regex=True)', globals=globals(), number=1)
0.324951599999622

>>> strings.purge_cached_regex_patterns()
>>> timeit.timeit('strings.contains(pattern, regex=True)', globals=globals(), number=1)
4.708719000000201
```

```python
# after PR contains
>>> import timeit
>>> strings = ak.array([f'{i%10} string {i%10}' for i in range(10**7)])
>>> pattern = 'ing \\d'
>>> timeit.timeit('strings.contains(pattern, regex=True)', globals=globals(), number=1)
1.7519627000001492

>>> strings.search(pattern)
<ak.Match object: matched=True, span=(5, 10); matched=True, span=(5, 10); matched=True, span=(5, 10); ... ; matched=True, span=(5, 10); matched=True, span=(5, 10); matched=True, span=(5, 10)>

>>> timeit.timeit('strings.contains(pattern, regex=True)', globals=globals(), number=1)
0.3537317999980587

>>> strings.purge_cached_regex_patterns()
>>> timeit.timeit('strings.contains(pattern, regex=True)', globals=globals(), number=1)
1.622022299998207
```

```python
# after PR startwith/endswith is similar except endswith only benefits if pattern+'$' is cached
>>> import timeit
>>> strings = ak.array([f'{i%10} string {i%10}' for i in range(10**7)])
>>> pattern = '\\d str'
>>> timeit.timeit('strings.startswith(pattern, regex=True)', globals=globals(), number=1)
1.6063028999997186

>>> strings.search(pattern)
<ak.Match object: matched=True, span=(0, 5); matched=True, span=(0, 5); matched=True, span=(0, 5); ... ; matched=True, span=(0, 5); matched=True, span=(0, 5); matched=True, span=(0, 5)>

>>> timeit.timeit('strings.startswith(pattern, regex=True)', globals=globals(), number=1)
0.28777529999933904

>>> pattern = 'ing \\d'
>>> timeit.timeit('strings.endswith(pattern, regex=True)', globals=globals(), number=1)
1.6858267999996315

>>> strings.search(pattern+'$')
<ak.Match object: matched=True, span=(5, 10); matched=True, span=(5, 10); matched=True, span=(5, 10); ... ; matched=True, span=(5, 10); matched=True, span=(5, 10); matched=True, span=(5, 10)>

>>> timeit.timeit('strings.endswith(pattern, regex=True)', globals=globals(), number=1)
0.3946396999999706
```
</details>

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -271,7 +271,7 @@ class Strings:
         """
         self._regex_dict = dict()
 
-    def _get_matcher(self, pattern: Union[bytes, str_scalars]):
+    def _get_matcher(self, pattern: Union[bytes, str_scalars], create: bool = True):
         """
         internal function to fetch cached Matcher objects
         """
@@ -282,11 +282,15 @@ class Strings:
             re.compile(pattern)
         except Exception as e:
             raise ValueError(e)
-        if pattern not in self._regex_dict:
+        matcher = None
+        if pattern in self._regex_dict:
+            matcher = self._regex_dict[pattern]
+        elif create:
             self._regex_dict[pattern] = Matcher(pattern=pattern,
                                                 parent_bytes_name=self.bytes.name,
                                                 parent_offsets_name=self.offsets.name)
-        return self._regex_dict[pattern]
+            matcher = self._regex_dict[pattern]
+        return matcher
 
     @typechecked
     def find_locations(self, pattern: Union[bytes, str_scalars]) -> Tuple[pdarray, pdarray, pdarray]:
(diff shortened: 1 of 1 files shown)


glitch (approved):
@pierce314159  walked me through the reasoning here.  Basically in certain cases if a matcher has already been created, we should use it, if not sometimes we want to make it and other times we don't want to bother because the other way is faster.

reuster986 (approved):

PR #966, opened 2021-11-08
Remove workarounds for Chapel 1.23 and earlier

ronawho:
Arkouda currently supports 1.24+, so remove older workarounds. While
here, reorder some Makefile flags so Chapel flags are together and -l
options come after -I/-L options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,10 +22,17 @@ else
 CHPL_FLAGS += --fast
 endif
 CHPL_FLAGS += -smemTrack=true -smemThreshold=1048576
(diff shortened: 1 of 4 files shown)


glitch (approved):
I was able to speak with @mhmerrill and he is on board with removing 1.23 support.

ronawho:
Should have captured this better in the PR message, but https://github.com/Bears-R-Us/arkouda/pull/841 changed the minimum version to 1.24, this is just removing old code that used to be used when 1.23 was supported.

Issue #903, opened 2021-08-26
Apache Parquet Support


bmcdonald3:
As I am starting to work on implementing reading in of Parquet files to Chapel arrays, I thought I would summarize what I have been finding here to get feedback and make sure I am going down the right path.

## Summary of Parquet I/O using GLib Interface
- Parquet files store columns of data, each column being of a single data type (although, different columns in the same file could be different data types)
- Columns are stored in "chunk arrays", which are not guaranteed to be contiguous in memory (which, among other reasons, make a simple pointer swap on the buffer difficult)
- Reading in Parquet columns (using the GLib interface) requires that the columns be read into Arrow arrays first, before being copied to Chapel arrays

## Interacting with Parquet files
From what I have gathered, the industry norm for interacting with Parquet files looks something like the following: 
1. Loading in the "table" (that is, all of the columns in the file) into memory
2. Reading in only a specific column on an as-needed basis to interact with
3. Keeping the table in memory for easy-access to other columns when needed

### First proposal: reading specific columns from a file
Based on this information, I have been taking the approach of allowing the reading of specific columns from the Parquet file, which looks something like this:
```chapel
var chplIntArr = getIntColumn(path="./data/file.parquet", col=0);
var chplStrArr = getStringColumn(path="./data/file.parquet", col=1);
```

### Second proposal: reading entire files at a time
```chapel
// tuple size matches the number of columns in the parquet file
(chplIntArr, chplStrArr) = readParquetFile(path="./data/file.parquet", numCols=2);
```

edit: adding this third proposal
### Third proposal: reading columns from in-memory table (not file)
When Arrow files are read, they are stored in tables and since the implementation for how this is done is largely out of our control (and based on the info in the Interacting with Parquet Files section above), this seems to be the most performant method for large files. The table is read in once and then passed to the various methods, with the data already in memory.
```chapel
var table = readParquetFile(path="./data/file.parquet");
var chplIntArr = getIntColumn(table=table, col=0);
var chplStrArr = getStringColumn(table=table, col=1);
```

## Remaining questions
1. Are you expecting multiple columns per file or just a single column per file?
2. Would it be valuable to allow the reading in of Parquet metadata/schema information to Chapel?
3. Is the entire file approach or the column-by-column approach preferred?
4. Does the information here align with your understanding?/Do you have any suggestions or alternate interface proposals?
[...]

PR #967, opened 2021-11-09
Add support for optionally reading and writing Parquet files

bmcdonald3:
This PR adds the ability to read and write int64 Parquet files in Arkouda through a `pdarray.save_parquet()` function and an `ak.read_parquet()` function. These functions behave in much the same way as the `pdarray.save()` and `ak.load()` functions (see docstrings). This is only the initial support, and more dtypes will likely be added as use necessitates.

For reading Parquet files, the Arrow C++ library was chosen due to library completeness and portability. The actual functionality is written in C++, with light C wrappers to allow leveraging of the Chapel C interoperability features, which requires the C++ code to be compiled into an object file prior to calling the functions within Chapel.

On the Chapel side, the functionality is modeled off of the HDF5 code, but currently only supports int64 and int32 Arrow datatypes, which are read into int64 pdarrays.

Performance numbers collected on a single node Cray CS:
Function  | Parquet (GiB/s) | HDF5 (GiB/s)
-- | -- | --
Read | 0.65 | 2.29
Write | 0.11 | 2.81

Potential next steps:
- Better error handling
- Improve performance
- Extend to support more reading of Arrow types to pdarrays
- Extend to support Arkouda Strings
- Support append mode
- Auto detect arrow/parquet support

Parquet as an optional dependency:
Running a `make` with this PR will build Arkouda as normal, without requiring Arrow. Only when the environment variable `ARKOUDA_SERVER_PARQUET_SUPPORT` is set will the Arrow-requiring files be pulled into the build. This functions in much the same way as ZMQ and HDF5, where the path can be specified in `Makefile.paths` and there is a provided `make install-arrow` command to download Arrow from offline and build from source.

Closes https://github.com/Bears-R-Us/arkouda/issues/903

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -6,6 +6,15 @@
 #include <parquet/arrow/reader.h>
 #include <parquet/arrow/writer.h>
 
+/*
+ C++ functions
+ -------------
+ These C++ functions are used to call into the Arrow library
+ and are then called to by their corresponding C functions to
+ allow interoperability with Chapel. This means that all of the
+ C++ functions must return types that are C compatible.
+*/
+
 int cpp_getNumRows(const char* filename) {
   std::shared_ptr<arrow::io::ReadableFile> infile;
   PARQUET_ASSIGN_OR_THROW(
@@ -120,6 +129,16 @@ const char* cpp_getVersionInfo(void) {
 void cpp_free_string(void* ptr) {
   free(ptr);
 }
+
+/*
+ C functions
+ -----------
+ These C functions provide no functionality, since the C++
+ Arrow library is being used, they merely call the C++ functions
+ to allow Chapel to call the C++ functions through C interoperability.
+ Each Arrow function must have a corresponding C function if wished
+ to be called by Chapel.
+*/
 
 extern "C" {
   int c_getNumRows(const char* chpl_str) {
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -1,0 +1,151 @@
+#include "ArrowFunctions.h"
+
+#include <iostream>
+#include <arrow/api.h>
+#include <arrow/io/api.h>
+#include <parquet/arrow/reader.h>
+#include <parquet/arrow/writer.h>
+
+int cpp_getNumRows(const char* filename) {
+  std::shared_ptr<arrow::io::ReadableFile> infile;
+  PARQUET_ASSIGN_OR_THROW(
+      infile,
+      arrow::io::ReadableFile::Open(filename,
+                                    arrow::default_memory_pool()));
+
+  std::unique_ptr<parquet::arrow::FileReader> reader;
+  PARQUET_THROW_NOT_OK(
+    parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
+  return reader -> parquet_reader() -> metadata() -> num_rows();
+}
+
+int cpp_getType(const char* filename, const char* colname) {
+  std::shared_ptr<arrow::io::ReadableFile> infile;
+  PARQUET_ASSIGN_OR_THROW(
+      infile,
+      arrow::io::ReadableFile::Open(filename,
+                                    arrow::default_memory_pool()));
+
+  std::unique_ptr<parquet::arrow::FileReader> reader;
+  PARQUET_THROW_NOT_OK(
+      parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
+
+  std::shared_ptr<arrow::Schema> sc;
+  std::shared_ptr<arrow::Schema>* out = &sc;
+  PARQUET_THROW_NOT_OK(reader->GetSchema(out));
+
+  int idx = sc -> GetFieldIndex(colname);
+  if(idx == -1) // TODO: error colname not in schema
+    idx = 0;
+  auto myType = sc -> field(idx) -> type();
+
+  if(myType == arrow::int64())
+    return ARROWINT64;
+  else if(myType == arrow::int32())
+    return ARROWINT32;
+  else // TODO: error type not supported
+    return ARROWUNDEFINED;
+}
+
+void cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colname, int numElems) {
+  auto chpl_ptr = (int64_t*)chpl_arr;
+
+  std::shared_ptr<arrow::io::ReadableFile> infile;
+  PARQUET_ASSIGN_OR_THROW(
+      infile,
+      arrow::io::ReadableFile::Open(filename,
+                                    arrow::default_memory_pool()));
+
+  std::unique_ptr<parquet::arrow::FileReader> reader;
+  PARQUET_THROW_NOT_OK(
+      parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
+  std::shared_ptr<arrow::ChunkedArray> array;
+
+  std::shared_ptr<arrow::Schema> sc;
+  std::shared_ptr<arrow::Schema>* out = &sc;
+  PARQUET_THROW_NOT_OK(reader->GetSchema(out));
+
+  auto idx = sc -> GetFieldIndex(colname);
+
+  // TODO: error: schema does not contain dsetname
+  if(idx == -1)
+    idx = 0;
+
+  PARQUET_THROW_NOT_OK(reader->ReadColumn(idx, &array));
+
+  int ty = cpp_getType(filename, colname);
+  std::shared_ptr<arrow::Array> regular = array->chunk(0);
+
+  if(ty == ARROWINT64) {
+    auto int_arr = std::static_pointer_cast<arrow::Int64Array>(regular);
+
+    for(int i = 0; i < numElems; i++)
+      chpl_ptr[i] = int_arr->Value(i);
+  } else if(ty == ARROWINT32) {
+      auto int_arr = std::static_pointer_cast<arrow::Int32Array>(regular);
+
+      for(int i = 0; i < numElems; i++)
+        chpl_ptr[i] = int_arr->Value(i);
+  }
+}
+
+void cpp_writeColumnToParquet(const char* filename, void* chpl_arr,
+                              int colnum, const char* dsetname, int numelems,
+                              int rowGroupSize) {
+  auto chpl_ptr = (int64_t*)chpl_arr;
+  arrow::Int64Builder i64builder;
+  for(int i = 0; i < numelems; i++)
+    PARQUET_THROW_NOT_OK(i64builder.AppendValues({chpl_ptr[i]}));
+  std::shared_ptr<arrow::Array> i64array;
+  PARQUET_THROW_NOT_OK(i64builder.Finish(&i64array));
+
+  std::shared_ptr<arrow::Schema> schema = arrow::schema(
+                 {arrow::field(dsetname, arrow::int64())});
+
+  auto table = arrow::Table::Make(schema, {i64array});
+
+  std::shared_ptr<arrow::io::FileOutputStream> outfile;
+  PARQUET_ASSIGN_OR_THROW(
+      outfile,
+      arrow::io::FileOutputStream::Open(filename));
+
+  PARQUET_THROW_NOT_OK(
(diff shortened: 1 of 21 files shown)


src/ArrowFunctions.cpp
```diff
@@ -0,0 +1,151 @@
...
+  PARQUET_ASSIGN_OR_THROW(
+      infile,
+      arrow::io::ReadableFile::Open(filename,
+                                    arrow::default_memory_pool()));
+
+  std::unique_ptr<parquet::arrow::FileReader> reader;
+  PARQUET_THROW_NOT_OK(
+    parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
+  return reader -> parquet_reader() -> metadata() -> num_rows();
+}
```
glitch, line 20: Are readers & file handles reusable or do you have to construct new ones each time?

Starting from the filename is good because you can't really screw it up, but I wonder if we need to be worried about the overhead (probably not since we generally are going to read a file once and be done).

Also another question, I noticed the reader  & file objects don't get explicitly closed.  Does that happen automatically / is it something we need to worry about?

src/GenSymIO.chpl
```diff
@@ -644,6 +644,168 @@ module GenSymIO {
         return new MsgTuple(repMsg,MsgType.NORMAL);
     }
 
+    proc readAllParquetMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+        if !hasParquetSupport {
+            throw getErrorWithContext(
+                   msg="Arkouda has been built without Parquet support",
```
glitch, line 650: 2 quick notes:

1. I'm not sure if we should throw an error here or return a MsgTuple with the `MsgType.Error` 
```chapel
return new MsgTuple(msg, MsgType.Error);
```
I know the server loop in `arkouda_sever` is in a try/catch and will handle it, so maybe it's fine.  It seems we have two ways of doing this and I'm not sure which one is "_most correct_" (note: you return a MsgType.Error tuple further below).  I'm curious what others think here.

2. OPTIONAL : I read this message a bit weird because of the passive negative voice (not sure what you call it).  Something like  "Arkouda was not built with Parquet support" might be clearer, your call.

src/GenSymIO.chpl
```diff
@@ -644,6 +644,168 @@ module GenSymIO {
...
+                return new MsgTuple(errorMsg, MsgType.ERROR);
+            }
+
+            var dsetdom = dsetlist.domain;
+            var filedom = filelist.domain;
+            var dsetnames: [dsetdom] string;
+            var filenames: [filedom] string;
+            dsetnames = dsetlist;
+
+            if filelist.size == 1 {
```
glitch, line 711: Is it possible for filelist.size to ever equal zero?  We test for 1 here (and then check to see if it is blank) or else we assume it is greater than 1.

src/GenSymIO.chpl
```diff
@@ -644,6 +644,168 @@ module GenSymIO {
...
+            } else {
+                filenames = filelist;
+            }
+
+            var fileErrors: list(string);
+            var fileErrorCount:int = 0;
+            var fileErrorMsg:string = "";
+            var sizes: [filedom] int;
+            var ty = getArrType(filenames[filedom.low],
+                                dsetlist[dsetdom.low]);;
```
glitch, line 738: Minor: you have a double semi-colon here.

src/GenSymIO.chpl
```diff
@@ -644,6 +644,168 @@ module GenSymIO {
...
+
+                // Only integer is implemented for now, do nothing if the Parquet
+                // file has a different type
+                if ty == ArrowTypes.int64 || ty == ArrowTypes.int32 {
+                  var entryVal = new shared SymEntry(len, int);
+                  readFilesByName(entryVal.a, filenames, sizes, dsetname);
+                  var valName = st.nextName();
+                  st.addEntry(valName, entryVal);
+                  rnames.append((dsetname, "pdarray", valName));
+                }
```
glitch, line 800: Should we return an error tuple if the type is not supported?  I know it's probably already stopped from being sent by the client.

src/GenSymIO.chpl
```diff
@@ -1235,6 +1397,71 @@ module GenSymIO {
...
     }
 
+    proc toparquetMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+        if !hasParquetSupport {
+            throw getErrorWithContext(
+                   msg="Arkouda has been built without Parquet support",
+                   lineNumber=getLineNumber(),
+                   routineName=getRoutineName(), 
+                   moduleName=getModuleName(),
+                   errorClass="ParquetBuildError");
```
glitch, line 1407: Same comment about `MsgType.Error` and error tuple as in `readAllParquetMsg`

src/Parquet.chpl
```diff
@@ -0,0 +1,109 @@
...
+  proc getVersionInfo() {
+    extern proc c_getVersionInfo(): c_string;
+    extern proc strlen(str): c_int;
+    extern proc c_free_string(ptr);
+    var cVersionString = c_getVersionInfo();
+    defer {
+      c_free_string(cVersionString: c_void_ptr);
+    }
+    var ret = try! createStringWithNewBuffer(cVersionString,
+                                             strlen(cVersionString));
```
glitch, line 26: If this fails it takes out the server process.  Should we throw an error here instead?

src/Parquet.chpl
```diff
@@ -0,0 +1,109 @@
...
+    }
+    return (subdoms, (+ reduce lengths));
+  }
+
+  proc domain_intersection(d1: domain(1), d2: domain(1)) {
+    var low = max(d1.low, d2.low);
+    var high = min(d1.high, d2.high);
+    if (d1.stride !=1) && (d2.stride != 1) {
+      //TODO: change this to throw
+      halt("At least one domain must have stride 1");
```
glitch, line 45: Yes, I agree, we should throw here.

glitch (commented):
Overall this looks really good!

My main questions are about closing parquet files & readers, whether we should throw an error vs. return MsgTuple with MsgType.Error from the read/write parquet chapel functions, and the `try!` versus throwing an error.

Great work!

ronawho:
@glitch already found this, but FYI this is based on https://github.com/Bears-R-Us/arkouda/pull/966 just because it reduces the amount of code we need to bring in for the parquet dependency check.

Ben also mentioned this a bit in the PR, but the intent of this is to be the first usable draft that we will then build on. The goal was to get something usable in that we can start getting feedback on as well as start running nightly correctness/performance testing. There are a number of known next steps (and we'd be interested in input on prioritization on those.) Note that better error-handling is one of those next steps. Today there are some halts and it's worth noting in the C++ code we currently throw an uncaught exception, which is effectively the same as a halt. Turning that into proper error-handling should be straight forward, but is a fair amount of boilerplate code and we didn't want to make those changes just yet in case larger changes to the core code were required.

src/ArrowFunctions.cpp
bmcdonald3: The issue here is we can't pass a C++ object to C/Chapel, so there is no way for us to get the opened file back to Chapel to reuse.

Yes, the files are closed automaticall (see https://arrow.apache.org/docs/cpp/api/formats.html#parquet-reader)

src/GenSymIO.chpl
bmcdonald3: Ya, I'd be open to either returning the error message or returning, so whatever anyone has a preference for I think both ways are fine. As Elliot mentioned though, we are aware the error handling is not perfect with this initial PR and were hoping to get this initial work merged before over-emphasizing the error handling (unless that is a problem).

src/GenSymIO.chpl
bmcdonald3: Regarding this and the following error-related comments, I was anticipating this is a follow-up PR as Elliot mentioned, but I do agree there are many places errors could be handled better.

src/GenSymIO.chpl
bmcdonald3: I suspect not -- this code was directly copy/pasted from the HDF5 code (see line 442), so since the HDF5 code seems to work just fine, I would expect this identical code to behave in the same way

src/ArrowFunctions.cpp
ronawho: FWIW I don't expect there to be a noticeable overhead from not reusing. The actual file read/write should dominate the time.

glitch (approved):
I'd like to get a few more reviews from other team members but overall I think it looks pretty good.

pierce314159 (approved):
Looks good to me! Nice Work

reuster986 (changes requested):
Great work, @bmcdonald3 ! I like the design here, especially how you handled the optional inclusion of parquet functionality at compile time in the makefile. I think it's a good prototype for how we can think about extending arkouda with other modular content.

I'm on board with the intention expressed by you and @ronawho to get basic functionality working in this PR and then come back and add more robust error handling, etc. later.

I just requested one minor change regarding how wildcards are handled, because I want us to adhere to `ak.read_all` usage, rather than `ak.load_all`. 

Well done!

ronawho:
> I'm on board with the intention expressed by you and @ronawho to get basic functionality working in this PR and then come back and add more robust error handling, etc. later.

@reuster986 do you (or others) have any input on prioritization of next steps? (there's an unordered list of potential next steps in the PR description)

bmcdonald3:
Thanks @reuster986, just pushed up some code that addresses your suggestion, assuming that I understand sufficiently. Also, I would note that the method for optional inclusion was work from @ronawho, so I cannot take credit for that!

tests/parquet_test.py
```diff
@@ -8,16 +8,15 @@
 verbose = True
 
 def compare_values(arr1, arr2):
-    for i in range(len(arr1)):
-        if arr1[i] != arr2[i]:
-            print(arr1[i], 'does not match', arr2[i], 'at index', i)
-            return 1
+    if (arr1 != arr2).all():
```
reuster986, line 11: I think this should be `if (arr1 != arr2).any():`. 

As written, arrays will compare equal unless *all* of their values differ.

reuster986 (changes requested):
Other than a small change on the test logic, everything looks good.

mhmerrill:
@reuster986 @bmcdonald3 we would like to merge this week, is this ready to go or do we need to do more?

tests/parquet_test.py
reuster986: @glitch I believe `arr1` and `arr2` are arkouda arrays. Although you could do `np.array_equal(arr1.to_ndarray(), arr2.to_ndarray())`, I guess I have a personal preference for staying in arkouda to avoid array transfers (even though those are fast now and these arrays are small, so at this point it's mostly a stylistic choice). 

Also, it's been a while since I looked, but I think most of the test suite uses the `(a == b).all()` pattern, so it might be good to use that for consistency.

reuster986:
@mhmerrill I reviewed earlier this morning and requested a small logic change in the tests (see discussion above). After that, I think it's ready to go.

tests/parquet_test.py
bmcdonald3: @reuster986 you are correct, `(a == b).all()` is the pattern elsewhere, and it seems that what other tests do most often is just `assertTrue`, rather than do a `compare_values` function like I was doing, so I have switched to that format for the test to be more consistent with other tests.

reuster986 (approved):

bmcdonald3:
@mhmerrill This is all that we had planned adding for this initial Parquet support effort. We do have a couple of follow up PRs nearly ready that improve the error handling, performance, and support additional Parquet types among a few other small changes, but we wanted to get a few nights worth of testing data on this initial work before adding in those other changes if that sounds good to you.

mhmerrill:
@bmcdonald3 @ronawho should we have a separate  CI process to check this? or does the CI already check this?

ronawho:
> @bmcdonald3 @ronawho should we have a separate CI process to check this? or does the CI already check this?

I think we discussed this offline, but it's tested by default in CI testing -- https://github.com/Bears-R-Us/arkouda/pull/967/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R7

PR #969, opened 2021-11-10
In1d benchmark

### LINKED ISSUE #968: Make `in1d` benchmark
The operation `ak.in1d(a, b)`, which tests each element of `a` for membership in `b`, is an important kernel in many workflows, but we don't have a benchmark for it. Arkouda uses 3 different algorithms for `in1d`, depending on the size of `b`:
1. When `b.size <= In1dMsg.sbound`, it performs a sequence of equality checks against elements of `b` and ORs the results.
  https://github.com/Bears-R-Us/arkouda/blob/39dd7a26ae20ef30939dee44f5f43bc8eefd8237/src/In1dMsg.chpl#L23
2. When `In1dMsg.sbound < b.size <= In1dMsg.mbound`, it sends `b` to each locale and creates an associative domain for fast lookup.
  https://github.com/Bears-R-Us/arkouda/blob/39dd7a26ae20ef30939dee44f5f43bc8eefd8237/src/In1dMsg.chpl#L28
3. When `b.size > In1dMsg.mbound`, it uses a sort-based approach inspired by numpy.

I don't think we need to benchmark the first regime, only the latter two. Also, `in1d` only operates on `int64` and `Strings`. So my plan for a benchmark comprises four subproblems:
- `int64` medium
- `int64` large
- `Strings` medium
- `Strings` large

Outside the scope of this issue, but the regime bounds could probably use some tuning, especially on IB.

reuster986:
Will close #968 

This PR implements benchmarks for `in1d` on the four subproblems described in #968 . The hard-coded size of the second argument to `in1d` is necessarily `2**25` (to access the largest regime), so the effective problem size has a large lower bound, regardless of the `--size` option. On my laptop, the benchmarks run fine in single-locale mode, but run out of memory in multi-locale mode. @ronawho will this be a problem for the benchmark systems?

### DIFF
--- a/benchmarks/str-in1d.py
+++ b/benchmarks/str-in1d.py
@@ -11,7 +11,7 @@ LARGE = 10**6 + 1
 MAXSTRLEN = 5
 
 def time_ak_in1d(size, trials):
-    print(">>> arkouda int64 in1d")
+    print(">>> arkouda string in1d")
     cfg = ak.get_config()
     N = size * cfg["numLocales"]
     a = ak.random_strings_uniform(1, MAXSTRLEN, N)
--- a/benchmarks/str-in1d.py
+++ b/benchmarks/str-in1d.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3                                                         
+
+import time, argparse
+import arkouda as ak
+
+# Must be less than src/SegmentedArray.chpl:in1dAssocSortThreshold, which defaults to 10**6
+MEDIUM = 10**6 - 1
+# Must be greater than mbound
+LARGE = 10**6 + 1
+# Controls how many unique values are possible
+MAXSTRLEN = 5
+
+def time_ak_in1d(size, trials):
+    print(">>> arkouda int64 in1d")
+    cfg = ak.get_config()
+    N = size * cfg["numLocales"]
+    a = ak.random_strings_uniform(1, MAXSTRLEN, N)
+
+    for regime, bsize in zip(('Medium', 'Large'), (MEDIUM, LARGE)):
+        print("{} regime: numLocales = {}  a.size = {:,}  b.size = {:,}".format(regime, cfg["numLocales"], N, bsize))
+        b = ak.random_strings_uniform(1, MAXSTRLEN, bsize)
+        timings = []
+        for _ in range(trials):
+            start = time.time()
+            c = ak.in1d(a, b)
+            end = time.time()
+            timings.append(end - start)
+        tavg = sum(timings) / trials
+        print("{} average time = {:.4f} sec".format(regime, tavg))
+        bytes_per_sec = (a.size * 8 + a.nbytes + b.size * 8 + b.nbytes) / tavg
+        print("{} average rate = {:.2f} GiB/sec".format(regime, bytes_per_sec/2**30))
+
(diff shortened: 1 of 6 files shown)


ronawho:
> On my laptop, the benchmarks run fine in single-locale mode, but run out of memory in multi-locale mode. @ronawho will this be a problem for the benchmark systems?

Should be fine. I think the smallest system has 128 GB of memory per node.

ronawho:
Here's results on 16-node-xc:

```
number of trials =  1
>>> arkouda int64 in1d
Medium regime: numLocales = 16  a.size = 160,000,000  b.size = 33,554,431
Medium average time = 68.3911 sec
Medium average rate = 0.02 GiB/sec
Large regime: numLocales = 16  a.size = 160,000,000  b.size = 33,554,433
Large average time = 1.2008 sec
Large average rate = 1.20 GiB/sec
```

So the medium version seems a lot slower, so we almost certainly have some threshold tuning to do, but that's for a different issue/PR (feel free to open and assign to me)

Let me play with the benchmark a little. It's a bit different from others since we're hardcoding part of the problem size and I'm wondering if there's a cleaner way we can keep the benchmark and the server thresholds in sync.

benchmarks/str-in1d.py
```diff
@@ -0,0 +1,64 @@
...
+
+# Must be less than src/SegmentedArray.chpl:in1dAssocSortThreshold, which defaults to 10**6
+MEDIUM = 10**6 - 1
+# Must be greater than mbound
+LARGE = 10**6 + 1
+# Controls how many unique values are possible
+MAXSTRLEN = 5
+
+def time_ak_in1d(size, trials):
+    print(">>> arkouda int64 in1d")
```
ronawho, line 14: int -> string?

ronawho:
And string results for 16-node-xc:


```
>>> arkouda str in1d
Medium regime: numLocales = 16  a.size = 160,000,000  b.size = 999,999
Medium average time = 0.3610 sec
Medium average rate = 4.78 GiB/sec
Large regime: numLocales = 16  a.size = 160,000,000  b.size = 1,000,001
Large average time = 2.7356 sec
Large average rate = 0.63 GiB/sec
```

benchmarks/str-in1d.py
reuster986: yes, good catch (copy/paste error)

reuster986:
Wow, looks like the `int64` threshold is way too high, and the string threshold is a bit low... would not have guessed that.

mhmerrill (commented):
@reuster986 looks ok to me. Is this ready to go?
@ronawho  are you ok with this?

reuster986:
@mhmerrill , since @ronawho had indicated he wanted to find a better way to synchronize the server and benchmark thresholds, I'll defer to him on whether he wants to add that work to this PR or save it for a separate one. If opts for a separate PR, then I think this one is ready to merge.

ronawho (approved):
Let's go with this and we can try to think of better ways to sync server/benchmark thresholds as part of https://github.com/Bears-R-Us/arkouda/issues/970

ronawho:
@reuster986 sorry for missing this before, but these need to be added to run_benchmarks.py

PR #974, opened 2021-11-16
Closes #971 by escaping double quotes in dataset names

### LINKED ISSUE #971: `read_all` and `ls_hdf` choke on unsanitized dataset names
Reproducer:

```python
my_arrays = {'foo"0"': ak.arange(100),
             'bar"': ak.arange(100)}
ak.save_all(my_arrays, './bad_dataset_names')
ak.read_all('./bad_dataset_names*')
[JSONDecodeError]

ak.ls_hdf('./bad_dataset_names*')
[JSONDecodeError] 
```

This might seem pretty obscure, but unfortunately we have real datasets whose column names are mangled in this exact way and it is quite difficult to change them upstream. Explicitly supplying the `datasets=` kwarg to leave out the mangled columns doesn't work because `read_all` now calls `get_datasets` to check for existence of all user-specified datasets. Using `allow_errors=True` with `read_all` doesn't seem to help either.

I suspect this behavior was introduced in #956 but I'm not sure. @glitch would you mind taking a look?

glitch:
Closes #971 by escaping double quotes in dataset names
for lshdfMsg and readAllHdfMsg/_buildReadAllHdfMsgJson

### DIFF
--- a/src/GenSymIO.chpl
+++ b/src/GenSymIO.chpl
@@ -279,12 +279,13 @@ module GenSymIO {
             repMsg = "[";  // Manual json building Chapel <= 1.24.1
             var first = true;
             for i in items {
+                i = i.replace(Q, ESCAPED_QUOTES, -1);
                 if first {
                     first = false;
                 } else {
                     repMsg += ",";
                 }
-                repMsg += '"' + i + '"';
+                repMsg += Q + i + Q;
             }
             repMsg += "]";
         } catch e : Error {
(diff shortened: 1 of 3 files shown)


src/GenSymIO.chpl
```diff
@@ -279,12 +279,13 @@ module GenSymIO {
             repMsg = "[";  // Manual json building Chapel <= 1.24.1
             var first = true;
             for i in items {
+                i = i.replace(Q, ESCAPED_QUOTES, -1);
```
pierce314159, line 282: i think `count = -1`  by default for `replace`

pierce314159 (approved):
Looks good to me!

src/GenSymIO.chpl
glitch: It is the default, I decided to be explicit in case the default changes.

reuster986 (approved):
thanks @glitch !

PR #976, opened 2021-11-16
Closes #975 by changing H5Oget_info_by_name call to explicitly use version 1.

### LINKED ISSUE #975: Compilation failure on newer versions of HDF5
@mhmerrill was getting a compilation failure with HDF5 version 1.12.1 and it looks like the macro compatibility of HDF5 between versions 1.10.x and 1.12.x  has an issue with function H5Oget_info_by_name (versions 1 vs 3, the proxy function now points to 3 which takes more args).

For now we should explicitly call version 1.

glitch:
Closes #975 by changing H5Oget_info_by_name call to explicitly use version 1.


mhmerrill (approved):
looks ok to me

mhmerrill:
since it's a one character change and gets me compiling again I am going to merge it

Issue #973, opened 2021-11-16
Create benchmark for sorting relevant distributions of data


reuster986:
I can't seem to find the discussion right now, but we've been mulling over whether to explore a different sorting algorithm, such as the `twoArrayRadixSort` already in Chapel. The current LSD radix sort has the advantages of simplicity and complete insensitivity to data distribution, which are powerful. However, there are conditions (such as mostly-sorted input) under which an MSD radix sort would perform significantly faster, and it's not clear to me which sort is best for the majority of data distributions we are likely to encounter in the wild.

To that end, I'd like to create a benchmark that evaluates the performance of sorting and co-sorting on several distributions of data. For now, only the existing LSD radix sort would be used, and I would expect the performance to be the same on all inputs of a given size, regardless of distribution. But if/when we include the Chapel `twoArrayRadixSort` as a runtime choice, the benchmarking infrastructure will exist to compare the two algorithms side by side.

Cases I would like to test:
- [x] uniform random integers
  - `[0, 2**16)`
  - `[0, 2**32)`
  - `[0, 2**64)`
- [x] power-law integers and floats
- [x] RMAT-generated vertex identifiers (ints)
- [x] block-sorted data, i.e. `concatenate(arrays)`, where each `arrays[i]` is sorted
- [x] refinements, e.g. `coargsort([a, b])`, where `a` is sorted but `b` is not
- [x] `datetime64[ns]`-like data, i.e. values whose range is much smaller than their magnitude
- [x] IPv4/IPv6-like data, e.g. with 90% of values in `[0, 2**32)` and 10% in `[2**32, 2**128)` (this would require a coargsort)
- [ ] Strings of uniformly distributed length
- [ ] Strings with log-normally distributed length

Any other suggestions or additional cases?

PR #977, opened 2021-11-16
Data distributions for sort benchmarking #973

reuster986:
This PR implements (co)argsort benchmarks for most of the data distributions described in #973 , with the exception of strings. It also creates a new suite of benchmarks for displaying them as a group.

@ronawho I've marked this as a draft because I think I need help getting the graph generation infrastructure to work. I believe it's correctly generating the `sort-cases.dat`, but for some reason, it won't plot the new benchmark graphs. Would you mind taking a look?


ronawho:
Just wanted to ACK that I've seen this review request, just haven't had a chance to get around to it yet. Hoping to take a look later today.

benchmarks/graph_infra/arkouda-sort-cases.graph
```diff
@@ -0,0 +1,41 @@
+perfkeys: uniform int64 16-bit average rate =, uniform int64 32-bit average rate =, uniform int64 64-bit average rate =, uniform float64 (0, 1) average rate =
+graphkeys: 16-bit, 32-bit, 64-bit, float64
+files: sort-cases.dat
```
ronawho, line 3: `files` is a comma separated list of files with one entry per perfkey/graphkey. So you probably want this to be:

```
files: sort-cases.dat, sort-cases.dat, sort-cases.dat, sort-cases.dat
```

Alternatively you can use `repeat-files: sort-cases.dat`, but this feature isn't as stable in our testing infrastructure so I'd probably just go the more annoying route of duplicating filenames.

reuster986: Thanks! I will hopefully get a chance to update this week.

reuster986: I updated the file list, but I still can't seem to get the new graphs to display when I run locally. Let's come back to this next week. Happy Thanksgiving!

ronawho: Not sure if this is the issue or something else, but note that you can't just "open" graphs locally (because of javascript cross-site security issues.) To browse locally you need to start an http server. https://github.com/Bears-R-Us/arkouda/blob/master/benchmarks/graph_infra/README.md should have instructions for how to do that.

reuster986: Oh yeah, I've been running a local http server, but no luck. But now I'm wondering if it's somehow caching the old index or something. Anyway, you're off this week, right? You seriously do not have to worry about this right now :)

ronawho: Sounds good. 

Mostly off :) In for a bit today to finish up some paperwork I put off, and I was using this to procrastinate even more :)

reuster986: Ah, say n'more, say n'more. Happy to help :)

benchmarks/sort-cases.py
```diff
@@ -0,0 +1,230 @@
...
+import arkouda as ak
+
+def is_cosorted(data):
+    # (b[0] > a[0]) | ((b[0] == a[0]) & recurse(a[1], b[1]))
+    def helper(x, right):
+        return (x[1:] > x[:-1]) | ((x[1:] == x[:-1]) & right)
+    right = ak.ones(data[0].size - 1, dtype=ak.bool)
+    for x in reversed(data):
+        right = helper(x, right)
+    return right.all()
```
pierce314159, line 14: Is this something that would be worth adding to arkouda as a counterpart to `ak.is_sorted`? I don't have a good sense of how often this would come up

Also I really like the way this function is implemented, it's super slick

pierce314159 (approved):
I'm not super familiar with the techniques used to generate the data for the different distributions (particularly `power_law`, `rmat`, and `IP_like`), so I can't speak to their correctness. Having said that, this looks good to me!

benchmarks/sort-cases.py
mhmerrill: we could definitely make this perform a lot better if we did it on the server side, IMHO.

mhmerrill:
@pierce314159 @reuster986 is this ready to go? it seems to me to be disjoint enough that if it needs to be fixed we can still put it in and fix if later...

pierce314159:
> @pierce314159 @reuster986 is this ready to go? it seems to me to be disjoint enough that if it needs to be fixed we can still put it in and fix if later...

I think it should be good to go! Unless @reuster986 thinks otherwise

reuster986:
@mhmerrill I was hoping to add `twoArrayRadixSort` to all these benchmarks, to compare the two algorithms. I almost got this done on Tuesday, but not quite.

reuster986:
@mhmerrill @pierce314159 @ronawho I added the `twoArrayRadixSort` to these benchmarks, and I believe this PR is ready for re-review. I also pulled in all the recent merges from master -- apologies if this makes things hard to view; my git-fu is rather weak.

pierce314159 (approved):

ronawho:
@reuster986 it looks like your final merge with master removed running this from run_benchmarks -- https://github.com/Bears-R-Us/arkouda/pull/977/commits/ac7eac6b60b8f596e5fcadfe60c1d5ff466b57c7

reuster986:
@ronawho oops! I'll put in a PR shortly.

Issue #979, opened 2021-11-18
`make test` hangs during categorical test on my iMac


mhmerrill:
`make test` hangs during categorical test on my iMac
This happened after a `brew upgrade` to get the newest version of chapel but I think I also got newer versions of HDF5 and ZMQ potentially.
I will check it out later today and provide version number and config  info here.

glitch:
This _may_ be related to Issue https://github.com/Bears-R-Us/arkouda/issues/924 and PR https://github.com/Bears-R-Us/arkouda/pull/925

Can you post the version of ZMQ that's being used?

mhmerrill:
the server is linking against these libraries:
Found ZMQ version: 4.3.4
Found HDF5 version: 1.12.1

glitch:
I should note I'm using `pyzmq -> 20.0.0` which in turn is using `zmq 4.3.3` and my HDF5 version is `1.10.6` via `h5py-> v2.10.0`
I believe @pierce314159 is using zmq 4.3.4, so I don't think that is the issue.

mhmerrill:
> I should note I'm using `pyzmq -> 20.0.0` which in turn is using `zmq 4.3.3` and my HDF5 version is `1.10.6` via `h5py-> v2.10.0`
> I believe @pierce314159 is using zmq 4.3.4, so I don't think that is the issue.

@glitch i think it's the HDF5

mhmerrill:
`2021-11-30:14:06:27 [arkouda_server] main Line 266 INFO [Chapel] >>> "lshdf" "[\"/Users/mike/chapel/arkouda/arkouda/categorical_test/tmp5zt4ii39/cat-save-test_LOCALE0000\"]"
Abort trap: 6`

glitch:
Linking issue #975 , I think I have a working fix for this.  Sent @mhmerrill a patch to test on his machine.  If it looks good I'll post a PR for it.  Essentially it adds an arkouda/chapel ccflag to force HDF5 API 110 over 112 as @ronawho mentioned in #975 and reverts the change of `help_h5ls.c` to use the proxy function & struct versions again versus the explicit versions.

ronawho:
Just as a point of reference https://github.com/glitch/arkouda/commit/83ecf6ef931717f82a0d0790202549792d798839 works for me. Without that I was seeing warnings during the build and the same hangs Mike was(I'm on my mac with HDF5 Version 1.12.1)

PR #987, opened 2021-12-01
Force HDF5 110 API for HDF5 1.10.x and 1.12.x support

### LINKED ISSUE #975: Compilation failure on newer versions of HDF5

glitch:
Closes #975 and #979
Forces HDF5 API version 110 so we can be compatible with both 1.10.x and 1.12.x versions of HDF5.  Updates help_h5ls.c to use H5 110 API

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@ default: $(DEFAULT_TARGET)
 VERBOSE ?= 0
 
 CHPL := chpl
+
+# We need to make the HDF5 API use the 1.10.x version for compatibility between 1.10 and 1.12
+CHPL_FLAGS += --ccflags="-DH5_USE_110_API"
+
 CHPL_DEBUG_FLAGS += --print-passes
 ifdef ARKOUDA_DEVELOPER
 CHPL_FLAGS += --ccflags="-O1"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,13 +13,21 @@ default: $(DEFAULT_TARGET)
 VERBOSE ?= 0
 
 CHPL := chpl
+
+# We need to make the HDF5 API use the 1.10.x version for compatibility between 1.10 and 1.12
+CHPL_HDF5_COMPAT := -DH5_USE_110_API
+
+# We need to build up compiler flags for Chapel --ccflags arg because they go in quotes
+CHPL_CC_FLAGS := $(CHPL_HDF5_COMPAT)
```
ronawho, line 21: Note that `--ccflags` stacks, so instead of building up all the args you could use it multiple times. e.g. instead of `--ccflags="-O1 -DH5_USE_110_API"` you could instead do `--ccflags="-O1" ccflags="-DH5_USE_110_API"`

That would probably make this diff smaller since I _think_ all you would need is `CHPL_FLAGS += --ccflags="-DH5_USE_110_API"`, but I'm not positive.

ronawho (approved):
Works for me (resolves build warnings and runtime failures). I left an inline comment that may clean up implementation, but I'm good with this as-is.

Makefile
glitch: Ah, I wasn't sure about the stacking that's good to know.  I was having issues with the quoted string making it over to the chapel based unit tests.  I'm going to see if I can do the stacking thing without making the escape-quoting stuff go crazy.

ronawho: I'm not sure that the quotes are even needed with stacking. I can't remember why I included them in the first place, it should only be needed if your have multiple options separated with a space.

ronawho (approved):

mhmerrill (approved):
looks ok to me

PR #993, opened 2021-12-06
Add handling of errors in Parquet code and minor clean up 

bmcdonald3:
With the initial Parquet work in https://github.com/Bears-R-Us/arkouda/pull/967, all errors in the C++ code are unhandled and cause the server to crash. This PR adds in Parquet error handling by catching the errors in the C++ code and passing the messages back to the Chapel code so helpful error messages can be passed back to the client. Also, some macros were added in the C++ code to help clean up some of the boilerplate code.

### DIFF
--- a/src/Parquet.chpl
+++ b/src/Parquet.chpl
@@ -13,6 +13,7 @@ module Parquet {
   extern var ARROWINT64: c_int;
   extern var ARROWINT32: c_int;
   extern var ARROWUNDEFINED: c_int;
+  extern var ARROWERROR: c_int;
 
   enum ArrowTypes { int64, int32, notimplemented };
 
@@ -90,7 +91,7 @@ module Parquet {
             var col: [filedom] int;
             if c_readColumnByName(filename.localize().c_str(), c_ptrTo(col),
                                   dsetname.localize().c_str(), filedom.size,
-                                  c_ptrTo(pqErr.errMsg)) < 0 {
+                                  c_ptrTo(pqErr.errMsg)) == ARROWERROR {
               pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
             }
             A[filedom] = col;
@@ -106,7 +107,7 @@ module Parquet {
     
     var size = c_getNumRows(filename.localize().c_str(),
                             c_ptrTo(pqErr.errMsg));
-    if size == -1 {
+    if size == ARROWERROR {
       pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
     }
     return size;
@@ -118,7 +119,7 @@ module Parquet {
     var arrType = c_getType(filename.localize().c_str(),
                             colname.localize().c_str(),
                             c_ptrTo(pqErr.errMsg));
-    if arrType < 0 {
+    if arrType == ARROWERROR {
       pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
     }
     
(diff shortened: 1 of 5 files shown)


src/Parquet.chpl
```diff
@@ -11,8 +13,34 @@ module Parquet {
...
+    
+    proc deinit() {
+      extern proc c_free_string(ptr);
+      c_free_string(errMsg);
+    }
+
+    proc parquetError(lineNumber, routineName, moduleName) throws {
+      extern proc strlen(a): int;
+      var err;
+      try! err = createStringWithNewBuffer(errMsg, strlen(errMsg));
```
glitch, line 34: Is `try!` really what we want here?  We're already throwing, if we can't allocate a new buffer maybe we can use a default message?  I doubt this would ever really happen, but if it does, it takes out the server process.

src/Parquet.chpl
```diff
@@ -11,8 +13,34 @@ module Parquet {
...
+      // free error message before throw since it can't be used again
```
glitch, line 35: I don't see the free explicitly called here, is that ok?  Presumably `deinit` gets invoked on exit?  Otherwise maybe you could use a `defer`?

src/Parquet.chpl
```diff
@@ -48,62 +76,98 @@ module Parquet {
...
+      
+
+  proc processParquetFilenames(filenames: [] string, matchingFilenames: [] string) throws {
+    var warnFlag: bool;
+    if matchingFilenames.size > 0 {
+      warnFlag = true;
+    } else {
+      warnFlag = false;
+    }
+    return warnFlag;
```
glitch, line 167: In theory you could condense this to `return matchingFilenames.size > 0;`

Also, this seems like a weird bit of logic for a proc named `processParquetFilenames`, is something supposed to be done to the filenames?  Is this a placeholder for something later?

test/UnitTestParquetCpp.chpl
```diff
@@ -3,16 +3,34 @@ use UnitTest;
...
-                                     dsetname, numelems, rowGroupSize);
+                                     dsetname, numelems, rowGroupSize,
+                                     errMsg): int;
+  extern proc c_free_string(a);
+  extern proc strlen(a): int;
+  var errMsg: c_ptr(uint(8));
+  defer {
+    c_free_string(errMsg);
+  }
+  var causeError = "asdasdasd":c_string;
```
glitch, line 16: Maybe this should have a better default value?  (Just now realizing I'm in the Unit test, so no big deal)

src/Parquet.chpl
bmcdonald3: Ya, this is based on the `processFilenames` function in `GenSymIO` and the idea is to handle the append case in there eventually, but since it isn't supported today it is a pretty minimal function just checking if we are overwriting any existing files.

This reminds me of something that I have been thinking of around supporting append for Parquet files though -- so, Parquet does not have an explicit function to just append something to a file, so we would need to read in the entire file and then append the column onto the read file and append the data like that before writing it back out to the file. With that in mind, is supporting append for Parquet still something that would be valuable, given the performance wouldn't be comparable with the regular case? (this may not be the best place to have this discussion)

src/Parquet.chpl
bmcdonald3: Ya, `deinit` is called on exit, so that should handle the freeing, I will remove that comment, that is a residual from an intermediate step that came in this development; thanks for pointing that out.

src/Parquet.chpl
bmcdonald3: I believe that it is, it seems to be the convention used in other places in the Arkouda code base, but I could add an alternative default message in the failing case if you would prefer it that way

glitch: We've been trying to remove the number of `try!` and `halts` and shift entirely to error handling.

src/Parquet.chpl
glitch: I'm not sure about the append case, but for the current `processParquetFilenames` impl, I'm good with your explanation.

glitch (approved):

src/Parquet.chpl
bmcdonald3: Alright, pushed that change up; is that something along the lines of what you had in mind?

glitch (approved):

reuster986 (approved):
looks good!

PR #995, opened 2021-12-07
Add Arrow versioning to ak.get_config() call

bmcdonald3:
This PR adds the Arrow version to calls to `ak.get_config()` to allow users to see the version of Arrow their server has been built with. It is a bit unfortunate, but due to the circular dependency that would be created by importing the `Parquet` module into the `ServerConfig` module, we can't import the call to get the version info from the Parquet module and need to copy and paste that code into the `ServerConfig` file. 

One other idea I had was splitting that single function call into it's own module that didn't need to include `ServerConfig`, something like `ArrowVersion`, but I think I prefer just copying the code since it is relatively short in this case,  but am open to taking that approach if others prefer (or have better ideas).

### DIFF
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -92,6 +92,7 @@ module ServerConfig
             const arkoudaVersion: string;
             const ZMQVersion: string;
             const HDF5Version: string;
+            const arrowVersion: string;
             const serverHostname: string;
             const ServerPort: int;
             const numLocales: int;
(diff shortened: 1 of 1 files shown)


bmcdonald3:
Another question that would be nice to get some input on: Currently, this PR display `arrowVersion: Not built` if the Arkouda server has been built without Parquet support. Does that seem alright or is there another message that would be more helpful to display there?

It seemed to me to make sense to keep the message there short, but `Not built` still seems a bit strange to me.

mhmerrill:
Maybe either:
1) not included 
Or
2) leave it out of the object/json

src/ServerConfig.chpl
```diff
@@ -108,10 +109,28 @@ module ServerConfig
...
+          extern proc strlen(str): c_int;
+          extern proc c_free_string(ptr);
+          var cVersionString = c_getVersionInfo();
+          defer {
+            c_free_string(cVersionString: c_void_ptr);
+          }
+          arrowVNum = try! createStringWithNewBuffer(cVersionString,
+                                                     strlen(cVersionString));
+        } else {
+          arrowVNum = "Not built";
```
glitch, line 126: An alternative is `Unsupported` or `Arrow support NOT detected` or as Mike mentioned just absent is fine too.

bmcdonald3: @mhmerrill Personally, I prefer a message like Kyle suggested of something like `Unsupported` rather than just leaving it out, and my reasoning there is that someone could be expecting that the Parquet code should work and this could be a way to help figure out why it's not in that case. 

This is definitely a hypothetical, but it seems something that could happen based on the elusiveness of the `ARKOUDA_SERVER_PARQUET_SUPPORT` variable. Are you alright with having it say `arrowVersion: Unsupported` in the default case, or would you still prefer the case of leaving it out?

pierce314159 (approved):
I think this looks good. I agree that "Arrow Version: Unsupported" seems like a good way to go

glitch (approved):

PR #996, opened 2021-12-10
Refresh akutil

reuster986:
This is a straight import of the latest version of akutil. It is huge, and it does not affect anything in the arkouda CI, so I don't think it's appropriate to review this PR as a whole. The purpose of this PR is just to get all the code into `arkouda/akutil` as a staging area, so that we can start migrating it into arkouda proper piece by piece (adding CI along the way).

### DIFF
--- a/akutil/akutil/__init__.py
+++ b/akutil/akutil/__init__.py
@@ -5,3 +5,14 @@ from akutil.alignment import *
 from akutil.plotting import *
 from akutil.join import *
 from akutil.hdbscan import *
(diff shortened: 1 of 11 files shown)


glitch:
I concur.  This is an add-on package at the moment and doesn't affect the main client or server code.
One thing to consider in the future is whether this is bundled with Arkouda or if it should get its own repo in the Bears-R-Us org, but I think that question is much further down the road.

glitch (approved):

PR #998, opened 2021-12-13
Remove out of date comment about poor sort performance on IB

ronawho:
Sort performance was significantly improved with Aggregation, which
especially benefited InfiniBand. Performance on EDR/HDR IB is now ahead
of Aries so the comment is no longer accurate.

### DIFF
--- a/pydoc/usage/argsort.rst
+++ b/pydoc/usage/argsort.rst
@@ -4,8 +4,6 @@
 Sorting
 ************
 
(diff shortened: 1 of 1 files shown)


glitch (approved):

pierce314159 (approved):

reuster986 (approved):
Great to retire this comment :) Nice work @ronawho !

PR #1001, opened 2021-12-14
Closes #990: Logic error in Categorical.in1d and other methods

### LINKED ISSUE #990: Logic error in `Categorical.in1d` and other methods
`ak.in1d(a, b)` gives the wrong answer when `a` and `b` are categoricals and `b` was constructed as a slice/gather. Reproducer:

```python
s = ak.array([str(i) for i in range(10)])
s12 = s[1:3]
answer = ak.in1d(s, s12)
cat = ak.Categorical(s)
cat12 = cat[1:3]
res = ak.in1d(cat, cat12)
assert (answer == res).all()
```

This assertion fails because `test` is all `True`. The reason is because the `Categorical.in1d` method was written with the assumption that every member of `.categories` is present in a Categorical array, but this is not the case when a Categorical is constructed from a slice/gather from another Categorical, or using the `.from_codes` constructor. 

Specificially, when `cat12` is created from the slice `cat[1:3]`, it inherits the full `cat.categories`, even though only two of them are present in the array. Then, when execution reaches the code block below, `test`/`cat12` has the same categories as `self`/`cat`, so the code infers that all the elements of `self`/`cat` are in `test`/`cat12`.

https://github.com/Bears-R-Us/arkouda/blob/20906e9fdc9918e6e7533176d1b3b48caf9a0754/arkouda/categorical.py#L453-L457

Two possible solutions that aren't mutually exclusive:
- Find all methods like `.in1d` that assume all categories are present in the array and fix them to explicitly find the set of categories
- Have a flag attribute like `.exhaustive` that tells whether all categories are actually present
- Modify the constructors to always eliminate unused categories on initialization (although users could still manually tinker with categories after initialization)

@pierce314159 and @glitch , I'm interested in your thoughts on how we should approach this.

pierce314159:
This PR (Closes #990):
- Resets categories for `Categoricals` built from slices
- Resets categories for `Categorical.unique` and `Categorical.in1d`

Example:

set up (same for before and after):
```python
s = ak.array([str(i) for i in range(10)])
cat = ak.Categorical(s)
cat_slice = cat[1:3]
cat_from_codes = ak.Categorical.from_codes(ak.array([1, 2]), s)
```

before PR:
```python
# unused categories present for slices
>>> cat_slice.categories
array(['6', '0', '5', '4', '9', '2', '7', '8', '1', '3'])

# user specified categories remain unchanged
>>> cat_from_codes.categories
array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'])

# in1d returns incorrect result
>>> ak.in1d(cat, cat_slice)
array([True True True True True True True True True True])
>>> cat.in1d(cat_from_codes)
array([True True True True True True True True True True])

# unique returns incorrect result
>>> cat_slice.unique()
array(['6', '0', '5', '4', '9', '2', '7', '8', '1', '3'])
>>> ak.unique(cat_from_codes)
array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'])
```
after PR:
```python
# unused categories removed for slices
>>> cat_slice.categories
array(['1', '2'])

# user specified categories remain unchanged (but still operate logically)
>>> cat_from_codes.categories
array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'])

# in1d returns expected result
>>> ak.in1d(cat, cat_slice)
array([False True True False False False False False False False])
>>> cat.in1d(cat_from_codes)
array([False True True False False False False False False False])

# unique returns expected result
>>> cat_slice.unique()
array(['2', '1'])
>>> ak.unique(cat_from_codes)
array(['1', '2'])
```

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -85,11 +85,12 @@ class Categorical:
         self.ndim = self.codes.ndim
         self.shape = self.codes.shape
         self.name : Optional[str] = None
+        self.uses_all_categories = kwargs['uses_all_categories'] if 'uses_all_categories' in kwargs else True
 
     @classmethod
     @typechecked
-    def from_codes(cls, codes : pdarray, categories : Strings, 
-                          permutation=None, segments=None) -> Categorical:
+    def from_codes(cls, codes: pdarray, categories: Strings,
+                   permutation=None, segments=None, uses_all_categories=False) -> Categorical:
         """
         Make a Categorical from codes and categories arrays. If codes and 
         categories have already been pre-computed, this constructor saves 
@@ -121,7 +122,7 @@ class Categorical:
         if codes.dtype != akint64:
             raise TypeError("Codes must be pdarray of int64")
         return cls(None, codes=codes, categories=categories, 
-                            permutation=permutation, segments=segments)
+                   permutation=permutation, segments=segments, uses_all_categories=uses_all_categories)
 
     def to_ndarray(self) -> np.ndarray:
         """
@@ -301,8 +302,8 @@ class Categorical:
         g = GroupBy(self.codes)
         idx = self.categories[g.unique_keys]
         newvals = g.broadcast(arange(idx.size), permute=True)
-        return Categorical.from_codes(newvals, idx, permutation=g.permutation, 
-                                      segments=g.segments)
+        return Categorical.from_codes(newvals, idx, permutation=g.permutation,
+                                      segments=g.segments, uses_all_categories=True)
 
     @typechecked
     def contains(self, substr : str) -> pdarray:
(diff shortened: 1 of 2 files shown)


reuster986:
@pierce314159 everything looks good here, but I wanted to suggest an optimization that might save computational work. Because `.reset_categories()` involves a `GroupBy`, it's pretty expensive, and we might want to use a flag to track whether it's necessary. I'm envisioning a `self.uses_all_categories` flag that obeys the following logic:
1. The regular `__init__` sets `self.uses_all_categories = True`
2. The `.from_codes` constructor sets `self.uses_all_categories` according to a keyword arg that defaults to `False`
3. The `.reset_categories()` method sets `uses_all_categories=True` in the result
4. The `.in1d` and `.unique` methods only call `self.reset_categories()` if `self.uses_all_categories == False` (and similarly for `test.reset_categories()`)

What do you think of this logic? Do you think it's better to merge this PR now and implement this optimization in a separate PR?

pierce314159:
> @pierce314159 everything looks good here, but I wanted to suggest an optimization that might save computational work. Because `.reset_categories()` involves a `GroupBy`, it's pretty expensive, and we might want to use a flag to track whether it's necessary. I'm envisioning a `self.uses_all_categories` flag that obeys the following logic:
> 
>     1. The regular `__init__` sets `self.uses_all_categories = True`
> 
>     2. The `.from_codes` constructor sets `self.uses_all_categories` according to a keyword arg that defaults to `False`
> 
>     3. The `.reset_categories()` method sets `uses_all_categories=True` in the result
> 
>     4. The `.in1d` and `.unique` methods only call `self.reset_categories()` if `self.uses_all_categories == False` (and similarly for `test.reset_categories()`)
> 
> 
> What do you think of this logic? Do you think it's better to merge this PR now and implement this optimization in a separate PR?

I think this sounds good! I'll go ahead and update the PR with the flag

pierce314159:
@reuster986, I added the `uses_all_categories` flag, does this line up with what you were thinking

reuster986 (approved):
looks great!

Issue #1005, opened 2021-12-14
Refactor I/O for hdf5 & parquet to combine common code


glitch:
We now have basic Parquet support in addition to HDF5 for file reading/writing.  This is a good opportunity to refactor into common code as we move towards better file I/O abstraction

- We should look at commonalities (such as file glob/pattern naming scheme generation) and combine common server code.
- Stream-lining client code for reading / loading data regardless of file type (i.e. client issues a `load_all` with a file prefix, determine type on the server based on magic number and load dataset)
- Add a new FileInfo command which explains the dataset (i.e. how many files were found, file format, number of pdarrays/objects along with their type and dataset name which exist in the files)
- Clean up and formalize the read_hdf5, read_parquet methods.  Power users should be able to use those methods directly.  Additionally, allow a keyword arg in the load_all / read_all functions where the user can specify the format (i.e. Discover, HDF5, Parquet).

PR #1007, opened 2021-12-17
Issue 1005 file io refactor

glitch:
Issue #1005 

~~The current changes are only moving code around, there are no functional changes to the server.~~

# UPDATE
This has been updated with the modular build changes which allowed us to invert the dependency between `GenSymIO` -> `HDF5 & Parquet` code.  The end result is that reading/writing HDF5 etc. has been moved to its own module and the message/commands now use the `CommandMap` module which makes HDF5 and Parquet part of the modular build process.

I have removed the FileReader/Writer aspect and kicked that can down the road a bit.

# Previous discussion etc.
## Design
The idea I'm pursing is to create a _CommonIO_ module (FileIO can likely be renamed to CommonIO) that holds common code for read/writing complex objects (symbol entries) to some type of persistence layer (currently file system but in theory expandable to any persistence layer).  Types of procs to go in common would be things like
- File name generation
- Reader/Writer interfaces
- Reader/Writer registry for dynamically holding the writers for things like HDF5, Parquet, etc. (this is with an eye on making these things pluggable, much like @bmcdonald3 Ak modular build PR)

~~Reader/Writer Registry
This would be another internal map which would map String -> ArkoudaWriter/ArkoudaReader.  The concept would be to have the build enable optional libraries (i.e. Parquet) and upon startup create an instance of each reader/writer.  This way we could de-couple the messages from the underlying implementation.  Here's an example:  GenSymIO currently holds the proc for reading Parquet and has a dependency on the Parquet module.  The same is true for HDF5.  Following Ben's work on the modular build which puts a function call ref in a cmd dispatch map it breaks the hard-coded dependency between optionally built modules and the `arkouda_server.chpl` code.  We could basically do the same thing with readers & writers.  Whatever gets compiled in can be added to the reader/writer map.  The read/write messages would be able to pull the reader/writer implementation out of the map and invoke the appropriate function.  This would make the message more generic where the message could be simplified to "readAllMsg" vs. "readAllHdfMsg" or "readAllParquetMsg" and the file type could be determined via magic number and the appropriate writer retrieved from the registry.  (That's the general idea).~~


src/FileIO.chpl
```diff
@@ -62,6 +62,78 @@ module FileIO {
...
+     */
+    proc getMatchingFilenames(prefix : string, extension : string) throws {
+        return glob("%s_LOCALE*%s".format(prefix, extension));    
+    }
+
+    /*
+     * Returns a tuple composed of a file prefix and extension to be used to
+     * generate locale-specific filenames to be written to.
+     */
+    proc getFileMetadata(filename : string) {
```
bmcdonald3, line 121: It seems to me that "metadata" is an overloaded term, so maybe we could call this `getFileNameInfo` or something like that? (I don't particularly like that name, so maybe someone else has another idea?)

With Parquet files, metadata refers to file information such as number of elements in the file, row group size, etc., so I would be in favor of not mixing the term to mean both.

src/FileIO.chpl
```diff
@@ -85,4 +157,33 @@ module FileIO {
...
+    
+    class ArkoudaFileWriter {
+
+        proc writeDistArrayToDisk(filename: string, dsetname, A) throws {
+
+        }
+
+    }
+
+    class FileWriterRegistry {
```
bmcdonald3, line 180: Reading from the PR description, my understanding here is that this is essentially another command map, but I don't know if I understand the need.

Is there a reason that we wouldn't just use the existing command map? It seems to me to be doing the same thing.

src/Parquet.chpl
```diff
@@ -1,6 +1,7 @@
 module Parquet {
   use SysCTypes, CPtr, IO;
   use ServerErrors, ServerConfig;
+  use FileIO ;
```
bmcdonald3, line 4: Just a space before the semicolon

src/GenHDF5IO.chpl
```diff
@@ -0,0 +1,1630 @@
+module GenHDF5IO {
```
bmcdonald3, line 1: It looks to me that you can compile Arkouda without adding a `use GenHDF5IO` in `arkouda_server.chpl`; is that correct? (assuming the function calls are removed from the `select` statement)

bmcdonald3 (commented):
Like the direction that this is moving in, just wanted to confirm that the `GenHDF5IO` file does not contain anything used in `arkouda_server.chpl` (or elsewhere outside that file). 

Also, I'm not sure that I understand the reader/writer registry and wonder if the same thing could be achieved with the command map that is being added in the modular approach, but I understand that work isn't done yet and this might be a good stepping stone for that.

src/FileIO.chpl
glitch: I agree with you, but I didn't want to change the `proc` names while moving them.  Although if everyone is cool with it we could do renames at the same time.

src/FileIO.chpl
glitch: This kind of goes with your question on `GenHDF5IO` and whether or not it needs to be included in `arkouda_server`.

We need to wire the writing pieces together in some way.  Right now I think this is done down in GenSymIO, so we need some level of indirection.  The `ArkoudaFileWriter` class would act as an interface in this capacity so GenSymIO could be made to function completely off of that class.

We could wire in specific commands (i.e. writeToHDF, writeToParquet, etc.) but depending how far we want to go with I/O abstraction we could have a single proc `writeToFile` and based on a format type passed in, we could pull out a compatible writer (same for reading, maybe base it on file magic numbers).  This is what I was thinking about with a separate writer registry.  I was worried about polluting the cmd map at the expense of extra complexity.

src/GenHDF5IO.chpl
glitch: See my other comment on `FileIO.chpl`.  The short answer is that `arkouda_server.chpl` still gets this dependency by way of `GenSymIO.chpl`.  It has the same issue that Parquet does.  It needs to be wired in somewhere and either mixing it into the command map or having a separate reader/write map/registry would be a way to unlink it (along with using a more general ArkoudaFileWriter/Reader class).

reuster986 (approved):
I like how things are factored apart in this.

dep/checkArrow.chpl
```diff
@@ -1,4 +1,25 @@
...
+  }
+  var ret: string;
+  try {
+    ret = createStringWithNewBuffer(cVersionString,
+                              strlen(cVersionString));
+  } catch e {
+    ret = "Error converting Arrow version message to Chapel string";
+  }
+  return ret;
+}
```
glitch, line 22: Note: This is duplicated here to remove the dependency on the Parquet module in Arkouda server source code since that ends up bringing in some other dependencies.

src/arkouda_server.chpl
```diff
@@ -311,11 +311,7 @@ proc main() {
                     var binaryRepMsg = segStrTondarrayMsg(cmd, args, st);
                     sendRepMsg(binaryRepMsg);
                   }
-                  when "lshdf"             {repTuple = lshdfMsg(cmd, args, st);}
-                  when "readAllHdf"        {repTuple = readAllHdfMsg(cmd, args, st);}
-                  when "tohdf"             {repTuple = tohdfMsg(cmd, args, st);}
-                  when "readAllParquet"    {repTuple = readAllParquetMsg(cmd, args, st);}
-                  when "writeParquet"      {repTuple = toparquetMsg(cmd, args, st);}
+
```
glitch, line 314: These commands moved into their respective modules and get registered through the CommandMap to take advantage of our new modular build capability.

src/Parquet.chpl
```diff
@@ -180,4 +184,238 @@ module Parquet {
...
+          }
+
+          repMsg = _buildReadAllMsgJson(rnames, false, 0, fileErrors, st);
+          pqLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
+          return new MsgTuple(repMsg,MsgType.NORMAL);
+      }
+  }
+
+  proc toparquetMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+    if !hasParquetSupport {
```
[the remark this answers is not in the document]
glitch: At the moment I think we should leave it as is since Parquet support is looped into the CI workflow.  I've been toying with modifying the config file to be either in `yaml` or `ini` format to gives us more options in the build process.  Right now there are two configuration points for Parquet (and HDF5), part in the `Makefile` and the other part in `ServerModules.cfg`.  Eventually we'll get one to work off of the other.

src/GenHDF5IO.chpl
```diff
@@ -0,0 +1,2193 @@
...
```
bmcdonald3: Maybe we want to add the `Msg` prefix to this module (and the Parquet module) to be more consistent with the other module names now that we have added the message processing functions to these modules?

src/Parquet.chpl
```diff
@@ -1,13 +1,28 @@
 module Parquet {
```
bmcdonald3, line 1: Similar to the HDF5 module, I think changing this to ParquetMsg makes more sense now.

src/Parquet.chpl
```diff
@@ -180,4 +184,238 @@ module Parquet {
...
+    proc readAllParquetMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+      if !hasParquetSupport {
+          throw getErrorWithContext(
+                  msg="Arkouda has been built without Parquet support",
+                  lineNumber=getLineNumber(),
+                  routineName=getRoutineName(), 
+                  moduleName=getModuleName(),
+                  errorClass="ParquetBuildError");
+      } else {
+          use Parquet;
```
bmcdonald3, line 197: This call to `use Parquet;` in the Parquet module should be removed

src/Parquet.chpl
```diff
@@ -180,4 +184,238 @@ module Parquet {
...
+  proc toparquetMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+    if !hasParquetSupport {
+        throw getErrorWithContext(
+                msg="Arkouda has been built without Parquet support",
+                lineNumber=getLineNumber(),
+                routineName=getRoutineName(), 
+                moduleName=getModuleName(),
+                errorClass="ParquetBuildError");
+    } else {
+        use Parquet;
```
bmcdonald3, line 359: Similarly, this should be removed.

bmcdonald3:
The environment variable that we have `ARKOUDA_SERVER_PARQUET_SUPPORT` is causing some problems with this branch. If you set `ARKOUDA_SERVER_PARQUET_SUPPORT` but have the module name commented out, it will not compile. I'm not sure what the best way to fix that would be, but I think it might be a good idea to remove that environment variable now if possible.

src/Parquet.chpl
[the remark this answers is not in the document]
bmcdonald3: I'm not so sure this makes sense anymore since Arkouda won't compile if this module is included anyway, so I don't think we get much value here anymore. The other, non-message functions would also need to be wrapped in this `!hasParquetSupport` thing for this to work as intended I think.

glitch:
Moving back to draft status as there are some changes Ben & I discussed making.

glitch:
Ready for re-review:  I worked out the issue with chapel based unit test failures.  It basically hinged on including all chapel-based source files to `start_test` and the removal of the `hasParquetSupport` switch along with the dependency inversion.  Basically the cpp code never got compiled and since a `require` statement in `ParquetMsg.chpl` existed, you'd get a compilation failure message regardless of whether or not the unit test you were compiling depended on `ParquetMsg`.

Changes: 

- `ServerConfig.chpl` was updated to remove explicit arrow calls for versioning.  I had to make the cfgStr variable so that ParquetMsg can dynamically add its version info to the string which I spliced in as part of the CommandMap registration process.
- complete removal of `hasParquetSupport` flag
- Parquet->ParquetMsg and GenHDF5IO -> HDF5Msg renaming
- `make test-chapel` now works with the server modularization
- Updated `start_test` COMOPTS to handle arrow function compilation when ARKOUDA_SERVER_PARQUET_SUPPORT is set.  Note, I probably could have removed the UnitTestParquetCpp.skipif but I left it in place for now as it shouldn't hurt anything; I did remove the compopts for that since it had the `hasParquetSupport` flag.

Everything compiles and passes all tests to the best of my knowledge and I tried it toggling parquet support on & off.
I also manually check that the Arrow version info is set in the config by querying for it from a client.

One final note, if you have ARKOUDA_SERVER_PARQUET_SUPPORT set but comment out the module ParquetMsg in the ServerRegistration.cfg that piece will not be built.  If you DO have `ParquetMsg` in the registration config but DO NOT have the ARKOUDA_SERVER_PARQUET_SUPPORT env var enabled (i.e. you don't have parquet support) it will output a warning and automatically remove it from the registration list.  This way general users don't have to manually turn it off every time, but developers with support turned on can choose to manually turn it off.

bmcdonald3 (approved):
This looks good! I think the last real question I have is: Should we have the Parquet module uncommented by default? 

Today, the default behavior makes you jump through some hoops to use Parquet and I think that makes sense given the dependency requirement, but now that we have the warning, I think that may be preferred since it provides more info, so I can definitely see both sides and don't know if there is a "right" answer, but I would say I would be in favor in having the `ParquetMsg` module commented out in `ServerModules.cfg` in the source repo.

glitch:
If we comment out `ParquetMsg` in the config, we have to jump through some hoops in the CI.yml workflow to turn it back on for testing.

src/GenSymIO.chpl
```diff
@@ -1,6 +1,4 @@
 module GenSymIO {
```
bmcdonald3: Thinking about this a little more, I think that the `GenSymIO` module name is weird for this file now. The functions provide functionality like creating a pdarray with `arrayMsg()` and converting a pdarray to an ndarray with `tondarrayMsg()`, and I'm not sure if these are necessarily "IO" functions, but I don't know what I would call it instead and maybe there is more to the name than I realize.

glitch: I want to keep the PR scope somewhat confined, so I'm not going to change it in this go-round; we can always rename it in another iteration.

src/HDF5Msg.chpl
```diff
@@ -0,0 +1,2193 @@
...
+                                // do A[intersection] = file[intersection - offset]
+                                var dataspace = C_HDF5.H5Dget_space(dataset);
+                                var dsetOffset = [(intersection.low - filedom.low): C_HDF5.hsize_t];
+                                var dsetStride = [intersection.stride: C_HDF5.hsize_t];
+                                var dsetCount = [intersection.size: C_HDF5.hsize_t];
+                                C_HDF5.H5Sselect_hyperslab(dataspace, C_HDF5.H5S_SELECT_SET, c_ptrTo(dsetOffset), 
+                                                                c_ptrTo(dsetStride), c_ptrTo(dsetCount), nil);
+                                var memOffset = [0: C_HDF5.hsize_t];
+                                var memStride = [1: C_HDF5.hsize_t];
+                                var memCount = [intersection.size: C_HDF5.hsize_t];
```
bradcray, line 576: @glitch: While coming across this code today, I wanted to point out a potential misunderstanding and opportunity to clean up / optimize / simplify this code (where it looks like this code is from @mhmerrill and the dawn of time, so I'm only tagging you for your info and as the last one to touch it rather than because you wrote it to begin with :) ).

From what I can tell, these declarations are creating 1-element arrays and then passing a pointer to the first/only element in the array into the HDF5 routines.  My guess is that this is because Mike believed that `c_ptrTo()` was only supported on arrays, but it's actually supported on any Chapel expression with storage.   So where one of these uses:

```chapel
var arr = [elt: t];
...c_ptrTo(arr)...
```

it could simply use:

```chapel
var val = elt: t;
...c_ptrTo(val)...
```

The savings would be the reduction in time to create an array and its governing domain for a single element ("Why play an ace when a two will do?").  

Specifically, `c_ptrTo()` should be able to be applied to any Chapel expression that has storage where arrays are a special case (in that it points to the array's initial element rather than to the array descriptor itself).

Since this is I/O code, my guess is that these O(1) overheads are washed away by the O(n) I/O cost and won't show any significant benefit from changing.  But it still seemed worth pointing out, if not for code cleanup purposes then to stem the proliferation of this idiom when it's not needed (e.g., if you're modifying or reviewing code going forward that uses it).

I also saw a variation on this pattern elsewhere in the code where the array was declared in a more verbose way:

```chapel
var A: [0..#1] t;
A[0] = val;
...c_ptrTo(A)...
```

which looked similarly like overkill.

It _could_ be that I'm missing some other subtle reason that these want to be arrays, but @bmcdonald3 took a look as well and came to similar conclusions (and I believe tried replacing some of them with scalars and things seemed to continue to work fine).

PR #1014, opened 2022-01-05
Optimize Parquet reading by reading batches rather than creating a copy

bmcdonald3:
Today, to read in a Parquet file, we are reading the Parquet file into a temporary array and then copying the contents of the temporary array into the Chapel array.

This optimization reads "batches" of 8,192 elements of int64 directly into the Chapel array, avoiding the extra array copy and need to copy elements one by one. For int32 Parquet columns, batches of single elements are read into a temporary variable and then copied to the Chapel array to avoid having to keep a temporary copy of the column around.

Here are reading performance numbers collected on chapcs
Parquet type | Current | This Branch
-- | -- | --
int64 | 0.69 GiB/s | 1.41 GiB/s
int32 | 0.65 GiB/s | 1.29 GiB/s

So, we see about a 2x improvement in performance case and also remove the extra array copy, so have a lower memory load.

Additionally, this test adds an int32 Parquet file under `test/resources` to test int32 reading in the Chapel unit testing.


src/ArrowFunctions.cpp
```diff
@@ -99,42 +99,63 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
...
-    COPYTOCHAPEL(Int32Array, regular);
+  
+  // Currently only supports int64 and int32 Arrow types
+  if(ty == ARROWINT64 || ty == ARROWINT32) {
+    std::unique_ptr<parquet::ParquetFileReader> parquet_reader =
+      parquet::ParquetFileReader::OpenFile(filename, false);
+
+    std::shared_ptr<parquet::FileMetaData> file_metadata = parquet_reader->metadata();
+    int num_row_groups = file_metadata->num_row_groups();
+    int num_columns = file_metadata->num_columns();
```
ronawho, line 113: I think `num_columns` is unused.

src/ArrowFunctions.cpp
```diff
@@ -99,42 +99,63 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
...
+          (void)reader->ReadBatch(batchSize, nullptr, nullptr, tmpArr, &values_read);
+          for (int64_t j = 0; j < values_read; j++)
+            chpl_ptr[i+j] = (int64_t)tmpArr[j];
+          i+=values_read;
+        }
+        free(tmpArr);
+      }
+    }
   }
   return 0;
```
ronawho, line 160: Here it just returns 0 for non int32/int64 types, but presumably this should error?

ronawho (approved):

glitch (approved):

PR #1016, opened 2022-01-07
Issue #914: Uses aggregator for SegString to HDF5 writes

glitch:
Issue #914: Uses aggregator for SegString to HDF5 writes and deprecates previous shuffling implementation.

NOTE: I haven't removed the previous implementation from the code, but these changes swap the implementation... meaning you can't call the previous impl from the client.  If we want to ability to switch back & forth, I can wire in an option for that in the `tohdf` Msg upon request.

Also can someone double check my `aggregator` logic, I think it is correct but this is my first time using it ;)

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -45,7 +45,7 @@ def ls_hdf(filename : str) -> List[str]:
 
 @typechecked
 def read_hdf(dsetName : str, filenames : Union[str,List[str]],
-             strictTypes: bool=True, allow_errors:bool = False, calc_string_offsets:bool = False) \
+             strictTypes: bool=True, allow_errors:bool = False, calc_string_offsets:bool = True) \
           -> Union[pdarray, Strings]:
     """
     Read a single dataset from multiple HDF5 files into an Arkouda
@@ -209,7 +209,7 @@ def read_all(filenames : Union[str, List[str]],
              iterative: bool = False,
              strictTypes: bool = True,
              allow_errors: bool = False,
-             calc_string_offsets = False)\
+             calc_string_offsets = True)\
              -> Union[pdarray, Strings, Mapping[str,Union[pdarray,Strings]]]:
     """
     Read datasets from HDF5 files.
@@ -331,7 +331,7 @@ def read_all(filenames : Union[str, List[str]],
 
 
 @typechecked
-def load(path_prefix : str, dataset : str='array', calc_string_offsets:bool = False) -> Union[pdarray,Strings]:
+def load(path_prefix : str, dataset : str='array', calc_string_offsets:bool = True) -> Union[pdarray,Strings]:
     """
     Load a pdarray previously saved with ``pdarray.save()``.
 
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -45,7 +45,7 @@ def ls_hdf(filename : str) -> List[str]:
 
 @typechecked
 def read_hdf(dsetName : str, filenames : Union[str,List[str]],
-             strictTypes: bool=True, allow_errors:bool = False, calc_string_offsets:bool = True) \
+             strictTypes: bool=True, allow_errors:bool = False, calc_string_offsets:bool = False) \
           -> Union[pdarray, Strings]:
     """
     Read a single dataset from multiple HDF5 files into an Arkouda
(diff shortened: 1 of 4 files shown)


mhmerrill (commented):
This looks good to me.
@ronawho I'd like to see this tested on multi-locale with shorter arrays to see if it works (but I think it does)
@reuster986 I asked @glitch to put this branch into the main repo so you can get it on your system to try out if you want

ronawho:
> Also can someone double check my aggregator logic, I think it is correct but this is my first time using it ;)

Aggregator logic looks good to me. 

> @ronawho I'd like to see this tested on multi-locale with shorter arrays to see if it works (but I think it does)

Are you saying you want something in nightly testing or want me to run something manually? If manual, what test am I running and with how many locales?

mhmerrill:
> > Also can someone double check my aggregator logic, I think it is correct but this is my first time using it ;)
> 
> Aggregator logic looks good to me.
> 
> > @ronawho I'd like to see this tested on multi-locale with shorter arrays to see if it works (but I think it does)
> 
> Are you saying you want something in nightly testing or want me to run something manually? If manual, what test am I running and with how many locales?

@ronawho I guess we could just accept the code and see if it breaks your nightly testing.

mhmerrill (approved):
looks good to me.

pierce314159 (approved):

glitch:
@ronawho if it is easy, we'd like to see if this branch: https://github.com/Bears-R-Us/arkouda/tree/914-hdf5-aggregator passes nightly testing.

Basically if *think* I fixed the issue from #904 and #914 where arrays smaller than the number of locales caused write failures to HDF5.  I have re-enabled the test that previously failed and ran it on 3 locales, but I'd feel better if it passed your nightly testing as well as gave us a general idea of whether the string array writes to HDF5 were comparable to the previous shuffling implementation.

ronawho:
> @ronawho if it is easy, we'd like to see if this branch: https://github.com/Bears-R-Us/arkouda/tree/914-hdf5-aggregator passes nightly testing...

Gotcha, it's not trivial to just switch the branch nightly is running (well it is, but it messes up perf results) so I'm just testing manually with that configuration by running ` ARKOUDA_NUMLOCALES=16 make test` on 16-node-xc. 

On master, I still see the crash if I revert https://github.com/Bears-R-Us/arkouda/commit/181d7c4ae9774dc2b8909f8c3503229a864b2297. With this PR I instead see an assertion failure for testSmallStringArrayToHDF5:

```
========================================================================== short test summary info ===========================================================================
FAILED tests/io_test.py::IOTest::testSmallStringArrayToHDF5 - AssertionError: "['ab', 'cd']" != "['', 'abu0000cd']"
================================================================= 1 failed, 23 passed, 9 warnings in 15.42s ==================================================================
```

Seems to occur with 3+ nodes. (`ARKOUDA_NUMLOCALES=3 python3 -m pytest tests/io_test.py` fails)

glitch:
Thanks @ronawho, that's helpful.  It looks like it writes correctly, it's just the `fixupSegBoundaries` that's off when reading back in and re-adjusting the offsets.

glitch:
Whoa, this is strange.  When I run with 4 locales using the two string case `["ab", "cd"]` and save it, 4 files are written and only 2 contain as expected.  What was not expected was that the two files were `Locale-0000` and `Locale-0002`.  For some reason `Locale-0001` was skipped.. which means it was not hosting offsets in the rect-dist.

When reading back in we get segSubdoms:  `{0..0} {1..0} {1..1} {2..1}` which corresponds to the data in the files etc.

At this point I think it's both easiest and safest to just scan the values array for the null-byte delimiter.  The change itself is just flipping the boolean `calc_string_offsets` flag in pdarrayIO.py to True instead of False.  In the meantime, I can look into figuring out a fix for `fixupSegBoundaries`.

ronawho:
`ARKOUDA_NUMLOCALES=16 make test` is passing now.

glitch:
I reverted the `calc_strings_offsets` switch in the client and it is now False again.  Instead I decided to check when we have empty domains in the offsets/segments component of the HDF5 file and if we do, then we trigger offset calculation from the values array (null-byte-delimiter) and skip the `fixupSegstringBoundaries` call.  This side-steps the issue when not all files have offsets and the files that do are non-contiguous (in which case the `fixupSegstringBoundaries` logic doesn't work).

I feel like this is a good fit.  We honor the offsets read in the majority of cases and derive them when needed or directed.

NOTE: I was able to test the previous failing test locally with a 6 locale single node/instance.  It goes from red to green with the fix.

reuster986 (approved):

Issue #970, opened 2021-11-12
Tune `in1d` thresholds


reuster986:
See discussion on #968 and #969 - assigning to @ronawho at his request. From his measurements, it looks like the `mbound` for `int64` is too high, at least for XC, and the `in1dAssocSortThreshold` for strings is too low. I wonder how different it looks between CS and XC...

ronawho:
As part of this, we should think about better ways to sync the server/benchmark implementation strategy thresholds.

ronawho:
16-node-cs-hdr testing has been down since Dec 13 (and we only started running this test in nightly on the 14 so this is the first time running) and it turns out int in1d is timing out for the `mbound` on CS.

The slow code is:
 https://github.com/Bears-R-Us/arkouda/blob/718b3f6e53286925089f9f7cde3b748b404c5def/src/In1d.chpl#L99-L100 
where the biggest issue for multi-locale and especially CS is that `ar2` is a block dist array so these are fine-grained GETs. We can improve this by doing bulk communication to localize to a local array first.

The second problem is that we're adding to the set serially. The `mbound` case is always faster than `lbound` when ran serially, but as the array size increases parallelism outweighs the cheaper implementation. On single node 24-core chapcs, `2**21` is faster for `mBound`, and `2**22` is slower. The current threshold is `2**25` so we at least want to go down 8x, but probably even more so when distributed and we have to localize and when there's more cores making the paralllel `lbound` code faster. 

```
>>> arkouda int64 in1d
Medium regime: numLocales = 1  a.size = 10,000,000  b.size = 4,194,303
Medium average time = 0.4440 sec
Medium average rate = 0.24 GiB/sec
Large regime: numLocales = 1  a.size = 10,000,000  b.size = 4,194,305
Large average time = 0.3926 sec
Large average rate = 0.27 GiB/sec
```

I'm wondering if it's even worth optimizing based on array size or if we should always just use the `lbound` implementation. On single node you're talking about saving maybe 0.1 seconds in the best case, though it's higher at scale for multi-locale because the startup cost for sort is much higher. Even if it is faster at scale, do you think in1d with a small second array is a common enough to be worth maintaining the additional code?

ronawho:
And a reminder that sort does have a non-trivial startup cost based on the total number of cores. At 75K cores we saw ~20 seconds of startup time in some previous runs:

[...]

PR #1021, opened 2022-01-11
Drop in1d down to 1 trial

ronawho:
This benchmark is timing out under 16-node-cs-hdr testing (which we just
got running again.) Decrease the number of trials to hopefully avoid the
timeout while we work to tune performance in #970.


mhmerrill (approved):
looks ok to me.

PR #1023, opened 2022-01-12
Change Parquet C++ types from int to int64_t and reorganization

bmcdonald3:
In the initial Parquet work, I was naively using `int` for all of my Arrow code, even in places where `int64_t` made more sense. What this means is that, before this PR, the maximum size for a Parquet file that could be read/written was `2^32` elements large. This PR updates those places to be `int64_t` instead of `int` and also moves includes to the header file.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -1,12 +1,5 @@
 #include "ArrowFunctions.h"
 
-#include <iostream>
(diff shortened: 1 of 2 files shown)


mhmerrill:
@bmcdonald3 is this an urgent PR?

reuster986 (approved):

PR #1024, opened 2022-01-13
Add support for reading timestamps in Parquet files

bmcdonald3:
This PR adds checking for the `timestamp` Arrow type in a Parquet column and treats it as an `int64`, which is its underlying physical type. Also, the type checking for `int64` and `int32` are updated to check against the type enum rather than a type instance.

Resolves https://github.com/Bears-R-Us/arkouda/issues/1026

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -84,10 +84,12 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
   }
   auto myType = sc -> field(idx) -> type();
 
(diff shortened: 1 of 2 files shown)


mhmerrill:
@bmcdonald3 there is a conflict. Also what issue does this go against?

bmcdonald3:
Alright, I have cleaned up that merge conflict and linked against issue #1026

mhmerrill (approved):
this looks ok to me.

pierce314159 (approved):

PR #1028, opened 2022-01-14
Optimize Parquet file writing with WriteBatch function

bmcdonald3:
This PR changes from using the simple `WriteTable` function provided by the Parquet API to a lower-level `WriteBatch` function that allows greater control of writes. This optimization to Parquet writing in Arkodua is very similar to the reading optimization from https://github.com/Bears-R-Us/arkouda/pull/1014. 

Performance numbers collected on chapcs:
Current | This Branch
-- | --
0.11 GiB/s | 1.09 GiB/s

So we are seeing about a 10x improvement for writing Parquet files on chapcs.


reuster986 (approved):
I don't speak C++ very well, but I think I understand the logic. Great performance improvement!

ronawho (approved):

pierce314159 (approved):

Issue #670, opened 2021-02-12
How do we become compatible with the most recent NumPy 1.20?


mhmerrill:
How do we become compatible with the most recent NumPy 1.20?
What is the real issue to grapple with?

hokiegeek2:
@mhmerrill the version available at our customer sites as well as mypy compatibility. I can make the latter a priority at your direction.

mhmerrill:
this appears to be an incompatibility with `my-you

hokiegeek2:
@mhmerrill After upgrading to Numpy 1.20.1, the following mypy errors occur:

![image](https://user-images.githubusercontent.com/10785153/108243088-ae76a980-711b-11eb-8661-333b74aa429c.png)

Some of the errors may be associated with this [deprecation notice](https://numpy.org/doc/stable/release/1.20.0-notes.html):

![image](https://user-images.githubusercontent.com/10785153/108243436-17f6b800-711c-11eb-87d1-c174c74adee7.png)

More info regarding Numpy data types is [here](https://numpy.org/doc/stable/user/basics.types.html#basics-types) and dtypes is [here](https://numpy.org/doc/stable/reference/arrays.dtypes.html#arrays-dtypes)

glitch:
I posted & linked a PR for this.  The gist of it is that I was able to update to both 1.20.3, 1.21.5 (and test against older 1.19.2).  The latest numpy at this time is 1.22.0 but it drops support for python 3.7, which we still support, so I stopped at 1.21.5.

PR #1033, opened 2022-01-19
Update compatibility to numpy - 1.21.5

glitch:
Issue #670 - This updates the compatibility for numpy to a max of version 1.21.5.

`mypy` and `pytest` unit testing pass for the following numpy verisons (I tested locally as well as with the github CI runner, you can check the logs under `arkouda_tests_linux -> Build/Install Arkouda to verify which version of numpy was installed)

- 1.19.2 (https://github.com/glitch/arkouda/actions/runs/1719532839)
- 1.20.3 (https://github.com/glitch/arkouda/actions/runs/1718965418)
- 1.21.5 (https://github.com/glitch/arkouda/actions/runs/1719181717)

Note: The latest version of numpy at this time is 1.22.0 but it dropped Python 3.7.x support so I stopped with the latest 3.7.x line available in `conda` which was 1.21.5

In short the main thing which happened was deprecation warnings were added to things like `np.bool` in later versions on numpy.  The documentations said it was really just a stand in for `builtins.bool` etc.  I made updates for each of the specific items which were deprecated to use the recommended type.

Also updated was the version of `mypy` to 0.931 along with its dependency `typed-ast`.

You may need to update your local environment to the new versions of dependencies, but they should all be available with either `pip` or `conda`.

### DIFF
--- a/arkouda/dtypes.py
+++ b/arkouda/dtypes.py
@@ -19,22 +19,23 @@ NUMBER_FORMAT_STRINGS = {'bool': '{}',
                          'np.float64': 'f'}
 
 dtype = np.dtype
-bool = np.dtype(np.bool)
+bool = np.dtype(bool)
 int64 = np.dtype(np.int64)
 float64 = np.dtype(np.float64)
 uint8 = np.dtype(np.uint8)
 str_ = np.dtype(np.str_)
-npstr = np.dtype(np.str)
+npstr = np.dtype(str)
 
 # Union aliases used for static and runtime type checking
-bool_scalars = Union[builtins.bool,np.bool]
+bool_scalars = Union[builtins.bool, np.bool_]
 float_scalars = Union[float,np.float64]
 int_scalars = Union[int,np.int64]
-numeric_scalars = Union[float,np.float64,int,np.int64]
-numpy_scalars = Union[np.float64,np.int64,np.bool,np.uint8,np.str,np.str_]
-str_scalars = Union[str, np.str, np.str_]
+numeric_scalars = Union[float,np.float64,int,np.int64,np.uint8]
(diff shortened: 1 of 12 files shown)


pierce314159 (approved):

hokiegeek2 (approved):
looks great!

arkouda/dtypes.py
```diff
@@ -19,22 +19,23 @@
...
 
 # Union aliases used for static and runtime type checking
-bool_scalars = Union[builtins.bool,np.bool]
+bool_scalars = Union[builtins.bool, np.bool_]
 float_scalars = Union[float,np.float64]
 int_scalars = Union[int,np.int64]
-numeric_scalars = Union[float,np.float64,int,np.int64]
-numpy_scalars = Union[np.float64,np.int64,np.bool,np.uint8,np.str,np.str_]
-str_scalars = Union[str, np.str, np.str_]
+numeric_scalars = Union[float,np.float64,int,np.int64,np.uint8]
```
pierce314159, line 33: I just spoke with @glitch about this. This PR adds `np.uint8`  to `numeric_scalars` which seems right to me. But I wanted to double check with someone who's been around longer to make sure it's omission was an oversight and it wasn't intentionally excluded. Maybe @reuster986 or @hokiegeek2 can verify

hokiegeek2: @pierce314159 good question! IIRC, just an oversight

reuster986: yeah, I think it was just an oversight. Ideally, we would accept all numpy dtypes as scalar input (except for maybe uint64), because they're all castable to something arkouda supports, but that's a discussion for another PR :)

reuster986 (approved):

PR #1034, opened 2022-01-19
Refactor binary operator function with doBinOp function

bmcdonald3:
As a stepping stone to adding 2D array support, this PR refactors the `binopvvMsg()` function by creating a generic `doBinOp()` function that can accept additional types of symbol table entries beyond just the default `symEntry`. This allows easily stamping out types that are not desired and additionally allows optionally adding new entries, such as a 2D entry, to share this binary operations code, but still be only included optionally.

Additionally, the `binopvvMsg()` function is 453 lines today, but this PR brings that number down to 309 lines, and that is now split between 2 functions. 

This change may slightly increase the difficulty of understanding/reading the `binopvvMsg()` code due to the additional layer of abstraction, but I believe that this type of refactoring is needed to allow additional symbol table entry types to optionally share code with the existing Arkouda server code.

Resolves https://github.com/Bears-R-Us/arkouda/issues/1045

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -1,5 +1,5 @@
 
-module Operators
+module BinOp
 {
   use ServerConfig;
   
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -37,10 +37,10 @@ module BinOp
   :throws: `UndefinedSymbolError(name)`
   */
   proc doBinOpvv(l, r, e, op: string, rname, pn, st) throws {
-    // Since we know that the result type is a boolean, we know
-    // that it either (1) is an operation between bools or (2) uses
-    // one of the boolean operators from the `boolOps` set above
     if e.etype == bool {
+      // Since we know that the result type is a boolean, we know
+      // that it either (1) is an operation between bools or (2) uses
+      // a boolean operator (<, <=, etc.)
       if l.etype == bool && r.etype == bool {
         select op {
           when "|" {
--- a/src/OperatorMsg.chpl
+++ b/src/OperatorMsg.chpl
@@ -8,6 +8,7 @@ module OperatorMsg
     use BitOps;
     use Reflection;
     use ServerErrors;
+    use Operators;
 
     use MultiTypeSymbolTable;
     use MultiTypeSymEntry;
@@ -66,15 +67,15 @@ module OperatorMsg
             
             if boolOps.contains(op) {
               var e = st.addEntry(rname, l.size, bool);
-              return doBinOp(l, r, e, op, rname, pn, st);
+              return doBinOpvv(l, r, e, op, rname, pn, st);
             } else if op == "/" {
               // True division is the only case in this int, int case
               // that results in a `real` symbol table entry.
               var e = st.addEntry(rname, l.size, real);
-              return doBinOp(l, r, e, op, rname, pn, st);
+              return doBinOpvv(l, r, e, op, rname, pn, st);
             }
             var e = st.addEntry(rname, l.size, int);
-            return doBinOp(l, r, e, op, rname, pn, st);
+            return doBinOpvv(l, r, e, op, rname, pn, st);
           }
           when (DType.Int64, DType.Float64) {
             var l = toSymEntry(left,int);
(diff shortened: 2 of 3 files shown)


mhmerrill:
after some comment additions and maybe some refactoring `doBinOp` into it's own module i think we should be good to go.

bmcdonald3:
I moved `doBinOp` to a module called `Operators.chpl`, but I wasn't exactly sure what to name it. The regular Arkodua precedence seems to indicate naming it just `Operator.chpl` based on the `OperatorMsg.chpl` name, but that seemed a bit strange to me. Are you alright with calling it `Operators.chpl`?

mhmerrill:
> I moved `doBinOp` to a module called `Operators.chpl`, but I wasn't exactly sure what to name it. The regular Arkodua precedence seems to indicate naming it just `Operator.chpl` based on the `OperatorMsg.chpl` name, but that seemed a bit strange to me. Are you alright with calling it `Operators.chpl`?

I might be inclined to call it `DoBinOp.chpl` or maybe just `BinOp.chpl` because it is specific to the binary operators.

pierce314159 (approved):
Looks good to me!

src/BinOp.chpl
```diff
@@ -0,0 +1,255 @@
...
+          }
+          when "*" {
+            e.a = l.a * r.a;
+          }
+          when "//" { // floordiv
+            ref ea = e.a;
+            ref la = l.a;
+            ref ra = r.a;
+            [(ei,li,ri) in zip(ea,la,ra)] ei = if ri != 0 then li/ri else 0;
+          }
```
pierce314159, line 121: I accidentally didn't save one of my comments, but it's about existing code and not new changes so kinda outside the scope of the PR anyway

No change requested:
I was thinking it might be nice to add a comment about the `ref`s and `zip`s for the first of these

Also do we need `ref`s of `l.a` and `r.a`? My (limited) understanding is `e.a` is the only one that needs a `ref` since it's the only one being updated. Would something like below work just as well or am I missing something?
```C
when "//" { // floordiv
  ref ea = e.a;
  [(ei,li,ri) in zip(ea,l.a,r.a)] ei = if ri != 0 then li/ri else 0;
}
```

glitch (approved):
2 things

1. I agree with Pierce on the comment thing & on the ref thing, but the code looks correct to the extend of my knowledge.
2. I was going to say we should probably separate the message tuple creation from the `e.a=...` stuff so you wouldn't need to pass the `rname`, nor `symtab` and leave it to be done in the invoking function (and throw an Error in the otherwise blocks, catching in the calling function).  However, it's not like it makes it shorter because you have to pay that cost somewhere.  So what would it gain you over not passing the SymTab and name around... not much.  So at that point it's mainly stylistic and there's not really a strong reason to change it.

I think you should probably fix the comment, but otherwise :+1:  Approved.

src/BinOp.chpl
mhmerrill: @bmcdonald3 @pierce314159 @bradcray the ref is required here because the array is wrapped in an object and the object is stored on locale-0. the refs are not required when you do a direct whole array operations like `e.a = l.a + r.a`, i think this is because the compiler can infer that the object references are constant through the whole operation.

mhmerrill (approved):
looks ok with the `setOps` comment fix mentioned

src/BinOp.chpl
```diff
@@ -0,0 +1,255 @@
...
+  :arg st: SymTab to act on
+  :type st: borrowed SymTab 
+
+  :returns: (MsgTuple) 
+  :throws: `UndefinedSymbolError(name)`
+  */
+  proc doBinOpvv(l, r, e, op: string, rname, pn, st) throws {
+    // Since we know that the result type is a boolean, we know
+    // that it either (1) is an operation between bools or (2) uses
+    // one of the boolean operators from the `boolOps` set above
```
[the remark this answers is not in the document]
mhmerrill: I agree, make sure this comment gets moved back to the call site.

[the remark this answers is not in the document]
bmcdonald3: I think the comment belongs here, but I have updated to be accurate after moving this to the new file.

src/BinOp.chpl
bradcray: @pierce314159 : For reference, this is something that's required for performance, not correctness, and which we hope to have the Chapel compiler automatically optimize in the future—so view it as an unfortunate workaround rather than anything we're particularly proud of or that should obviously make sense.  The following issue in our repo tracks it:  https://github.com/chapel-lang/chapel/issues/10160

PR #1036, opened 2022-01-20
fixes float precision bug #964

### LINKED ISSUE #964: Warn about float precision on `GroupBy.aggregate(float64)`
Because arkouda implements grouped aggregations as differences on scans, these operations can amplify floating point precision errors beyond what is typical in numpy/pandas. We should consider raising a warning when the user calls an aggregation on a `float64` pdarray. The warning message should be something like, "Grouped aggregations on float64 can magnify errors due to floating point imprecision."

We might also consider adding a section to the documentation on this issue, and linking to that page in the warning message.

Example:
```python
N = 10**8
G = 10**7
groupnum = ak.randint(0, G, N, seed=1)
intval = ak.randint(0, 2**32, N, seed=2)
floatval = ak.cast(intval, ak.float64)
g = ak.GroupBy(groupnum)
_, intmean = g.mean(intval) # Exact mean
_, floatmean = g.mean(floatval) # Approximate; WARN here
ak_mse = ak.mean((intmean - floatmean)**2)
print(ak_mse)
1.1131222...

df = pd.DataFrame()
df['groupnum'] = groupnum.to_ndarray()
df['intval'] = intval.to_ndarray()
df['floatval'] = floatval.to_ndarray()
gdf = df.groupby('groupnum').agg({'intval': 'mean', 'floatval': 'mean'})
pd_mse = ((gdf.intval - gdf.floatval)**2).mean()
print(pd_mse)
0.0
```

In this example, the mean squared error in arkouda between int and float results is on the order of 1, whereas the mse in pandas/numpy is 0 (below machine tolerance).

reuster986:
I had a stroke of inspiration that I could fix #964 by adapting the `ResettingOrScanOp` I created for doing grouped bitwise OR and AND. Using the same approach for summation, the cumulative sum gets reset to zero at every segment boundary, stopping any precision errors from carrying over. I added a unit test to verify that grouped summation gives equivalent results for int and float.

The big question is, what does it do to performance? Currently, on 16 node CS, grouped sum runs at 93 GiB/s, whereas OR and AND are about 63 GiB/s. If the difference is due entirely to the scan class, then we might expect sum performance to drop about 30%, to the level of AND and OR. This would be regrettable, but I'd say the precision is worth it.

### DIFF
--- a/src/ReductionMsg.chpl
+++ b/src/ReductionMsg.chpl
@@ -411,35 +411,127 @@ module ReductionMsg
        and then reduce over each chunk using the operator <Op>. The return array 
        of reduced values is the same size as <segments>.
      */
-    proc segSum(values:[] ?intype, segments:[?D] int, skipNan=false) throws {
+    /* proc segSum(values:[] ?intype, segments:[?D] int, skipNan=false) throws { */
+    /*   type t = if intype == bool then int else intype; */
+    /*   var res: [D] t; */
+    /*   if (D.size == 0) { return res; } */
+    /*   var cumsum; */
+    /*   if (isFloatType(t) && skipNan) { */
+    /*     var arrCopy = [elem in values] if isnan(elem) then 0.0 else elem; */
+    /*     // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit */
+    /*     overMemLimit(numBytes(t) * arrCopy.size); */
+    /*     cumsum = + scan arrCopy; */
+    /*   } */
+    /*   else { */
+    /*     // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit */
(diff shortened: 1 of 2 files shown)


mhmerrill:
@reuster986 @ronawho I think I am ok with the performance reduction in light of the accuracy gain.

mhmerrill (approved):
I think this looks good, understanding the accuracy gain at the loss of performance.

pierce314159 (approved):
I def understand `ReduceScanOp` classes better after reviewing this!

This looks good to me!

Aside:
This took me longer than it should've mostly because I misunderstood what `segSum` was supposed to be doing in the first place so I was over here like "I think I understand it but idk how it's getting the right answer?" until i looked at the original... :man_facepalming: i got there eventually... :sweat_smile:

PR #1037, opened 2022-01-20
Reduce argsort memory footprint

ronawho:
Reduce the memory overhead of argsort/radixSortLSD_ranks from ~5x to
~4x. The big arrays created are 2 temp arrays with the key and the rank
and a return array with just the rank. Previously all 3 arrays existed
at the same time for a ~5x overhead, but this refactors the code to
destroy one of the temp arrays before creating the return array, which
changes the peak overhead to those 2 temp arrays or a ~4x overhead.

While here remove some internal `overMemLimit` checks on a scan since
it's expected that callsites use radixSortLSD_memEst before calling
argsort and that includes the scan arrays.

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -154,103 +154,103 @@ module RadixSortLSD
         try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                        "type = %s nBits = %t".format(t:string,nBits));
         
-        // form (key,rank) vector
-        var kr0: [aD] (t,int) = [(key,rank) in zip(a,aD)] (key,rank);
         var kr1: [aD] (t,int);
-        
-        // create a global count array to scan
-        var gD = newBlockDom({0..#(numLocales * numTasks * numBuckets)});
-        var globalCounts: [gD] int;
-        var globalStarts: [gD] int;
-        
-        // loop over digits
-        for rshift in {0..#nBits by bitsPerDigit} {
-            const last = (rshift + bitsPerDigit) >= nBits;
-            try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                                        "rshift = %t".format(rshift));
-            // count digits
-            coforall loc in Locales {
-                on loc {
-                    coforall task in Tasks {
-                        // bucket domain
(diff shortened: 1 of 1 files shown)


ronawho:
> While here remove some internal overMemLimit checks on a scan since
it's expected that callsites use radixSortLSD_memEst before calling
argsort and that includes the scan arrays.

Opened as a draft because that isn't actually true. I found that many callsites are computing their own memory estimate instead of using this common one. I was going to change those, but then I wondered if it just makes sense to move the mem limit check from the callsite into the top of the radixSort routines. 

@mhmerrill Curious for your thoughts on that and this approach to lowering the memory overhead of `radixSortLSD_ranks`. Note that it's easiest to look at ignoring whitespace -- https://github.com/Bears-R-Us/arkouda/pull/1037/files?w=1

mhmerrill:
@ronawho I have several thoughts about this...

- this is a good catch and a nice reduction on temp space usage
- I added `radixSortLSD_memEst()` intending to use it all over the code where ever the radix sort is called but didn't follow through on that intention.
- I never intended `radixSortLSD_ranks()` to `throw`, all/most of the `throw`ing should be done before we get into the bare-chapel-array level of the code.
- please add an issue to link this PR.

ronawho:
> I added radixSortLSD_memEst() intending to use it all over the code where ever the radix sort is called but didn't follow through on that intention.
> I never intended radixSortLSD_ranks() to throw, all/most of the throwing should be done before we get into the bare-chapel-array level of the code.

Ok, so you would be ok if I made these non-throwing, and updated all call sites to use `radixSortLSD_memEst`?

mhmerrill:
> > I added radixSortLSD_memEst() intending to use it all over the code where ever the radix sort is called but didn't follow through on that intention.
> > I never intended radixSortLSD_ranks() to throw, all/most of the throwing should be done before we get into the bare-chapel-array level of the code.
> 
> Ok, so you would be ok if I made these non-throwing, and updated all call sites to use `radixSortLSD_memEst`?

YES!

reuster986 (approved):

ronawho:
Closing in favor of https://github.com/Bears-R-Us/arkouda/pull/1061.

PR #650, opened 2021-02-03
cleaned up verbosity message info #649

hokiegeek2:

PR #1041, opened 2022-01-21
Add support for passing server args to runClient/run_benchmarks.py

### LINKED ISSUE #1006: Update `runClient` with additional server args
Currently, the `util/test/runClient` script only takes `-nl` as an argument, but it would be great to be able to specify things like `--partition` and `--walltime`. It should just be able to pass those arguments to `arkouda_server`.

ronawho:
Add `--server-args` argument to `runClient`/`run_benchmarks.py`. This allows
passing arbitrary options to the server. While here, unify some of the options
for these scripts (support both version of num locales, add port to runClient)
and  change `start_arkouda_server` `log`/`verbose` args to `trace`. The actual
server options were replaced in #650/#760, so just update the start server
wrapper args to reflect the name change.

Example usage:

```
./benchmarks/run_benchmarks.py argsort --server-args=--CommAggregation.dstBuffSize=256 --server-args=--CommAggregation.srcBuffSize=12
./util/test/runClient tests/check.py --server-args=--walltime=5:00
```

Resolves #1006

### DIFF
--- a/benchmarks/run_benchmarks.py
+++ b/benchmarks/run_benchmarks.py
@@ -91,8 +91,9 @@ def create_parser():
     # TODO support alias for a larger default N
     #parser.add_argument('--large', default=False, action='store_true', help='Run a larger problem size')
 
-    parser.add_argument('-nl', '--num-locales', default=get_arkouda_numlocales(), help='Number of locales to use for the server')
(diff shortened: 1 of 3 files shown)


glitch (approved):

mhmerrill (approved):
looks ok to me also

Issue #1046, opened 2022-01-21
Allow modules from outside the Arkouda src directory to be modularly built


bmcdonald3:
Today, the modular Arkouda build only works for modules in the Arkouda `src/` directory. Adding support for including modules from other directories is desirable.

The main challenges with this approach is adding client-side support for the new functions. One approach would be to have a generic function that could call any server function, but then users would be limited in return types, so something like this:
```
def server_function(cmd, args):
    repMsg = generic_msg(cmd=cmd, args=args)
    return repMsg
```

Another approach is expecting users to create their own client-side functions and require them to be imported into the Arkouda instance:
```
import arkouda as ak
from arkouda.client import generic_msg

def test_command():
    repMsg = generic_msg(cmd='test-command')
    return repMsg

ak.__dict__["test_command"] = test_command
```
This approach seems the most reasonable, since it allows the user full control over the arguments, any preprocessing of those arguments, and how the return type is handled, but it is definitely more work than the `server_function()` approach.

I have been considering additional solutions to somehow automate this process, but I just don't see how that could be done since there are any number of things that a user would want to do on the client side to handle their server call.

glitch:
When trying to use this locally (setting `ARKOUDA_SERVER_USER_MODULES=/home/glitch/projects/arkouda-contrib` etc.) I get an error when issuing a `make`.  The full command and error look like this:
```
chpl --print-passes    --ccflags="-DH5_USE_110_API" --ccflags="-O1" -smemTrack=true -smemThreshold=1048576 --no-cache-remote --interleave-memory -I/home/glitch/miniconda3/envs/ak-base/include -L/home/glitch/miniconda3/envs/ak-base/lib --ldflags="-Wl,-rpath,/home/glitch/miniconda3/envs/ak-base/lib" -lhdf5 -lhdf5_hl -lzmq -sarkoudaVersion="\"pre-release-v2022.01.20+32.g28742e5a.dirty\"" /home/glitch/projects/arkouda//src/arkouda_server.chpl -M /home/glitch/projects/arkouda//src/compat/ge-125 /home/glitch/projects/arkouda-contrib -o arkouda_server
error: file '/home/glitch/projects/arkouda-contrib' does not have a recognized suffix
make: *** [Makefile:270: arkouda_server] Error 1
```

[...]

PR #1047, opened 2022-01-21
Allow modules from outside the Arkouda source dir to be built into the server

bmcdonald3:
Today, the modular Arkouda build only works for modules in the Arkouda `src/` directory, but this PR enables users to specify modules from any directory by supplying the absolute path in an environment variable called `ARKOUDA_SERVER_USER_MODULES`.

This means a user could add `TestMsg` module name to the `ServerModules.cfg` file and then build Arkouda with the command `ARKOUDA_SERVER_USER_MODULES=/path/to/TestMsg.chpl make` and that module would be included in the build. The path in the environment variable is appended onto the `chpl` command so that the compiler can find the file.

Resolves: https://github.com/Bears-R-Us/arkouda/issues/1046


src/serverModuleGen.py
```diff
@@ -22,6 +25,21 @@ def generateServerIncludes(config_filename, reg_filename):
 
     serverfile.write("}\n")
 
+def generateSearchPaths(filename):
+    with open(filename) as configfile:
+        modules = configfile.readlines()
+        ret = ""
+        for module in modules:
+            module = (module.rstrip()).lstrip()
```
glitch, line 33: I think you can simplify this as `module = module.strip()`  which should trim whitespace from both ends.

src/serverModuleGen.py
```diff
@@ -22,6 +25,21 @@ def generateServerIncludes(config_filename, reg_filename):
...
+            # Represents a path to a file
+            if len(module) > 0 and module[0] == '/':
```
glitch, line 35: You can simplify this as `if module.startswith("/")`

bmcdonald3: Thanks for the suggestions, both are better for sure!

glitch (approved):

MODULAR.md
```diff
@@ -12,9 +12,7 @@ New config files must follow the same format, listing one module name per line w
 
 ### Adding new modules into the build process
 
-Adding a custom module into the Arkouda buld process is treated the same way as a regular module: added in a single line in the server configuration file.
-
-Currently, modules will only be found by the Arkouda build if located in the `src/` directory.
+The only unique aspect of adding modules from other directories into the Arkouda build process is that a complete path must be specified for all modules not found in the Arkouda `src/` directory. This would actually be done by adding a line such as `/Users/test-user/path/to/mod` to `ServerModules.cfg` to add the module `mod.chpl` to the Arkouda build. Aside from that, adding a custom module into the Arkouda buld process is treated the same way as a regular module: added in a single line in the server configuration file.
```
pierce314159, line 15: very small note:
I think it's clearer here to say "This is done by" instead of "This would actually be done by"

pierce314159 (approved):
Looks good to me!

MODULAR.md
bmcdonald3: Makes sense, thanks for pointing that out; added.

mhmerrill:
IMO this is a step forward but I'm a little taken back by the need to spec absolute paths. Maybe we could also spec with an env var or something like `$OTHER_PKG/PkgMsg.chpl` or something? make sense? I am not sure I want my local directory structure exposed if I gave the module list to someone else...

bmcdonald3:
Ya, that makes sense to me, I think having an environment variable to specify the direct paths is a better idea, I can work on switching over to that.

glitch:
A minor request to have the env var prefixed with `ARKOUDA_`, so it limits any potential clash with a users environment (and makes it easy to `env|grep ARK` and dump out all of the arkouda variables.

mhmerrill:
> A minor request to have the env var prefixed with `ARKOUDA_`, so it limits any potential clash with a users environment (and makes it easy to `env|grep ARK` and dump out all of the arkouda variables.

is that something we should enforce? or a usage convention?

glitch:
I'm realizing we could really leave it in the user's hands.  In theory they can use whatever env var or directory they want right?

mhmerrill:
> I'm realizing we could really leave it in the user's hands. In theory they can use whatever env var or directory they want right?

i think so... maybe they have a chapel package managed by mason and it might also have an Arkouda extension...

pierce314159 (approved):

MODULAR.md
```diff
@@ -12,9 +12,7 @@ New config files must follow the same format, listing one module name per line w
 
 ### Adding new modules into the build process
 
-Adding a custom module into the Arkouda buld process is treated the same way as a regular module: added in a single line in the server configuration file.
-
-Currently, modules will only be found by the Arkouda build if located in the `src/` directory.
+The only unique aspect of adding modules from other directories into the Arkouda build process is that a complete path must be specified for all modules not found in the Arkouda `src/` directory. This is done by adding a line such as `/Users/test-user/path/to/mod` to `ServerModules.cfg` to add the module `mod.chpl` to the Arkouda build. Aside from that, adding a custom module into the Arkouda buld process is treated the same way as a regular module: added in a single line in the server configuration file.
```
pierce314159, line 15: small typo `buld process` -> `build process`

pierce314159 (approved):

MODULAR.md
bmcdonald3: Good catch; updated.

bmcdonald3:
Alright, updated this PR to use an environment variable named `ARKOUDA_SERVER_USER_MODULES` rather than requiring a full path in the `ServerModules.cfg` file.

src/serverModuleGen.py
```diff
@@ -5,9 +5,12 @@ def getModules(filename):
...
         ret = []
         for module in modules:
-            module = (module.rstrip()).lstrip()
+            module = module.strip()
             if len(module) > 0 and module[0] != '#':
-                ret.append(module)
+                if module[0] == '/':
+                    ret.append(module[module.rfind('/')+1:])
+                else:
+                    ret.append(module)
```
glitch, line 13: This works so no change requested, but if you wanted to tighten it up a bit
```python
    for module in modules:
        module = module.split("#")[0].split("/")[-1].strip()
        if module:
            ret.append(module)
```
This also handles trailing comments i.e. `MyMsg  # This is my fancy module` -> `MyMsg`

If you want to go full blown crazy the whole func in 3 lines:
```python
def getModules(filename):
    with open(filename) as configfile:
        return [m.split("/")[-1].strip() for m in configfile.readlines() if m.split("#")[0].strip()]
```
but that is a bit unreadable.

glitch (approved):

src/serverModuleGen.py
bmcdonald3: Ooh, I like the fancy Python syntax, switched to the first option, thanks!

PR #1048, opened 2022-01-21
Add optional support for 2 dimensional arrays

bmcdonald3:
This PR adds support for optional 2D arrays in the server. These 2D arrays essentially only support binary operations and random 2D array generation in this PR, but that could be extended to other areas of the codebase as seen fit.

This is able to stay modular by containing everything pertaining to 2D arrays in files separate from the 1D array code. This 2D array support is located in `SymEntry2D.chpl`, which contains the 2D symbol table entry, and `Arr2DMsg.chpl`, which contains all of the 2D array functions.

This PR is based on the binary operations refactor in https://github.com/Bears-R-Us/arkouda/pull/1034 (in other words, the diff for `OperatorMsg.chpl` is identical to that other PR), where the 2D array code is able to share the `doBinOp` function to avoid code duplication while maintaining modularity of the 2D arrays.

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -205,6 +205,8 @@ class pdarray:
             if self.size != other.size:
                 raise ValueError("size mismatch {} {}".format(self.size,other.size))
             cmd = "binopvv"
+            if self.ndim != 1:
+                cmd += "2d"
             args= "{} {} {}".format(op, self.name, other.name)
             repMsg = generic_msg(cmd=cmd,args=args)
             return create_pdarray(repMsg)
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -457,6 +457,9 @@ class pdarray:
                 # Interpret negative key as offset from end of array
                 key += self.size
             if (key >= 0 and key < self.size):
+                if self.ndim == 2:
+                    repMsg = generic_msg(cmd="[int2d]", args=f"{self.name} {key}")
+                    return create_pdarray(repMsg)
                 repMsg = generic_msg(cmd="[int]", args="{} {}".format(self.name, key))
                 fields = repMsg.split()
                 # value = fields[2]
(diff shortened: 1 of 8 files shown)


arkouda/pdarrayclass.py
```diff
@@ -205,6 +205,8 @@ def _binop(self, other : pdarray, op : str) -> pdarray:
             if self.size != other.size:
                 raise ValueError("size mismatch {} {}".format(self.size,other.size))
             cmd = "binopvv"
+            if self.ndim != 1:
+                cmd += "2d"
```
Ethan-DeBandi99, line 209: I am wondering if it would be better to check specifically for a dimension of 2 here. This shouldn't be an issue now, since we are only expecting dimensions of 1 or 2, but could help identify anything out of the oridinary with the dimension in the future.

src/Arr2DMsg.chpl
```diff
@@ -0,0 +1,280 @@
...
+      n = nStr: int;
+    } catch {
+      var errorMsg = "Error parsing/decoding either dtypeBytes, m, or n";
+      gsLogger.error(getModuleName(), getRoutineName(), getLineNumber(), errorMsg);
+      return new MsgTuple(errorMsg, MsgType.ERROR);
+    }
+
+    overMemLimit(2*m*n);
+
+    if dtype == DType.Int64 {
```
Ethan-DeBandi99, line 39: Is uint64 support needed for 2d arrays?

src/Arr2DMsg.chpl
```diff
@@ -0,0 +1,280 @@
...
+    var repMsg: string; // response message
+    // split request into fields
+    var (dtypeStr,aMinStr,aMaxStr,mStr,nStr,seed) = payload.splitMsgToTuple(6);
+    var dtype = str2dtype(dtypeStr);
+    var m = mStr:int;
+    var n = nStr:int;
+    var rname = st.nextName();
+    
+    select (dtype) {
+      when (DType.Int64) {
```
Ethan-DeBandi99, line 82: If uint64 support is required, it would need to be added here as well.

Ethan-DeBandi99:
I was not sure if uint64 support for 2d arrays is outside this scope of this pr, but noted it above. There are other locations where it would be needed as well. If support is outside the scope of this work, please disregard.

bmcdonald3:
Closing this since we are hoping to merge this into the `arkouda-contrib` repo instead: https://github.com/Bears-R-Us/arkouda-contrib/pull/1

PR #1050, opened 2022-01-26
Parallelize Parquet file reading on-node

bmcdonald3:
This PR turns the `for` loop for reading files into a `forall` loop to parallelize file reading. Since Parquet is thread-safe, we are able to parallelize reads on-node, which can't be done with HDF5. `test_multi_file()` in `tests/parquet_test.py` tests this functionality.

Here are some performance figures collected on 16-node-cs-hdr using 400 .25 GiB files before and after vs HDF5:
| Nodes | HDF5   | Ser Parquet (before) | Par parquet (after) |
| ----: | -----: | ----------: | ----------: |
|  1      |  136.1s   |    126.9s  |        12.8s   |
| 16     |  8.4s  |    17.8s        |         4.0s    |

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -93,13 +93,13 @@ module ParquetMsg {
     var (subdoms, length) = getSubdomains(sizes);
 
     coforall loc in A.targetLocales() do on loc {
-      var pqErr = new parquetErrorMsg();
       var locFiles = filenames;
       var locFiledoms = subdoms;
-      for (filedom, filename) in zip(locFiledoms, locFiles) {
+      forall (filedom, filename) in zip(locFiledoms, locFiles) {
         for locdom in A.localSubdomains() {
           const intersection = domain_intersection(locdom, filedom);
           if intersection.size > 0 {
+            var pqErr = new parquetErrorMsg();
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Wow! What a speedup!! Looks good to me

bmcdonald3:
On doing some more testing, I discovered an off by one error when reading more than 20 files, will need to explore that more before merging.

bmcdonald3:
This was actually an issue that I had already encountered in the Chapel code; in the unit test I was appending the number straight to the end of the file name, which caused "file10" to come before "file1", causing the read ins to come back in a weird order, so it was a bug with the test, not the parallelization code.

glitch (approved):

mhmerrill (approved):
looks ok to me

ronawho:
@reuster986 FYI in case you haven't seen pings in other places.

reuster986:
@ronawho @bmcdonald3 This is awesome! And yes I did miss it -- I have like 4 different jobs these days, so thank you for pinging again!

@mhmerrill @glitch once #1055 gets merged (since it's a correctness fix), I think we should cut another release to test parquet read performance internally.

PR #1054, opened 2022-02-01
Issue 1046: Fixes issue with missing -M for user contrib server modules in external directory

glitch:
Issue 1046: Fixes issue with missing -M for user contrib server modules in external directory

Small fix to test for and add `-M` if missing to `ARKOUDA_SERVER_USER_MODULES`


bmcdonald3:
See comments on other issue; there was a design discussion for not using `-M` and instead using the complete path to the module.

Makefile
```diff
@@ -262,13 +262,19 @@ else
 	ARKOUDA_COMPAT_MODULES += -M $(ARKOUDA_SOURCE_DIR)/compat/lt-125
 endif
 
+ifdef ARKOUDA_SERVER_USER_MODULES
+	ifeq (,$(findstring \-M,$(ARKOUDA_SERVER_USER_MODULES)))
```
mhmerrill, line 266: does this grock a list of directories contained in the `ARKOUDA_SERVER_USER_MODULES` env var?

mhmerrill: so we can include modules from multiple directories.

glitch:
I'm going to decline/close this one.  I think I understand @bmcdonald3 better now.  I mistook to the file path versus module name  and module directory path (I did the latter, set full path to the module directory and specified the module name in the cfg file, but I should have specified full path to the file containing the module.)  I'll add a wiki entry on this with an example to make it a bit more clear.

PR #1055, opened 2022-02-01
#964 Fix sum precision for real this time

reuster986:
History:
- #964 identified the issue with poor float precision in grouped sums
- #1036 attempted to fix the precision by applying the resetting scan operator developed for grouped AND/OR to grouped sum
- #1043 reverted the fix after a bug was discovered that produced incorrect results at high core counts (which is why it was not caught by the CI)

I first wrote a reproducer, `test/ScanOpTest.chpl`, which reproduces the bug when compiled against a previous version of arkouda in multi-locale mode and run with oversubscription, e.g.:

`CHPL_RT_OVERSUBSCRIBED=yes CHPL_RT_NUM_THREADS_PER_LOCALE=8 test-bin/ScanOpTest -nl 4 --SIZE=32`

This helped me determine (via much trial and error), that the bug was a fairly simple mistake in the `ResettingPlusScanOp.accumulateOntoState()` method in how it handled (or failed to handle) the reset history of the earlier state being accumulated onto the current state. The fix itself is most easily seen [here](https://github.com/Bears-R-Us/arkouda/commit/5837347245eebfb7fc3ba0f888f5e891814be3f2#diff-4eccb88989269fe625d4e7e7d089f6e26ee8210fe5801fabb112347a3f06de55).

Because this same bug should also affect grouped AND/OR, I applied the fix to those scan operators as well.

Even though the CI still won't create the conditons that triggered the original bug, I left the chapel test in place, in case we need to test manually in the future.

Finally, there is a python test for verifying that the precision of grouped sum is correct, since that was the original purpose of this whole expedition.


reuster986:
@ronawho would you mind compiling `test/ScanOpTest.chpl` and running under the configuration where you started seeing this bug? I'm fairly confident that I found a configuration on my laptop that covers it, but I just want to be sure we don't break the nightly runs again.

ronawho:
> @ronawho would you mind compiling test/ScanOpTest.chpl and running under the configuration where you started seeing this bug? I'm fairly confident that I found a configuration on my laptop that covers it, but I just want to be sure we don't break the nightly runs again.

The following passes for me on 16-node-cs-hdr:

```sh
make test-bin/ScanOpTest
CHPL_RT_NUM_THREADS_PER_LOCALE=128 ./test-bin/ScanOpTest -nl 16
```

I also checked that the full pytest suite passes:

```
ARKOUDA_NUMLOCALES=16 make test
...
====================================================================== 276 passed in 903.27s (0:15:03) =======================================================================
```

reuster986:
@ronawho thanks!

pierce314159 (approved):

bmcdonald3 (approved):

Issue #1059, opened 2022-02-01
Operations that localize string segments run slow with poorly distributed data


reuster986:
Regex operations and string hashing both use `lowLevelLocalizingSlice` to pull each string's bytes onto the locale where its segment pointer resides, in order to compute a function locally over those bytes. This approach works fantastically when most strings' bytes live on the same locale as their segment pointers, which is true for most arrays of strings. However, users have run into several degenerate data distributions where the majority of bytes are on different locales than their corresponding segment pointers. In these cases, the current approach is very slow because it's doing individual GETs for most of the string bytes.

One of these degenerate cases that crops up somewhat frequently is when you have an array of strings that are sorted by length. In this case, it is virtually guaranteed that a majority of bytes will live on a different locale than their segment pointers. In one such instance, a large array like this took about 20 minutes just to hash, but when we randomly permuted the same array (thus breaking the degeneracy), hashing took only a few seconds. A couple other users have mentioned running into this same type of problem.

In light of these examples, I'm about to put in a draft PR that demonstrates a different approach, in which the computation of the function (e.g. sipHash) on a given string happens on the locale where the bytes start, rather than the locale where the segment pointer resides.

We can also use this issue to propose other approaches and discuss their merits.

ronawho:
👍  Very interested in seeing the draft PR and good to know about this issue. We discussed this a bit in https://github.com/Bears-R-Us/arkouda/issues/266, but at the time I think we decided to wait for this to be an issue in user runs and sounds like we have that now.

Like you said, today we distribute work based on the segments array and the localize any remote data. Sounds like you want to distribute work based on the data and then maybe we have a PUT the results back since the result array has the same distribution as the segments. ​This should definitely be faster though I imagine there will be some difficulties to localize strings that cross locale boundaries. I also don't expect performance to be perfectly uniform in this case because for instance the rate of SipHash is faster for larger strings so the locale that owns the large strings will likely finish faster. This should still be way better than what we have today, but is something to be aware of.

>  In these cases, the current approach is very slow because it's doing individual GETs for most of the string bytes.

Nit, but `lowLevelLocalizingSlice` should do a single GET for all bytes in a string and not a GET per byte. This will still be slow, but it's not completely fine-grained. I'm not sure that you said otherwise, but just wanted to clarify in case anybody reading at home assumed it was element at a time GETs.

reuster986:
[...]

PR #1060, opened 2022-02-01
Draft: #1059 String ops follow byte locality

reuster986:
This PR is a demonstration for an approach to solving #1059 by running per-string computations on the locale where the string's bytes start, rather than the locale where its segment pointer lives. This PR currently only does this for sipHash128, but I believe it could be extended to regex operations.

In the case of an array of strings sorted by length (described in the linked issue), the existing code would have to remotely GET the bytes of most strings using `lowLeveLocalizingSlice`. However, the approach in this PR would only have to remotely GET the bytes of at most `numLocales - 1` strings. I think the only way to incur a lot of communication would be to have an array dominated by one or a few very long strings -- certainly possible, but more of an edge case than what is currently causing problems.

I plan to measure performance on our internal systems later this week, but if others want to run tests, I would welcome it.

Mainly, I'm looking for feedback from @pierce314159 and @ronawho about the design of this code:
- Do you think this code could be easily extended to other operations like regex?
- I had originally tried to pass a function pointer to `computeOnSegments`, but I couldn't get first-class functions to work right, so I went with the enum. Any other suggestions?
- There are two places in `ComputeOnSegments` (lines 68 and 71) where I use a range index to gather (potentially) non-local values, and I'm wondering if those should use aggregation instead.


ronawho:
I see a rather significant performance degradation for this with the SipHashSpeedTest benchmark (tested on 16-node-cs-hdr with the following patch):


<details>

```diff
diff --git a/src/SegmentedComputation.chpl b/src/SegmentedComputation.chpl
index 3938255..9b9929e 100644
--- a/src/SegmentedComputation.chpl
+++ b/src/SegmentedComputation.chpl
@@ -1,4 +1,6 @@
 module SegmentedComputation {
+  use CommAggregation;
+  use SipHash;

   proc computeSegmentOwnership(segments: [?D] int, vD) {
     // Locale that owns each segment's bytes
diff --git a/test/SipHashSpeedTest.chpl b/test/SipHashSpeedTest.chpl
index cbfbab0..a3de8c1 100644
--- a/test/SipHashSpeedTest.chpl
+++ b/test/SipHashSpeedTest.chpl
@@ -6,6 +6,8 @@ config const NINPUTS = 100_000;
 config const INPUTSIZE = 64;
 config const SEED = "none";

+config const computeOnSegments = false;
+
 enum testMode {fixed, variable};
 config const mode = testMode.variable;
 config const compareTypes = false;
@@ -39,8 +41,13 @@ proc testVariableLength(n:int, meanSize:int, type t) {
   }
   var hashes: [segs.domain] 2*uint;
   d.start();
-  forall (h, i, l) in zip(hashes, segs, lengths) {
-    h = sipHash128(tohash, i..#l);
+  if computeOnSegments {
+    use SegmentedComputation;
+    hashes = computeOnSegments(segs, vals, SegFunction.SipHash128, 2*uint(64));
+  } else {
+    forall (h, i, l) in zip(hashes, segs, lengths) {
+      h = sipHash128(tohash, i..#l);
+    }
   }
   d.stop(printTime=false);
   return (d.elapsed(), vals.size * numBytes(t));
```

</details>


```
./test-bin/SipHashSpeedTest -nl 16 --NINPUTS=10_000_000_001 --INPUTSIZE=64 --mode=testMode.variable
Hashed 10000000001 uint(8) blocks (615296.0 MB) in 2.54 seconds (241974.54 MB/s)

./test-bin/SipHashSpeedTest -nl 16 --NINPUTS=10_000_000_001 --INPUTSIZE=64 --mode=testMode.variable --computeOnSegments
Hashed 10000000001 uint(8) blocks (615259.0 MB) in 60.38 seconds (10189.69 MB/s)
```

Looking at comm counts I see:

```
./test-bin/SipHashSpeedTest -nl 16 --NINPUTS=10_000_000_001 --INPUTSIZE=64 --mode=testMode.variable --printDiagsSum
(GETS = 2563030, PUTS = 0, ONS = 30)

./test-bin/SipHashSpeedTest -nl 16 --NINPUTS=10_000_000_001 --INPUTSIZE=64 --mode=testMode.variable --computeOnSegments --printDiagsSum
(GETS = 6591724, PUTS = 1154, ONS = 9360)
```

So 2.5x more GETs and a lot more ONs. I can narrow down where the comm is coming from if that's helpful.

reuster986:
Thanks for running the benchmark!

Hmm, that's disappointing, and somewhat counterintuitive -- I can't think where that much communication is coming from, unless I made a mistake in calculating which locale owns each string's bytes.

Anyway, yes I'd love to narrow down where the comms are coming from...

reuster986:
I think I might have realized at least one thing that was going on. If you look at [this commit](https://github.com/Bears-R-Us/arkouda/pull/1060/commits/65d002918ffb12b570f1e82fa79629d9fba4d862) that I just pushed, I had previously been using a range to slice a BlockDist array, which I think results in another BlockDist array. I just changed the slice to a domain, so that the resulting array will be local, which is what I had originally intended. The arrays need to be local because they are driving the `forall`, which should run on all local tasks, not on all locales.

It's hard to really get a sense for the scaling of comms on my laptop, but this change does seem to reduce the number of GETs and ONs.

ronawho:
Ah, yeah -- that was where a lot of comm was coming from. 16-node-cs-hdr machine is down at the moment, but that reduced a lot of comm on a different machine. Creating the BlockDist array created a lot more communication, but it also create tasks on all remote locales to eagerly privatize and some of those tasks were blocked from running by other tasks in our SPMD section so beyond the extra comm some tasks were stalled.

Will regather full numbers when 16-node-cs-hdr is online.

reuster986:
I used the following snippet to measure hashing performance on a typical case and the degenerate case (sorted by length):

```python
from time import time
N = 10**9
# Generate typical case
s = ak.random_strings_uniform(2, 100, N, seed=1)
t0 = time()
h1 = s.hash()
orig_time = time() - t0
# Sort by length for degenerate case
perm = ak.argsort(s.get_lengths())
ssorted = s[perm]
t0 = time()
h2 = ssorted.hash()
sorted_time = time() - t0
```

Here are timings on 10 locales of UM

| branch | orig_time (s) | sorted_time (s) | orig_rate (GiB/s) | sorted_rate (GiB/s) |
| ------- | ----------- | ----------- | ---------- | ---------- |
| master | 0.226 | 681 | 212 | 0.070 |
| follow-seg-bytes | 0.335 | 1.48 | 143 | 32.4 |

So it looks like `computeOnSegments` is about 50% slower for the typical case (I confirmed this with `N = 10**10`) but almost 500x faster for the degenerate case.

ronawho:
Nice! I think it's definitely worth making this switch to help the degenerate case. I will see if there's any obvious optimizations to reduce the overhead for the less degenerate cases. I also think it'd be worth adding a benchmark to track this (surprise, surprise I know). I'm a little surprised the non-sorted and sorted performance isn't a little closer with the new scheme. I guess in this case the last locale will have all the shortest strings to hash and so the SipHash rate will be slowest there.

My machine is still unavailable but should be back up tomorrow or Monday for me to investigate any obvious perf optimizations.

reuster986:
@ronawho I think in addition to the locale with all the short strings spending the most time on SipHash, it will also have to send the most return values remotely via aggregation, so that might compound the load imbalance. I think the current results should be acceptable, though, at least for the use cases I can think of. But I won't stop you from investigating further optimizations :)

src/SegmentedComputation.chpl
```diff
@@ -0,0 +1,89 @@
...
+    
+    // Start task parallelism
+    coforall loc in Locales {
+      on loc {
+        const myFirstSegIdx = startSegInds[loc.id];
+        const myNumSegs = numSegs[loc.id];
+        const mySegInds = {myFirstSegIdx..#myNumSegs};
+        // Segment offsets whose bytes are owned by loc
+        // Block transfer -- should use aggregation?
+        const mySegs = segments[mySegInds];
```
ronawho, line 70: @bradcray -- this is doing something like:

```chpl
const DR: [{start..#length}] int = BlockDist[{start..#length}];
```

Where we'll only be pulling from 1 or 2 remote locales. Surprisingly this results in privatizing something today, but I'm not sure why. Does that makes sense to you? Switching to:

```chpl
var DR: [{start..#length}] int = 0;
DR = BlockDist[{start..#length}];
```

gets rid of the privatization, but is uglier. 

Using `-schpl_serializeSlices` also gets rid of the privatization, but for this I'm more curious what we're even privatizing here and why.

ronawho: Not sure what I was thinking here -- of course slicing results in privatization today and that's why serializing slices is a benefit (I got myself confused here -- I had an older version that changed to split declaration and assignment and that got rid of privatization, but that was due to some other changes I had in the Chapel repo.)

@reuster986 today this slice assignment is fairly expensive due to the eager privatization (will be solved by https://github.com/chapel-lang/chapel/pull/18911, which I think we're hoping to get in this release) and would be further optimized by fixing https://github.com/chapel-lang/chapel/issues/14479. With hacks for both of those in place, this assignment gets much faster for me.

In the meantime I think the following aggregated version will be faster:

```chpl
var mySegs, myLens: [mySegInds] int;
forall i in mySegInds with (var agg = new SrcAggregator(int)) {
  agg.copy(mySegs[i], segments[i]);
  agg.copy(myLens[i], lengths[i]);
}
```

With my hacky optimizations in place the slice-assignment is faster than the aggregated version so I'm not sure which version we want to go with for right now. I might merge with what you have now, and I'll follow up with a PR to switch to aggregation after we get some performance data, and then open up an issue to ensure we revert that change when slice-serialization is enabled by default.

ronawho:
16-node-cs-hdr results currently:

```
./test-bin/SipHashSpeedTest -nl 16 --trace=false --NINPUTS=10_000_000_001 --INPUTSIZE=64 --mode=testMode.{fixed,variable} --computeOnSegments={false,true}
```

| method            | fixed-length | variable-length |
| ----------------- | -----------: | --------------: |
| localizeSlice     | 0.49s        | 2.27s           |
| computeOnSegments | 1.79s        | 1.88s           |

So at least for this config even the slightly variable-length test-case performs better than before. The fixed length is slower, but that's not surprising (and I suspect is incredibly unlikely in real use-cases). We still have to do all the work we were doing before in the fixed-length case to hash and then there's extra time for `computeSegmentOwnership` and copying the portions of the lengths/segments arrays.

Here are the comm counts:

| method            |       | fixed |      |   |         | var   |      |
| ----------------- | ----: | ----: | ---: | - | ------: | ----: | ---: |
|                   | GETS  | PUTS  | ONS  |   | GETS    | PUTS  | ONS  |
| localizeSlice     | 585   |   0   |   30 |   | 2258513 |   0   |   30 |
| computeOnSegments | 9090  | 135   | 1920 |   |   38937 | 299   | 2114 |

---

I'm starting by looking at the fixed length case to see if there's obvious places to reduce the amount of comm. For `./test-bin/SipHashSpeedTest -nl 16 --trace=false --NINPUTS=10_000_000_001 --INPUTSIZE=64  --mode=testMode.fixed --computeOnSegments` I see roughly 0.5s for `computeSegmentOwnership` and 1.0 seconds for the spmd coforall that contains the hashing code.

In `computeSegmentOwnership` most of the time is spent default initializing and then assigning to arrays with domain `D`. You can play some games to avoid default init. The following patch get's this from 0.50s to 0.42s (0.44s with the slightly less ugly version noted in a comment)

<details>

```diff
diff --git a/src/SegmentedComputation.chpl b/src/SegmentedComputation.chpl
index abbc2ce..9ac9833 100644
--- a/src/SegmentedComputation.chpl
+++ b/src/SegmentedComputation.chpl
@@ -3,16 +3,10 @@ module SegmentedComputation {
   use SipHash;

   proc computeSegmentOwnership(segments: [?D] int, vD) {
-    // Locale that owns each segment's bytes
-    var startLocales: [D] int;
     const low = vD.low;
     const size = vD.size;
-    // Vector op on segments, should be fast
-    forall (sl, seg) in zip(startLocales, segments) {
-      // which locale do my bytes start on?
-      sl = ((seg - low)*numLocales) / size;
-    }
-
+    // Locale that owns each segment's bytes
+    var startLocales: [D] int = [seg in segments] ((seg - low)*numLocales) / size;
     // Index of first segment each locale owns bytes for
     var startSegInds: [LocaleSpace] int;
     // Mark true where owning locale changes
@@ -35,7 +29,13 @@ module SegmentedComputation {
       }
     }

-    var lengths: [D] int;
+    // Reuse startLocales space to save time creating another array. I find
+    // this ugly and not worth the time. An in-between option would to do
+    // something like:
+    //     const lengths = [(s, i) in zip(segments, D)]
+    //       if (i == D.high) then vD.size - s else segments[i+1] - s;
+    // to avoid the default-init of lengths
+    ref lengths = startLocales;
     forall (l, s, i) in zip(lengths, segments, D) {
       if (i == D.high) {
         l = vD.size - s;
```

</details>

Where I don't like the `ref` reuse of an older array, but I don't find the code that initializes while declaring to be offensive. The performance difference is relatively small so I think go with whatever you find easiest to read.

In the SPMD region there is a fair bit of comm to do the slice-assignment. See https://github.com/Bears-R-Us/arkouda/pull/1060/files#r802663518. 

There was also some surprise comm from creating the aggregators. I captured this in https://github.com/Bears-R-Us/arkouda/issues/1081 and expect to address that today.

ronawho:
I left some comments about potential performance optimizations, but feel free to leave those for me to do in a followup PR. Overall I think this is a big win and avoids some serious performance issues.

src/SegmentedComputation.chpl
reuster986: Thanks! I updated to use aggregators for this.

ronawho:
Updated performance figures for `./test-bin/SipHashSpeedTest -nl 16 --trace=false --NINPUTS=10_000_000_001 --INPUTSIZE=64 --mode=testMode.{fixed,variable} --computeOnSegments={false,true}`


| method                      | fixed-length | variable-length |
| --------------------------- | -----------: | --------------: |
| localizeSlice               | 0.49s        | 2.27s           |
| computeOnSegments (76667f2) | 1.79s        | 1.88s           |
| computeOnSegments (9d0575e) | 1.47s        | 1.48s           |

ronawho (approved):

Issue #1058, opened 2022-02-01
Reduce memory usage for `argsort`/`radixSortLSD_ranks`


ronawho:
Today `radixSortLSD_ranks` has 2 temp arrays that hold the keys and ranks, and then at the very end just strips off the ranks to return that. This leads to ~5x memory overhead, but I think we can destroy 1 of the temp arrays before creating the return array to limit that to ~4x.

PR #1061, opened 2022-02-01
Reduce argsort memory overhead

ronawho:
Reduce the memory overhead of `argsort`/`radixSortLSD_ranks` from ~5x to
~4x. Previously, `radixSortLSD_ranks` had 3 large arrays in existence at
the same time -- 2 temp key+rank arrays and a return rank array. This
creates a `radixSortLSD` wrapper that returns the key+ranks, and updates
`radixSortLSD_ranks` to call that. This means we only have 1 temp array
when we create the ranks arrays, which reduces the peak memory to the 2
temp key+ranks arrays.

I created `radixSortLSD` instead of just adjusting `radixSortLSD_ranks`
to destroy one of the temp arrays early since we can use it to optimize
in1d, which needs the sorted keys and the ranks, but today does that
inefficiently with a rank sort and gathering values into a sorted array.
See #1062 for more information on this optimization.

Also update `radixSortLSD_memEst` to reflect the reduced memory usage
and use it in additional places that were duplicating the calculation.

Resolves https://github.com/Bears-R-Us/arkouda/issues/1058

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -307,8 +307,9 @@ module ArgSortMsg
         // TODO support arbitrary size with array-of-arrays or segmented array
         proc mergedArgsort(param numDigits) throws {
 
-          overMemLimit(((4 + 3) * size * (numDigits * bitsPerDigit / 8))
-                       + (2 * here.maxTaskPar * numLocales * 2**16 * 8));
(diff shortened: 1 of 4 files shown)


mhmerrill (approved):
This looks ok to me

PR #1063, opened 2022-02-02
Optimize sorting calls when both keys and ranks are needed

### LINKED ISSUE #1062: Optimize sorting when both keys and ranks are needed
Today when both the sorted keys and ranks are needed a `radixSortLSD_ranks` call is done and then the keys are gathered into sorted location with the ranks. This is wasteful though because radixSort is already sorting the keys as part of the rank sort and `radixSortLSD_ranks` just discards them at the end. 

We should be able to optimize this by using the `radixSortLSD` version added in #1061 that returns both the sorted keys and the ranks.

ronawho:
Previously, code that needed both the sorted keys and the ranks was
doing a rank sort and then gathering to sort the keys. Sorting the keys
is already done as part of the main radixSort code, so call the new
version that returns both the keys and values and then split that into 2
arrays. In some cases it would be pretty easy to just use the combined
array, but I wasn't sure if that was worthwhile.

This also removes some explicit `isSorted` checks since the sort code
will already do that check and removes an unused `assumeSorted` from
`uniqueGroup`

On 16-node-cs-hdr I see large int in1d go from 4.8 to 5.4 GiB/s and
setdiff go from 3.6 to 4.1 GiB/s. I expect some improvements for codes
using the other areas that were optimized.

Resolves #1062
Part of #970

### DIFF
--- a/src/In1d.chpl
+++ b/src/In1d.chpl
@@ -74,10 +74,10 @@ module In1d
         /* ar[{u1.size..#u2.size}] = u2; */
         ar[D.interior(u2.size)] = u2;
         // Sort unique arrays together to find duplicates
-        var order = radixSortLSD_ranks(ar);
         var sar: [D] int;
-        forall (s, o) in zip(sar, order) with (var agg = newSrcAggregator(int)) {
-            agg.copy(s, ar[o]);
+        var order: [D] int;
+        forall (s, o, so) in zip(sar, order, radixSortLSD(ar)) {
+          (s, o) = so;
         }
         // Duplicates correspond to values in both arrays
         var flag: [D] bool;
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -7,7 +7,7 @@ module SegmentedArray {
   use UnorderedCopy;
   use SipHash;
   use SegStringSort;
-  use RadixSortLSD only radixSortLSD_ranks;
+  use RadixSortLSD;
   use PrivateDist;
   use ServerConfig;
   use Unique;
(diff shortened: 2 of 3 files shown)


ronawho:
Draft since it relies on #1061

src/In1d.chpl
```diff
@@ -74,10 +74,11 @@ module In1d
         /* ar[{u1.size..#u2.size}] = u2; */
         ar[D.interior(u2.size)] = u2;
         // Sort unique arrays together to find duplicates
-        var order = radixSortLSD_ranks(ar);
+        // TODO could further optimize to avoid splitting into 2 arrays
```
ronawho, line 77: e.g. the following diff would allow us to avoid creating the 2 split arrays (slight faster since we avoid array creation for 2 arrays and avoid the copy):

```diff
diff --git a/src/In1d.chpl b/src/In1d.chpl
index 6df1bc6c..e448568e 100644
--- a/src/In1d.chpl
+++ b/src/In1d.chpl
@@ -74,20 +74,15 @@ module In1d
         /* ar[{u1.size..#u2.size}] = u2; */
         ar[D.interior(u2.size)] = u2;
         // Sort unique arrays together to find duplicates
-        // TODO could further optimize to avoid splitting into 2 arrays
-        var sar: [D] int;
-        var order: [D] int;
-        forall (s, o, so) in zip(sar, order, radixSortLSD(ar)) {
-          (s, o) = so;
-        }
+        var sar_order = radixSortLSD(ar);
         // Duplicates correspond to values in both arrays
         var flag: [D] bool;
-        flag[D.interior(-(D.size-1))] = (sar[D.interior(D.size-1)] == sar[D.interior(-(D.size-1))]);
+        flag[D.interior(-(D.size-1))] = (sar_order[D.interior(D.size-1)][0] == sar_order[D.interior(-(D.size-1))][0]);
         // Get the indices of values from u1 that are also in u2
         // Because sort is stable, original index of left duplicate will always be in u1
         var ret: [D] bool;
-        forall (o, f) in zip(order, flag) with (var agg = newDstAggregator(bool)) {
-            agg.copy(ret[o], f);
+        forall (o, f) in zip(sar_order, flag) with (var agg = newDstAggregator(bool)) {
+            agg.copy(ret[o[1]], f);
         }
         // Use the inverse index to map from u1 domain to ar1 domain
         var truth: [aD1] bool;

```

I think it makes the code harder to follow and no longer index-neutral, but curious for other input. Maybe from @mhmerrill in particular.

The performance difference is minor, maybe 0.5% for in1d

mhmerrill: @ronawho I think I'd opt for easier to understand code at this point.

mhmerrill:
@ronawho is this ready for review? does it still need to be a draft?

ronawho:
Ready for review now.

mhmerrill (approved):
looks good to me

reuster986 (commented):
I'm late to the party, but this looks great!

PR #1065, opened 2022-02-02
Add a --saveUsedModules flag to save modules used in a server run

### LINKED ISSUE #1064: Add a flag to output all modules that were used in a arkouda_server session
With the modularization of the Arkouda build, it can sometimes be difficult to figure out which modules are needed for a particular run, so if there were a way to report the modules that were used upon server close or something like that it could make it simpler for a user to figure out what modules to include in their next run.

bmcdonald3:
This PR adds a --saveUsedModules flag that prints the name of all
modules that had a mesasge function called in a UsedModules.cfg file
that can be reused to speed up the next build. The function is called
upon `ak.shutdown()`.

To use this flag on a benchmark, it can be run:
`./benchmarks/run-benchmarks.py IO --server-args='--saveUsedModules`.
To use this flag on an interactive run of the Arkouda server, run:
`./arkouda_server --saveUsedModules`.

Resolves #1064

### DIFF
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -917,20 +917,20 @@ module SegmentedMsg {
   
   proc registerMe() {
     use CommandMap;
-    registerFunction("segmentLengths", segmentLengthsMsg);
-    registerFunction("segmentedHash", segmentedHashMsg);
-    registerFunction("segmentedSearch", segmentedSearchMsg);
-    registerFunction("segmentedFindLoc", segmentedFindLocMsg);
-    registerFunction("segmentedFindAll", segmentedFindAllMsg);
-    registerFunction("segmentedPeel", segmentedPeelMsg);
-    registerFunction("segmentedSub", segmentedSubMsg);
-    registerFunction("segmentedIndex", segmentedIndexMsg);
-    registerFunction("segmentedBinopvv", segBinopvvMsg);
-    registerFunction("segmentedBinopvs", segBinopvsMsg);
-    registerFunction("segmentedGroup", segGroupMsg);
-    registerFunction("segmentedIn1d", segIn1dMsg);
-    registerFunction("randomStrings", randomStringsMsg);
-    registerFunction("segStr-assemble", assembleStringsMsg);
-    registerBinaryFunction("segStr-tondarray", segStrTondarrayMsg);
+    registerFunction("segmentLengths", segmentLengthsMsg, getModuleName());
+    registerFunction("segmentedHash", segmentedHashMsg, getModuleName());
+    registerFunction("segmentedSearch", segmentedSearchMsg, getModuleName());
+    registerFunction("segmentedFindLoc", segmentedFindLocMsg, getModuleName());
+    registerFunction("segmentedFindAll", segmentedFindAllMsg, getModuleName());
+    registerFunction("segmentedPeel", segmentedPeelMsg, getModuleName());
+    registerFunction("segmentedSub", segmentedSubMsg, getModuleName());
+    registerFunction("segmentedIndex", segmentedIndexMsg, getModuleName());
+    registerFunction("segmentedBinopvv", segBinopvvMsg, getModuleName());
+    registerFunction("segmentedBinopvs", segBinopvsMsg, getModuleName());
+    registerFunction("segmentedGroup", segGroupMsg, getModuleName());
+    registerFunction("segmentedIn1d", segIn1dMsg, getModuleName());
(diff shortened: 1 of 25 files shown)


benchmarks/run_benchmarks.py
```diff
@@ -105,6 +105,7 @@ def create_parser():
     parser.add_argument('--annotations', default='', help='File containing annotations')
     parser.add_argument('--configs', help='comma seperate list of configurations')
     parser.add_argument('--start-date', help='graph start date')
+    parser.add_argument('--save-used-modules', default=False, action='store_true', help='Save used modules in a `UsedModules.cfg` file')
```
ronawho, line 108: I don't think this will be used widely enough to warrant it's own flag and we can get the same behavior with `--server-args --saveUsedModules`

If we decide to keep this here, we should move it below `--server-args` and add it to `runClient` as well (but I'm in favor of not adding it)

ronawho (approved):
👍 

Modular builds have been a huge time saver for me when doing performance tuning. Today it takes me a few iterations to figure out what modules I actually need for a particular benchmark, but this makes that way simpler/automated.

src/SegmentedMsg.chpl
```diff
@@ -917,20 +917,20 @@ module SegmentedMsg {
...
+    registerFunction("segmentedFindAll", segmentedFindAllMsg, getModuleName());
+    registerFunction("segmentedPeel", segmentedPeelMsg, getModuleName());
+    registerFunction("segmentedSub", segmentedSubMsg, getModuleName());
+    registerFunction("segmentedIndex", segmentedIndexMsg, getModuleName());
+    registerFunction("segmentedBinopvv", segBinopvvMsg, getModuleName());
+    registerFunction("segmentedBinopvs", segBinopvsMsg, getModuleName());
+    registerFunction("segmentedGroup", segGroupMsg, getModuleName());
+    registerFunction("segmentedIn1d", segIn1dMsg, getModuleName());
+    registerFunction("randomStrings", randomStringsMsg, getModuleName());
+    registerFunction("segStr-assemble", assembleStringsMsg, getModuleName());
```
glitch, line 933: I don't know how much Reflection costs, so if it's negligible dis-regard:  You could save `getModuleName()` at the beginning of `registerMe` and pass the string instead of calling the function every time.  i.e.
```chapel
const mod_name = getModuleName();
registerFunction("cmd1", myFunc1, mod_name);
registerFunction("cmd2", myFunc2, mod_name);
...
```


Also, separately, why did `registerBinaryFunction` below not get the same treatment?

glitch (commented):
A couple of thoughts here.

1. If you're adding this for `registerFunction` why not add it to `registerBinaryFunction`?
2. You made this optional by adding a second function, if the goal is to record all of the modules used, shouldn't it be required?
3. I'm a bit on the fence here, if the goal is to determine which modules are needed to build for iterative development & benchmarks it seems slightly intrusive... but then again, I guess we don't really have a way to easily figure it out today... and I mean, what are log statements really other than notes to developers saying what happened.

I think I talked myself into being on board :smile: .  At some point I think we need to introduce some type of rudimentary dependency management in `ServerModules.cfg` and maybe add some extra developer tooling in util, but that fills a slightly different need than this.

hokiegeek2 (approved):
this looks solid as-is. Longer term, it would be really cool to have a way to add function annotations to, in this case, specify the functions to register, which would obviate the need for all the registerMe function calls (not a criticism, just a wish list item 😄). As always, good job, @bmcdonald3. 👍

src/SegmentedMsg.chpl
bmcdonald3: Ya, I will add it to `registerBinaryFunction` as well

bmcdonald3: The reason that I left the overload of `registerFunction` was that I didn't want `arkouda_server` and other base modules to be added to the list of output modules, since that would not be a valid configuration file. I suppose we could instead just check if it is `arkouda_server` or in a set of the base modules or something, but I think I prefer the overload as it is today.

pierce314159 (approved):

glitch (changes requested):
I saw you added the `modName` to `registerBinaryFunction`, but it looks like you forgot to add the corresponding linkage in `arkouda_server` - `else if commandMapBinary.contains(cmd)`... line 366.  When a binary command is invoked you'll need to add its module to the set of usedModules for completeness.

bmcdonald3:
@glitch good catch, added

glitch (approved):

PR #1066, opened 2022-02-03
Rework IO benchmarks to use the same benchmarks for Parquet and HDF5

bmcdonald3:
Today, the HDF5 and Parquet benchmarks are nearly identical, just
with the different Parquet/HDF5 API calls. This PR combines them
both into a single benchmark that defaults to HDF5 and then adds
a separate Parquet version with just a main function and parser.

This also allows the IO benchmarks to run just write/read tests
if you only wish to test one of the 2 and wanted to keep the files
around afterwards, for example.

Additionally, a benchmark is added to test writing/reading multiple
files, which represents a more common use case and illustrates how
Parquet and HDF5 stack up for multiple file reading, which seems to
be a more realistic use case that is also made more interesting
after having parallelized Parquet file reads.


benchmarks/IO.py
```diff
@@ -7,51 +7,78 @@
...
+    
     readtimes = []
     for i in range(trials):
         start = time.time()
-        a.save(path)
-        end = time.time()
-        writetimes.append(end - start)
-        start = time.time()
-        b = ak.load(path)
+        a = ak.load(path+'*') if not parquet else ak.read_parquet(path+'*')
```
reuster986, line 44: In general, `ak.load` is not supposed to take wildcards. This always trips me up, but if an array was saved with `a.save(path)`, then it should be loaded with `ak.load(path)` -- the exact same argument -- otherwise it might do crazy things. If a wildcard is necessary, it should be used with `ak.read_all()` or `ak.read_hdf()`.

benchmarks/IO.py
```diff
@@ -7,51 +7,78 @@
...
+        if multifile:
+            b = ak.randint(0, 1, N, dtype=ak.float64, seed=seed)
 
-    a.save(path)
-    b = ak.load(path)
-    for f in glob(path+"_LOCALE*"):
+    a.save(f"{path}{1}") if not parquet else a.save_parquet(f"{path}{1}")
+    if multifile:
+        b.save(f"{path}{2}") if not parquet else b.save_parquet(f"{path}{2}")
+    c = ak.load(path+'*') if not parquet else ak.read_parquet(path+'*')
```
reuster986, line 74: Same comment as above.

reuster986 (changes requested):
This will be really helpful for IO comparison! Just one minor change requested.

Issue #1011, opened 2021-12-30
Add `uint64` dtype


reuster986:
We've been toying with the idea of an unsigned integer type for a while, but I think we've finally reached a critical mass of use cases where it would be helpful. Most of the code can just be copied from what is defined for `DType.Int64`, unless we want to use this as an opportunity to consolidate code with dispatch or generic helper functions.

For the CI, we can `uint64` to the operator test and measure correctness against numpy.

Also, we can't forget I/O and casting.

mhmerrill:
@Bears-R-Us/arkouda-core-dev we should have a discussion of this before we run off and do it. I have been puzzling over the operator code organization. there are a few different ways to organize the code and the current way keeps all of the operators for a specific type combination together. Remember we don't implement all operators for all type combinations.

mhmerrill:
There will be additions to the client and the server...
I think the steps look something like this:
- [ ] add new dtype `uint64` to the server with conversions to chapel `uint` type
- [ ] add new dtype `uint64` to the client aliases and such
- [ ] add binOps for `uint64`
- [ ] add indexing for `uint64`
- [ ] add sorting for `uint64` : `argsort`, `sort`, `coargsort`
- [ ] add setops and stuff for `uint64` : `in1d`, `unique`, `union1d`,  etc...
- [ ] more stuff

bmcdonald3:
With some help from Elliot, Pierce, and Mike, we've gotten https://github.com/Bears-R-Us/arkouda/pull/1070 passing CI tests and ready for review. 

This PR includes:
 * argsort, sort, coargsort (Thank you Elliot)
 * indexing for uint with [int] and [pdarray]
 * casting to/from uint64
 * Parquet uint64 reading/writing
 * Random uint array generation
 * '+' binary operation for uint+uint pdarrays
 
This PR does not implement:
 - most uint binary operations
 - HDF5 uint support 
 - setops for uints

glitch:
For incremental follow up, here's the brief synopsis of my PR comments and spots I noticed we need to still address

- [ ] `sorting.py` - `sort` and `argsort` if the pdarray is size 0 we return `zeros(0, dtype=int64)` and there is also a check on the pdarray dtype which follows the size test.  We may want to flip the conditions and check the incoming dtype first.  Then we can make the zeros call do `zeros(0, dtype=pda.dtype)`.
[...]

PR #1070, opened 2022-02-04
Add support for arrays of unsigned integers 

bmcdonald3:
This PR adds support for the uint64 dtype.

The supported functionality in this initial work include:
 * argsort, sort, coargsort (Thank you Elliot)
 * indexing with [int] and [pdarray]
 * casting to/from uint64
 * Parquet uint64 reading/writing
 * Random uint array generation
 * '+' binary operation for uint+uint pdarrays

For other remaining work, see https://github.com/Bears-R-Us/arkouda/issues/1011.


arkouda/dtypes.py
```diff
@@ -16,25 +16,27 @@
...
 float64 = np.dtype(np.float64)
 uint8 = np.dtype(np.uint8)
+uint64 = np.dtype(np.uint64)
 str_ = np.dtype(np.str_)
 npstr = np.dtype(str)
 
 # Union aliases used for static and runtime type checking
 bool_scalars = Union[builtins.bool, np.bool_]
 float_scalars = Union[float,np.float64]
 int_scalars = Union[int,np.int64]
```
mhmerrill, line 34: add uint64 here

arkouda/dtypes.py
```diff
@@ -66,16 +69,17 @@ def __repr__(self) -> str: # type: ignore
 
 ARKOUDA_SUPPORTED_INTS = (int,np.int64)
```
mhmerrill, line 70: add uint64 here

mhmerrill: also below https://github.com/Bears-R-Us/arkouda/blob/293da309874bcd010b2918965fe569df16ad3724/arkouda/dtypes.py#L146-L149
needs to be modified to split the `uint` and `int`  interpretation apart.

src/BinOp.chpl
```diff
@@ -251,6 +261,7 @@ module BinOp
       var repMsg = "created %s".format(st.attrib(rname));
       return new MsgTuple(repMsg, MsgType.NORMAL);
     }
+    writeln(l.dtype, r.dtype);
```
mhmerrill, line 264: debugging?

src/NumPyDType.chpl
```diff
@@ -14,6 +14,7 @@ module NumPyDType
     */
     proc whichDtype(type etype) param : DType {
       if (etype == int) {return DType.Int64;}
+      if (etype == uint(64)) {return DType.UInt64;}
```
mhmerrill, line 17: I think the default size is 64 for both int and uint. we don't carry the `(64)` around in the code because it is the default. @bradcray what do you think?

src/RandMsg.chpl
```diff
@@ -30,14 +30,12 @@ module RandMsg
         var (lenStr,dtypeStr,aMinStr,aMaxStr,seed) = payload.splitMsgToTuple(5);
         var len = lenStr:int;
         var dtype = str2dtype(dtypeStr);
+        writeln(dtype);
```
mhmerrill, line 33: debugging?

src/RandMsg.chpl
```diff
@@ -30,14 +30,12 @@ module RandMsg
...
 
         // get next symbol name
         var rname = st.nextName();
+
         
-        // if verbose print action
```
mhmerrill, line 37: why remove this?

src/NumPyDType.chpl
bradcray: I agree that `uint(64)` is unnecessary and that `uint` should suffice.  Even if/when we add `uint(128)` or `int(128)` in the future, I don't think we'll be redefining `int` or `uint` or making it "implementation-defined" as C does.  

I can imagine different codes choosing to use either notation from a "completely self-documenting" vs. "brevity" angle.  In cases like this, I'd probably just follow the lead of the existing code and use `uint` ("if it was good enough for them, it's good enough for us").

src/RandMsg.chpl
bradcray: Poor Ben's going to be shamed into never opening another draft PR again.  :D  I'd hold off on reviewing details like this until he gets to the point of considering this no longer to be in draft form.  If he develops like me (throw in a bunch of code that I'll eventually remove), he'll look at the diff and boil it down to the essentials between now and then.

mhmerrill:
@bmcdonald3 my branch has some other changes you may want because I took a different tact at the problem. I was trying to get everything to pass testing and just add
```python
a = ak.zeros(10,dtype=ak.uint64)
print(a)
a.fill(np.uint64(1))
print(a)
```
i'll link my branch here when I am able to push it to the repo.

src/RandMsg.chpl
mhmerrill: I don't think he should feel ashamed this is a great help to us.
I learn stuff from his code all the time and I don't have time to be ashamed of anything.
Sorry if I am too hard. I wanted to see the code and thought I would make a few suggestions to get this off the ground ASAP.

@bmcdonald3 I think you and your code are awesome and I don't butter people up ;-)

bmcdonald3: Thanks Mike and I definitely appreciate the feedback, I am far from expert in the client-side code especially.

pierce314159:
most of the assertion errors are because we hard coded what the error message should be so when we added uint64 , it's not matching.

```python
 self = <extrema_test.ArgMaxKTest testMethod=test_error_handling>

    def test_error_handling(self):
        testArray = ak.randint(0, 100, 100)
    
        with self.assertRaises(TypeError) as cm:
            ak.argmaxk(list(range(0,10)), 1)
        self.assertEqual('type of argument "pda" must be arkouda.pdarrayclass.pdarray; got list instead',
                         cm.exception.args[0])
    
        with self.assertRaises(TypeError) as cm:
            ak.argmaxk(testArray, '1')
>       self.assertEqual('type of argument "k" must be one of (int, int64); got str instead',
                         cm.exception.args[0])
E       AssertionError: 'type of argument "k" must be one of (int, int64); got str instead' != 'type of argument "k" must be one of (int, uint64, int64); got str instead'
E       - type of argument "k" must be one of (int, int64); got str instead
E       + type of argument "k" must be one of (int, uint64, int64); got str instead
E       ?                                          ++++++++
```

 I think it's probably enough to say `self.assertRaises(TypeError)` and not check the exact wording of the error message. But if we want to avoid changing too much in this PR we can just change those tests to something like

```python
self.assertEqual('type of argument "k" must be one of (int, uint64, int64); got str instead',
                         cm.exception.args[0])
```

glitch (approved):
Overall looks good.  I didn't see any show stoppers and only minor improvements with some of those probably being beyond the scope of the PR and certainly in the realm of incremental imrpovements.

:+1:

PR #1071, opened 2022-02-07
Issue #1070: Adds UInt64 support to arrayMsg and tondarrayMsg

glitch:
Issue #1070: Adds UInt64 support to arrayMsg and tondarrayMsg
plus unit test.

### DIFF
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -596,7 +596,7 @@ class IOTest(ArkoudaTest):
             b2 = ak.load(f"{tmp_dirname}/single_string", dataset="b1")
             self.assertEqual(str(b1), str(b2))
 
-    def testUint64ToHDF5(self):
+    def testUint64ToFromArray(self):
         """
         Test conversion to and from numpy array / pdarray using unsigned 64bit integer (uint64)
         """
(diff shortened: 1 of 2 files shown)


tests/io_test.py
```diff
@@ -596,6 +596,15 @@ def testSmallStringArrayToHDF5(self):
             b2 = ak.load(f"{tmp_dirname}/single_string", dataset="b1")
             self.assertEqual(str(b1), str(b2))
 
+    def testUint64ToHDF5(self):
+        """
+        Test conversion to and from numpy array / pdarray using unsigned 64bit integer (uint64)
+        """
+        npa1 = np.array([18446744073709551500, 18446744073709551501, 18446744073709551502])
```
bmcdonald3, line 603: Do we know this is uint64 just based on the number of digits?

tests/io_test.py
```diff
@@ -596,6 +596,15 @@ def testSmallStringArrayToHDF5(self):
             b2 = ak.load(f"{tmp_dirname}/single_string", dataset="b1")
             self.assertEqual(str(b1), str(b2))
 
+    def testUint64ToHDF5(self):
```
ronawho, line 599: Not an HDF5 test, is it?

tests/io_test.py
glitch: 18446744073709551615 is uint64 max value I think (based on https://docs.microsoft.com/en-us/dotnet/api/system.uint64.maxvalue?view=net-6.0)  although I can add a specific dtype on this to make sure.

bmcdonald3 (approved):

pierce314159 (approved):
Looks good!

PR #1073, opened 2022-02-07
Issue 1011: Adds support for reading/writing uint64 to HDF5

glitch:
Issue #1011 
Adds support for reading/writing `uint64` to HDF5
Adds unit test for verification.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -1834,7 +1834,7 @@ module HDF5Msg {
         return idx == numLocales-1;
     }
 
-        /* 
+    /**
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
Looks good to me

glitch:
Note: @pierce314159 pointed out I was using the wrong ticket number in the commit message & ticket description.  I force-pushed this branch to fix that info.

reuster986 (approved):
Looks great! 

As I understand it, the only HDF5 dtype that will get read in as arkouda uint64 is HDF5 uint64; all other HDF5 integer types -- (u)int(8|16|32) and int64 -- will get read in as arkouda int64. Is that accurate? I approve of that design; just making sure I understand.

PR #1080, opened 2022-02-09
Add unary operations for uint pdarrays

bmcdonald3:
This PR adds support for popcount, clz, ctz, cumsum, and parity for uint pdarrays.

Related to: https://github.com/Bears-R-Us/arkouda/issues/1011

### DIFF
--- a/src/EfuncMsg.chpl
+++ b/src/EfuncMsg.chpl
@@ -212,6 +212,38 @@ module EfuncMsg
                     }
                 }
             }
+            when (DType.UInt64) {
+                var e = toSymEntry(gEnt,uint);
+                select efunc
+                {
+                    when "popcount" {
+                        var a = st.addEntry(rname, e.size, uint);
+                        a.a = popcount(e.a);
+                    }
+                    when "clz" {
+                        var a = st.addEntry(rname, e.size, uint);
+                        a.a = clz(e.a);
+                    }
+                    when "ctz" {
+                        var a = st.addEntry(rname, e.size, uint);
+                        a.a = ctz(e.a);
+                    }
+                    when "cumsum" {
+                        // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit
+                        overMemLimit(numBytes(uint) * e.size);
+                        st.addEntry(rname, new shared SymEntry(+ scan e.a));
+                    }
+                    when "parity" {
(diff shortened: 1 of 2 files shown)


glitch (approved):

src/EfuncMsg.chpl
```diff
@@ -212,6 +212,29 @@ module EfuncMsg
                     }
                 }
             }
+            when (DType.UInt64) {
+                var e = toSymEntry(gEnt,uint);
+                select efunc
+                {
+                    when "popcount" {
```
mhmerrill, line 219: @reuster986 maybe these should mirror more of the `int64` ops like `parity`, also should we return an `int64` from these functions?

mhmerrill: I thought about it and I think the return values should be `int64` or `int` in Chapel.

mhmerrill: I am thinking we need `cumsum` and `parity`.
`cumsum` returns `uint64` or `uint` in Chapel
`parity` returns `uint64` or `uint` in Chapel

bmcdonald3: Alright, I have changed them to be of the form `a.a = clz(e.a):int;`.

Do you think that approach is what we would want or instead something like: `a.a = clz(e.a:int);`, where the input is cast to an `int` rather than the result. I believe that there could be different behavior in some cases between the two.

reuster986: I don't have a strong opinion either way, but I'm curious why an `int64` return is preferable, since all results are guaranteed to be non-negative anyway?

mhmerrill: > I don't have a strong opinion either way, but I'm curious why an `int64` return is preferable, since all results are guaranteed to be non-negative anyway?

ok, i don't have a comeback for that ;-) except that it's the same as the return for `int64` but maybe that doesn't matter.
I hate to whip @bmcdonald3 around but I am ok with changing them back to `uint`.

mhmerrill (approved):
this looks good.

src/EfuncMsg.chpl
reuster986: thanks! no big deal either way on the return type, but this way seems consistent with what chapel already does.

reuster986 (approved):

PR #1082, opened 2022-02-09
Optimize `myLocaleSpace` assignment in aggregators

### LINKED ISSUE #1081: Optimize `LocaleSpace` assignment in aggregators
Today the aggregators do `const myLocaleSpace = LocaleSpace;` to store their peers. This `LocaleSpace` copy was done was done to avoid old overheads with how Chapel tracked arrays declared over domains. That has since been optimized, but during some other work I noticed there is a many-to-one GET from all remote locales to make that assignment because `LocaleSpace` lives on locale 0. Switch to just using `numLocales` since it's a POD value that gets replicated at program startup, so will have no comm.

ronawho:
Change `myLocaleSpace = LocaleSpace` to `myLocaleSpace = 0..<numLocales`
to eliminate communication when creating aggregators. `LocaleSpace`
lives on locale 0, so there was some communication to copy it. Using
`numLocales` eliminates that because `numLocales` is just an `int` that
is replicated at program startup. This eliminates a many-to-one GET that
may have a minor performance benefit, particularly at scale.

While here, also change the non-ugni default buffer size from `8096` to
`8192`. I meant for this to be a power of two in #732 but either made a
typo or a think-o.

I expect some minor performance benefits from these changes on
InfiniBand system.

Resolves #1081

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -6,7 +6,7 @@ module CommAggregation {
   use CommPrimitives;
 
   // TODO should tune these values at startup
-  param defaultBuffSize = if CHPL_COMM == "ugni" then 4096 else 8096;
+  param defaultBuffSize = if CHPL_COMM == "ugni" then 4096 else 8192;
   private config const yieldFrequency = getEnvInt("ARKOUDA_SERVER_AGGREGATION_YIELD_FREQUENCY", 1024);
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
looks ok to me.

pierce314159 (approved):

reuster986 (approved):

PR #1072, opened 2022-02-07
Add a ChplConfig compatibility module for < 1.26

bradcray:
In Chapel 1.26, we're replacing auto-available module `ChapelEnv` with
must-be-used/imported `ChplConfig` and references to `CHPL_*` start
generating warnings as a result on the main Chapel development branch.
This PR updates the sources to use `ChplConfig` to avoid these warnings
while providing a(n empty) compatibility module for past releases of Chapel.

Resolves #1076.

PR #1084, opened 2022-02-09
Closes #1079: `checkRE2.chpl` version bug

### LINKED ISSUE #1079: `checkRE2.chpl` version checking bug
`checkRE2.chpl`  attempts to import the appropriate regex module for the chpl version used 
```C
  if chplVersion < createVersion(1,25) {
    use Regexp;
    var myRegex = compile("a+");
  }
  else {
    use Regex;
    var myRegex = compile("a+");
  }
```
However `use` statements are resolved early in the compiler, such that both `Regex` and `Regexp` will both attempt to be parsed by the compiler regardless of version number.

This results in an arkouda `make` failure for anyone using 1.24.x. An (unfortunate) workaround for this in the meantime is to make arkouda without checks which will skip this check as well as `zmq` and `hdf5`
```
make ARKOUDA_SKIP_CHECK_DEPS=true
```

https://github.com/Bears-R-Us/arkouda/pull/1032/files#r802103781

pierce314159:
This PR (closes #1079):
- Replaces `checkRE2.chpl` with version specific files (`checkRE2_124.chpl` and `checkRE2_125.chpl`)
- Updates `Makefile` to choose the appropriate check based on chpl version

This resolves a bug where users with chpl versions pre `1.25` cannot make arkouda. It also fixes the `Regexp` deprecation warnings for those running `1.25.x`. I have verified this works locally with chpl versions `1.24.1` and `1.25.0`

Bug was encountered by @zhihuidu and reported by @bradcray


bradcray:
@pierce314159 : At a glance, this looks like it should work to me.  I haven't done anything to try and test it (but could if you'd like).

I was thinking this morning about how I would solve this, and ended up thinking that I would try a single `checkRE2.chpl` file that contained `use ArkoudaRegexCompat;` since that matches what Arkouda itself does, so would leverage the compatibility logic that already exists.  However, that would also require adding the same `-M compat/*` flags (via the ARKOUDA_COMPAT_MODULES Makefile variable) to this check test compilation lines, which we don't currently do.  E.g., something like:

>         $(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@

OTOH, the new Makefile logic you're adding is, in a way, replicating the style of logic that creates those flags in the first place, so maybe reusing the existing logic would be better?

It looks like @glitch added or modified that logic most recently, so maybe he'd have thoughts?  @ronawho and I have been talking about how to achieve inter-version compatibility recently as part of my #1072 effort, so tagging him as well.

(Only tangentially related to this effort, I would've also been inclined to add a `Regex.chpl` compatibility module in the `src/compat/lt-125/` directory rather than adding an `ArkoudaRegexCompat` module to both `lt-125` and `ge-125` in order to keep the source file referring to things that we'd hope it would in the long-term and just provide a workaround for 1.24.x).

glitch (approved):
I'm good with this approach.

We could move things around to pick up the ArkoudaRegexCompat.chpl, adding a main to it and then try to compile it instead etc. but at the end of the day this change set will be removed when we drop 1.24 support.  So as long as it's getting the job done in an easy to understand manner I think we're good.

I also tested this on Chapel 1.24 and it worked for me. (`make check-re2` and regular `make`)

mhmerrill (approved):
looks ok

PR #1085, opened 2022-02-09
Add in1d and setops support for uint pdarrays

bmcdonald3:
This PR adds support for `in1d`, `setdiff1d`, `setxor1d`,
`union`, `intersect1d`, `concatenate`, and `unique` for 
`uint` pdarrays.

The testing in `tests/setops_test.py` has been extended to
account for the new `uint` support.

### DIFF
--- a/src/ArraySetopsMsg.chpl
+++ b/src/ArraySetopsMsg.chpl
@@ -53,28 +53,34 @@ module ArraySetopsMsg
         var intersect_maxMem = max(gEnt_sortMem, gEnt2_sortMem);
         overMemLimit(intersect_maxMem);
 
-        select(gEnt.dtype) {
-          when (DType.Int64) {
-             if (gEnt.dtype != gEnt2.dtype) {
-                 var errorMsg = notImplementedError("newIntersect1d",gEnt2.dtype);
-                 asLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);                             
-                 return new MsgTuple(errorMsg, MsgType.ERROR);
-             }
-             var e = toSymEntry(gEnt,int);
-             var f = toSymEntry(gEnt2, int);
-             
-             var aV = intersect1d(e.a, f.a, isUnique);
-             st.addEntry(vname, new shared SymEntry(aV));
+        select(gEnt.dtype, gEnt2.dtype) {
+          when (DType.Int64, DType.Int64) {
+            var e = toSymEntry(gEnt,int);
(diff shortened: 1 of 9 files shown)


reuster986 (approved):
looks great!

pierce314159 (approved):
Looks great to me!

It looks like there's a couple of places in `ArraySetOps` and `Concat` where we might be able to pull out common code as a helper function and pass in the `SymEntry` if that's possible? But either way I think that should be in a separate PR

src/ArraySetopsMsg.chpl
```diff
@@ -53,28 +53,34 @@ module ArraySetopsMsg
...
+
+            var aV = intersect1d(e.a, f.a, isUnique);
+            st.addEntry(vname, new shared SymEntry(aV));
+
+            repMsg = "created " + st.attrib(vname);
+            asLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
+            return new MsgTuple(repMsg, MsgType.NORMAL);
+          }
+          otherwise {
+            var errorMsg = notImplementedError("newIntersect1d",gEnt.dtype);
```
mhmerrill, line 80: is the string here intended? `"newIntersect1d"`

mhmerrill:
> This PR adds support for `in1d`, `setdiff1d`, `setxor1d`, `union`, and `intersect1d` for `uint` pdarrays.
> 
> The testing in `tests/setops_test.py` has been extended to account for the new `uint` support.

@bmcdonald3 check me but I think also being added are `unique` and `concatenate`, also adding support for `uint64` index arrays. these should be added to the description of the PR.

src/ArraySetopsMsg.chpl
pierce314159: I think that string was present before this PR

bmcdonald3: Ya, that string does predate this PR. I thought it was weird and just kept it at first, but now that I think about it a bit more, when I was an intern and implemented this intersect code I had 2 versions of intersect around and called this new one `newIntersect1d`, so I think that just slipped past my intern project... I will fix that here.

bmcdonald3:
Unique and concatenate were added, but I did not thread the `uint` indexing through the message layer to make that work on the client side, only for the support needed in the server.

Would it be valuable to extend this PR to add uint indexing as a server operation?

mhmerrill (approved):
looks good! thanks!

PR #1087, opened 2022-02-09
Update tests and benchmarks correctness tests for uint and add uint docs

bmcdonald3:
In the rush to get the uint support into Arkouda, the tests/benchmarks/
docs got slightly neglected, so this PR fills in some of those gaps.
All benchmarks with supported options have had their dtypes updated
to reflect uint support and also several unit tests have been updated
to reflect their new uint support, such as the sorting test, bitops
test, setops test, and numeric test.

Also in this PR are minor fixes from the initial uint PR:
* fixing `sorting.py` to return the correct dtype for an empty array
* removing duplicate logic in `ArrowFunctions.cpp` for uint reading

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -88,8 +88,10 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
     return ARROWINT64;
   else if(myType->id() == arrow::Type::INT32)
     return ARROWINT32;
-    else if(myType->id() == arrow::Type::UINT64)
+  else if(myType->id() == arrow::Type::UINT64)
     return ARROWUINT64;
+  else if(myType->id() == arrow::Type::UINT32)
+    return ARROWUINT32;
   else if(myType->id() == arrow::Type::TIMESTAMP)
     return ARROWTIMESTAMP;
   else // TODO: error type not supported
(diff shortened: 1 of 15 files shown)


glitch (changes requested):

src/ArrowFunctions.cpp
```diff
@@ -128,16 +128,9 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
...
-          static_cast<parquet::Int64Reader*>(column_reader.get());
-
-        while (reader->HasNext()) {
-          (void)reader->ReadBatch(batchSize, nullptr, nullptr, &chpl_ptr[i], &values_read);
-          i+=values_read;
-        }
-      } else if(ty == ARROWUINT64) {
+      // Since int64 and uint64 Arrow dtypes share a physical type and only differ
+      // in logical type, they must be read from the file in the same way
+      if(ty == ARROWINT64 || ty == ARROWUINT64) {
```
ronawho, line 133: This seems surprising. At the very least I would expect you need to cast the chapel array to `uint64_t*` and I'm surprised the parquet types aren't different. Can you add a test that verifies reading in `max(uint)` works correctly?

And anything needed for reading in uint32?

glitch (changes requested):

src/ArrowFunctions.cpp
bmcdonald3: > I would expect you need to cast the chapel array to uint64_t*

That was my initial thought as well, but Parquet does not even expose anything to read in `uint` values, so that does not work. They are both 64 bit and share the same physical type, so I think that makes sense.

> Can you add a test that verifies reading in max(uint) works correctly?

Added.

> And anything needed for reading in uint32?

I hadn't considered this since I just had `uint64` support in my head, but I've added it now. Since we don't support writing uint32 (or int32), we won't be testing it, but I saw it work locally with some struggles to create the file in pyarrow.

glitch (approved):

PR #1094, opened 2022-02-10
Fix `computeOnSegments` when there's more locales than strings

### LINKED ISSUE #1093: categorical and coargsort failures in nightly testing
Since #1060 many tests that use string coargsort are failing in nightly multilocale testing. Seems to fail with 11+ locales. Running with `ARKOUDA_DEVELOPER` I see:

```
SegmentedComputation.chpl:54: error: halt reached - With a negative count, the range must have a last index.
```

I think this is a simple case of there not being enough data for each locale and we're not skipping some computation on empty locales.

ronawho:
Many tests were failing in nightly 16 locale testing after #1060. This
was caused by trying to hash fewer strings than there were locales. The
empty locales should be a no-op, but ended up trying to do a count with
a negative value, which isn't supported on ranges without a high bound.
Fix that by just making the number of segments 0 if it was negative.

Alternatively we could skip further computation if the size is 0, but
that adds an extra level of indentation, which didn't seem worthwhile.
The operations will be no-ops anyways so it shouldn't matter.

Resolves #1093

### DIFF
--- a/src/SegmentedComputation.chpl
+++ b/src/SegmentedComputation.chpl
@@ -50,7 +50,7 @@ module SegmentedComputation {
     coforall loc in Locales {
       on loc {
         const myFirstSegIdx = startSegInds[loc.id];
-        const myNumSegs = numSegs[loc.id];
+        const myNumSegs = max(0, numSegs[loc.id]);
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
looks good to me!

PR #1096, opened 2022-02-11
Have each arkouda server use a unique identifier for arrays

### LINKED ISSUE #694: CI fails for Python 3.8.8

### LINKED ISSUE #924: Sporadic Unknown symbol error for pytests
On 09/16 some of our nightly testing configurations started failing for registration_test.py with an unknown symbol error. Full test failure:

<details>

```
=================================== FAILURES ===================================
_____________________ RegistrationTest.test_in_place_info ______________________

self = <registration_test.RegistrationTest testMethod=test_in_place_info>

    def test_in_place_info(self):
        """
        Tests the class level info method for pdarray, String, and Categorical
        """
        cleanup()
        my_pda = ak.ones(10, ak.int64)
        self.assertFalse(any([sym['registered'] for sym in json.loads(my_pda.info())]),
                        msg='no components of my_pda should be registered before register call')
        my_pda.register('my_pda')
        self.assertTrue(all([sym['registered'] for sym in json.loads(my_pda.info())]),
                        msg='all components of my_pda should be registered after register call')

        my_str = ak.random_strings_uniform(1, 10, UNIQUE, characters='printable')
        self.assertFalse(any([sym['registered'] for sym in json.loads(my_str.info())]),
                        msg='no components of my_str should be registered before register call')
        my_str.register('my_str')
        self.assertTrue(all([sym['registered'] for sym in json.loads(my_str.info())]),
                        msg='all components of my_str should be registered after register call')

>       my_cat = ak.Categorical(ak.array([f"my_cat {i}" for i in range(1, 11)]))

tests/registration_test.py:248:
arkouda/categorical.py:79: in __init__
    self.codes = g.broadcast(arange(self.categories.size), permute=True)
python-deps/lib/python3.7/site-packages/typeguard/__init__.py:891: in wrapper
    retval = func(*args, **kwargs)
arkouda/groupbyclass.py:970: in broadcast
    repMsg = generic_msg(cmd=cmd,args=args)
arkouda/client.py:546: in generic_msg
    return _send_string_message(cmd=cmd, args=args, recv_binary=recv_binary)

cmd = 'broadcast', recv_binary = False, args = 'id_45 id_48 id_52 True 10'

    def _send_string_message(cmd : str, recv_binary : bool=False,
                             args : str=None) -> Union[str, memoryview]:
        """
        Generates a RequestMessage encapsulating command and requesting
        user information, sends it to the Arkouda server, and returns
        either a string or binary depending upon the message format.

        Parameters
        ----------
        cmd : str
            The name of the command to be executed by the Arkouda server
        recv_binary : bool, defaults to False
            Indicates if the return message will be a string or binary data
        args : str
            A delimited string containing 1..n command arguments

        Returns
        -------
        Union[str,memoryview]
            The response string or binary data sent back from the Arkouda server

        Raises
        ------
        RuntimeError
            Raised if the return message contains the word "Error", indicating
            a server-side error was thrown
        ValueError
            Raised if the return message is malformed JSON or is missing 1..n
            expected fields
        """
        message = RequestMessage(user=username, token=token, cmd=cmd,
                              format=MessageFormat.STRING, args=args)

        logger.debug('sending message {}'.format(message))

        socket.send_string(json.dumps(message.asdict()))

        if recv_binary:
            frame = socket.recv(copy=False)
            view = frame.buffer
            if bytes(view[0:len(b"Error:")]) == b"Error:":
            return view
        else:
            raw_message = socket.recv_string()
            try:

E                   RuntimeError: UnknownSymbolError Line 225 In MultiTypeSymbolTable.checkTable: Error: lookup: unknown symbol: id_48

arkouda/client.py:349: RuntimeError
----------------------------- Captured stderr call -----------------------------
[GroupBy] Line 185 DEBUG: created id_48 int64 10 1 (10) 8 ,created id_49 int64 10 1 (10) 8
```

</details>

@reuster986 started seeing this too in https://github.com/Bears-R-Us/arkouda/pull/922.

The configs testing chapel 1.24.1 are failing, so I don't think this is from any Chapel changes and the Arkouda code didn't change for a few days around when our testing started failing so I don't think it's from changes to Arkouda either. My current guess is that this is related to upgrading to pyzmq version 22.3.0, which came out on the 16th. https://github.com/zeromq/pyzmq/pull/1590 seems like the most suspicious PR. I'm not sure if there's a bug on the pyzmq size, or if this exposed a bug on our end.

It's also not clear to me why this isn't more consistent and easier to reproduce. I can't repro locally and I can't even repro trying to set up my configuration to match the nightly testing.

For now my recommendation is to limit to pyzmq 22.2.1 to see if that cleans up nightly testing, and come back to investigate when we have more time.

ronawho:
Previously, a server would just generate names like `id_1` and use a
monotonically increasing counter so that each server instance had unique
ids. This works well for a single client/server session, but can cause
problems when a single python instance connects to multiple servers. In
this case it's possible that stale references to an old server are only
garbage-collected when connected to a new server. This can cause
accidental deletion of live arrays on the new server. This has
occasionally been an issue for users and it has been sporadically biting
us in CI testing, which took us a while to figure out.

Switch to generating a short random token to add to the ids so each
server should have a unique identifier. e.g. `id_a4e9f43_1`. This way
connections to a new server can't delete ids acquired from an old
server.

Over the past year or so we had seen sporadic unknown symbol errors in
pytests for CI testing and occasionally in nightly testing that was
caused by this. For pytests each file will start/stop the server, but
all the clients are running in the same python instance so they could
have stale references to an old server. This was sporadic because slight
changes in when the GC would run would change whether we'd see this and
changes to our dependencies or python version would also change this.

This fixes the issues and removes workarounds we had in place before.
Specifically:
 - Stop pinning to old pyzmq versions
 - Stop manually registering some arrays for a regex test
 - Stop avoid python portability testing

Resolves #694
Resolves #924


mhmerrill (approved):
looks good.

glitch (approved):
This looks good.  I'm curious if it completely solves our underlying issue / issues.

Just to try to clarify the potential issues; I think there are 2, semi-intertwined ones.
Both involve a long running client that connects / disconnects to an original sever instance and connects to a new server instance.

1. User makes pdarray A -> id_7.  User disconnects from server and connects to a new server.  Upon python garbage collection the client issues a deletion request to the new server with the old ID which DOES NOT EXIST.
2. User makes pdarray A -> id_7 on original server, disconnects & connects to a new server.  User creates more pdarrays and gets pdarray-B -> id_7 on the new server.  Python GC happens cleaning pdarrayA which issues a delete on id_7 which should have pointed to the old server; however, that id happens to exist on the new server and gets deleted.  This means a rug-pull on pdarrayB and when the user tries to use pdarrayB it gets the entry not found.

I believe this PR solves the case number 2 above, but not case number 1.

I also don't know which case was causing our problem (hopefully it was case 2, which means this would fix it :+1: )

Good work!

PR #1098, opened 2022-02-11
Add warning to String documentation

### LINKED ISSUE #1097: Update client Strings class documentation about `__del__()`
@mhmerrill and I were talking through some of the garbage collection and deletion code when we noticed Strings on the client did not implement `__del__` in the python code.  Our initial reaction was that this was an error.  However, after closer inspection we realized this was ok because we want the python Garbage collector to naturally invoke the deletion methods on the internal components.  If we override the method we are on the hook to enforce clean-up and this introduces room for error.

Since the Strings class _has a_ client side pdarray python object which already implements `__del__` it naturally cleans itself up on the server.

The Task here is to add documentation to the strings class denoting that Strings should NOT implement `__del__` as a warning to future developers.

Pl4tt:
closes #1097
(@Pl4tt the above line is what I meant)
This just adds a little warning hint to the documentation of the string class that implementing __del__() method can lead to an error

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -176,12 +176,12 @@ class Strings:
         self._regex_dict: Dict = dict()
         self.logger = getArkoudaLogger(name=__class__.__name__) # type: ignore
 
-     """
-     NOTE:
-          The Strings.__del__() method should NOT be implemented. Python will invoke the __del__() of any components by default.
-          Overriding this default behavior with an explicitly specified Strings.__del__() method may introduce errors in the event that additional components are added to Strings and the method is not updated.
-          By allowing Python's garbage collecting to handle this automatically, we avoid extra maintenance
-     """
+    """
+    NOTE:
+         The Strings.__del__() method should NOT be implemented. Python will invoke the __del__() of any components by default.
+         Overriding this default behavior with an explicitly specified Strings.__del__() method may introduce errors in the event that additional components are added to Strings and the method is not updated.
+         By allowing Python's garbage collecting to handle this automatically, we avoid extra maintenance
+    """
      
     def __iter__(self):
         raise NotImplementedError('Strings does not support iteration. To force data transfer from server, use to_ndarray')
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -55,7 +55,12 @@ class Strings:
     -----
     Strings is composed of two pdarrays: (1) offsets, which contains the
     starting indices for each string and (2) bytes, which contains the 
-    raw bytes of all strings, delimited by nulls.    
+    raw bytes of all strings, delimited by nulls.
+    
+    Warnings
+    --------
+    As causing some errors, you should not implement the __del__() method.
+    Instead strings clean up by themselves on the server.
     """
 
     BinOps = frozenset(["==", "!="])
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -60,7 +60,8 @@ class Strings:
     Warnings
     --------
     As causing some errors, you should not implement the __del__() method.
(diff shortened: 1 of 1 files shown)


mhmerrill:
@Pl4tt please link to the issue so say "closes #1097" in the description or link to the issue on the sidebar of the PR.

arkouda/strings.py
```diff
@@ -55,7 +55,12 @@ class Strings:
     -----
     Strings is composed of two pdarrays: (1) offsets, which contains the
     starting indices for each string and (2) bytes, which contains the 
-    raw bytes of all strings, delimited by nulls.    
+    raw bytes of all strings, delimited by nulls.
+    
+    Warnings
+    --------
+    As causing some errors, you should not implement the __del__() method.
+    Instead strings clean up by themselves on the server.
```
mhmerrill, line 63: this needs a little more specificity. say something like "the self.entry attribute is a parry which gets cleaned up automatically through its __del__() method" or something like that.

Pl4tt:
@mhmerrill thanks for your feedback!
Just updated the description of the warning and added "closes #1097 " to the description of the new commit :)

arkouda/strings.py
```diff
@@ -55,7 +55,13 @@ class Strings:
...
     Strings is composed of two pdarrays: (1) offsets, which contains the
     starting indices for each string and (2) bytes, which contains the 
-    raw bytes of all strings, delimited by nulls.    
+    raw bytes of all strings, delimited by nulls.
+    
+    Warnings
+    --------
+    As causing some errors, you should not implement the __del__() method.
+    Instead the self.entry attribute is a pdarray which gets cleaned up
+    automatically through its __del__() method.
```
pierce314159, line 64: I think speaking a bit more to the reason we don't want to implement a `__del__` method would be useful for any devs who are wondering why it's missing. A good potential rewording for this is something like:

```python
"""
NOTE:
  The Strings.__del__() method should NOT be implemented. Python will invoke the __del__() of any components by default.
  Overriding this default behavior with an explicitly specified Strings.__del__() method may introduce errors in the event that additional components are added to Strings and the method is not updated.
  By allowing Python's garbage collecting to handle this automatically, we avoid extra maintenance
"""
```

I also think moving this message to be right before the `__iter__` method (line 178 of strings.py) would be most useful (since this is where I would likely try to put an `__del__` method). Though I could be convinced otherwise if others have differing opinions

https://github.com/Bears-R-Us/arkouda/blob/d0fda3917835e46f3a6abe080bd8a1cc3fa377d2/arkouda/strings.py#L178-L180

pierce314159 (changes requested):
Thanks for taking this on! We love to see a PR from a first time contributor! :)

I've made a note of a potential rewording and relocation for this warning that I think would be useful

arkouda/strings.py
mhmerrill: @Pl4tt are you going to make the change so we can get your PR accepted and merged?

Pl4tt: sure, I'll do it straight away :) Thanks for your feedback!

pierce314159 (approved):
Looks good!

mhmerrill:
@pierce314159 should this be a doc string style comment??? is that what killed the CI?

pierce314159:
> @pierce314159 should this be a doc string style comment??? is that what killed the CI?

@mhmerrill and @Pl4tt, everything seems to be fine now. Looks like everything was indented by 1 space too many (mypy/docs can be finicky). I went ahead and push a commit to unindent which can be seen at https://github.com/Bears-R-Us/arkouda/pull/1098/commits/0147945d42687838736ce42a010aed7385225ae8

PR #1099, opened 2022-02-14
Create a core RadixSortLSD that can sort arbitrary data

### LINKED ISSUE #1068: Combine radixSortLSD sorting routines
Today there are 2 versions of radixSortLSD. One that just sorts the keys and ones that sorts the keys and ranks. The code is virtually the same, except for how the key is extracted from the array (just the element for keys, and the first index of a tuple for the key ranks case.)

Here's the diff between the keys and keys/ranks code:

<details>

```diff
diff --git a/sort.chpl b/sort.chpl
index c3cd535b..7bc40b84 100644
--- a/sort.chpl
+++ b/sortc.chpl
@@ -1,10 +1,9 @@
-    proc radixSortLSD_keys(a: [?aD] ?t, checkSorted: bool = true): [aD] t {
-
+    proc radixSortLSD(a:[?aD] ?t, checkSorted: bool = true): [aD] (t, int) {
         // check to see if array is already sorted
         if (checkSorted) {
             if (isSorted(a)) {
-                var sorted: [aD] t = a;
-                return sorted;
+                var keyranks: [aD] (t,int) = [i in aD] (a[i], i);
+                return keyranks;
             }
         }
 
@@ -12,8 +11,9 @@
         try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                        "type = %s nBits = %t".format(t:string,nBits));
 
-        var kr0: [aD] t = a;
-        var kr1: [aD] t;
+        // form (key,rank) vector
+        var kr0: [aD] (t,int) = [(key,rank) in zip(a,aD)] (key,rank);
+        var kr1: [aD] (t,int);
 
         // create a global count array to scan
         var gD = newBlockDom({0..#(numLocales * numTasks * numBuckets)});
@@ -41,7 +41,8 @@
                                    "locid: %t task: %t tD: %t".format(loc.id,task,tD));
                         // count digits in this task's part of the array
                         for i in tD {
-                            var bucket = getDigit(kr0[i], rshift, last, negs); // calc bucket from key
+                            const (key,_) = kr0[i];
+                            var bucket = getDigit(key, rshift, last, negs); // calc bucket from key
                             taskBucketCounts[bucket] += 1;
                         }
                         // write counts in to global counts in transposed order
@@ -85,9 +86,10 @@
                         }
                         // calc new position and put (key,rank) pair there in kr1
                         {
-                            var aggregator = newDstAggregator(t);
+                            var aggregator = newDstAggregator((t,int));
                             for i in tD {
-                                var bucket = getDigit(kr0[i], rshift, last, negs); // calc bucket from key
+                                const (key,_) = kr0[i];
+                                var bucket = getDigit(key, rshift, last, negs); // calc bucket from key
                                 var pos = taskBucketPos[bucket];
                                 taskBucketPos[bucket] += 1;
                                 aggregator.copy(kr1[pos], kr0[i]);
@@ -105,4 +107,4 @@
             }
         } // for rshift
         return kr1;
-    }//proc radixSortLSD_keys
+    }//proc radixSortLSD
```

</details>

I think we could merge the sort code into a single routine if we had an interface to extract the key. For this, I think we can copy what the Chapel sort module does -- create a comparator that provides a `key` method to extract the key. Then pass this comparator to `getBitWidth`/`getDigit`/`isSorted`.

See https://chapel-lang.org/docs/modules/packages/Sort.html#comparators (the .key method in particular) for how this is done in Chapels's sort.

ronawho:
Previously, the main differences between `RadixSortLSD_keys` and
`RadixSortLSD_ranks` were transformations on the input data and how the
key was extracted from that transformed data. For key sorting no
transformations are done and the array element is just the key. For rank
sorting a tuple of key/rank is created and the key has to extracted.

This makes the core sort agnostic to those two aspects. Now the array
transformation is done before calling sort and a comparator is used to
extract the key. Comparators are based off the ones in Chapel's sort
https://chapel-lang.org/docs/modules/packages/Sort.html#comparators

Note that this core-sort is in-place so for cases where a transformation
is not needed, a copy must be made to avoid changing the input array.

Resolves #1068

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -156,10 +156,10 @@ module RadixSortLSD
        In-place radix sort a block distributed array
        comparator is used to extract the key from array elements
      */
-    private proc radixSortLSDCore(kr0:[?aD] ?t, nBits, negs, comparator) {
+    private proc radixSortLSDCore(a:[?aD] ?t, nBits, negs, comparator) {
         try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                        "type = %s nBits = %t".format(t:string,nBits));
-        var kr1: [aD] t;
+        var temp: [aD] t;
         
         // create a global count array to scan
         var gD = newBlockDom({0..#(numLocales * numTasks * numBuckets)});
@@ -180,14 +180,14 @@ module RadixSortLSD
                         // allocate counts
                         var taskBucketCounts: [bD] int;
                         // get local domain's indices
-                        var lD = kr0.localSubdomain();
+                        var lD = a.localSubdomain();
                         // calc task's indices from local domain's indices
                         var tD = calcBlock(task, lD.low, lD.high);
                         try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                    "locid: %t task: %t tD: %t".format(loc.id,task,tD));
                         // count digits in this task's part of the array
                         for i in tD {
-                            const key = comparator.key(kr0[i]);
+                            const key = comparator.key(a[i]);
                             var bucket = getDigit(key, rshift, last, negs); // calc bucket from key
                             taskBucketCounts[bucket] += 1;
                         }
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
these changes look good to me.

Ethan-DeBandi99 (approved):
Looks good

mhmerrill (approved):
looks good to me.

PR #1100, opened 2022-02-14
Switch `SymEntry` to use const domains

### LINKED ISSUE #1083: Can `SymEntry` use const domains?
Currently, the domain in `SymEntry` is declared as `var`: https://github.com/Bears-R-Us/arkouda/blob/bafab6114c9ea05301e164b69d8fa9d3209c65ff/src/MultiTypeSymEntry.chpl#L186-L187

In Chapel 1.23 there were some optimizations made for `const` domains to avoid having to track arrays declared over them for the purpose of resizing arrays when the domain is resized (no need to track for const domains because const domains can't be resized.) This is detailed starting on slide 25 of https://chapel-lang.org/releaseNotes/1.23/04-perf-opt.pdf. 

Here I'm wondering if we can switch to use a const domain, which should have some minor performance benefits to array creation. I don't think Arkouda arrays are ever resized so I think this should be safe.

ronawho:
Arkouda arrays are never resized so there's no reason for the domains to
not be const. Making them const allows Chapel to skip some tracking code
that's used for potential array resizing, so this slightly optimizes
array creation and deletion.

For more information on this optimization see the content starting on
slide 25 of https://chapel-lang.org/releaseNotes/1.23/04-perf-opt.pdf.

Resolves #1083

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -183,7 +183,7 @@ module MultiTypeSymEntry
         'aD' is the distributed domain for 'a' whose value and type
         are defined by makeDistDom() to support varying distributions
         */
-        var aD: makeDistDom(size).type;
+        const aD: makeDistDom(size).type;
         var a: [aD] etype;
         
         /*


mhmerrill (approved):
I think this looks ok.

Issue #985, opened 2021-11-30
Tracking Parquet follow up work


bmcdonald3:
Here is a list of the Parquet follow up work that has been identified so far, items will be checked off as they are merged into master in follow up PRs. Please speak up if you have an opinion on prioritization or have identified any additional work that has been left out of this list. 

- [x] Better error handling (added in https://github.com/Bears-R-Us/arkouda/pull/993)
- [x] Improve performance for reading
    - [x] batch reading optimization (added in https://github.com/Bears-R-Us/arkouda/pull/1014)
    - [x] parallel file reading https://github.com/Bears-R-Us/arkouda/pull/1050
- [ ] Improve write performance
    - [x] batch writing https://github.com/Bears-R-Us/arkouda/pull/1028
    - [ ] parallel file writing
- [x] Writing with Snappy/RLE https://github.com/Bears-R-Us/arkouda/pull/1104
- [x] Extend to support more reading of Arrow types to pdarrays
    - [x] bool
    - [x] float
    - [x] uint32 (just need to thread through server code, C++ already has this)
    - [x] uint64 https://github.com/Bears-R-Us/arkouda/pull/1070
    - [x] timestamps https://github.com/Bears-R-Us/arkouda/pull/1024
- [x] Add capability for string reading
- [x] Support append mode
   - [x] strings
   - [x] other types
- [x] Auto detect Parquet/HDF5 file type at runtime
- [ ] Auto detect arrow/parquet support at build time

reuster986:
A couple things to add:
- [ ] Extend `ak.get_datasets()` to parquet, or add an analogous function to list the columns available for reading from a set of parquet files. I think by default only return columns that arkouda can actually read, but maybe with a keyword arg to return all existing columns?
- [ ] If no column/dset names are passed to `ak.read_parquet()`, read all supported columns by default

reuster986:
Rough priorities on the remaining components:
1. Reading strings
2. Support for `get_datasets` or equivalent
3. Read all supported columns by default
4. Reading float and bool
5. Append mode for writing
6. Auto-detect arrow/parquet

Longer-term or speculative elements (should not displace the above, but we should keep them in the back of our minds):
1. API consolidation and standardization across HDF5 and parquet
2. Support for parquet columns with nested types (like List), potentially mapping to akutil.SegArray.

pierce314159:
[...]

PR #1104, opened 2022-02-14
Add support for writing Parquet files with Snappy/RLE

bmcdonald3:
This PR adds support for writing Parquet with Snappy compression
and RLE/Bit-Packing Hybrid encoding. This can be toggled through
the `compressed` keyword argument on `pdarray.save_parquet()`.

Here are performance numbers collected on a 16-node CS:
Locales	| Current Write	| Snappy/RLE Write
------- | ------------: | ---------------:
1	| 1.00 GiB/s	| 0.36 GiB/s
16	| 7.78 GiB/s	| 5.21 GiB/s

Locales | Current Read | Snappy/RLE Read | Current 400-file Read | Snappy/RLE 400-file Read
------- | -----------: | --------------: | --------------------: | -----------------------:
1       | 1.50 GiB/s   | 0.72 GiB/s      | 8.33 GiB/s            | 8.13 GiB/s
16      | 19.12 GiB/s  | 10.40 GiB/s     | 22.73 GiB/s           | 21.74 GiB/s

So there is a rather large hit for reading/writing a single large file,
but read performance does not take much of a hit for I/O-bound reads,
such as reading in 400 files in parallel.

The current Parquet IO benchmarks continue to write without compression
to allow a more fair comparison against HDF5 and also to maintain
continuity with historical numbers for the benchmarks.

Captured in https://github.com/Bears-R-Us/arkouda/issues/985

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -1044,7 +1044,8 @@ class pdarray:
                            format(self.name, dataset, m, json_array, self.dtype)))
 
     @typechecked
-    def save_parquet(self, prefix_path : str, dataset : str='array', mode : str='truncate') -> str:
+    def save_parquet(self, prefix_path : str, dataset : str='array', mode : str='truncate',
+                     compressed : bool = False) -> str:
         """
         Save the pdarray to Parquet. The result is a collection of Parquet files,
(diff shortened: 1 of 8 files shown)


mhmerrill (approved):
looks good to me.

PR #1106, opened 2022-02-15
#1095 benchmarks for str ops with good vs poor locality

### LINKED ISSUE #1095: Add string benchmark with poor data locality
Most of the current string benchmarks have pretty uniform data distribution and locality, but this is not always the case in real world use so it'd be good to have a benchmark to track this.

A good starting point would be the code from https://github.com/Bears-R-Us/arkouda/pull/1060#issuecomment-1029450424 for hashing unsorted and sorted strings:

```python
import arkouda as ak
import numpy as np
from time import time
import sys

ak.connect(server=sys.argv[1], port=sys.argv[2])

N = 10**9
# Generate typical case
s = ak.random_strings_uniform(2, 100, N, seed=1)
t0 = time()
h1 = s.hash()
orig_time = time() - t0
# Sort by length for degenerate case
perm = ak.argsort(s.get_lengths())
ssorted = s[perm]
t0 = time()
h2 = ssorted.hash()
sorted_time = time() - t0
```

Where I think capturing the time to argsort and gather the permutation would also be worth timing. Though that's probably already captured by other benchmarks (string groupby and string gather)

Ideally we would have something like this for the regex codes too, but generating data that's useful for computing regex's that's not uniform is trickier.

reuster986:
Closes #1095 

This PR adds one benchmark script that generates 4 benchmark plots, each of which compare performance of an operation on strings with good locality vs. strings with poor locality. The operations are:
- `strings.hash()`
- `strings.substring_search()`
- `ak.cast(strings, ak.float64)`
- `strings == literal`

I did not measure the performance of sorting the lengths and permuting the generated strings because those operations are already measured by the `argsort` and `str-gather` benchmarks. However, it would be easy to capture these measurements if desired.

### DIFF
--- a/benchmarks/str-locality.py
+++ b/benchmarks/str-locality.py
@@ -19,7 +19,7 @@ def time_all_ops(N, trials, seed, correctnessOnly):
     if correctnessOnly:
         N = 10**4
     else:
-        print(">>> arkouda substring search")
+        print(">>> arkouda string locality tests")
         print("numLocales = {}, N = {:,}".format(cfg["numLocales"], N))
     random_strings, sorted_strings, perm = generate_data(N, seed)
     nbytes = random_strings.nbytes
--- a/benchmarks/str-locality.py
+++ b/benchmarks/str-locality.py
@@ -73,7 +73,7 @@ def create_parser():
     parser = argparse.ArgumentParser(description="Measure the performance of various string operations on strings with good locality (random) and poor locality (sorted).")
     parser.add_argument('hostname', help='Hostname of arkouda server')
     parser.add_argument('port', type=int, help='Port of arkouda server')
-    parser.add_argument('-n', '--size', type=int, default=10**8, help='Problem size: Number of Strings to search')
+    parser.add_argument('-n', '--size', type=int, default=10**7, help='Problem size: Number of Strings to search')
     parser.add_argument('-t', '--trials', type=int, default=1, help='Number of times to run the benchmark')
     parser.add_argument('--correctness-only', default=False, action='store_true', help='Only check correctness, not performance.')
     parser.add_argument('-s', '--seed', default=None, type=int, help='Value to initialize random number generator')
--- a/benchmarks/str-locality.py
+++ b/benchmarks/str-locality.py
@@ -14,13 +14,14 @@ def generate_data(N, seed):
     sorted_strings = random_strings[perm]
     return random_strings, sorted_strings, perm
 
-def time_all_ops(N, trials, seed, correctnessOnly):
-    cfg = ak.get_config()
+def time_all_ops(N_per_locale, trials, seed, correctnessOnly):
     if correctnessOnly:
         N = 10**4
     else:
         print(">>> arkouda string locality tests")
-        print("numLocales = {}, N = {:,}".format(cfg["numLocales"], N))
+        nl = ak.get_config()["numLocales"]
+        N = nl * N_per_locale
+        print("numLocales = {}, N = {:,}".format(nl, N))
     random_strings, sorted_strings, perm = generate_data(N, seed)
     nbytes = random_strings.nbytes
 
--- a/benchmarks/str-locality.py
+++ b/benchmarks/str-locality.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+
+import time
+import argparse
+import arkouda as ak
+
+def generate_data(N, seed):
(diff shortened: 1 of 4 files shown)


benchmarks/str-locality.py
```diff
@@ -0,0 +1,93 @@
...
+    perm = ak.argsort(random_strings.get_lengths())
+    sorted_strings = random_strings[perm]
+    return random_strings, sorted_strings, perm
+
+def time_all_ops(N, trials, seed, correctnessOnly):
+    cfg = ak.get_config()
+    if correctnessOnly:
+        N = 10**4
+    else:
+        print(">>> arkouda substring search")
```
ronawho, line 22: Not just `substring search`. Maybe `string locality tests`?

reuster986: good catch -- serves me right for copy/pasting :man_facepalming:

reuster986:
@ronawho thanks for those measurements!

> Casting is relatively slow to begin with and poor locality doesn't help. I'm unsure how much optimization can be done here or how much is worthwhile -- is this used much in practice? (edit just saw the PR to improve cast performance, let me recheck with that)

I have had a couple users asking lately for the ability to convert large string arrays to numbers, so I think it would be nice to have, although the need for it could probably be avoided with good engineering of the input data. I'm curious to see whether my PR improves things at all, although I would not expect the good-locality case to change. Intuitively, I would guess that converting a string to a number should be about as fast as a regex search (am I missing something?), so it's surprising to me that it's over two orders of magnitude slower. The heart of it is just a chapel cast -- are there any known performance pitfalls there?

> Comparing to scalar is quite slow to begin with and also slowed by poor locality. I'm not familiar with the code used to implement this, so unsure about optimization potential.

This is very old code from the dawn of string support, and there are definitely things we can do to speed it up even independent of `computeOnSegments`. I plan to tackle this next.

ronawho:
>> Casting is relatively slow to begin with and poor locality doesn't help. I'm unsure how much optimization can be done here or how much is worthwhile -- is this used much in practice? (edit just saw the PR to improve cast performance, let me recheck with that)

> I have had a couple users asking lately for the ability to convert large string arrays to numbers, so I think it would be nice to have, although the need for it could probably be avoided with good engineering of the input data. I'm curious to see whether my PR improves things at all, although I would not expect the good-locality case to change. Intuitively, I would guess that converting a string to a number should be about as fast as a regex search (am I missing something?), so it's surprising to me that it's over two orders of magnitude slower. The heart of it is just a chapel cast -- are there any known performance pitfalls there?

Performance is much, much better even for good locality in your other PR. I did not look to see why it was so slow before, but I can if we're curious. I was originally thinking this was casting to str, which can be slowed by the need to allocate, but casting strings to numeric should be quite fast on the chapel side.

>> Comparing to scalar is quite slow to begin with and also slowed by poor locality. I'm not familiar with the code used to implement this, so unsure about optimization potential.

> This is very old code from the dawn of string support, and there are definitely things we can do to speed it up even independent of computeOnSegments. I plan to tackle this next.

👍

benchmarks/str-locality.py
```diff
@@ -0,0 +1,93 @@
...
+    prefix = ak.random_strings_uniform(minlen=1, maxlen=16, size=N, seed=seed, characters='numeric')
+    if seed is not None:
+        seed += 1
+    suffix = ak.random_strings_uniform(minlen=1, maxlen=16, size=N, seed=seed, characters='numeric')
+    random_strings = prefix.stick(suffix, delimiter='.')
+    perm = ak.argsort(random_strings.get_lengths())
+    sorted_strings = random_strings[perm]
+    return random_strings, sorted_strings, perm
+
+def time_all_ops(N, trials, seed, correctnessOnly):
```
ronawho, line 17: Hmm, for most benchmarks we scale the problem size by the number of locales (`N = N_per_locale * cfg["numLocales"]`), but I'm realizing this test (and substring_search.py/flatten.py do not do that)

reuster986: Another good catch! I fixed it here, but this was actually a copy/paste from `benchmarks/substring_search.py`, so it exists there too. I will start another issue for that.

ronawho: Cool, thanks. Probably need to drop problem size down to `10**7`, at least until optimizations for all these operations go in.

reuster986: Good point. Done.

ronawho (approved):

bmcdonald3 (approved):

Issue #1101, opened 2022-02-14
Use `computeOnSegments` for operations on string bytes


reuster986:
PR #1060 introduced a way to do computations on strings that follow the locality of the bytes, rather than the segment pointers. For hashing strings, this approach has a little more overhead but significantly improves worst-case behavior, enough that we decided to switch. There are several similar cases where it might make sense to switch, so I am tracking those here. It would be good to have benchmarks like #1095 in place beforehand to establish a baseline.

- [X] `sipHash128` (#1060 )
- [ ] Regex 
  - [ ] `findMatchLocations`
  - [ ] `sub`
  - [ ] `substringSearch`
  - [ ] `peelRegex`
- [ ] `compare`
  - [ ] with scalar (while we're at it, fix compare with empty string)
  - [ ] between vectors?
- [ ] `castStringToSymEntry`

PR #1107, opened 2022-02-15
#1101 Use `computeOnSegments` for casting strings to numeric

reuster986:
This is part of #1101 and should improve performance of casting string arrays to numeric values when the locality of the string bytes is poor.

Performance is tracked in one of the benchmarks added in #1106 , but I have not yet been able to run this at a significant scale to verify the expected improvement.

While here, also add more robust error handling to str-to-num casts. It now supports three modes:
- strict (default): raise an exception if any value cannot be converted
- ignore: never raise an exception, bad inputs result in a dtype-dependent default/nan value
- return_validity: same result as 'ignore', but also return a bool array showing where the conversion succeeded.

I am not wild about adding another `Enum` for the error handling mode, because it feels rather annoying as a user. Can anyone think of a better way?

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -9,14 +9,20 @@ from arkouda.dtypes import _as_dtype
 from arkouda.pdarrayclass import pdarray, create_pdarray
 from arkouda.pdarraysetops import unique
 from arkouda.strings import Strings
+from enum import Enum
 
 Categorical = ForwardRef('Categorical')
 
 __all__ = ["cast", "abs", "log", "exp", "cumsum", "cumprod", "sin", "cos", 
-           "hash", "where", "histogram", "value_counts", "isnan"]
+           "hash", "where", "histogram", "value_counts", "isnan", "ErrorMode"]
+
+class ErrorMode(Enum):
+    strict = 'strict'
+    ignore = 'ignore'
+    return_validity = 'return_validity'
 
 @typechecked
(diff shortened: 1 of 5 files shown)


ronawho:
16-node-cs-hdr performance:

Before:
```
Casting good locality Average time = 22.1167 sec
Casting good locality Average rate = 0.0758 GiB/sec
Casting poor locality Average time = 45.3778 sec
Casting poor locality Average rate = 0.0369 GiB/sec
```


Now:
```
Casting good locality Average time = 0.0950 sec
Casting good locality Average rate = 17.6440 GiB/sec
Casting poor locality Average time = 0.2211 sec
Casting poor locality Average rate = 7.5834 GiB/sec
```

In table form:

| config        | Before     | Now        |
| ------------- | ---------: | ---------: |
| good locality | 0.07 GiB/s | 17.6 GiB/s |
| poor locality | 0.03 GiB/s |  7.5 GiB/s |

So great improvements for good and poor locality.

reuster986:
Wow, cool! I will have to go back and rationalize that speedup to myself, because I didn't expect it, but will definitely take it :)

Thanks again for running the measurement!

pierce314159:
The only idea I have to avoid the `Enum` would be to have a separate cast function to handle the `ignore` and `return_validity` cases. So `cast` would handle the strict case and `relaxed_cast(return_validity = False)` would handle the other two. It's hard to tell if this would be more or less annoying though

ronawho (approved):

bmcdonald3 (approved):

pierce314159 (approved):
Looks good to me!

PR #1111, opened 2022-02-15
Closes #1088 - Remove "legacy_placeholder" from String.get_lengths()

### LINKED ISSUE #1088: Remove "legacy_placeholder" from Strings.get_lengths() function
### Background
As part of #786 Strings objects which consist of two pdarrays (segments and values aka offsets and bytes) were internalized on the server as a single "Entry" in the symbol table; previously they were two loosely coupled independent entries.  In an effort to mitigate risk from such a big change, the message formats were left intact.  The original implementation needed to pass two id's (for each pdarray component) but the current implementation only has one.  A "legacy_placeholder" string literal was put in place for the second id across all of the string messages on both the client & server.  A number of these exist, but this ticket will seek to only address a single end-to-end case to make it an easier task for a new Arkouda dev to gain experience working in both parts of the codebase.

### Task
Remove the "legacy_placeholder" id from both the client & server for the [Strings.get_lengths ](https://github.com/Bears-R-Us/arkouda/blob/master/arkouda/strings.py#L309) function.

In general you will need to find the message / cmd name used by the client & server and find the corresponding server side Chapel procedure.  The Client message arguments will need to be shortened appropriately and the server side processing of the message will need to be modified to correspond to the message change.

### Files of interest

- [strings.py](https://github.com/Bears-R-Us/arkouda/blob/master/arkouda/strings.py)
- [SegmentedMsg.chpl](https://github.com/Bears-R-Us/arkouda/blob/master/src/SegmentedMsg.chpl)

### Testing
Unit tests for strings.py::Strings::get_lengths() should already be in place (i.e. `tests/string_test/StringTest.test_get_lengths`) so they should fail/pass based on the changes.

Ethan-DeBandi99:
This PR (closes #1088):

- Removed legacy_placeholder from String.get_lengths() and SegmentedMsg.segmentLengthMsg().

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -321,7 +321,7 @@ class Strings:
             Raised if there is a server-side error thrown
         """
         cmd = "segmentLengths"
-        args = "{} {} {}".format(self.objtype, self.entry.name, "legacy_placeholder")
+        args = "{} {}".format(self.objtype, self.entry.name)
         return create_pdarray(generic_msg(cmd=cmd,args=args))
 
     @typechecked
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -130,15 +130,15 @@ module SegmentedMsg {
   proc segmentLengthsMsg(cmd: string, payload: string, 
                                           st: borrowed SymTab): MsgTuple throws {
     var pn = Reflection.getRoutineName();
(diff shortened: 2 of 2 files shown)


pierce314159 (approved):
Looks great!

mhmerrill (approved):
looks good.

glitch (approved):

Issue #1039, opened 2022-01-20
Update peel/stick to use aggregation


ronawho:
Noted in https://github.com/Bears-R-Us/arkouda/issues/762#issuecomment-820621005 and https://github.com/Bears-R-Us/arkouda/issues/222#issuecomment-809465943f, but peel/stick is the only remaining code that uses `unorderedCopy`, which will perform poorly everywhere except Cray XCs.

We should convert these to use aggregation. Note that peel/stick are still listed as experimental (but maybe that could be removed if performance is improved.) Note that the flatten benchmarks creates some input arrays using peel/stick, and this is one of the things preventing us from bumping the problem size for that benchmark up from the current `10**6` (normal benchmark problem size is `10**8`)

PR #1112, opened 2022-02-15
Closes #1109: Scale `substring_search.py` and `flatten.py` benchmarks with `numLocales`

### LINKED ISSUE #1109: Should substring search benchmark scale by numLocales?
Most benchmarks treat the `--size` argument as a per-locale size and multiply by `numLocales` when generating the test data. However, `benchmarks/substring_search.py` does not scale by `numLocales`:

https://github.com/Bears-R-Us/arkouda/blob/d0fda3917835e46f3a6abe080bd8a1cc3fa377d2/benchmarks/substring_search.py#L8-L14

Was this intentional, or should we update it?

pierce314159:
This PR (closes #1109):
- Corrects oversight which kept the `substring_search.py` and `flatten.py` benchmarks from scaling with `numLocales` (as the other benchmarks do)
- Changes the `size` variable to `N_per_locale` in `str-in1d.py` to be consistent with other benchmarks
- Changes the default problem size for flatten from `10**6` to `10**5`

### DIFF
--- a/benchmarks/flatten.py
+++ b/benchmarks/flatten.py
@@ -5,9 +5,10 @@ import argparse
 import arkouda as ak
 
 
-def time_flatten(N, trials):
+def time_flatten(N_per_locale, trials):
     print(">>> arkouda flatten")
     cfg = ak.get_config()
+    N = N_per_locale * cfg["numLocales"]
     print("numLocales = {}, N = {:,}".format(cfg["numLocales"], N))
 
     thirds = [ak.cast(ak.arange(i, N*3, 3), 'str') for i in range(3)]
@@ -67,7 +68,7 @@ def create_parser():
     parser = argparse.ArgumentParser(description="Measure the performance of regex and non-regex flatten on Strings.")
     parser.add_argument('hostname', help='Hostname of arkouda server')
     parser.add_argument('port', type=int, help='Port of arkouda server')
-    parser.add_argument('-n', '--size', type=int, default=10**6, help='Problem size: Number of Strings to flatten')
+    parser.add_argument('-n', '--size', type=int, default=10**5, help='Problem size: Number of Strings to flatten')
     parser.add_argument('-t', '--trials', type=int, default=1, help='Number of times to run the benchmark')
     parser.add_argument('--correctness-only', default=False, action='store_true', help='Only check correctness, not performance.')
     return parser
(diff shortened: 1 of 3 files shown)


benchmarks/flatten.py
```diff
@@ -5,9 +5,10 @@
 import arkouda as ak
 
 
-def time_flatten(N, trials):
+def time_flatten(size, trials):
     print(">>> arkouda flatten")
     cfg = ak.get_config()
+    N = size * cfg["numLocales"]
```
ronawho, line 11: Flatten is pretty slow today (partially the kernel, but mostly the input array creation using `stick` (https://github.com/Bears-R-Us/arkouda/issues/1039)).

Think we'll need to drop the problem size from `10**6` to `10**5` for flatten for now.

benchmarks/flatten.py
```diff
@@ -5,9 +5,10 @@
 import arkouda as ak
 
 
-def time_flatten(N, trials):
+def time_flatten(size, trials):
```
ronawho, line 8: Nit, but the other benchmarks call this `N_per_locale`

pierce314159: oooh yeah i was looking at `Str-in1d`. But it does look like most of the others do `N_per_locale`

benchmarks/array_transfer.py
```diff
@@ -7,9 +7,10 @@
 
 TYPES = ('int64', 'float64')
 
-def time_ak_array_transfer(N, trials, dtype, random, seed):
+def time_ak_array_transfer(N_per_locale, trials, dtype, random, seed):
     print(">>> arkouda {} array creation".format(dtype))
     cfg = ak.get_config()
+    N = N_per_locale * cfg["numLocales"]
```
ronawho, line 13: This one is actually meant to be N and not scale with the number of locales. (It's transferring between the single client and server, so we're limited by single client node.)

pierce314159: ohhhh gotcha! thanks, I'll fix this

ronawho (approved):

Ethan-DeBandi99 (approved):
I do not see any issues.

reuster986 (approved):

glitch (approved):

PR #1113, opened 2022-02-15
Closes #1105 - Remove Assertion test from Error Message

### LINKED ISSUE #1105: Remove `AssertEqual` for exact wording of error messages
I've seen this come up a few times and just encountered it again with @Ethan-DeBandi99.

There are few cases where we are checking the exact wording of an error message and sometimes this causes test failures when there's nothing wrong. This often happens when different yet functionally equivalent types are returned (`int64` vs `np.int64`). An example from `numeric_test.py` that I just saw with Ethan is:

https://github.com/Bears-R-Us/arkouda/blob/d0fda3917835e46f3a6abe080bd8a1cc3fa377d2/tests/numeric_test.py#L78-L81

Here we check that:
- `ak.histogram(pda, bins='1')` raises a `TypeError`
- the error message is exactly `'type of argument "bins" must be one of (int, int64, uint64); got str instead'`

With Ethan the error message differed slightly (`'type of argument "bins" must be one of (int, np.int64, uint64); got str instead'`) which caused the test to fail

I think checking the kind of error is sufficient and the exact wording of the error message is not needed. Unless there's some reason to keep these around that I'm not seeing, I suggest we remove all checks of this form from `tests`

```python
self.assertEqual('exact_wording_of_error',  cm.exception.args[0]) 
```

Ethan-DeBandi99:
This PR (closes #1105):

-Removed Assert tests checking for specific wording within error messages from `tests`

### DIFF
--- a/tests/categorical_test.py
+++ b/tests/categorical_test.py
@@ -38,8 +38,6 @@ class CategoricalTest(ArkoudaTest):
         
         with self.assertRaises(ValueError) as cm:
             ak.Categorical(ak.arange(0,5,10))
-        self.assertEqual('Categorical: inputs other than Strings not yet supported', 
-                         cm.exception.args[0])        
         
     def testCategoricalFromCodesAndCategories(self):
         codes = ak.array([7,5,9,8,2,1,4,0,3,6])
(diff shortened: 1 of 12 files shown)


pierce314159 (approved):
Looks good!

mhmerrill (approved):
looks ok to me.

PR #1115, opened 2022-02-15
1114 enable global server in pytest

hokiegeek2:


ronawho:
At least for CI and nightly testing, `CLASS_SERVER` is probably our preferred mode. The startup overhead is negligible for us, and the benefit to isolating errors down to a single class when something kills the server is beneficial.

glitch:
Looking at your commits, it seems your remote is still pointing at -> https://github.com/mhmerrill/arkouda
You should update that to Bears-R-Us/arkouda
If this gets squash merged I don't _think_ it would be an issue, but it may do something... _hokie_  (cue the CSI-Miami sunglasses meme)

mhmerrill:
@hokiegeek2 after 3+ hours I cancelled the CI runs for this PR, I am thinking there is something wrong... maybe. also update your `origin` to Bears-R-Us. ;-)

glitch:
Hmm... If I can hazard a guess, I bet the server went down on one of the tests, and the ZMQ client is waiting to reconnect which results in a test hanging.

hokiegeek2:
> Hmm... If I can hazard a guess, I bet the server went down on one of the tests, and the ZMQ client is waiting to reconnect which results in a test hanging.

Good point, seen that before, but it looks like the tests never got going:

image

hokiegeek2:
@mhmerrill argh, sorry, forgot to commit new file that started the test harness in GLOBAL_SERVER mode. Added that file and pulled in latest upstream master changes, CI running now.

hokiegeek2:
Okay, this looks more-better:

image

The issue was I forgot to commit tests/conftest.py. I added that and the tests run fine.

hokiegeek2:
> @hokiegeek2 after 3+ hours I cancelled the CI runs for this PR, I am thinking there is something wrong... maybe. also update your `origin` to Bears-R-Us. ;-)

@mhmerrill also, fixed my remote upstream on my laptop, will be reflected in my next commit

image

hokiegeek2:
all CI steps passed, ready for review

pytest.ini
```diff
@@ -34,7 +34,7 @@ python_functions = test*
 env =
     D:ARKOUDA_SERVER_HOST=localhost
     D:ARKOUDA_SERVER_PORT=5555
-    D:ARKOUDA_FULL_STACK_TEST=True
+    D:ARKOUDA_RUNNING_MODE=GLOBAL_SERVER
```
ronawho, line 37: At least for CI and nightly testing, `CLASS_SERVER` is probably our preferred mode. The startup overhead is negligible for us, and the benefit to isolating errors down to a single class when something kills the server is beneficial.

My preference would be to leave `CLASS_SERVER` the default for that reason, but if others prefer `GLOBAL_SERVER`, I can update CI and nightly to override.

hokiegeek2: @ronawho I apologize, I am not following what you're writing here. The test cases are executed sequentially, so not sure how GLOBAL_SERVER vs CLASS_SERVER impacts tracing down a test case failure.

ronawho: If the first test class does something that kills the server we lose the remaining testing. For example https://github.com/Bears-R-Us/arkouda/issues/1093 was causing failures in nightly testing for 3 or 4 different classes, but in the `GLOBAL_SERVER` world we only would have seen the first real failure and all other tests would have failed because the server wasn't running.

hokiegeek2: @ronawho ah, I see, yes, you are correct

mhmerrill: I have two questions:

1. Do we still have a mode to connect to an already running server?
2. Are we going to make the default `CLASS_SERVER`?

hokiegeek2: @mhmerrill made the changes, pulled in upstream master, built/tested pushed to PR, CI successful, ready for review

mhmerrill (approved):
looks ok to me.

Ethan-DeBandi99 (approved):
Looks good

PR #1116, opened 2022-02-15
Use `computeOnSegments` for string comparisons

reuster986:
This is part of https://github.com/Bears-R-Us/arkouda/issues/1101 and should improve performance of comparing string arrays to string literals or other string arrays when the locality of the string bytes is poor.

Performance is tracked in one of the benchmarks added in https://github.com/Bears-R-Us/arkouda/pull/1106 , but I have not yet tested at scale.

This PR includes the commits in #1107 , which has not yet been merged. My git-fu is poor, so I'm not sure if there is a better way to extend a pending PR or what I should do if/when it gets merged.

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -1044,6 +1044,11 @@ module SegmentedArray {
                            moduleName = getModuleName(),
                            errorClass="ArgumentError");
     }
+    if useHash {
+      const lh = lss.hash();
+      const rh = rss.hash();
+      return if polarity then (lh == rh) else (lh != rh);
+    }
     ref oD = lss.offsets.aD;
     // Start by assuming all elements differ, then correct for those that are equal
     // This translates to an initial value of false for == and true for !=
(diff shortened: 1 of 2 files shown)


ronawho:
> This PR includes the commits in https://github.com/Bears-R-Us/arkouda/pull/1107 , which has not yet been merged. My git-fu is poor, so I'm not sure if there is a better way to extend a pending PR or what I should do if/when it gets merged.

This is typically the way I open PRs that are dependent or based on each other. After #1107 is merged I would then do something like:

```
git checkout master
git pull origin master
git checkout str-compare
git rebase -i master
git push origin str-compare --force
```

Where basically this is checking out master, pulling in the changes from 1107, doing an interactive rebase of this branch to master with 1107 included, and then force pushing up to this PR. Rebase and force push is generally frowned upon for things that are already committed to a repo, but in general is fine for touching branches only you're using (rebasing and force pushing rewrites history, so is frowned upon if other people already have that history.) If you don't want to rebase/force-push, alternatively I think you could do `git merge master` instead of rebase but that's not a workflow I'm familiar with offhand.

pierce314159 (approved):
This all looks good to me!
I just looked at the last two commits since the first two are in 1107

also 👍 for what @ronawho suggested for updating this once 1107 is merged. I do the same rebase on top of master and force push workflow

Ethan-DeBandi99 (approved):

ronawho (approved):

PR #1117, opened 2022-02-16
Closes #1108 Update README.md Apple Instructions

### LINKED ISSUE #1108: Document Apple M1 Install Variations
While configuring my environment, a few items with the Apple MacOS install were not accurate for the new M1 processors. Additionally, some of the directions are out of date. The following points need to be documented in the README.

- Apple MacBooks running the M1, M1 Pro, or M1 Max processors are not compatible with the full install of Anaconda. Currently, the only distribution configured to run is miniforge.
- Miniforge will require /etc/paths to be updated for python3 to load properly
- Chapel does not require a make when installing via Homebrew

Ethan-DeBandi99:
This PR (closes #1108):

Updates the MacOS Environment section of the README.md to account for variances when equipped with Apple Silicon. Instructions should now reflect both Intel (x86) chips and the new Apple Silicon (arm64) chips. 

The install has been broken out into 2 pieces

- Configure Using Homebrew
- Configure Manually

Doing this provides full install instructions based on the method chosen. @pierce314159 and I thought this was more intuitive. Each Section details out how to install Chapel and configure the Python Environment using the respective method. Additionally, anything that is different on Macs with Apple Silicon is noted. Looking for feedback on the organization structure.

I included both miniforge and Anaconda3 as options for configuring Anaconda. Miniforge is compatible with all chipsets, but does not include as many packages out of the box. I would recommend removing Anaconda3 as an option since it is limited (although it will most likely support Apple Silicon in the near future). It is worth noting the Miniconda is not compaitble with the arm64 architecture yet either.

Specific links to the installs are included for easily accessing the correct installs based on chipset, but this may be overkill. We should probably at least indicate the correct install, though they may change as Anaconda/Python are updated in the future. It may be better to provide the base link to the installs and and detail out the naming conventions that indicate which install is appropriate for a given hardware set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,8 +91,8 @@ This yielded a >20TB dataframe in Arkouda.
 1. [Prerequisites](#prereq-main)
    - [Requirements](#prereq-reqs)
    - [MacOS](#prereq-mac)
-     - [Installing Chapel](#prereq-mac-chapel)
-     - [Python environment - Anaconda](#prereq-mac-anaconda)
+     - [Configure with Homebrew](#prereq-mac-brew)
+     - [Configure Manually](#prereq-mac-manual)
    - [Linux](#prereq-linux)
      - [Install Chapel](#prereq-linux-chapel)
      - [Python environment - Anaconda](#prereq-linux-anaconda)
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
I think this looks good.

PR #1120, opened 2022-02-17
Make some minor optimizations to RadixSortLSD

### LINKED ISSUE #1119: Make some micro-optimizations to sort
https://github.com/Bears-R-Us/arkouda/pull/1099 hurt sort performance slightly because it added an extra whole array copy on the last iteration. Should be able to get rid of that and while there make a few other small improvements (avoid extra array for scans, use localAccess in some places)

ronawho:
While cleaning up the sort implementation in #1099 I saw a few minor
optimization opportunities. They won't have a major impact, but given
the importance of sort it seemed worthwhile.

Specifically this:
- Avoids an extra copy for sort. Part of 1099 added an extra copy
  between the input array and the temp array. Remove that here by
  switching the order the arrays are used so we end up PUT'ing sorted
  results into the input/return array instead of the temp one.
- Eliminates a copy for the scan on globalStarts in sort. Today scan
  always creates a new array and returns it. By assigning the result of
  a scan into an existing array we were just adding an array copy and
  not eliminating an allocation. Switch to just assigning into a new
  variable, which will steal the returned buffer and avoid the initial
  allocation for globalStarts and the assignment to it during scans.
  Longer term, we have discussed having a way to provide the output
  array for scans, but there's no timeframe for that.
- Uses `localAccess` for some array accesses in sort. `localAccess` is a
  minor optimization that avoids some locality checks when array
  accesses are local. The performance benefit is small, but not trivial
  and given the importance of sort it seemed worthwhile.

Resolves #1119

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -159,7 +159,7 @@ module RadixSortLSD
     private proc radixSortLSDCore(a:[?aD] ?t, nBits, negs, comparator) {
         try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                        "type = %s nBits = %t".format(t:string,nBits));
-        var temp: [aD] t;
+        var temp = a;
         
         // create a global count array to scan
         var gD = newBlockDom({0..#(numLocales * numTasks * numBuckets)});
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
I think this looks good.

PR #1123, opened 2022-02-17
Closes #1110 legacy_placeholder removal

### LINKED ISSUE #1110: Remove "legacy_placeholder" from remaining locations
Background
As part of [786](https://github.com/Bears-R-Us/arkouda/issues/786) Strings objects which consist of two pdarrays (segments and values aka offsets and bytes) were internalized on the server as a single "Entry" in the symbol table; previously they were two loosely coupled independent entries. In an effort to mitigate risk from such a big change, the message formats were left intact. The original implementation needed to pass two id's (for each pdarray component) but the current implementation only has one. A "legacy_placeholder" string literal was put in place for the second id across all of the string messages on both the client & server. 
[1088](https://github.com/Bears-R-Us/arkouda/issues/1088) handles the removal of these from the String.get_lengths() function. This issue is to remove the remaining references found in the code.

**Task**
Remove the "legacy_placeholder" id from both the client & server.

In general you will need to find the message / cmd name used by the client & server and find the corresponding server side Chapel procedure. The Client message arguments will need to be shortened appropriately and the server side processing of the message will need to be modified to correspond to the message change.

Ethan-DeBandi99:
This PR (close #1110):

The remaining files have been updated to remove legacy_placeholder both from server and client. The TODO tags indicating locations where changes were required have been removed. String formatting updated to reflect the removals.

### DIFF
--- a/arkouda/match.py
+++ b/arkouda/match.py
@@ -134,12 +134,12 @@ class Match:
         from arkouda.strings import Strings
         cmd = "segmentedFindAll"
         args = "{} {} {} {} {} {} {}".format(self._objtype,
-                                                self._parent_entry_name,
-                                                self._matched.name,
-                                                self._starts.name,
-                                                self._lengths.name,
-                                                self._indices.name,
-                                                return_match_origins)
+                                             self._parent_entry_name,
+                                             self._matched.name,
+                                             self._starts.name,
+                                             self._lengths.name,
+                                             self._indices.name,
+                                             return_match_origins)
         repMsg = cast(str, generic_msg(cmd=cmd, args=args))
         if return_match_origins:
             arrays = repMsg.split('+', maxsplit=2)
@@ -188,9 +188,9 @@ class Match:
         # We don't cache the locations of groups, find the location info and call findAll
         cmd = "segmentedFindLoc"
         args = "{} {} {} {}".format(self._objtype,
-                                       self._parent_entry_name,
-                                       group_num,
-                                       json.dumps([self.re]))
+                                    self._parent_entry_name,
+                                    group_num,
+                                    json.dumps([self.re]))
         repMsg = cast(str, generic_msg(cmd=cmd, args=args))
         created_map = json.loads(repMsg)
         global_starts = create_pdarray(created_map["Starts"])
@@ -211,12 +211,12 @@ class Match:
         lengths = global_lengths[global_indices[matched]]
         cmd = "segmentedFindAll"
         args = "{} {} {} {} {} {} {}".format(self._objtype,
-                                                self._parent_entry_name,
-                                                matched.name,
-                                                starts.name,
-                                                lengths.name,
-                                                indices.name,
-                                                return_group_origins)
+                                             self._parent_entry_name,
+                                             matched.name,
+                                             starts.name,
+                                             lengths.name,
+                                             indices.name,
+                                             return_group_origins)
         repMsg = cast(str, generic_msg(cmd=cmd, args=args))
         if return_group_origins:
             arrays = repMsg.split('+', maxsplit=2)
--- a/arkouda/match.py
+++ b/arkouda/match.py
@@ -133,9 +133,8 @@ class Match:
         """
         from arkouda.strings import Strings
         cmd = "segmentedFindAll"
-        args = "{} {} {} {} {} {} {} {}".format(self._objtype,
+        args = "{} {} {} {} {} {} {}".format(self._objtype,
                                                 self._parent_entry_name,
-                                                "legacy_placeholder",
                                                 self._matched.name,
                                                 self._starts.name,
                                                 self._lengths.name,
@@ -188,9 +187,8 @@ class Match:
 
         # We don't cache the locations of groups, find the location info and call findAll
         cmd = "segmentedFindLoc"
-        args = "{} {} {} {} {}".format(self._objtype,
+        args = "{} {} {} {}".format(self._objtype,
                                        self._parent_entry_name,
-                                       "legacy_placeholder",
                                        group_num,
                                        json.dumps([self.re]))
         repMsg = cast(str, generic_msg(cmd=cmd, args=args))
@@ -212,9 +210,8 @@ class Match:
         starts = global_starts[global_indices[matched]]
         lengths = global_lengths[global_indices[matched]]
         cmd = "segmentedFindAll"
-        args = "{} {} {} {} {} {} {} {}".format(self._objtype,
+        args = "{} {} {} {} {} {} {}".format(self._objtype,
                                                 self._parent_entry_name,
-                                                "legacy_placeholder",
                                                 matched.name,
                                                 starts.name,
                                                 lengths.name,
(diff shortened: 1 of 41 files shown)


src/ConcatenateMsg.chpl
```diff
@@ -55,8 +55,7 @@ module ConcatenateMsg
             var valSize: int;
             select objtype {
                 when "str" {
-                    var legacy_placeholder: string;
-                    (name, legacy_placeholder) = rawName.splitMsgToTuple('+', 2);
+                    name = rawName;
```
pierce314159, line 58: I think it's cleaner to just change the looping variable from `rawName` to `name` (on line 53) to be something like
```C
for (name, i) in zip(names, 1..) {
```
and avoid the extra variable

src/FindSegmentsMsg.chpl
```diff
@@ -132,11 +132,9 @@ module FindSegmentsMsg
               [(u, s, i) in zip(ukeylocs, permKey, paD)] if ((i > paD.low) && (permKey[i-1] != s))  { u = true; }
             }
             when "str" {
-              // TODO remvove legacy_placeholder
-              var (myNames1,legacy_placeholder) = name.splitMsgToTuple('+', 2);
               fsLogger.info(getModuleName(),getRoutineName(),getLineNumber(),
-                                  "findSegmentsMessage myNames1: {} legacy_placeholder: {}".format(myNames1,legacy_placeholder));
-              var str = getSegString(myNames1, st);
+                                  "findSegmentsMessage myNames1: {}".format(name));
```
pierce314159, line 136: the logging message should be updated to read `"findSegmentsMessage name: {}".format(name)` to account for the change from `myNames1` to `name`

test/UnitTestPeelStick.chpl
```diff
@@ -133,19 +133,18 @@ proc testMessageLayer(substr, n, minLen, maxLen) throws {
...
   var roname = parseName(roAttribs);
-  reqMsg = "stick str %s %s str %s %s False %jt".format(loname, "legacy_placeholder", roname, "legacy_placeholder", [""]);
+  reqMsg = "stick str %s str %s False %jt".format(loname, roname, [""]);
   writeReq(reqMsg);
   repMsg = segBinopvvMsg(cmd="segBinopvv", payload=reqMsg, st).msg;
   writeRep(repMsg);
-  //TODO remove legacy_placeholder
-  var (rtoAttribs,legacy_placeholder) = repMsg.splitMsgToTuple('+', 2);
+  var rtoAttribs = repMsg;
   var rtoname = parseName(rtoAttribs);
```
pierce314159, line 148: I think we could just call `parseName(repMsg)` and not declare an extra var

pierce314159 (commented):
I made some small notes but I think this looks good overall

arkouda/pdarraysetops.py
```diff
@@ -64,31 +66,31 @@ def unique(pda : Union[pdarray,Strings,'Categorical'], # type: ignore
...
     elif isinstance(pda, pdarray):
-        repMsg = generic_msg(cmd="unique", args="{} {} {}".\
+        repMsg = generic_msg(cmd="unique", args="{} {} {}". \
                              format(pda.objtype, pda.name, return_counts))
         if return_counts:
-            vc = cast(str,repMsg).split("+")
+            vc = cast(str, repMsg).split("+")
             logger.debug(vc)
-            return create_pdarray(cast(str,vc[0])), create_pdarray(cast(str,vc[1]))
+            return create_pdarray(cast(str, vc[0])), create_pdarray(cast(str, vc[1]))
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I believe PyCharm probably did do this. I will make sure to note this for the future.

arkouda/match.py
```diff
@@ -133,9 +133,8 @@ def find_matches(self, return_match_origins: bool = False):
         """
         from arkouda.strings import Strings
         cmd = "segmentedFindAll"
-        args = "{} {} {} {} {} {} {} {}".format(self._objtype,
+        args = "{} {} {} {} {} {} {}".format(self._objtype,
                                                 self._parent_entry_name,
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I updated the alignments for these cases, so all the indentation should now be on the same level.

arkouda/sorting.py
```diff
@@ -57,7 +57,7 @@ def argsort(pda : Union[pdarray,Strings,'Categorical'], algorithm : SortingAlgor
     if pda.size == 0 and hasattr(pda, "dtype"):
         return zeros(0, dtype=pda.dtype)
     if isinstance(pda, Strings):
-        name = '{}+{}'.format(pda.entry.name, "legacy_placeholder")
+        name = '{}'.format(pda.entry.name)
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I updated this to use the ternary.

src/FindSegmentsMsg.chpl
Ethan-DeBandi99: Completely missed that. It has been updated.

src/ConcatenateMsg.chpl
Ethan-DeBandi99: Extra var has been removed.

arkouda/matcher.py
```diff
@@ -44,9 +44,8 @@ def find_locations(self) -> None:
         sym_tab = list_symbol_table()
         if not self.populated or any([getattr(self, pda).name not in sym_tab for pda in self.LocationsInfo]):
             cmd = "segmentedFindLoc"
-            args = "{} {} {} {} {}".format(self.objtype,
+            args = "{} {} {} {}".format(self.objtype,
                                            self.parent_entry_name,
-                                           "legacy_placeholder",
                                            0,  # groupNum is 0 for regular matches
                                            json.dumps([self.pattern]))
```
pierce314159, line 50: (1/2) there's still two places where the indentation got overlooked

arkouda/strings.py
```diff
@@ -694,9 +689,8 @@ def contains(self, substr: Union[bytes, str_scalars], regex: bool = False) -> pd
         if matcher is not None:
             return matcher.get_match(MatchType.SEARCH, self).matched()
         cmd = "segmentedSearch"
-        args = "{} {} {} {} {}".format(self.objtype,
+        args = "{} {} {} {}".format(self.objtype,
                                        self.entry.name,
-                                       "legacy_placeholder",
                                        "str",
                                        json.dumps([substr]))
```
pierce314159, line 695: (2/2) indentation got overlooked

pierce314159 (approved):
I think this is good to go! 

I had 2 comment but it's just nitpicking formatting stuff

arkouda/strings.py
Ethan-DeBandi99: Corrected the spacing

arkouda/sorting.py
```diff
@@ -126,7 +123,7 @@ def coargsort(arrays: Sequence[Union[Strings, pdarray, 'Categorical']], algorith
             atypes.append(a.objtype)
         elif isinstance(a, Strings):
             atypes.append(a.objtype)
-            anames.append('{}+{}'.format(a.entry.name, "legacy_placeholder"))
+            anames.append(a.entry.name)
```
glitch, line 126: No change requested: Just thinking down the line, I wonder if adding a `getName()` function to `pdarray` and `Strings` would simplify some of the casting / conditional logic.  In theory it would start to collapse to `pda.getname()`

glitch (approved):
Looks good, nice job!

PR #1125, opened 2022-02-18
Cleanup the int in1d implementation

ronawho:
Make a few implementation cleanups to in1d. There may be a small
performance benefit as well, but most of the time is spent sorting so
nothing major. Cleanups are:
 - Add a helper function to concatenate arrays for `in1dSort`
 - Drop the unused `counts` from `uniqueSort*` in `in1dSort`
 - Simplify the check for duplicates in` in1dSort`
 - Add in1d wrapper to select strategy instead of choosing at each callsite

Part of #970

### DIFF
--- a/src/ArraySetops.chpl
+++ b/src/ArraySetops.chpl
@@ -13,6 +13,7 @@ module ArraySetops
     use RadixSortLSD;
     use Unique;
     use Indexing;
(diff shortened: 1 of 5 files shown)


reuster986 (approved):
nice!

PR #1131, opened 2022-02-22
Tune int in1d threshold

ronawho:
Tune the threshold for selecting in1d implementations. Even with recent
optimizations, the threshold for selecting between the set and sort
based strategies is a little too high. Drop from `2**25` down to
`2**23`, which is pretty good on 16-node-cs-hdr, 16-node-xc, and
512-node-xc. The ideal threshold depends on core count, node count, and
network bandwidth, but I think this will work well in practice. For some
configs `2**24` may be a little better, but we I'd prefer to stick lower
since we know the sort based methods scales well, it just has high
startup overhead.

Part of #970

### DIFF
--- a/benchmarks/in1d.py
+++ b/benchmarks/in1d.py
@@ -5,10 +5,10 @@ import arkouda as ak
 
 TYPES = ('int64', 'uint64')
 
-# Must be less than src/In1dMsg.chpl:mbound, which defaults to 2**25
-MEDIUM = 2**25 - 1
-# Must be greater than mbound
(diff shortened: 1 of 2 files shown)


ronawho:
I'd probably wait a day or two before merging to allow nightly perf testing on #1125 to run.

reuster986 (approved):

ronawho:
No major perf changes from https://github.com/Bears-R-Us/arkouda/pull/1125 (expected), so I think this can be merged.

PR #1132, opened 2022-02-22
Use int in1d to implement the hashed based string in1d

ronawho:
The implementation was similar, so take advantage of optimizations and
tuning done for ints instead of duplicating the logic for strings.

Part of #970


ronawho:
Drafted since it's based on #1131, but I think this is ready to review.

reuster986 (approved):
good call on reusing the in1d code

Issue #1008, opened 2021-12-20
Multi-dimensional Array View


pierce314159:
Create an `ArrayView` allowing pdarrays to be viewed/indexed in a multidimensional way. The underlying pdarray will still be stored in 1d

These `ArrayView`s should be constructed using:
- `pdarray.reshape`
  -  i.e. `ak.arange(27).reshape((3, 3, 3))`
- `ak.array`
  -  i.e. `ak.array([[0, 0], [0, 1], [1, 1]])`

This is how numpy handles N-dimensional arrays ([`ndarray`](https://numpy.org/doc/stable/reference/arrays.ndarray.html))

This functionality has been discussed for a while and may be related to other issues (such as #39, #462, #702)

pierce314159:
I think I'm going to go ahead and close this since the initial functionality has been implemented. New issues can be made to track the additional functionality

PR #1142, opened 2022-02-23
Initial Multi-dim ArrayView

pierce314159:
Part of Issue #1008:
- Introduces `ArrayView`
- Creation using `ak.array` or `pdarray.reshape`
  - `ak.array([[0, 0], [0, 1], [1, 1]])`
  - `ak.arange(27).reshape(3, 3, 3)`
- Handles `row_major`/`C` order and `column_major`/`F`
-  Support integer and slice indexing for `__getitem__`
- Support integer indexing for `__setitem__`
- `__str__` and `__repr__` methods implemented by converting to numpy and calling corresponding methods
- Implements `to_ndarray`
- Adds basic testing following along with [numpy basic indexing tutorial](https://numpy.org/doc/stable/user/basics.indexing.html)

<details>
<summary> Why advanced indexing isn't implemented:</summary>

the reason indexing by arrays or "advanced indexing" as numpy calls it is not yet supported is because numpy behavior with 2+ arrays is different than I expected. I don't *think* it should be too hard to modify the existing code to get the right functionality, but it's been a while since I started working this issue and wanted to go ahead and get out the first iteration.

This example shows how indexing by arrays can be a bit different. This is talked about a bit in https://numpy.org/doc/stable/user/basics.indexing.html

but I expected slice and array to behave the same when their indices matched up, but that's not always the case
```python
>>> n = np.arange(4).reshape(2,2)
# sometimes they line up
>>> n[:,:]
array([[0, 1],
       [2, 3]])

>>> n[:,[0,1]]
array([[0, 1],
       [2, 3]])

>>> n[[0,1],:]
array([[0, 1],
       [2, 3]])
# sometimes they do not
>>> n[[0,1],[0,1]]
array([0, 3])
```
basically when you have 2+ arrays, the functionality switches from the Cartesian product of coordinates to more coordinate-wise.

so `n[:, :]` gets `[0,0], [0,1], [1,0], [1,1]` whereas `n[[0,1],[0,1]]` only gets `[0,0], [1,1]`
</details>

### DIFF
--- a/arkouda/array_view.py
+++ b/arkouda/array_view.py
@@ -1,6 +1,6 @@
 from arkouda.pdarrayclass import pdarray, parse_single_value, create_pdarray
-from arkouda.pdarraycreation import array, arange
-from arkouda.numeric import cumprod
+from arkouda.pdarraycreation import array, arange, zeros
+from arkouda.numeric import cumprod, where, cast as akcast
 import numpy as np  # type: ignore
 from arkouda.client import generic_msg
 from arkouda.dtypes import translate_np_dtype, resolve_scalar_dtype
@@ -11,6 +11,29 @@ OrderType = Enum('OrderType', ['ROW_MAJOR', 'COLUMN_MAJOR'])
 
 
 class ArrayView:
+    """
+    A multi-dimensional view of a pdarray. Arkouda ``ArraryView`` behaves similarly to numpy's ndarray.
+    The base pdarray is stored in 1-dimension but can be indexed and treated logically as if it were multi-dimensional
+
+    Attributes
+    ----------
+    base: pdarray
+        The base pdarray that is being viewed as a multi-dimensional object
+    dtype: dtype
+        The element type of the base pdarray (equivalent to base.dtype)
+    size: int_scalars
+        The number of elements in the base pdarray (equivalent to base.size)
+    shape: pdarray[int]
+        A pdarray specifying the sizes of each dimension of the array
+    ndim: int_scalars
+         Number of dimensions (equivalent to shape.size)
+    itemsize: int_scalars
+        The size in bytes of each element (equivalent to base.itemsize)
+    order: str {'C'/'row_major' | 'F'/'column_major'}
+        Index order to read and write the elements.
+        By default or if 'C'/'row_major', read and write data in row_major order
+        If 'F'/'column_major', read and write data in column_major order
+    """
     def __init__(self, base: pdarray, shape, order='row_major'):
         self.objtype = type(self).__name__
         self.shape = array(shape)
@@ -20,7 +43,9 @@ class ArrayView:
             raise ValueError(f"cannot reshape array of size {base.size} into shape {self.shape}")
         self.base = base
         self.size = base.size
+        self.dtype = base.dtype
         self.ndim = self.shape.size
+        self.itemsize = self.base.itemsize
         if order.upper() in {'C', 'ROW_MAJOR'}:
             self.order = OrderType.ROW_MAJOR
         elif order.upper() in {'F', 'COLUMN_MAJOR'}:
@@ -31,7 +56,7 @@ class ArrayView:
         self._reverse_shape = self.shape if self.order is OrderType.COLUMN_MAJOR else self.shape[::-1]
         if self.shape.min() == 0:
             # avoid divide by 0 if any of the dimensions are 0
-            self._dim_prod = array([0] * self.shape.size)
+            self._dim_prod = zeros(self.shape.size, self.dtype)
         else:
             # cache dim_prod to avoid recalculation, reverse if row_major
             self._dim_prod = cumprod(self.shape)//self.shape if self.order is OrderType.COLUMN_MAJOR else cumprod(self._reverse_shape)//self._reverse_shape
@@ -78,17 +103,17 @@ class ArrayView:
             pass
         if isinstance(key, pdarray):
             kind, _ = translate_np_dtype(key.dtype)
-            if kind != "int":
+            if kind not in ("int", "uint"):
                 raise TypeError(f"unsupported pdarray index type {key.dtype}")
             # Interpret negative key as offset from end of array
-            key = array([key[i] + self.shape[i] if key[i] < 0 else key[i] for i in range(key.size)])
+            key = where(key < 0, akcast(key + self.shape, kind), key)
             # Capture the indices which are still out of bounds
             out_of_bounds = (key < 0) | (self.shape <= key)
             if out_of_bounds.any():
                 out = arange(key.size)[out_of_bounds][0]
                 raise IndexError(f"index {key[out]} is out of bounds for axis {out} with size {self.shape[out]}")
             coords = key if self.order is OrderType.COLUMN_MAJOR else key[::-1]
-            repMsg = generic_msg(cmd="arrayViewIntIndex", args="{} {} {}".format(self.base.name, self._dim_prod.name, coords.name))
+            repMsg = generic_msg(cmd="arrayViewIntIndex", args=f"{self.base.name} {self._dim_prod.name} {coords.name}")
             fields = repMsg.split()
             return parse_single_value(' '.join(fields[1:]))
         else:
@@ -98,7 +123,7 @@ class ArrayView:
             key = key if self.order is OrderType.COLUMN_MAJOR else key[::-1]
             for i in range(len(key)):
                 x = key[i]
-                if np.isscalar(x) and resolve_scalar_dtype(x) == 'int64':
+                if np.isscalar(x) and (resolve_scalar_dtype(x) == 'int64' or 'uint64'):
                     orig_key = x
                     if x < 0:
                         # Interpret negative key as offset from end of array
@@ -160,10 +185,10 @@ class ArrayView:
             pass
         if isinstance(key, pdarray):
             kind, _ = translate_np_dtype(key.dtype)
-            if kind != "int":
+            if kind not in ("int", "uint"):
                 raise TypeError(f"unsupported pdarray index type {key.dtype}")
             # Interpret negative key as offset from end of array
-            key = array([key[i] + self.shape[i] if key[i] < 0 else key[i] for i in range(key.size)])
+            key = where(key < 0, akcast(key + self.shape, kind), key)
             # Capture the indices which are still out of bounds
             out_of_bounds = (key < 0) | (self.shape <= key)
             if out_of_bounds.any():
@@ -171,7 +196,7 @@ class ArrayView:
                 raise IndexError(f"index {key[out]} is out of bounds for axis {out} with size {self.shape[out]}")
             coords = key if self.order is OrderType.COLUMN_MAJOR else key[::-1]
             generic_msg(cmd="arrayViewIntIndexAssign", args="{} {} {} {} {}".format(self.base.name,
-                                                                                    self.base.dtype.name,
+                                                                                    self.dtype.name,
                                                                                     self._dim_prod.name,
                                                                                     coords.name,
                                                                                     self.base.format_other(value)))
@@ -211,7 +236,7 @@ class ArrayView:
 
         Examples
         --------
-        >>> a = ak.arange(0, 6, 1).reshape(2,3)
+        >>> a = ak.arange(6).reshape(2,3)
         >>> a.to_ndarray()
         array([[0, 1, 2],
                [3, 4, 5]])
--- a/arkouda/array_view.py
+++ b/arkouda/array_view.py
@@ -38,7 +38,7 @@ class ArrayView:
         self.objtype = type(self).__name__
         self.shape = array(shape)
         if not isinstance(self.shape, pdarray):
-            raise TypeError(f"Shape cannot be type {type(self.shape)}")
+            raise TypeError(f"ArrayView Shape cannot be type {type(self.shape)}. Expecting pdarray.")
         if base.size != self.shape.prod():
             raise ValueError(f"cannot reshape array of size {base.size} into shape {self.shape}")
         self.base = base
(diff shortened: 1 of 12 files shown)


arkouda/array_view.py
```diff
@@ -0,0 +1,224 @@
...
+        if not isinstance(self.shape, pdarray):
+            raise TypeError(f"Shape cannot be type {type(self.shape)}")
+        if base.size != self.shape.prod():
+            raise ValueError(f"cannot reshape array of size {base.size} into shape {self.shape}")
+        self.base = base
+        self.size = base.size
+        self.ndim = self.shape.size
+        if order in {'C', 'row_major', 'ROW_MAJOR'}:
+            self.order = OrderType.ROW_MAJOR
+        elif order in {'F', 'column_major', 'COLUMN_MAJOR'}:
```
Ethan-DeBandi99, line 26: May be a good idea to use `order.lower()` or `order.upper()` here. That way we cover things like Row_Major as well. Not entirely necessary though.

src/IndexingMsg.chpl
```diff
@@ -13,9 +13,136 @@ module IndexingMsg
...
+                when "slice" {
+                    var (start, stop, stride) = jsonToTuple(typeCoords[i+1], 3*int);
+                    var slice: range(stridable=true) = convertSlice(start, stop, stride);
+                    var scaled: [0..#slice.size] int = slice * dimProdEntry.a[i/2];
+                    for j in 0..#slice.size {
+                        scaledCoords[offsets[i/2]+j] = scaled[j];
+                    }
+                }
+                // Advanced indexing not yet supported
+                // when "pdarray" {
```
bmcdonald3, line 69: Were you seeing issues running this code?

arkouda/array_view.py
pierce314159: good idea!

src/IndexingMsg.chpl
pierce314159: ah yeah i should make a note of this. This code was working the way I intended but numpy doesn't handle array indexing the way I expected. This is talked about a bit in https://numpy.org/doc/stable/user/basics.indexing.html

but I expected slice and pdarray to behave the same when their indices matched up, but that's not always the case
```python
>>> n = np.arange(4).reshape(2,2)
# sometimes they line up
>>> n[:,:]
array([[0, 1],
       [2, 3]])

>>> n[:,[0,1]]
array([[0, 1],
       [2, 3]])

>>> n[[0,1],:]
array([[0, 1],
       [2, 3]])
# sometimes they do not
>>> n[[0,1],[0,1]]
array([0, 3])
```
basically when you have 2+ arrays, the functionality switches from the Cartesian product of coordinates to more coordinate-wise.

so `n[:, :]` gets `[0,0], [0,1], [1,0], [1,1]` whereas `n[[0,1],[0,1]]` only gets `[0,0], [1,1]`

Ethan-DeBandi99: We should make sure to document this when documentation is added. At least link to the numpy link to indicate this is the expected functionality.

Ethan-DeBandi99:
@pierce314159 - Following up on our discussion yesterday regarding Advanced Sort combined with Basic Sorts. I found an interesting take on the subject in general. I am not sure that this is how NumPy is doing it, but it makes sense. I am not sure that this would be the most efficient way to handle the combined case, but would be interested to see how it would work out.  Unfortunately, the source was all theory and only had a single sentence on it, but I tried to run with it. I am just putting a high level summary below so that I do not forget about it.

1. Perform all the "pre-calculations" as you already are.
2. Identify that the indexing provided is a mixture of basic/advanced indexing (if not we are in the other "easier" cases we discussed yesterday (All Basic, All Advanced, single mixed). Though, I believe the last one here will be removed because it will essentially function the same as when you have multiple integer arrays triggering advanced indexing.
3. Compute all basic indexing. In doing this, you are going to essentially treat each advanced index as a full slice on that dimension.
4. Once you have the subset of the array from basic indexing, you then apply the advanced indexing on the appropriate dimensions.

I know this is in no way as technical as it should be, but it seems like the simplest thing to do. 

If you want to discuss this more, let me know.

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+        if len(key) < self.ndim:
+            # append self.ndim-len(key) many ':'s to fill in the missing dimensions
+            for i in range(self.ndim - len(key)):
+                key.append(slice(None, None, None))
+        try:
+            # attempt to convert to a pdarray (allows for view[0,2,1] instead of view[ak.array([0,2,1])]
+            # but pass on RuntimeError to allow things like view[0,:,[True,False,True]] to be correctly handled
+            key = array(key)
+        except:  # bare except is bad practice but we want to pass on any errors/warnings
+            pass
```
Ethan-DeBandi99, line 103: This just seems like there should be a better way to handle this. Maybe at the very least, if we are only expecting a `RuntimeError`, on catch that in the except? That would at least error on any unexpected errors. 

I think a helper function here may be warranted to essentially handle these cases and check what key contains. If key is only numerics, we can create the pdarray. I think a helper function to make the decision here rather than a `try except` would be better suited.

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
```
Ethan-DeBandi99, line 185: See above. The same helper function should apply here as well.

Ethan-DeBandi99 (changes requested):
Mostly minor changes for error messages. Overall looks good. The logic on all the indexing looks correct and I ran some tests and did not find any issues with that. Overall looks good, just needs a few error handling tweaks in my opinion.

joshmarshall1 (approved):
Ethan already covered everything I noticed. So everything else looks good to me

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+    itemsize: int_scalars
+        The size in bytes of each element (equivalent to base.itemsize)
+    order: str {'C'/'row_major' | 'F'/'column_major'}
+        Index order to read and write the elements.
+        By default or if 'C'/'row_major', read and write data in row_major order
+        If 'F'/'column_major', read and write data in column_major order
+    """
+    def __init__(self, base: pdarray, shape, order='row_major'):
+        self.objtype = type(self).__name__
+        self.shape = array(shape)
```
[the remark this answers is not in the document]
pierce314159: calling `array` on a pdarray should return itself without any changes, so I don't think this should be an issue. And for everything else it converts to an `ndarray` before turning it into a `pdarray` so I think tuples should be good to go for `shape` too

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+        if not isinstance(self.shape, pdarray):
+            raise TypeError(f"Shape cannot be type {type(self.shape)}")
```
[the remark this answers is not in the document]
pierce314159: Yeah I can update this. I don't actually expect this to ever be rasied, I think i mostly did this for typechecking reasons since I want to treat shape as a pdarray and `ak.array` can return a `Strings`

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+        return self.size
+
+    def __repr__(self):
+        from arkouda.client import pdarrayIterThresh
+        if self.size <= pdarrayIterThresh:
+            return self.to_ndarray().__repr__()
+        else:
+            vals = [f"'{self.base[i]}'" for i in range(3)]
+            vals.append('... ')
+            vals.extend([f"'{self.base[i]}'" for i in range(self.size-3, self.size)])
```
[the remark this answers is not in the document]
pierce314159: ooooo this is cool! yeah I'll update it to this

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+        return f"[{', '.join(vals)}], shape {self.shape}"
+
+    def __getitem__(self, key):
+        if isinstance(key, int) or isinstance(key, slice):
+            key = [key]
+        elif isinstance(key, tuple):
+            key = list(key)
```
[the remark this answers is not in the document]
pierce314159: Yeah this is less checking expected types and more convert non-list like things into a list. I'll update it to be of the form
```python
if isinstance(key, pdarray):
elif isinstance(key, list):
else:
    raise TypeError(f"Unhandled key type: {key} ({type(key)})")
```
to make sure we error on things that weren't converted to lists

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+        try:
+            # attempt to convert to a pdarray (allows for view[0,2,1] instead of view[ak.array([0,2,1])]
+            # but pass on RuntimeError to allow things like view[0,:,[True,False,True]] to be correctly handled
+            key = array(key)
+        except:  # bare except is bad practice but we want to pass on any errors/warnings
+            pass
+        if isinstance(key, pdarray):
+            kind, _ = translate_np_dtype(key.dtype)
+            if kind not in ("int", "uint"):
+                raise TypeError(f"unsupported pdarray index type {key.dtype}")
```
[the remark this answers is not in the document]
pierce314159: I didn't really think about indexing with just a boolean array, but I guess it wouldn't be too hard. There only seems to be 2 cases
- All true, we want every dimension so return the entire `ArrayView`
- Any false, we said no to any of the dimensions so return an empty `ArrayView`

Numpy seems to agree with this 
```python
>>> a = np.arange(30).reshape((5, 3, 2))
>>> a[True,True,True]
array([[[[ 0,  1],
         [ 2,  3],
         [ 4,  5]],

        [[ 6,  7],
         [ 8,  9],
         [10, 11]],

        [[12, 13],
         [14, 15],
         [16, 17]],

        [[18, 19],
         [20, 21],
         [22, 23]],

        [[24, 25],
         [26, 27],
         [28, 29]]]])

>>> a[False,True,True]
array([], shape=(0, 5, 3, 2), dtype=int64)

>>> a[True,False,True]
array([], shape=(0, 5, 3, 2), dtype=int64)

>>> a[True,True,False]
array([], shape=(0, 5, 3, 2), dtype=int64)
```

src/IndexingMsg.chpl
```diff
@@ -13,9 +13,166 @@ module IndexingMsg
...
+                when "int" {
+                    scaledCoords[offsets[i/2]] = typeCoords[i+1]:int * dimProdEntry.a[i/2];
+                }
+                when "slice" {
+                    var (start, stop, stride) = jsonToTuple(typeCoords[i+1], 3*int);
+                    var slice: range(stridable=true) = convertSlice(start, stop, stride);
+                    var scaled: [0..#slice.size] int = slice * dimProdEntry.a[i/2];
+                    for j in 0..#slice.size {
+                        scaledCoords[offsets[i/2]+j] = scaled[j];
+                    }
```
bmcdonald3, line 66: I don't totally understand the calculation that is going on here, but I think that since `scaledCoords` is a distributed array, there are going to be some remote accesses. 

Since `dimProdEntry` is also a distributed array, there will also be some remote accesses there, which make things a little more difficult than just doing a `coforall` over `scaledCoords` local subdomains. 

This may be an unhelpful comment since I don't really know what a solution would be, but I think some better locality could be achieved potentially. @ronawho I'm curious if you have any input here.

arkouda/array_view.py
pierce314159: a helper function might be a good idea. I wasn't only getting a runtime error, there was also a numpy deprecation warning that i was trying to avoid from the `np.array` done within `ak.array`
```python
>>> np.array([0,1,[True,False,True]])
/home//miniconda3/envs/ak-work/bin/ipython:1: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  #!/home//miniconda3/envs/ak-work/bin/python
array([0, 1, list([True, False, True])], dtype=object)

>>> In [9]: ak.array([0,1,[True,False,True]])
/home//proj/arkouda/arkouda/pdarraycreation.py:178: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  a = np.array(a)
<ipython-input-9-329ab0e78122> in <module>
----> 1 ak.array([0,1,[True,False,True]])

~/proj/arkouda/arkouda/pdarraycreation.py in array(a, dtype)
    208     # If not strings, then check that dtype is supported in arkouda
    209     if a.dtype.name not in DTypes:
--> 210         raise RuntimeError("Unhandled dtype {}".format(a.dtype))
    211     # Do not allow arrays that are too large
    212     size = a.size

RuntimeError: Unhandled dtype object
```

So even catching the `RuntimeError` and handling it correctly would still result in a distracting deprecation. I couldn't think of a better way to do it than this but it's far from ideal, so I'm open to suggestions

arkouda/array_view.py
```diff
@@ -0,0 +1,249 @@
...
+                                                                                         self._dim_prod.name,
+                                                                                         json.dumps(indices)))
+            reshape_dim = reshape_dim_list if self.order is OrderType.COLUMN_MAJOR else reshape_dim_list[::-1]
+            return create_pdarray(repMsg).reshape(reshape_dim, order=self.order.name)
+
+    def __setitem__(self, key, value):
+        if isinstance(key, int) or isinstance(key, slice):
+            key = [key]
+        elif isinstance(key, tuple):
+            key = list(key)
```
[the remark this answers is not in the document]
pierce314159: See above

arkouda/array_view.py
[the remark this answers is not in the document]
pierce314159: I'm gonna head out a bit early today but will add this monday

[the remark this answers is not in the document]
pierce314159: @Ethan-DeBandi99 this functionality should be working now
```python
>>> a = ak.arange(30).reshape((5, 3, 2))

>>> a[True,True,True]
array([[[[ 0,  1],
         [ 2,  3],
         [ 4,  5]],

        [[ 6,  7],
         [ 8,  9],
         [10, 11]],

        [[12, 13],
         [14, 15],
         [16, 17]],

        [[18, 19],
         [20, 21],
         [22, 23]],

        [[24, 25],
         [26, 27],
         [28, 29]]]])

>>> a[False,True,True]
array([], shape=(0, 5, 3, 2), dtype=int64)

>>> a[True,False,True]
array([], shape=(0, 5, 3, 2), dtype=int64)

>>> a[True,True,False]
array([], shape=(0, 5, 3, 2), dtype=int64)
```

arkouda/array_view.py
Ethan-DeBandi99: @pierce314159 - I think this is fine for now. Can we add an issue to look into a more robust solution and work it when we have time? I don't think we need to hold the PR up for this right now.

Ethan-DeBandi99 (approved):
This looks good to me. I think we should just add a separate issue for that `try except`.

arkouda/array_view.py
pierce314159: Issue #1333 created to track this

PR #1146, opened 2022-02-25
Change ak.read_parquet to read all supported datasets by default

bmcdonald3:
Previously, `read_parquet()` was defaulting to reading a dataset
called `array`. This PR updates the client-side code to get all
datasets if the user does not provide any datasets and then pass
those to the server to be read.

Related issue: https://github.com/Bears-R-Us/arkouda/issues/985

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -178,7 +178,7 @@ def read_parquet(filenames : Union[str, List[str]],
         datasets = [datasets]
 
     nonexistent = set(datasets) - \
-        (set(get_datasets_allow_errors(filenames)) if allow_errors else set(get_datasets(filenames[0])))
+        (set(get_datasets_allow_errors(filenames, True)) if allow_errors else set(get_datasets(filenames[0], True)))
     if len(nonexistent) > 0:
         raise ValueError("Dataset(s) not found: {}".format(nonexistent))
 
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -177,6 +177,11 @@ def read_parquet(filenames : Union[str, List[str]],
     if isinstance(datasets, str):
         datasets = [datasets]
 
+    nonexistent = set(datasets) - \
+        (set(get_datasets_allow_errors(filenames)) if allow_errors else set(get_datasets(filenames[0])))
+    if len(nonexistent) > 0:
+        raise ValueError("Dataset(s) not found: {}".format(nonexistent))
+
     rep_msg = generic_msg(cmd="readAllParquet", args=
                           f"{strictTypes} {len(datasets)} {len(filenames)} {allow_errors} {json.dumps(datasets)} | {json.dumps(filenames)}")
     rep = json.loads(rep_msg)  # See GenSymIO._buildReadAllHdfMsgJson for json structure
@@ -429,6 +434,8 @@ def get_datasets_allow_errors(filenames: List[str], is_parquet=False) -> List[st
     ----------
     filenames : List[str]
         A list of HDF5 files visible to the arkouda server
+    is_parquet : bool
+        Is filename a Parquet file; false by default
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


reuster986 (approved):

arkouda/pdarrayIO.py
```diff
@@ -418,7 +420,7 @@ def get_datasets(filename : str, is_parquet=False) -> List[str]:
     return datasets
 
 @typechecked
-def get_datasets_allow_errors(filenames: List[str]) -> List[str]:
+def get_datasets_allow_errors(filenames: List[str], is_parquet=False) -> List[str]:
     """
     Get the names of datasets in an HDF5 file
     Allow file read errors until success
```
pierce314159, line 426: The `Parameters` of this docstring should be updated to include `is_parquet`

arkouda/pdarrayIO.py
```diff
@@ -172,11 +172,13 @@ def read_parquet(filenames : Union[str, List[str]],
     """
     if isinstance(filenames, str):
         filenames = [filenames]
-    if isinstance(dsetname, str):
-        dsetname = [dsetname]
+    if datasets is None:
+        datasets = get_datasets_allow_errors(filenames, True) if allow_errors else get_datasets(filenames[0], True)
+    if isinstance(datasets, str):
+        datasets = [datasets]
```
pierce314159, line 178: This is my first time looking at this code so I'm not sure but

Should we add something similar to the `else` block in `read_all` to make sure we throw a `ValueError` listing the missing datasets?
https://github.com/Bears-R-Us/arkouda/blob/883d7010ebd6464751c6e948f2914c48125d7fda/arkouda/pdarrayIO.py#L284-L294
It's not entirely clear to me what the `nonexistent`would look like since L289-290 should never run because it's identical to the check at L286-287. I'm not sure who the author of `read_all` is but maybe @reuster986 or @mhmerrill would know?

pierce314159 (approved):
One small documentation change and one question but looks good overall!

arkouda/pdarrayIO.py
mhmerrill: @pierce314159, @reuster986 wrote the original code for this although I don't remember who touched it last... @bmcdonald3 @glitch ???

bmcdonald3: I think you are right that 289-290 would never run in the current state, so that can likely be removed, but I think what nonexistent is checking is whether or not the datasets don't exist in the file.

For example, say you had a file with this schema: `{"col1": int_col, "col2": str_col}` and you call `ak.read_all("my-file", ['col1', 'bad-name'])` then nonexistent would be `set('bad-name')` because it is the set of the input datasets without the ones that actually exist in the file. It hits the `else` case because it is not just a string, but a list of datasets.

Admittedly, I did not think about this deeply when I was just copy/pasting the code over to the Parquet work, but now that I have looked at it more, I think that check should be moved outside of the else and should always run over the datasets to make sure no bad datasets slip by.

I'll update this PR to reflect that change; should I only change it in the Parquet code for this PR or should I do it in both places? (`read_all` and `read_parquet`)

bmcdonald3: After rereading Pierce's original comment, I realize that I responded to a different question.

> Should we add something similar to the else block in read_all to make sure we throw a ValueError listing the missing datasets?

I think we should move the existing code that is in the else block that throws the value error outside of the else block to ensure it is run every time, but isn't that code in the else block already today? 

I think it was missing from the Parquet side, which I have now added.

pierce314159: ohhh yeah that makes sense! It's @mhmerrill's call but I'm in favor of updating `read_all` as well

pierce314159: oh yeah, I don't think I really knew what I was asking lol. I just figured the check probably needed to happen both places and was confused about the unreachable code in L289-290. Your explanation helped a lot

mhmerrill: can we merge something today and do another PR to resolve this?

arkouda/pdarrayIO.py
```diff
@@ -172,11 +172,18 @@ def read_parquet(filenames : Union[str, List[str]],
...
+
+    nonexistent = set(datasets) - \
+        (set(get_datasets_allow_errors(filenames)) if allow_errors else set(get_datasets(filenames[0])))
```
pierce314159, line 181: I think these need to be `get_datasets*(filenames, True)`

pierce314159 (approved):

PR #1147, opened 2022-02-25
Closes #1121 uint64 hash return

### LINKED ISSUE #1121: Change output of `.hash()` to `(uint64, uint64)`
The `.hash()` methods that compute SipHash on data should return `uint64` values. Currently, the uint64 output in Chapel is converted to signed integers before storing in the symbol table and returned as `(ak.int64, ak.int64)`

Ethan-DeBandi99:
This PR (closes #1121):

`.hash()` functions have been updated so that chapel is using uint64 for hashes. This results in uint64 returning to the client.

- `hash64` returns a single uint64 
- `hash128` returns (uint64, uint64)

### DIFF
--- a/src/EfuncMsg.chpl
+++ b/src/EfuncMsg.chpl
@@ -85,18 +85,18 @@ module EfuncMsg
                     }
                     when "hash64" {
                         overMemLimit(numBytes(int) * e.size);
-                        var a = st.addEntry(rname, e.size, int);
+                        var a = st.addEntry(rname, e.size, uint);
                         forall (ai, x) in zip(a.a, e.a) {
-                            ai = sipHash64(x): int(64);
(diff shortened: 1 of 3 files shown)


reuster986 (changes requested):
This looks good! However, there is actually one more call site to update:

https://github.com/Bears-R-Us/arkouda/blob/883d7010ebd6464751c6e948f2914c48125d7fda/src/SegmentedMsg.chpl#L510-L523

Ethan-DeBandi99:
@reuster986 - Thanks for the reminder. I remember looking at that and completely forgot to update it. I just pushed it up.

reuster986 (approved):

pierce314159 (approved):
Looks good to me

PR #1148, opened 2022-02-25
Fix quoting when getting vars from the Makefile

ronawho:
The makefile supports an arbitrary `print-%` command that will print
whatever makefile variable you want. This is used by the chapel unit
tests to get the value of `$CHPL_FLAGS`. Previously, we just `echo`d,
but the shell interprets/removes quotes causing issues for things like:
`CHPL_FLAGS="--ccflags=\"-Wno-expansion-to-defined -Wno-nullability-completeness\""`

Fix that here by using the Makefile `info` command instead of `@echo`


mhmerrill (approved):
seems to work.

PR #1150, opened 2022-02-25
Fix Parquet unit test version number test when Arrow version < 5

bmcdonald3:
Previously, the Parquet unit test was checking if the Arrow version
was greater than 5, but 5 is not a requirement, so I am removing
that check and instead just ensuring that the version major can be
cast to an integer value.

### DIFF
--- a/test/UnitTestParquetCpp.chpl
+++ b/test/UnitTestParquetCpp.chpl
@@ -122,9 +122,13 @@ proc testVersionInfo() {
   }
   var ret;
   try! ret = createStringWithNewBuffer(cVersionString);
-  if ret[0]: int >= 5 {
+  try {
+    // Ensure that version number can be cast to int
+    // Not checking version number for compatability
+    var vMajor = ret[0]:int;
     return 0;
-  } else {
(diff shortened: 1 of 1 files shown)


test/UnitTestParquetCpp.chpl
```diff
@@ -122,9 +122,13 @@ proc testVersionInfo() {
   }
   var ret;
   try! ret = createStringWithNewBuffer(cVersionString);
-  if ret[0]: int >= 5 {
+  try {
+    // Ensure that version number can be cast to int
+    // Not checking version number for compatability
```
mhmerrill, line 127: is there a minimum version number for Aarow/Parquet?

bmcdonald3: Not that I am aware of. Arrow version 4 is the oldest that we have tested with, so we could just require version 4, but I suspect it works with older versions as well. Note that the most recent version is `7.0.0`.

mhmerrill (approved):
Good to go. If we need in the future to limit version, we can.

PR #1152, opened 2022-02-25
Closes #1145: Add `uint64` support for `ak.arange`

### LINKED ISSUE #1145: add `uint64` support to `ak.arange`
currently `ak.arange()` only supports `int64` output.
so, add `uint64` support to `ak.arange()`

also maybe while we're fixing this we should also move `ak.arange()` and `ak.linspace()` into their own module, maybe call it `SequenceMsg.chpl`, or something like that.

pierce314159:
This PR (Closes #1145):
- Adds `dtype` to `ak.arange` which defaults to `int64` and allows casting to any arkouda dtype supported by `ak.cast`
  - `ak.arange(7, dtype=ak.uint64)`
  - `ak.arange(0, 7, dtype=ak.bool)`
  - `ak.arange(0, 7, -1, dtype=ak.float64)`
  - Note: this is equivalent to `ak.cast(ak.arange(7), dt=ak.MYDTYPE)`
- Adds a test in `pdarray_creation_test.py` of the `dtype` functionality against numpy
- Moves `arange` and `linspace` from `MsgProcessing.chpl` to `SequenceMsg.chpl` 
  - Note: this is just a relocation, the logic of `arange` and `linspace` did not change since casting happens on client
- Adds `SequenceMsg` to `ServerModules.cfg` and modifies `arkouda_server` to add these to `CommandMap` appropriately


arkouda/pdarraycreation.py
```diff
@@ -459,21 +462,21 @@ def arange(*args) -> pdarray:
...
         stop = args[1]
         stride = 1
 
-    #if three args are given then first arg is start,
-    #second is stop, third is stride
+    # if three args are given then first arg is start,
+    # second is stop, third is stride
     if len(args) == 3:
         start = args[0]
         stop = args[1]
```
Ethan-DeBandi99, line 482: Stride is never set here. stride = args[2] should be added.

Ethan-DeBandi99 (commented):
Is there a reason we are doing the casting client side and not storing the desired dtype server side?

arkouda/pdarraycreation.py
pierce314159: I was really confused for a second lol. Stride is set, it's just collapsed on github

pierce314159:
> Is there a reason we are doing the casting client side and not storing the desired dtype server side?

We talked about this a bit offline, but I'll drop a comment for other's awareness. The cast is still being done on the server-side just not in this function but using the `ak.cast` method. This is a good question because the right thing to do isn't obvious

In my head, the trade-off I'm making here by calling `ak.cast` is adding an extra server call in exchange for clearer code and less duplicated logic. We could certainly pass the `dtype` along with this message and do all the appropriate casting on the server (`switch`/`select` on `dtype`), but that strikes me as more error prone to have essentially the same block copied multiple times when we could just call a cast. This is in line with how `numpy` handles this since the tests match. This way also has the benefit of getting `float64` and `bool` for free. 

But like I said, it's totally not obvious this is ideal and if others think this `cast` is not performant enough, im totally down to rework the chapel code for `arange`

mhmerrill:
> > Is there a reason we are doing the casting client side and not storing the desired dtype server side?
> 
> @Ethan-DeBandi99 and I talked about this a bit offline, but I'll drop a comment for others' awareness. The cast is still being done on the server-side, not in this function, but by using the `ak.cast` method. This is a good question because the right thing to do isn't obvious
> 
> In my head, the trade-off I'm making here by calling `ak.cast` is adding an extra server call in exchange for clearer code and avoiding duplicated logic. We could certainly pass the `dtype` along with this message and do all the appropriate casting on the server (`switch`/`select` on `dtype`), but that strikes me as more error prone to have essentially the same block copied multiple times when we could just call a cast. This is in line with how `numpy` handles this since the tests match. This way also has the benefit of getting `float64` and `bool` for free.
> 
> But like I said, it's totally not clear if this is the right way to go and if others think this `cast` is not performant enough, im totally down to rework the chapel code for `arange`

@pierce314159 I am ok with this solution as long as we capture the desired `numpy` behavior, so that we return the same values or at least as close as we can make them.

mhmerrill (approved):
looks good to me.

PR #1156, opened 2022-02-25
Add support for reading and writing bool Parquet columns

bmcdonald3:
This PR adds the ability to read and write columns of boolean
type with Parquet files.

Related issue: https://github.com/Bears-R-Us/arkouda/issues/985


src/ParquetMsg.chpl
```diff
@@ -28,10 +28,12 @@ module ParquetMsg {
   extern var ARROWINT64: c_int;
   extern var ARROWINT32: c_int;
   extern var ARROWUINT64: c_int;
+  extern var ARROWBOOLEAN: c_int;
```
pierce314159, line 31: do we need an `extern var ARROWUINT32: c_int;`here? And also other places in this file? The only reason I think to ask is because `uint32` got added to the `ArrowTypes` enum

tests/parquet_test.py
```diff
@@ -14,13 +14,15 @@ def read_write_test(dtype):
         ak_arr = ak.randint(0, 2**32, SIZE)
     elif dtype =='uint64':
         ak_arr = ak.randint(0, 2**32, SIZE, dtype=ak.uint64)
+    elif dtype =='bool':
+        ak_arr = ak.randint(0, 2**32, SIZE, dtype=ak.bool)
```
pierce314159, line 18: for the `bool` case, it might be more representative to do `ak.randint(0, 1, adjusted_size, dtype=ak.bool)`

just because in it's current state, it'll only ever be false when we happen to hit a 0 which seems pretty rare. There are 2 other places in this file where this would need to be updated

tests/parquet_test.py
```diff
@@ -14,13 +14,15 @@ def read_write_test(dtype):
...
         
     ak_arr.save_parquet("pq_testcorrect", "my-dset")
     pq_arr = ak.read_parquet("pq_testcorrect*", "my-dset")
     
     for f in glob.glob('pq_test*'):
         os.remove(f)
-
+
```
pierce314159, line 25: Just whitespace:
This is extremely nitpicky and you shouldn't bother to change it unless there's another reason to update the PR but I wanted to make a note that it looks like this whitespace was introduced

pierce314159 (approved):
This looks good to me

tests/parquet_test.py
bmcdonald3: Ah, I was not aware of this, thanks!

src/ParquetMsg.chpl
bmcdonald3: Yes, we should probably add that, but I think it is for another PR, I'll remove the references to `uint32` that were added in this one. (I was testing some stuff and that slipped through when it shouldn't have, thanks for catching)

I've made note of this in https://github.com/Bears-R-Us/arkouda/issues/985 so we don't lose track of that.

mhmerrill (approved):
looks ok to me.

Issue #1151, opened 2022-02-25
`ak.get_datasets()` prepends space to parquet column names


reuster986:
In the list of column names returned by `ak.get_datasets(..., is_parquet=True)`, all but the first have a leading space added to them.

```
>>> ak.get_datasets(filename, is_parquet=True)
['col1',
 ' col2',
 ' col3',
 ...
 ' coln']

PR #1159, opened 2022-02-26
Remove trailing space for ak.get_datasets

bmcdonald3:
When implemented `ak.get_datasets` for Parquet,
I had left a trailing space in the string of
dataset names that was constructed in the C++ code,
which was causing the column names to incorrect for
all but the first column.

This slipped through testing because there is currently
no way to write multi-column Parquet files in Arkouda
and there is no support for appending to Parquet files
today. If we were able to add that functionality, testing
for Parquet would become much stronger.

Resolves: https://github.com/Bears-R-Us/arkouda/issues/1151


Ethan-DeBandi99 (approved):
I see no issues.

pierce314159 (approved):
I don't 100% understand what the `testMultiDset` proc is doing. But its structure is similar enough to the others that it seems the big thing being checked is that there isn't a space in `"col1,col3"`, so I feel comfortable approving this! And even though my `C++` isn't the strongest, the 1 character change was easy enough to follow :)

reuster986 (approved):

PR #1163, opened 2022-03-01
Add support for reading string columns from Parquet files

bmcdonald3:
This PR adds the initial, unoptimized code for reading
string columns from Parquet files using the same
`ak.read_parquet()` interface as the other Parquet dtypes.

For this initial implementation, performance was not a focus
in an attempt to add functionality and iteratively improve
performance, but it is worth noting that the performance and
scalability of this code is not good with this initial support
PR. The `segmentedCalcOffsets()` function (that is used in HDF5
as well) is used and it appears to have a large amount of comm
and does not scale as locales are added. Additionally, there does
not seem to be a query to figure out the byte size of a Parquet
string column, so we have to pass over the file twice; first to
get the number of bytes per row to be able to construct the values
array prior to reading in the values, then, a second time, to
actually read in the data.

Since there is no support for writing Parquet string columns in
Arkouda today, I wasn't able to come up with a good Arkouda unit
test for this functionality, but there is a Chapel unit test
testing the C++ code.


bmcdonald3:
```python
import pyarrow as pa
import pyarrow.parquet as pq
import numpy as np
import pandas as pd
import arkouda as ak
ak.connect()

ak_strs = ak.random_strings_uniform(minlen=1, maxlen=8, size=100)
df = pd.DataFrame({'array': ak_strs.to_ndarray()})
table = pa.Table.from_pandas(df)
pq.write_table(table, 'strings-frompa.parquet')

ak_vals = ak.read_parquet("strings-frompa.parquet", "array")

assert((ak_vals == ak_strs).all())
```

Here is a unit test that tests this code (albeit, incompletely) that we couldn't fit into the Arkouda unit tests since we do not want to add pyarrow to the dependency list, but this can be run manually if anyone wants to make sure this code works as expected.

reuster986:
Exciting!

I just stumbled across this repo, https://github.com/apache/parquet-testing, which seems to have a lot of sample files for testing parquet functionality. We might be able to use some of that for correctness checks (but not performance checks, as none of it is large).

Ethan-DeBandi99 (approved):
I do not see any issues

src/ArrowFunctions.h
```diff
@@ -19,6 +19,7 @@ extern "C" {
 #define ARROWUINT32 3
 #define ARROWBOOLEAN 4
 #define ARROWTIMESTAMP ARROWINT64
+#define ARROWSTRING 6
```
pierce314159, line 22: this prob doesn't matter but should `ARROWSTRING` be 5 since `ARROWTIMESTAMP` points to `ARROWINT64`/0

reuster986:
Regarding `segmentedCalcOffsets()`, I see a couple potential paths forward:
1. Refactor `segmentedCalcOffsets()` to use aggregation, which I believe would make it sufficiently performant and scalable, since it is the same approach used elsewhere, e.g. `FindSegments` https://github.com/Bears-R-Us/arkouda/blob/1193c8a1daf632445c28a04734ad7aaaa31bf542/src/FindSegmentsMsg.chpl#L191-L200
2. Avoid using `segmentedCalcOffsets()` altogether and compute the offsets inside `cpp_getStringColumnNumBytes`. In each iteration of the loop below, the `size` variable becomes the next offset and can be stored off in an array
https://github.com/bmcdonald3/arkouda/blob/92f9e44fa67dcff789ed42e462f92d5c685cf46c/src/ArrowFunctions.cpp#L136-L140

Option 1 is simpler to implement, but it makes me nervous to rely on NULL bytes always and only marking the end of a string. For that reason, I would like to eventually get rid of `segmentedCalcOffsets()` from the codebase altogether, and I would prefer to go with option 2 here.

That said, if we want to do this incrementally, I'm fine with merging this PR first and then moving forward with one of these options (or another proposal) in a separate PR. @bmcdonald3 I'll let you make that call.

src/ParquetMsg.chpl
```diff
@@ -285,13 +301,20 @@ module ParquetMsg {
...
     var rnames: list((string, string, string)); // tuple (dsetName, item type, id)
 
     for (dsetidx, dsetname) in zip(dsetdom, dsetnames) do {
         for (i, fname) in zip(filedom, filenames) {
             var hadError = false;
             try {
-                (sizes[i], types[dsetidx]) = (getArrSize(fname), getArrType(fname, dsetname));
+                types[dsetidx] = getArrType(fname, dsetname);
+                if types[dsetidx] == ArrowTypes.stringArr {
+                  sizes[i] = getStrColSize(filenames[0], dsetname);
```
pierce314159, line 314: is there a reason to do `filenames[0]` here instead of `fname`?

test/UnitTestParquetCpp.chpl
```diff
@@ -17,8 +17,8 @@ proc testReadWrite(filename: c_string, dsetname: c_string, size: int) {
   
   var a: [0..#size] int;
   for i in 0..#size do a[i] = i;
-  
-  if c_writeColumnToParquet(filename, c_ptrTo(a), 0, dsetname, size, 10000, false, 1, errMsg) < 0 {
+
+  if c_writeColumnToParquet(filename, c_ptrTo(a), 0, dsetname, size, 10000, 1, false, errMsg) < 0 {
```
pierce314159, line 21: were these 2 params switched intentionally?

pierce314159: Also I think this is why the CI is failing, cause this would also need to be updated here
https://github.com/Bears-R-Us/arkouda/blob/1193c8a1daf632445c28a04734ad7aaaa31bf542/test/UnitTestParquetCpp.chpl#L21

pierce314159 (changes requested):
Looks good overall except for small change that needs to be made to one of the unit tests

test/UnitTestParquetCpp.chpl
pierce314159: I think this is why the CI is failing cause this doesn't match the function signature
https://github.com/Bears-R-Us/arkouda/blob/1193c8a1daf632445c28a04734ad7aaaa31bf542/src/ArrowFunctions.h#L46-L48

src/ArrowFunctions.h
bmcdonald3: I put ARROWSTRING as 6 because I have another PR in the works for adding float support and that one I reserved for 5, it is really arbitrary though

src/ParquetMsg.chpl
bmcdonald3: Nope, that was an error; good catch

test/UnitTestParquetCpp.chpl
pierce314159: wait this does match?? okay nevermind me, im confused lol

src/ArrowFunctions.h
pierce314159: oh okay that makes sense!

bmcdonald3:
@reuster986 thank you for looking into this -- I agree that doing the calculation of the offsets array makes the most sense here, so I will plan to proceed with option 2.

> That said, if we want to do this incrementally, I'm fine with merging this PR first

I would probably lean towards merging this initial PR first and then incrementally working on performance because that would make it a little less cumbersome to keep branches and things straight and make sure I am not too divergent from master.

pierce314159 (approved):
Looks good to me!

test/UnitTestParquetCpp.chpl
bmcdonald3: I don't know what I was thinking when I did this, I wish I had an explanation!

reuster986 (approved):

Issue #1162, opened 2022-02-28
`nonexistent datasets` check in `read_all`


pierce314159:
Capturing conversation in #1146:

The `else` block in `read_all` builds `nonexistent` with any datasets in `datasets` which can't be found and raises a`ValueError`
https://github.com/Bears-R-Us/arkouda/blob/883d7010ebd6464751c6e948f2914c48125d7fda/arkouda/pdarrayIO.py#L284-L294

It look like since L289-290 should never run because it's identical to the check at L286-287.

@bmcdonald3 noted (paraphrased):
> I think that 289-290 will never run in the current state, so that can likely be removed. Nonexistent is checking is whether the datasets exist in the file.
For example, say you had a file with this schema: {"col1": int_col, "col2": str_col} and you call ak.read_all("my-file", ['col1', 'bad-name']) then nonexistent would be set('bad-name') because it is the set of the input datasets without the ones that actually exist in the file. It hits the else case because it is not just a string, but a list of datasets.
I think that check should be moved outside of the `else` and should always run over the datasets to make sure no bad datasets slip by

As ben suggested, I think `read_all` should be updated to be something like
```python
    if datasets is None:
        datasets = get_datasets_allow_errors(filenames) if allow_errors else get_datasets(filenames[0])
    if isinstance(datasets, str):
        datasets = [datasets]

    nonexistent = set(datasets) - \
        (set(get_datasets_allow_errors(filenames)) if allow_errors else set(get_datasets(filenames[0])))
    if len(nonexistent) > 0:
        raise ValueError("Dataset(s) not found: {}".format(nonexistent))
```
Feedback from @reuster986 would be great since (I believe) he is the original author of much of the IO code and can hopefully let us know if we're missing something here

Note: if this change is made, `io_test.py` would likely need to be updated as bad dataset names will be caught sooner and will result in a `ValueError` instead of a `RuntimeError`

_Originally posted by @pierce314159 in https://github.com/Bears-R-Us/arkouda/pull/1146#discussion_r816151850_

reuster986:
I agree with the consensus here -- removing the redundant check in 289-290 and moving the "nonexistent" check outside the if/else so it always runs. I'm not sure why I wrote it that way to begin with, it's been so long... I was probably just coding fast and made a mistake. Good catch @pierce314159 and @bmcdonald3 !

Ethan-DeBandi99:
Noting that this code has moved to `pdarrayIO.py:read()`
[...]

PR #1169, opened 2022-03-01
Move nonexistent check in read_all outside of else block so it always runs

bmcdonald3:
In some work in the Arkouda client side, it was identified
that there was a check to ensure that the datasets that the
user provided to be read in would only be run if multiple
datasets were provided, not just a single, string dataset.

This PR moves that check outside of the else block so it
will run on all provided datasets.

Closes: https://github.com/Bears-R-Us/arkouda/issues/1162


bmcdonald3:
Closing this as I do not totally understand how the unit tests should be updated to account for this new difference; someone with more knowledge of the differences between `read_all`/`read_hdf`/`load` should likely make this change. (my apologies for the noise)

PR #1170, opened 2022-03-01
Add support for reading and writing float64 Arrow columns

bmcdonald3:
This PR adds support for writing float64 (real) arrays in
Chapel to Parquet files and reading both float and double
data types from Parquet files. The written float64 values
are written to Parquet as Arrow's double datatype.


pierce314159 (approved):
This looks good to me!

I struggled a bit determining where we need to handle `ARROWFLOAT` and `ARROWDOUBLE` vs just handling `ARROWDOUBLE`, but I got there. It seems to be just when we're reading we handle both (and convert to double/float64). Then when writing everything should be a double

Ethan-DeBandi99 (approved):
I am assuming the TODO tags for throwing errors are intended to be handled in another PR.

I see no issues.

mhmerrill (approved):
looks ok to me.

Issue #1153, opened 2022-02-25
Server crashes on bad column names when reading parquet


reuster986:
Related to #1151 , in that I discovered this issue when trying to read parquet files using column names from `ak.get_datasets()`, which prepends a space to most names. When these incorrect names are used, arkouda should raise an exception, but instead the server crashes:

```python
>>> columns = ak.get_datasets(filenames[0], is_parquet=True)
>>> data = ak.read_parquet(filenames, dsetname=columns)
```

Sample output in arkouda_server terminal:
```
uncaught TaskErrors: 4 errors: Error: ParquetError 108 on_fn:ParquetMsg Dataset: <dataset_name> does not exist in file: <file_path>
  /path/to/arkouda/src/ParquetMsg.chpl:99: thrown here
  /path/to/arkouda/src/ParquetMsg.chpl:99: uncaught here
[repeat for each locale]
```

PR #1175, opened 2022-03-03
Get Parquet string reading offsets in C++ function

bmcdonald3:
Previously, Parquet string files that had been read in were
calculating their offset arrays with the `segmentedCalcOffsets`
function from `GenSymIO.chpl`, but that function is not optimized
and was resulting in large amounts of communication, meaning that
performance decreased as more nodes were added.

This PR switches the computation of the offsets array to happen
at the point where the byte sizes are determined from the file,
resulting in much better scaling.

Here are some performance numbers calculated on a CS using
50 files of 1 million uniformly distributed string elements:

nodes | master  | this branch |
----- | ------: | ----------: |
1     | 2.366 s | 0.3248 s    |
2     | minutes | 0.2899 s    |
4     | minutes | 0.2625 s    |
8     | minutes | 0.2135 s    |

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -122,13 +122,34 @@ module ParquetMsg {
     }
   }
 
-  proc getStrColSize(filename: string, dsetname: string) throws {
-    extern proc c_getStringColumnNumBytes(filename, colname, errMsg): int;
+  proc calcSizesAndOffset(offsets: [] ?t, byteSizes: [] int, filenames: [] string, sizes: [] int, dsetname: string) throws {
+    var (subdoms, length) = getSubdomains(sizes);
+    
+    coforall loc in offsets.targetLocales() do on loc {
+      var locFiles = filenames;
+      var locFiledoms = subdoms;
+      forall (i, filedom, filename) in zip(sizes.domain, locFiledoms, locFiles) {
+        for locdom in offsets.localSubdomains() {
+          const intersection = domain_intersection(locdom, filedom);
+          if intersection.size > 0 {
+            var pqErr = new parquetErrorMsg();
+            var col: [filedom] t;
+            byteSizes[i] = getStrColSize(filename, dsetname, col);
+            offsets[filedom] = col;
+          }
+        }
+      }
+    }
+  }
+
+  proc getStrColSize(filename: string, dsetname: string, offsets: [] int) throws {
+    extern proc c_getStringColumnNumBytes(filename, colname, offsets, errMsg): int;
     var pqErr = new parquetErrorMsg();
     
     var byteSize = c_getStringColumnNumBytes(filename.localize().c_str(),
-                                          dsetname.localize().c_str(),
-                                          c_ptrTo(pqErr.errMsg));
+                                             dsetname.localize().c_str(),
+                                             c_ptrTo(offsets),
+                                             c_ptrTo(pqErr.errMsg));
     if byteSize == ARROWERROR then
       pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
     return byteSize;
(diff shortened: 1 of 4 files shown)


mhmerrill (approved):
this looks good to me.

src/ParquetMsg.chpl
```diff
@@ -122,13 +122,34 @@ module ParquetMsg {
...
+    coforall loc in offsets.targetLocales() do on loc {
+      var locFiles = filenames;
+      var locFiledoms = subdoms;
+      forall (i, filedom, filename) in zip(sizes.domain, locFiledoms, locFiles) {
+        for locdom in offsets.localSubdomains() {
+          const intersection = domain_intersection(locdom, filedom);
+          if intersection.size > 0 {
+            var pqErr = new parquetErrorMsg();
+            var col: [filedom] t;
+            byteSizes[i] = getStrColSize(filename, dsetname, col);
```
reuster986, line 137: I forget, what happens if this function throws an error from inside a `forall`? I seem to remember having issues throwing from within a `forall`, but maybe that was with an old version of Chapel.

reuster986 (approved):
This looks great, and the performance gain is awesome! I had some in-line comments, but I think I convinced myself that everything looks good, so I'm approving.

src/ParquetMsg.chpl
bmcdonald3: This code is almost identical to the code that was causing the errors in https://github.com/Bears-R-Us/arkouda/issues/1153, so when you are in a `forall`/`coforall`, Chapel will sum up all the errors and return a `TaskErrors` which has all the errors together. The fix that implemented for the bug you were seeing was sort of a workaround for that, and I am not sure why the throwing `TaskErrors` is not caught by the main Arkouda `try`/`catch` in `arkouda_server.chpl`. 

I am planning on investigating that some more in the near feature, but I think you are correct that there may be some issues with throwing in these loops today. Since there are already other instances doing this in the codebase today (specifically https://github.com/Bears-R-Us/arkouda/blob/20f0315bf8f3723f4bc9a8803f1f89cc01d98ae8/src/ParquetMsg.chpl#L100-L123), I was thinking that we should get this in now and then I will hopefully know more about the throwing errors soon, but I have no problem waiting until I've pinned that down if you think that is a safer approach.

src/ParquetMsg.chpl
```diff
@@ -122,13 +122,34 @@ module ParquetMsg {
...
 
-  proc getStrColSize(filename: string, dsetname: string) throws {
-    extern proc c_getStringColumnNumBytes(filename, colname, errMsg): int;
+  proc calcSizesAndOffset(offsets: [] ?t, byteSizes: [] int, filenames: [] string, sizes: [] int, dsetname: string) throws {
+    var (subdoms, length) = getSubdomains(sizes);
+    
+    coforall loc in offsets.targetLocales() do on loc {
+      var locFiles = filenames;
+      var locFiledoms = subdoms;
+      forall (i, filedom, filename) in zip(sizes.domain, locFiledoms, locFiles) {
```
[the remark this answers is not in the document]
bmcdonald3: Yes, I think you've got it right, there shouldn't be duplicated work because the `forall` is with default rectangular arrays on `loc`.

If you look closely, you will see that the column that we are using here is declared as `var col: [filedom] t;`, when it should ideally be defined as `var col: [intersection] t;` and then only read a subset of the data, rather than all of the contents of the file, but due to how we have implemented the Parquet stuff up til today, we are unfortunately reading in entire files even if the intersection on a single locale is only, say, a single element.

I hate to keep saying we have plans to resolve this soon, but this is another thing we plan to resolve soon. We have been investigating ways to only read the necessary portions of the file, and think we have figured out the path forward, which is to skip to the `intersection.low` in the Parquet file and assign only the necessary amount of elements directly into the local slice of the distributed array, which will also avoid the extra copy that `col` is introducing here and in the main `readFilesByName` loop.

src/ParquetMsg.chpl
reuster986: Thanks for the explanation! I am ok with merging now while the investigation of error handling continues.

src/ParquetMsg.chpl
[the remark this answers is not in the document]
reuster986: Ah, that's a good point. It's totally fine that this is on the "plan to resolve soon" list :)

PR #1177, opened 2022-03-04
Closes #1165 drop on axis

### LINKED ISSUE #1165: `ak.DataFrame` Drop on Axis
`pandas.DataFrame` allows for dropping columns and rows using a `axis` parameter. The default is `axis=0` which will drop specified row indexes. Setting `axis=1` allows columns to be dropped. `ak.DataFrame` needs to be updated to so that the drop functionality allows for an `axis` parameter and setting its values replicate the functionality of `pandas.DataFrame.drop`

For reference: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html

Ethan-DeBandi99:
This PR (closes #1165):

Pandas `DataFrame.drop()` utilizes an `axis` parameter to trigger drops on rows or columns. This functionality has been added to Arkouda here.

- `axis` should be 0 or 1. 0 = rows, 1 = columns.
- Tests have been added to: validate column drop using axis param, validate row drop, validate exception cases are caught as expected.
- Documentation has been updated to include `.drop()` for `DataFrame`. 

For testing I added functions to build an `ak.DataFrame` and a `pd.DataFrame`. I only implemented this for drop because there is an existing issue in to handle this for all tests.

This functionality is currently only available in-place at this time.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -407,20 +407,16 @@ class DataFrame(UserDict):
         last_idx = -1
         # sort to ensure we go in ascending order.
         keys.sort()
-        for i, k in enumerate(keys):
+        for k in keys:
             if not isinstance(k, int):
                 raise TypeError("Index keys must be integers.")
-            pda = self.index[(last_idx+1):k:1]
-            if pda.size > 0:
-                idx_list.append(pda)
+            idx_list.append(self.index[(last_idx+1):k])
             last_idx = k
 
-        pda = self.index[(last_idx+1):]
-        if pda.size > 0:
-            idx_list.append(pda)
+        idx_list.append(self.index[(last_idx+1):])
 
         idx_to_keep = concatenate(idx_list)
-        for key, val in self.items():
+        for key in self.keys():
             # using the UserDict.__setitem__ here because we know all the columns are being reset to the same size.
             # This avoids the size checks we would do when only setting a single column
             UserDict.__setitem__(self, key, self[key][idx_to_keep])
@@ -433,8 +429,8 @@ class DataFrame(UserDict):
         ----------
         keys : str, int or list
             The labels to be dropped on the given axis
-        axis : int
-            The axis on which to drop from. 0 - drop rows, 1 - drop columns
+        axis : int or str
+            The axis on which to drop from. 0/'index' - drop rows, 1/'columns' - drop columns
 
         Examples
         ----------
@@ -450,10 +446,10 @@ class DataFrame(UserDict):
         if isinstance(keys, str) or isinstance(keys, int):
             keys = [keys]
 
-        if axis == 0:
+        if axis == 0 or axis == 'index':
             #drop a row
             self._drop_row(keys)
-        elif axis == 1:
+        elif axis == 1 or axis == 'columns':
             #drop column
             self._drop_column(keys)
         else:
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -409,7 +409,7 @@ class DataFrame(UserDict):
         keys.sort()
         for i, k in enumerate(keys):
             if not isinstance(k, int):
-                raise ValueError("Index keys must be integers.")
+                raise TypeError("Index keys must be integers.")
             pda = self.index[(last_idx+1):k:1]
             if pda.size > 0:
                 idx_list.append(pda)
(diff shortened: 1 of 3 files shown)


mhmerrill (approved):
looks ok to me.

arkouda/dataframe.py
```diff
@@ -382,25 +382,83 @@ def _repr_html_(self):
...
 
             del self[key]
 
+    def _drop_row(self, keys):
+        """
+        Drop a row or rows from the dataframe, in-place.
+
+        keys : list
+            The indexes to be dropped on the given axis
+        """
```
pierce314159, line 405: i think we should allow for users to pass in a pdarray for keys in addition to a list, but that can be handled in a different PR

arkouda/dataframe.py
```diff
@@ -382,25 +382,83 @@ def _repr_html_(self):
...
+            if pda.size > 0:
+                idx_list.append(pda)
+            last_idx = k
+
+        pda = self.index[(last_idx+1):]
+        if pda.size > 0:
+            idx_list.append(pda)
+
+        idx_to_keep = concatenate(idx_list)
+        for key, val in self.items():
```
pierce314159, line 423: i don't thing `val` is being used here, so we can just do `for key in self.keys():`
or it might have to be `self.data.keys()`

pierce314159 (changes requested):
A few things I'd like you to take a look at but I think the overall functionality looks good

arkouda/dataframe.py
```diff
@@ -382,25 +382,83 @@ def _repr_html_(self):
...
+            if not isinstance(k, int):
+                raise TypeError("Index keys must be integers.")
+            pda = self.index[(last_idx+1):k:1]
+            if pda.size > 0:
+                idx_list.append(pda)
+            last_idx = k
+
+        pda = self.index[(last_idx+1):]
+        if pda.size > 0:
+            idx_list.append(pda)
```
pierce314159, line 420: I just chatted with ethan about this offline. but if the `pda.size >0` check isn't necessary (i.e. `concatenate([],...)` behaves the way we want it to with an empty list), then we can simplify the logic here and avoid creating the extra `pda` with something like

```python
idx_list.append(self.index[(last_idx+1):k])
# and
idx_list.append(self.index[(last_idx+1):])

File: arkouda/dataframe.py [hunk 4]
```diff
@@ -382,25 +382,83 @@ def _repr_html_(self):
...
+            keys = [keys]
+
+        if axis == 0:
+            #drop a row
+            self._drop_row(keys)
+        elif axis == 1:
+            #drop column
+            self._drop_column(keys)
+        else:
+            raise ValueError("axis must be 0 or 1")
```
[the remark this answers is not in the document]
Ethan-DeBandi99: updated to allow both and updated docstring to include new values. Updated error message.

arkouda/dataframe.py
```diff
@@ -382,25 +382,83 @@ def _repr_html_(self):
...
+            The indexes to be dropped on the given axis
+        """
+        idx_list = []
+        last_idx = -1
+        # sort to ensure we go in ascending order.
+        keys.sort()
+        for i, k in enumerate(keys):
+            if not isinstance(k, int):
+                raise TypeError("Index keys must be integers.")
+            pda = self.index[(last_idx+1):k:1]
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Removed :1

arkouda/dataframe.py
```diff
@@ -382,25 +382,83 @@ def _repr_html_(self):
...
+        Drop a row or rows from the dataframe, in-place.
+
+        keys : list
+            The indexes to be dropped on the given axis
+        """
+        idx_list = []
+        last_idx = -1
+        # sort to ensure we go in ascending order.
+        keys.sort()
+        for i, k in enumerate(keys):
```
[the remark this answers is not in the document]
Ethan-DeBandi99: updated loop to not call enumerate. This was left over from testing prior to submitting PR.

arkouda/dataframe.py
Ethan-DeBandi99: Updated to use only the keys in the loop.

arkouda/dataframe.py
Ethan-DeBandi99: Removed the `if pda.size > 0` and updated append to those listed above.

pierce314159 (approved):
Looks good!

PR #1179, opened 2022-03-04
Add UsedModules.cfg to .gitignore

bmcdonald3:
The UsedModules.cfg file, which is generated from --saveUsedModules
was not being ignored by git, and I don't think we would ever want
a UsedModules.cfg pushed into the repo, so moving it into the
gitignore will resolve any annoyance from having to selectively not
include that file.


mhmerrill (approved):
this looks good.

PR #1183, opened 2022-03-07
Closes #1181: use apache/parquet-testing files

### LINKED ISSUE #1181: Test parquet against data from apache/parquet-testing
To improve CI coverage of parquet functionality, I think we should incorporate some of the test files from https://github.com/apache/parquet-testing/tree/master/data into arkouda and run unit tests against them. As a start, I would recommend:
- `alltypes_plain.parquet` (except `timestamp_col`, which has `Int96` dtype)
- `alltypes_plain.snappy.parquet` (except `timestamp_col`, which has `Int96` dtype)
- `delta_byte_array.parquet`

The parent repo is under the Apache 2.0 license, which if I'm reading it correctly allows redistribution of files with other projects if we include the original license. I think if we create a new dir `resources/parquet-testing` and put the data files and their `LICENSE.txt` in there, we would satisfy the requirements.

We can then write tests to check that:
- [ ] `ak.get_datasets` returns all columns (except `timestamp_col`)
- [ ] `ak.read_parquet` reads all columns without error (probably not necessary to validate the data)

reuster986:
This PR incorporates standard test files from github.com/apache/parquet-testing and adds a unit tests that checks `ak.get_datasets` and `ak.read_parquet` against them. Attribution and license are included.

Currently, the new unit test is skipped by default because it fails, but it can be run manually with `python3 -m pytest tests/parquet_test.py --optional-parquet`.


tests/parquet_test.py
```diff
@@ -120,3 +120,35 @@ def test_get_datasets(self):
...
+                    'c_birth_country',
+                    'c_login',
+                    'c_email_address',
+                    'c_last_review_date']
+        for basename, ans in zip(filenames, (columns1, columns1, columns2)):
+            filename = os.path.join(datadir, basename)
+            columns = ak.get_datasets(filename, is_parquet=True)
+            self.assertListEqual(columns, ans)
+            # Merely test that read succeeds, do not check output
+            data = ak.read_parquet(filename, datasets=columns)
```
bmcdonald3, line 154: Is there a reason that we are choosing not to test the output? It seems the correct output data is stored in that repo under some MD files (example: https://github.com/apache/parquet-testing/blob/master/data/delta_byte_array.md), so we could manually copy that data in, similarly to how the columns are added here.

reuster986: The expected output for that file is in https://github.com/apache/parquet-testing/blob/master/data/delta_byte_array_expect.csv, which is a 1000 line CSV file -- a little too much hard-code in :)

Once `ak.read_parquet` can successfully read the file, it might be a good idea to have it read the expected CSV with `pandas.read_csv` and compare, but today `ak.read_parquet` crashes on that file, so it's hard to write the test.

bmcdonald3: Got a chance to look into this a bit; as far as the byte array file goes, it appears to be `DELTA_BYTE_ARRAY-encoded`, which is something that just is not supported by the installs of Arrow that I (we?) have and I don't think is a limitation with our implementation, but I am disappointed to see all of these server crashes popping up and hope to iron some of those out this week.

For the `alltypes_plain.parquet` file, we are seeing the `timestamp96` column slip by because our check is that the type is an instance of the timestamp Arrow type, which it technically is, so I may add an explicit check to make sure we don't try to read timestamp96 columns, which appear to be able to be read in, but the results are incorrect (of course). The `float_col` is what is making the call to `ak.read_parquet()` not work on this file. I am surprised this is failing, but I will try to get this fixed. I have confirmed that we can read double files that are generated by Pyarrow, but am seeing now that float columns fail for some reason, which is quite surprising to me... 

It will be great to get these files testing to get confidence with the Parquet implementation, seeing this bugs come up is definitely exposing some holes in the current implementation.

bmcdonald3: Opened https://github.com/Bears-R-Us/arkouda/pull/1186 to fix the float reading issue and it now is able to read through the whole `alltypes_plain.parquet` file correctly, but is still reading the timestamp96 column.

reuster986: Good news on #1186!

For the DELTA_BYTE_ARRAY, I didn't realize that was a special encoding with complicated support. I just chose that file because it had some longer and more varied string columns, not because we have a definite need to support the encoding. I wish I had a better way to figure out exactly what encodings and parameters our sample data actually uses, but I haven't determined how to parse that info from the output of `parquet-tools inspect`. 

I would be ok with removing the DELTA_BYTE_ARRAY file, but it might also be good to leave it in to check the error handling... what do you think?

bmcdonald3: Ya, I agree that leaving it in as another error check makes sense; I'll keep you updated as I work on a fix for that.

bmcdonald3: Alright, I've opened https://github.com/Bears-R-Us/arkouda/pull/1188 to catch all Parquet errors which should fix the server crashes that you have seen.

I'd be curious on any input in that approach though, which is just catching all errors and reporting them without really doing anything more advanced to handle the errors more gracefully, but at least it gets around server crashes.

Ethan-DeBandi99 (changes requested):
`@pytest.mark.optional_parquet` needs to be updated to `@pytest.mark.run_optional_parquet`, or `config.addinivalue_line("markers", "run_optional_parquet: mark test as slow to run")` should be updated to `config.addinivalue_line("markers", "optional_parquet: mark test as slow to run")`. Otherwise, you will get the following warning that the custom marker is not registered:
```
tests/parquet_test.py:124
  /Users/ethandebandi/Documents/git/arkouda/tests/parquet_test.py:124: PytestUnknownMarkWarning: Unknown pytest.mark.optional_parquet - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.optional_parquet
```

tests/parquet_test.py
```diff
@@ -120,3 +120,35 @@ def test_get_datasets(self):
 
         for f in glob.glob('pq_test*'):
             os.remove(f)
+
+    @pytest.mark.optional_parquet
```
Ethan-DeBandi99, line 124: Updated to `@pytest.mark.run_optional_parquet` or update the marker.

tests/conftest.py
```diff
@@ -1,8 +1,24 @@
...
+    if config.getoption("--run-optional-parquet"):
+        # --run-optional-parquet given in cli: do not skip optional parquet tests
+        return
+    skip_parquet = pytest.mark.skip(reason="need --run-optional-parquet option to run")
+    for item in items:
+        if "optional_parquet" in item.keywords:
+            item.add_marker(skip_parquet)
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "run_optional_parquet: mark test as slow to run")
```
Ethan-DeBandi99, line 21: Update `run_optional_parquet` or update the decorator

tests/parquet_test.py
```diff
@@ -120,3 +120,35 @@ def test_get_datasets(self):
...
+        columns2 = ['c_customer_id',
+                    'c_salutation',
+                    'c_first_name',
+                    'c_last_name',
+                    'c_preferred_cust_flag',
+                    'c_birth_country',
+                    'c_login',
+                    'c_email_address',
+                    'c_last_review_date']
+        for basename, ans in zip(filenames, (columns1, columns1, columns2)):
```
pierce314159, line 149: just want to double check that the repeating of `columns1` is intended

pierce314159 (approved):
I think this looks good once the marker and decorator are updated to match like Ethan suggests

tests/parquet_test.py
```diff
@@ -120,3 +120,35 @@ def test_get_datasets(self):
...
+        columns1 = ['id',
+                    'bool_col',
+                    'tinyint_col',
+                    'smallint_col',
+                    'int_col',
+                    'bigint_col',
+                    'float_col',
+                    'double_col',
+                    'date_string_col',
+                    'string_col']
```
bmcdonald3, line 139: As far as the exclusion of the timestamp column goes, I think we should actually include that column since it is deprecated and we technically can read the column, we just won't be getting correct results.

I looked into this for a bit and it seems difficult to differentiate between a timestamp of int96 and int64 programmatically, and I think part of that is because the int96 timestamps have been deprecated.

If we are concerned about being able to detect int96 vs int64 timestamps and skip the int96 timestamp columns, I could invest more time in that, but I wonder the likelihood of that scenario given that int96 timestamps are deprecated.

Is the skipping of int96 timestamp columns essential?

tests/parquet_test.py
pierce314159: After looking at this more, I realize the `columns1` is supposed to be repeated. it corresponds with the file being read. Both `'alltypes_plain.parquet'` and `'alltypes_plain.snappy.parquet'` expect `columns1`. And `'delta_byte_array.parquet'` expects `columns2`

mhmerrill (approved):
looks ok to me now

PR #1185, opened 2022-03-07
Update Arkouda to work with CTypes

bradcray:
In Chapel 1.26, various C types from 'SysCTypes', 'SysBasic' and
'CPtr' are being brought together into a single (new) module,
'CTypes'.  In this PR, I'm adding a new user-level 'CTypes'
compatibility module that assembles the pieces that Arkouda relies
upon together so that it can use the new names and organization yet
still be compiled with older compilers.


mhmerrill (approved):
this looks good to me.

Issue #1158, opened 2022-02-26
Improve Parquet error handling


bmcdonald3:
On my first pass of the error handling code for Parquet, I did not have a super strong grasp on error handling in Arkouda and there are several places in the code that should be updated now that I am understanding the error handling better. Specifically, all throwing Parquet functions should be wrapped in a `try/catch` block and all HDF5 specific error handling code should be removed. I was naively assuming that the Parquet and HDF5 error handling would be very similar, but now I am realizing that is not the case.

Ideally, there would be no way for any Parquet function calls to crash the server, but that is not the case today.

PR #1188, opened 2022-03-08
Catch all Parquet errors and report them to the client

bmcdonald3:
In the initial Parquet error handling code, the only errors that
were being reported were errors that came from status codes of
calls to the Parquet API, but all errors in the C++ code went
unhandled and would result in a server crash.

This PR wraps all the C++ functions in a try/catch to catch all
errors to be reported back to the client rather than crashing the
server. This approach seems to align with the philosophy of
"The server should never crash, errors should just be reported to
the client".

Additionally, a bug was identified with throwing errors in `forall`
loops within a function, which would cause a try/catch wrapping
the throwing function to not catch the resulting `TaskErrors` error
throw, but a workaround for that is to wrap `forall` loops in a
try/catch which just throws the error, which then allows the
overarching try/catch to catch and report the error, where previously
it would crash the server.

Related issue: https://github.com/chapel-lang/chapel/issues/19378
Closes: https://github.com/Bears-R-Us/arkouda/issues/1158

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -50,68 +50,127 @@ bool check_status_ok(arrow::Status status, char** errMsg) {
 */
 
 int64_t cpp_getNumRows(const char* filename, char** errMsg) {
-  std::shared_ptr<arrow::io::ReadableFile> infile;
-  ARROWRESULT_OK(arrow::io::ReadableFile::Open(filename, arrow::default_memory_pool()),
-                 infile);
+  try {
+    std::shared_ptr<arrow::io::ReadableFile> infile;
+    ARROWRESULT_OK(arrow::io::ReadableFile::Open(filename, arrow::default_memory_pool()),
+                   infile);
 
-  std::unique_ptr<parquet::arrow::FileReader> reader;
-  ARROWSTATUS_OK(parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
+    std::unique_ptr<parquet::arrow::FileReader> reader;
+    ARROWSTATUS_OK(parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
   
-  return reader -> parquet_reader() -> metadata() -> num_rows();
+    return reader -> parquet_reader() -> metadata() -> num_rows();
+  } catch (const std::exception& e) {
+    *errMsg = strdup(e.what());
+    return ARROWERROR;
+  }
 }
 
 int cpp_getType(const char* filename, const char* colname, char** errMsg) {
-  std::shared_ptr<arrow::io::ReadableFile> infile;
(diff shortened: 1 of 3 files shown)


src/ArrowFunctions.cpp
```diff
@@ -50,68 +50,127 @@ bool check_status_ok(arrow::Status status, char** errMsg) {
...
+      return ARROWTIMESTAMP;
+    else if(myType->id() == arrow::Type::BOOL)
+      return ARROWBOOLEAN;
+    else if(myType->id() == arrow::Type::STRING)
+      return ARROWSTRING;
+    else if(myType->id() == arrow::Type::FLOAT)
+      return ARROWFLOAT;
+    else if(myType->id() == arrow::Type::DOUBLE)
+      return ARROWDOUBLE;
+    else // TODO: error type not supported
```
Ethan-DeBandi99, line 111: Since error handling is being configured in this PR, should we go ahead and add the error here? Looking at the places where this is called, it looks like it will continue on returning `ARROWUNDEFINED` and it looks like actually producing/handling the error at this point would be more accurate.

src/ArrowFunctions.cpp
```diff
@@ -50,68 +50,127 @@ bool check_status_ok(arrow::Status status, char** errMsg) {
...
+        while (ba_reader->HasNext()) {
+          parquet::ByteArray value;
+          (void)ba_reader->ReadBatch(1, nullptr, nullptr, &value, &values_read);
+          size += value.len + 1; // add 1 to account for zero that will be added in Chapel array
+          offsets[i++] = value.len + 1;
+        }
+      }
+      return size;
+    }
+    return ARROWUNDEFINED;
```
Ethan-DeBandi99, line 163: Would we want an error in the event the something is not a string here? Returning undefined does not seem wrong, but if the type does not meet the criteria I would think we would want an error. I may be wrong in my thought process here.

bmcdonald3: In the case where this function is called and it isn't a string, I think it should be an error, but in order for that to happen something would have to go pretty wrong on the server side, so it is pretty unlikely. The `ARROWUNDEFINED` and `ARROWERROR` are actually the same thing, so maybe we should just condense them to be only `ARROWERROR` to remove some confusion. I don't think having the second variable is all that valuable.

Ethan-DeBandi99: That make sense. I was not sure if that was the intended functionality or not, which is why I brought it up. I definitely agree that it would make sense to only use `ARROWERROR` to avoid confusion.

Ethan-DeBandi99:
Everything looks good to me, other than needing to add a `throws` in `ParquetMsg.chpl`.

Ethan-DeBandi99 (approved):
Everything looks good to me.

mhmerrill (approved):
this looks ok to me.

PR #1192, opened 2022-03-09
Closes #1022 collapse unique and value_counts server messages

### LINKED ISSUE #1022: collapse unique and value_counts server messages
collapse unique and value_counts server messages because they are duplicative.
I think value_counts has all the functionality.
the only reason not to do this may be performance but I don't think so on the surface.

Ethan-DeBandi99:
This PR (closes #1022):

- Removed `value_countsMsg` function
- Removes registration of `value_count` messages

All functionality was handled already in `unique` messages. These messages are configured to handle returning value_counts via a flag that is passed in the message.

### DIFF
--- a/src/UniqueMsg.chpl
+++ b/src/UniqueMsg.chpl
@@ -158,83 +158,9 @@ module UniqueMsg
            }
         }
     }
-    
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
looks ok to me.

pierce314159 (approved):

PR #1194, opened 2022-03-09
Update marker name to match decorator

pierce314159:
missed a couple places where `--run-optional-parqet` needed to be updated


bmcdonald3 (approved):
I tested this locally and the change that is being made to switch to `--optional-parquet` works as expected, but there are still some failures in the file.

I would assume that is not part of this PR, so I think this is good to go, but should I be working to resolve those failures at the moment?

pierce314159:
I rebased on top of master to get Ben's latest arrow error handling which is showing up here as a lot of noise (both in commits and files changed) but all of this is from master and not new functionality

The only changes I've made are:
- 12d968a0ae858bd64af1cae6b799b703aeeab67f

mhmerrill (approved):
wow, look ok to me, i guess these are a bunch of changes from master into this branch.

PR #1195, opened 2022-03-09
Add support for reading binary data columns in Parquet

bmcdonald3:
Parquet has a couple different representations of string columns,
one that is called "string", which guarantees characters will
be UTF8, and one of "binary", which is the same as string, but
without the UTF8 guarantee. This PR adds support for reading
the binary data type by treating it as a string.


mhmerrill (approved):
looks ok to me.

PR #1196, opened 2022-03-10
Update Arrow version to 7 and add snappy to install

bmcdonald3:
This PR updates the Arrow version to the latest install and adds snappy support to the install, since support for reading/writing with snappy was recently added.


pierce314159:
After installing snappy with conda, I was able to successfully make arkouda with snappy support using this PR

pierce314159 (approved):

PR #1201, opened 2022-03-10
Issue #1189: Fix type of H5Literate argument to H5_iter_order_t

### LINKED ISSUE #1189: Bad arguments to C_HDF5.H5Literate()
The following calls to C_HDF5.H5Literate() have the wrong argument types for argument 3:

https://github.com/Bears-R-Us/arkouda/blob/master/src/HDF5Msg.chpl#L182-L187

Specifically, the third argument should be one of the following values:

```chapel
// H5_iter_order_t enum                                                     
    extern type H5_iter_order_t = c_int;
    extern const H5_ITER_UNKNOWN :H5_iter_order_t;
    extern const H5_ITER_INC :H5_iter_order_t;
    extern const H5_ITER_DEC :H5_iter_order_t;
    extern const H5_ITER_NATIVE :H5_iter_order_t;
    extern const H5_ITER_N :H5_iter_order_t;
```

where today, it is:

```c
C_HDF5.H5Literate(fid, C_HDF5.H5_INDEX_NAME, C_HDF5.H5_INDEX_NAME, idx_p, c_ptrTo(_get_item_count), c_ptrTo(nfields));
```

Chapel doesn't catch this error because of the way that these external C enums are represented as Chapel code; but turning on a bunch of C warnings in the generated C code will catch it.  I believe what's happening in practice is that the enum value whose integral value happens to match that of `C_HDF5.H5_INDEX_NAME` is being sent in instead and working out well enough that nobody has noticed.

I'm not familiar enough with these calls or their use in Arkouda to know what value should be getting used here instead.

glitch:
Fixes #1189 by updating the arg type in H5Literate calls to H5_iter_order_t.
Uses the "NATIVE" version since we really don't care about the order we're just using it to get the item count and allocate space.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -179,12 +179,12 @@ module HDF5Msg {
         
         // First iteration to get the item count so we can ballpark the char* allocation
         var nfields:c_int = 0:c_int;
-        C_HDF5.H5Literate(fid, C_HDF5.H5_INDEX_NAME, C_HDF5.H5_INDEX_NAME, idx_p, c_ptrTo(_get_item_count), c_ptrTo(nfields));
+        C_HDF5.H5Literate(fid, C_HDF5.H5_INDEX_NAME, C_HDF5.H5_ITER_NATIVE, idx_p, c_ptrTo(_get_item_count), c_ptrTo(nfields));
         
         // Allocate space for array of strings
         var c_field_names = c_calloc(c_char, 255 * nfields);
         idx_p = 0:C_HDF5.hsize_t; // reset our iteration counter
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

mhmerrill (approved):
ok, i understand it now and it looks ok to me.

bradcray:
I don't know this code or call well enough to know what the call does or what value "should" be passed in, but given @glitch's comment in the OP and the fact that it hasn't seemed sensitive to us passing in a semi-random value before, I think this is good to go.  The new value should definitely fix the C compiler complaint that caused me to open the issue in any case.

bradcray (approved):

PR #1204, opened 2022-03-11
Only read necessary elements in Chapel array for Parquet

bmcdonald3:
Previously, Parquet was having each node read the entire file of
any file that had an intersection with its local domain into a
temporary array and then assigning that local array to the
distributed array, which was resulting in extra communication
as well as keeping around a copy of the entire file, which would
mean that memory footprints on each node were much higher than
they needed to be.

This PR switches that to instead read directly into the local
portion of the distributed array and only the necessary number
of elements, skipping to the offset location of each file.

Here are some performance numbers collected on a 16-node-cs-hdr
using the `parquetIO` benchmark:

locales | master      | this branch
------- | ----------- | -----------
1       | 3.11 GiB/s  | 3.44 GiB/s
4       | 11.67 GiB/s | 14.02 GiB/s
8       | 22.43 GiB/s | 27.42 GiB/s
16      | 40.05 GiB/s | 49.27 GiB/s

String reading:
locales | master     | this branch
------- | ---------- | -----------
1       | 1.10 GiB/s | 1.66 GiB/s
4       | 1.09 GiB/s | 1.30 GiB/s
8       | 0.90 GiB/s | 1.22 GiB/s
16      | 0.88 GiB/s | 0.93 GiB/s

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -122,13 +122,12 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
   }
 }
 
-int cpp_getStringColumnNumBytes(const char* filename, const char* colname, void* chpl_offsets, char** errMsg) {
+int cpp_getStringColumnNumBytes(const char* filename, const char* colname, void* chpl_offsets, int64_t numElems, int64_t startIdx, char** errMsg) {
   try {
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99 (approved):
This looks good to me

pierce314159 (approved):

mhmerrill (approved):
looks ok to me.

PR #1206, opened 2022-03-11
Closes #1203 - `SegArray` - Allow Empty Segments

### LINKED ISSUE #1203: `SegArray` - Allow Empty Segments
`SegArray` is not currently configured to allow for any empty segmets. In most cases, this is very easy to correct by only changing some of the restrictions within the `__init__()` method.

An issue still remains when the empty segment is the last element of the SegArray. The current thought for handling this is to use a boolean array with the same size as segments to indicate which indexes in the `segments` array are empty segments. This will allow for empty segments to be ignored during computation, but still be represented in the dataset, preventing operations like broadcast from failing.

The intention here is to use Segments of length 0 to indicate an empty segment. For example [0, 3, 3, 5] indicates that a segment starts at index 0 and ends at index 2, an empty segment follows, a segment starting at index 3 and ending at index 4 is third and the final segment starts at index 5. This would look something like:
```json
SegArray([
[0, 1, 2]
[]
[3, 4]
[5, 6, 7]
])
```

This issue is required before #1197 will be full function.

Ethan-DeBandi99:
This PR (closes #1203):

- Modifies the `SegArray` Class to allow for `SegArray` objects to be created such that they contain empty segments.
- Updates functionality so that `SegArray` objects with empty segments do not error during some operations
    - `get_ngrams()`
    - `append_single()`
    - `remove_repeats()` - This functioned, but dropped all empty segments. I configured it to maintain any empty segments, but we can alter this if needed.
 - Updated testing to ensure proper functionality when `SegArray` contains empty segments.

Creating a`SegArray` with an empty segment is simple. If the empty segment at `n-1` (where `n` is the number of segments), the empty segment is denoted by duplicate indexes in `SegArray.segments`. If the empty segment is at `n`, the index into `SegArray.values` is `SegArray.values.size`. Examples below:
```python
# SegArray with first segment empty
values = ak.array([0, 1, 2, 3, 4, 5, 6 , 7, 8])
segments = ak.array([0, 0, 3, 7])
segarr = ak.SegArray(segments, values)
segarr

SegArray([
[]
[0 1 2]
[3 4 5 6]
[7 8]
])

# SegArray with 3rd segment empty
values = ak.array([0, 1, 2, 3, 4, 5, 6 , 7, 8])
segments = ak.array([0, 2, 5, 5, 8])
segarr = ak.SegArray(segments, values)
segarr

SegArray([
[0 1]
[2 3 4]
[]
[5 6 7]
[8]
])

# SegArray with last segment empty
# SegArray with 3rd segment empty
values = ak.array([0, 1, 2, 3, 4, 5, 6 , 7, 8])
segments = ak.array([0, 2, 5, 9])
segarr = ak.SegArray(segments, values)
segarr

SegArray([
[0 1]
[2 3 4]
[5 6 7 8]
[]
])
```

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -95,10 +95,10 @@ class SegArray:
         """
         if not isinstance(segments, pdarray) or segments.dtype != akint64:
             raise TypeError("Segments must be int64 pdarray")
-        if not is_sorted(segments) or (unique(segments).size != segments.size):
+        if not is_sorted(segments):
             raise ValueError("Segments must be unique and in sorted order")
         if segments.size > 0:
-            if segments.min() != 0 or segments.max() >= values.size:
+            if segments.min() != 0:
                 raise ValueError("Segments must start at zero and be less than values.size")
         elif values.size > 0:
             raise ValueError("Cannot have non-empty values with empty segments")
@@ -115,12 +115,22 @@ class SegArray:
         else:
             self.lengths = lengths
         self.dtype = values.dtype
+
+        self.non_empty = zeros(self.size, dtype=akbool)
+        self.non_empty_count = 0
+        for i in range(self.lengths.size):
+            if self.lengths[i] == 0:
+                self.non_empty[i] = False
+            else:
+                self.non_empty[i] = True
+                self.non_empty_count += 1
+
         if grouping is None:
             if self.size == 0:
                 self.grouping = GroupBy(zeros(0, dtype=akint64))
             else:
                 # Treat each sub-array as a group, for grouped aggregations
-                self.grouping = GroupBy(broadcast(self.segments, arange(self.size), self.valsize))
+                self.grouping = GroupBy(broadcast(self.segments[self.non_empty], arange(self.non_empty_count), self.valsize))
         else:
             self.grouping = grouping
 
@@ -384,7 +394,7 @@ class SegArray:
 
         ngrams = []
         notsegstart = ones(self.valsize, dtype=akbool)
-        notsegstart[self.segments] = False
+        notsegstart[self.segments[self.non_empty]] = False
         valid = ones(self.valsize - n + 1, dtype=akbool)
         for i in range(n):
             end = self.valsize - n + i + 1
@@ -393,7 +403,14 @@ class SegArray:
                 valid &= notsegstart[i:end]
         ngrams = [char[valid] for char in ngrams]
         if return_origins:
-            origin_indices = self.grouping.broadcast(arange(self.size), permute=True)[:valid.size][valid]
+            # set the proper indexes for broadcasting. Needed to alot for empty segments
+            seg_idx = zeros(self.non_empty_count, dtype=akint64)
+            idx = 0
+            for i in range(self.size):
+                if self.non_empty[i]:
+                    seg_idx[idx] = i
+                    idx += 1
+            origin_indices = self.grouping.broadcast(seg_idx, permute=True)[:valid.size][valid]
             return ngrams, origin_indices
         else:
             return ngrams
@@ -560,7 +577,7 @@ class SegArray:
         else:
             lastscatter = newsegs + newlens - 1
         newvals[lastscatter] = x
-        origscatter = arange(self.valsize) + self.grouping.broadcast(arange(self.size), permute=True)
+        origscatter = arange(self.valsize) + self.grouping.broadcast(arange(self.non_empty_count), permute=True)
         if prepend:
             origscatter += 1
         newvals[origscatter] = self.values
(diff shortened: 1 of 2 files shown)


arkouda/segarray.py
```diff
@@ -115,12 +115,22 @@ def __init__(self, segments, values, copy=False, lengths=None, grouping=None):
         else:
             self.lengths = lengths
         self.dtype = values.dtype
+
+        self.non_empty = zeros(self.size, dtype=akbool)
+        self.non_empty_count = 0
```
pierce314159, line 120: Do we want these to be visible to users? I'm thinking we should rename these to `self._non_empty` and `self._non_empty_count` if we are planning on them only being accessed by internal code

arkouda/segarray.py
```diff
@@ -115,12 +115,22 @@ def __init__(self, segments, values, copy=False, lengths=None, grouping=None):
...
+        for i in range(self.lengths.size):
+            if self.lengths[i] == 0:
+                self.non_empty[i] = False
+            else:
+                self.non_empty[i] = True
+                self.non_empty_count += 1
+
```
pierce314159, line 127: I think we might be able to get a cleaner initialization if we do something like this

```python
self.non_empty = self.lengths > 0
self.non_empty_count = self.non_empty.sum()
```
This way we don't have to iterate through `self.lengths` on the client

arkouda/segarray.py
```diff
@@ -393,7 +403,14 @@ def get_ngrams(self, n, return_origins=True):
...
         ngrams = [char[valid] for char in ngrams]
         if return_origins:
-            origin_indices = self.grouping.broadcast(arange(self.size), permute=True)[:valid.size][valid]
+            # set the proper indexes for broadcasting. Needed to alot for empty segments
+            seg_idx = zeros(self.non_empty_count, dtype=akint64)
+            idx = 0
+            for i in range(self.size):
+                if self.non_empty[i]:
+                    seg_idx[idx] = i
+                    idx += 1
```
pierce314159, line 412: I think this can be accomplished with this
```python
seg_idx = arange(self.size)[self.non_empty]
```

pierce314159 (commented):
A few places where we might be able to eliminate iteration on the client

arkouda/segarray.py
Ethan-DeBandi99: Updated to `_non_empty` and `_non_empty_count`. I was not sure if they would have any user facing value, but rethinking it, I am not sure they would.

arkouda/segarray.py
Ethan-DeBandi99: I agree. This has been updated.

arkouda/segarray.py
Ethan-DeBandi99: Updated to your suggestion.

arkouda/segarray.py
```diff
@@ -588,12 +594,20 @@ def remove_repeats(self, return_multiplicity=False):
...
+        # Correct segments to properly assign empty lists - prevents dropping empty segments
+        if not self._non_empty.all():
+            empty_segs = []
+            for i in range(self.size):
+                if not self._non_empty[i]:
+                    if i < truesegs.size:
+                        empty_segs.append(truesegs[i])
+                    else:
+                        empty_segs.append(truepaths.size)
+            truesegs = concatenate([truesegs, array(empty_segs)])
```
pierce314159, line 610: I think this simplifies this a bit, it's still not ideal (i dont love the way `trailing_empties` is constructed). I tried to use `ak.where` but the lengths didn't line up the way i wanted
```python
if not self._non_empty.all():
    # empties after the first truesegs.size will be out of bounds, we append truepaths.size to compensate
    trailing_empties = array([truepaths.size] * (self.size - truesegs.size))
    empty_segs = concatenate([truesegs, trailing_empties])[~self._non_empty]
    truesegs = concatenate([truesegs, empty_segs])
```
also just for fun.... awful one liner! lol
```python
truesegs = truesegs if self._non_empty.all() else concatenate([truesegs, concatenate([truesegs, array([truepaths.size] * (self.size - truesegs.size))])[~self._non_empty]])

```

pierce314159 (approved):
the tests are a bit hard to review because it's difficult telling what a SegArray will look like from it's pieces. I left one note about how to *maybe* simplifiy a bit of logic, but I think it's good in it's current state

arkouda/segarray.py
pierce314159: note: we might want to update this once `ak.full` from #1207 is implemented

tests/segarray_test.py
```diff
@@ -321,3 +439,15 @@ def test_remove_repeats(self):
...
+        segments = ak.array([0, len(a), len(a), len(a)+len(b)])
+        segarr = ak.SegArray(segments, flat)
+        dedup = segarr.remove_repeats()
+        self.assertListEqual(dedup.lengths.to_ndarray().tolist(), [3, 0, 3, 0])
+        self.assertListEqual(dedup[0].tolist(), list(set(a)))
+        self.assertListEqual(dedup[1].tolist(), [])
+        self.assertListEqual(dedup[2].tolist(), list(set(b)))
+        self.assertListEqual(dedup[3].tolist(), [])
+
+        # TODO - how to configure so that this will not remove empty segments
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I removed the completed TODOs. Thanks for catching that.

arkouda/segarray.py
Ethan-DeBandi99: Updated this to avoid using the loop.

Ethan-DeBandi99:
> In doing some additional testing this morning, I discovered one scenario where the `remove_repeats()` code does not produce the desired output. Converting this to draft until the issue is resolved.
> 
> ```python
> # original SegArray
> SegArray([
> [1 1 1 2 3]
> []
> []
> [10 11 11 12]
> ])
> 
> # remove_repeats() result.
> SegArray([
> [1 2 3]
> []
> [10 11 12]
> []
> ])
> ```

Updated logic and added a test case to make sure edge case that was failing is now handled. Once CI completes, will move from draft status.

reuster986 (approved):
Well done on handling empty segments, and it is really great to have this functionality covered by the CI!

PR #1214, opened 2022-03-15
Closes #1174 ak.randint() fails for ranges greater than 2**63

### LINKED ISSUE #1174: ak.randint() fails for ranges greater than 2**63
`ak.randint(-(2**63),0,100)` appears to work as expected
`ak.randint(0,2**63-1,100)` appears to work as expected
`ak.randint(-(2**63),2**63-1,100)`, returns only repeats of same value, i.e. `-(2**63)`

range of `2**64-1`  results in 2 unique values, and so on.

Ethan-DeBandi99:
This PR (closes #1174):

- Updates `fillInt()` in `arkouda/src/RandArray.chpl` to use an unsigned `modulus`. This allows us to use the full 64 bits and prevent the overflow seen in #1174
- Added unit test to validate functionality to `tests/pd_array_creation_test.py`.

<details>
  <summary>Process:</summary>

The 2 factors that came into play here:
-  The way we converted `a` to unsigned (the code below does an absolute value), essentially zeroes out the signed bit, so we lose one bit of range
```chapel
[ai in a] if (ai < 0) { ai = -ai; }
```
- `modulus` being stored as a signed 64 bit resulted in overflows, so for the example in the issue
```
modulus = aMax - aMin = (2**63-1) - (-(2**63)) = 2**64-1
```

Our proposed solution converts both of these to unsigned
```chapel
const a2 = a:uint;
if (aMax > aMin) {
  const modulus = (aMax - aMin):uint;
  a = (a2 % modulus):t + aMin:t;
}
```
</details>

NOTE:
- We are converting the signed ints to unsigned differently, so randint with a seed will provide different results before and after this PR. Unfortunately, this means anywhere `randint()` is called using seed, we will see different results.
- We are creating a uint copy of `a` where the original code was in place. We might be able to avoid this if we don't define `a` as a `const`, but the it's not clear if this is better since the compiler is able to do optimizations with const

### DIFF
--- a/src/RandArray.chpl
+++ b/src/RandArray.chpl
@@ -23,7 +23,7 @@ module RandArray {
       }
       const a2 = a:uint;
       if (aMax > aMin) {
-        const modulus = (aMax - aMin):uint;
+        const modulus = aMax:uint - aMin:uint;
         a = (a2 % modulus):t + aMin:t;
       }
   }
--- a/src/RandArray.chpl
+++ b/src/RandArray.chpl
@@ -21,14 +21,10 @@ module RandArray {
         var seed = (seedStr:int) + here.id;
         fillRandom(a, seed);
       }
-      [ai in a] if (ai < 0) { ai = -ai; }
+      const a2 = a:uint;
       if (aMax > aMin) {
-        const modulus = aMax - aMin;
-        if modulus > 0 {
-          [x in a] x = ((x % modulus) + aMin):t;
-        } else {
-          [x in a] x = (x + aMin):t;
-        }       
+        const modulus = (aMax - aMin):uint;
+        a = (a2 % modulus):t + aMin:t;
       }
   }
 
--- a/src/RandArray.chpl
+++ b/src/RandArray.chpl
@@ -24,7 +24,11 @@ module RandArray {
       [ai in a] if (ai < 0) { ai = -ai; }
       if (aMax > aMin) {
         const modulus = aMax - aMin;
-        [x in a] x = ((x % modulus) + aMin):t;
+        if modulus > 0 {
+          [x in a] x = ((x % modulus) + aMin):t;
+        } else {
+          [x in a] x = (x + aMin):t;
+        }       
       }
   }
 
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -124,6 +124,11 @@ class PdarrayCreationTest(ArkoudaTest):
         self.assertEqual(ak.bool, test_array.dtype)
         
         test_ndarray = test_array.to_ndarray()
+
+        # test resolution of modulus overflow - issue #1174
+        test_array = ak.randint(-(2**63), 2**63-1, 10)
+        to_validate = np.full(10, -(2**63))
+        self.assertFalse((test_array.to_ndarray() == to_validate).all())
         
         for value in test_ndarray:
             self.assertTrue(value in [True,False])
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -159,7 +159,8 @@ class PdarrayCreationTest(ArkoudaTest):
 
     def test_randint_with_seed(self):
         values = ak.randint(1, 5, 10, seed=2)
-        self.assertTrue((ak.array([4, 3, 1, 3, 4, 4, 2, 4, 3, 2]) == values).all())
+        print(values)
+        self.assertTrue((ak.array([4, 3, 1, 3, 2, 4, 4, 2, 3, 4]) == values).all())
 
         values = ak.randint(1, 5, 10, dtype=ak.float64, seed=2)
         self.assertTrue((ak.array([2.9160772326374946, 4.353429832157099, 4.5392023718621486, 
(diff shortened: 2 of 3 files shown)


src/RandArray.chpl
```diff
@@ -24,7 +24,11 @@ module RandArray {
       [ai in a] if (ai < 0) { ai = -ai; }
       if (aMax > aMin) {
         const modulus = aMax - aMin;
-        [x in a] x = ((x % modulus) + aMin):t;
+        if modulus > 0 {
```
mhmerrill, line 27: this bends my brain a little, maybe we should comment this with the reason why we do this.

Ethan-DeBandi99: @pierce314159 and I initially thought this would work, but thinking about it more, we can expand this.

The `if else` handles detection of an integer overflow in `modulus`, `aMax - aMin > 2**64`. In the case that there is an integer overflow, this solution only provides `2**63` of range. This essentially overrides `aMax` to be `aMin + 2**63`

pierce314159: We think we got a better solution by converting `uint` and back. This has the benefit of getting the full 64 bits of range and is simpler logic

mhmerrill (dismissed):
looks ok.

src/RandArray.chpl
```diff
@@ -21,10 +21,10 @@ module RandArray {
         var seed = (seedStr:int) + here.id;
         fillRandom(a, seed);
       }
-      [ai in a] if (ai < 0) { ai = -ai; }
+      const a2 = a:uint;
       if (aMax > aMin) {
-        const modulus = aMax - aMin;
-        [x in a] x = ((x % modulus) + aMin):t;
+        const modulus = (aMax - aMin):uint;
```
ronawho, line 26: This `aMax - aMin` can still result in integer overflow, which is currently undefined behavior. e.g. `max(int) - (-1)` is undefined, not `min(int)`. 

Note that unsigned integer overflow is well defined and wraps. See https://github.com/chapel-lang/chapel/issues/6160 for more discussion on overflow.

We'll probably want to do something to convert to uint before doing the math. My mind goes to something like:

```chpl
const modules = if aMin > 0 then aMax:uint - aMin:uint else aMax:uint + abs(aMin):uint`
```

But that's not quite right since `abs(min(int))` is undefined as well (at least I think so, if it is defined the result is still `min(int)` which still causes problems.) Let me think on this for a bit and see what we do in our standard library for cases like this.

bradcray: This is the kind of question I've found Paul @cassella really excels at, if he's got any free cycles.  If I haven't mis-internalized his past guidance, my instinct would be that one could just write `(aMax:uint - aMin:uint) and get the right answer, assuming `aMax > aMin`—i.e., that any signed/unsigned wraparounds will simply do the right thing.  Paul, want to check my work?

Unrelated to this PR (and with the caveat that I don't know this code), seeing a modulus in the context of random number generation worries me slightly, though.  E.g., if I'm generating random 8-bit sequences and then using % to try and constrain them to be between 0..250, statistically speaking, I'm going to get twice as many entries in the 0..5 range than the 6..250 range due to wraparound, aren't I?  I think the more typical practice is to drop out-of-bound or "leftover" values on the floor and "re-roll".  I've lost track of the full interface of our random number module, but after a quick review of it the other week, I thought it had some capabilities along these lines.  Tagging @mppf here who is more expert in RNG than I am.

mhmerrill: remember we are trying to work like 
https://numpy.org/doc/stable/reference/random/generated/numpy.random.randint.html

mppf: I agree with Brad's concern about `%`. (But it will not matter for all use cases). In the Random module we handle getting a bounded value with a re-roll that doesn't affect the RNG state. (see the getNext with min/mai here https://chapel-lang.org/docs/modules/standard/Random/PCGRandom.html#PCGRandom.PCGRandomStream.getNext ). We could potentially wire that up to `fillRandom` so you could give min/max arguments to `fillRandom`.

mhmerrill: that would be great but we still need a short term fix for this issue.

pierce314159: Ethan and I could attempt to write the reroll and avoid using `%`. Or if a member of the chapel team writes the code we can add in that code as a workaround until that becomes the lowest supported chapel version

Ethan-DeBandi99: I just pushed an update based on Brad's feedback.
@pierce314159 added an issue for tracking modulo bias and ensuring uniformity of the data (#1220).

pierce314159 (approved):

mhmerrill (dismissed):
this looks ok to me.

tests/pdarray_creation_test.py
```diff
@@ -124,6 +124,11 @@ def test_randint(self):
         self.assertEqual(ak.bool, test_array.dtype)
         
         test_ndarray = test_array.to_ndarray()
+
+        # test resolution of modulus overflow - issue #1174
+        test_array = ak.randint(-(2**63), 2**63-1, 10)
+        to_validate = np.full(10, -(2**63))
```
mhmerrill, line 130: on a second look this doesn't seem correct, looking at the issue.
you should get values across the entire range [-(2**63), 2**63-1] and not just the value of (-(2**63))

pierce314159: I believe the assertion here is testing that the output is not `-(2**63)` repeated `size` times

```python
self.assertFalse((test_array.to_ndarray() == to_validate).all())
```

Ethan-DeBandi99: Just to confirm, what @pierce314159 stated was my intention with the test.

PR #1222, opened 2022-03-18
Add initial support for writing string Parquet files

bmcdonald3:
This PR adds a `save_parquet()` function to the `strings` object,
allowing Arkouda strings to be written to Parquet files. The Chapel
code is based on the HDF5 aggregators function.

Parquet string reading does not allow saving the offsets array today,
nor are you able to load in the offsets array like you can with HDF5,
but I don't think that is as big of a performance bottleneck for
Parquet reading as it is for HDF5 reading.


Ethan-DeBandi99 (approved):
Looks good

mhmerrill (approved):
looks ok to me.

PR #1224, opened 2022-03-18
Unify HDF5 and Parquet ls calls into a single function

bmcdonald3:
This PR moves the `ls_hdf()` function, which accepts an
`is_parquet` argument to a generic `ls()` function. This
allows the user to call `ak.ls(filename)` on either a
Parquet or HDF5 file without having to specify its type.

This is done by having a generic command that is sent to the
server and then allowing the server to differentiate between
the two different file types and calling the correct server
function.

While we would like to get here with all of the IO code, this
function seemed like a good starting point since both functions
have the same arguments (just the filename) and it is a
relatively simple function. Discussion on the naming and how we
would like to handle the differentiating of the file types would
be appreciated.

Closes: https://github.com/Bears-R-Us/arkouda/issues/1234

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -7,29 +7,27 @@ from arkouda.pdarrayclass import pdarray, create_pdarray
 from arkouda.strings import Strings
 from arkouda.categorical import Categorical
 
-__all__ = ["ls_hdf", "read_hdf", "read_all", "load", "get_datasets",
+__all__ = ["ls_any", "read_hdf", "read_all", "load", "get_datasets",
            "load_all", "save_all", "read_parquet"]
 
 ARKOUDA_HDF5_FILE_METADATA_GROUP = "_arkouda_metadata"
 
-
 @typechecked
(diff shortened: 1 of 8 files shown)


Ethan-DeBandi99 (commented):
The logic in the code looks good. 
The CI issues appear to be at least partially due to missing test files, but the code looks good.

pierce314159 (approved):
This looks correct to me! Well done!

I'll leave discussions about how we want to name generic io to @reuster986 and @mhmerrill

bmcdonald3:
Ya, I think we should have a discussion about naming, possibly at the Arkouda weekly call on Tuesday?

Ethan-DeBandi99 (approved):

reuster986 (approved):
Code looks great!

Regarding the name of the function, I would lean towards `ak.ls()`, or if that's too ambiguous, then `ak.ls_file()`.

bmcdonald3:
Alright, I have updated the name to just `ak.ls()`.

PR #1227, opened 2022-03-22
Add append mode for Parquet file writing

bmcdonald3:
This PR adds append mode for all dtypes in Parquet besides
strings. Since Parquet does not support any method of appending
columns to a file, all values are read from the file before
appending the column and rewriting out the entire file. Due to
this, as well as using the higher level `ReadTable` and
`WriteTable` functions instead of `ReadBatch` and `WriteBatch`,
the performance of writing Parquet files using append mode is
significantly worse than writing a completely new Parquet file.

Some additional limitations with the Parquet implementation
require that columns appended to an existing match the number
of rows of other columns and that the dataset name is not
duplicated in the file (in other words, this only allows you
to add a new column to the file, not add values to an existing
column).

command                     | exec time 1 node | exec time 8 nodes |
--------------------------- | ----------------:| -----------------:|
truncate (default)          | 0.0967 s         | 0.072 s           |
append with 1 existing col  | 0.42 s	       | 0.106 s           |
append with 2 existing cols | 0.59 s           | 0.110 s           |

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1421,7 +1421,7 @@ class Strings:
         if mode.lower() in 'truncate':
             m = 0
         else:
-            raise ValueError("Allowed modes are 'truncate'")
+            raise ValueError("Allowed modes are 'truncate', append not yet supported on strings")
 
         try:
             json_array = json.dumps([prefix_path])
@@ -1429,7 +1429,7 @@ class Strings:
             raise ValueError(e)
 
         cmd = "writeParquet"
-        args = f"{self.entry.name} {dataset} {json_array} str {compressed}"
+        args = f"{self.entry.name} {dataset} {m} {json_array} str {compressed}"
         return cast(str, generic_msg(cmd, args))
         
     def is_registered(self) -> np.bool_:
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -23,6 +23,8 @@ module ParquetMsg {
 
   private config const logLevel = ServerConfig.logLevel;
   const pqLogger = new Logger(logLevel);
+  config const TRUNCATE: int = 0;
+  config const APPEND: int = 1;
   
   private config const ROWGROUPS = 512*1024*1024 / numBytes(int); // 512 mb of int64
   // Undocumented for now, just for internal experiments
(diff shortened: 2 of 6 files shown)


src/ParquetMsg.chpl
```diff
@@ -375,18 +400,56 @@ module ParquetMsg {
...
+      if !allexist || (matchingFilenames.size != filenames.size) {
+        throw getErrorWithContext(
+                   msg="appending to existing files must be done with the same number " +
+                      "of locales. Try saving with a different directory or filename prefix?",
+                   lineNumber=getLineNumber(), 
+                   routineName=getRoutineName(), 
+                   moduleName=getModuleName(), 
+                   errorClass='MismatchedAppendError'
+              );
+      }
```
Ethan-DeBandi99, line 433: I may be missing something here, but it appears that we are duplicating our checks for file existence. I am not sure the `exists()` check on the file names is needed. It appears redundant to me with 
```python
var matchingFilenames = glob("%s_LOCALE*%s".format(filename, ".parquet"));

var warnFlag = processParquetFilenames(filenames, matchingFilenames, mode);
```
being called. 
Is there a reason that we cannot just check 
```python
if matchingFilenames.size == 0{
    # throw error
}
if matchingFilename.size != filenames.size {
    # throw error
}
```

Ethan-DeBandi99 (changes requested):
Nothing major. I think we need to update the `if mode.lower in` lines I have commented on. This will not be an issue 99% of the time, but could present an issue.

The `processParquetFilenames()` does not seem like an issue to me, but possibly repeated work. I may be missing something, but if not we should simplify this.

src/ParquetMsg.chpl
bmcdonald3: I believe the reason that we do two similar checks that would both fail in the same case is that the two cases are different and so we want to generate a different error message.

If there are no files, we let the user know we haven't found any, if there aren't the correct number of files, we let the user know that. So they both would fail under the same conditions, but presenting a different error message in both cases is helpful.

Note that this is drawn from the HDF5 code, so a change here would likely mean we want to change the HDF5 code as well.

mhmerrill: My two cents... make it match the HDF5 code then file an issue, we should also look into unifying the duplicated code.

Ethan-DeBandi99: @mhmerrill - sounds good. Issue #1229 submitted.

arkouda/strings.py
```diff
@@ -1421,15 +1421,15 @@ def save_parquet(self, prefix_path : str, dataset : str='strings_array',
         if mode.lower() in 'truncate':
```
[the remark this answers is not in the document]
Ethan-DeBandi99: After discussing with @bmcdonald3, a new issue has been created to address this so globally so that the code stays in sync for HDF5 and Parquet for the time being. Issue #1230 has been added to address this.

Ethan-DeBandi99:
Issues #1229 and #1230 have been added to address change requests at a later date. I resolved my requests as a result. Everything else looks good to go.

Ethan-DeBandi99 (approved):

PR #1231, opened 2022-03-23
Bulk append values in Parquet and switch some places to int64

bmcdonald3:
This PR updates the Parquet append function to switch from
adding values one at a time to instead bulk add all of the
values. Additionally, there were some places in the C++/C code
that should have been `int64_t` instead of plain `int`, and those
have been updated here.

Performance with 10m int elems on 8 nodes of a Cray CS:
version     | 1 existing column | 2 existing columns |
----------- | ----------------: | -----------------: |
master      | 0.327 s           | 0.460 s            |
this branch | 0.270 s           | 0.368 s            |

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -122,7 +122,7 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good

mhmerrill (approved):
looks ok.

PR #1235, opened 2022-03-24
Closes #1190 Index for Indexing DataFrame

### LINKED ISSUE #1190: `ak.DataFrame` Use `Index` for Indexing
The current configuration of `ak.DataFrame` uses indexes that are configured as a "column". This means that `Dataframe._columns` contains a value `index` and the indexes are stored in an `pdarray` as a column. Pandas uses an `Index` object to manage indexing of DataFrames. 

`ak.DataFrame` needs to be updated to utilize an `Index` object.
- Add an index property to the class
- Configure an index object.
- Update functionality utilizing indexes to work with new indexes
- Remove the `index` column.

Ethan-DeBandi99:
This PR (closes #1190):

Arkouda DataFrames were previously using a column labeled `'index'` for indexing and then artificially providing an `index` property via a function that returned `df.data['index']`. In order to configure this to function more in line with pandas, Arkouda DataFrames have been updated to have a property that uses Arkouda Index objects as the index.

This new configuration prevents the need to filter the columns to avoid pulling out the index. Currently, the functionality of Arkouda Index objects are not fully comparable to pandas.Index. This does not currently impact our functionality, but is noted as it may need to be improved upon in the future.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -297,8 +297,8 @@ class DataFrame(UserDict):
             rtn_data = {}
             s = key
             for k in self._columns:
-                rtn_data[k] = UserDict.__getitem__(self, k)[s.start:s.stop:s.step]
-            return DataFrame(initialdata=rtn_data, index=list(range(self.size)[s.start:s.stop:s.step]))
+                rtn_data[k] = UserDict.__getitem__(self, k)[s]
+            return DataFrame(initialdata=rtn_data, index=arange(self.size)[s])
         else:
             raise IndexError("Invalid selector: unknown error.")
 
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -148,10 +148,6 @@ class DataFrame(UserDict):
             self._columns = initialdata._columns
             if index == None:
                 self.index = initialdata.index
-            elif isinstance(index, Index):
-                self.index = index
-            elif isinstance(index, pd.Index):
-                self.index = index.values.tolist()
             else:
                 self.index = index
             self.data = initialdata.data
@@ -167,10 +163,6 @@ class DataFrame(UserDict):
 
             if index == None:
                 self.index = initialdata.index.values.tolist()
-            elif isinstance(index, Index):
-                self.index = index
-            elif isinstance(index, pd.Index):
-                self.index = index.values.tolist()
             else:
                 self.index = index
             self.data = {}
@@ -236,10 +228,6 @@ class DataFrame(UserDict):
             # creating a new one.
             if self.index is None:
                 self.index = arange(0, self._size, 1)
-            elif isinstance(index, Index):
-                self.index = index
-            elif isinstance(index, pd.Index):
-                self.index = index.values.tolist()
             else:
                 self.index = index
 
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -21,6 +21,7 @@ from arkouda.client import maxTransferBytes
 from arkouda.row import Row
 from arkouda.alignment import in1dmulti
 from arkouda.series import Series
+from arkouda.index import Index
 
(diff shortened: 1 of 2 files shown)


arkouda/dataframe.py
```diff
@@ -287,14 +294,11 @@ def __getitem__(self, key):
...
+            rtn_data = {}
             s = key
             for k in self._columns:
-                data[k] = UserDict.__getitem__(self, k)[s.start:s.stop:s.step]
-            # result._columns.append(k)
-            # result._empty = False
-            return DataFrame(initialdata=data)
-
+                rtn_data[k] = UserDict.__getitem__(self, k)[s.start:s.stop:s.step]
+            return DataFrame(initialdata=rtn_data, index=list(range(self.size)[s.start:s.stop:s.step]))
```
pierce314159, line 301: i know this is how it was done in the existing code but I don't think you need to rebuild the slice. I think something like this works
```python
UserDict.__getitem__(self, k)[s]
return DataFrame(initialdata=rtn_data, index=arange(self.size)[s])
```
I also think we should use the `ak.arange` function since a list of size `self.size` could be large and `._set_index` will just convert to a pdarray anyway

arkouda/dataframe.py
```diff
@@ -217,25 +223,26 @@ def __init__(self, initialdata=None):
...
             if len(sizes) > 0:
                 self._size = sizes.pop()
-            # If the index column was passed in, use that instead of
+
+            # If the index param was passed in, use that instead of
             # creating a new one.
-            if self.data['index'] is None:
-                self.data['index'] = arange(0, self._size, 1)
+            if self.index is None:
+                self._set_index(arange(0, self._size, 1))
```
pierce314159, line 230: i know this was existing code and certainly does not need to be changed but I wanted to mention it. the start and stride don't need to be specified here since it's the default `arange(self._size)` should work. This happens in a couple of places

arkouda/dataframe.py
```diff
@@ -547,25 +552,26 @@ def drop_duplicates(self, subset=None, keep='first'):
...
+            gp = akGroupBy([self.data[col] for col in subset])
 
         if keep == 'last':
-            _segment_ends = concatenate([_.segments[1:] - 1, array([_.permutation.size - 1])])
-            return self[_.permutation[_segment_ends]]
+            _segment_ends = concatenate([gp.segments[1:] - 1, array([gp.permutation.size - 1])])
+            return self[gp.permutation[_segment_ends]]
         else:
-            return self[_.permutation[_.segments]]
+            print(gp.permutation[gp.segments])
```
pierce314159, line 573: left a print statement

pierce314159 (approved):
This looks good. Left a few comments that are all very minor and most are about things that were already there

arkouda/dataframe.py
Ethan-DeBandi99: removed the print.

arkouda/dataframe.py
Ethan-DeBandi99: I left this for now since it is not impacting functionality. I added issue #1236 to track.

pierce314159 (changes requested):
Updating to Request Changes. Even though it's a small thing, I do think this is worth fixing to avoid having a python copy of index

> i know this is how it was done in the existing code but I don't think you need to rebuild the slice. I think something like this works
```python
UserDict.__getitem__(self, k)[s]
return DataFrame(initialdata=rtn_data, index=arange(self.size)[s])
```
>I also think we should use the `ak.arange` function since a list of size `self.size` could be large and `._set_index` will just convert to a pdarray anyway

arkouda/dataframe.py
Ethan-DeBandi99: I updated this to the requested configuration. Sorry I totally missed this when I was making changes this morning!

pierce314159 (approved):
Looks good!

PR #1244, opened 2022-03-30
Add null checking for strings

bmcdonald3:
This PR adds a check for if `values_read` is greater than 0
before attempting to store the value into the Arkouda array
because a value of 0 represents either a null value or a problem
reading.

Closes https://github.com/Bears-R-Us/arkouda/issues/1247


reuster986 (approved):
This looks great! The way you implemented it (assigning null values the empty string) is perfect. While I won't be able to test it until tomorrow, I think null handling is something arkouda should have anyway, even if it doesn't resolve the current issue. We're bound to encounter files (like your test file) with null strings at some point. 

For that reason, I'm actually in favor of merging this PR now, regardless of the outcome of my test.

PR #1248, opened 2022-03-31
Closes #1241: Use `computeOnSegments` for `substringSearch`

### LINKED ISSUE #1241: Use `computeOnSegments` for `substringSearch`
Use `computeOnSegments` for `substringSearch`

pierce314159:
This PR (Closes #1241):
- Updates `substringSearch` to use `computeOnSegments`

Edit: The NOTE should be resolved thanks to ronawho's suggestion

<details>
  <summary>NOTE: </summary>

The way I've done this recompiles the regex pattern each iteration instead of (I believe) once per task using `with`
```Chapel
forall ... with (var myRegex = _unsafeCompileRegex(pattern))
```
I'm not sure if this is better or worse/if there's a better way to do this. Feedback on this would be appreciated
</details>

### DIFF
--- a/src/SegmentedComputation.chpl
+++ b/src/SegmentedComputation.chpl
@@ -44,6 +44,7 @@ module SegmentedComputation {
     StringToNumericReturnValidity,
     StringCompareLiteralEq,
     StringCompareLiteralNeq,
+    StringSearch,
   }
   
   proc computeOnSegments(segments: [?D] int, values: [?vD] ?t, param function: SegFunction, type retType, const strArg: string = "") throws {
@@ -89,6 +90,9 @@ module SegmentedComputation {
               }
               when SegFunction.StringCompareLiteralNeq {
                 agg.copy(res[i], stringCompareLiteralNeq(values, start..#len, strArg));
+              }
+              when SegFunction.StringSearch {
+                agg.copy(res[i], stringSearch(values, start..#len, strArg));
               }
               otherwise {
                 compilerError("Unrecognized segmented function");
--- a/src/SegmentedComputation.chpl
+++ b/src/SegmentedComputation.chpl
@@ -71,31 +71,34 @@ module SegmentedComputation {
         }
         try {
           // Apply function to bytes of each owned segment, aggregating return value to res
-          forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType)) {
-            select function {
-              when SegFunction.SipHash128 {
-                agg.copy(res[i], sipHash128(values, start..#len));
-              }
-              when SegFunction.StringToNumericStrict {
-                agg.copy(res[i], stringToNumericStrict(values, start..#len, retType));
-              }
-              when SegFunction.StringToNumericIgnore {
-                agg.copy(res[i], stringToNumericIgnore(values, start..#len, retType));
-              }
-              when SegFunction.StringToNumericReturnValidity {
-                agg.copy(res[i], stringToNumericReturnValidity(values, start..#len, retType[0]));
-              }
-              when SegFunction.StringCompareLiteralEq {
-                agg.copy(res[i], stringCompareLiteralEq(values, start..#len, strArg));
-              }
-              when SegFunction.StringCompareLiteralNeq {
-                agg.copy(res[i], stringCompareLiteralNeq(values, start..#len, strArg));
-              }
-              when SegFunction.StringSearch {
-                agg.copy(res[i], stringSearch(values, start..#len, strArg));
-              }
-              otherwise {
-                compilerError("Unrecognized segmented function");
+          if function == SegFunction.StringSearch {
+            forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType), var myRegex = _unsafeCompileRegex(strArg)) {
+              agg.copy(res[i], stringSearch(values, start..#len, myRegex));
+            }
+          } else {
+            forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType)) {
+              select function {
+                when SegFunction.SipHash128 {
+                  agg.copy(res[i], sipHash128(values, start..#len));
+                }
+                when SegFunction.StringToNumericStrict {
+                  agg.copy(res[i], stringToNumericStrict(values, start..#len, retType));
+                }
+                when SegFunction.StringToNumericIgnore {
+                  agg.copy(res[i], stringToNumericIgnore(values, start..#len, retType));
+                }
+                when SegFunction.StringToNumericReturnValidity {
+                  agg.copy(res[i], stringToNumericReturnValidity(values, start..#len, retType[0]));
+                }
+                when SegFunction.StringCompareLiteralEq {
+                  agg.copy(res[i], stringCompareLiteralEq(values, start..#len, strArg));
+                }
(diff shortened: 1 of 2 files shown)


src/SegmentedComputation.chpl
```diff
@@ -70,28 +71,34 @@ module SegmentedComputation {
...
-              when SegFunction.StringCompareLiteralNeq {
-                agg.copy(res[i], stringCompareLiteralNeq(values, start..#len, strArg));
-              }
-              otherwise {
-                compilerError("Unrecognized segmented function");
+          if function == SegFunction.StringSearch {
+            forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType), var myRegex = _unsafeCompileRegex(strArg)) {
+              agg.copy(res[i], stringSearch(values, start..#len, myRegex));
+            }
+          } else {
```
Ethan-DeBandi99, line 78: This is mostly a question for furthering my understanding of Chapel. Is breaking the `StringSearch` function into its own `if` block more efficient because the regex is defined in the `with`? Since we already checked and compiled the regex, I would think that this would have little impact but may be wrong. I am just thinking that moving this to the `select` block below would read a little easier, but if there is a performance reason, this works.

Adding code for clarity if this were to be moved into select.
```chapel
select function {
    when SegFunction..StringSearch {
        var myRegex = _unsafeCompileRegex(strArg)
        agg.copy(res[i], stringSearch(values, start..#len, myRegex));
    }
}
```

Ethan-DeBandi99 (approved):
This looks good. I am approving as a the comment I did make is a question for my understanding. Although, if there is no performance gain, I think removing the extra `forall` definition would be cleaner. Overall, nothing should negatively impacted functionality.

src/SegmentedComputation.chpl
pierce314159: We spoke a bit about this offline but you're exactly right that this is done for performance reasons. There's kinda 2 ways to have it all under the same `forall` which each have their drawbacks
- We obviously don't want to compile the regex for the `SegFunction`s that don't require it. This is just unnecessary computation (so just adding the `with var myRegex` to the forall for everyone wouldn't work)
- What you suggest, falls under the second. By using `forall ...with var myRegex`, we compile the regex once per task. Doing this compile inside the forall would compile the regex every iteration of the forall loop. This is essentially what i did originally before using ronawho's suggestion above

in `SegmentedComputation`
```chapel
forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType)) {
...
agg.copy(res[i], stringSearch(values, start..#len, strArg));
}
```
in `SegmentedArray`
```chapel
inline proc stringSearch(values, rng, pattern) throws {
  return checkCompile(pattern).search(interpretAsString(values, rng, borrow=true)).matched;
}
```

ronawho: Right, compiling the regex just once per task is important for performance (generally speaking compiling regex's is expensive, but using them once compiled is cheap -- this is true for all regex libraries I know of, not just ours.)

This current approach looks good to me, but I think a potential alternative would be to collapse back down into a single forall and have a default-init'd regex for the non-regex paths. 

It's worth noting that when Arkouda regex's were first coming online that wouldn't have compiled, but it should now. e.g. the following wouldn't have worked:


```chpl
var myRegex = _unsafeCompileRegex(strArg)
forall ... {
    agg.copy(res[i], stringSearch(values, start..#len, myRegex));
}
```

but that was fixed in the 1.25.1 timeframe. So once that's the minimum version supported, maybe this code could be cleaned up a bit. Here's the diff compared to the current PR that I tested:

<details>

```diff
diff --git a/src/SegmentedArray.chpl b/src/SegmentedArray.chpl
index b4a18084..394bfd2b 100644
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -637,8 +637,8 @@ module SegmentedArray {
     */
     proc substringSearch(const pattern: string) throws {
       // We need to check that pattern compiles once to avoid server crash from !try in _unsafecompile
-      checkCompile(pattern);
-      return computeOnSegments(offsets.a, values.a, SegFunction.StringSearch, bool, pattern);
+      const regexPattern = checkCompile(pattern);
+      return computeOnSegments(offsets.a, values.a, SegFunction.StringSearch, bool, regexArg=regexPattern);
     }
 
     /*
diff --git a/src/SegmentedComputation.chpl b/src/SegmentedComputation.chpl
index fb8cf28d..ef81b1ae 100644
--- a/src/SegmentedComputation.chpl
+++ b/src/SegmentedComputation.chpl
@@ -4,6 +4,7 @@ module SegmentedComputation {
   use ServerErrors;
   private use Cast;
   use Reflection;
+  use Regex;
 
   proc computeSegmentOwnership(segments: [?D] int, vD) {
     const low = vD.low;
@@ -47,7 +48,7 @@ module SegmentedComputation {
     StringSearch,
   }
   
-  proc computeOnSegments(segments: [?D] int, values: [?vD] ?t, param function: SegFunction, type retType, const strArg: string = "") throws {
+  proc computeOnSegments(segments: [?D] int, values: [?vD] ?t, param function: SegFunction, type retType, const strArg: string = "", regexArg = new regex(string)) throws {
     // type retType = if (function == SegFunction.StringToNumericReturnValidity) then (outType, bool) else outType;
     var res: [D] retType;
     if (D.size == 0) {
@@ -71,11 +72,6 @@ module SegmentedComputation {
         }
         try {
           // Apply function to bytes of each owned segment, aggregating return value to res
-          if function == SegFunction.StringSearch {
-            forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType), var myRegex = _unsafeCompileRegex(strArg)) {
-              agg.copy(res[i], stringSearch(values, start..#len, myRegex));
-            }
-          } else {
             forall (start, len, i) in zip(mySegs, myLens, mySegInds) with (var agg = newDstAggregator(retType)) {
               select function {
                 when SegFunction.SipHash128 {
@@ -96,12 +92,14 @@ module SegmentedComputation {
                 when SegFunction.StringCompareLiteralNeq {
                   agg.copy(res[i], stringCompareLiteralNeq(values, start..#len, strArg));
                 }
+                when SegFunction.StringSearch {
+                  agg.copy(res[i], stringSearch(values, start..#len, regexArg));
+                }
                 otherwise {
                   compilerError("Unrecognized segmented function");
                 }
               }
             }
-          }
         } catch {
           throw new owned ErrorWithContext("Error computing %s on string or segment".format(function:string),
                                            getLineNumber(),
```

</details>

But for now I'd stick with the approach in this PR.

ronawho (approved):

ronawho:
FYI here are perf results for 16-node-cs-hdr for the str-locality test:

```
# before:
./benchmarks/run_benchmarks.py -nl 16 str-locality
Regex searching good locality Average rate = 40.37 GiB/sec
Regex searching poor locality Average rate =  0.07 GiB/sec

# now:
./benchmarks/run_benchmarks.py -nl 16 str-locality
Regex searching good locality Average rate = 17.96 GiB/sec
Regex searching poor locality Average rate = 11.96 GiB/sec

# now with bumped problem size:
./benchmarks/run_benchmarks.py -nl 16 str-locality --size=$((10**8))
Regex searching good locality Average rate = 25.58 GiB/sec
Regex searching poor locality Average rate = 13.08 GiB/sec
```

(and this is what we expect from computeOnSegments, worse performance with good locality, but much, much better performance with poor locality)

Issue #1249, opened 2022-04-01
Remove snappy compression on null strings test file


bmcdonald3:
The null strings test file that was initially added has snappy compression, which is supported by Arkouda, but some nightly tests currently use an Arrow version that does not have snappy installed, so are failing due to that. Since snappy isn't essential to this test, it seems the easiest short-term fix is to just use a new file that is uncompressed, but we have plans to update the Arrow install soon.

PR #1250, opened 2022-04-01
Update null test file to be uncompressed

bmcdonald3:
This PR replaces the null Parquet test file to be uncompressed.

The previous null test file was generated with Snappy compression
because that is the default for pyarrow, but that was not an
essential part of the test and means that Arrow must have been
installed with the optional Snappy build component, which is
supported by Arkouda today, but can cause some annoyance for users
who may have built Arrow without Snappy support.

Closes: https://github.com/Bears-R-Us/arkouda/issues/1249


mhmerrill (approved):
looks ok to me.

Issue #1261, opened 2022-04-11
Server crash when reading Parquet string columns


bmcdonald3:
The server crashes sometimes when a single file is shared between two locales and a string column is read from the file.

PR #1262, opened 2022-04-11
Fix server crash when reading string columns from multiple locales

bmcdonald3:
There was a bug caught with the Parquet string reading where
reading a string column from a single file that was shared between
multiple locales was causing a crash. I believe this was caused by
a segfault in the C++ code where the buffer allocated on the
Chapel size was too small due to incorrectly handling null bytes.
This PR simplifies the string reading Parquet code to avoid this
type of problem.

This partially reverts an optimization made in https://github.com/Bears-R-Us/arkouda/pull/1204, so we expect that
performance will take a minor hit.

Closes https://github.com/Bears-R-Us/arkouda/issues/1261


reuster986 (approved):
Thanks for all your help diagnosing and fixing this!

Ethan-DeBandi99 (approved):
Looks good

Issue #1268, opened 2022-04-13
arkouda.Strings object being initialized with name = None


joshmarshall1:
Found during testing of #1086 

The __init__ of the Strings class is setting the name of all objects to None: `self.name: Optional[str] = None` which results in references to the name of a newly created Strings object to be None instead of their server-generated name.

Proposed change is to move the `self.name: Optional[str] = None` into the try block and update it to be `self.name: Optional[str] = self.entry.name` which will set the newly initialized Strings object to the name associated with the strings_pdarray it's created from.

PR #1269, opened 2022-04-13
Closes #1086, Generic Attach Method

### LINKED ISSUE #1086: Create a generic `ak.attach()`
Currently, when attaching to registered arrays, the user must know ahead of time whether the array is a pdarray or Strings and call either `ak.attach_pdarray()` or `ak.Strings.attach()` accordingly. This makes programmatic attachment to multiple arrays difficult. For example, we'd like to be able to create a DataFrame with both string and numeric columns, register the columns with a common naming scheme, and then later attach to all of them to recover the DataFrame. See the `register_all` and `attach_all` functions in `akutil/util.py` for usage.

However, this would require a generic `ak.attach()` function that runs the "attach" server command and parses the output to determine whether it is a pdarray or Strings and returns the appropriate type. I believe this can be done with:

```python
def attach(name):
    repMsg = generic_msg(cmd="attach", args=name)
    dtype = repMsg.split()[2]
    if dtype == "str":
        return Strings.from_return_msg(repMsg)
    else:
        return create_pdarray(repMsg)
```

joshmarshall1:
This PR (closes #1086):

Added the method outlined in the original issue to arkouda/util.py. During testing I found that on creation of a Strings object, .name is always equal to None - created issue #1268 for this. Until that issue is worked, to get the name of a Strings object you have to use `<Strings object>.entry.name`.

Example of the functionality:
```python
import arkouda as ak

ak.connect()

a = ak.arange(5)
a_name = a.name

s = ak.array(["abc","def","123"])
s_name = s.entry.name

ak.diconnect()
ak.connect()

a2 = ak.util.attach(a_name)
s2 = ak.util.attach(s_name)

```

### DIFF
--- a/arkouda/util.py
+++ b/arkouda/util.py
@@ -4,15 +4,15 @@ import numpy as np  # type: ignore
 import h5py #type: ignore
 import os
 
-from arkouda import __version__
+from arkouda import __version__, Strings
 from arkouda.client_dtypes import BitVector, BitVectorizer, IPv4
 from arkouda.timeclass import Datetime, Timedelta
-from arkouda.pdarrayclass import attach_pdarray, pdarray
+from arkouda.pdarrayclass import attach_pdarray, pdarray, create_pdarray
 from arkouda.pdarraysetops import concatenate as pdarrayconcatenate
 from arkouda.pdarraycreation import arange
 from arkouda.pdarraysetops import unique
 from arkouda.pdarrayIO import read_hdf
-from arkouda.client import get_config, get_mem_used
+from arkouda.client import get_config, get_mem_used, generic_msg
 from arkouda.groupbyclass import GroupBy, broadcast, coargsort
 from arkouda.infoclass import information, AllSymbols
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good to me.

Ethan-DeBandi99:
Extending it should be fairly easy. One note though, @pierce314159 and I had removed the `SegArray` register/attach when it was moved to arkouda. I think because it was not server side, but reviewing it now, that should not be an issue. I will add an issue to get that added into `ak.SegArray` and subsequently add it to the generic attach we are adding here.

I think a unit test would be good just to ensure we don't have any issues.

joshmarshall1:
After looking into adding `Categorical` support to the generic attach method, and a discussion with @Ethan-DeBandi99, we found it will be more complicated to add than we initially thought due to how Categoricals are registered. Ticket #1284 will support this addition and will be updated once we have a plan on how to handle this correctly.

PR #1271, opened 2022-04-13
Closes #1268, Strings object name initialization

### LINKED ISSUE #1268: arkouda.Strings object being initialized with name = None

joshmarshall1:
This PR (closes #1268):

Moving the name initialization into the try block and setting it to the passed pdarray's name value allows for a generic reference to the name of the object regardless of if the object is a Strings object or pdarray.

Existing references to the Strings object entry.name value will still function, i.e. the name of Strings object 's' can now be accessed through `s.name` or `s.entry.name`. Both of these methods will return the same value.

This solves the issue of using the newly created generic attach method still requiring knowledge of if the array is a pdarray or Strings object in order to get the name of the Strings object correctly, instead of "None", when attaching to an element. 

Example: 
```python
import arkouda as ak

ak.connect()

a = ak.arange(5)
s = ak.array(["abc","123","def"])

a_name = a.name
#Originally, this would set s_name = None
s_name = s.name

```

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -168,11 +168,11 @@ class Strings:
             self.nbytes = bytes_size  # This is a deficiency of server GenSymEntry right now
             self.ndim = self.entry.ndim
             self.shape = self.entry.shape
+            self.name: Optional[str] = self.entry.name
         except Exception as e:
             raise ValueError(e)   
 
         self.dtype = npstr
-        self.name:Optional[str] = None
         self._regex_dict: Dict = dict()
         self.logger = getArkoudaLogger(name=__class__.__name__) # type: ignore
 


Ethan-DeBandi99 (approved):
This looks good to me. @pierce314159 you have more background on strings than I do. Was there a reason we were setting the name to None for String objects?

pierce314159:
> This looks good to me. @pierce314159 you have more background on strings than I do. Was there a reason we were setting the name to None for String objects?

If i remember correctly, there wasn't a `Strings.name` attribute at all originally. When we added it `Strings` was not yet a full fledged entry in the symbol table (so we didn't have an entry/entry.name yet) and instead we had two pdarray components we were keeping up with (offsets and values). So it wasn't clear which component name should be the name of the `Strings`

This could be entirely wrong, I think @glitch was the one who added `Strings.name` and made the switch to composite entries existing in the symbol table. So maybe he can correct me

pierce314159 (approved):
I think this looks good

reuster986 (approved):

PR #1274, opened 2022-04-13
Closes #1265: Add `to/is_lower` and `to/is_upper` methods to `Strings`

### LINKED ISSUE #1265: Add `.toLower()` and `.toUpper()` methods to Strings
We would like to be able to convert strings to all lowercase or all uppercase. For now, this can just be in the ASCII range, as I'm not even sure how to determine case in broader UTF-8.

pierce314159:
This PR (Closes #1265):
- Adds `Strings.to_lower()`
  - Returns a new Strings with all uppercase characters from the original replaced with their lowercase equivalent
- Adds `Strings.to_upper()`
  - Returns a new Strings with all lowercase characters from the original replaced with their uppercase equivalent
- Adds `Strings.is_lower()`
  - Returns a boolean pdarray where index `i` indicates whether string `i` of the Strings is entirely lowercase
- Adds `Strings.is_upper()`
  - Returns a boolean pdarray where index `i` indicates whether string `i` of the Strings is entirely uppercase
- Adds a test for the new functionality

NOTES:
- This PR follows the basic outline laid out by @ronawho in the issue and returns a new Strings. I can work on a version that modifies in-place if we want to go in that direction
- `is_lower/upper` takes advantage of `computeOnSegments`, but it seemed more involved for `to_lower/upper`. This is something that should be looked into more
- `interpretAsBytes` is moved from `Flatten.chpl` to `SegmentedArray.chpl`. The function is not changed

Examples:
```python
>>> mixed_case = ak.array([f'StrINgS {i}' for i in range(3)])
>>> mixed_case 
array(['StrINgS 0', 'StrINgS 1', 'StrINgS 2'])

>>> lower = mixed_case.to_lower()
>>> lower
array(['strings 0', 'strings 1', 'strings 2'])

>>> upper = mixed_case.to_upper()
>>> upper
array(['STRINGS 0', 'STRINGS 1', 'STRINGS 2'])

>>> strings = ak.concatenate([lower, mixed_case , upper])
>>> strings
array(['strings 0', 'strings 1', 'strings 2', 'StrINgS 0', 'StrINgS 1', 'StrINgS 2', 'STRINGS 0', 'STRINGS 1', 'STRINGS 2'])

>>> strings.is_lower()
array([True True True False False False False False False])

>>> strings.is_upper()
array([False False False False False False True True True])
```

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -327,6 +327,128 @@ class Strings:
         return create_pdarray(generic_msg(cmd=cmd,args=args))
 
     @typechecked
+    def to_lower(self) -> Strings:
+        """
+        Returns a new Strings with all uppercase characters from the original replaced with their lowercase equivalent
+
+        Returns
+        -------
+        Strings
+            Strings with all uppercase characters from the original replaced with their lowercase equivalent
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there is a server-side error thrown
+
+        See Also
+        --------
+        Strings.to_upper
+
+        Examples
+        --------
+        >>> strings = ak.array([f'StrINgS {i}' for i in range(5)])
+        >>> strings
+        array(['StrINgS 0', 'StrINgS 1', 'StrINgS 2', 'StrINgS 3', 'StrINgS 4'])
+        >>> strings.to_lower()
+        array(['strings 0', 'strings 1', 'strings 2', 'strings 3', 'strings 4'])
+        """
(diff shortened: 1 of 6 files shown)


joshmarshall1 (approved):
Looks good. 

Initial thoughts were for the chapel methods `lower()` and `upper()` to be renamed to `to_lower()` and `to_upper()` for consistency, but upon further consideration that change might instead cause confusion with chapel's built in `toLower()` and `toUpper()` methods.

Ethan-DeBandi99 (approved):
I think this is good to go

reuster986 (approved):

PR #1281, opened 2022-04-14
Closes #1236 - Cleanup of arange unnecessary arguments

### LINKED ISSUE #1236: Dataframe `arange()` usage Clean-up
This was identified during review of #1190. There are several instances where `arange()` is being called as `arange(0, self._size, 1)` (or similar). Because `0` is the default for `start` and `1` is the default for `stride`, This can be simplified to `arange(self._size)` since the defaults will automatically be applied.

**Files to Update**
- `arkouda/dataframe.py`

joshmarshall1:
This ticket (closes #1236):

All instances of arange in `arkouda/dataframe.py` have been checked and updated to include only the self._size argument if the leading argument was 0 and the third argument was 1 or there were only two arguments with the leading argument being 0.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -227,7 +227,7 @@ class DataFrame(UserDict):
             # If the index param was passed in, use that instead of
             # creating a new one.
             if self.index is None:
-                self._set_index(arange(0, self._size, 1))
+                self._set_index(arange(self._size))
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

Ethan-DeBandi99 (approved):
Looks good

Issue #1242, opened 2022-03-29
Naming and conventions of generic IO functions 


bmcdonald3:
## Summary
In an attempt to unify the Parquet and HDF5 (and any future supported IO types), we started with the `ls()` function. This function calls a generic server-side function that will first determine the file type and then call the correct function, allowing it to be used on both Parquet and HDF5 files without the user being aware.

Deciding on naming and error handling is a bit more difficult with the read functions that do not always share the same signature and also have some additional error checking.

## Reading files
Here are a couple of proposals for a generic read in Arkouda:
1. Have a single function called just `read()` that is equivalent to today's `read_all()`, but works on both HDF5 and Parquet and requires a hijacking of the existing `read()` function
2. Change `read_all()` to be able to read both file types, so the behavior doesn't change very much from what it is today
3. Create a new function, perhaps `ak.read_any()`, that can read both files and allows us to keep the existing functions around for specific uses

### Allow errors
One of the problems with unifying file read naming is the additional flags, such as `allow_errors`. If a user calls `read_all()` with `allow_errors` today, the files will all be attempted to be read in as an HDF5 file and then reported back if one isn't found (or another error occurs). If we want to support an `allow_errors` like this for a generic function, we may be in a position where we are unable to determine the file type because the files passed to the server do not exist, for example.

The existing prototypes have been basing the filetype off of the first file name passed to Arkouda, and if that filename doesn't exist or there is a problem with the file, then we aren't able to determine the type and `allow_errors` behavior changes slightly. 

Is this a problem? How do we want to handle this case where the first file in the glob expression has an error and `allow_errors` has been passed to the server? (it will likely be slightly different behavior than today, regardless of what is chosen).

## Writing files
Today, we have a `save()` and a `save_parquet()`. Of course, we can't determine for the user whether or not an HDF5 or a Parquet file should be written, but are we happy with those names? Here are a few proposals:
1. Separate `save()` and `save_parquet()` functions for HDF5 and Parquet (this is what we have today)
2. Separate `save_hdf()` and `save_parquet()` functions
3. One `save()` function that accepts an `is_parquet` (or other argument) to determine file type

bmcdonald3:
Hi @mhmerrill @reuster986 I've put together a few proposals on IO unification naming here. Let me know if you have an opinion on any of these. Sorry the issue ended up being a bit lengthy. Thanks.

reuster986:
[...]

PR #1288, opened 2022-04-14
Unify various file reading functions into a single read function

bmcdonald3:
This PR takes `read_hdf()`, `read_all()`, and `read_parquet()` and
combines them into a single `read()` function. This new function is
identical to the old `read_all()` function with the exception of the
`file_format` keyword argument. This defaults to `"infer"`, which
means that the server will determine the type of the file and it does
not need to be specified by the user.

This file type inference changes certain error checking behaviors
(e.g., if you have an error in one of your filenames, this will behave
differently that before with `allow_errors=True`). To still allow the
old functionality, the `file_format` keyword argument can be explicitly
set to either `HDF5` or `Parquet` (case insensitive), and it will
behave as before.

Additionally, an `ak.get_filetype()` function has been added that will
return the type of a file, or the type of the first file in a glob
expression.

Part of https://github.com/Bears-R-Us/arkouda/issues/1242

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -270,7 +270,7 @@ def get_datasets(filename : str) -> List[str]:
         datasets.remove(ARKOUDA_HDF5_FILE_METADATA_GROUP)
     return datasets
 
-def get_filetype(filename: Union[str, list[str]]) -> str:
+def get_filetype(filenames: Union[str, List[str]]) -> str:
     """
     Get the type of a file accessible to the server. Supported
     file types and possible return strings are 'HDF5' and 'Parquet'.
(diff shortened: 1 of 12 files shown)


arkouda/pdarrayIO.py
```diff
@@ -492,7 +369,7 @@ def load_all(path_prefix: str) -> Mapping[str, Union[pdarray, Strings, Categoric
 
     See Also
     --------
-    save_all, load, read_hdf, read_all
+    save_all, load, read, read
```
Ethan-DeBandi99, line 372: Is read meant to be listed twice?

Ethan-DeBandi99 (approved):
This looks good. This solution seems much cleaner than what we had in place before.

arkouda/pdarrayIO.py
bmcdonald3: Ah, good catch -- I did a find and replace on both `read_hdf` and `read_all`, so that is why there is a double here (and probably elsewhere)

joshmarshall1 (approved):

PR #1292, opened 2022-04-15
Closes #1286: `test_broadcast_*` issue in `groupby_test`

### LINKED ISSUE #1286: Issue in `test_broadcast_*` in `groupby_test`
https://github.com/Bears-R-Us/arkouda/blob/78d3104a1211de039e5e84a88d987c99fe6bc996/tests/groupby_test.py#L169-L182
https://github.com/Bears-R-Us/arkouda/blob/78d3104a1211de039e5e84a88d987c99fe6bc996/tests/groupby_test.py#L184-L197

These tests aren't actually checking the equality of the arrays, they need to be updated to this form
```python
self.assertTrue((np.array(...) == results.to_ndarray()).all())
```
They also seem to be failing when updated to this form, this definitely needs to be investigated more

This was discovered in PR #1283, this applies to the newly added `test_broadcast_uints`

pierce314159:
This PR (closes #1286):
- Updates `test_broadcast_ints` and `test_broadcast_booleans` to actually compare against the result array by following the format
```python
self.assertTrue((np.array(...) == results.to_ndarray()).all())
```
- The tests were incorrectly comparing the non-permuted expected against the result from `permute=True`. Updated tests to compare againt `permute=True` and `permute=False`

### DIFF
--- a/tests/groupby_test.py
+++ b/tests/groupby_test.py
@@ -185,19 +185,28 @@ class GroupByTest(ArkoudaTest):
         self.assertTrue((bcast[:-1] >= bcast[1:]).all())
         
     def test_broadcast_ints(self):
-        keys,counts = self.igb.count()
+        keys, counts = self.igb.count()
 
-        self.assertTrue((np.array([1,4,2,1,2]) == counts.to_ndarray()).all())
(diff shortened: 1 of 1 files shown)


joshmarshall1 (approved):

Ethan-DeBandi99 (approved):
Looks good to me

PR #1293, opened 2022-04-15
Add a file_format keyword argument to save

bmcdonald3:
This PR adds a `file_format` keyword argument to `ak.save()`.
Accepted values are `HDF5` and `Parquet` and the default value
is `HDF5`.

Additionally, the `save_parquet()` function is now just a wrapper
over `ak.save()` and a similar `save_hdf()` function has been
added for HDF5.

Part of https://github.com/Bears-R-Us/arkouda/issues/1242


arkouda/pdarrayclass.py
```diff
@@ -1137,28 +1158,80 @@ def save_parquet(self, prefix_path : str, dataset : str='array', mode : str='tru
...
+        Examples
+        --------
+        >>> a = ak.arange(0, 100, 1)
+        >>> a.save_hdf('arkouda_range')
+
+        Array is saved in numLocales files with names like ``arkouda_range_LOCALE0000``
+
+        The array can be read back in as follows
+
+        >>> b = ak.read_all('arkouda_range')
```
joshmarshall1, line 1230: As above for `read_all()`

joshmarshall1 (changes requested):
All looks good aside from the docstrings listing `read_all` `read_hdf` and `read_parquet` which are being removed in PR #1288 I would recommend updating these to have `read()` instead to reflect those changed.

bmcdonald3:
The reason that I chose to leave those in is that if I change them in this PR as well as in the other one then we will have merge conflicts with the 2 PRs. They are updated in the other PR, so once both branches are merged, that should be resolved I believe (unless I am missing something).

joshmarshall1:
That makes sense, my mistake. I saw they were highlighted green so didn't suspect there would be overlap with the other PR

joshmarshall1 (approved):

bmcdonald3:
Oh, no you are right, good catch, I dismissed that too quickly, my mistake. The new docstrings that I added won't be conflicting and won't change with the other PR, so those are what need to change -- thanks.

arkouda/pdarrayclass.py
```diff
@@ -1045,7 +1045,7 @@ def save(self, prefix_path : str, dataset : str='array', mode : str='truncate',
 
         See Also
         --------
-        save_all, load, read_hdf, read_all
+        save_all, load, read_hdf, read
```
joshmarshall1, line 1048: `read_hdf` should also be removed here

joshmarshall1 (changes requested):
It looks like you missed one. The rest look good

joshmarshall1 (approved):

Ethan-DeBandi99 (approved):
Looks good

joshmarshall1 (approved):

mhmerrill:
@bmcdonald3 where is the corresponding issue?

PR #1296, opened 2022-04-18
Closes #1229 and #1295 - Updating save checks in parquet/hdf5 msg files to be more efficient

### LINKED ISSUE #1229: Parquet/HDF5 Write Validate using `matchingFilenames`
This issue was identified during PR review, #1227. 

`processParquetFilenames` takes in `filenames` and `matchingFilenames`. `matchingFilenames` is computed via
```python
var matchingFilenames = glob("%s_LOCALE*%s".format(filename, ".parquet"));
```
and then passed into `processParquetFilenames()`.  The function then loops `filenames` and checks for existence. This appears to be unnecessary because the checks utilizing the values set in the loop can be updated:
```python
# FROM
    if !anyexist {
            throw getErrorWithContext(
               msg="Cannot append a non-existent file, please save without mode='append'",
               lineNumber=getLineNumber(), 
               routineName=getRoutineName(), 
               moduleName=getModuleName(), 
               errorClass='WriteModeError'
                                      );
    }
    if !allexist || (matchingFilenames.size != filenames.size) {
      throw getErrorWithContext(
                 msg="appending to existing files must be done with the same number " +
                    "of locales. Try saving with a different directory or filename prefix?",
                 lineNumber=getLineNumber(), 
                 routineName=getRoutineName(), 
                 moduleName=getModuleName(), 
                 errorClass='MismatchedAppendError'
            );
    }

# TO
    if matchingFilenames.size == 0 {
              throw getErrorWithContext(
                 msg="Cannot append a non-existent file, please save without mode='append'",
                 lineNumber=getLineNumber(), 
                 routineName=getRoutineName(), 
                 moduleName=getModuleName(), 
                 errorClass='WriteModeError'
                                        );
      }
      if matchingFilenames.size != filenames.size{
        throw getErrorWithContext(
                   msg="appending to existing files must be done with the same number " +
                      "of locales. Try saving with a different directory or filename prefix?",
                   lineNumber=getLineNumber(), 
                   routineName=getRoutineName(), 
                   moduleName=getModuleName(), 
                   errorClass='MismatchedAppendError'
              );
      }
```
This code was designed using HDF5 as a model and HDF5 will require the same type of modification.

### LINKED ISSUE #1295: Attempting to append to a non-existant Parquet file throwing incorrect error
Found while testing #1229 and determined to be pre-existing.

When attempting to append to a non-existant file and error message of `Cannot append a non-existent file, please save without mode='append'` is expected. Instead, only for pdarrays being saved to parquet format, an error of `Invalid argument: Argument type mismatch in argument 0 (in string.format) ` is being thrown.

To reproduce:
```python
import arkouda as ak
ak.connect()
a = ak.arange(25)
#Will only error if there is no _LOCALE0000.parquet file in the directory
a.save_parquet("/home/josh/testing/", "1229_pdarray", "append")
```

This results in:
```python
Input In [4], in <cell line: 1>()
----> 1 a.save_parquet("/home/josh/testing/", "1229_pdarray", "append")

File ~/anaconda3/envs/test-env/lib/python3.8/site-packages/typeguard/__init__.py:1015, in typechecked.<locals>.wrapper(*args, **kwargs)
   1013 memo = _CallMemo(python_func, _localns, args=args, kwargs=kwargs)
   1014 check_argument_types(memo)
-> 1015 retval = func(*args, **kwargs)
   1016 try:
   1017     check_return_type(retval, memo)

File ~/git/arkouda/arkouda/pdarrayclass.py:1159, in pdarray.save_parquet(self, prefix_path, dataset, mode, compressed)
   1157 except Exception as e:
   1158     raise ValueError(e)
-> 1159 return cast(str, generic_msg(cmd="writeParquet", args="{} {} {} {} {} {}".\
   1160                              format(self.name, dataset, m, json_array, self.dtype,
   1161                                     compressed)))

File ~/git/arkouda/arkouda/client.py:548, in generic_msg(cmd, args, payload, send_binary, recv_binary)
    546     else:
    547         assert payload is None
--> 548         return _send_string_message(cmd=cmd, args=args, recv_binary=recv_binary)
    550 except KeyboardInterrupt as e:
    551     # if the user interrupts during command execution, the socket gets out
    552     # of sync reset the socket before raising the interrupt exception
    553     socket = context.socket(zmq.REQ)

File ~/git/arkouda/arkouda/client.py:351, in _send_string_message(cmd, recv_binary, args)
    349 # raise errors or warnings sent back from the server
    350 if return_message.msgType == MessageType.ERROR:
--> 351     raise RuntimeError(return_message.msg)
    352 elif return_message.msgType == MessageType.WARNING:
    353     warnings.warn(return_message.msg)

RuntimeError: Invalid argument: Argument type mismatch in argument 0 (in string.format)
```

joshmarshall1:
This ticket (closes #1229 and closes #1295):

Updated the `processParquetFilenames()` and `processFilenames()` methods to reflect the recommended updates in #1229 . This resulted in the removal of a `for` loop and a few variables that were no longer required in both files. 

In addition, during testing I found that attempting to save in append mode to a non-existing files was throwing the wrong error message and in further debugging, I found the bug to be pre-existing in the codebase. Issue #1295 describes this bug.

With Ethan's assistance we debugged the root cause of the incorrect error and corrected it in this PR. The issue was due to trying to use a non-existant filename to check for the existence of an array within the file before checking if the file itself existed. The solution was to move the file check above the array check so it would error correctly.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -1318,23 +1318,13 @@ module HDF5Msg {
        * meaning that 1..n files will be overwritten.
        */
       if (mode == APPEND) {
-          var allexist = true;
-          var anyexist = false;
-          
-          for f in filenames {
-              var result =  try! exists(f);
-              allexist &= result;
-              if result {
-                  anyexist = true;
-              }
-          }
 
           /*
            * Check to see if any exist. If not, this means the user is attempting to append
            * to 1..n files that don't exist. In this situation, the user is alerted that
            * the dataset must be saved in TRUNCATE mode.
            */
-          if !anyexist {
+          if matchingFilenames.size == 0 {
               throw getErrorWithContext(
                  msg="Cannot append a non-existent file, please save without mode='append'",
                  lineNumber=getLineNumber(), 
@@ -1350,7 +1340,7 @@ module HDF5Msg {
            * a file append is attempted where the number of locales between the file 
            * creates and updates changes.
            */
-          if !allexist || (matchingFilenames.size != filenames.size) {
+          if matchingFilenames.size != filenames.size {
               throw getErrorWithContext(
                    msg="appending to existing files must be done with the same number " +
                       "of locales. Try saving with a different directory or filename prefix?",
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -281,6 +281,11 @@ module ParquetMsg {
       var suffix = '%04i'.format(i): string;
       filenames[i] = filename + "_LOCALE" + suffix + ".parquet";
     }
+    
+    var matchingFilenames = glob("%s_LOCALE*%s".format(filename, ".parquet"));
+
+    var warnFlag = processParquetFilenames(filenames, matchingFilenames, mode);
+
     if mode == APPEND {
       var datasets = getDatasets(filenames[0]);
       if datasets.contains(dsetname) then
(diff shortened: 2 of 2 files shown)


bmcdonald3 (approved):
Looks good 👍

Ethan-DeBandi99 (approved):
Looks good. Good catch on the append issue with non-existing files.

Issue #1298, opened 2022-04-18
Add a function to get the null indices from a Parquet string column


bmcdonald3:
Currently, Parquet null strings are just read in as empty strings into Arkouda, so there is no way to know if that was really just an empty string, or a null value.

If there was a way to get the null indices from the file, that could provide valuable information as well as help debug and figure out where things could be going wrong.

In the Parquet file format, null values and empty strings are two distinct concepts, so having a distinction between them can be helpful in identifying potential issues and keeping the two separate. Null values represent an absence of data, whereas empty strings represent a 0-length string.

mhmerrill:
So does Parquet format have a `null` value concept for all data types i.e. `int` and `float`?

bmcdonald3:
Yes, it does, but it is not supported in all the implementations. The Spark and Go implementations support `null` values on numeric columns, but I don't think the C version and some of the other language implementations do.

If we were to add explicit handling of the nulls for the numeric columns, it would be a bit of a performance hit, so I think it is a bit of a tradeoff there if we want to support the null reading in the numeric columns preemptively. I am open to both, but I would definitely lean towards only adding that support if/when it is definitely needed.

PR #1299, opened 2022-04-18
Add a get_null_indices function for Parquet files

bmcdonald3:
This PR adds support for an `ak.get_null_indices()` function that
accepts a list of Parquet filenames and a list of datasets and will
return a pdarray representing the null indices of that column. This
currently only works with String Parquet columns.

In the returned array, 0 represents a non-null value and 1 represents
a null index.

Closes https://github.com/Bears-R-Us/arkouda/issues/1298


Ethan-DeBandi99 (approved):
I don't see any issues.

pierce314159 (approved):
I don't see any issues with this, nice work!

Issue #1306, opened 2022-04-19
Strings object did not get updated `save()` method


bmcdonald3:
https://github.com/Bears-R-Us/arkouda/pull/1293 added the `file_format` keyword to `save()` to simplify the Parquet/HDF5 API, but I forgot to update the `strings.py` file to reflect this change. This function should be updated to have the same API as the one in `pdarrayclass.py`.

While in this code, I noticed that the strings object `save` and the pdarray `save` are passing commands with differing lengths to the server and I don't think that is what we want, so I am also hoping to fix that message handling.

PR #1307, opened 2022-04-19
Update strings object to match new save function convention

bmcdonald3:
In #1293, the `pdarray.save()` method was updated with a `file_format`
keyword that allowed specification of Parquet or HDF5 files to make
the API simpler to use. In that PR, strings were overlooked and had
not been updated to match this new API design.

This PR updates the `strings.save()` function to use a `file_format`
keyword argument like the other PR and also adds `save_hdf` and
`save_parquet` wrappers around that function.

Additionally documentation has been updated and a discrepency between
the HDF5 strings/pdarray `tohdf` command arguments was cleaned up. An
unused argument was removed and the strings and pdarray message
arguments now send the same number of arguments (7) to the server,
whereas they used to send differing numbers (6 for pdarrays and 8
for strings).

Closes https://github.com/Bears-R-Us/arkouda/issues/1306

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -1019,9 +1019,9 @@ class pdarray:
             By default, truncate (overwrite) output files, if they exist.
             If 'append', attempt to create new dataset in existing files.
         compressed : bool
-            Arkouda only supports writing with compression on Parquet files.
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99 (approved):
Looks good

PR #1309, opened 2022-04-20
Closes #1303 save pdarray parquet

### LINKED ISSUE #1303: write dataframe or multiple aligned pdarrays into parquet file(s) as a Table
I need to save multiple columns/arrays in parquet so that it can be read back as a Table. I attempted to save the pdarrays consecutively using foo.save_parquet(... mode='append'), but I kept getting errors. Even if those errors were fixed, I'm not sure it would solve my problem. I need to read the data into another application that is expecting Table objects. However, even if we ignore my specific application, I believe that most producers/consumers of parquet format files would expect Table objects rather than individual arrays.

Ethan-DeBandi99:
This PR (closes #1303):

`ak.DataFrame` has been updated and now includes a `save_table()` method. This method works by calling `save_all()` on the data mapping. The `Index` is added to the data mapping if a users sets `index=True`.

`pdarrayIO.save_all()` has been updated to now allow saving as parquet or hdf5. This allows the user to save a list of `pdarray` to a file. Testing only provided for Parquet at this time. `pd.read_parquet()` allows use to provide a directory to read all the files in for multilocale. 

Everything is configured so that once strings is updated to use the new `file_format` parameter, that will be functional for dataframe saving as well.


Ethan-DeBandi99:
Tests that are currently broken will be resolved with `Strings.save()` update to support new configuration. Keeping this PR in draft until PR #1307 is merged.

arkouda/dataframe.py
```diff
@@ -12,7 +12,8 @@
 from arkouda.pdarraycreation import arange, array
 from arkouda.groupbyclass import GroupBy as akGroupBy
 from arkouda.pdarraysetops import concatenate, unique, intersect1d, in1d
-from arkouda.pdarrayIO import save_all
+from arkouda.pdarrayIO import save_all, load_all
+from arkouda.pdarrayIO import save_all, load_all
```
joshmarshall1, line 16: Duplicate import

joshmarshall1 (approved):
All looks good except a duplicated `from arkouda.pdarrayIO import save_all, load_all` in `dataframe.py`

arkouda/dataframe.py
```diff
@@ -1068,6 +1069,41 @@ def save(self, path, index=False):
...
+        columns: List
+            List of columns to include in the file. If None, writes out all columns
+        file_format: str
+            'HDF5' or 'Parquet'. Defaults to 'HDF5'
+        index: Bool
+            If true, include the index values in the save file.
+
+        Notes
+        ______
+        This function current uses 'truncate' mode to ensure the file exists before appending.
```
pierce314159, line 1089: `current` > `currently`

arkouda/dataframe.py
```diff
@@ -1068,6 +1069,41 @@ def save(self, path, index=False):
...
+        if index:
+            data["Index"] = self.index
+        save_all(data, prefix_path=prefix_path,
+                 file_format=file_format)
+    @classmethod
+    def load_table(cls, prefix_path):
+        return cls(load_all(prefix_path))
+
+
+
```
pierce314159, line 1106: this is just white space so but we have 0 newlines between `save_table` & `load_table` but 3 between `load_table` & `argsort`

tests/io_test.py
```diff
@@ -529,6 +529,7 @@ def testAppendMixedStringsDataset(self):
         
         strings = strings_array.to_ndarray()
         strings.sort()
+        print(r_mixed)
```
pierce314159, line 532: stray print statement

pierce314159 (approved):
Everything looks good! Couple small formatting things

tests/io_test.py
Ethan-DeBandi99: Removed

arkouda/dataframe.py
Ethan-DeBandi99: Updated. Good catch.

arkouda/dataframe.py
Ethan-DeBandi99: Removed duplicate import

arkouda/dataframe.py
Ethan-DeBandi99: Updated.

PR #1313, opened 2022-04-21
Categorical updates

reuster986:
This PR brings Categoricals a little more in line with user intuition:
- adds an N/A value for representing missing or null data
- keeps categories constant for all operations that return a Categorical, except when the user explicitly (re)sets the categories. This behavior is useful in situations where the user hand-picked the categories or expects them to be defined the same across multiple arrays.
- adds functions for setting explicit categories and standardizing categories across multiple Categoricals (these operations are out-of-place, not in-place)
- adds support for calling `ak.lookup()` with Categorical values
- adds a few new tests for this functionality

### DIFF
--- a/arkouda/__init__.py
+++ b/arkouda/__init__.py
@@ -24,3 +24,4 @@ from arkouda.row import *
 from arkouda.index import *
 from arkouda.series import *
 from arkouda.alignment import *
+from arkouda.plotting import *
--- a/arkouda/__init__.py
+++ b/arkouda/__init__.py
@@ -23,4 +23,4 @@ from arkouda.dataframe import *
 from arkouda.row import *
 from arkouda.index import *
 from arkouda.series import *
-
+from arkouda.alignment import *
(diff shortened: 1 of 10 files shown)


reuster986:
@pierce314159 and @Ethan-DeBandi99 this should be ready to review now. I had a subtle logic error where I made an assumption about the ordering of an array produced by `concatenate(..., ordered=False)` that was false in the multi-locale case, but the CI caught it and I refactored the code so that it no longer relies on that assumed ordering.

reuster986:
@Ethan-DeBandi99 I found myself manually appending an NAvalue to Categoricals in several mission workflows (mostly when I have to do a ak.lookup() with Categorical values and need something to put for values not in the table), so I figured I would just put it in by default.

Ethan-DeBandi99:
> @Ethan-DeBandi99 I found myself manually appending an NAvalue to Categoricals in several mission workflows (mostly when I have to do a ak.lookup() with Categorical values and need something to put for values not in the table), so I figured I would just put it in by default.

Makes sense to me. Thanks for clarifying!

Issue #1322, opened 2022-04-22
Compile C++ code with sanitizers if the server is being built with sanitizers


bmcdonald3:
Today, if Chapel is built with `CHPL_SANITIZE_EXE` set, then the server will be built with that sanitizer. The C++ code should get the same sanitizer flags as the Chapel code so that we can catch memory errors in the C++ code as well.

PR #1323, opened 2022-04-22
Add sanitizer check for C++ compilation in Makefile

bmcdonald3:
This PR adds a check to see if the Chapel executable was built
with sanitizers and, if it is, it will also compile the Arrow C++
code with those sanitizers. This change will help with nightly
sanitizer testing.

Closes https://github.com/Bears-R-Us/arkouda/issues/1322

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,9 +149,14 @@ CHECK_DEPS = check-chpl check-zmq check-hdf5 check-re2 $(OPTIONAL_CHECKS)
 endif
 check-deps: $(CHECK_DEPS)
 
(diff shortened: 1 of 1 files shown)


ronawho (approved):
What's here looks good. Are you planning a followup to build the dependencies with sanitizers enabled?

bmcdonald3:
@ronawho Were you imagining that all the dependencies would be built with sanitizers enabled? If so, I would put that as a new follow up PR. If only Arrow, I would include that in this PR I think.

On the other hand, it could be good to get some runs working with this simpler form before making more code changes?

Ethan-DeBandi99 (approved):

Issue #1328, opened 2022-04-25
File I/O Inconsistencies


Ethan-DeBandi99:
When working issue #1324, I noticed some inconsistencies with File I/O between Parquet and HDF5. I have listed them out below. I spoke with @mhmerrill and we believe that both should function as HDF5 does

- HDF5 is able to be saved with or without a file extension using the `save()` or `save_all()` methods. Parquet files always have `.parquet` appended when writing the files in `ParquetMsg.chpl`. 
- Loading HDF5 is configured so that loading is done by passing a `prefix_path` and then using `os.path.splitext()` to get the prefix and then extension and use those to load the file. This would allow `load()` to work with Parquet when doing something like `load('path_to_dir/name_prefix.parquet')`. However, when saving, the user does not have to provide the `.parquet `so this is not exactly clear when loading.
- Adding the `file_format` variable for loading requires HDF5 to retain the same functionality that was there before and then if it is set to `file_format='Parquet'` the extension must be set to `.parquet`. I have a solution doing this that I will be pushing a PR for shortly. I will link that once it is up.

I may not be aware of some reasoning behind configuring things to work this way. Is there a reason that we have this variance or can we go ahead an update the Parquet support to be more in line with the flexibility of HDF5?

bmcdonald3:
No, the adding of the `.parquet` was totally arbitrary. At one time in history, I believe the HDF5 files were saved with a `.hdf` extension (there were, and probably still are, some documentation showing that is how they are saved), but of course they aren't today.

I would be fine with removing the `.parquet` for writing parquet files. Mostly I just liked it because I could distinguish easier between an HDF5 and Parquet file that was written with Arkouda, but I'm not sure if anyone else has an opinion on this.

reuster986:
@Ethan-DeBandi99 I agree with the proposed updates to unify HDF5 and Parquet behavior.

Ethan-DeBandi99:
10-4 I will get on that. Thanks!

Ethan-DeBandi99:
Issue #1331 has been added to align the behavior of parquet with HDF5.

PR #1329, opened 2022-04-25
Closes #1324 - `load_all()` use `file_format`

### LINKED ISSUE #1324: `load_all()` use `file_format`
Most of the save code has been updated to use `file_format` to indicate if saving should be done via Parquet or HDF5 (defaulting to HDF5). Working on #1303, I noticed that `pdarrayIO:load_all()` is not using this. But instead uses `os.path.splitext(path_prefix)` to access the filetype and format the filepath for configuring the dataset. The issue here is that a user could save to a directory with no name appended to the front of `_LOCAL####`. This results in `os.path.splitext` not properly identifying the files to load. 

In order to not break any existing code that is formatted to work properly with `load_all`, we should still check if an extension is found using the existing method. If it is, we will utilize that, otherwise, we can use the optional `file_format` parameter to identify the file type and set the extension.

Ethan-DeBandi99:
Closes #1324 

- Updates `ak.load()` and `ak.load_all()` to make use of the `file_format` parameter that has been added for save methhods.
- Adds testing for `ak.load()` and `ak.load_all()` functionality with `file_format` parameter to `tests/io_test.py`.

The code here is configured to work around some of the inconsistencies raised in #1328. Because Parquet files always have `.parquet` extension, but HDF5 has more flexibility the code is configured to add the extension for parquet and continue using `os.path.splitext()` for HDF5 files.

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -122,6 +122,7 @@ def read(filenames : Union[str, List[str]],
     and read all of them. Use ``get_datasets`` to show the names of datasets
     to HDF5 files.
     """
+    print(filenames)
     if isinstance(filenames, str):
         filenames = [filenames]
     if datasets is None:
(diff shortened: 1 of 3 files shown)


arkouda/pdarrayIO.py
```diff
@@ -122,6 +122,7 @@ def read(filenames : Union[str, List[str]],
     and read all of them. Use ``get_datasets`` to show the names of datasets
     to HDF5 files.
     """
+    print(filenames)
```
reuster986, line 125: I'm not sure we want to print out the filenames. Sometimes there can be thousands of them.

reuster986 (changes requested):
Looks good overall! Just one request to remove a print statement.

arkouda/pdarrayIO.py
Ethan-DeBandi99: Agreed, this was left over from some debugging I was doing. I just pushed up the removal of the print.

reuster986 (approved):

PR #1330, opened 2022-04-26
846 refactor unique to handle groupby

### LINKED ISSUE #846: Refactor GroupBy, findSegments, and Unique
I have two goals for this issue, and wanted to start a discussion about them:
* Separate grouping and sorting semantics
  * `argsort` and `coargsort` should actually sort the array(s). Currently, calling `coargsort` on a list including a Strings or Categorical will only group, not sort.
  * `GroupBy` should guarantee grouping, but not necessarily sorting. Strings and Categorical should have separate APIs for sorting and grouping, and `GroupBy` should call the latter.

* Consolidate GroupBy/findSegments logic and migrate to Chapel
  * The `uniqueMsg` function in Chapel actually does 95% of what `GroupBy` needs. I propose refactoring `uniqueMsg` and its sub-fuctions to
    * handle all groupable types: `int64` pdarray, Strings, and Categorical, as well as lists of these
    * optionally return a permutation, segments, and unique key indices, in addition to the unique values. These are already computed internally (or are trivially derivable from what is), and comprise all the information necessary to construct a `GroupBy`. 
  * Doing so will reduce code (by rendering `findSegmentsMsg` unnecessary) and improve performance in some cases (e.g. when arrays are packed into chapel tuples for coargsort).
  * This work would be greatly simplified by having a `MultiArray` class in the server -- a `GenSymEntry` that holds multiple equal-length columns, like a dataframe but without all the methods.

reuster986:
Closes #846 . This PR significantly refactors `GroupBy` on both the client and server side to make use of an enhanced `unique` function. From a user perspective, `GroupBy` usage and semantics is unchanged, so backwards compatibility is preserved. This refactor has three main benefits:
- Users can now call `unique` on lists/tuples of arrays to get the unique "rows". This is a common pattern that previously had to be accomplished via hacks like `GroupBy(list_of_arrays).unique_keys`
- The `findSegmentsMsg` server function has been eliminated, with the functionality rolled into `uniqueMsg`. This not only consolidates code, but should improve performance by avoiding recompuatation of some internal arrays used in finding segments. I would expect the performance of all GroupBy benchmark cases to increase with this change, but especially on 16-node CS.
- The logic for `GroupBy` involving lists of arrays with more than 16 combined radix digits (128 bits) has been revamped to rely on hashing, which should greatly improve scaling and performance in the limit of large numbers of arrays. This change should further improve the 8- and 16-array GroupBy benchmark cases, bringing them closer to the 1- and 2-array cases (again, especially on 16-node CS).
- These changes support the goal of eventually decoupling grouping semantics from (co)sorting semantics, see #846


reuster986:
@Ethan-DeBandi99 I just noticed you self-assigned the corresponding issue #846 last week -- apologies for not communicating about this sooner.

reuster986:
@bmcdonald3 or @ronawho would either of you be able to do an out-of-cycle benchmark run for this PR, just to check that the expected performance improvements materialize? I know the Chapel review is happening now (I'm listening as we speak!), so if you don't have spare time, we could just merge and revert if we see a problem in the nightly run.

bmcdonald3:
@reuster986 I can run this after the review and get back to you later today since I am on PT if that timeline is alright.

bmcdonald3:
Hey @reuster986, had a chance to run this branch with the `groupby` benchmark on 16-node CS and this is what I saw:

  | Current master | This branch
-- | -- | --
1 array | 7.8533 GiB/s | 7.6119 GiB/s
2 array | 6.6566 GiB/s | 7.3021 GiB/s
8 array | 2.7024 GiB/s | 5.4505 GiB/s
16 array | 1.3854 GiB/s | 4.9655 GiB/s

So pretty significant performance improvements as the number of arrays increase.

reuster986:
Excellent, thanks @bmcdonald3 ! On the one hand, I'm shocked that my code actually had the intended effect! On the other hand, I expected a small speedup for the 1-array case because rolling findSegments into unique eliminates a redundant gather. But now that I think about it, the old code handled single arrays in a streamlined way that avoided packing the array into tuples, whereas the new code does the tuple packing on everything. I could probably special-case single arrays in the new code, but I think I should save that for a separate PR.

src/UniqueMsg.chpl
```diff
@@ -26,137 +27,220 @@ module UniqueMsg
...
-                    }
-                  }
-                  otherwise {
-                      var errorMsg = notImplementedError("unique",gEnt.dtype);
-                      umLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);                
-                      return new MsgTuple(errorMsg, MsgType.ERROR);
-                  }
-                }
+    proc uniqueMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+      var (returnGroupsStr, nstr, rest) = payload.splitMsgToTuple(3);
```
Ethan-DeBandi99, line 37: Indentation is off from the rest of the function here.

Ethan-DeBandi99 (approved):
Looks good to me. I saw one whitespace thing, but that should not cause any issues.

joshmarshall1 (approved):
Looks good

src/AryUtil.chpl
```diff
@@ -5,6 +5,14 @@ module AryUtil
     use Reflection;
     use Logging;
     use ServerConfig;
+    use MultiTypeSymbolTable;
+    use MultiTypeSymEntry;
+    use ServerErrors;
+    use BitOps;
+
+    param bitsPerDigit = RSLSD_bitsPerDigit;
```
mhmerrill, line 13: @reuster986 and I talked about reworking the `RadixSortLSD` code layout between modules. I would like to see it refactored into just two modules maybe call them `RadixSortLSD_core` and `RadixSortLSD_support`. This would merge the stuff from this file and `ServerConfig` into `RadixSort_support`, so all the helper functions and parameters. And we would switch to using ``private config param` for commandline parameters.

mhmerrill (changes requested):
needs a little refactor between modules in the chapel server code.

pierce314159 (commented):
Couple small things, my biggest question is should the `storeUniqueKeys` call be commented out? 

Also I'm not sure this should close Issue #846, because it doesn't address the `Separate grouping and sorting semantics` bullet 

I think either we should make that a parent issue and make sub-issues off it or just create another issue to track the string/categorical sorting vs grouping

src/AryUtil.chpl
mhmerrill: i'll add this to issue on refactor of sort code

PR #1332, opened 2022-04-27
#1199 - Dynamic versioning for quickstart documentation

### LINKED ISSUE #1199: Set Documentation Version with Versioneer
Previously, Arkouda documentation was being generated with a hardcoded version number. This was updated in issue #1176. The fix at that point was to correct the `./arkouda_server` output to show the current format and use a format string to depict the version and to update `release='2020.07.07' ` to `release=''`The version that is actually printed is generated via `versioneer`.  After discussing with @mhmerrill, we want to look into having versioneer supply a version for the documentation as well.

- Set `release='<current_version>'` in `pydoc/conf.py`
- We will likely need a `pydoc/_version.py` like in `arkouda`.

joshmarshall1:
This ticket (closes #1199):

Versioneer doesn't support writing the version to multiple locations, so the workaround for this is to import the `_version.py` file under arkouda into any files that need to access the version. Using the `get_versions()` method from `_version.py` I was able to pull the current version into the documentation's config file.

Sphinx doesn't natively support replacing strings within `code-blocks`, so because the versions in the quickstart guide are within a code-block I found an extension that enables substitution inside `code-blocks` [here](https://stackoverflow.com/questions/44761197/how-to-use-substitution-definitions-with-code-blocks/53267394#53267394?newreg=7a731bc3a74141e9a74d7cd46fcf74fa)

I added the extension under `pydocs/_ext` and setup the substitution variable to make it work. Tested with `make docs` and verified the substitution happened in both instances of the version being displayed as well as in the page title by looking at the output file and hovering over the tab in my browser.

Example output:

``` bash
$ ./arkouda_server -nl 1
server listening on tcp://<your_machine>.local:5555
arkouda server version = v2022.04.15+22.g8027d1f6.dirty
memory limit = 15461882265
bytes of memory used = 0
```

### DIFF
--- a/pydoc/_ext/ak_sphinx_extensions.py
+++ b/pydoc/_ext/ak_sphinx_extensions.py
@@ -1,0 +1,40 @@
+"""
+Custom Sphinx extensions.
+"""
+
+from typing import List
+
+from sphinx.application import Sphinx
+from sphinx.directives.code import CodeBlock
+
+class SubstitutionCodeBlock(CodeBlock):  # type: ignore
+    """
+    Similar to CodeBlock but replaces placeholders with variables.
+    """
+
+    def run(self) -> List:
+        """
+        Replace placeholders with given variables.
+        """
+        app = self.state.document.settings.env.app
+        new_content = []
+        self.content = self.content  # type: List[str]
+        existing_content = self.content
+        for item in existing_content:
+            for pair in app.config.substitutions:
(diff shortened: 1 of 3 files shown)


Ethan-DeBandi99 (approved):
Looks good to me. Thanks for figuring this out!

pierce314159 (approved):

PR #1334, opened 2022-04-27
Closes #1300 - Improve Performance of DataFrame Display

### LINKED ISSUE #1300: Improve performance of dataframe display
Interactive users working in Jupyter notebooks frequently print out the first and last few rows of a dataframe to see sample values or check the results of a transformation. The syntax for this is just to execute `df` as a statement, which shows an html table of the head and tail of the dataframe, just like pandas does. With pandas, this operation is instantaneous. I believe it is also fast with single-node arkouda, but when running with many nodes, it can take 5-10 seconds to show this "quick" view of the dataframe. As a developer, this feels like an annoyingly trivial place to worry about performance, but for the user, viewing a dataframe is probably the operation with the widest gap between expected and actual runtime, and it is so common that it tends to bog down interactive workflows. As a goal, I'd love to get to the point where a 20-column dataframe can display in under 1 second.

Probably the first thing to do is create a benchmark that builds a 10_000 x 20 dataframe of mixed dtypes (including Categoricals, which are especially slow) and simulates displaying it by calling `df._repr_html_()`. That call is the one to measure for the benchmark. I'm hoping that a 16 node CS run of this benchmark would replicate the several seconds of latency we're experiencing on our typical node allocations.

In theory, we should be able to make this operation go quite fast. The amount of data transferred to the client and displayed by this operation is bounded by the dimensions of the dataframe and by the pandas options `'display.max_rows'` and `'display.max_columns'`, which default to something under 100 and in practice never get set above a few thousand. For our test dataframe with 20 columns and a `display.max_rows=100`, that means 2k elements and ~16KB (depending on strings) transferred, so a very small amount of data. The problem today is that the client-side code for `df._repr_html_()` loops over columns and slices the head and tail of each column as a separate indexing message, which incurs both message overhead and comm aggregation startup for each column. I think we could address this by making a dedicated server message that processes all columns within a single message and bypasses aggregation for this small amount of data.

Ethan-DeBandi99:
Closes #1300 

- Updates the `_get_head_tail()` method to function by creating a single server message instead of 1 per column in the DataFrame. This was done by adding `_get_head_tail_server()` which allows us to benchmark the new code against the old. The old function has been retained for benchmarking.
- Adds a `DataFrameIndexingMsg.chpl` to parse and process the server message for indexing the columns. The message is configured so the the column type, column name (in the df), column object name (server-side) are sent to the server to indexing to the appropriate head/tail. 
- Benchmarking configured to test `_rep_html()` which now calls `_get_head_tail_server()`. The benchmark also checks `_get_head_tail_server()` and `_get_head_tail()` directly to allow for easy comparison. 

Per the issue description, @joshmarshall1 and I did this without using aggregation because the datasets should be relatively small.

Benchmarking on a single node shows an performance already, but we will need to benchmark this on a multi-node system. Results from single node:

```bash
array size = 10,000
number of trials =  5
>>> arkouda dataframe display
numLocales = 1, N = 10,000
  _repr_html_ Average time = 0.0401 sec
  _repr_html_ Average rate = 0.03 GiB/sec
  _get_head_tail_server Average time = 0.0406 sec
  _get_head_tail_server Average rate = 0.03 GiB/sec
  _get_head_tail Average time = 0.0825 sec
  _get_head_tail Average rate = 0.01 GiB/sec

```

@reuster986 - I will leave it up to you if you would like to request an out of band benchmark.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -5,12 +5,13 @@ from warnings import warn
 import pandas as pd  # type: ignore
 import random
 import json
+from typing import cast
 
 from arkouda.segarray import SegArray
 from arkouda.pdarrayclass import pdarray
 from arkouda.categorical import Categorical
 from arkouda.strings import Strings
-from arkouda.pdarraycreation import arange, array
+from arkouda.pdarraycreation import arange, array, create_pdarray
 from arkouda.groupbyclass import GroupBy as akGroupBy
 from arkouda.pdarraysetops import concatenate, unique, intersect1d, in1d
 from arkouda.pdarrayIO import save_all, load_all
@@ -440,23 +441,33 @@ class DataFrame(UserDict):
         # Being 1 above the threshold causes the PANDAS formatter to split the data frame vertically
         idx = array(list(range(maxrows // 2 + 1)) + list(range(self._size - (maxrows // 2), self._size)))
         # TODO - pass names to server of
-        #   - idx
-        #   - if categorical -> send codes name
-        #   - otherwise send name
-        #   - number of elements
-        #   - each element type [(type, name), ...]
+        #   - Link column to index
+        #   - SegArray implementation. Locations outlined
         msg_list = []
         for col in self._columns:
-            # TODO - handle segarray
             if isinstance(self[col], Categorical):
                 msg_list.append(f"Categorical+{self[col].codes.name}+{self[col].categories.name}")
+            elif isinstance(self[col], SegArray):
+                msg_list.append(f"SegArray+{self[col].codes.name}+{self[col].categories.name}")
             elif isinstance(self[col], Strings):
                 msg_list.append(f"Strings+{self[col].name}")
             else:
                 msg_list.append(f"pdarray+{self[col].name}")
 
-        generic_msg(cmd="dataframe_idx", args="{} {} {}".
-                    format(len(msg_list), idx.name, json.dumps(msg_list)))
+        repMsg = cast(str, generic_msg(cmd="dataframe_idx", args="{} {} {}".
+                                       format(len(msg_list), idx.name, json.dumps(msg_list))))
+        msgList = json.loads(repMsg)
+        for m in msgList:
+            msg = m.split("+", 1)
+            t = msg[0]
+            if t == "Strings":
+                Strings.from_return_msg(msg[1])
+            elif t == "SegArray":
+                return NotImplemented
+            else:
+                create_pdarray(msg[1])
+
+
 
     def _shape_str(self):
         return "{} rows x {} columns".format(self.size, self._ncols())
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -446,28 +446,34 @@ class DataFrame(UserDict):
         msg_list = []
         for col in self._columns:
             if isinstance(self[col], Categorical):
-                msg_list.append(f"Categorical+{self[col].codes.name}+{self[col].categories.name}")
+                msg_list.append(f"Categorical+{col}+{self[col].codes.name}+{self[col].categories.name}")
             elif isinstance(self[col], SegArray):
-                msg_list.append(f"SegArray+{self[col].codes.name}+{self[col].categories.name}")
+                msg_list.append(f"SegArray+{col}+{self[col].segments.name}+{self[col].values.name}")
             elif isinstance(self[col], Strings):
-                msg_list.append(f"Strings+{self[col].name}")
+                msg_list.append(f"Strings+{col}+{self[col].name}")
             else:
-                msg_list.append(f"pdarray+{self[col].name}")
+                msg_list.append(f"pdarray+{col}+{self[col].name}")
 
         repMsg = cast(str, generic_msg(cmd="dataframe_idx", args="{} {} {}".
                                        format(len(msg_list), idx.name, json.dumps(msg_list))))
         msgList = json.loads(repMsg)
+
+        df_dict = {}
         for m in msgList:
             msg = m.split("+", 1)
             t = msg[0]
             if t == "Strings":
-                Strings.from_return_msg(msg[1])
+                df_dict[msg[1]] = Strings.from_return_msg(msg[2])
             elif t == "SegArray":
-                return NotImplemented
+                eles = msg[2].split("+")
+                df_dict[msg[1]] = SegArray(create_pdarray(eles[0]), create_pdarray(eles[1]))
             else:
-                create_pdarray(msg[1])
+                df_dict[msg[1]] = create_pdarray(msg[2])
 
+        new_df = DataFrame(df_dict)
+        new_df._set_index(idx)
 
+        return new_df.to_pandas(retain_index=True)[self._columns]
 
     def _shape_str(self):
         return "{} rows x {} columns".format(self.size, self._ncols())
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -461,6 +461,7 @@ class DataFrame(UserDict):
             msg = m.split("+", 2)
             t = msg[0]
             if t == "Strings":
+                # Categorical is returned as a strings by indexing categories[codes[idx]]
                 df_dict[msg[1]] = Strings.from_return_msg(msg[2])
             elif t == "SegArray":
                 # split creates for segments and values
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -4,6 +4,7 @@ from collections import UserDict
 from warnings import warn
 import pandas as pd  # type: ignore
 import random
+import json
 
 from arkouda.segarray import SegArray
 from arkouda.pdarrayclass import pdarray
@@ -17,7 +18,7 @@ from arkouda.dtypes import int64 as akint64
 from arkouda.dtypes import float64 as akfloat64
 from arkouda.sorting import argsort, coargsort
 from arkouda.numeric import where
-from arkouda.client import maxTransferBytes
+from arkouda.client import maxTransferBytes, generic_msg
 from arkouda.row import Row
 from arkouda.alignment import in1dmulti
 from arkouda.series import Series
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99:
Added @joshmarshall1 as a reviewer even though he helped write this code. There are some elements that I handled that would be good for him to review as well.

joshmarshall1 (commented):
As a participant in developing this, I won't approve it, but after a second review I think we should go through and add comments throughout the chpl code, since there are almost none, just to make future maintenance easier. I also noticed a few places that TODO tags were left in and a stray #

benchmarks/dataframe.py
```diff
@@ -0,0 +1,112 @@
...
+TYPES = ('int64', 'uint64',)
+
+def generate_dataframe(N):
+    types = [ak.Categorical, ak.pdarray, ak.Strings, ak.SegArray]
+
+    # generate random columns to build dataframe
+    df_dict = {}
+    for x in range(20):  # loop to create 20 random columns
+        key = f"c_{x}"
+        d = types[random.randint(0, len(types)-1)]
```
reuster986, line 20: I'm thinking we should cycle deterministically through the types, rather than choosing them randomly. If today's run gets 10 Categorical columns and tomorrow's run gets 2, there will be a big difference in runtime that doesn't reflect anything about the code being benchmarked. Generating random *data* of a particular type is fine, though.

benchmarks/dataframe.py
```diff
@@ -0,0 +1,112 @@
...
+        if d == ak.Categorical:
+            str_arr = ak.array(["".join(random.choices(string.ascii_letters + string.digits, k=5)) for _ in range(N)])
```
reuster986, line 22: This is fine for this benchmark's default problem size of 10**4, but just for future reference, the scalable way to generate a N-long ak.Strings array of 5-character strings is `ak.random_strings_uniform(minlen=5, maxlen=6, size=N)`.

benchmarks/dataframe.py
```diff
@@ -0,0 +1,112 @@
...
+            df_dict[key] = ak.Categorical(str_arr)
+        elif d == ak.pdarray:
+            df_dict[key] = ak.array(np.random.randint(0, 2 ** 32, N))
+        elif d == ak.Strings:
+            df_dict[key] = ak.array(["".join(random.choices(string.ascii_letters + string.digits, k=5)) for _ in range(N)])
```
reuster986, line 27: `ak.random_strings_uniform()` would be useful here too.

arkouda/dataframe.py
```diff
@@ -440,8 +491,8 @@ def _repr_html_(self):
         """
         Return html-formatted version of the dataframe.
         """
+        prt = self._get_head_tail_server()
 
-        prt = self._get_head_tail()
```
reuster986, line 444: I really love that you benchmarked both the old and new versions of this function! Although the single-node results look really promising, I'm wondering if we should look at how those benchmarks compare on the 16-node cases before we switch the implementation. We could merge this PR while still using `self._get_head_tail()`, and then if the benchmarks look good, it would be a one-line PR to change the implementation. On the other hand, it would save us some trouble if there's a hidden gotcha in the multi-locale case.

reuster986 (changes requested):
Looks good overall! Just a couple requested changes in the benchmark, and a suggestion about when/how we switch to the new implementation.

arkouda/dataframe.py
```diff
@@ -440,8 +491,9 @@ def _repr_html_(self):
         """
         Return html-formatted version of the dataframe.
         """
+        #
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Removed

src/DataFrameIndexingMsg.chpl
```diff
@@ -0,0 +1,206 @@
...
+				}
+				otherwise {
+					var errorMsg = notImplementedError(pn, ele_parts[0]);
+					dfiLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+					throw new IllegalArgumentError(errorMsg);
+				}
+			}
+		}
+
+		// TODO - return values need to provide what type the data is so that we can properly configure the object on the client
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Removed

src/DataFrameIndexingMsg.chpl
```diff
@@ -0,0 +1,206 @@
...
+							rpm = "%jt".format(dfIdxHelper(idx, col_vals, st, col_name));
+						}
+						otherwise {
+							var errorMsg = notImplementedError(pn,dtype2str(gCol.dtype));
+							dfiLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+							throw new IllegalArgumentError(errorMsg);
+						}
+					}
+				}
+				// TODO - when SegArray
```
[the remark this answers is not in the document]
Ethan-DeBandi99: removed

benchmarks/dataframe.py
Ethan-DeBandi99: I updated this to use `x%4` to evenly distribute the column types.

benchmarks/dataframe.py
Ethan-DeBandi99: I did not know that we had this. I updated to use `ak.random_strings_uniform`.

benchmarks/dataframe.py
Ethan-DeBandi99: Updated to use `ak.random_strings_uniform`

arkouda/dataframe.py
Ethan-DeBandi99: This makes sense to me. I update `_repr_html_()` and `__repr__()` to use `self._get_head_tail()`. The benchmark has been updated to remove `_repr_html_()` from the operations to test and will now only directly compare `_get_head_tail()` and `get_head_tail_server()`

Ethan-DeBandi99:
Updated benchmark results on single node with requested updates from @reuster986 implemented.

```bash
array size = 10,000
number of trials =  5
>>> arkouda dataframe display
numLocales = 1, N = 10,000
  _get_head_tail_server Average time = 0.0261 sec
  _get_head_tail_server Average rate = 0.05 GiB/sec
  _get_head_tail Average time = 0.0547 sec
  _get_head_tail Average rate = 0.02 GiB/sec
```

reuster986 (approved):
Looks great, thanks!

benchmarks/dataframe.py
```diff
@@ -0,0 +1,111 @@
...
+    pd.set_option("display.max_columns", 20)
+
+    printdf = df._get_head_tail_server() # measure the pandas df returned
+    # Mainly want to verify shape for the print
+    assert(printdf.shape[0] == 101)
+    assert(printdf.shape[1] == 20)
+
+
+def create_parser():
+    parser = argparse.ArgumentParser(description="Run the setops benchmarks: intersect1d, union1d, setdiff1d, setxor1d")
```
pierce314159, line 79: The description needs updating

benchmarks/dataframe.py
```diff
@@ -0,0 +1,111 @@
...
+    parser.add_argument('hostname', help='Hostname of arkouda server')
+    parser.add_argument('port', type=int, help='Port of arkouda server')
+    parser.add_argument('-n', '--size', type=int, default=10**4, help='Problem size: length of arrays A and B')
```
pierce314159, line 82: I'm not sure but `--size` might need updating as well

src/DataFrameIndexingMsg.chpl
```diff
@@ -0,0 +1,204 @@
...
+    use ServerErrors;
+    use Logging;
+    use Message;
+    use SegmentedMsg;
+    use AryUtil;
+
+    use MultiTypeSymEntry;
+    use MultiTypeSymbolTable;
+
+    use CommAggregation;
```
pierce314159, line 16: do we need this import if we're not using aggregation?

src/DataFrameIndexingMsg.chpl
```diff
@@ -0,0 +1,204 @@
...
+                when ("SegArray"){
+                    dfiLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),"Element at %i is SegArray".format(i));
+                    ref segments_name = ele_parts[2];
+                    ref values_name = ele_parts[3];
+                    dfiLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),"Segments Name: %s, Values Name: %s".format(segments_name, values_name));
+
+                    var gSeg: borrowed GenSymEntry = getGenericTypedArrayEntry(segments_name, st);
+                    var segments = toSymEntry(gSeg, int);
+                    var gVal: borrowed GenSymEntry = getGenericTypedArrayEntry(values_name, st);
+                    var values = toSymEntry(gVal, int);
```
pierce314159, line 182: does segarray also support `uint` as possible type for values?

pierce314159 (commented):
Couple comments, nothing major. The only thing that might need to be updated is "support uint for segarray values" question? But the logic all looks good to me

benchmarks/dataframe.py
```diff
@@ -0,0 +1,111 @@
...
+            r = fxn()
+            end = time.time()
+            timings[op].append(end - start)
+            results[op] = r
+
+    tavg = {op: sum(t) / trials for op, t in timings.items()}
+
+    for op, t in tavg.items():
+        print("  {} Average time = {:.4f} sec".format(op, t))
+        bytes_per_sec = (df.size * 64 * 2) / t
```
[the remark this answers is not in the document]
Ethan-DeBandi99: To be completely honest, I copied this out of another benchmark and did not update it. I will take a look at argsort and update accordingly.

src/DataFrameIndexingMsg.chpl
Ethan-DeBandi99: updated to properly handle supported types.

arkouda/dataframe.py
```diff
@@ -422,6 +424,55 @@ def _get_head_tail(self):
...
+                                       format(len(msg_list), idx.name, json.dumps(msg_list))))
+        msgList = json.loads(repMsg)
+
+        df_dict = {}
+        for m in msgList:
+            # Split to [datatype, column, create]
+            msg = m.split("+", 2)
+            t = msg[0]
+            if t == "Strings":
+                df_dict[msg[1]] = Strings.from_return_msg(msg[2])
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Added comment.

benchmarks/dataframe.py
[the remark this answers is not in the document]
Ethan-DeBandi99: We talked offline about this, but updated this to address nbytes properly.

benchmarks/dataframe.py
Ethan-DeBandi99: Updated

benchmarks/dataframe.py
Ethan-DeBandi99: Updated the description.

src/DataFrameIndexingMsg.chpl
Ethan-DeBandi99: Good point. This has been removed.

src/DataFrameIndexingMsg.chpl
```diff
@@ -0,0 +1,204 @@
...
+
+    // gather indexing by integer index vector
+    proc dfIdxHelper(idx: borrowed SymEntry(int), columnVals: borrowed SymEntry(?t), st: borrowed SymTab, col: string, rtnName: bool=false): string throws {
+        param pn = Reflection.getRoutineName();
+        // get next symbol name
+        var rname = st.nextName();
+
+        if (columnVals.size == 0) && (idx.size == 0) {
+            var a = st.addEntry(rname, 0, t);
+            var repMsg = "pdarray+%s+created ".format(col) + st.attrib(rname);
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Updated.

pierce314159 (approved):
Looks good!

Ethan-DeBandi99:
@reuster986 not going to lie, I am a bit confused as to why a list can't index into rows. Can you give me some details on your thought process here? I will get it pulled out though

PR #1341, opened 2022-04-29
Closes # 1310 - Import akutil updates

### LINKED ISSUE #1310: Import akutil updates
For context, akutil is a package of client-side enhancements for arkouda popular among users who can't contribute directly to arkouda. Most of the akutil functionality has been ported into arkouda proper (thanks everyone!), but there have been a few updates to akutil since then.

I have brought in three files recently updated in akutil and placed them in a new branch (https://github.com/Bears-R-Us/arkouda/tree/dataframe-updates2), but I need some help with the merge logic. Comparing against master, many of the "changes" in these files would effectively revert some of the work that has been done to port into arkouda, which of course we don't want to do. I'm trying to figure out how to separate these old, unwanted "changes" from the actual updates intended by the user who wrote them. It feels like git should have a way to do this, e.g. by resetting the branch to the time when akutil was first ported into arkouda, so that the porting commits get applied on top of these files. Can someone help me figure out how to do this?

Ethan-DeBandi99:
Closes #1310 

- Adds `accessor.py`
- Updates to `dataframe.py` manually configured against current master
- Updates to `series.py` manually configured against current master
- Added code to check if user passes a column named `index`, and if they do, the index `df.index` will be built from that and the index is not added to the columns.

Linking to the diff of the branch posted with the issue for reference: [Akutil Updates Branch](https://github.com/Bears-R-Us/arkouda/compare/dataframe-updates2)

A large majority of the differences displayed are due to whitespace and were omitted.

Please Note: if reviewing the diff, some changes do not completely line up due to the fact that `ak.DataFrame` has been updated to utilize and `Index` object and set the index using a property instead of as a column in the data.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
+from datetime import datetime
+from ipaddress import ip_address
+from tabulate import tabulate
 from collections import UserDict
 from warnings import warn
 import pandas as pd  # type: ignore
+import numpy as np  # type: ignore
 import random
 import json
 from typing import cast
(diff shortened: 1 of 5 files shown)


arkouda/dataframe.py
```diff
@@ -233,6 +316,17 @@ def __init__(self, initialdata=None, index=None):
 
             self.update_size()
 
+    def __getattr__(self, key):
+        # print("key =", key)
+        if key not in self.columns:
+            # return self.__getattribute__(key)
```
pierce314159, line 322: can probably remove the commented out lines here

joshmarshall1 (approved):
Looks good

PR #1342, opened 2022-04-29
Closes #1337 - Adding support to generic attach for categorical and segarray

### LINKED ISSUE #1337: Add SegArray and Categorical support to Generic Attach
Right now, the generic attach method only supports pdarray and Strings. After talking to @Ethan-DeBandi99 we came up with a way to move some of this functinality into the Chapel that will allow for Categorical and SegArray generic attach to also be supported. GroupBy should be able to be done in the same way, but will require some more processing to find it's components which is why it will most likely be added with a second issue.

joshmarshall1:
This ticket (closes #1337):

Added methods to `RegistrationMsg.chpl` for generic attach command which takes in an optional `type` or will attempt to infer the type if not passed. Passing in the optional type will skip directly to the portion of the method that compiles the response message. This should allow users to pass the type to increase performance.

I added a method to `MultiTypeSymbolTable.chpl` that checks existence of a given name and returns a bool result. I didn't see any other methods that returned bool for checking existence, only `checkTable` which returned void if the name exists, or error if it doesn't. This new method prevents me from having to use a `try/catch` block with `checkTable` or having to access the `tab` property of the `SymTab` class directly.

Also added were `from_return_message` methods to `categorical.py` and `segarray.py` so the return could be passed from the generic attach method in `util.py` and parsed correctly by the corresponding class.

The existing test cases for `pdarray` and `Strings` generic attach were expanded to include a given and an inferred type test. Two new tests were added, one for `Categorical` and one for `SegArray`, also each with a given and an inferred type.


arkouda/categorical.py
```diff
@@ -124,6 +124,58 @@ def from_codes(cls, codes: pdarray, categories: Strings,
...
+        Raises
+        ------
+        RuntimeError
+            Raised if a server-side error is thrown in the process of creating
+            the categorical instance
+        """
+        # parts[0] is "categorical". Used by the generic attach method to identify the
+        # response message as a Categorical
+
+        repParts = repMsg.split(";")
```
Ethan-DeBandi99, line 153: I am thinking that we should stick with the convention of `+` delimited creates in the response messages. This will require gluing the strings components together for the `Strings.repParts[1]`, but I think that is worth the consistency.

arkouda/segarray.py
```diff
@@ -232,6 +233,37 @@ def concat(cls, x, axis=0, ordered=True):
...
+
+        Raises
+        ------
+        RuntimeError
+            Raised if a server-side error is thrown in the process of creating
+            the categorical instance
+        """
+        # parts[0] is "segarray". Used by the generic attach method to identify the
+        # response message as a SegArray
+        parts = repMsg.split(";")
```
Ethan-DeBandi99, line 260: Again, this should be `+` delimited for consistency.

arkouda/util.py
```diff
@@ -282,13 +282,21 @@ def convert_if_categorical(values):
...
     """
-    repMsg = generic_msg(cmd="attach", args=name)
-    dtype = repMsg.split()[2]
-    if dtype == "str":
-        return Strings.from_return_msg(repMsg)
+    repMsg = cast(str, generic_msg(cmd="genericAttach", args=f"{dtype}+{name}"))
+
+    if repMsg.split(";")[0] == "categorical":
+        return Categorical.from_return_msg(repMsg)
+    elif repMsg.split(";")[0] == "segarray":
```
Ethan-DeBandi99, line 294: These will also need to be updated to split on `+`

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+        }
+
+        repMsg = "categorical;created %s".format(cats);
+        // Check if the categories is numeric or string, if string add byte size
+        if (isStringAttrib(cats)) {
+            var s = getSegString("%s.categories".format(name), st);
+            repMsg += "+created bytes.size %t".format(s.nBytes);
+        }
+
+        repMsg += ";created %s".format(codes);
```
Ethan-DeBandi99, line 144: Use `+` to delimit created statements.

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+
+        // Optional components of categorical
+        if st.contains("%s.permutation".format(name)) {
+            var perm = st.attrib("%s.permutation".format(name));
+            if (perm.startsWith("Error:")) { 
+                var errorMsg = perm;
+                regLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+                return new MsgTuple(errorMsg, MsgType.ERROR); 
+            }
+            repMsg += ";created %s".format(perm);
```
Ethan-DeBandi99, line 154: replace `;` with `+`

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+        }
+        if st.contains("%s.segments".format(name)) {
+            var segs = st.attrib("%s.segments".format(name));
+            if (segs.startsWith("Error:")) { 
+                var errorMsg = segs;
+                regLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+                return new MsgTuple(errorMsg, MsgType.ERROR); 
+            }
+            repMsg += ";created %s".format(segs);
```
Ethan-DeBandi99, line 163: replace ; with +

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+        }
+
+        var lens = st.attrib("%s_lengths".format(name));
+        if (lens.startsWith("Error:")) { 
+            var errorMsg = lens;
+            regLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+            return new MsgTuple(errorMsg, MsgType.ERROR); 
+        }
+
+        repMsg = "segarray;created %s;created %s;created %s".format(segs, vals, lens);
```
Ethan-DeBandi99, line 210: replace ; with +

Ethan-DeBandi99 (changes requested):
Logic all looks good. One thing that I would like to change would be using `+` for all delimiters instead of `;`. I see why you are using a separate delimiter here (Categorical.categories being a `Strings` object). That should be easy enough to glue together to create the `Strings` object properly using `Strings.from_response_msg()`.

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+            regLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                            "Attempting to find type of registered element '%s'".format(name));
+
+            if st.contains(name) {
+                // Easy case where full name given matches an entry, pdarray or Strings
+                dtype = "simple";
+            } else if st.contains("%s.categories".format(name)) {
+                dtype = "categorical";
+            } else if st.contains("%s_segments".format(name)) {
+                // Important to note that categorical has a .segments while segarray uses _segments
```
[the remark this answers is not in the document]
Ethan-DeBandi99: The _ can be ignored here. I misread the line as `._segments` not just `_segments`

arkouda/categorical.py
```diff
@@ -124,6 +124,58 @@ def from_codes(cls, codes: pdarray, categories: Strings,
...
+
+        if len(repParts) > 2:
+            name = repParts[3].split()[0]
+            if ".permutation" in name:
+                parts["permutation"] = create_pdarray(repParts[3])
+            else:
+                parts["segments"] = create_pdarray(repParts[3])
+
+        if len(repParts) == 4:
+            parts["segments"] = create_pdarray(repParts[4])
```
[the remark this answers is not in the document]
joshmarshall1: Updated to be an if/elif/else that raises a ValueError if an unknown field is present

src/RegistrationMsg.chpl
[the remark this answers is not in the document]
joshmarshall1: Updated to check for the existence of the required fields for `Categorical` and `SegArray`

arkouda/categorical.py
joshmarshall1: Updated to use `+` delimited string in the response message and glued the components back together for the `Strings` component in `Categorical`

arkouda/segarray.py
joshmarshall1: Updated to use `+` delimited string in the response message

arkouda/util.py
joshmarshall1: Updated to use `+` delimited string in the response message

src/RegistrationMsg.chpl
joshmarshall1: Updated to use `+` delimited string in the response message

src/RegistrationMsg.chpl
joshmarshall1: Updated to use `+` delimited string in the response message

src/RegistrationMsg.chpl
joshmarshall1: Updated to use `+` delimited string in the response message

src/RegistrationMsg.chpl
joshmarshall1: Updated to use `+` delimited string in the response message

Ethan-DeBandi99 (approved):
Looks good

mhmerrill:
@joshmarshall1 there is a basic conflict on this PR now. I could fix it but I rather you did so I don't screw it up with the web editor ;-)

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+
+    :returns: MsgTuple response message
+    */
+    proc attachSegArrayMsg(cmd: string, name: string, st: borrowed SymTab): MsgTuple throws {
+        regLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                            "%s: Collecting SegArray components for '%s'".format(cmd, name));
+
+        var repMsg: string;
+
+        var segs = st.attrib("%s_segments".format(name));
```
pierce314159, line 189: i hate having to use `_` as our delimiter for segarray rather than `.`

I get that's to avoid conflicts with registering a categorical with the same name, but I want to be able to use a consistent delimiter for components

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+        var repMsg: string; // response message
+
+        var ele_parts = payload.split("+");
+        var dtype = ele_parts[0];
+        var name = ele_parts[1];
+
+        if dtype == "infer" {
+            // Try to determine the type from the entries in the symbol table
+            // Should try __RegisteredSymbols__ instead
+            var symbols = st.info("__RegisteredSymbols__");
```
pierce314159, line 240: do you ever actually use `symbols`

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+                                    errorClass="ValueError"
+                                    );
+            }
+        }
+
+        // type possibilities for pdarray and strings
+        var simpleTypes: list(string) = ["pdarray","int64", "uint8", "uint64", "float64", "bool", "strings", "string", "str"];
+        if simpleTypes.contains(dtype.toLower()) {
+            dtype = "simple";
+        }
```
pierce314159, line 267: I don't think this is necessary because we should have already overrode the dtype above?

In this if:
```Chapel
            if st.contains(name) {
                // Easy case where full name given matches an entry, pdarray or Strings
                dtype = "simple";
```

src/RegistrationMsg.chpl
```diff
@@ -96,6 +99,199 @@ module RegistrationMsg
...
+            }
+            when ("categorical") {
+                return attachCategoricalMsg(cmd, name, st);
+            }
+            when ("segarray") {
+                return attachSegArrayMsg(cmd, name, st);
+            }
+            otherwise {
+                regLogger.warn(getModuleName(),getRoutineName(),getLineNumber(), 
+                            "Unsupported type provided: '%s'. Supported types are: int64, str, categorical, segarray".format(dtype));
```
pierce314159, line 282: should this log message say `pdarray` instead of `int64`?

pierce314159 (approved):
Couple gripes but overall the logic looks good!

src/RegistrationMsg.chpl
mhmerrill: i don't think it's wrong to tell the user that name is already in use for another kind of object.

arkouda/util.py
```diff
@@ -282,13 +282,21 @@ def convert_if_categorical(values):
...
+    elif repMsg.split("+")[0] == "segarray":
+        return SegArray.from_return_msg(repMsg)
     else:
-        return create_pdarray(repMsg)
+        dtype = repMsg.split()[2]
+
+        if dtype == "str":
+            return Strings.from_return_msg(repMsg)
+        else:
+            return create_pdarray(repMsg)
```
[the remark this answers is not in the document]
joshmarshall1: It's possible, but we'd have to make sure it's updated to handle that format everywhere else the build from repMsg is used.

src/RegistrationMsg.chpl
joshmarshall1: This might be a good issue for Jim to work on when he gets setup. That should a very straight forward change

src/RegistrationMsg.chpl
joshmarshall1: No, that was leftover from debugging - good catch

src/RegistrationMsg.chpl
joshmarshall1: Not exactly. If the dtype given is infer, then it can hit that `if` where the dtype is set to `"simple"` only if the name matches a name in the table.

However, if a user passes in the result from `my_pdarray.dtype` then we have to handle `int64`. This is checking all possible supported types that could be provided from the user and simplifying them to a single case to be used in the `select` block

src/RegistrationMsg.chpl
joshmarshall1: Yes it should

arkouda/categorical.py
```diff
@@ -238,6 +238,60 @@ def set_categories(self, new_categories, NAvalue=None):
...
+            else:
+                raise ValueError(f"Unknown field, {name}, found in Categorical.")
+
+        if len(repParts) == 4:
+            parts["segments"] = create_pdarray(repParts[5])
+
+        # To get the name split the message into Categories, Codes, Permutation, Segments
+        # then split the categories into it's components, Name being second: name.categories
+        # split the name on . and take the first half to get the given name
+        name = repParts[1].split()[1].split(".")[0]
```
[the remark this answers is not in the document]
joshmarshall1: Comment added

arkouda/util.py
```diff
@@ -282,13 +282,21 @@ def convert_if_categorical(values):
     return values
 
 
-def attach(name):
+def attach(name: str, dtype: str = "infer"):
```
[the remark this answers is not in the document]
joshmarshall1: I think this should be handled in another ticket for all of `arkouda/util`

src/RegistrationMsg.chpl
pierce314159: ahh i see!

pierce314159 (approved):
Looks good!

arkouda/util.py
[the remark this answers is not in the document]
pierce314159: This is being tracked by #1373

PR #1344, opened 2022-05-02
Closes #1302 - .yml files for setup of user and developer conda environment

### LINKED ISSUE #1302: Add yaml environment file to configure anaconda environment
Add a yaml file to allow for easy configuration of conda environment and dependency installation for using arkouda.

joshmarshall1:
This ticket (closes #1302):

Added two yaml files, one for standard configuration and one for developer configuration of conda environments. The packages included were the ones listed in `setup.py`.

These files successfully generated conda environments in testing that were able to run arkouda without errors, but we should find a way to test these in a completely clean environment without other previous installations or existing packages.

### DIFF
--- a/arkouda-env-dev.yml
+++ b/arkouda-env-dev.yml
@@ -24,5 +24,5 @@ dependencies:
   - pip:
     - typeguard==2.10.0
     # Developer dependencies
-    - mypy==0.931
+    - mypy>=0.931
   
--- a/arkouda-env-dev.yml
+++ b/arkouda-env-dev.yml
@@ -1,0 +1,28 @@
+name: arkouda-env-dev
+channels:
+  - conda-forge
+dependencies:
+  - python>=3.7   # minimum 3.7
+  - numpy>=1.16.5,<=1.21.5
+  - pandas>=1.1.0
+  - pyzmq>=20.0.0
+  - tabulate
+  - pyfiglet
+  - versioneer
+  - matplotlib
+  - h5py
+  - pip
+  # Developer dependencies
+  - pexpect
+  - pytest
+  - pytest-env
+  - Sphinx
+  - sphinx-argparse
+  - sphinx-autoapi
+  - typed-ast
+  
+  - pip:
+    - typeguard==2.10.0
+    # Developer dependencies
+    - mypy==0.931
+  
--- a/arkouda-env.yml
+++ b/arkouda-env.yml
@@ -1,0 +1,16 @@
+name: arkouda-env
+channels:
+  - conda-forge
+dependencies:
+  - python>=3.7   # minimum 3.7
+  - numpy>=1.16.5,<=1.21.5
+  - pandas>=1.1.0
+  - pyzmq>=20.0.0
+  - tabulate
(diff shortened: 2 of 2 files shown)


arkouda-env-dev.yml
```diff
@@ -0,0 +1,28 @@
...
+  - pyfiglet
+  - versioneer
+  - matplotlib
+  - h5py
+  - pip
+  # Developer dependencies
+  - pexpect
+  - pytest
+  - pytest-env
+  - Sphinx
```
Ethan-DeBandi99, line 19: This should be lowercase `sphinx`.

arkouda-env-dev.yml
```diff
@@ -0,0 +1,28 @@
...
+  - sphinx-argparse
+  - sphinx-autoapi
+  - typed-ast
+  
+  - pip:
+    - typeguard==2.10.0
+    # Developer dependencies
+    - mypy==0.931
```
Ethan-DeBandi99, line 27: According to `setup.py` this should be `mypy>=0.931`. I know there were issues when your version was higher. Can we track that down and determine which case is correct and set it? We can probably add a separate issue to track this.

pierce314159 (approved):
Not a YAML or env config expert but this looks good to me

Ethan-DeBandi99 (changes requested):
Nothing major, just the one word that needs lowercase. The other comment should probably be addressed in its own PR as what is in both locations will function properly.

arkouda-env-dev.yml
mhmerrill: this is supposed to be uppercase I think.
https://github.com/Bears-R-Us/arkouda/blob/b9e53938bce78a34be0b6f1631525f5b1db95f52/setup.py#L154-L156

arkouda-env-dev.yml
mhmerrill: agree with this comment.
https://github.com/Bears-R-Us/arkouda/blob/b9e53938bce78a34be0b6f1631525f5b1db95f52/setup.py#L154-L156

mhmerrill (changes requested):
see comments

Ethan-DeBandi99:
> If we're opting to make parquet a requirement, it might be good to add the parquet dependencies to this .yml as well. @Ethan-DeBandi99 @pierce314159 @mhmerrill Thoughts?

@joshmarshall1 - we will be doing this in the future, but currently `pyarrow` and `Parquet` are optional. I agree when we switch to make these dependencies, we should add it to the `.yml`, but for now it should not be added.

arkouda-env-dev.yml
joshmarshall1: Updated to by `mypy>=0.931`
I will add an issue to track down potential version issue

Ethan-DeBandi99 (approved):
looks good.

pierce314159 (approved):

Ethan-DeBandi99:
@mhmerrill - are there further changes you want made before this PR is ready?

mhmerrill (approved):

PR #1354, opened 2022-05-04
Closes #1331 - Parquet alignment with HDF5

### LINKED ISSUE #1331: Parquet alignment with HDF5
Inconsistencies are as listed below:

- HDF5 is able to be saved with or without a file extension using the `save()` or `save_all()` methods. Parquet files always have `.parquet` appended when writing the files in `ParquetMsg.chpl`. 
- Loading HDF5 is configured so that loading is done by passing a `prefix_path` and then using `os.path.splitext()` to get the prefix and then extension and use those to load the file. This would allow `load()` to work with Parquet when doing something like `load('path_to_dir/name_prefix.parquet')`. However, when saving, the user does not have to provide the `.parquet `so this is not exactly clear when loading.
- Adding the `file_format` variable for loading requires HDF5 to retain the same functionality that was there before and then if it is set to `file_format='Parquet'` the extension must be set to `.parquet`. I have a solution doing this that I will be pushing a PR for shortly. I will link that once it is up.

Following discussion in #1328, we will be aligning the functionality of Parquet with that of HDF5.

Ethan-DeBandi99:
Closes #1331

- Updates load functionality to support `file_format='INFER'`. 
- Updates saving files as `Parquet` to not automatically add the file extension.
- `Parquet` save/load should now function the same as `HDF5`. The file extension is now optional and can be added by the user.
- Updated `IO.rst` and docstrings to reflect current I/O functionality for `HDF5` and `Parquet`.

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -305,14 +305,24 @@ module ParquetMsg {
                                         dsetname, numelems,
                                         dtype, compressed,
                                         errMsg): int;
-    var filenames: [0..#A.targetLocales().size] string;
+    // var filenames: [0..#A.targetLocales().size] string;
     var dtypeRep = toCDtype(dtype);
-    for i in 0..#A.targetLocales().size {
-      var suffix = '%04i'.format(i): string;
-      filenames[i] = filename + "_LOCALE" + suffix + ".parquet";
-    }
+    // for i in 0..#A.targetLocales().size {
+    //   var suffix = '%04i'.format(i): string;
+    //   filenames[i] = filename + "_LOCALE" + suffix + ".parquet";
+    // }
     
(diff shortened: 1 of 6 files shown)


reuster986 (approved):

joshmarshall1 (approved):
Looks good

src/ParquetMsg.chpl
```diff
@@ -305,14 +305,24 @@ module ParquetMsg {
                                         dsetname, numelems,
                                         dtype, compressed,
                                         errMsg): int;
-    var filenames: [0..#A.targetLocales().size] string;
+    // var filenames: [0..#A.targetLocales().size] string;
```
mhmerrill, line 308: i noticed a bunch of `//` commented out lines in this file. i guess these are the old code. do we really need them still?

Issue #1308, opened 2022-04-19
Should `mode='append'` create a file if the file is non-existent?


bmcdonald3:
During a conversation with @kellyjoy15 on https://github.com/Bears-R-Us/arkouda/issues/1304, the question was raised: "Why doesn't `mode='append'` create a file if the file does not exist already?" and this seems like a reasonable feature to me.

It seems if you plan to append to a file, writing it with `mode='append'` the first time may make sense, since you would think something along the lines of "I am going to append to this file, so it probably needs to be created in append mode." Today, this will just generate an error and the first file needs to be generated with `mode='truncate'`.

Ethan-DeBandi99:
@bmcdonald3 - i think this would be a good feature. If we don't want it on all the time, maybe a flag that a user can specify if they want to allow file creation if it does not exist.

pierce314159:
I don't have the most experience with IO stuff so I'll defer to others but I feel like it makes sense to be able to create a file in append mode

kellyjoy15:
As a reference, in [python's built-in open function](https://docs.python.org/3/library/functions.html#open), "append" mode means open for writing, appending to the end of the file if it exists.

"<html>
<body>


'a' | open for writing, appending to the end of file if it exists
-- | --



</body>
</html>

pierce314159:
It seems like it would be a good idea to change this behavior. Does anyone have a reason not to? @mhmerrill @reuster986 @ronawho

bmcdonald3:
To me, I think the difference between Python's `open` function and our `save()` function is we aren't just opening the file, we are 1. Opening the file 2. writing to the file and 3. closing the file, so our API is different from the Python API in that sense.

If Arkouda had a more similar workflow and API, maybe something like this:
```
my-file = ak.open_file('my-file', mode='append')
my-file.write(pdarray1)
my-file.write(pdarrau2)
...
my-file.close()
```
Then I think that would be a good reason to base our API off of Python, but I think that would require a reworking of the API. I am not arguing against that design and I actually think that would make a lot of sense, but I think it would be a change from what we have today is all.

mhmerrill:
I really don't like leaving files open. All Arkouda operations currently leave state "complete" upon message completion. That said I am ok with file creation when appending to a nonexistent file.

mhmerrill:
[...]

PR #1358, opened 2022-05-04
Create nonexistent parquet files when writing with mode='append'

### LINKED ISSUE #1366: Parquet add File on append
Result of discussion in https://github.com/Bears-R-Us/arkouda/issues/1308.

When a user saves with mode APPEND, if the file does not exist, it should be created. If the file exists, it should function as append currently does. If the file exists, but the number of files does not match the locales, we should error.

bmcdonald3:
This PR updates the current Parquet append functionality to create
the files when they do not exist. Previously, the server would
send an error message when trying to append to nonexistent files.

Related to https://github.com/Bears-R-Us/arkouda/issues/1308
Closes #1366

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -315,16 +315,23 @@ module ParquetMsg {
     var matchingFilenames = glob("%s_LOCALE*%s".format(filename, ".parquet"));
 
     var warnFlag = processParquetFilenames(filenames, matchingFilenames, mode);
+    var createFiles: bool = false;
 
     if mode == APPEND {
-      var datasets = getDatasets(filenames[0]);
-      if datasets.contains(dsetname) then
-        throw getErrorWithContext(
-                 msg="A column with name " + dsetname + " already exists in Parquet file",
-                 lineNumber=getLineNumber(), 
-                 routineName=getRoutineName(), 
-                 moduleName=getModuleName(), 
-                 errorClass='WriteModeError');
+      // Files do not exist, so we can just create them by treating as truncate
+      if warnFlag {
+        createFiles = true;
+        warnFlag = false;
+      } else {
+        var datasets = getDatasets(filenames[0]);
+        if datasets.contains(dsetname) then
+          throw getErrorWithContext(
+                    msg="A column with name " + dsetname + " already exists in Parquet file",
+                    lineNumber=getLineNumber(), 
+                    routineName=getRoutineName(), 
+                    moduleName=getModuleName(), 
+                    errorClass='WriteModeError');
+      }
     }
     
     coforall (loc, idx) in zip(A.targetLocales(), filenames.domain) do on loc {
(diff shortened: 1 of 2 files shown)


reuster986 (approved):

Ethan-DeBandi99 (changes requested):
The current configuration will work, but reading through it is a bit confusing, especially if there is no prior knowledge of what is happening. I am requesting that we change the variable `warnFlag` to `filesExist` as it is more clear when reading the code of the purpose it serves. Doing this will also trim down on a few if blocks.

We will also need to update the `return warnFlag;` to `return fileExists`.

We should probably update the `pdarrayIO.py:save_all` function logic as it will still utilize the `truncate` workflow if the file does not exist. I am not sure this should be required for this PR though. I will add an issue for this

I am assuming that we will want this to function for HDF5 as well? Otherwise, we will be back in a situation where the 2 file formats function differently on save. @mhmerrill - do you want me to add an issue to handle that?

bmcdonald3:
I have a branch taht handles HDF5 as well, no need to create another issue

src/ParquetMsg.chpl
```diff
@@ -484,9 +484,9 @@ module ParquetMsg {
       }
     } else if mode == TRUNCATE {
       if matchingFilenames.size > 0 {
-        warnFlag = true;
+        filesExist = true;
       } else {
-        warnFlag = false;
+        filesExist = false;
       }
```
Ethan-DeBandi99, line 490: Since we are now defaulting `filesExist` to true, we could simplify this block to 

```chapel
if matchingFilenames.size == 0 {
    filesExist = false;
}
```

Ethan-DeBandi99 (approved):
This looks good. Thanks for making the changes to `fileExists`. I made one note that we could simplify the logic block for setting it, but that is not required for functionality or understanding.

src/ParquetMsg.chpl
```diff
@@ -496,7 +496,7 @@ module ParquetMsg {
                  moduleName=getModuleName(), 
                  errorClass='IllegalArgumentError');
     }
-    return warnFlag;
+    return filesExist;
```
Ethan-DeBandi99, line 499: Can we update this to match the other return statement?

Ethan-DeBandi99:
Added issue and tagged as closing that issue for tracking purposes.

Ethan-DeBandi99 (changes requested):
The return needs to be updated for correct logic. Other than that looks good.

src/ParquetMsg.chpl
bmcdonald3: I don't think we want that in the processParquetFilenames function, only in the write dist array function.

Ethan-DeBandi99: You are correct. I missed the return for the write and somehow thought this was it. My bad!

Ethan-DeBandi99 (approved):
Looks good

pierce314159 (approved):

PR #1364, opened 2022-05-05
Create nonexistent HDF5 files when writing with mode='append'

### LINKED ISSUE #1365: HDF5 Create File on Append
Result of discussion in #1308.

When a user saves with mode `APPEND`, if the file does not exist, it should be created. If the file exists, it should function as append currently does. If the file exists, but the number of files does not match the locales, we should error.

bmcdonald3:
This PR updates the current HDF5 append functionality to create
the files when they do not exist. Previously, the server would
send an error message when trying to append to nonexistent files.

Closes https://github.com/Bears-R-Us/arkouda/issues/1308
Closes #1365

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -1062,7 +1062,7 @@ module HDF5Msg {
     proc write1DDistStringsAggregators(filename: string, mode: int, dsetName: string, entry:SegStringSymEntry, writeOffsets:bool) throws {
         var prefix: string;
         var extension: string;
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
This looks good. Thanks for carrying the changes I requested on Parquet over!

Ethan-DeBandi99:
Added a ticket specific to HDF5 for tracking this and linked here

pierce314159 (approved):

PR #1391, opened 2022-05-13
Closes #1389: Bug fix for categorical._akNAcode with generic attach

### LINKED ISSUE #1389: Update generic attach to return Categorical `_akNAcode`
When the generic attach support for `SegArray` and `Categorical` was added, `Categorical` didn't have the `_akNAcode` field. Now that is does, using generic attach to attach to a Categorical won't pull back `_akNAcode` and trying to `unregister` a generically attached `Categorical` will leave `_akNAcode` behind.

joshmarshall1:
This PR (closes #1389):

Added `_akNAcode` to the generic attach and categorical `from_return_msg` processing. Also fixed where in Categorical `init` the Categorical's `.segments` was getting set to `.permutation` causing a similar bug.

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -76,7 +76,7 @@ class Categorical:
                 # Permutation and segments should only ever be supplied together from
                 # the .from_codes() method, not user input
                 self.permutation = cast(pdarray, self.permutation)
-                self.segments = cast(pdarray, self.permutation)
+                self.segments = cast(pdarray, self.segments)
                 unique_codes = self.codes[self.permutation[self.segments]]
             else:
                 unique_codes = unique(self.codes)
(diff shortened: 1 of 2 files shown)


arkouda/categorical.py
```diff
@@ -268,20 +268,19 @@ def from_return_msg(repMsg):
...
-        if len(repParts) == 4:
-            parts["segments"] = create_pdarray(repParts[5])
+        if len(repParts) > 5:
+            for i in range(5, len(repParts)):
+                name = repParts[i].split()[1]
+                if ".permutation" in name:
+                    parts["permutation"] = create_pdarray(repParts[i])
+                elif ".segments" in name:
+                    parts["segments"] = create_pdarray(repParts[i])
+                else:
```
pierce314159, line 282: I think this only works if both `.permutation` and `.segments` exists and won't if only one does. I think to handle either existing, you need to do something similar to what was originally there

```python
if len > 4:
  # at least one exists

if len == 5:
   # both exist so handle the other one
```

I think this original code wasn't super obvious, so it might be worth adding some comments explaining

pierce314159 (changes requested):
small change in `from_return_msg`

arkouda/categorical.py
pierce314159: ohhhh i think i misread this! Nevermind this looks perfect!

pierce314159 (approved):

mhmerrill (approved):
looks ok to me.

mhmerrill:
Is this comprehended in a test? how did it slip through testing before?

Issue #1399, opened 2022-05-18
`ak.util.register()` throws an error


joshmarshall1:
While working #1381 I discovered attempting to register a `Series` object results in an error message. 

The root cause seems to be the `ak.util.register` method attempting to parse the arkouda version number as a `datetime`, which in the case of the current master branch is `2022.05.09+14.gd0b5bef2`. The datetime parser doesn't know what to do with this and as a result it's throwing `ParserError: Unknown string format: 2022.05.09+14.gd0b5bef2`

It's likely this works just fine for release versions that are only the date and don't have the extra hash. So I would guess that users aren't experiencing this issue.

Issue #1381, opened 2022-05-11
Add `is_registered` to SegArray and Series


joshmarshall1:
`SegArray` and `Series` are the only register-able objects that don't contain a method for checking if they've been registered so this method should be added.

PR #1400, opened 2022-05-18
Part of #1381 - Added `SegArray.is_registered` method

joshmarshall1:
This PR is part of #1381 

The issue this is for is to add `is_registered` to `SegArray` and `Series`. However, currently the `register` functionality isn't working for `Series` objects and needs to be corrected before `is_registered` can be added and tested. Issue #1399 tracks the fix of `Series.register`

In this PR I added `is_registered()` to `SegArray` and modified the existing registration test case to also check the three cases for is_registered; All parts registered, Some parts registered, or No parts registered.

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -1091,7 +1091,6 @@ class SegArray:
 
         if any(regParts) and not all(regParts):
             warn(f"SegArray expected {len(regParts)} components to be registered, but only located {sum(regParts)}")
-            return False
-        else:
-            return all(regParts)
+
+        return all(regParts)
         


arkouda/segarray.py
```diff
@@ -1070,3 +1071,29 @@ def attach(cls, name, segment_suffix='_segments', value_suffix='_values', length
...
+        # SegArray contains 3 parts - segments, values, and lengths
+        regParts = sum([self.segments.is_registered(), self.values.is_registered(), self.lengths.is_registered()])
+
+        if regParts == 3:
+            return True
+        elif regParts == 0:
+            return False
+        else:
+            warn(f"SegArray expected 3 components to be registered, but only located {regParts}")
+            return False
```
pierce314159, line 1098: It's personal preference, but I really like using `any` and `all`, I feel it's more descriptive since reading the `if` tells you exactly what the problem is

```python
parts_registered = [self.segments.is_registered(), self.values.is_registered(), self.lengths.is_registered()]

if any(parts_registered) and not all(parts_registered):
  warn(f"SegArray expected {len(parts_registered)} components to be registered, but only located {sum(parts_registered)}")
return all(parts_registered)    
```

this has the added benefit of requiring fewer lines be changed if we add more registerable components (such as `self.grouping`)

pierce314159 (approved):
This logic looks good! I left a suggestion but it's personal preference

pierce314159 (approved):

Ethan-DeBandi99 (approved):

PR #1403, opened 2022-05-18
Closes #1399: Removing version check from `ak.util.register`

### LINKED ISSUE #1399: `ak.util.register()` throws an error

joshmarshall1:
This PR (closes #1399):

The datetime parse in `ak,util.register()` was checking if the current arkouda version was `>= '2021.04.14'` and doing one process if True, and another if False in support of previous versions. This is likely a leftover of when `ak.util` was separate from the main codebase as the current arkouda version will never be less than or equal to `'2021.04.14'` in the current release, `'2022.05.09'` 

Removing this check will allow the `register` method to function properly, but we should look into either updating to be fully generic or removing this method entirely and splitting it's special case into `pdarray.register` if necessary.

### DIFF
--- a/arkouda/util.py
+++ b/arkouda/util.py
@@ -67,12 +67,7 @@ def register(a, name):
     Register an arkouda object with a user-specified name. Backwards compatible
     with earlier arkouda versions.
     """
-    if pd.to_datetime(__version__.lstrip('v')) >= pd.to_datetime('2021.04.14'):
-        # New versions register in-place and don't need callback
-        cb = identity
-    else:
-        # Older versions return a new object that loses higher-level dtypes and must be re-converted
-        cb = get_callback(a)
+    cb = identity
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

Ethan-DeBandi99 (approved):
looks good

PR #1407, opened 2022-05-19
1406 Fix bug in ak.lookup

### LINKED ISSUE #1406: ak.lookup gives incorrect result when keys are not ordered

reuster986:
Closes #1406 

Previously, the `ak.lookup()` function, which treats key-value pairs as a function to be applied over args, implicitly assumed that the user-supplied keys were in the order output by `ak.unique()` and `ak.GroupBy()`. This is the most common usage, but it is not inherent to the semantics of the function, and when a user (legitimately) supplied keys that were unique but unordered, `ak.lookup()` would return the wrong answer. 

This PR fixes that behavior by performing `ak.GroupBy(keys)` to ensure the anticipated ordering. Because this can be an expensive step, it also adds a `keys_from_unique` keyword arg as a shortcut for when keys come from unique/GroupBy.

This PR also adds a unit test for `ak.lookup` in the `join_test.py` file (because lookup is a special case of a join).

### DIFF
--- a/tests/join_test.py
+++ b/tests/join_test.py
@@ -105,6 +105,27 @@ class JoinTest(ArkoudaTest):
         with self.assertRaises(ValueError):
             l, r = ak.join.inner_join(left, right, wherefunc=ak.intersect1d, whereargs=(ak.arange(10), ak.arange(5)))
 
+    def test_lookup(self):
+        keys = ak.arange(5)
+        values = 10*keys
+        args = ak.array([5, 3, 1, 4, 2, 3, 1, 0])
+        ans = np.array([-1, 30, 10, 40, 20, 30, 10, 0])
+        # Simple lookup with int keys
+        # Also test shortcut for unique-ordered keys
+        res = ak.lookup(keys, values, args, fillvalue=-1, keys_from_unique=True)
+        self.assertTrue((res.to_ndarray() == ans).all())
+        # Compound lookup with (str, int) keys
+        res2 = ak.lookup((ak.cast(keys, ak.str_), keys), values, (ak.cast(args, ak.str_), args), fillvalue=-1)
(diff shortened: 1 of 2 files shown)


joshmarshall1 (approved):

tests/join_test.py
```diff
@@ -105,6 +105,27 @@ def test_inner_join(self):
...
 
+    def test_lookup(self):
+        keys = ak.arange(5)
+        values = 10*keys
+        args = ak.array([5, 3, 1, 4, 2, 3, 1, 0])
+        ans = np.array([-1, 30, 10, 40, 20, 30, 10, 0])
+        # Simple lookup with int keys
+        # Also test shortcut for unique-ordered keys
+        res = ak.lookup(keys, values, args, fillvalue=-1, keys_from_unique=True)
+        self.assertTrue((res.to_ndarray() == ans).all())
```
pierce314159, line 116: I know it's not consistent across our tests but i prefer
```python
self.assertListEqual(res.to_ndarray().tolist(), ans.tolist())
```
because it gives more precise information about where the lists are different

for example
```python
>       self.assertTrue((ak.array([1,2,3]) == ak.array([1,1,3])).all())
E       AssertionError: False is not true
```
vs
```python
>       self.assertListEqual(ak.array([1,2,3]).to_ndarray().tolist(), ak.array([1,1,3]).to_ndarray().tolist())
E       AssertionError: Lists differ: [1, 2, 3] != [1, 1, 3]
E       
E       First differing element 1:
E       2
E       1
E       
E       - [1, 2, 3]
E       ?     ^
E       
E       + [1, 1, 3]
E       ?     ^
```

mhmerrill (approved):
looks ok to me.

PR #1408, opened 2022-05-19
Closes #953 - Import/Export Functionality

### LINKED ISSUE #953: Investigate importing hdf5 file created from Pandas/PyTables
A community member working through the TaxiCab dataset (https://github.com/Bears-R-Us/ArkoudaNotebooks/blob/master/NYCTaxi.ipynb) had modified the notebook to write the dataframe out to HDF5 via pandas dataframe.
i.e.
```python
df.to_hdf(out_filename, key="data", format="table", mode="w")
```

While Arkouda is capable of reading HDF5 as a format, output like above is not something it can currently handle.   We should investigate what it would take to read in a simple data frame / table exported from Pandas.  If it's a "quick win" this may be a good alternative for folks to import data into Arkouda instead of trying to pass the data over the ZMQ socket.

Note: Under the hood I believe the Pandas implementation uses PyTables to format and write the data into HDF5.
- https://www.pytables.org/
- https://github.com/PyTables/PyTables
- https://portal.hdfgroup.org/display/HDF5/HDF5

Ethan-DeBandi99:
Closes #953 

- Adds Import/Export functionality between Arkouda and Pandas. This is currently all done client side and is intended as an initial iteration. More support will be added as needed.
- Updates DataFrame loading to properly order columns when loading from a Parquet file.
- Adds testing of import/export
- IO documentation updated to include Import/Export.

The current implementation utilizes pandas for the most functionality. 
Import Workflow:
1. Load pandas dataframe from file
2. Build arkouda dataframe from pandas object

Export Workflow:
1. Load arkouda dataframe from file
2. Build pandas dataframe from arkouda object with `to_pandas()` method

Import and export can be configured to save a new file formatted properly and/or return the resulting object.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1,8 +1,5 @@
 from __future__ import annotations
 
-from datetime import datetime
-from ipaddress import ip_address
-from tabulate import tabulate
 from collections import UserDict
 from warnings import warn
 import pandas as pd  # type: ignore
@@ -11,6 +8,7 @@ import random
 import json
 from typing import cast, Union, List, Callable
 from typeguard import typechecked
+import os
 
 from arkouda.segarray import SegArray
 from arkouda.pdarrayclass import pdarray
@@ -22,7 +20,7 @@ from arkouda.groupbyclass import GroupBy as akGroupBy, unique
 from arkouda.pdarraysetops import concatenate, intersect1d, in1d
 
 from arkouda.pdarrayIO import save_all, load_all
-from arkouda.dtypes import int64 as akint64, uint64 as akuint64, float64 as akfloat64, bool as akbool
+from arkouda.dtypes import int64 as akint64, float64 as akfloat64, bool as akbool
 from arkouda.sorting import argsort, coargsort
 from arkouda.numeric import where
 from arkouda.client import maxTransferBytes, generic_msg
@@ -31,6 +29,7 @@ from arkouda.alignment import in1dmulti
 from arkouda.series import Series
 from arkouda.index import Index
 from arkouda.numeric import cast as akcast, isnan as akisnan
+from arkouda.pdarrayIO import get_filetype
 
 # This is necessary for displaying DataFrames with BitVector columns,
 # because pandas _html_repr automatically truncates the number of displayed bits
@@ -1289,7 +1288,16 @@ class DataFrame(UserDict):
 
     @classmethod
     def load_table(cls, prefix_path, file_format='INFER'):
-        return cls(load_all(prefix_path, file_format=file_format))
+        prefix, extension = os.path.splitext(prefix_path)
+        first_file = "{}_LOCALE0000{}".format(prefix, extension)
+        filetype = get_filetype(first_file)
+        # columns load backwards
+        df = cls(load_all(prefix_path, file_format=filetype))
+        if filetype == "HDF5":
+            return df
+        else:
+            # return the dataframe with them reversed so they match what was saved. This is only an issue with parquet
+            return df[df.columns[::-1]]
 
     def argsort(self, key, ascending=True):
         """
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -683,8 +683,6 @@ def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None
 
     Raises
     ______
-    RuntimeWarning
-        - Export attempted on Parquet file. Arkouda formatted Parquet files are readable by pandas.
     RuntimeError
         - Unsupported file type
 
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -591,7 +591,7 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path : 
 
 
 def import_data(read_path: str, write_file: str = None,
-                return_obj: bool = False, index: bool = False):
+                return_obj: bool = True, index: bool = False):
     """
     Import data from Pandas to Arkouda readable format
 
@@ -602,7 +602,7 @@ def import_data(read_path: str, write_file: str = None,
     write_to: str, optional
         path to file to write arkouda formatted data to. Only write file if provided
     return_obj: bool, optional
-        Default False. When True return the Arkouda DataFrame object
+        Default True. When True return the Arkouda DataFrame object, otherwise return None
     index: bool, optional
         Default False. When True, maintain the indexes loaded from the pandas file
 
@@ -663,7 +663,7 @@ def import_data(read_path: str, write_file: str = None,
 
 
 def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None,
-           return_obj: bool = False, index: bool = False):
+           return_obj: bool = True, index: bool = False):
     """
     Export data from arkouda to Pandas readable format
 
@@ -678,7 +678,7 @@ def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None
     write_file: str, optional
         path to file to write arkouda formatted data to. Only write the file if this is set
     return_obj: bool, optional
-        Default False. When True return the Arkouda DataFrame object
+        Default True. When True return the Pandas DataFrame object, otherwise return None
 
 
     Raises
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -669,8 +669,8 @@ def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None
 
     Parameters
     __________
-    rd_filename: str
-        path to file where pandas data is stored.
+    read_path: str
+        path to file where arkouda data is stored.
     dataset_name: str
         name to store dataset under
     index: bool
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -590,6 +590,7 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path : 
             arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode='append')
 
 
+@typechecked
 def import_data(read_path: str, write_file: str = None,
                 return_obj: bool = True, index: bool = False):
     """
@@ -632,7 +633,6 @@ def import_data(read_path: str, write_file: str = None,
     from arkouda.dataframe import DataFrame
     # verify file path
     is_glob = False if os.path.isfile(read_path) else True
-    print(is_glob)
     file_list = glob.glob(read_path)
     if len(file_list) == 0:
         raise FileNotFoundError(f"Invalid read_path, {read_path}. No files found.")
@@ -660,7 +660,7 @@ def import_data(read_path: str, write_file: str = None,
     if return_obj:
         return df
 
-
+@typechecked
 def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None,
            return_obj: bool = True, index: bool = False):
     """
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -676,7 +676,7 @@ def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None
     index: bool
         Default False. When True, maintain the indexes loaded from the pandas file
     write_file: str, optional
-        path to file to write arkouda formatted data to. Only write the file if this is set
+        path to file to write pandas formatted data to. Only write the file if this is set
     return_obj: bool, optional
         Default True. When True return the Pandas DataFrame object, otherwise return None
 
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -627,9 +627,7 @@ def import_data(read_path: str, write_file: str = None,
 
     Notes
     _____
-    - Import can only be performed from hdf5 or parquet files.
-    - Import is only designed to import data from Pandas. Additional format support will require attribute validation
-    to occur before parsing the file.
+    - Import can only be performed from hdf5 or parquet files written by pandas.
     """
     from arkouda.dataframe import DataFrame
     # verify file path
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -601,7 +601,7 @@ def import_data(read_path: str, write_file: str = None,
     __________
     read_path: str
         path to file where pandas data is stored. This can be glob expression for parquet formats.
-    write_to: str, optional
+    write_file: str, optional
         path to file to write arkouda formatted data to. Only write file if provided
     return_obj: bool, optional
         Default True. When True return the Arkouda DataFrame object, otherwise return None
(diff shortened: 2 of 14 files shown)


Ethan-DeBandi99:
Because the CI runs with Parquet support enabled, the tests are running for parquet. Because the import export is client side, pyarrow is now required. I added this to the setup.py. Since we are looking into making parquet a dependency, I do not see this as a problem. If anyone else has an issue with this, please let me know.

arkouda/dataframe.py
```diff
@@ -21,7 +19,7 @@
 from arkouda.pdarraysetops import concatenate, intersect1d, in1d
 
 from arkouda.pdarrayIO import save_all, load_all
-from arkouda.dtypes import int64 as akint64, uint64 as akuint64, float64 as akfloat64, bool as akbool
+from arkouda.dtypes import int64 as akint64, float64 as akfloat64, bool as akbool
```
mhmerrill, line 22: i noticed you removed uint64 here... do we need it?

arkouda/dataframe.py
```diff
@@ -1248,7 +1247,16 @@ def save_table(self, prefix_path, columns=None, index=False, file_format='HDF5')
...
     def load_table(cls, prefix_path, file_format='INFER'):
-        return cls(load_all(prefix_path, file_format=file_format))
+        prefix, extension = os.path.splitext(prefix_path)
+        first_file = "{}_LOCALE0000{}".format(prefix, extension)
+        filetype = get_filetype(first_file)
+        # columns load backwards
+        df = cls(load_all(prefix_path, file_format=filetype))
+        if filetype == "HDF5":
+            return df
+        else:
```
kellyjoy15, line 1298: What about explicitly checking the filetype, then raising an exception if it is not "HDF5" or "Parquet"? This might save trouble down the road if some other filetype is added... Mostly pointing out because there have been other bugs caused by similar assumptions...
```
elif filetype == "Parquet":
            # return the dataframe with them reversed so they match what was saved. This is only an issue with parquet
            return df[df.columns[::-1]]
else:
            #raise some appropriate exception...
```

arkouda/dataframe.py
```diff
@@ -21,7 +19,7 @@
 from arkouda.pdarraysetops import concatenate, intersect1d, in1d
 
 from arkouda.pdarrayIO import save_all, load_all
-from arkouda.dtypes import int64 as akint64, uint64 as akuint64, float64 as akfloat64, bool as akbool
```
kellyjoy15, line 25: I am probably just unaware, but is there a reason we no longer need to import uint64 datatype here?

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+    See Also
+    ________
+    pandas.DataFrame.to_parquet, pandas.DataFrame.to_hdf,
+    pandas.DataFrame.read_parquet, pandas.DataFrame.read_hdf,
+    ak.export
+
+    Notes
+    _____
+    - Import can only be performed from hdf5 or parquet files.
+    - Import is only designed to import data from Pandas. Additional format support will require attribute validation
```
kellyjoy15, line 630: These two import comments seem contradictory. I believe a longer explanation is needed, or a reference to a deep-dive into the difference in how pandas and arkouda write parquet/hdf5 files.

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+        path to file to write arkouda formatted data to. Only write file if provided
+    return_obj: bool, optional
+        Default False. When True return the Arkouda DataFrame object
+    index: bool, optional
+        Default False. When True, maintain the indexes loaded from the pandas file
+
+    Raises
+    ______
+    RuntimeWarning
+        - Export attempted on Parquet file. Arkouda formatted Parquet files are readable by pandas.
```
kellyjoy15, line 612: I'm confused by the warning. Are we importing or exporting? What is an "Arkouda formatted parquet file"?

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
             first_iter = False
         else:
             arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode='append')
+
+
+def import_data(read_path: str, write_file: str = None,
+                return_obj: bool = False, index: bool = False):
+    """
+    Import data from Pandas to Arkouda readable format
```
kellyjoy15, line 596: Maybe a longer explanation here? "import data" makes me think we are importing data into arkouda. Are we just re-writing it in a different format? Which pandas formats are allowed? Is it only from a file or can I pass it a pandas object directly?

Question: Let's say I have an arkouda dataframe, akdf, with special column types, like bitvector, ip address, etc. I transfer this to a local pandas object by akdf.to_pandas(), then write it out to a parquet or hdf5 file using pandas. If I use this "import_data" function, will I get back what I started with? If not, that would be important to explain here.

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+        df.save_table(write_file, index=index, file_format=filetype)
+
+    if return_obj:
+        return df
+
+
+def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None,
+           return_obj: bool = False, index: bool = False):
+    """
+    Export data from arkouda to Pandas readable format
```
kellyjoy15, line 668: Which pandas-readable format(s)?

arkouda/dataframe.py
Ethan-DeBandi99: @mhmerrill - it was unused, so I removed it

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+
+    Parameters
+    __________
+    rd_filename: str
+        path to file where pandas data is stored.
```
kellyjoy15, line 673: Parameter doesn't match the function definition. Also, I am confused about whether we are reading or writing pandas data. Is this where we *will* write out the "pandas readable" data to?

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+    dataset_name: str
+        name to store dataset under
+    index: bool
+        Default False. When True, maintain the indexes loaded from the pandas file
+    write_file: str, optional
+        path to file to write arkouda formatted data to. Only write the file if this is set
+    return_obj: bool, optional
+        Default False. When True return the Arkouda DataFrame object
```
kellyjoy15, line 681: Intuitively, I feel like the default should be True.

"When True **return the Arkouda DataFrame object**": This seems to contradict line 694, which says a pd.DataFrame object will be returned.

pydoc/usage/IO.rst
```diff
@@ -97,4 +97,17 @@ Loading persisted DataFrame data from disk
 -------------------------------------------
 This functionality allows the columns be loaded as datasets, which creates a mapping of column names to column data. This structure is supported by the ``DataFrame`` constructor and is used to reconstruct the ``DataFrame``
 
-.. autofunction:: arkouda.DataFrame.load_table
\ No newline at end of file
+.. autofunction:: arkouda.DataFrame.load_table
+
+Import/Export
+=============
+Arkouda is able to import Pandas data from save files. Additionally, Arkouda is able to export data to pandas formatting. The file formats supported are:
```
kellyjoy15, line 104: Confusing. An explanation is needed. After reading, #953, I think I understand what is going on. I wouldn't have been able to figure it out from reading this.

mhmerrill:
@Ethan-DeBandi99 fix merge conflict in imports please.

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+    Parameters
+    __________
+    rd_filename: str
+        path to file where pandas data is stored.
+    dataset_name: str
+        name to store dataset under
+    index: bool
+        Default False. When True, maintain the indexes loaded from the pandas file
+    write_file: str, optional
+        path to file to write arkouda formatted data to. Only write the file if this is set
```
kellyjoy15, line 679: Sorry. confusing again. I thought we were exporting into a pandas-readable format.

Ethan-DeBandi99:
> @Ethan-DeBandi99 fix merge conflict in imports please.

@mhmerrill - done

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,146 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+    df = DataFrame(df_def)
+
+    if write_file:
+        df.save_table(write_file, index=index, file_format=filetype)
+
+    if return_obj:
+        return df
+
+
+def export(read_path: str, dataset_name: str = "ak_data", write_file: str = None,
```
kellyjoy15, line 665: If this is meant as complementary to "import_data", should it be called "export_data" instead?

kellyjoy15 (commented):
Changes primarily requested to documentation.

arkouda/dataframe.py
Ethan-DeBandi99: The uint64 import was unused in the file so it was removed.

arkouda/pdarrayIO.py
Ethan-DeBandi99: `import` is a reserved word in python, so the function name needed to be something different for `import`, hence `import_data`. Export is not a reserved word, so we could simply use export. I can update this to be export_data if that is what is preferred.

arkouda/pdarrayIO.py
Ethan-DeBandi99: This is actually not valid anymore. I had something working differently that did not account for multi-locale files and neglected to remove the leftover warning. It has been removed.

arkouda/pdarrayIO.py
Ethan-DeBandi99: Again, missed fixing the docstring. Initially I had this always writing the file, but because that is also optional, it is probably best to default to returning the object. I updated both import and export to return the object and corrected the docstring wording.

arkouda/pdarrayIO.py
Ethan-DeBandi99: You are correct here, apparently I did a terrible job of updating this docstring. This has been corrected.

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,144 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+    Notes
+    _____
+    - Import can only be performed from hdf5 or parquet files.
+    - Import is only designed to import data from Pandas. Additional format support will require attribute validation
+    to occur before parsing the file.
+    """
+    from arkouda.dataframe import DataFrame
+    # verify file path
+    is_glob = False if os.path.isfile(read_path) else True
+    print(is_glob)
```
pierce314159, line 636: stray print statement

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,144 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
             first_iter = False
         else:
             arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode='append')
+
+
+def import_data(read_path: str, write_file: str = None,
```
pierce314159, line 593: should these be `@typechecked`?

arkouda/pdarrayIO.py
pierce314159: i don't think these are contradictory but a rewording could be nice. Maybe something like
```
- Currently import can only be performed on hdf5 and parquet files.
- Currently import only supports data written using pandas into one of the above file formats.
```

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,144 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+                return_obj: bool = True, index: bool = False):
+    """
+    Import data from Pandas to Arkouda readable format
+
+    Parameters
+    __________
+    read_path: str
+        path to file where pandas data is stored. This can be glob expression for parquet formats.
+    write_to: str, optional
```
pierce314159, line 602: this doesn't match the parameter name in the function def (`write_file`)

arkouda/pdarrayIO.py
Ethan-DeBandi99: I updated the description. I also added a note under the notes section that files saved by arkouda when written out to a pandas format will be the same file type. Parquet --> Parquet and HDF5 --> HDF5.

arkouda/pdarrayIO.py
```diff
@@ -581,3 +588,144 @@ def save_all(columns : Union[Mapping[str,pdarray],List[pdarray]], prefix_path :
...
+    See Also
+    ________
+    pandas.DataFrame.to_parquet, pandas.DataFrame.to_hdf,
+    pandas.DataFrame.read_parquet, pandas.DataFrame.read_hdf,
+    ak.import_data
+
+    Notes
+    _____
+    - Import can only be performed from hdf5 or parquet files.
+    - Import is only designed to import data from Pandas.
```
pierce314159, line 706: these should be updated/removed for export

pydoc/usage/IO.rst
pierce314159: a potential rewording could be
```
Import allows users to import data written by pandas into arkouda. Export allows users to write arkouda data into a format pandas can read. The file formats supported are:
```

pierce314159 (approved):
few notes, looks good to me!

pydoc/usage/IO.rst
Ethan-DeBandi99: I updated the explanation to hopefully give a better understanding of what the functionality is.

arkouda/pdarrayIO.py
Ethan-DeBandi99: I updated the docstring to hopefully help. You have the option to return the object or save a file. Import/Export is intended to be done using a saved file. The DataFrame d`__init__` already accepts and pandas dataframe object to build from. That is actually how objects are built, so to answer your scenario, yes you should get back what you started with.

arkouda/dataframe.py
Ethan-DeBandi99: @pierce314159 & @kellyjoy15 I initially had it that way, but it seemed redundant to do the same check again. @kellyjoy15 do you have any issues with leaving this?

arkouda/pdarrayIO.py
Ethan-DeBandi99: @pierce314159 - I reworded slightly different (one bullet), but I think it captures what you are stating.

pydoc/usage/IO.rst
Ethan-DeBandi99: @pierce314159 - I further updated this to use your wording for the first section. I think that combined with the additional details should capture what we need.

arkouda/pdarrayIO.py
Ethan-DeBandi99: removed

arkouda/pdarrayIO.py
Ethan-DeBandi99: Added. Good catch.

arkouda/pdarrayIO.py
Ethan-DeBandi99: Updated

arkouda/pdarrayIO.py
Ethan-DeBandi99: Updated to match suggestion for import, but reworded for export.

Ethan-DeBandi99:
@kellyjoy15 / @pierce314159 - I believe I addressed everything. Please let me know if there is anything I missed or if any of my changes do not satisfy what you were asking for. Thanks!

pierce314159 (approved):

joshmarshall1 (approved):
Looks good, all prior issues appear to be resolved

PR #1410, opened 2022-05-19
Closes #1381: Adding `is_registered()` to Series and Index

### LINKED ISSUE #1381: Add `is_registered` to SegArray and Series

joshmarshall1:
This ticket (closes #1381):

This PR adds the `is_registered` method to `Series` and `Index`, as well as a test for the `register` and `attach` functionality with `Series`, but it should be noted that neither `Series` nor `Index` have an `unregister` method

### DIFF
--- a/arkouda/index.py
+++ b/arkouda/index.py
@@ -1,4 +1,5 @@
 import pandas as pd  # type: ignore
+from typing import Optional
 
 from arkouda.pdarrayclass import pdarray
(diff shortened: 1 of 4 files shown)


pierce314159 (approved):
Looks good!

Ethan-DeBandi99 (approved):
Looks good

PR #1413, opened 2022-05-19
Closes #1412 - Add .flake to gitignore

### LINKED ISSUE #1412: Add .flake to git ignore
Using `flake8` there is a configuration file `.flake` that is added to the root of arkouda. This needs to be added to the git ignore.

Ethan-DeBandi99:
Closes #1412 

- Adds `.flake` to `.gitignore`
- Adds `.DS_Store` to `gitignore`


pierce314159 (changes requested):
Full disclosure, i was the one who suggested using `.flake8` so this on me, but i figured out that we can instead use `setup.cfg`, which we are already using for `versioneer`. So I'm requesting that we update `setup.cfg` to look like this

```
[versioneer]
VCS = git
style = pep440
versionfile_source = arkouda/_version.py
tag_prefix =
parentdir_prefix = arkouda-

[isort]
profile = black

[flake8]
max-line-length = 88
extend-ignore = E203
```

We can still git ignore `.flake8` in case anyone wants to override our default

Ethan-DeBandi99:
> [isort]
> profile = black
> 
> [flake8]
> max-line-length = 88
> extend-ignore = E203

@pierce314159 - `setup.cfg` has been updated.

Ethan-DeBandi99:
@pierce314159 - we may want to update max line length. At least in my editor, it is showing lines as too long that aren't even close.

pierce314159:
> @pierce314159 - we may want to update max line length. At least in my editor, it is showing lines as too long that aren't showing as long at all. Based on what I am seeing I would recommend setting it at 110. Though, I would like to see if maybe my editor is just wider than others or something.

yeah this is one of the few cases where i don't like PEP 8. I'm with you, i prefer longer lines for readability

Just so everyone is aware, black already exceeds [PEP 8's recommendation of 79 characters per line](https://peps.python.org/pep-0008/#maximum-line-length) with [it's default of 88 characters](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length)

It appears like line length is one of the only configurable things in `black` with the `--line-length` flag. I don't know if this can be set to higher than 88 or only lower tho. But i'm down for a more sensible default as long as we're all consistent about it since we all have large monitors.... and i love my long one liners :)

pierce314159 (approved):

Ethan-DeBandi99:
I updated the `max_line_length` to 105. This looks like it is a good setting. This requires black to be run as `black --line-length 105 <filename>`.  If using black as an external tool in pycharm this can be set to run automatically.

PR #1414, opened 2022-05-19
Closes #1401 - DataFrame`save_table` bug

### LINKED ISSUE #1401: DataFrame Index Save
DataFrames have been updated to utilze `ak.Index` objects for their indexes. When saving, `ak.Index` does not have a save function. Either `ak.DataFrame.save_table` needs to be updated to accomodate for this or `ak.Index.save` needs to be added.

Ethan-DeBandi99:
Closes #1401 

- Adds `ak.Index.save` method to save index objects.
- Adds test for `ak.Index.save`
- Adds test to `ak.DataFrame` testing to validate function of save with `index=True`

### DIFF
--- a/arkouda/index.py
+++ b/arkouda/index.py
@@ -1,15 +1,19 @@
 import pandas as pd  # type: ignore
 from typing import Optional
+import json
+from typing import cast as typecast
 
 from arkouda.pdarrayclass import pdarray
 from arkouda.pdarraycreation import arange, ones
 from arkouda.pdarraysetops import argsort, in1d
 from arkouda.sorting import coargsort
-from arkouda.dtypes import int64, float64, bool
+from arkouda.dtypes import int64 as akint64, float64 as akfloat64, bool as akbool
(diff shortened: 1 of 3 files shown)


joshmarshall1 (approved):

arkouda/index.py
```diff
@@ -113,6 +116,96 @@ def lookup(self, key):
...
+            is case insensitive.
+
+        Returns
+        -------
+        string message indicating result of save operation
+
+        Raises
+        ------
+        RuntimeError
+            Raised if a server-side error is thrown saving the pdarray
```
pierce314159, line 152: `saving the Index`?

pierce314159 (approved):
looks good

arkouda/index.py
Ethan-DeBandi99: @pierce314159 - I left it as pdarray because it is only saving self.index which is a pdarray. It made sense in my head, but I am happy to update if you think it should be changed.

PR #1423, opened 2022-05-20
Closes # 1167 - `ak.DataFrame` Add `inplace` parameter

### LINKED ISSUE #1167: `ak.DataFrame` Add `inplace` parameter
`pandas.DataFrame` offers the ability to return a new object or replace the calling object when performing certain actions. The default in `pandas` is `inplace=False`.  Because the currently functionality in `ak.DataFrame` (ported from `akutil.DataFrame`) is to return a new object in some instances and replace in others, the default  in `ak.DataFrame` functions will vary if we maintain the current functionality. I would suggest aligning with `pandas`. 

Functions to update:
- `ak.DataFrame.drop`
- `ak.DataFrame.rename`
- `ak.DataFrame.reset_index`

Ethan-DeBandi99:
Closes #1167 

Adds the `inplace` parameter to the following functions:
- `drop`
- `rename`
- `reset_index`

These functions will perform the operation on `self` when `inplace` is `True`. When `False` a copy is made using `ak.DataFrame.copy()` and the operation is performed. The resulting object is returned. When operations are performed in place `None` is returned. This maintains consistency with `Pandas`. 

Minor adjustments were needed to `ak.DataFrame.copy` to prevent drop column from applying to all objects. 

Testing has been updated to validate the updated functions work for both `in-place` and `out-of-place` operations.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -9,7 +9,8 @@ import pandas as pd  # type: ignore
 import numpy as np  # type: ignore
 import random
 import json
-from typing import cast
+from typing import cast, Union, List
+from typeguard import typechecked
 
 from arkouda.segarray import SegArray
 from arkouda.pdarrayclass import pdarray
@@ -610,7 +611,8 @@ class DataFrame(UserDict):
             The labels to be dropped on the given axis
         """
         for key in keys:
-            del self[key]
+            self.pop(key, None)
+            # del self[key]
 
     def _drop_row(self, keys):
         """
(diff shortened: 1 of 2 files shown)


arkouda/dataframe.py
```diff
@@ -610,7 +611,8 @@ def _drop_column(self, keys):
             The labels to be dropped on the given axis
         """
         for key in keys:
-            del self[key]
+            self.pop(key, None)
+            # del self[key]
```
joshmarshall1, line 615: Looks like you left in the commented out original `del self[key]`

arkouda/dataframe.py
```diff
@@ -1434,11 +1475,13 @@ def copy(self, deep=True):
             res._size = self._size
             res._bytes = self._bytes
             res._empty = self._empty
-            res._columns = self._columns
+            res._columns = self._columns[:]  # if this is not a slice, droping columns modifies both
 
             for key, val in self.items():
                 res[key] = val[:]
 
+            res._set_index(Index(self.index.index))
```
joshmarshall1, line 1483: Just double checking this is supposed to be `self.index.index` and not `self.index`?

joshmarshall1 (approved):
Looks good, just a couple things to point out/question

mhmerrill (approved):
looks ok to me.

Issue #1426, opened 2022-05-20
akVersion allocation is 1 too small


bmcdonald3:
In the `addArkoudaHdf5VersioningMetadata` function, there is a buffer being allocated for `akVersion` that is 1 too small, causing a bad memory access.

PR #1427, opened 2022-05-20
Fix allocated buffer size for akVersion

bmcdonald3:
When calculating the Arkouda version when writing an HDF5 file,
the buffer size was 1 too low, which meant there was a bad memory
access going when setting the last value to 0 to ensure null
termination.

Closes https://github.com/Bears-R-Us/arkouda/issues/1426


Ethan-DeBandi99 (approved):
Looks good

pierce314159 (approved):

PR #1438, opened 2022-05-23
Adjust source files to be compatible with changes I made to the Math module

lydia-duncan:
The Chapel team is in the process of splitting the Math standard library into a
portion that continues to be included by default and one that will require a
`use` or `import` statement going forward.  To maintain compatibility with both
the previous releases and main/the upcoming release, update these files to no
longer use explicit naming to access the symbols in that module (since the ones
that are being included by default are no longer in the module named "Math" but
are instead in a new module named "AutoMath" that won't exist for previous
releases)

While here, I also updated test/TestBase.chpl to fix the issue test/UnitTestSort.chpl
was encountering with latest Chapel main.  That test was complaining it could not
find AryUtil, because the 1.27 pre-release has changed our handling of `public use`
to no longer bring in the name of the module in addition to the symbols it contains,
unless an `as` clause is used.  Add an `as` clause so that the name is now
brought in again.  (see https://github.com/chapel-lang/chapel/pull/19306#issuecomment-1122425356
for how the scoping rules are expected to change in the upcoming 1.27 release)

### DIFF
--- a/src/Cast.chpl
+++ b/src/Cast.chpl
@@ -94,7 +94,7 @@ module Cast {
       num = stringToNumericStrict(values, rng, toType);
     } catch {
       if toType == real {
-        num = Math.NAN;
+        num = NAN;
       } else if toType == int {
         // Use pandas.NaT, i.e. -2**63, as NaN for int
         num = min(int);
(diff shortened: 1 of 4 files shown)


lydia-duncan:
(Draft so I can get CI testing started while I run testing locally with main, which seems to be going well so far)

lydia-duncan:
@mhmerrill or @reuster986 - would one of you be able to review this PR?

Ethan-DeBandi99 (approved):
Looks good

PR #1441, opened 2022-05-24
Closes #1363: ak.Series with only values and more robust typechecking

### LINKED ISSUE #1363: `ak.Series` with only values and more robust typechecking
First seen in issue #1347 by @kellyjoy15

```python
In [5]: n = np.array([True,False,True,True])
In [6]: p = pd.Series(n)
In [7]: p.index
Out[7]: RangeIndex(start=0, stop=4, step=1)

In [8]: p.values
Out[8]: array([ True, False,  True,  True])

In [9]: p
Out[9]:
0     True
1    False
2     True
3     True
dtype: bool

In [10]: a = ak.Series(n)
In [11]: a.index
Out[11]: True

In [12]: a.values
Out[12]: False

In [13]: a
Out[13]: 
AttributeError: 'numpy.bool_' object has no attribute 'to_ndarray'
```

We differ from pandas here. This is because series expects a tuple of `(index_arr, values_arr)` as input
 https://github.com/Bears-R-Us/arkouda/blob/67a5b542e28647bcc5d4b283b8f3e4bcfc006191/arkouda/series.py#L74-L77
So when `ak.Series(a)` is called it's expects `index` to be the first element and `values` to be the second element. So `index` is set to `True` and `values` is set to `False`

This is similar to the issue seen in #1267, and we need more robust type checking to error when the `Series` is initialized

In order to get the expected behavior we have to do this
```python
In [5]: a = ak.array([True,False,True,True])

In [6]: s = ak.Series((ak.arange(a.size), a))

In [7]: s.index
Out[7]: array([0 1 2 3])

In [8]: s.values
Out[8]: array([True False True True])

In [9]: s
Out[9]:
0     True
1    False
2     True
3     True
dtype: bool
```

We should update `ak.Series` so `ak.Series(values)` will behave like `ak.Series((ak.arange(values.size), values))` by default

jeichert60:
This PR (closes #1363):

- Deprecates the `ar_tuple` parameter inside of the `__init__` method.
- Adjusts the tests to use the other 2 parameters `(index, data)` when instantiating.

### DIFF
--- a/arkouda/series.py
+++ b/arkouda/series.py
@@ -1,8 +1,10 @@
+from typeguard import typechecked
+from typing import List, Optional, Tuple, Union, Iterable
 from arkouda.pdarrayclass import pdarray, argmaxk, attach_pdarray
 from arkouda.pdarraycreation import arange, array, zeros
 from arkouda.pdarraysetops import argsort, concatenate, in1d
 from arkouda.index import Index
-from arkouda.groupbyclass import GroupBy
+from arkouda.groupbyclass import GroupBy, groupable_element_type
 from arkouda.dtypes import int64, float64
 from arkouda.numeric import value_counts, cast as akcast
 from arkouda.util import get_callback
@@ -75,28 +77,55 @@ class Series:
 
     Parameters
     ----------
-    ar_tuple : 2-tuple of arkouda arrays with the first being the grouping key(s) and the
-             second being the value. The grouping key(s) will be the index of the series.
+    index : pdarray, Strings
+        an array of indices associated with the data array.
+        If empty, it will default to a range of ints whose size match the size of the data.
+        optional
+    data : Tuple, List, groupable_element_type
+        a 1D array. Must not be None if ar_tuple is not provided.
 
+    Raises
+    ------
+    TypeError
+        Raised if index is not a pdarray or Strings object
+        Raised if data is not a pdarray, Strings, or Categorical object
+    ValueError
+        Raised if the index size does not match data size
+
+    Notes
+    -----
+    The Series class accepts either positional arguments or keyword arguments.
+    If entering positional arguments,
+        2 arguments entered:
+            argument 1 - index
+            argument 2 - data
+        1 argument entered:
+            argument 1 - data
+    If entering 1 positional argument, it is assumed that this is the data argument.
+    If entering keywords,
+        'data' (see Parameters)
+        'index' (optional) must match size of 'data'
     """
 
-    def __init__(self, ar_tuple=None, data=None, index=None):
-        if ar_tuple is not None:
-            self.index = Index.factory(ar_tuple[0])
-            self.values = ar_tuple[1]
-        elif data is None:
-            raise TypeError("ar_tuple and data cannot both be null")
-
+    @typechecked
+    def __init__(self, data: Union[Tuple, List, groupable_element_type],
+                 index: Optional[Union[pdarray, Strings]] = None):
+        # TODO: Allow index to be an Index when index.py is updated
+        if isinstance(data, (tuple, list)) and len(data) == 2:
+            # handles the previous `ar_tuple` case
+            if not isinstance(data[0], (pdarray, Strings)):
+                raise TypeError("indices must be a pdarray or Strings")
+            if not isinstance(data[1], (pdarray, Strings, Categorical)):
+                raise TypeError("values must be a pdarray, Strings, or Categorical")
+            self.values = data[1]
+            self.index = Index.factory(data[0])
         else:
-            if not isinstance(data, (pdarray, Strings, Categorical)):
-                data = array(data)
-            self.values = data
+            # When only 1 positional argument it will be treated as data and not index
+            self.values = array(data) if not isinstance(data, (Strings, Categorical)) else data
+            self.index = Index.factory(index) if index is not None else Index(arange(self.values.size))
 
-            if index is None:
-                index = arange(data.size)
-            self.index = Index.factory(index)
         if self.index.size != self.values.size:
-            raise ValueError("Index and data must have same length")
+            raise ValueError("Index size does not match data size")
         self.size = self.index.size
 
     def __len__(self):
@@ -196,7 +225,7 @@ class Series:
         else:
             # scalar value
             idx = self.index == key
-        return Series((self.index[idx], self.values[idx]))
+        return Series(index=self.index.index[idx], data=self.values[idx])
 
     @classmethod
     def _make_binop(cls, operator):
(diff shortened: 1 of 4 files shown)


arkouda/series.py
```diff
@@ -75,28 +79,37 @@ class Series:
...
 
-    def __init__(self, ar_tuple=None, data=None, index=None):
+    @typechecked
+    def __init__(self, ar_tuple: Union[tuple, list] = None, index: pdarray = None,
+                 data: groupable_element_type = None):
+        # TODO: Allow index to be an Index when index.py is updated
         if ar_tuple is not None:
             self.index = Index.factory(ar_tuple[0])
             self.values = ar_tuple[1]
+            warnings.warn('ar_tuple will be deprecated in future releases', DeprecationWarning)
```
joshmarshall1, line 101: We might want to add to this warning message how to pass it with the new method so the user doesn't have to guess or look into the code

joshmarshall1 (changes requested):
Overall looks good, one small thing in the tail method and then a recommendation

pierce314159 (approved):
Couple minor suggestions, looks good overall!

arkouda/series.py
```diff
@@ -233,7 +246,12 @@ def add(self, b):
         index = self.index.concat(b.index).index
 
         values = concatenate([self.values, b.values], ordered=False)
-        return Series(GroupBy(index).sum(values))
+
+        vals = GroupBy(index).sum(values)
+        idx = vals[0]
+        val = vals[1]
+        return Series(data=val, index=idx)
+        # return Series(GroupBy(index).sum(values))
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I agree with @pierce314159 here, `idx, vals = ` just works better for my brain.

Ethan-DeBandi99 (changes requested):
Josh and Pierce touched on most things already. After discussing with @pierce314159 yesterday, I believe we can actually remove the `ar_tuple` parameter, but allow for `data` to handle a `Tuple` and `List` as well. This effectively will not change the api because it looks like most places ar_tuple is being used, it is being done positionally. We could then add an error using `**kwargs` to detect if `ar_tuple` is passed as a keyword. I think this eliminates some technical debt moving forward and removes the need for the deprecation warning. Alternatively, if we want to leave the deprecation warning, but still allow `ar_tuple` to be usable for a release cycle, I think that would be fine as well. This would also remove the need to update a lot of the usage for the `Series.__init__` to utilize the new parameter format.

joshmarshall1 (approved):

arkouda/series.py
```diff
@@ -75,28 +77,38 @@ class Series:
...
-             second being the value. The grouping key(s) will be the index of the series.
+    ar_tuple : tuple, list.
+        2-tuple of arkouda arrays with the first being the grouping key(s) and the
+        second being the value. The grouping key(s) will be the index of the series.
+        This will be deprecated in future releases.
+    index : pdarray
+        an array of indices associated with the data array.
+        If empty, it will default to a range of ints whose size match the size of the data.
+    data : pdarray, Strings, Categorical
+        a 1D array. Must not be None if ar_tuple is not provided.
```
pierce314159, line 88: should we remove `ar_tuple` from the doc string since we removed it as a named parameter?

Also we should update the accepted types of data and index in docstring to match the `__init__`

arkouda/series.py
```diff
@@ -75,28 +77,38 @@ class Series:
...
-            raise TypeError("ar_tuple and data cannot both be null")
-
+    @typechecked
+    def __init__(self, data: Union[Tuple, List, groupable_element_type],
+                 index: Optional[Union[pdarray, Strings]] = None):
+        # TODO: Allow index to be an Index when index.py is updated
+        if isinstance(data, (tuple, list)) and len(data) == 2:
+            # handles the previous `ar_tuple` case
+            if not isinstance(data[0], (pdarray, Strings)):
+                raise TypeError("indices must be a pdarray, iterable, or Strings")
```
pierce314159, line 99: remove `iterable` from this TypeError message

arkouda/series.py
```diff
@@ -75,28 +77,38 @@ class Series:
...
+            if not isinstance(data[1], (pdarray, Strings, Categorical)):
+                raise TypeError("values must be a pdarray, Strings, or Categorical")
+            # i = array(data[0]) if isinstance(data[0], (tuple, list)) else data[0]
```
pierce314159, line 102: remove this commented out line

arkouda/series.py
```diff
@@ -233,7 +245,10 @@ def add(self, b):
         index = self.index.concat(b.index).index
 
         values = concatenate([self.values, b.values], ordered=False)
-        return Series(GroupBy(index).sum(values))
+
+        idx, vals = GroupBy(index).sum(values)
+        return Series(data=vals, index=idx)
+        # return Series(GroupBy(index).sum(values))
```
pierce314159, line 251: remove this commented out line

tests/series_test.py
```diff
@@ -7,21 +7,52 @@
 class SeriesTest(ArkoudaTest):
 
     def test_series_creation(self):
-        ar_tuple = (ak.arange(3), ak.array(['A', 'B', 'C']))
-        s = ak.Series(ar_tuple=ar_tuple)
+        # Use positional arguments
+        ar_tuple = ak.array(ak.arange(3)), ak.array(ak.arange(3))
```
pierce314159, line 11: `ak.arange` returns a pdarray, so you shouldn't need to wrap it in an `ak.array`

I think this line can be changed to 
```python
ar_tuple = ak.arange(3), ak.arange(3)
```

pierce314159 (commented):
few small things

arkouda/series.py
jeichert60: Done.  Beefed up doc string.

arkouda/series.py
```diff
@@ -75,28 +77,55 @@ class Series:
...
+    ValueError
+        Raised if the index size does not match data size
+
+    Notes
+    -----
+    The Series class accepts either positional arguments or keyword arguments.
+    If entering positional arguments,
+        2 arguments entered:
+            argument 1 - index
+            argument 2 - data
```
pierce314159, line 101: I think these got swapped at some point. It looks to me like `data` should be first now

arkouda/series.py
```diff
@@ -75,28 +77,55 @@ class Series:
...
     ----------
-    ar_tuple : 2-tuple of arkouda arrays with the first being the grouping key(s) and the
-             second being the value. The grouping key(s) will be the index of the series.
-
+    index : pdarray, Strings
+        an array of indices associated with the data array.
+        If empty, it will default to a range of ints whose size match the size of the data.
+        optional
+    data : Tuple, List, groupable_element_type
+        a 1D array. Must not be None if ar_tuple is not provided.
```
pierce314159, line 85: we should prob remove the reference to `ar_tuple` here

pierce314159 (commented):
few things, mostly documentation

arkouda/series.py
pierce314159: resolving these, they will be covered in the reformatting

PR #1443, opened 2022-05-24
Closes #1382: Adding `Series` support to generic attach

### LINKED ISSUE #1382: Add `Series` support to generic attach
As another register-able object, `Series` should be supported by our generic attach functionality. This will also include having to create a `build_from_message` function within the `Series` class

joshmarshall1:
This PR (closes #1382):

This is waiting on PR #1441 to be merged due to file conflicts. Once merged, I'll update this PR to work with @jeichert60's changes.

This PR adds Chapel logic for determining if a given name represents a `Series` as well as adding logic to the symbol table that allows for a regex string to be passed and returns a list of all matching names contained in the table. This method, `findAll`, serves the purpose of finding all `<name>_key_<n>` names that that are associated with a `MultiIndex`. This method also opens the door for `GroupBy` support to be added in the future.

On the Python side, I added the `from_return_msg` method to `Series` which will create a `Series` object based on the returned message string from the generic attach method

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -6,7 +6,7 @@ module MultiTypeSymbolTable
     use Reflection;
     use ServerErrors;
     use Logging;
-    use Regex;
+    use ArkoudaRegexCompat;
     
     use MultiTypeSymEntry;
     use Map;
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -484,13 +484,12 @@ module MultiTypeSymbolTable
         proc findAll(pattern: string): [] string throws {
             var regex = compile(pattern);
             var infoStr = "";
-            for name in tab {
+            forall name in tab.keysToArray() with (+ reduce infoStr) {
                 if name.match(regex) {
                     infoStr += name + "+";
                 }
             }
-            infoStr = infoStr.strip("+");
-            return infoStr.split("+");
+            return infoStr.strip("+").split("+");
         }
     }
 
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -6,6 +6,7 @@ module MultiTypeSymbolTable
     use Reflection;
     use ServerErrors;
     use Logging;
+    use Regex;
     
     use MultiTypeSymEntry;
     use Map;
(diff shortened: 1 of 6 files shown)


src/MultiTypeSymbolTable.chpl
```diff
@@ -470,6 +471,27 @@ module MultiTypeSymbolTable
...
+            var regex = compile(pattern);
+            var infoStr = "";
+            for name in tab {
+                if name.match(regex) {
+                    infoStr += name + "+";
+                }
+            }
+            infoStr = infoStr.strip("+");
+            return infoStr.split("+");
+        }
```
pierce314159, line 494: using a [Reduce Intent](https://chapel-lang.org/docs/technotes/reduceIntents.html#reduce-intents), you can implement this in parallel (provided you don't care about order)
```Chapel
proc findAll(pattern: string): [] string throws {
    var regex = compile(pattern);
    var infoStr = "";
    forall name in tab with (+ reduce infoStr) {
        if name.match(regex) {
            infoStr += name + "+";
        }
    }
    return infoStr.strip("+").split("+");
}
```

pierce314159: I'm curious if @ronawho has any suggestions here

src/MultiTypeSymbolTable.chpl
```diff
@@ -6,6 +6,7 @@ module MultiTypeSymbolTable
     use Reflection;
     use ServerErrors;
     use Logging;
+    use Regex;
```
pierce314159, line 9: So the `Regex` module changed names between chapel version 1.24 (`Regexp`) and 1.25 (`Regex`). If you change this to `use ArkoudaRegexCompat;` it should do the right import depending on the version

src/MultiTypeSymbolTable.chpl
ronawho: I suspect it doesn't really matter here, and I'd go with whatever code is cleanest. My guess (though I don't really know) is that there's probably not more than a couple dozen or hundred sym entries at any given time.

If you have "enough" entries a parallel reduce will definitely be faster, but for a small reduce the array creation and task creation overheads will outweigh any parallelism benefits. That said, the overhead of those is small so even if things are slower, it's imperceptibly slower to the Arkouda user waiting so I maybe I'd go with the parallel approach just in case there are ever a lot of entries.

And note that this doesn't get more efficient when running on multiple nodes. `keysToArray` is returning a local non-distributed array and operations on it will still be local. (unless you have like a million or more entries, I don't think you want to distribute this so this is the correct behavior)

pierce314159: Awesome! Thanks ronawho

src/MultiTypeSymbolTable.chpl
joshmarshall1: Updated

pierce314159 (approved):

src/MultiTypeSymbolTable.chpl
ronawho: I do not have a ticket against Arkouda for this (just an internal one for our sprint plans.) Feel free to create one for it

Ethan-DeBandi99 (approved):

mhmerrill:
@joshmarshall1 fix conflicts please.

PR #1444, opened 2022-05-24
Optimize domain_intersection() when both domains are non-strided

bradcray:
As currently written, domain_intersection() will always return a
'stridable' domain because its return expression is of the form
'low..high by stride'.  In Chapel, stridable domains can result in a
performance hit relative to non-stridable ones since they require more
generalized math (since the stride is not known statically).  For
example, an array '[1..n by s] uint(8)' requires a division by 's' in
order to calculate the offset from the array's base pointer.

Here, I'm optimizing the return path so that if neither of the input
domains are stridable, the output domain won't be either.

Note that, for its current uses, I don't expect this to have much of
an impact on the execution time.  In part because (so far as I could
tell), the resulting domain didn't seem to be used to declare arrays
(which is where the highest overheads are), and since the calls are
largely in I/O code, it's likely that the I/O overheads will overwhelm
any benefit from this change.

But, it could still be beneficial in case further uses of the utility
are made in the future.

Finally, note that if this is a simple domain intersection, indexing
one domain by another in Chapel computes a domain intersection.  Thus,
this could be replaced with ``d1[d2]``.  That's a bigger change than I
wanted to take on here, though, so I just took the baby step that
seemed obvious to me.

### DIFF
--- a/src/FileIO.chpl
+++ b/src/FileIO.chpl
@@ -179,8 +179,12 @@ module FileIO {
             //TODO: change this to throw
             halt("At least one domain must have stride 1");
         }
-        var stride = max(d1.stride, d2.stride);
-        return {low..high by stride};
+        if !d1.stridable && !d2.stridable {
+            return {low..high};
+        } else {
+            var stride = max(d1.stride, d2.stride);
+            return {low..high by stride};
+        }
     }
 
     proc getFirstEightBytesFromFile(path:string):bytes throws {


bradcray:
@glitch : For your consideration.  This is not a big deal, just something I stumbled across while making other changes.

bradcray:
(also tagging @bmcdonald3 just to keep him in the loop because we were discussing this).

bradcray:
@glitch: Sorry, I have not done a good job of keeping my mental map of GitHub usernames -> real peoples' names straight, and Ben just reminded me that you're not developing Arkouda anymore.  Please feel free to ignore this and the other comment I just tagged you on.

Tagging @reuster986 on this for his general information since Ben says that he's the main person with whom he's been coordinating on I/O changes.  Bill, for reference, the other comment I was referring to is here:  https://github.com/Bears-R-Us/arkouda/pull/1007/files#r881055259 (where the goal was to make sure that some of MIke's early habits didn't rub off on people, if not to clean up the code itself a bit).

Ethan-DeBandi99 (approved):
Based on your explanation, this looks good to me

glitch:
@bradcray Sorry folks, I wasn't able to get into my github account the other day, but for what it's worth, I approve!

Issue #1338, opened 2022-04-28
refactor `RadixSortLSD` into a class so multiple sorts can be instantiated at the same time


mhmerrill:
refactor `RadixSortLSD` into a class so multiple sorts can be instantiated at the same time. This will enable us to to support multiple sort configurations at the same time.

PR #1447, opened 2022-05-25
Preliminary implementation of `RadixSortLSDClass`

pierce314159:
First steps towards #1338

This is a draft PR as i want to get initial impressions/feedback on this implementation of refactoring Radix Sort as a class before I actually change `RadixSortLSD.chpl`

The way these are instantiated in chapel are
```Chapel
// default params are numTasks = here.maxTaskPar, bitsPerDigit = RSLSD_bitsPerDigit
var rs = new RadixSortLSDInstance();
var rs2 = new RadixSortLSDInstance(numTasks, bitsPerDigit);
rs.radixSortLSD_keys(a);
rs2.radixSortLSD_keys(a);
```

Currently only `ak.sort` uses `RadixSortLSDInstance` as my proof of concept
```python
>>> a = ak.randint(1, 15, 10)
>>> a
array([13 11 4 5 6 11 4 8 8 6])

>>> ak.sort(a)
array([4 4 5 6 6 8 8 11 11 13])

>>> ak.sort(a, bits_per_digit=4)
array([4 4 5 6 6 8 8 11 11 13])

>>> ak.sort(a, num_tasks=22, bits_per_digit=4)
array([4 4 5 6 6 8 8 11 11 13])
```

### DIFF
--- a/src/RadixSortLSDClass.chpl
+++ b/src/RadixSortLSDClass.chpl
@@ -209,14 +209,12 @@ module RadixSortLSDClass {
 
         proc radixSortLSD_memEst(size: int, itemsize: int) {
             // 2 temp key+ranks arrays + globalStarts/globalClounts
-            var numBuckets = 1 << bitsPerDigit;
             return (2 * size * (itemsize + numBytes(int))) +
                 (2 * numLocales * numTasks * numBuckets * numBytes(int));
         }
 
         proc radixSortLSD_keys_memEst(size: int, itemsize: int) {
             // 2 temp key arrays + globalStarts/globalClounts
-            var numBuckets = 1 << bitsPerDigit;
             return (2 * size * itemsize) +
                 (2 * numLocales * numTasks * numBuckets * numBytes(int));
         }
--- a/src/RadixSortLSDClass.chpl
+++ b/src/RadixSortLSDClass.chpl
@@ -13,21 +13,6 @@ module RadixSortLSDClass {
     private config const logLevel = ServerConfig.logLevel;
     const rsLogger = new Logger(logLevel);
 
-    proc radixSortLSD_memEst(size: int, itemsize: int, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit) {
-        // 2 temp key+ranks arrays + globalStarts/globalClounts
-        var numBuckets = 1 << bitsPerDigit;
-        return (2 * size * (itemsize + numBytes(int))) +
-            (2 * numLocales * numTasks * numBuckets * numBytes(int));
-    }
-
-    proc radixSortLSD_keys_memEst(size: int, itemsize: int, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit) {
-        // 2 temp key arrays + globalStarts/globalClounts
-        var numBuckets = 1 << bitsPerDigit;
-        return (2 * size * itemsize) +
-            (2 * numLocales * numTasks * numBuckets * numBytes(int));
-
-    }
-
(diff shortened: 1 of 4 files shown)


src/RadixSortLSDClass.chpl
```diff
@@ -0,0 +1,225 @@
...
+    use IO;
+    use CTypes;
+    use Reflection;
+    use Logging;
+    use ServerConfig;
+
+    private config const logLevel = ServerConfig.logLevel;
+    const rsLogger = new Logger(logLevel);
+
+    proc radixSortLSD_memEst(size: int, itemsize: int, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit) {
```
mhmerrill, line 16: move memEst routines into the class and make them use the class consts for tasks, etc.

src/RadixSortLSDClass.chpl
```diff
@@ -0,0 +1,225 @@
...
+                                aggregator.flush();
+                            }
+                        }//coforall task
+                    }//on loc
+                }//coforall loc
+
+                // copy back to temp for next iteration
+                // Only do this if there are more digits left
+                if !last {
+                temp = a;
```
mhmerrill, line 177: we just accepted a PR to use the swap operator `temp <=> a`, need to reflect the change here.

mhmerrill:
- move some params out of `ArrayUtil.chpl` into the support module , maybe call it `RadixSortLSD_support.chpl` or a better name if you think of one.
https://github.com/Bears-R-Us/arkouda/blob/ba545acc5c7f1c112f1ee412f03f2add074821e5/src/AryUtil.chpl#L13-L15
- move the support routines (`getBitWidth`,`getDigit`) out of `ArrayUtil.chpl` into the support module
https://github.com/Bears-R-Us/arkouda/blob/ba545acc5c7f1c112f1ee412f03f2add074821e5/src/AryUtil.chpl#L206-L298

bmcdonald3 (commented):
There are a couple of places where the spacing seems off in the Chapel code, but I think the concept makes sense. I can run some performance experiments on this a bit later today to see how it compares to the radix sort standalone function as I think there may be some performance implications of using a class, but I don't think I have a strong enough grasp on what that is to explain.

src/RadixSortLSDClass.chpl
pierce314159: ah yeah I wasn't really thinking here. In my head, i was thinking we'd want to run the mem estimates before creating a `RadixSortLSDInstance`, but the memory overhead is in running the sort, not creating an instance of the class. I'll get those moved!

bmcdonald3:
Had a chance to run some performance tests and, unfortunately, the performance of this sort relative to the standalone function appears to be abysmal. This is a problem with Chapel, not your code by the way Pierce! 

Here are some preliminary numbers on 16-node-cs-hdr:
```
./test-bin/UnitTestSort -nl 1 --elemsPerLocale=$((10))
  radixSortLSD_keys(A:[] int(64)) -- (rand uint(32) vals) -- 0.00 MiB/s per node (0.02s)
 radixSortLSD_ranks(A:[] int(64)) -- (rand uint(32) vals) -- 0.00 MiB/s per node (0.02s)
class radixSortLSD_ranks(A:[] int(64)) -- (rand uint(32) vals) -- 0.00 MiB/s per node (5.09s)
```

You will notice that this is ran single locale with 10 elements per locale, and that is because the execution time is so long for the class radix sort that I wasn't able to wait around for them to finish, but for this extremely small problem size, you see about a 260x performance hit with the class.

I am keep looking into this to see where all the time is going, but just noting that there is going to need to be some work to get a class implementation of sort to be on par with a standalone function.

Here is the branch that I was using to test this: https://github.com/bmcdonald3/arkouda/tree/radix-class

pierce314159:
Closing this PR for reasons above, I'll put up a new one with the implementation changes discussed

PR #1451, opened 2022-05-27
Closes #1450 - Import/Export Delete Folder on Startup

### LINKED ISSUE #1450: `import`/`export` tests update to delete folder on startup
If the `import`/`export` tests fail, the folder used for testing is not deleted. We should add a check to delete if the folder exists at the beginning of the run.

Ethan-DeBandi99:
Closes #1450:

Updated the tests to check if the directory exists or not before running `os.mkdir`.


pierce314159 (approved):
Looks good!

mhmerrill (approved):
looks good

Issue #1395, opened 2022-05-17
`black` auto formatter


pierce314159:
Recent discussions seem to be in favor of using the [Black](https://github.com/psf/black) auto formatting tool for our development team

Black follows all the [PEP-8](https://www.python.org/dev/peps/pep-0008/) rules with some extra for reasons like determinism and smaller diffs. Some benefits of using black:

- Code is formatted in a way to minimize diffs, making code review easier because diffs will be functionality changes instead of formatting
- Abstract Syntax Trees are used to verify the formatted code is effectively equivalent to the original
- Widely adopted:
  - Used by many open-source projects including pytest and pandas
  - Used by organizations including Tesla, Facebook, Dropbox, Mozilla, Quora

There's also really nice [interoperability with pycharm](https://black.readthedocs.io/en/stable/integrations/editors.html#pycharm-intellij-idea) (which a fair percentage of  the dev team uses)
______

I think the way forward is:
- the team installs `black`
- we put up a single PR per python file containing only the black formatting changes for that file
- once the entire python code base is formatted, we'll add a linter such as `flake8` to the CI to verify new code meets our standards
- (optionally) [add pre commit git hooks](https://ljvmiranda921.github.io/notebook/2018/06/21/precommits-using-black-and-flake8/) for black and flake8 

other popular code quality tools to consider:
- [isort](https://github.com/PyCQA/isort) (import sorting & formatting)
- [Pylint](https://github.com/PyCQA/pylint) (code quality)
- [coverage.py](https://github.com/nedbat/coveragepy) (test coverage)

Related to #837

pierce314159:
> @pierce314159 did we address the final thing on the checklist yet?

We did! I'm gonna go ahead and close the issue!

PR #1473, opened 2022-06-03
Closes #1471 - Update Install Doc

### LINKED ISSUE #1471: Update Install Doc
The documentation needs to be updated to include `make install-arrow` now that we require Parquet.

Ethan-DeBandi99:
Closes #1471 

- Updates the `README.md` to include Arrow in the `make install-deps` section. Adds information to individually install dependencies as well.
- Updates configuration files (arkouda-env.yml, arkouda-env-dev.yml, and setup.py) to include the dependencies for Arrow that need to be installed before `make install-arrow` will successfully complete. In the event that the `Makefile` fix does not work properly.
- Updates `Makefile` to add `-DARROW_DEPENDENCY_SOURCE=AUTO`. This triggers the usage of Chapel deps instead of requiring them to be installed via a package manager.
- Adds `defaluts` channel to conda env yamls because some Arrow dependencies need that to install properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,9 +173,7 @@ This step only needs to be done once. Once dependencies are installed, you will 
 - Arrow
 
 #### Arrow Dependencies
-The arrow install will fail unless supporting packages are installed. Please run the install below based on your package manager
-
-#### Arrow Dependencies
+The arrow install will fail unless supporting packages are installed. Please run the install below based on your package manager.
 
 ```
 #using conda to install
--- a/README.md
+++ b/README.md
@@ -163,8 +163,48 @@ sh /opt/cmake-$CM_VERSION-Linux-x86_64.sh --skip-license --include-subdir
 export PATH=./cmake-$CM_VERSION-Linux-x86_64/bin:$PATH
 ```
 
-# If zmq and hdf5 have not been installed previously, execute make install-deps
-make install-deps
+### Install Dependencies
+This step only needs to be done once. Once dependencies are installed, you will not need to run again. You can installl all dependencies with a single command or install individually for a customized build.
+
+#### Dependencies
+
+- ZMQ
+- HDF5
+- Arrow
+
+#### Arrow Dependencies
+The arrow install will fail unless supporting packages are installed. Please run the install below based on your package manager
+
+##### All Dependencies 
+
+`make install-deps`
+
+#### Individual Installs
+
+```
+# Install ZMQ Only
+make install-zmq
+
+# Install HDF5 Only
+# Install HDF5 Only
+make install-hdf5
+
+# Install Arrow Only
+make install-arrow
+```
+
+#### Trouble Installing Arrow?
+
+If arrow is not installing, trying installing the dependencies below. Use the command suited to the package manager that you are using.
+
+```
+#using conda to install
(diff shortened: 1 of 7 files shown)


Ethan-DeBandi99:
Marking as draft until I can identify why CI fails when dependencies for arrow added to setup.py to install via pip.

Ethan-DeBandi99:
@ronawho helped me get everything straightened out. Conda is automatically used for dependencies by cmake when running in an active conda environment. Adding `-DARROW_DEPENDENCY_SOURCE=AUTO` to the make command for `make install--arrow` will trigger it to use the Chapel deps.

README.md
```diff
@@ -163,8 +163,45 @@ sh /opt/cmake-$CM_VERSION-Linux-x86_64.sh --skip-license --include-subdir
...
+`make install-deps`
+
+#### Individual Installs
+
+```
+# Install ZMQ Only
+make install-zmq
+
+# Install HDF5 Only
+# Install HDF5 Only
```
bmcdonald3, line 186: looks like this line is doubled here

README.md
```diff
@@ -214,8 +251,6 @@ are pushed to the Arkouda or Arkouda fork _master branch_. Next, navigate to the
 option. The Github Pages docs url will be displayed once the source option is selected. Click on the link and the
 Arkouda documentation homepage will be displayed.
 
-</details>
```
bmcdonald3, line 217: Was this deleted intentionally? It looks like we just have an open details tag there now.

README.md
Ethan-DeBandi99: Removed the dup

README.md
Ethan-DeBandi99: It was, but it was the wrong fix. Found where a comment block was improperly configured and fixed.

arkouda-env-dev.yml
```diff
@@ -1,20 +1,20 @@
 name: arkouda-env-dev
 channels:
   - conda-forge
+  - defaults
```
pierce314159, line 4: i had more luck when testing this locally with the order of the channels swapped

arkouda-env-dev.yml
```diff
@@ -1,20 +1,20 @@
...
   - versioneer
   - matplotlib
   - h5py
   - pip
   - types-tabulate
-  - tables
+  - pytables
   - pyarrow
 
   # Developer dependencies
```
pierce314159, line 20: do we want to add any other developer dependencies? Ones that come to mind are `jupyter` (which will bring along ipython) and the formatting ones (`black`, `flake8`, `isort`)

pierce314159 (approved):
This all looks good to me!

arkouda-env-dev.yml
Ethan-DeBandi99: Mine works without issue with them in this order. The order should not matter

arkouda-env-dev.yml
Ethan-DeBandi99: `Jupyter` I would say no because Arkouda does not require. The formatting stuff could be added. However, it looks like you have an issue in for documenting the formatting (#1395). I would add it with the documentation. This modification was mainly to resolve issues related to the parquet dependency.

bmcdonald3 (approved):
Looks good to me

PR #1477, opened 2022-06-06
Closes #1465: Add typechecking to `Series`

### LINKED ISSUE #1465: Typechecking in `series.py`
Add Typechecking to `series.py`
related to #1373

jeichert60:
This PR (Closes #1465):
- Adds typechecking to the methods in `Series`

### DIFF
--- a/arkouda/series.py
+++ b/arkouda/series.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import List, Optional, Tuple, Union
 from warnings import warn
 
@@ -5,6 +7,8 @@ import numpy as np  # type: ignore
 import pandas as pd  # type: ignore
 from pandas._config import get_option  # type: ignore
 from typeguard import typechecked
+
+import arkouda.dataframe
 from arkouda.accessor import CachedAccessor, DatetimeAccessor, StringAccessor
 from arkouda.alignment import lookup
 from arkouda.categorical import Categorical
@@ -13,7 +17,7 @@ from arkouda.groupbyclass import GroupBy, groupable_element_type
 from arkouda.index import Index
 from arkouda.numeric import cast as akcast
 from arkouda.numeric import value_counts
-from arkouda.pdarrayclass import argmaxk, attach_pdarray, pdarray, create_pdarray
+from arkouda.pdarrayclass import argmaxk, attach_pdarray, create_pdarray, pdarray
 from arkouda.pdarraycreation import arange, array, zeros
 from arkouda.pdarraysetops import argsort, concatenate, in1d
 from arkouda.strings import Strings
@@ -102,6 +106,7 @@ class Series:
         1 argument entered:
             argument 1 - data
     If entering 1 positional argument, it is assumed that this is the data argument.
+    If only 'data' argument is passed in, Index will automatically be generated.
     If entering keywords,
         'data' (see Parameters)
         'index' (optional) must match size of 'data'
@@ -149,33 +154,30 @@ class Series:
             prt = pd.concat(
                 [self.head(maxrows // 2 + 2).to_pandas(), self.tail(maxrows // 2).to_pandas()]
             )
-            length_str = "\nLength {len(self)}"
+            length_str = f"\nLength {len(self)}"
         return (
             prt.to_string(
-                dtype=prt.dtype,
-                min_rows=get_option("display.min_rows"),
-                max_rows=maxrows,
-                length=False,
+                dtype=prt.dtype, min_rows=get_option("display.min_rows"), max_rows=maxrows, length=False,
             )
             + length_str
         )
 
     def __getitem__(self, key):
-        if type(key) == Series:
+        if isinstance(key, Series):
             # @TODO align the series indexes
             key = key.values
         return Series((self.index[key], self.values[key]))
 
     dt = CachedAccessor("dt", DatetimeAccessor)
-    str = CachedAccessor("str", StringAccessor)
-    # cat = CachedAccessor("cat", CategoricalAccessor)
+    str_acc = CachedAccessor("str", StringAccessor)
 
     @property
     def shape(self):
         # mimic the pandas return of series shape property
         return (self.values.size,)
 
-    def isin(self, lst):
+    @typechecked
+    def isin(self, lst: Union[pdarray, Strings, List]) -> Series:
         """Find series elements whose values are in the specified list
 
         Input
@@ -192,31 +194,26 @@ class Series:
         boolean = in1d(self.values, lst)
         return Series(data=boolean, index=self.index)
 
-    def locate(self, key):
+    @typechecked
+    def locate(self, key: Union[int, pdarray, Series, List, Tuple]) -> Series:
         """Lookup values by index label
 
-        The input can be a scalar, a list of scalers, or a list of lists (if the series has
-        a MultiIndex).
-        As a special case, if a Series is used as the key, the series labels are preserved
-        with its values used as the key.
+        The input can be a scalar, a list of scalers, or a list of lists (if the series has a
+        MultiIndex). As a special case, if a Series is used as the key, the series labels are
+        preserved with its values use as the key.
 
         Keys will be turned into arkouda arrays as needed.
 
         Returns
         -------
-
         A Series containing the values corresponding to the key.
         """
-        t = type(key)
         if isinstance(key, Series):
             # special case, keep the index values of the Series, and lookup the values
-            labels = key.index
-            key = key.values
-            v = lookup(self.index.index, self.values, key)
-            return Series((labels, v))
+            return Series(index=key.index, data=lookup(self.index.index, self.values, key.values))
         elif isinstance(key, pdarray):
             idx = self.index.lookup(key)
-        elif t == list or t == tuple:
+        elif isinstance(key, (list, tuple)):
             key0 = key[0]
             if isinstance(key0, list) or isinstance(key0, tuple):
                 # nested list. check if already arkouda arrays
@@ -226,7 +223,13 @@ class Series:
 
             elif not isinstance(key0, pdarray):
                 # a list of scalers, convert into arkouda array
-                key = array(key)
+                try:
+                    val = array(key)
+                    if isinstance(val, pdarray):
+                        key = val
+                except Exception:
+                    raise TypeError("'key' parameter must be convertible to pdarray")
+
             # else already list if arkouda array, use as is
             idx = self.index.lookup(key)
         else:
@@ -237,7 +240,7 @@ class Series:
     @classmethod
     def _make_binop(cls, operator):
         def binop(self, other):
-            if type(other) == Series:
+            if isinstance(other, Series):
                 if self.index._check_aligned(other.index):
                     return cls((self.index, operator(self.values, other.values)))
                 else:
@@ -264,7 +267,8 @@ class Series:
 
         return aggop
 
-    def add(self, b):
+    @typechecked
+    def add(self, b: Series) -> Series:
 
         index = self.index.concat(b.index).index
 
@@ -273,7 +277,8 @@ class Series:
         idx, vals = GroupBy(index).sum(values)
         return Series(data=vals, index=idx)
 
-    def topn(self, n=10):
+    @typechecked
+    def topn(self, n: int = 10) -> Series:
         """Return the top values of the series
 
         Parameters
@@ -292,7 +297,8 @@ class Series:
 
         return Series(index=k.index[idx], data=v[idx])
 
-    def sort_index(self, ascending=True):
+    @typechecked
+    def sort_index(self, ascending: bool = True) -> Series:
         """Sort the series by its index
 
         Returns
@@ -303,7 +309,8 @@ class Series:
         idx = self.index.argsort(ascending=ascending)
         return Series(index=self.index.index[idx], data=self.values[idx])
 
-    def sort_values(self, ascending=True):
+    @typechecked
+    def sort_values(self, ascending: bool = True) -> Series:
         """Sort the series numerically
 
         Returns
@@ -316,33 +323,37 @@ class Series:
                 # For numeric values, negation reverses sort order
                 idx = argsort(-self.values)
             else:
-                # For non-numeric values, need the descending arange because
-                # reverse slicing not supported
+                # For non-numeric values, need the descending arange because reverse slicing
+                # is not supported
                 idx = argsort(self.values)[arange(self.values.size - 1, -1, -1)]
         else:
             idx = argsort(self.values)
         return Series(index=self.index.index[idx], data=self.values[idx])
 
-    def tail(self, n=10):
+    @typechecked
+    def tail(self, n: int = 10) -> Series:
         """Return the last n values of the series"""
 
         idx_series = self.index[-n:]
         return Series(index=idx_series.index, data=self.values[-n:])
 
-    def head(self, n=10):
+    @typechecked
+    def head(self, n: int = 10) -> Series:
         """Return the first n values of the series"""
 
         idx_series = self.index[0:n]
         return Series(index=idx_series.index, data=self.values[0:n])
 
-    def to_pandas(self):
+    @typechecked
+    def to_pandas(self) -> pd.Series:
         """Convert the series to a local PANDAS series"""
 
         idx = self.index.to_pandas()
         val = convert_if_categorical(self.values)
         return pd.Series(val.to_ndarray(), index=idx)
 
-    def value_counts(self, sort=True):
+    @typechecked
+    def value_counts(self, sort: bool = True) -> Series:
         """Return a Series containing counts of unique values.
 
         The resulting object will be in descending order so that the
@@ -350,7 +361,6 @@ class Series:
 
         Parameters
         ----------
-
         sort : Boolean. Whether or not to sort the results.  Default is true.
         """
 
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (changes requested):
I think you need to rerun `black` on the file. There are a few things within the diff here that are off. I checked out the PR and ran `black` and there were things that were not in the diff here.

You will also need to git rebase with master before we can merge this.

arkouda/series.py
```diff
@@ -228,7 +228,13 @@ def locate(self, key):
 
             elif not isinstance(key0, pdarray):
                 # a list of scalers, convert into arkouda array
-                key = array(key)
+                try:
+                    val = array(key)
+                    if isinstance(val, pdarray):
+                        key = val
+                except Exception:
+                    raise TypeError("Must be convertible to pdarray")
```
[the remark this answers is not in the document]
jeichert60: Inserted that val must be convertible.

arkouda/series.py
```diff
@@ -525,20 +540,26 @@ def concat(arrays, axis=0, index_labels=None, value_labels=None):
...
+            raise TypeError(f"Items must all have same type: {types}")
+
+        if isinstance(arrays[0], tuple):
             arrays = [Series(i) for i in arrays]
 
         if axis == 1:
             # Horizontal concat
             if value_labels is None:
-                value_labels = [f"val_{i}" for i in range(len(arrays))]
+                value_labels = ["val_{}".format(i) for i in range(len(arrays))]
```
pierce314159, line 553: Can we update this to use `f` string as well so
```python
value_labels = [f"val_{i}" for i in range(len(arrays))]
```

pierce314159 (commented):
Two comments about f strings, looks good other than that.

Also your PR description is empty, would you mind updating that? Try to use closing keywords to link with the issue. Something like
```
This PR (Closes #1465):
- Adds typechecking to the methods in `Series`
```
Would be great

arkouda/series.py
jeichert60: Done

arkouda/series.py
```diff
@@ -404,11 +418,12 @@ def register(self, label):
         """
 
         retval = self.index.register(label)
-        register(self.values, f"{label}_value")
+        register(self.values, "f{label}_value")
```
Ethan-DeBandi99, line 421: `f` need to move this outside the quotes.

arkouda/series.py
```diff
@@ -98,8 +98,8 @@ class Series:
     The Series class accepts either positional arguments or keyword arguments.
     If entering positional arguments,
         2 arguments entered:
-            argument 1 - data
-            argument 2 - index
+            argument 1 - index
+            argument 2 - data
         1 argument entered:
             argument 1 - data
     If entering 1 positional argument, it is assumed that this is the data argument.
```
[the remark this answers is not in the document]
jeichert60: Done

arkouda/series.py
```diff
@@ -404,11 +418,12 @@ def register(self, label):
         """
 
         retval = self.index.register(label)
-        register(self.values, f"{label}_value")
+        register(self.values, "{}_value".format(label))
```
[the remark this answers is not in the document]
jeichert60: Done

arkouda/series.py
```diff
@@ -102,9 +103,11 @@ class Series:
         1 argument entered:
             argument 1 - data
     If entering 1 positional argument, it is assumed that this is the data argument.
+    If only 'data' argument is passed in, Index will automatically be generated.
     If entering keywords,
         'data' (see Parameters)
         'index' (optional) must match size of 'data'
+    'ar_tuple' will no longer be provided. 'data' was updated to handle the tuples.
```
Ethan-DeBandi99, line 110: This line is not needed.

arkouda/series.py
```diff
@@ -110,9 +110,9 @@ class Series:
...
     @typechecked
     def __init__(
-        self,
-        data: Union[Tuple, List, groupable_element_type],
-        index: Optional[Union[pdarray, Strings]] = None,
+            self,
+            data: Union[Tuple, List, groupable_element_type],
+            index: Optional[Union[pdarray, Strings]] = None,
     ):
         # TODO: Allow index to be an Index when index.py is updated
```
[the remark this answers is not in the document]
Ethan-DeBandi99: We need to add an issue to handle this because #1482 has be merged.

arkouda/series.py
```diff
@@ -227,7 +227,13 @@ def locate(self, key):
 
             elif not isinstance(key0, pdarray):
                 # a list of scalers, convert into arkouda array
-                key = array(key)
+                try:
+                    val = array(key)
+                    if isinstance(val, pdarray):
+                        key = val
+                except Exception:
+                    raise TypeError(f"{val} Must be convertible to pdarray")
```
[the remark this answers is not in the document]
jeichert60: Updated to: f"'key' parameter must be convertible to pdarray"

arkouda/series.py
jeichert60: Removed

arkouda/series.py
```diff
@@ -227,7 +225,13 @@ def locate(self, key):
 
             elif not isinstance(key0, pdarray):
                 # a list of scalers, convert into arkouda array
-                key = array(key)
+                try:
+                    val = array(key)
+                    if isinstance(val, pdarray):
+                        key = val
+                except Exception:
+                    raise TypeError(f"'key' parameter must be convertible to pdarray")
```
Ethan-DeBandi99, line 233: You need to remove the `f` before the string. This is what is causing the CI to fail on `flake8`.

Ethan-DeBandi99 (changes requested):

arkouda/series.py
```diff
@@ -193,13 +192,13 @@ def isin(self, lst):
         boolean = in1d(self.values, lst)
         return Series(data=boolean, index=self.index)
 
-    def locate(self, key):
+    @typechecked
+    def locate(self, key: Union[int, pdarray, "Series", List, Tuple]):
```
pierce314159, line 196: here and in other places, we need the returns part of the typechecking
```python
def locate(self, key: Union[int, pdarray, Series, List, Tuple]) -> Series:
```

arkouda/series.py
```diff
@@ -5,6 +5,7 @@
 import pandas as pd  # type: ignore
 from pandas._config import get_option  # type: ignore
 from typeguard import typechecked
+
```
pierce314159, line 8: if we add `from __future__ import annotations` to our imports, we won't have to add the quotes around `"Series"` in the typechecking anymore and can just say `Series`

pierce314159 (changes requested):
A few things. Note most of these don't have in line comments since these refer to sections of the file that have not been changed

if we add `from __future__ import annotations` to our imports, we won't have to add the quotes around `"Series"` in the typechecking anymore and can just say `Series`

A lot of the methods need the return part of the type checking for example
```python
@typechecked
def sort_values(self, ascending: bool = True) -> Series:
```

`isin` seemed to be missed for typechecking. We also want to update the `Returns` section of the docstring to specify that a `Series` is returned

We should go ahead and find all the places we do explict type comparison and change it to use `isinstance`
```python
if type(key) == Series:
# becomes
if isinstance(key, Series)
```
and in the same vein we should update `pdconcat` to do the same thing we did in `concat`. To verify every element is the same type instead of assuming if the first element is a `Tuple` that every element is
```python
if type(arrays[0]) == tuple:
# becomes
types = set([type(x) for x in arrays])
if len(types) != 1:
    raise TypeError(f"Items must all have same type: {types}")

if isinstance(arrays[0], tuple):
```

arkouda/series.py
jeichert60: Done

arkouda/series.py
jeichert60: Done

arkouda/series.py
Ethan-DeBandi99: @pierce314159 This looks good to me now, but leaving for you to double check.

Ethan-DeBandi99 (changes requested):
This looks like Josh's mypy>=0.931 are showing up here. We need to update the commit history so that is not the case, which is why I am requesting changes. Other than that this looks good.

arkouda/series.py
```diff
@@ -193,31 +193,29 @@ def isin(self, lst):
         boolean = in1d(self.values, lst)
         return Series(data=boolean, index=self.index)
 
-    def locate(self, key):
+    @typechecked
+    def locate(self, key: Union[int, pdarray, Series, List, Tuple]) -> Series:
```
Ethan-DeBandi99, line 197: Should we be allowing an `Index` object as a key here?

arkouda/series.py
```diff
@@ -444,14 +463,14 @@ def is_registered(self):
...
-                f"but only located {sum(regParts)}"
+                f"Series expected {len(regParts)} components to be registered, but only located"
+                f" {sum(regParts)}"
             )
 
         return all(regParts)
 
     @staticmethod
-    def from_return_msg(repMsg):
+    def from_return_msg(repMsg) -> Series:
```
Ethan-DeBandi99, line 473: Should update to `repMsg: String`

Ethan-DeBandi99 (changes requested):
Few more things to review. Also, I noticed `pydoc/requirements.txt` and `setup.py` are in here, but probably should not be. The steps to resolve this should be pretty similar to what we did for PR #1507. Let myself or @pierce314159 know if you need help with that.

arkouda/series.py
```diff
@@ -1,10 +1,14 @@
...
+
 from typing import List, Optional, Tuple, Union
 from warnings import warn
 
 import numpy as np  # type: ignore
 import pandas as pd  # type: ignore
 from pandas._config import get_option  # type: ignore
 from typeguard import typechecked
+
+import arkouda.dataframe
```
pierce314159, line 11: if you change this to 
```python
from arkouda.dataframe import DataFrame
```
you won't need to do `arkouda.dataframe.DataFrame` and you'll be able to do just `DataFrame` instead

arkouda/series.py
```diff
@@ -111,7 +116,7 @@ class Series:
     def __init__(
         self,
         data: Union[Tuple, List, groupable_element_type],
-        index: Optional[Union[pdarray, Strings, Tuple, List, Index]] = None,
+        index: Optional[Union[pdarray, Strings, Tuple, List]] = None,
```
pierce314159, line 119: This is the line that was added in #1508 to enable `index` being an `Index`

I'm pretty sure what happened here is you had a merge conflict and you took your changes over the ones in `master`, but we def want the ones in `master` since this was part of a critical bug fix. So if you put this back, you won't need to touch `dataframe.py`

arkouda/series.py
```diff
@@ -149,33 +154,29 @@ def __repr__(self):
...
-    # cat = CachedAccessor("cat", CategoricalAccessor)
+    str_acc = CachedAccessor("str", StringAccessor)
 
     @property
     def shape(self):
         # mimic the pandas return of series shape property
         return (self.values.size,)
 
-    def isin(self, lst):
+    def isin(self, lst: Union[pdarray, Strings]) -> Series:
```
pierce314159, line 179: i think you missed the `@typechecked` above `isin` here

arkouda/series.py
```diff
@@ -192,31 +193,29 @@ def isin(self, lst):
...
         """
-        t = type(key)
         if isinstance(key, Series):
             # special case, keep the index values of the Series, and lookup the values
             labels = key.index
-            key = key.values
-            v = lookup(self.index.index, self.values, key)
+            val = key.values
+            v = lookup(self.index.index, self.values, val)
             return Series((labels, v))
```
pierce314159, line 215: we don't need this to be so many lines
```python
v = lookup(self.index.index, self.values, key.values)
return Series((key.index, v))
```
also i prefer the explicit approach of initializing a `Series` (not relying on tuple case), so perhaps a cleaner way to do this is
```python
return Series(index = key.index, data = lookup(self.index.index, self.values, key.values))
```

arkouda/series.py
```diff
@@ -316,41 +325,43 @@ def sort_values(self, ascending=True):
...
-    def head(self, n=10):
+    @typechecked
+    def head(self, n: int = 10) -> Series:
         """Return the first n values of the series"""
 
         idx_series = self.index[0:n]
         return Series(index=idx_series.index, data=self.values[0:n])
 
-    def to_pandas(self):
+    def to_pandas(self) -> pd.Series:
```
pierce314159, line 349: should we add `@typechecked`?

arkouda/series.py
```diff
@@ -424,7 +444,7 @@ def attach(label, nkeys=1):
 
         return Series((k, v))
 
-    def is_registered(self):
+    def is_registered(self) -> bool:
```
pierce314159, line 447: should we add `@typechecked`?

arkouda/series.py
```diff
@@ -443,14 +463,14 @@ def is_registered(self):
...
-                f"but only located {sum(regParts)}"
+                f"Series expected {len(regParts)} components to be registered, but only located"
+                f" {sum(regParts)}"
             )
 
         return all(regParts)
 
     @staticmethod
-    def from_return_msg(repMsg):
+    def from_return_msg(repMsg: str) -> Series:
```
pierce314159, line 473: should we add `@typechecked`?

pierce314159 (changes requested):
The biggest issue is this PR undoes changes from PR #1508, which allow `index` to be an `Index`

arkouda/series.py
Ethan-DeBandi99: @pierce314159 this was done due to an issue with the circular import that using `from` on the import causes. I know it is not ideal, but `import arkouda.dataframe` is more flexible.

arkouda/series.py
Ethan-DeBandi99: @reuster986 - should locate allow for an index to be passed in?

arkouda/series.py
jeichert60: Put back the code in dataframe the way it was.

arkouda/series.py
jeichert60: Done

arkouda/series.py
jeichert60: Done

arkouda/series.py
jeichert60: Done

arkouda/series.py
jeichert60: Done

Ethan-DeBandi99 (approved):
I think this looks good.

arkouda/series.py
pierce314159: ah i see! i'm cool with this approach

arkouda/series.py
```diff
@@ -486,26 +507,28 @@ def from_return_msg(repMsg):
...
-        If a list of grouped arkouda arrays is passed they are converted to a series.
-        Each grouping is a 2-tuple with the first item being the key(s) and the second being the value.
-
-        If horizontal, each series or grouping must have the same length and the same index.
-        The index of the series is converted to a column in the dataframe.
-        If it is a multi-index,each level is converted to a column.
+    @typechecked
+    def concat(
+        arrays: List, axis: int = 0, index_labels: List[str] = None, value_labels: List[str] = None
+    ) -> Union[arkouda.dataframe.DataFrame, Series]:
```
pierce314159, line 525: Just want to double check that `index_labels` shouldn't be `Union[List[str], str]`. I don't know for certain but the doc string leads me to believe this could be a single name

arkouda/series.py
```diff
@@ -546,31 +574,34 @@ def concat(arrays, axis=0, index_labels=None, value_labels=None):
...
-
-        return retval
+            return Series(index=idx.index, data=v)
 
     @staticmethod
-    def pdconcat(arrays, axis=0, labels=None):
+    @typechecked
+    def pdconcat(
+        arrays, axis: int = 0, labels: Strings = None
+    ) -> Union[arkouda.dataframe.DataFrame, Series, pd.Series, pd.DataFrame]:
```
pierce314159, line 596: we missing type annotations on `arrays`. Also this should only return `pd.Series` and `pd.DataFrame`

pierce314159 (commented):
a few things in the typechecking to look at

arkouda/series.py
pierce314159: We spoke offline, it seems `index_labels` should be the same length as `arrays`. We're going to leave it as is for now

arkouda/series.py
```diff
@@ -149,33 +154,30 @@ def __repr__(self):
...
+    str_acc = CachedAccessor("str", StringAccessor)
 
     @property
     def shape(self):
         # mimic the pandas return of series shape property
         return (self.values.size,)
 
-    def isin(self, lst):
+    @typechecked
+    def isin(self, lst: Union[pdarray, Strings, list]) -> Series:
```
pierce314159, line 180: for the sake of consistency with the other type hints, this should be `List`

pierce314159 (approved):
I'm going to approve since my comment is borderline nitpicky. It should work, but it would be nice to be consistent between the typehints

Either way, i think this is really close/done! I think this is the last comment i'll have

arkouda/series.py
jeichert60: Done

pierce314159 (approved):

PR #1490, opened 2022-06-08
Update additional files to avoid using SysCTypes, CPtr

bradcray:
This is a follow-up to PR #1185 which updated Arkouda to use with
the new 'CTypes' module introduced in Chapel 1.26.  Specifically,
I'd overlooked that other files still relied on SysCTypes and CPtr,
such as those used for the check-* rules in the make process.

Here, I'm updating those files to also use the new 'CTypes' modules
and updating the Makefile to add the ARKOUDA_COMPAT_MODULES
module search path to their compilations as well.  This seems like a
reasonable change in that (as I understand it) the point of the
check-* rules is to see whether the main compilation will work, so
making the 'chpl' command lines more similar seems like it's only
making the check that much better (?).

This work was motivated by the fact that I removed the deprecated
modules on `main` in preparing for this month's Chapel 1.27 and
broke several of our nightly test configurations trying to use Chapel's
`main` branch as a result.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,14 +174,14 @@ endif
 ZMQ_CHECK = $(DEP_INSTALL_DIR)/checkZMQ.chpl
 check-zmq: $(ZMQ_CHECK)
 	@echo "Checking for ZMQ"
-	$(CHPL) $(CHPL_FLAGS) $< -o $(DEP_INSTALL_DIR)/$@
+	$(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@
 	$(DEP_INSTALL_DIR)/$@ -nl 1
 	@rm -f $(DEP_INSTALL_DIR)/$@ $(DEP_INSTALL_DIR)/$@_real
 
 HDF5_CHECK = $(DEP_INSTALL_DIR)/checkHDF5.chpl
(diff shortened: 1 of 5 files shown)


bradcray:
@mhmerrill: Would you be up for reviewing this as a follow-on to #1185?  I think it's pretty straightforward and no-brainer-ish, but let me know if you have questions or concerns.

Ethan-DeBandi99 (approved):
Looks good to me

pierce314159 (approved):
Looks good!

also just wanna check if we also want to update `check-re2` to be 
```
$(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@
```
it doesn't seem to use `SysCTypes` or `CPtr`, but neither does `check-zmq` which got updated

bradcray:
Oh, that's a good point @pierce314159 , and I'm not sure what made me change zmq but not re2—it may have just been that it's a check that I've run afoul of more often, so was more aware of.  I could either back out the change to the zmq test or add it to re2 as you prefer.  I think an argument for adding it would be that if/when other compatibility changes are required to these tests going forward, we'd already be set.  But I could also imagine taking the more minimalist approach of not requiring flags that don't have any impact at present.

bradcray:
(We could also merge this to get testing back on track, and then do any cleanup as a follow-on PR).

PR #1502, opened 2022-06-10
1478 typechecking join

### LINKED ISSUE #1478: Typechecking in join.py
Add typechecking to join.py
Related to #1373

jeichert60:
This PR (Closes #1478):

Adds typechecking to the methods in join.py

### DIFF
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -1,4 +1,4 @@
-from typing import Tuple, Union, cast
+from typing import Callable, Tuple, Union, cast
 
 import numpy as np  # type: ignore
 from typeguard import typechecked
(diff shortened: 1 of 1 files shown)


arkouda/join.py
```diff
@@ -115,7 +115,8 @@ def join_on_eq_with_dt(
     return (resI, resJ)
 
 
-def gen_ranges(starts, ends):
+@typechecked
+def gen_ranges(starts: pdarray, ends: pdarray):
```
pierce314159, line 119: we should add the return type here and in the other functions. something like
```python
@typechecked
def gen_ranges(starts: pdarray, ends: pdarray) -> Tuple[pdarray, pdarray]:
```

pierce314159 (commented):
We need to add the return types but looks good other than that

Ethan-DeBandi99 (commented):
Pierce already noted that typing needs to be added to the returns. Other than that, looks good.

joshmarshall1 (approved):

PR #1504, opened 2022-06-10
fix channel order in conda yml file

### LINKED ISSUE #1503: switch order of `conda-forge` and `defaults` channel in `.yml` conda config files
i had a bunch of difficulty with the version of `pyarrow` on my x86_64 mac running catalina os.

basically i get unknown x86_64 symbols in `pyarrow` library.
i will note that this worked fine with brew installed chapel 1.25.1.1

`conda-forge` supplies `pyarrow` ver 7.0.0
`defaults` channel supplies `pyarrow` ver 4.0.1

this seems to be a bad interaction with the new brew install of chapel 1.26, the linker, and `pyarrow` 7.0.0, this could be an artifact of the way chapel 1.26 brew package is built or something.

the fix for me was to swap the channel order and put `defaults` first which installs `pyarrow` 4.0.1

mhmerrill:
closes #1503 

`pyarrow` 7.0.0 from the conda-forge channel does not work for me on my x86_64 mac running catalina os.

flipping the order of the channels installs `pyarrow` 4.0.1 which works... any explanation beyond this is in the issue.

### DIFF
--- a/arkouda-env-dev.yml
+++ b/arkouda-env-dev.yml
@@ -1,7 +1,7 @@
 name: arkouda-env-dev
 channels:
-  - conda-forge
   - defaults
+  - conda-forge
 dependencies:
   - python>=3.8   # minimum 3.8
   - numpy>=1.16.5,<=1.21.5
--- a/arkouda-env.yml
+++ b/arkouda-env.yml
@@ -1,7 +1,7 @@
 name: arkouda-env
 channels:
+  - defaults
   - conda-forge
-  - -defaults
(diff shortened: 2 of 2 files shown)


Ethan-DeBandi99 (approved):
This looks fine to me. My machine has the old order and is function correctly, but I believe that may be because I am on the arm64 processor.

Also, I know @pierce314159 had similar issues on x86 chipsets, so it is probably better to change it and make note of any variance for the arm chips like we do for installs.

PR #1507, opened 2022-06-13
Closes #1360: Utilize append functionality in `ak.save_all()` 

### LINKED ISSUE #1360: `ak.save_all()` utilize `append` updates
While reviewing PR #1358, I noticed that `ak.save_all()` had not been updated to take advantage of the new append mode functionality to create the file(s) if they do not exist. The logic should be updated to take advantage of this.

We should review `.save()` functionality elsewhere to ensure this new feature is taken advantage of because it will simplify client side code.

jeichert60:
This PR (closes #1360):
    
  - takes advantage of the new append mode functionality to create files if they don't exist.

### DIFF
--- a/arkouda/pdarrayIO.py
+++ b/arkouda/pdarrayIO.py
@@ -631,17 +631,13 @@ def save_all(
         pdarrays = cast(List[pdarray], columns)
         if names is None:
             datasetNames = [str(column) for column in range(len(columns))]
-    if (mode.lower() not in "append") and (mode.lower() not in "truncate"):
+    if mode.lower() not in ["append", "truncate"]:
         raise ValueError("Allowed modes are 'truncate' and 'append'")
-    first_iter = True
+
     for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
-        """Append all pdarrays to existing files as new datasets EXCEPT the first one,
-        and only if user requests truncation"""
-        if mode.lower() != "append" and first_iter:
-            arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode="truncate")
-            first_iter = False
-        else:
-            arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode="append")
+        arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode=mode)
+        if mode.lower() == "truncate":
+            mode = "append"
 
 
 @typechecked


Ethan-DeBandi99 (changes requested):
After reviewing this, I don't think we have quite hit the mark with this. We have added complexity in order to address the issue. Reviewing the code from master, I think it is closer to what we want than when this issue was posted.

arkouda/pdarrayIO.py
```diff
@@ -633,15 +633,20 @@ def save_all(
...
+    else:
+        first_iter = True
+        for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
+            """Append all pdarrays to existing files as new datasets EXCEPT the first one,
+            and only if user requests truncation"""
+            if first_iter:
+                arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode="truncate")
+                first_iter = False
+            else:
+                arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode="append")
```
Ethan-DeBandi99, line 649: I am not sure this hits the mark for what we want. It is adding an additional loop that instead of 1. I would recommend something like 

```python
    for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
        arr.save(prefix_path=prefix_path, dataset=name, file_format=file_format, mode=mode)
        if mode.lower() == "truncate":
            mode = "append"
```

If we wanted a simple fix, we could change line 640 from master to 
```python
if mode.lower() == "truncate" and first_iter
```

Overall, this does not look like it is as much of an issue as initially thought. I would suggest going with one of the solutions above to address the issue.

Side note: I forgot to click add when I did my review initially, sorry!

jeichert60: I like your first suggestion.

arkouda/pdarrayIO.py
```diff
@@ -633,15 +633,20 @@ def save_all(
             datasetNames = [str(column) for column in range(len(columns))]
     if (mode.lower() not in "append") and (mode.lower() not in "truncate"):
```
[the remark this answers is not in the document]
pierce314159: I think the reason it was done this way was to allow mode to be `'a', 'app', 't',` instead of just `'append'` and `'truncate'`

That being said, i like this better. I think it's much cleaner and it matches the error message and the docstring

Ethan-DeBandi99 (approved):
Looks good

joshmarshall1 (approved):

PR #1513, opened 2022-06-15
Closes #1514: Remove trailing comma from 1 tuples in array creation message

### LINKED ISSUE #1514: Make Arkouda compatible with Chapel main trailing commas
In https://github.com/chapel-lang/chapel/pull/19713, trailing commas
were added to 1 tuples and array creation was splitting on commas,
which caused a problem with the new comma. This PR removes the trailing
comma in the case of a 1-tuple for array creation.

bmcdonald3:
In https://github.com/chapel-lang/chapel/pull/19713, trailing commas
were added to 1 tuples and array creation was splitting on commas,
which caused a problem with the new comma. This PR removes the trailing
comma in the case of a 1-tuple for array creation.

Closes #1514

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -1636,6 +1636,10 @@ def create_pdarray(repMsg: str) -> pdarray:
         mydtype = fields[2]
         size = int(fields[3])
         ndim = int(fields[4])
+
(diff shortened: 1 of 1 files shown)


mhmerrill (approved):
looks ok to me. i always get confused because Chapel uses `(10,)` to disambiguate between 1-tuple and other things.

PR #1515, opened 2022-06-15
This PR (Closes #1465):

### LINKED ISSUE #1465: Typechecking in `series.py`

jeichert60:
- Adds Groupby.size() method


arkouda/groupbyclass.py
```diff
@@ -263,9 +263,12 @@ def __init__(
                 self.keys, return_groups=True, assume_sorted=self.assume_sorted
             )
 
-        self.size = self.permutation.size
+        self.length = self.permutation.size
         self.ngroups = self.segments.size
 
+    def size(self):
```
Ethan-DeBandi99, line 269: Add type checking here.

arkouda/groupbyclass.py
```diff
@@ -263,9 +263,12 @@ def __init__(
...
+        return self.count()
+
     def count(self) -> Tuple[groupable, pdarray]:
```
Ethan-DeBandi99, line 272: Missing the `@typechecked`. Since we are in the file and modifying this functionality, we should add it here.

arkouda/groupbyclass.py
```diff
@@ -263,9 +263,12 @@ def __init__(
                 self.keys, return_groups=True, assume_sorted=self.assume_sorted
             )
 
-        self.size = self.permutation.size
+        self.length = self.permutation.size
         self.ngroups = self.segments.size
 
+    def size(self):
+        return self.count()
```
Ethan-DeBandi99, line 270: We need to add a docstring to this function. It should indicate that this alias was added to conform with Pandas API.

tests/groupby_test.py
```diff
@@ -187,6 +187,11 @@ def test_boolean_arrays(self):
         self.assertEqual(ct[1], true_ct)
         self.assertListEqual(k.to_ndarray().tolist(), [False, True])
 
+        k, ct = g.size()
```
Ethan-DeBandi99, line 190: Can we add a comment that size is being added and tested for the issue number? Its not completely necessary, just helps keep track of why things were added and when the testing was added.

Ethan-DeBandi99 (changes requested):
Few things need to be cleaned up.

arkouda/groupbyclass.py
jeichert60: Done

arkouda/groupbyclass.py
jeichert60: Done

tests/groupby_test.py
jeichert60: Done

arkouda/groupbyclass.py
Ethan-DeBandi99: The docstring at a minimum should also have a `See Also` section that points lists count. I would also add a Returns for users using `Groupby.size?` in ipython to view docstring, just makes it easier for them to know what to expect.

Ethan-DeBandi99 (changes requested):

arkouda/groupbyclass.py
```diff
@@ -263,10 +263,41 @@ def __init__(
                 self.keys, return_groups=True, assume_sorted=self.assume_sorted
             )
 
-        self.size = self.permutation.size
+        self.length = self.permutation.size
         self.ngroups = self.segments.size
 
-    def count(self) -> Tuple[groupable, pdarray]:
+    @typechecked
+    def size(self) -> Tuple[Union[Tuple, pdarray, Strings], pdarray]:
```
pierce314159, line 270: I think the typechecking for `size` and `count` should be the same as `count` was before i.e.
```python
def count(self) -> Tuple[groupable, pdarray]:
```

This will include the elements you have here `pdarray` and `Strings`. As well as `Categorical` and or other sequence of these (`Tuple`, `List`, ...)
https://github.com/Bears-R-Us/arkouda/blob/b6122cba6b267c2c12ce6c7d850b47c37e84391f/arkouda/groupbyclass.py#L28-L29

pierce314159 (commented):
Change requested for the typechecking, looks good other than that!

Also the title of the PR seems like it was meant to be in the body. I think editing it to be something like below would be perfect
```
# Title
Closes #1353: Add `Groupyby.size()` to match pandas API

# Body
This PR (Closes #1353): 
- Adds `Groupby.size()` method (an alias for `Groupby.count()`) in order to match the pandas API
- Changes attribute `Groupby.size` to be `Groupby.length`
```

arkouda/groupbyclass.py
jeichert60: When I tested it using the original type hints, I got an error that said Categorical was undefined.

pierce314159: I just pulled down the PR and i see what you mean, we should still support `Categorical`. So let's restore the original type hints and remove `@typechecked`

I also think we should add a comment near the top (perhaps right under the definition of `groupable`) that says 
```
Note: we won't be typechecking GroupBy until we can figure out a way to handle the circular import with Categorical
```

jeichert60: Done

arkouda/groupbyclass.py
jeichert60: Done

arkouda/groupbyclass.py
```diff
@@ -263,9 +265,38 @@ def __init__(
                 self.keys, return_groups=True, assume_sorted=self.assume_sorted
             )
 
-        self.size = self.permutation.size
+        self.length = self.permutation.size
         self.ngroups = self.segments.size
 
+    def size(self) -> Tuple[groupable, pdarray]:
+        """
+        This alias for "count" was added to conform with Pandas API
```
Ethan-DeBandi99, line 273: I would move this into a `Notes` section in the docstring. The descriptions here should be similar to count.

Ethan-DeBandi99 (changes requested):
Just a small rearrangement. Sorry I was not more specific on the last round.

pierce314159 (approved):
Looks good! I think this will be good to go once ethan's comment is addressed

joshmarshall1 (approved):

Ethan-DeBandi99 (approved):

PR #1521, opened 2022-06-17
Closes #1520: Always recompile Arrow when running make or make check-arrow

### LINKED ISSUE #1520: Always recompile Arrow when running `make` or `make check-arrow`
If Arrow versions are changed, the Makefile currently won't recompile `ArrowFunctions.cpp` because, to the Makefile, nothing has changed for that file. This was causing problems in certain environments.

bmcdonald3:
This PR makes it so that the Arrow C++ code will always recompile
on a `make` or `make check-arrow`. When upgrading Arrow versions,
if the C++ code is not recompiled, you could end up having problems
with the version that the C++ code was compiled with and the version
that the Chapel code is trying to link with, which will cause
problems.

Also, this PR adds removal of `ArrowFunctions.o` to `make clean`.

Closes: https://github.com/Bears-R-Us/arkouda/issues/1520

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,7 +153,7 @@ ARROW_SANITIZE=-fsanitize=$(SANITIZER)
 endif
 
 .PHONY: compile-arrow-cpp
-compile-arrow-cpp: $(ARROW_CPP) $(ARROW_H)
+compile-arrow-cpp:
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

mhmerrill (approved):
thanks. this will cut down on some headaches.

PR #1526, opened 2022-06-22
Closes #1525 - security alert for `numpy<=1.21.6`

### LINKED ISSUE #1525: security alert for numpy<=1.21.6
there is a security alert about numpy<=1.21.6 

I really don't understand the message but it also says versions effected are numpy<=1.21.6

(https://github.com/Bears-R-Us/arkouda/security/dependabot/3)

"Incomplete string comparison in the numpy.core component in NumPy1.9.x, which allows attackers to fail the APIs via constructing specific string objects."

Ethan-DeBandi99:
Closes #1525 

- Updated numpy requirement to `numpy>=1.22.2`.

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -13,7 +13,7 @@ The dependencies listed here have varying installation instructions depending up
 The following python packages are required by the Arkouda client package.
 
 - `python>=3.8`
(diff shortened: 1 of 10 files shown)


pierce314159 (approved):
Looks good!

PR #1530, opened 2022-06-23
Closes #1529: Quick fix for np > 1.21

### LINKED ISSUE #1529: Quick fix for numpy >1.21
Convert checks in `numeric_test` from `==` to `np.isclose` and investigate mypy complaints

pierce314159:
This PR (Closes #1529):
- Updates `numeric_test` comparisons with numpy operations using floats to use `np.isclose` instead of `==` to avoid test failure due to numeric precision issues (differing in last decimal place)
- Adds `# type: ignore` to an instance of `np.linspace` because mypy struggles sometimes to get correct typing information from numpy. Verified locally with version `1.22.2` that `np.linspace` does work with a mix of integer size (`np.int16`, `np.int32`,...)

### DIFF
--- a/tests/numeric_test.py
+++ b/tests/numeric_test.py
@@ -140,7 +140,7 @@ class NumericTest(ArkoudaTest):
         na = np.linspace(1, 10, 10)
         pda = ak.array(na)
 
-        self.assertTrue(np.isclose(np.log(na), ak.log(pda).to_ndarray()).all())
+        self.assertTrue(np.allclose(np.log(na), ak.log(pda).to_ndarray()))
         with self.assertRaises(TypeError):
             ak.log([range(0, 10)])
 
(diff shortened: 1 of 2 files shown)


tests/numeric_test.py
```diff
@@ -140,23 +140,23 @@ def testLog(self):
         na = np.linspace(1, 10, 10)
         pda = ak.array(na)
 
-        self.assertTrue((np.log(na) == ak.log(pda).to_ndarray()).all())
+        self.assertTrue(np.isclose(np.log(na), ak.log(pda).to_ndarray()).all())
```
mhmerrill, line 143: could you use `np.allclose` ?

pierce314159: sure!

mhmerrill (approved):
looks good.

PR #1532, opened 2022-06-23
Closes #1535: More efficient `lookup` algorithm

### LINKED ISSUE #1535: More efficient `lookup` algorithm
The previous implementation of `lookup` required up to 3 `GroupBy` operations. Re-implement to only require one. Initial testing suggests a ~40% speedup

reuster986:
This PR (Closes #1535): 
- Refactors the `lookup` function to be more efficient. The previous implementation required up to 3 GroupBy ops, whereas this one requires only one. Initial testing suggests a ~40% speedup in a real use case.

### DIFF
--- a/arkouda/alignment.py
+++ b/arkouda/alignment.py
@@ -7,7 +7,7 @@ from arkouda.dtypes import uint64 as akuint64
 from arkouda.dtypes import float64 as akfloat64
 from arkouda.groupbyclass import GroupBy, broadcast, unique
 from arkouda.pdarrayclass import is_sorted, pdarray
-from arkouda.pdarraycreation import arange, ones, zeros
+from arkouda.pdarraycreation import arange, ones, zeros, full
 from arkouda.pdarraysetops import argsort, concatenate, in1d
 from arkouda.strings import Strings
 from arkouda.numeric import where
@@ -281,8 +281,7 @@ def lookup(keys, values, arguments, fillvalue=-1, keys_from_unique=False):
     # Find arguments in keys array
     idx = find(arguments, keys)
     # Initialize return values with fillvalue for missing values
-    retvals = ak.zeros(idx.size, dtype=values.dtype)
-    retvals.fill(fillvalue)
+    retvals = full(idx.size, fillvalue, dtype=values.dtype)
     # Where arguments were found in keys, put corresponding fuction values
     found = idx >= 0
     retvals[found] = values[idx[found]]
--- a/arkouda/alignment.py
+++ b/arkouda/alignment.py
@@ -11,6 +11,7 @@ from arkouda.pdarraycreation import arange, ones, zeros, full
 from arkouda.pdarraysetops import argsort, concatenate, in1d
 from arkouda.strings import Strings
 from arkouda.numeric import where
+from warnings import warn
 
 
 def unsqueeze(p):
@@ -172,23 +173,24 @@ def in1dmulti(a, b, assume_unique=False, symmetric=False):
         else:
             return atruth
 
+
 def find(query, space):
     """
     Return indices of query items in a search list of items (-1 if not found).
-    
+
     Parameters
     ----------
     query : (sequence of) array-like
         The items to search for. If multiple arrays, each "row" is an item.
     space : (sequence of) array-like
         The set of items in which to search. Must have same shape/dtype as query.
-        
+
     Returns
     -------
     indices : pdarray, int64
         For each item in query, its index in space or -1 if not found.
     """
-    
+
     # Concatenate the space and query in fast (block interleaved) mode
     if isinstance(query, (pdarray, Strings, Categorical)):
         if type(query) != type(space):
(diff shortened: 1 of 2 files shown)


arkouda/alignment.py
```diff
@@ -223,36 +275,18 @@ def lookup(keys, values, arguments, fillvalue=-1, keys_from_unique=False):
...
-    # Set values for arguments present in keys
-    # Set them to the values corresponding to elements queried
-    uvals[scattermask] = values[gathermask]
-    # Broadcast return values back to non-unique arguments
-    return g.broadcast(uvals, permute=True)
+    # Find arguments in keys array
+    idx = find(arguments, keys)
+    # Initialize return values with fillvalue for missing values
+    retvals = ak.zeros(idx.size, dtype=values.dtype)
+    retvals.fill(fillvalue)
```
pierce314159, line 285: I think at least part of the problem is the `ak.` shouldn't be here. Also you can do this in one line using `full`
```python
retvals = full(idx.size, fillvalue, dtype=values.dtype)
```

reuster986: Ah, thanks!

arkouda/alignment.py
```diff
@@ -172,6 +175,63 @@ def in1dmulti(a, b, assume_unique=False, symmetric=False):
...
+            raise TypeError("Arguments must have same type")
+        c = concatenate((space, query), ordered=False)
+        spacesize = space.size
+        querysize = query.size
+    else:
+        atypes = np.array([ai.dtype for ai in query])
+        btypes = np.array([bi.dtype for bi in space])
+        if not (atypes == btypes).all():
+            raise TypeError("Array dtypes of arguments must match")
+        c = [concatenate((si, qi), ordered=False) for si, qi in zip(space, query)]
```
pierce314159, line 207: should we do a check that `query` and `space` are the same size? I think the `(atype == btypes).all()` will fail if they aren't, but I don't think the error message would be very helpful

arkouda/alignment.py
```diff
@@ -172,6 +175,63 @@ def in1dmulti(a, b, assume_unique=False, symmetric=False):
...
+        spacesize = space[0].size
+        querysize = query[0].size
```
pierce314159, line 209: is every element of `space`/`query` guaranteed to have the same size as the first?

arkouda/alignment.py
```diff
@@ -172,6 +175,63 @@ def in1dmulti(a, b, assume_unique=False, symmetric=False):
...
+    # Combined index of space and query elements, in block interleaved order
+    # All space indices are less than all query indices
+    i = concatenate((arange(spacesize), arange(spacesize, spacesize + querysize)), ordered=False)
```
pierce314159, line 212: I might be missing something but it's not immediately obvious to me how this is different than
```python
i = arange(spacesize + querysize)
```

tests/join_test.py
```diff
@@ -129,7 +130,7 @@ def test_lookup(self):
         res3 = ak.lookup(keys[::-1], values[::-1], args, fillvalue=-1)
         self.assertTrue((res3.to_ndarray() == ans).all())
         # Non-unique keys should raise error
-        with self.assertRaises(ak.NonUniqueError):
+        with pytest.warns(UserWarning):
```
pierce314159, line 133: in other places in our tests, we use
```python
with self.assertWarns(UserWarning):
```
I'm not sure if one is better than the other, but we should prob be consistent

pierce314159 (commented):
A couple places i'd like some clarification. I don't see any issues with the logic, but also haven't looked at it super hard

arkouda/alignment.py
reuster986: It ends up being different because of the `ordered=False`, which block-interleaves the two arrays, rather than sequentially concatenating them. The goal is to get a block-interleaved array of indices that mirrors the block-interleaved array of elements from `space` and `query`.

tests/join_test.py
reuster986: Ah, thanks! I was looking for a `self` method related to warnings, but couldn't find it.

arkouda/alignment.py
reuster986: Good point -- will add a check

arkouda/alignment.py
reuster986: Another good point. Aside: it might be cool to create a `MultiArray` class that ensures all arrays are the same size and carries a `.size` attribute. Like a `DataFrame` without all the methods -- useful for all these functions that accept either arrays or lists of same-length arrays.

arkouda/alignment.py
pierce314159: ah okay! that makes sense

pierce314159 (approved):
Looks good to me!

arkouda/alignment.py
```diff
@@ -172,6 +175,69 @@ def in1dmulti(a, b, assume_unique=False, symmetric=False):
...
+        if len(query) != len(space):
+            raise TypeError("Multi-array arguments must have same number of arrays")
+        spacesize = set(s.size for s in space)
+        querysize = set(q.size for q in query)
+        if len(spacesize) != 1 or len(querysize) != 1:
+            raise TypeError("Multi-array arguments must be non-empty and have equal-length arrays")
+        spacesize = spacesize.pop()
+        querysize = querysize.pop()
+        atypes = np.array([ai.dtype for ai in query])
+        btypes = np.array([bi.dtype for bi in space])
```
Ethan-DeBandi99, line 212: Noting that this will break if anything is Categorical because `.dtype` is not yet supported.

Ethan-DeBandi99 (approved):
This looks good to me. Noted one spot where categorical will be a problem. We do this in so many locations, I do not think we should hold up the PR. I am going to look at working the issue for Categorical to support `dtype` today.

joshmarshall1 (approved):

jeichert60 (approved):

PR #1533, opened 2022-06-23
Draft: Second iteration of radix sort instances, using `record RadixSortLSDPlan` 

pierce314159:
Part of #1338
Very basic iteration, all we do here is wrap the parameters for `RadixSortCore` into a record and pass them in

I think eventually we want to allocate the arrays `a` and `temp` and store them in `RadixSortLSDPlan` as well. I'm not sure how this will affect performance because right now, i'm passing `plan` in as a `const` and I don't we'll be able to do that then since those arrays will be changing

I think once i figure out the above, we'll build some sort of map so we can cache `RadixSortLSDPlan`s. That way if we request one with the same configuration as an existing one, we can reuse the existing one. This'll save us the cost of allocating and deallocating arrays for repetitive sorts with the same configuration. Though we'll prob want to cap this somehow to avoid too much memory being used by the cache

Once again, `ak.sort` is my guinea pig function
```python
>>> a = ak.randint(1, 15, 10)
>>> a
array([13 11 4 5 6 11 4 8 8 6])

>>> ak.sort(a)
array([4 4 5 6 6 8 8 11 11 13])

>>> ak.sort(a, bits_per_digit=4)
array([4 4 5 6 6 8 8 11 11 13])

>>> ak.sort(a, num_tasks=22, bits_per_digit=4)
array([4 4 5 6 6 8 8 11 11 13])
```

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -27,6 +27,12 @@ module RadixSortLSD {
         }
     }
 
+    proc makeRadixSortLSDPlan(numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit): RadixSortLSDPlan {
+        // for now this factory method is equivalent to calling the constructor but in the future
+        // it could return a cached plan object if one already exists with the same config info
+        return new RadixSortLSDPlan(numTasks, bitsPerDigit);
+    }
+
     record KeysComparator {
       inline proc key(k) { return k; }
     }
@@ -163,21 +169,20 @@ module RadixSortLSD {
         } // for rshift
     }//proc radixSortLSDCore
 
-    proc radixSortLSD(a:[?aD] ?t, checkSorted: bool = true, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit): [aD] (t, int) {
+    proc radixSortLSD(a:[?aD] ?t, checkSorted: bool = true, const plan: RadixSortLSDPlan): [aD] (t, int) {
         var kr: [aD] (t,int) = [(key,rank) in zip(a,aD)] (key,rank);
         if (checkSorted && isSorted(a)) {
             return kr;
         }
         var (nBits, negs) = getBitWidth(a);
-        var plan = new RadixSortLSDPlan(numTasks, bitsPerDigit);
-        radixSortLSDCore(kr, nBits, negs, new KeysRanksComparator(), plan);
+        radixSortLSDCore(kr, nBits, negs, new KeysRanksComparator(), plan = plan);
         return kr;
     }
 
     /* Radix Sort Least Significant Digit
        radix sort a block distributed array
        returning a permutation vector as a block distributed array */
-    proc radixSortLSD_ranks(a:[?aD] ?t, checkSorted: bool = true, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit): [aD] int {
+    proc radixSortLSD_ranks(a:[?aD] ?t, checkSorted: bool = true, const plan: RadixSortLSDPlan): [aD] int {
         if (checkSorted && isSorted(a)) {
             var ranks: [aD] int = [i in aD] i;
             return ranks;
@@ -185,8 +190,7 @@ module RadixSortLSD {
 
         var kr: [aD] (t,int) = [(key,rank) in zip(a,aD)] (key,rank);
         var (nBits, negs) = getBitWidth(a);
-        var plan = new RadixSortLSDPlan(numTasks, bitsPerDigit);
-        radixSortLSDCore(kr, nBits, negs, new KeysRanksComparator(), plan);
+        radixSortLSDCore(kr, nBits, negs, new KeysRanksComparator(), plan = plan);
         var ranks: [aD] int = [(_, rank) in kr] rank;
         return ranks;
     }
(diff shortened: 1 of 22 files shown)


src/RadixSortLSD.chpl
```diff
@@ -25,6 +13,20 @@ module RadixSortLSD
     private config const logLevel = ServerConfig.logLevel;
     const rsLogger = new Logger(logLevel);
 
+    record RadixSortLSDPlan {
```
mhmerrill, line 16: I would also add/prefer a factory function called `makeRadixSortLSDPlan()` which creates and returns a plan, that way in the future we can do more than just `init()` then plan.

src/RadixSortLSD.chpl
```diff
@@ -159,52 +163,55 @@ module RadixSortLSD
         } // for rshift
     }//proc radixSortLSDCore
 
-    proc radixSortLSD(a:[?aD] ?t, checkSorted: bool = true): [aD] (t, int) {
+    proc radixSortLSD(a:[?aD] ?t, checkSorted: bool = true, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit): [aD] (t, int) {
```
mhmerrill, line 166: I would like you to pull the plan creation up higher in the call stack, in this case into the `sortMsg()` that way in the future we can choose a plan based on the input size and have less parameters.

src/RadixSortLSD.chpl
```diff
@@ -159,52 +163,55 @@ module RadixSortLSD
...
+        var plan = new RadixSortLSDPlan(numTasks, bitsPerDigit);
+        radixSortLSDCore(kr, nBits, negs, new KeysRanksComparator(), plan);
         return kr;
     }
 
     /* Radix Sort Least Significant Digit
        radix sort a block distributed array
        returning a permutation vector as a block distributed array */
-    proc radixSortLSD_ranks(a:[?aD] ?t, checkSorted: bool = true): [aD] int {
+    proc radixSortLSD_ranks(a:[?aD] ?t, checkSorted: bool = true, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit): [aD] int {
```
mhmerrill, line 180: Same comment as before: I would like you to pull the `plan` creation up higher in the call stack, in this case into the `sortMsg()` that way in the future we can choose a plan based on the input size and have less parameters.

src/RadixSortLSD.chpl
```diff
@@ -159,52 +163,55 @@ module RadixSortLSD
...
+        radixSortLSDCore(kr, nBits, negs, new KeysRanksComparator(), plan);
         var ranks: [aD] int = [(_, rank) in kr] rank;
         return ranks;
     }
 
     /* Radix Sort Least Significant Digit
        radix sort a block distributed array
        returning sorted keys as a block distributed array */
-    proc radixSortLSD_keys(a: [?aD] ?t, checkSorted: bool = true): [aD] t {
+    proc radixSortLSD_keys(a: [?aD] ?t, checkSorted: bool = true, numTasks: int = here.maxTaskPar, bitsPerDigit: int = RSLSD_bitsPerDigit): [aD] t {
```
mhmerrill, line 197: Same comment as before: although in this case it would be `argsortMsg` and that would be more invasive of a change.

src/RadixSortLSD.chpl
```diff
@@ -159,52 +163,55 @@ module RadixSortLSD
...
         }
         var (nBits, negs) = getBitWidth(a);
-        radixSortLSDCore(copy, nBits, negs, new KeysComparator());
+        var plan = new RadixSortLSDPlan(numTasks, bitsPerDigit);
+        radixSortLSDCore(copy, nBits, negs, new KeysComparator(), plan);
         return copy;
     }
 
-    proc radixSortLSD_memEst(size: int, itemsize: int) {
+    proc radixSortLSD_memEst(size: int, itemsize: int, numTasks: int = here.maxTaskPar, numBuckets: int = 1 << RSLSD_bitsPerDigit) {
```
mhmerrill, line 208: same comment as before.

src/RadixSortLSD.chpl
```diff
@@ -159,52 +163,55 @@ module RadixSortLSD
...
         // 2 temp key+ranks arrays + globalStarts/globalClounts
         return (2 * size * (itemsize + numBytes(int))) +
                (2 * numLocales * numTasks * numBuckets * numBytes(int));
     }
 
-    proc radixSortLSD_keys_memEst(size: int, itemsize: int) {
+    proc radixSortLSD_keys_memEst(size: int, itemsize: int, numTasks: int = here.maxTaskPar, numBuckets: int = 1 << RSLSD_bitsPerDigit) {
```
mhmerrill, line 214: same comment as before

src/SortMsg.chpl
```diff
@@ -30,7 +30,7 @@ module SortMsg
     proc sortMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
       param pn = Reflection.getRoutineName();
       var repMsg: string; // response message
-      var (algoName, name) = payload.splitMsgToTuple(2);
+      var (algoName, name, numTasksStr, bitsPerDigitStr) = payload.splitMsgToTuple(4);
```
mhmerrill, line 33: in the future we would choose the best plan unless the were set to sane values.

src/SortMsg.chpl
```diff
@@ -65,7 +68,10 @@ module SortMsg
             return b;
           }
           when SortingAlgorithm.RadixSortLSD {
-            return radixSortLSD_keys(a);
+            var numTasks = if nt != -1 then nt else here.maxTaskPar;
+            var bitsPerDigit = if bpd != -1 then bpd else RSLSD_bitsPerDigit;
+
```
mhmerrill, line 73: for now I would create the plan here calling `const plan = makeRadixSortLSDPlan()` or something.

src/RadixSortLSD.chpl
pierce314159: it took me a min to figure out how this was different that using the constructor, but i think i understand now.

So for the time being, it'll be equivalent to initializing a plan, but in the future it could return a preexisting plan with the same configuration (which was cached from a previous sort)

am i understanding this correctly?

src/RadixSortLSD.chpl
pierce314159: yeah that was always the plan, but did it this way in the interest of not changing too many things at once to keep the draft a bit easier to read. I'll go ahead and make these changes

src/RadixSortLSD.chpl
pierce314159: I'm not as sure about doing this in the `memEst` functions. Because if we move towards having the `RadixSortLSDPlan` contain the arrays, we would want to call the memory estimates before we actually created the `plan`, right? I'll make this change for now, since we're not currently having the plan object store the created arrays and it's easy enough to switch back

src/RadixSortLSD.chpl
mhmerrill: yes, and it could throw exceptions which you can't do in a `class`es `init()`

mhmerrill:
just so it doesn't get lost:

- we had talked about making a  `makeRadixSortLSDPlanMsg` message to the server so there is one place to provide the parameters for making a plan.
- and then all of the places where `RadixSortLSD` is called we would call `chooseRadixSortLSDPlan`, this would remove the parameters from all the messages for making a plan.

Ethan-DeBandi99:
This work did not yield the results expected. Closing this because nothing further is being done at this time.

PR #1534, opened 2022-06-24
This PR (Closes #1518):

### LINKED ISSUE #1518: Add `to_title` and `is_title` methods to `Strings`
Create `to_title` and `is_title` methods for `Strings` which behave like their corresponding methods on python `str`'s ([title](https://docs.python.org/3/library/stdtypes.html#str.title) and [istitle](https://docs.python.org/3/library/stdtypes.html#str.istitle)) applied to every element of the SegString. You should use chapel builtins for strings ([toTitle](https://chapel-lang.org/docs/builtins/String.html#String.string.toTitle) and [isTitle](https://chapel-lang.org/docs/builtins/String.html#String.string.isTitle))

Note this is very similar to #1274 which will be a good example to follow

jeichert60:
- Adds Strings.to_title()
    - Returns a new Strings with lower, mixed, or upper characters from the original replaced with their titlecase equivalent
- Adds Strings.is_title()
    - Returns a boolean pdarray where index i indicates whether string i of the Strings is titlecase
- Adds a test for the new functionality


arkouda/strings.py
```diff
@@ -459,6 +489,40 @@ def is_upper(self) -> pdarray:
...
+            True for elements that are titlecase, False otherwise
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there is a server-side error thrown
+
+        See Also
+        --------
+        Strings.is_title
```
pierce314159, line 510: same here it should be `See Also is_upper and is_lower`

tests/string_test.py
```diff
@@ -562,6 +566,13 @@ def test_case_change(self):
         expected = ak.arange(30) >= 20
         self.assertListEqual(isupper.to_ndarray().tolist(), expected.to_ndarray().tolist())
 
+        # first 10 all lower, middle 10 mixed case (not lower or upper), last 10 all title
+        t = ak.concatenate([lower, s, title])
+
+        istitle = t.is_title()
+        expected = ak.arange(30) >= 20
+        self.assertListEqual(istitle.to_ndarray().tolist(), expected.to_ndarray().tolist())
+
```
pierce314159, line 575: I think the better way to do this is to modify the existing `lu` to something like this
```python
# first 10 all lower, second 10 mixed case (not lower, upper, or title), third 10 all upper, last 10 all title
lmut = ak.concatenate([lower, s, upper, title])

islower = lmut.is_lower()
expected = 10 > ak.arange(40)
self.assertListEqual(islower.to_ndarray().tolist(), expected.to_ndarray().tolist())

isupper = lmut.is_upper()
expected = (30 > ak.arange(40)) & (ak.arange(40) >= 20)
self.assertListEqual(isupper.to_ndarray().tolist(), expected.to_ndarray().tolist())

istitle = lmut.is_title()
expected = ak.arange(40) >= 30
self.assertListEqual(istitle.to_ndarray().tolist(), expected.to_ndarray().tolist())
```

pierce314159: I know i wrote the original, so this one is on me. But I would also rename `s` to `mixed` so it's more clear

pierce314159 (approved):
Looks good to me! Nice work

Ethan-DeBandi99 (approved):
Looks good to me

joshmarshall1 (approved):

Issue #1398, opened 2022-05-18
DataFrame Display Server-Side Performance Issue


Ethan-DeBandi99:
@ronawho ran the benchmark to test `_get_head_tail_server`'s performance. The results are

```
numLocales = 16, N = 160,000
  _get_head_tail_server Average time = 14.2314 sec
  _get_head_tail_server Average rate = 0.00 GiB/sec
  _get_head_tail Average time = 3.1467 sec
  _get_head_tail Average rate = 0.01 GiB/sec
```

This is much worse performance than we were hoping for. This initial issue was #1300.

ronawho:
Adding some subtimers to the client on 16-node-cs-hdr I see:

10.7s on the actual server call:

https://github.com/Bears-R-Us/arkouda/blob/aba0b4149cf37767635531096d8fa62132bd61a5/arkouda/dataframe.py#L550-L551

0.65s creating `df_dict`: 

https://github.com/Bears-R-Us/arkouda/blob/aba0b4149cf37767635531096d8fa62132bd61a5/arkouda/dataframe.py#L554-L567

and 0.3s in the `to_pandas` call:

https://github.com/Bears-R-Us/arkouda/blob/aba0b4149cf37767635531096d8fa62132bd61a5/arkouda/dataframe.py#L571


Digging into the server call, I found most of the time was spent localizing the segments in `df_seg_array_idx`:

https://github.com/Bears-R-Us/arkouda/blob/aba0b4149cf37767635531096d8fa62132bd61a5/src/DataFrameIndexingMsg.chpl#L77-L82

---

What I'm seeing is that most segments were of len 5, but one was of len 799500, and gathering that one is where al the time is going. This is because orig_segs looks like:


> 0 5 10 15 20 25 30 35 40 45 50 55 60 65 70 75 80 85 90 95 100 105 110 115 120 125 130 135 140 145 150 155 160 165 170 175 180 185 190 195 200 205 210 215 220 225 230 235 240 245 250 799750 799755 799760 799765 799770 799775 799780 799785 799790 799795 799800 799805 799810 799815 799820 799825 799830 799835 799840 799845 799850 799855 799860 799865 799870 799875 799880 799885 799890 799895 799900 799905 799910 799915 799920 799925 799930 799935 799940 799945 799950 799955 799960 799965 799970 799975 799980 799985 799990 799995

[...]

PR #1537, opened 2022-06-24
Closes #1536: Switch to using a lowLevelLocalizingSlice for DF indexing message

### LINKED ISSUE #1536: Improve dataframe get head tail server performance
Since we know that we are only indexing by a relatively small number of elements from the _get_head_tail_server function, we can optimize based on our knowledge of the small array.

bmcdonald3:
This PR is a performance improvement for the `_get_head_tail_server`
function. Since we know that the index arrays will be small that
we are using to index in the DataFrameIndexingMsg function, we can
hardcode in that expectation to avoid some overhead that is helpful
only when doing large-array indexing.

To get around this overhead, a `smallIdx` argument was added to
pdarray string indexing by integer arrays indicating that the
indexing array is small and would not benefit from all of the
large array index overhead.

Instead of doing the regular indexing, the `smallIdx` branch will
localize each string using `lowLevelLocalizingSlice` before copying
it into the output array. This performs better than the naive
indexing approach taken when `CHPL_COMM=none` for distributed runs
for arrays over a certain threshold and significantly outperforms
the large array path.

This optimization could be extended to dynamically check the size of
the index array and then determine whether or not we want to do the
indexing optimized for large or small arrays, but pinning down those
thresholds can be difficult and machine-dependent.

Running 50 `_get_head_tail_server` string calls on 8
nodes of a cs-hdr:
String size | This branch | Naive copy | Master
----------- | ----------- | ---------- | ------
5-6         | 0.7960 s    | 0.7883 s   | 1.1169 s
50-51       | 0.8262 s    | 1.0042 s   | 1.2797 s
100-101     | 0.8736 s    | 1.2195 s   | 1.3165 s
200-201     | 0.8778 s    | 1.6184 s   | 1.3511 s
300-301     | 0.9518 s    | 1.9948 s   | 1.4371 s
1000-1001   | 1.363 s     | --         | 1.8706 s
2000-2001   | 1.8799 s    | --         | 2.3605 s

Running `dataframe.py` benchmark on 16 nodes of a cs-hdr:
This branch | Master
----------- | ---------
0.8991 s    | 1.0048 s

Closes: https://github.com/Bears-R-Us/arkouda/issues/1536
Part of: https://github.com/Bears-R-Us/arkouda/issues/1398

### DIFF
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -760,7 +760,7 @@ module SegmentedMsg {
   }
 
   proc segPdarrayIndex(objtype: string, args: [] string, 
-                                 st: borrowed SymTab): MsgTuple throws {
+                       st: borrowed SymTab, smallIdx=false): MsgTuple throws {
     var pn = Reflection.getRoutineName();
 
     // check to make sure symbols defined
@@ -781,7 +781,10 @@ module SegmentedMsg {
                 select gIV.dtype {
                     when DType.Int64 {
                         var iv = toSymEntry(gIV, int);
-                        var (newSegs, newVals) = strings[iv.a];
+                        // When smallIdx is set to true, some localization work will
+                        // be skipped and a localizing slice will instead be done,
+                        // which can perform better by avoiding those overhead costs.
+                        var (newSegs, newVals) = strings[iv.a, smallIdx=smallIdx];
                         var newStringsObj = getSegString(newSegs, newVals, st);
                         newStringsName = newStringsObj.name;
                         nBytes = newStringsObj.nBytes;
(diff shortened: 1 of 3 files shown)


Ethan-DeBandi99:
This looks like it is part of #1398. The performance had been improved already, but this furthers it and @ronawho and I reopened that issue for further improvement. Linking here so the connection is not lost.

src/SegmentedMsg.chpl
```diff
@@ -781,7 +781,10 @@ module SegmentedMsg {
                 select gIV.dtype {
                     when DType.Int64 {
```
pierce314159, line 782: is it worthwhile to update the `DType.UInt64` case too?

pierce314159 (approved):
Looks good! One question about updating the `uint` case

src/SegmentedMsg.chpl
bmcdonald3: I don't think that code path can be hit by the data frame indexing message case and I just ran a quick `writeln` test to make sure, so I don't think that would help this specific effort, but could be added if we choose to support this small array approach elsewhere.

joshmarshall1 (approved):

PR #1541, opened 2022-06-29
Closes #1540 - Adds `.objType` and `.dtype` to Categorical

### LINKED ISSUE #1374: Arkouda Categoricals do not have a .dtype attribute
Arrays and Dataframes containing Categorical objects do not support the .dtype attribute. While this is likely related to the lack of a corresponding underlying numpy datatype for Categorical objects, the lack of the dtype attribute causes difficulty, particularly when performing operations on ak.DataFrame objects when trying to determine operations appropriate for individual columns. 

Code to replicate: 

`ak.Categorical(ak.array(['see','felix','run'])).dtype`

produces output: 
```

<ipython-input-16-579dfd5cb717> in <module>
----> 1 ak.Categorical(ak.array(['see','felix','run'])).dtype

AttributeError: 'Categorical' object has no attribute 'dtype'
```

### LINKED ISSUE #1385: Standardize `.objtype` across Arkouda Classes
Should we standardize `.objtype`? It seems to vary quite a lot
```python
>>> ak.pdarray.objtype
# 'pdarray'
>>> ak.Strings.objtype
# 'str'
>>> ak.Categorical.objtype
# 'category'
```
with a lot classes not having an class attribute `objtype` at all (`Groupby`, `ArrayView`, `Dataframe`,...). 
___
It seems to me this is behaving like a slightly unpredictable alias for `.__name__`. So I think it makes sense to make a decorator, so we can straight up alias `__name__`
```python
def objtypedec(orig_cls):
    orig_cls.objtype = orig_cls.__name__
    return orig_cls
```
So in the arkouda classes we just have to do 
```python
from arkouda.decorators import objtypedec

@objtypedec
class myClass:
```
This will look like the following

```python
>>> ak.pdarray.objtype
# 'pdarray'
>>> ak.Strings.objtype
# 'Strings'
>>> ak.Categorical.objtype
# 'Categorical'
>>> ak.GroupBy.objtype
# 'GroupBy'
>>> ak.ArrayView.objtype
# 'ArrayView'
>>> ak.DataFrame.objtype
# 'DataFrame'
>>> ak.SegArray.objtype
# 'SegArray'
```

### LINKED ISSUE #1540: `.objType` & `.dtype` for Categorical
Part of #1539

Add `.objType='Categorical'` and `.dtype=np.str_` to `Categorical`.

This is to serve as a trial for the configuration specified in #1539

Ethan-DeBandi99:
Closes #1540 
Closes #1374
Closes #1385

Adds decorator to access object type using the class name.

Adds `.objType="Categorical"` property to Categorical
Adds `.dtype=np.str_` property to Categorical

References to the Categorical objType have been updated from `category` to `Categorical`


arkouda/categorical.py
```diff
@@ -137,6 +137,8 @@ def __init__(self, values, **kwargs) -> None:
         self.nlevels = self.categories.size
         self.ndim = self.codes.ndim
         self.shape = self.codes.shape
+        self.dtype = np.str_
+        self.objType = "Categorical"
```
reuster986, line 141: Let's change the attribute name to `objtype` (lowercase) to keep consistent with `pdarray`: 

https://github.com/Bears-R-Us/arkouda/blob/9b400e6560d1fccac3206f4f74f01ba0fc70758e/arkouda/pdarrayclass.py#L163

Also, I know you literally just implemented what I wrote in the issue, but I have some feedback for myself :)
* we may want to use the `str_` dtype object imported from `arkouda.dtypes` instead of `np.str_`, just to keep everthing coming from one place
* I like @pierce314159 's proposal to use `self.__class__.__name__` as the objtype instead of hard-coding the name

reuster986 (changes requested):

arkouda/categorical.py
Ethan-DeBandi99: Updated

reuster986: Thanks!  However, it looks like `self.objType` is still camel case (should be `self.objtype`)... or am I missing a commit?

reuster986 (approved):

pierce314159 (approved):
Looks good!

PR #1544, opened 2022-06-29
Recommend Chapel 1.26.0 and use it for CI testing

### LINKED ISSUE #1543: Upgrade to Chapel 1.26
Chapel 1.26 has been out for a while, but I'm late getting around to upgrading. 1.26 was mostly about improved GPU and `CHPL_COMM=ofi support`, neither of which are currently important to Arkouda so we weren't missing much, but I think it'd be good to upgrade in prep for 1.27 which includes some optimizations that do matter to arkouda.

ronawho:
Chapel 1.26 was mostly about new GPU features and better `CHPL_COMM=ofi`
support, neither of which are important to Arkouda currently, which is
largely why I haven't upgraded until now.

Note that this changes the documentation to "require" 1.26.0, but only
recommends it in the Makefile instead of making it an error. Arkouda
should continue to work with Chapel 1.24+ so existing users don't
necessarily have to upgrade, but I think it makes sense to recommend
that users reading the doc should use the latest version.

Note that 1.24 is pretty old at this point and we can probably drop some
support for some older versions, but I'll leave that for others.

Resolves #1543

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     container:
-      image: chapel/chapel:1.25.1
+      image: chapel/chapel:1.26.0
     steps:
     - uses: actions/checkout@v2
     - name: Install dependencies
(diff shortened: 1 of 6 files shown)


ronawho:
Note that I'm open to different phrasing that makes it more clear that older versions are still supported, but my thinking was that it's mostly new users reading docs so it's simpler to stick with 1 version to limit confusion.

Ethan-DeBandi99 (approved):
This looks good to me. Thanks for updating all the docs too!

pierce314159 (approved):
Looks good!

PR #1549, opened 2022-07-01
Closes #1548: Fix `interval_lookup` bug and enable multi-array input

### LINKED ISSUE #1548: Fix `interval_lookup` and update to allow multi-array input
There are two issues with `interval_lookup` and the related `search_intervals`: one is a bug having to do with the `assume_unique=` keyword arg, and the other is that support for multi-array input, which existed in akutil, somehow went away.

Reproducer for both:
```python
>>> starts = (ak.zeros(3), ak.array([0, 10, 20]))
>>> ends = (ak.zeros(3), ak.array([5, 15, 25]))
>>> vals = ak.arange(3)
>>> queries = (ak.zeros(3), ak.array([23, 13, 3]))
>>> expected = ak.array([2, 1, 0])
>>> result = ak.interval_lookup((starts, ends), vals, queries)
TypeError: search_intervals() got an unexpected keyword argument 'assume_unique'

# Bypassing interval_lookup and going directly to search_intervals
# This works in akutil
>>> inds = ak.search_intervals(vals, (starts, ends))
TypeError: arguments must be numeric arrays
```

For the second issue, my guess is that the arg validation just needs to be expanded to support sequences of numeric arrays (what I call multi-arrays).

pierce314159:
This PR (Closes #1548):
- Updates calls to `search_intervals` to not use `assume_unique` (this keyword arg was removed in #1494)
- Adds ability to pass multi-array input into `interval_lookup` and `search_intervals`
- Adds a test for the multi-array input

Note:
I'll admit this implementation is definitely not the most elegant. Since this is an important bug, I went with what seemed most straightforward, so I fully expect there's a lot that can be done to make this code more efficient and prettier. Stylistic suggestions are very welcome

### DIFF
--- a/arkouda/alignment.py
+++ b/arkouda/alignment.py
@@ -1,3 +1,4 @@
+import functools
 from warnings import warn
 
 import numpy as np  # type: ignore
@@ -11,7 +12,8 @@ from arkouda.groupbyclass import GroupBy, broadcast, unique
 from arkouda.numeric import where
 from arkouda.pdarrayclass import is_sorted, pdarray
 from arkouda.pdarraycreation import arange, full, ones, zeros
-from arkouda.pdarraysetops import argsort, concatenate, in1d
+from arkouda.pdarraysetops import concatenate, in1d
+from arkouda.sorting import argsort
 from arkouda.strings import Strings
 
 
@@ -238,7 +240,7 @@ def find(query, space):
     return spaceidx[i >= spacesize]
 
 
-def lookup(keys, values, arguments, fillvalue=-1, keys_from_unique=False):
+def lookup(keys, values, arguments, fillvalue=-1):
     """
     Apply the function defined by the mapping keys --> values to arguments.
 
@@ -254,9 +256,6 @@ def lookup(keys, values, arguments, fillvalue=-1, keys_from_unique=False):
         (or tuple of dtypes, for a sequence) as keys.
     fillvalue : scalar
         The default value to return for arguments not in keys.
-    keys_from_unique : bool
-        If True, keys are assumed to be the output of ak.unique, e.g. the
-        .unique_keys attribute of a GroupBy instance.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


arkouda/alignment.py
```diff
@@ -383,61 +383,139 @@ def search_intervals(vals, intervals):
...
+            or any([not isinstance(lo, pdarray) for lo in low])
+            or any([not isinstance(hi, pdarray) for hi in high])
+        ):
+            raise TypeError("All elements of Multi-array arguments must be pdarrays")
+
+        valsize = set(v.size for v in vals)
+        lowsize = set(lo.size for lo in low)
+        highsize = set(hi.size for hi in high)
+        if len(valsize) != 1 or len(lowsize) != 1 or len(highsize) != 1:
+            raise TypeError("Multi-array arguments must be non-empty and have equal-length arrays")
```
Ethan-DeBandi99, line 461: Do we really need everything to be the same size? Obviously, at index `i`, `vals[i].size == low[i].size == high[i].size` is required. But I am not sure there is a requirement for the various indexes to be the same size. Looking at the rest of the code, we are definitely relying on this, so I just wanted to check that we don't think there will be variance in sizes between elements.

Ethan-DeBandi99 (commented):
The code looks good to me on initial pass (I will probably go through one more time to review the logic a bit closer). I left 1 comment. I understand why we are checking that everything is the same size in the multi-array case, but wanted to validate that the actual use cases support these checks.

arkouda/alignment.py
pierce314159: good point! this is definitely an assumption i was making that I'm not sure is necessarily the case. I was thinking about a comment @reuster986 made about [`MultiArray`s having all arrays be the same size ](https://github.com/Bears-R-Us/arkouda/pull/1532#discussion_r906307677)

But I'm not sure, what's right here. @reuster986, do u know if it's reasonable to expect all arrays stored in an a multi-array to have the same size?

joshmarshall1 (approved):
All the logic looks good to me, As long as Bill doesn't request changes based on the conversation Ethan proposed, I approve.

arkouda/alignment.py
Ethan-DeBandi99: Based on @reuster986 input, I think this is good to go, resolving the thread.

Issue #1546, opened 2022-06-29
Prep for and upgrade Arkouda to Chapel 1.27


ronawho:
Chapel 1.27.0 will be out June 30. It contains some performance optimizations for scans that benefit Arkouda. Make any changes required to quiet deprecation warnings and upgrade to 1.27 once it's out.

PR #1553, opened 2022-07-01
Quiet deprecation warnings for Chapel 1.27

ronawho:
As part of Chapel's continued push to library stabilization and Chapel
2.0 there are a few methods that have changed or will change that result
in deprecation warnings for Arkouda. Adjust them here to be compatible
with 1.27 while maintaining compatible with previously supported
versions (currently 1.24)

Individual commits have more details, but the 3 main changes are:
 - Change `datetime.today()` to `datetime.now()`
 - Change `string.match(regex)` to `regex.match(string)`
 - Avoid calling `.low`/`.high` on strided ranges/domains

Part of https://github.com/Bears-R-Us/arkouda/issues/1546

### DIFF
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -739,7 +739,7 @@ module SegmentedMsg {
                 return new MsgTuple(errorMsg, MsgType.ERROR);
             }
             // TO DO: in the future, we will force the client to handle this
-            var slice: range(stridable=true) = convertPythonSliceToChapel(start, stop, stride);
+            var slice = convertPythonSliceToChapel(start, stop);
             // Compute the slice
             var (newSegs, newVals) = strings[slice];
             // Store the resulting offsets and bytes arrays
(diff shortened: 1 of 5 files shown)


src/SegmentedMsg.chpl
```diff
@@ -756,16 +756,12 @@ module SegmentedMsg {
...
-    else if (start <= stop) & (stride > 0) {slice = start..(stop-1) by stride;}
-    // BAD FORM start < stop and stride is negative
-    else {slice = 1..0;}
-    return slice;
+  proc convertPythonSliceToChapel(start:int, stop:int): range(stridable=false) {
+    if (start <= stop) {
+      return start..(stop-1);
+    } else {
+      return 1..0;
+    }
```
pierce314159, line 764: I was confused how this would be equivalent until i realized we force `stride` to be one before we actually call this function
```Chapel
if (stride != 1) { 
    var errorMsg = notImplementedError(pn, "stride != 1"); 
    smLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);      
    return new MsgTuple(errorMsg, MsgType.ERROR);
}
var slice = convertPythonSliceToChapel(start, stop);
```

pierce314159 (approved):
Looks good to me!

src/SegmentedMsg.chpl
ronawho: Yeah, probably should have included the comment in the PR description, but the commit message detailed this -- https://github.com/Bears-R-Us/arkouda/pull/1553/commits/a1ad94412799c260a098e576404f93c5386074b7

pierce314159: ah cool! i missed that!

PR #1554, opened 2022-07-05
Closes #1546: Recommend Chapel 1.27.0 and use it for CI testing

### LINKED ISSUE #1546: Prep for and upgrade Arkouda to Chapel 1.27

ronawho:
Chapel 1.27 includes scan optimizations and scans are used heavily in
Arkouda so it improves the performance of quite a few operations.

More details about the 1.27 release can be found at:
https://chapel.discourse.group/t/announcing-chapel-1-27-0/13627

Closes https://github.com/Bears-R-Us/arkouda/issues/1546

### DIFF
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -2,7 +2,7 @@
 
 The dependencies listed here have varying installation instructions depending upon your preferred operating system. Please use in the installation instructions provided in [INSTALL.md](INSTALL.md).
 
-- `Chapel 1.26.0 or later`
+- `Chapel 1.27.0 or later`
 - `cmake>=3.11.0`
 - `zeromq>=4.2.5`
 - `hdf5`
--- a/pydoc/setup/prerequisites.rst
+++ b/pydoc/setup/prerequisites.rst
@@ -7,7 +7,7 @@ Prerequisites
 *******************
 Chapel
 *******************
-(version 1.26.0 or greater)
+(version 1.27.0 or greater)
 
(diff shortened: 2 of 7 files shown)


joshmarshall1 (approved):

REQUIREMENTS.md
```diff
@@ -2,7 +2,7 @@
 
 The dependencies listed here have varying installation instructions depending upon your preferred operating system. Please use in the installation instructions provided in [INSTALL.md](INSTALL.md).
 
-- `Chapel 1.26.0 or later`
+- `Chapel 1.27.0 or later`
```
Ethan-DeBandi99, line 5: Are we not supporting 1.26.0 at all any further? I would update this to 

`Chapel 1.26.0 or later` (1.27.0 Recommended)

pydoc/setup/prerequisites.rst
```diff
@@ -7,7 +7,7 @@ Prerequisites
 *******************
 Chapel
 *******************
-(version 1.26.0 or greater)
+(version 1.27.0 or greater)
```
Ethan-DeBandi99, line 10: Again, are we moving to only support 1.27.0?

REQUIREMENTS.md
Ethan-DeBandi99: I completely agree with what you are saying here. I would prefer that everyone stick with the latest version, but I am not sure that is truly going to be a reality. I am fine with keeping it at `1.27.0`, but then I would remove the `or greater`. My thinking here is that we don't have a higher version and we only want them using `1.27.0`. 

This is a good point though, we are really only testing the current `Chapel` version. My thought would then be to say we only actively support that version. I don't think we can eliminate "support" for older versions though because there is often a delay in customers updating.

PR #1556, opened 2022-07-05
Rename `SegString.hash()` to `SegString.siphash()`

### LINKED ISSUE #1555: Fix Arkouda compatibility with Chapel main
Recent changes with override checking have broken builds on main.

ronawho:
`hash()` is currently a special method in Chapel and we recently added
some extra checking upstream that disallows using the name without the
`override` keyword. We haven't settled on the name `hash()` yet and are
considering special methods with names more unique as to not interfere
with user names, but until that's settled I think it's easiest to just
give the segmented string hash a different name to stay compatible.

Resolves #1555

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -175,7 +175,7 @@ module ArgSortMsg
     }
 
     proc incrementalArgSort(s: SegString, iv: [?aD] int): [] int throws {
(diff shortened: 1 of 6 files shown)


Ethan-DeBandi99 (approved):
Looks good

joshmarshall1 (approved):

Issue #1218, opened 2022-03-16
Extend `pdarray` setops to work for multiple arrays


Ethan-DeBandi99:
Write a PR that extends the ordinary set ops to work on multiple arrays, much like alignment.py:in1d_multi(a, b) is the multi-array analog of pdarraysetops.py:in1d(a, b). Stylistically, I think the logic of in1d_multi should be rolled into in1d, which should accept either singleton arrays or tuples/lists of arrays as arguments. Similar logic should suffice to make the other set operations work for multiple arrays.

Adds support for multiple arrays to the existing `pdarray` set operations:
- `union1d`
- `intersection1d`
- `setdiff1d`
- `setxor1d`

As an example, use `alignment.py:in1d_multi(a, b)` as a guide. It is the multi-array analog of `pdarraysetops.py:in1d(a, b)`.

@reuster986 mentioned a preference for the logic to be consolidated into a single function that takes in single arrays or tuples/lists of arrays as arguments.

Example utilization:

```python
def set_intx(cls, a, b):
    a_seg_inds = a.grouping.broadcast(arange(a.size))
    b_seg_inds = b.grouping.broadcast(arange(b.size))
    (new_seg_inds, new_values) = intersect1d((a_seg_inds, a.values), (b_seg_inds, b.values))
    g = ak.GroupBy(new_seg_inds)
    # TODO: Possibly add logic to insert empty segments where intersection is empty
    return SegArray(g.segments, new_values[g.permutation])
```

This issue is needed to make the work for #1197 (PR #1213) better performing. That PR has been converted to draft until this work is completed. Once completed, the PR will need to be modified to utilize the updated `pdarray` set functionality.

reuster986:
I spoke with @Ethan-DeBandi99 offline, but I'm re-opening because #1225 works on SegArrays, but the actual format of the input data should be different (my fault for being slow to review and not catching it before it was merged).

For multi-array setops, think of each input argument as a table (e.g. columns from a DataFrame), where the columns are arkouda arrays and each "row" is an element of the set. The set logic thus operates on rows, as follows:

```
>>> employeeFirstNames = ak.array(['Alice', 'Bob', 'Alice', 'Bob', 'Bob'])
>>> employeeLastNames = ak.array(['Jones', 'Smith', 'Baker', 'Smith', 'Miller'])
>>> employeeAges = ak.array([42, 35, 38, 56, 52])
[...]

PR #1213, opened 2022-03-15
Closes #1197 & #1245 `SegArray` Set Operations

Ethan-DeBandi99:
This PR (closes #1197 and closes #1245):

Adds the requests set operation functionality to `SegArray`. The following are implemented:
- `intersect()` computes the intersection
- `union()` computes the union
- `difference()` computes the difference 
- `xor()` computes the symmetric difference.

Maintaining this as a draft until benchmarks are added for the method used.

Unit tests have been added to `tests/segarray_test.py` to validate functionality. The tests also cover validation when segments are empty.

PR #1557, opened 2022-07-06
Closes #1495 - Move `in1dmulti` into `arkouda/pdarraysetops.py`

### LINKED ISSUE #1495: Move `in1dmulti` Into `arkouda/pdarraysetops.py`
`arkouda/pdarraysetops` has been updated so that other set operations (`interset1d`, `union1d`, `setxor1d`, and `setdiff1d`) are able to handle operations on `a` and `b` where both are `pdarray`s or when both are a sequence of `pdarray`s. This functionality exists for `in1d`, but is located in `arkouda/pdarraysetops`. The functionality of `in1dmulti` should into `in1d` to conform to the format of other setops. Additionally, we should add a deprecation warning to the function call in `arkouda/alignment.py` and make it an alias for `in1d` with the updated functionality.

Ethan-DeBandi99:
Closes #1495

Moves `in1dmulti` from `arkouda/alignment.py` to `arkouda/pdarraysetops.py`. `in1dmulti` has been renamed to `in1d`. 

The original `in1d` is renamed to `_in1d_single` so that we did not need to reconfigure the layout to handle single or multi `pdarray` operations.


reuster986 (changes requested):
Code looks good! But can we alias `in1dmulti` to `in1d` somewhere in the arkouda namespace for backwards compatibility?

pierce314159 (approved):

mhmerrill (changes requested):
I don't want to be a stick in the mud... but I think this deviates from NumPy `in1d` behavior. Verify that and if it does, have two entry points `in1d` which adheres to NumPy and `in1dmulti` which have the code as is.

Ethan-DeBandi99:
> I don't want to be a stick in the mud... but I think this deviates from NumPy `in1d` behavior. Verify that and if it does, have two entry points `in1d` which adheres to NumPy and `in1dmulti` which have the code as is.

I do not agree with this. Our other setop functionality also has single entry points for multiple array support. This is a case where I believe that adding the additional functionality under the single entry point makes sense. @pierce314159, @reuster986 - what do you both think? I will update it if that is what is desired, but based on the other configurations, this request does not make sense to me.

joshmarshall1 (approved):

pierce314159:
> > I don't want to be a stick in the mud... but I think this deviates from NumPy `in1d` behavior. Verify that and if it does, have two entry points `in1d` which adheres to NumPy and `in1dmulti` which have the code as is.
> 
> I do not agree with this. Our other setop functionality also has single entry points for multiple array support. This is a case where I believe that adding the additional functionality under the single entry point makes sense. @pierce314159, @reuster986 - what do you both think? I will update it if that is what is desired, but based on the other configurations, this request does not make sense to me.

It's true this change is in line with #1218, where we extended the pdarray setops to work with multiple arrays despite their numpy equivalents only handling single arrays ([np.union1d](https://numpy.org/doc/stable/reference/generated/numpy.union1d.html), [np.intersect1d](https://numpy.org/doc/stable/reference/generated/numpy.intersect1d.html))

Digging back a bit further, the push to incorporate `in1d_multi` into `in1d` seems to come from [this comment by bill](https://github.com/Bears-R-Us/arkouda/pull/1213#discussion_r827361233) (which motivated the above issue). Specifically it says:
> Stylistically, I think the logic of in1d_multi should be rolled into in1d, which should accept either singleton arrays or tuples/lists of arrays as arguments. Similar logic should suffice to make the other set operations work for multiple arrays.

jeichert60 (approved):

mhmerrill (approved):

Issue #1547, opened 2022-06-29
Drop support for older chapel versions


ronawho:
Historically Arkouda has supported the latest 2 chapel releases. However, at the moment I believe 1.24.x, 1.25.x, 1.26.x, and soon 1.27.x are supported. Dropping some of the older versions will allow us to remove some of the compatibly shims and reduce the amount of testing we'd need for #1545.

1.25.1 included some signifiant performance optimizations, so left to my own devices I'd require that as the minimum, but I'll defer to arkouda core devs.

Also note that in the past chapel released twice a year and we've recently switched to four times a year so it's not obvious to me how many chapel versions arkouda will want to support at any given time anymore. If you're still game for just supporting the last 2 versions I think that'd be great to minimize how many compatibly shims are required, but your call.

pierce314159:
I'm of the opinion that we should guarantee support for no more than 1 version back. To some extent, i believe we should only support the latest chapel version since each release tends to bring about pretty significant changes. We could just be on a delay (only update our minimum version/ add functionality specific to the latest version once we know our core users have gained access to the latest chapel version)

But I can def seem the reasoning for being backwards compatible for 1 version. I think anything more than that seems overly cautious

@Bears-R-Us/arkouda-core-dev, what do y'all think?

joshmarshall1:
I'm in agreement with Pierce. I believe one version back would be a good minimum supported version. I would like to add though that in the event of a major bug in a Chapel version, I think we should require the version that fixes that bug without support for the previous version (Once the core users have access to the latest version). In that case I don't think it makes sense to support versions with known major issues.

ronawho:
Offline I believe @mhmerrill indicated he'd at least like to keep support for chpl versions from the last ~6 months (so support the current and the past 2 releases with Chapel's new quarterly release schedule).

Given that I'm planning on closing this since we've removed support for 1.24 and the latest supported version is 1.25 where 1.25.1 was released in December of 2021.

For future reference https://github.com/Bears-R-Us/arkouda/pull/1568 is a good starting point for when we're ready to drop 1.25 support.

ronawho:
And I also want to note that I have no strong preference for how many versions you want to support. If you have more discussions internally and decide we can deprecate chpl versions faster just let me know. I was mostly just trying to wrap up some ongoing efforts before I go on vacation, not trying to rush or stifle any discussion here.

pierce314159:
[...]

PR #1560, opened 2022-07-07
Drop support for Chapel 1.24

ronawho:
Drop support for Chapel 1.24 and bump the minimum requirement to 1.25.
This allows us to remove some workarounds for broken json formatting and
remove some old transitory comments.

Part of #1547


ronawho:
This is all I found with a quick search, I didn't do anything too intelligent here other than replace the workaround code with the comment that was suggested nearby. I'm also not sure if there's other places that could be updated that didn't have such comments.

ronawho:
CI testing is hanging, so I think this broke something -- need to dig

pierce314159 (approved):

PR #1561, opened 2022-07-07
Closes #1161: Update `make test-chapel` to run unit tests

### LINKED ISSUE #1161: fix `make test-chapel` to not only compile arkouda-chapel unit tests but also run them
fix `make test-chapel` to not only compile arkouda-chapel unit tests but also run them

pierce314159:
This PR (Closes #1161):
- Updates `make test-chapel` to run unit tests in addition to compiling them
- This now runs the same commands the CI uses i.e.
```bash
python3 src/serverModuleGen.py ServerModules.cfg
start_test test
```
NOTE:
I'm not the best with `make`, so please feel free to suggest a different approach. I got this working using `.SILENT:` but that feels like overkill. I think we really just need `no-print-directory`, but I'm not sure how to get that to work. Without the silent, `make` will exclaim everytime it switches into the test directory and that interupts the compile statement

Tagging @mhmerrill, @bmcdonald3, and @ronawho as they are better with `make` than me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -429,7 +429,10 @@ $(eval $(call create_help_target,test-help,TEST_HELP_TEXT))
 test: test-python
 
 .PHONY: test-chapel
-test-chapel: $(TEST_TARGETS)
+.SILENT:
+test-chapel:
+	@echo $(shell python3 $(MODULE_GENERATION_SCRIPT) $(ARKOUDA_CONFIG_FILE));
+	start_test $(TEST_SOURCE_DIR)
 
 .PHONY: test-all
 test-all: test-python test-chapel


Makefile
```diff
@@ -429,7 +429,10 @@ $(eval $(call create_help_target,test-help,TEST_HELP_TEXT))
 test: test-python
 
 .PHONY: test-chapel
-test-chapel: $(TEST_TARGETS)
+.SILENT:
+test-chapel:
+	@echo $(shell python3 $(MODULE_GENERATION_SCRIPT) $(ARKOUDA_CONFIG_FILE));
```
ronawho, line 434: So far as I know the chapel tests don't need the config file, why is this generation needed?

ronawho: Looks like it was added in https://github.com/Bears-R-Us/arkouda/commit/e622dc5fbdc3c309ac8701f5f9929a95b53c4d47 -- @bmcdonald3 can you take an action item to look into whether this is needed.

Ethan-DeBandi99 (commented):
I don't think the line that @ronawho was saying may not be required will hurt anything by being there. I do want to note a few things to see if anyone else thinks they are an issue

- The .SILENT appears to be causing the compile statements to disappear during `make`. Not sure if this is an issue, but may make debugging more of a pain.
- I am seeing a failure during `test/UnitTestParquetCpp.chpl`. This is also happening for me on master. Good chance it may be due to my configuration, but wanted note in case anyone else is seeing it. Below is the output from the `.tmp`
```
timedexec: target program died with signal 11, without coredump
```

Ethan-DeBandi99:
I did a bit more digging on the issue I was seeing with the `test/UnitTestParquetCpp.chpl`. It appears as though there is a good chance that this may be due to the chipset. I know @jeichert60 has an arm64 Mac as well, so I am going to get him to run it and see if his has the same issue. If he does, I am tempted to say that there may be something between x86 and arm64 that are causing the error.

Ethan-DeBandi99:
> I did a bit more digging on the issue I was seeing with the `test/UnitTestParquetCpp.chpl`. It appears as though there is a good chance that this may be due to the chipset. I know @jeichert60 has an arm64 Mac as well, so I am going to get him to run it and see if his has the same issue. If he does, I am tempted to say that there may be something between x86 and arm64 that are causing the error.

On Friday, I had Jim run the test as well and it failed there. Thus it appears that this is an issue with the arm64 chipset when running the parquet test. I will try to find some time this week to look into it, and will probably want to touch base with @bmcdonald3 since he wrote our current Parquet code.

jeichert60 (approved):

joshmarshall1 (approved):

PR #1562, opened 2022-07-07
Closes #1545: Add portability testing to older chpl versions

### LINKED ISSUE #1545: Add chpl portability testing to CI
Historically Arkouda has supported the latest 2 chapel releases. However, at the moment I believe 1.24.x, 1.25.x, 1.26.x, and soon 1.27.x are supported. Given the variety of versions I think we should add CI testing to ensure the versions we want to work continue to do so.

I think we can also drop support for some older versions, which would allow us to remove some of the compatibility shims. 

Also note that in the past chapel released twice a year and we've recently switched to four times a year so it's not obvious to me how many chapel versions arkouda will want to support at any given time anymore.

ronawho:
Add portability testing for chpl 1.25/1.26. The documentation currently
"requires" 1.27, but we expect older versions to continue to work. Add
some testing to lock in that the versions we expect to work continue to
do so. I think we're planning to drop support for 1.25 soon, but wanted
to start with it for now.

Closes #1545

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -71,7 +71,7 @@ jobs:
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: ${{ matrix.python-version }}
+        python-version: ${{matrix.python-version}}
     - name: Install dependencies
       run: |
         apt-get update && apt-get install -y -V ca-certificates lsb-release wget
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
This looks good to me

pierce314159 (approved):
Looks good

joshmarshall1 (approved):

PR #1564, opened 2022-07-07
Closes #1168 - DataFrame Index Renaming

### LINKED ISSUE #1168: `ak.DataFrame` Index Renaming
`ak.DataFrame` currently supports renaming `Columns`. However, `pandas.DataFrame` also supports renaming of `indexes`. 

`pandas` uses `index` and `column` parameters to indicate which is to be updated. `ak.DataFrame` should be updated to replicate this functionality.

Ethan-DeBandi99:
Closes #1168 

Adds support for renaming indexes. This support currently requires that the replacement value is the same type as the original.

Adds testing for index renaming.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -930,13 +930,13 @@ class DataFrame(UserDict):
     @typechecked
     def _rename_column(
         self, mapper: Union[Callable, Dict], inplace: bool = False
-    ) -> Union[None, DataFrame]:
+    ) -> Optional[DataFrame]:
         """
         Rename columns within the dataframe
 
         Parameters
         ----------
-        mapper : callable or dict-like, Optional
+        mapper : callable or dict-like
             Function or dictionary mapping existing columns to new columns.
             Nonexistent names will not raise anerror.
         inplace: bool
@@ -983,13 +983,13 @@ class DataFrame(UserDict):
     @typechecked
     def _rename_index(
         self, mapper: Union[Callable, Dict], inplace: bool = False
-    ) -> Union[None, DataFrame]:
+    ) -> Optional[DataFrame]:
         """
-        Rename indexes withing the dataframe
+        Rename indexes within the dataframe
 
         Parameters
         ----------
-        mapper : callable or dict-like, Optional
+        mapper : callable or dict-like
             Function or dictionary mapping existing indexes to new indexes.
             Nonexistent names will not raise anerror.
         inplace: bool
@@ -1009,7 +1009,6 @@ class DataFrame(UserDict):
             the same type as the existing value.
         """
         obj = self if inplace else self.copy()
-        # TODO - does the replacement need to be the same type????
         if callable(mapper):
             for i in range(obj.index.size):
                 oldval = obj.index[i]
@@ -1036,9 +1035,9 @@ class DataFrame(UserDict):
         column: Optional[Union[Callable, Dict]] = None,
         axis: Union[str, int] = 0,
         inplace: bool = False,
-    ) -> Union[None, DataFrame]:
+    ) -> Optional[DataFrame]:
         """
-        Rename columns in-place according to a mapping.
+        Rename indexes or columns according to a mapping.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
This looks good!

few comments-- all docstrings

arkouda/dataframe.py
```diff
@@ -977,6 +980,141 @@ def rename(self, mapper: Union[Callable, dict], inplace: bool = False) -> Union[
...
+        """
+        obj = self if inplace else self.copy()
+        # TODO - does the replacement need to be the same type????
+        if callable(mapper):
+            for i in range(obj.index.size):
+                oldval = obj.index[i]
+                newval = mapper(oldval)
+                if type(oldval) != type(newval):
+                    raise TypeError("Replacement value must have the same type as the original value")
+                obj.index.values[obj.index.values == oldval] = newval
```
pierce314159, line 1019: i thought of this a second too late for my first review, but would something like this work here? it might break the `Nonexistent names will not raise an error.` part, i haven't thought about it too hard
```python
if callable(mapper):
    obj.index.values = mapper(obj.index.values)
```

Ethan-DeBandi99: I do not believe that we would be able to do something like this, because it relies on `mapper`  taking in a `pdarray` which is not necessarily a guarantee.

arkouda/dataframe.py
```diff
@@ -928,30 +928,33 @@ def update_size(self):
...
-        Rename columns in-place according to a mapping.
+        Rename columns within the dataframe
 
         Parameters
         ----------
-        mapper : callable or dict-like
-            Function or dictionary mapping existing column names to
-            new column names. Nonexistent names will not raise an
-            error.
+        mapper : callable or dict-like, Optional
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Updated

arkouda/dataframe.py
```diff
@@ -977,6 +980,141 @@ def rename(self, mapper: Union[Callable, dict], inplace: bool = False) -> Union[
             return obj
         return None
 
+    @typechecked
+    def _rename_index(
+        self, mapper: Union[Callable, Dict], inplace: bool = False
+    ) -> Union[None, DataFrame]:
+        """
+        Rename indexes withing the dataframe
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Corrected

arkouda/dataframe.py
```diff
@@ -928,30 +928,33 @@ def update_size(self):
             self._size = sizes.pop()
 
     @typechecked
-    def rename(self, mapper: Union[Callable, dict], inplace: bool = False) -> Union[None, DataFrame]:
+    def _rename_column(
+        self, mapper: Union[Callable, Dict], inplace: bool = False
+    ) -> Union[None, DataFrame]:
```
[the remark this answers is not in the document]
Ethan-DeBandi99: You are totally right. I updated this.

joshmarshall1 (approved):

arkouda/dataframe.py
```diff
@@ -977,6 +980,140 @@ def rename(self, mapper: Union[Callable, dict], inplace: bool = False) -> Union[
...
+        mapper: Optional[Union[Callable, Dict]] = None,
+        index: Optional[Union[Callable, Dict]] = None,
+        column: Optional[Union[Callable, Dict]] = None,
+        axis: Union[str, int] = 0,
+        inplace: bool = False,
+    ) -> Optional[DataFrame]:
+        """
+        Rename indexes or columns according to a mapping.
+
+        Parameters
```
jeichert60, line 1042: Just a nit but you might want to put the parameters in the docstring in the same sequence as in the function definition (i.e. column <-> index).

arkouda/dataframe.py
pierce314159: oh good point. I wrote this right before i left for the day. All i really did to try it out was pull down the PR, make the change, and check that the tests still passed. I figured that was enough to leave a comment for it to be looked into more

PR #1566, opened 2022-07-08
Closes #1565 - Add `Index.is_unique`

### LINKED ISSUE #1565: Add `ak.Index.is_unique`
We should add an `is_unique` attribute for arkouda `Index`s similar to [pandas.Index.is_unique](https://pandas.pydata.org/docs/reference/api/pandas.Index.is_unique.html). I think this will allow us to do less expensive computation when we know we have uniqueness

Ethan-DeBandi99:
Closes #1565 

Adds `ak.Index.is_unique`. Returns `True` if all values of the Index are unique, otherwise returns `False`.

Adds testing for `ak.Index.is_unique`.

Once this and PR #1168 are merged, we can update the renaming workflow for indexes to only update a specific value if the Index object is unique.

### DIFF
--- a/arkouda/index.py
+++ b/arkouda/index.py
@@ -302,6 +302,7 @@ class Index:
             ),
         )
 
+
 class MultiIndex(Index):
     def __init__(self, values):
         if not (isinstance(values, list) or isinstance(values, tuple)):
--- a/arkouda/index.py
+++ b/arkouda/index.py
@@ -79,6 +79,21 @@ class Index:
     def shape(self):
         return (self.size,)
 
+    @property
+    def is_unique(self):
+        """
+        Property indicating if all values in the index are unique
+        Returns
+        -------
+            bool - True if all values are unique, False otherwise.
+        """
+        g = GroupBy(self.values)
+        key, ct = g.count()
+        if (ct > 1).any():
+            return False
+        else:
+            return True
+
     @staticmethod
     def factory(index):
         t = type(index)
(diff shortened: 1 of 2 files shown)


jeichert60 (approved):

joshmarshall1 (approved):

arkouda/index.py
```diff
@@ -79,6 +79,21 @@ def index(self):
...
+        Returns
+        -------
+            bool - True if all values are unique, False otherwise.
+        """
+        g = GroupBy(self.values)
+        key, ct = g.count()
+        if (ct > 1).any():
+            return False
+        else:
+            return True
```
pierce314159, line 95: this is a nitpick but i think this is cleaner
```python
return (ct == 1).all()
```

tests/index_test.py
```diff
@@ -33,6 +33,13 @@ def test_multiindex_creation(self):
         with self.assertRaises(ValueError):
             idx = ak.MultiIndex([ak.arange(5), ak.arange(3)])
 
+    def test_is_unique(self):
+        i = ak.Index(ak.array([0, 1, 2]))
+        self.assertTrue(i.is_unique)
+
+        i = ak.Index(ak.array([0, 1, 1]))
+        self.assertTrue(i.is_unique is False)
```
pierce314159, line 41: go back and forth on which is more readable, but i just wanted to throw out there that `assertFalse` is an option

mhmerrill (approved):
looks good to me.

pierce314159 (commented):
A few things to look at

PR #1568, opened 2022-07-08
Drop support for Chapel 1.25

ronawho:
Drop support for Chapel 1.25 and bump the minimum requirement to 1.26.
This allows us to remove the Regex compatibility shim as well a
CType/ChplConfig shims.

Part of #1547


ronawho:
Opened as a draft (and closing) since I think we're not actually ready for this given the discussion in https://github.com/Bears-R-Us/arkouda/issues/1547, but I had the branch lying around so I wanted to get it somewhere more visible as a starting point for when 1.25 is dropped in the future.

PR #1569, opened 2022-07-09
 Closes #1519: Add strip method to Strings

### LINKED ISSUE #1519: Add `strip` method to `Strings`
Create a `strip` method for `Strings` which behaves like python's [`str.strip`](https://docs.python.org/3/library/stdtypes.html#str.strip) applied to every element of the SegString (i.e. removing the specified characters with whitespace as default). You should use chapel builtins for strings ([string.strip](https://chapel-lang.org/docs/builtins/String.html#String.string.strip)).

Note this is very similar to Issue #1518, but there's the additional complication of the sizes changing between the original SegString and the return SegString. There are examples of dealing with this in SegString (`sub` being a good, although more complicated, example) https://github.com/Bears-R-Us/arkouda/blob/c07ba2ef034ee824a21aafa6168d53b098ebd724/src/SegmentedArray.chpl#L618

jeichert60:
This PR (Closes #1519):

- Add ability to strip leading and trailing characters from original Strings
	- If a set of characters is provided, it is used as a list of leading and trailing characters to remove
	- If no set of characters is provided, the leading and trailing characters default to spaces

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -526,6 +526,51 @@ class Strings:
         )
 
     @typechecked
+    def strip(self, chars: Optional[Union[bytes, str_scalars]] = "") -> Strings:
+        """
+        Returns a new Strings object with all leading and trailing occurrences of characters contained
+        in chars removed. The chars argument is a string specifying the set of characters to be removed.
+        If omitted, the chars argument defaults to removing whitespace. The chars argument is not a
+        prefix or suffix; rather, all combinations of its values are stripped.
+
+        Parameters
+        ----------
+        chars : the set of characters to be removed
+
+        Returns
+        -------
+        Strings
+            Strings object with the leading and trailing characters matching the set of characters in
+            the chars argument removed
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there is a server-side error thrown
+
+        Examples
+        --------
+        >>> strings = ak.array(['Strings ', '  StringS  ', 'StringS   '])
+        >>> s = strings.strip()
+        >>> s
+        array(['Strings', 'StringS', 'StringS'])
+
+        >>> strings = ak.array(['Strings 1', '1 StringS  ', '  1StringS  12 '])
+        >>> s = strings.strip(' 12')
+        >>> s
+        array(['Strings', 'StringS', 'StringS'])
+        """
+        # The following "if" block is a hack to get around the limitation of sending arguments where
+        # the first character of the argument is a space. Ordinarily, this would result in the space not
+        # being part of the argument.
+        # TODO - Remove this "if" block when we go to JSON for message passing.
+        if isinstance(chars, bytes):
+            chars = chars.decode()
+        args = "{} {} {}".format(self.objtype, self.entry.name, chars,)
+        rep_msg = generic_msg(cmd="segmentedStrip", args=args)
+        return Strings.from_return_msg(cast(str, rep_msg))
+
+    @typechecked
     def cached_regex_patterns(self) -> List:
         """
         Returns the regex patterns for which Match objects have been cached
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -560,10 +560,6 @@ class Strings:
         >>> s
         array(['Strings', 'StringS', 'StringS'])
         """
-        # The following "if" block is a hack to get around the limitation of sending arguments where
-        # the first character of the argument is a space. Ordinarily, this would result in the space not
-        # being part of the argument.
-        # TODO - Remove this "if" block when we go to JSON for message passing.
         if isinstance(chars, bytes):
             chars = chars.decode()
         args = "{} {} {}".format(self.objtype, self.entry.name, chars,)
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -476,6 +476,31 @@ module SegmentedMsg {
     return new MsgTuple(repMsg, MsgType.NORMAL);
   }
 
+  proc segmentedStripMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+    var pn = Reflection.getRoutineName();
+    var repMsg: string;
+
+    var (objtype, name, chars) = payload.splitMsgToTuple(" ", 3);
+
+    // check to make sure symbols defined
+    st.checkTable(name);
+
+    select (objtype) {
+      when ("str") {
+        var strings = getSegString(name, st);
+        var (off, val) = strings.strip(chars);
+        var retString = getSegString(off, val, st);
+        repMsg = "created " + st.attrib(retString.name) + "+created bytes.size %t".format(retString.nBytes);
+        return new MsgTuple(repMsg, MsgType.NORMAL);
+      }
+      otherwise {
+          var errorMsg = notImplementedError(pn, "%s".format(objtype));
+          smLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+          return new MsgTuple(errorMsg, MsgType.ERROR);
+      }
+    }
+  }
+
   proc createPeelSymEntries(lo, lv, ro, rv, st: borrowed SymTab) throws {
     var leftEntry = getSegString(lo, lv, st);
     var rightEntry = getSegString(ro, rv, st);
@@ -1005,6 +1030,7 @@ module SegmentedMsg {
     registerFunction("segmentedFindAll", segmentedFindAllMsg, getModuleName());
     registerFunction("segmentedPeel", segmentedPeelMsg, getModuleName());
     registerFunction("segmentedSub", segmentedSubMsg, getModuleName());
+    registerFunction("segmentedStrip", segmentedStripMsg, getModuleName());
     registerFunction("segmentedIndex", segmentedIndexMsg, getModuleName());
     registerFunction("segmentedBinopvv", segBinopvvMsg, getModuleName());
     registerFunction("segmentedBinopvs", segBinopvsMsg, getModuleName());
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -719,6 +719,54 @@ module SegmentedString {
     }
 
     /*
+      Strip out all of the leading and trailing characters of each element of a segstring that are
+      called out in the "chars" argument.
+
+      :arg chars: the set of characters to be removed
+      :type chars: string
+
+      :returns: Strings – substrings with stripped characters from the original string and the offsets into those substrings
+    */
+    
+    proc strip(chars: string) throws {
+      ref origOffsets = this.offsets.a;
+      ref origVals = this.values.a;
+      const lengths = this.getLengths();
+
+      saLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                                             "chars: %s - origOffsets: %t - origVals: %t"
+                                             .format(chars, origOffsets, origVals:bytes));
+
+      var replacedLens: [this.offsets.aD] int;
+
+      forall (off, len, rlen) in zip(origOffsets, lengths, replacedLens) {
+        if chars.isEmpty() {
+          rlen = interpretAsBytes(origVals, off..#len).strip().size + 1;
+        } else {
+          rlen = interpretAsBytes(origVals, off..#len).strip(chars:bytes).size + 1;
+        }
+      }
+      var retVals: [makeDistDom(+ reduce replacedLens)] uint(8);
+      var retOffs = (+ scan replacedLens) - replacedLens;
+
+      forall (off, len, roff) in zip(origOffsets, lengths, retOffs) with (var valAgg = newDstAggregator(uint(8))) {
+        var i = 0;
+        if chars.isEmpty() {
+          for b in interpretAsBytes(origVals, off..#len).strip() {
+            valAgg.copy(retVals[roff+i], b:uint(8));
+            i += 1;
+          }
+        } else {
+          for b in interpretAsBytes(origVals, off..#len).strip(chars:bytes) {
+            valAgg.copy(retVals[roff+i], b:uint(8));
+            i += 1;
+          }
+        }
+      }
+      return (retOffs, retVals);
+    }
+
+    /*
       Returns list of bools where index i indicates whether the regular expression, pattern, matched string i of the SegString
 
       Note: the regular expression engine used, re2, does not support lookahead/lookbehind
--- a/tests/string_test.py
+++ b/tests/string_test.py
@@ -540,6 +540,42 @@ class StringTest(ArkoudaTest):
         lengths = s1.get_lengths()
         self.assertTrue((ak.array([3, 3, 5, 4, 4]) == lengths).all())
 
+    def test_strip(self):
+        s = ak.array([" Jim1", "John1   ", "Steve1 2"])
+        strip = s.strip(" 12")
+        expected = ak.array(["Jim", "John", "Steve"])
+        self.assertListEqual(strip.to_ndarray().tolist(), expected.to_ndarray().tolist())
+
+        s = ak.array([" Jim1", "John1   ", "Steve1 2"])
+        strip = s.strip("12 ")
+        expected = ak.array(["Jim", "John", "Steve"])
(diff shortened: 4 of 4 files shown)


arkouda/strings.py
```diff
@@ -525,6 +525,50 @@ def is_title(self) -> pdarray:
...
+    @typechecked
+    def strip(self, pattern: Union[bytes, str_scalars] = " ") -> Strings:
+        """
+        Returns a new Strings with all leading and trailing spaces (or set of given characters)
+        stripped from the original.
+
+        Returns
+        -------
+        Strings
+            Strings with all leading and trailing spaces (or set of given characters) stripped from
```
Ethan-DeBandi99, line 537: This should be updated to something like `Strings  object with all leading and trailing values matching pattern removed."

arkouda/strings.py
```diff
@@ -525,6 +525,50 @@ def is_title(self) -> pdarray:
...
+            the original
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there is a server-side error thrown
+
+        See Also
+        --------
+        Strings.to_upper
```
Ethan-DeBandi99, line 547: I am not sure to_upper needs to be referenced here.

arkouda/strings.py
```diff
@@ -525,6 +525,50 @@ def is_title(self) -> pdarray:
...
+        >>> strings = ak.array(['Strings 1', '1 StringS  ', '  1StringS  12 '])
+        >>> strings.strip(' 12')
+        >>> strings
+        array(['Strings', 'StringS', 'StringS '])
+        """
+        if isinstance(pattern, bytes):
+            pattern = pattern.decode()
+        contains_space = " " in pattern
+        if contains_space:
+            pattern = pattern.replace(" ", "")
```
Ethan-DeBandi99, line 565: Can we add a note here so that we remember to update this when we figure out JSON on the server? I just want to make sure we have it tagged for the update.

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
...
+                                             "pattern: %s - origOffsets: %t - origVals: %t"
+                                             .format(pattern, origOffsets, origVals:bytes));
+
+      var replacedLens: [this.offsets.aD] int;
+      var newPattern: string;
+
+      // Pattern contains only whitespace characters or is empty
+      if containsSpace == true {
+        newPattern = pattern + " ";
+      }
```
Ethan-DeBandi99, line 755: Again, this is going to be an issue, because we remove all spaces, not just trailing ones. There are a few ways you could get around this with the message parsing, or you could record the position of the spaces, the message parsing will be much easier though.

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
...
+
+      saLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),"pattern: (%t)".format(newPattern));
+
+      forall (off, len, rlen) in zip(origOffsets, lengths, replacedLens) with (var valAgg = newDstAggregator(uint(8))) {
```
Ethan-DeBandi99, line 759: `with (var valAgg = newDstAggregator(uint(8))` is not needed here because you do not use the `valAgg` until the next loop.

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
...
+      //   forall (off, len, roff) in zip(origOffsets, lengths, retOffs) with (var valAgg = newDstAggregator(uint(8))) {
+      //     var i = 0;
+      //     for b in interpretAsBytes(origVals, off..#len).strip(pattern:bytes) {
+      //       valAgg.copy(retVals[roff+i], b:uint(8));
+      //       i += 1;
+      //     }
+      //   }
+      //   return (retOffs, retVals);
+    }
+
```
Ethan-DeBandi99, line 791: If we aren't using this, it should be removed.

src/SegmentedMsg.chpl
```diff
@@ -489,6 +489,38 @@ module SegmentedMsg {
     return new MsgTuple(repMsg, MsgType.NORMAL);
   }
 
+  proc segmentedStripMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+    var pn = Reflection.getRoutineName();
+    var repMsg: string;
+
+    var (subcmd, objtype, name, containsSpace, pattern) = payload.splitMsgToTuple(5);
```
Ethan-DeBandi99, line 496: It would probably be ideal to update the parsing here so that pattern can include spaces. You can limit the number of splits done on payload to do this. I think this would be better than adding the check for contains space.

arkouda/strings.py
```diff
@@ -525,6 +525,50 @@ def is_title(self) -> pdarray:
             generic_msg(cmd="checkChars", args=f"isTitle {self.objtype} {self.entry.name}")
         )
 
+    @typechecked
+    def strip(self, pattern: Union[bytes, str_scalars] = " ") -> Strings:
```
Ethan-DeBandi99, line 529: The parameters should be listed in the docstring. We need to note that pattern is treated as a set of chars and not as a string.

Ethan-DeBandi99 (changes requested):
Few things to clean up. One thing I would like to look at is we should be able to handle the a space in the pattern with proper message parsing. I think that simplifies/removes a few checks for us.

arkouda/strings.py
```diff
@@ -525,6 +525,50 @@ def is_title(self) -> pdarray:
...
+        --------
+        >>> strings = ak.array(['Strings ', '  StringS  ', 'StringS   '])
+        >>> strings.strip()
+        >>> strings
+        array(['Strings', 'StringS', 'StringS '])
+
+        >>> strings = ak.array(['Strings 1', '1 StringS  ', '  1StringS  12 '])
+        >>> strings.strip(' 12')
+        >>> strings
+        array(['Strings', 'StringS', 'StringS '])
```
joshmarshall1, line 559: the third element in your results here both still have a trailing space, despite space being part of what's being stripped. Shouldn't these results contain no remaining trailing spaces?

joshmarshall1 (approved):
Overall looks good, only one comment related to your example results

src/SegmentedMsg.chpl
pierce314159: oh yeah ethan is right, this is the way to do it

The only thing i don't love about this approach is if we ever add additional parameters we have to remember that `pattern` must remain the last parameter in order for this to keep working (i lost a day debugging this with another function early in my arkouda development). So we should add a comment on the client and server saying that `pattern` should be always be the last parameter since it can contain spaces

arkouda/strings.py
pierce314159: I agree this could use some rewording. I think in cases like this where we are mimicing an existing function, it's often useful to look to their docstring for inspiration. [str.strip](https://docs.python.org/3/library/stdtypes.html#str.strip) says:
>Return a copy of the string with the leading and trailing characters removed. The chars argument is a string specifying the set of characters to be removed. If omitted or None, the chars argument defaults to removing whitespace. The chars argument is not a prefix or suffix; rather, all combinations of its values are stripped:

src/SegmentedArray.chpl
pierce314159: I don't think it should matter because `pattern` is the set of characters being stripped, so position/multiplicity of the characters shouldn't change anything

arkouda/strings.py
pierce314159: This is a bit of a nitpick, but i feel like `pattern` should be renamed to `chars` since that is the name of the parameter for both the python and chapel version of `strip`

pierce314159 (changes requested):
See in line comments for more details about why, but the biggest issues that need to get addressed
- `strip` won't work if the provided character set doesn't contain spaces
- `strip` with no parameters (default) only removes spaces, not all whitespace

I pulled down the PR to verify this and result for some examples are below
```python
# first issue, `strip` won't work if the character set doesn't contain spaces
>>> abc_list = ['abcStringsbac', 'cabStringScc', 'bbabStringS abc']
>>> ak_abc = ak.array(abc_list)
# ak.strip output
>>> ak_abc.strip('abc')
array(['abcStringsbac', 'cabStringScc', 'bbabStringS abc'])
# expected output (str.strip applied to each string)
>>> [s.strip('abc') for s in abc_list]
['Strings', 'StringS', 'StringS ']

# second issue, `strip` with no parameters (default) only removes spaces, not all whitespace
>>> whitespace_list = ['\nStrings ', ' \n StringS \r', ' \t   StringS \r\t ']
>>> ak_ws = ak.array(whitespace_list)
# ak.strip output
>>> ak_ws.strip()
array(['nStrings', 'n StringS u000D', 'u0009   StringS u000Du0009'])
# expected output (str.strip applied to each string)
>>> [s.strip() for s in whitespace_list]
['Strings', 'StringS', 'StringS']
```

arkouda/strings.py
jeichert60: Revised to align with your suggestions.

arkouda/strings.py
jeichert60: Yep. Cut and paste leftover.

arkouda/strings.py
jeichert60: Added a TODO note.

src/SegmentedArray.chpl
jeichert60: I agree with Pierce that position doesn't matter. Also, I'm not sure what you mean by "remove all spaces" but if there are spaces inside the original string that are bordered by non-matching characters then they are not removed. However, I changed the Python code to remove the "containsSpace" flag.

src/SegmentedArray.chpl
jeichert60: Yep. Removed.

src/SegmentedArray.chpl
jeichert60: Removed

src/SegmentedMsg.chpl
jeichert60: This doesn't work when the first character of the argument is a space, so I changed the python code to change  the position of the space to the last character in the argument if it was originally in the first position.

arkouda/strings.py
jeichert60: Revised the examples to all be out of place.

arkouda/strings.py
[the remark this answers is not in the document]
jeichert60: Setting the default to "" so that all whitespace will be stripped.

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
       return (subbedOffsets, subbedVals, numReplacements);
     }
 
+    /*
+      Strip out all of the leading and trailing characters of each element of a segstring that are
+      called out in the "pattern" argument.
+
+      :arg pattern: pattern used to find matches of the leading and trailing characters in the segstring
+      :type pattern: string
```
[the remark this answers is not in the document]
jeichert60: Replaced pattern with characters along with replacement of the "pattern" argument with "chars".

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
...
+
+      :arg containsSpace: the string to replace pattern matches with
+      :type replStr: string
+
+      :arg initCount: If count is nonzero, at most count splits occur. If zero, substitute all occurences of pattern
+      :type initCount: int
```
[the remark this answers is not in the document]
jeichert60: Yep. It has been removed.

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
...
+
+      :arg containsSpace: If True, appends a space character to the pattern
+      :type containsSpace: bool
```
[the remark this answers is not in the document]
jeichert60: Replaced with your comment.

src/SegmentedArray.chpl
```diff
@@ -717,6 +717,78 @@ module SegmentedArray {
...
+
+      :returns: Strings – substrings with stripped characters from the original string and the offsets into those substrings
+    */
+
+    proc strip(pattern: string, containsSpace: bool) throws {
+      checkCompile(pattern);
```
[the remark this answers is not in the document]
jeichert60: Removed

src/SegmentedArray.chpl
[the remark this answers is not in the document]
jeichert60: Not using containsSpace anymore.

src/SegmentedMsg.chpl
pierce314159: We need to investigate the first character being a space issue further to see if any other methods are affected, and update them accordingly (or at least denote it with a comment). I think the right answer for all this is to increase our efforts in moving our message processing to JSON

arkouda/strings.py
```diff
@@ -525,6 +525,55 @@ def is_title(self) -> pdarray:
             generic_msg(cmd="checkChars", args=f"isTitle {self.objtype} {self.entry.name}")
         )
 
+    @typechecked
+    def strip(self, chars: Optional[Union[bytes, str_scalars]] = "") -> Strings:
+        """
+        Returns a new Strings object with all leading and trailing characters removed. The chars
```
pierce314159, line 531: I know this based on the `str.strip` doc strings, but i think we should be explicit that it's the characters in `chars`

`Returns a new Strings object with all leading and trailing occurrences of characters contained in chars removed`

src/SegmentedMsg.chpl
```diff
@@ -489,6 +489,31 @@ module SegmentedMsg {
     return new MsgTuple(repMsg, MsgType.NORMAL);
   }
 
+  proc segmentedStripMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+    var pn = Reflection.getRoutineName();
+    var repMsg: string;
+
+    var (subcmd, objtype, name, chars) = payload.splitMsgToTuple(4);
```
pierce314159, line 496: we don't really need the `subcmd` since it's never used

pierce314159 (commented):
Couple minor things but the logic looks good to me!

Also some small things from last review still need to be addressed:
 - the PR description repeats the bullet `Add ability to strip leading and trailing characters from original Strings`
 - I think you need to rerun black on `strings.py`

pierce314159 (approved):
The 2 things from my previous reviews are smaller and shouldn't hold up this PR but i'll list them here again:
- The PR description repeats the bullet `Add ability to strip leading and trailing characters from original Strings`
- You need to rerun black on `strings.py`

Other than that, it looks good! All of my in-line feedback has been addressed

src/SegmentedArray.chpl
Ethan-DeBandi99: I left this comment by mistake. I had `remove` functionality in my head, not `strip` that is my bad. Sorry!

src/SegmentedMsg.chpl
Ethan-DeBandi99: I believe I tracked down the issue with the first char being a space. Per the Chapel docs on `.split`, [here](https://chapel-lang.org/docs/builtins/String.html?highlight=split#String.string.split), the split is being done on runs of whitespace when only the number of chunks is supplied. However, if you look at the function allowing the separation string to be provided, [here](https://github.com/Bears-R-Us/arkouda/blob/40e16f155a67e006c33fe0f44100bcb878edd443/src/ServerConfig.chpl#L240), you can specify the single space as the separator. 

As a result, I think we should update the message split to something like:
```
var (objtype, name, chars) = payload.splitMsgToTuple(" ", 3);
```

This will then allow for 
```
        if chars:
            chars = chars[1:] + " " if chars[0] == " " else chars
```
to be removed.

Ethan-DeBandi99 (changes requested):
I left details of the changes on my previous request to update the message parsing. Please update it to use the alternate version of `splitMsgToTuple` that allows the separator to be specified.

Ethan-DeBandi99 (approved):

src/SegmentedMsg.chpl
Ethan-DeBandi99: Updated msg split and removed the if.

PR #1579, opened 2022-07-12
DataFrame display perf experiments

ronawho:
Just wanted to capture a branch I've been using to explore https://github.com/Bears-R-Us/arkouda/issues/1398. This adds server timers to parts of `_get_head_tail_server` to see where the time is going. Note that the benchmark was updated to just time `ak.Strings` since that is the slow operation (so is `ak.Categorical`, but the code paths are roughly the same between the two)


ronawho:
Closing, just wanted someplace less ephemeral than a branch to capture this exploration.

PR #1582, opened 2022-07-12
Optimize string indexing by removing extra interior domain creations

bmcdonald3:
In pdarray indexing previously, there were 3 calls to `interior`
on the domain of the array that was causing overhead from creating
the additional domains from those calls. This PR instead just loops
over the entire domain to avoid needing to create those extra
domains.

This impact is most beneficial for small problem sizes where
creation overheads are more pronounced, but results in a slight
performance improvement for all cases.

String gather benchmark on 16-node-cs-hdr:
master        | this branch   |
------------- | ------------- |
23.20 GiB/sec | 24.74 GiB/sec |

### DIFF
--- a/src/SegmentedArray.chpl
+++ b/src/SegmentedArray.chpl
@@ -269,10 +269,11 @@ module SegmentedArray {
         srcIdx = 1;
         var diffs: [D] int;
         diffs[D.low] = left[D.low]; // first offset is not affected by scan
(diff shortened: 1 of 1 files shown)


ronawho (approved):
Beyond the perf improvements from getting rid of distributed domain creation and distributed array slicing, I personally find this code easier to read (I always have to test to remember what interior does)

PR #1592, opened 2022-07-15
Closes #1591 - add `client_dtype_test` to `pytest.ini`

### LINKED ISSUE #1591: Add `client_dtypes_test` to `pytest.ini` and issue with `test_is_ipv6`
While working #1531, I ran into an failures with `test_is_ipv6` before i made any changes. I wasn't seeing these with `make test` and was confused until I realized that it's not being run by `make test`. We should investigate `test_is_ipv6` and add `client_dtypes_test` to `pytest.ini` (and any others that are missing)

Ethan-DeBandi99:
Closes #1591 

Updates the `test_is_ipv6` to use proper types.

Adds `self.name = None` in order to silence warnings during testing when the objects cannot be created.

Adds `client_dtypes_test` to `pytest.ini`. Reordered tests so that the order matches the order of the files so it is easier to locate things.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -37,6 +37,7 @@ testpaths =
     tests/setops_test.py
     tests/sort_test.py
     tests/string_test.py
+    tests/summarization_test.py
     tests/util_test.py
     tests/where_test.py
 norecursedirs = .git dist build *egg* tests/deprecated/*
--- a/pytest.ini
+++ b/pytest.ini
@@ -7,24 +7,27 @@ testpaths =
     tests/bitops_test.py
     tests/categorical_test.py
     tests/check.py
+    tests/client_dtypes_test.py
     tests/client_test.py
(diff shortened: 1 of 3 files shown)


Ethan-DeBandi99:
Converting to draft because the error @pierce314159 experienced is occurring now in the CI.

pytest.ini
```diff
@@ -34,9 +37,8 @@ testpaths =
     tests/setops_test.py
     tests/sort_test.py
     tests/string_test.py
+    tests/util_test.py
```
pierce314159, line 40: it seems we're still missing `summarization_test.py` which should come just before this one

pierce314159 (commented):
Looks good other than one test that's still missing from `pytest.ini`

mhmerrill (approved):
looks ok to me.

PR #1595, opened 2022-07-15
Closes #1594: Add -s to make commands in compopts and remove .SILENT

### LINKED ISSUE #1594: Remove SILENT from Makefile
The SILENT in the Makefile is preventing commands from displaying their output and should be removed.

bmcdonald3:
This PR remove the `.SILENT` line from the Makefile, which
was preventing the output of `make` commands from displaying
and instead adds an explicit `-s` to the `make` commands that
were being executed in the `test/COMPOPTS` directory, which was
causing problems in certain environments.

Closes https://github.com/Bears-R-Us/arkouda/issues/1594

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -433,7 +433,6 @@ $(eval $(call create_help_target,test-help,TEST_HELP_TEXT))
 test: test-python
 
 .PHONY: test-chapel
-.SILENT:
 test-chapel:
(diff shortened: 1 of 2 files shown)


ronawho (approved):

pierce314159 (approved):
Thanks Ben! This works much better than the `.SILENT`!

PR #1599, opened 2022-07-16
Closes #1598 Interval lookup improvements

### LINKED ISSUE #1598: `search_intervals()` enhancements
I took too long to review #1549 , which was excellent work by @pierce314159 , but I wanted to extend it a bit to handle overlapping intervals and tiebreaks, so I am about to create a PR for that.

reuster986:
This PR updates `search_intervals()` to handle potentially overlapping intervals and without requiring the arguments to be passed in sorted order. The function now uses a `tiebreak` array to choose which interval to return when a value is contained in multiple intervals. 

This PR also updates the CI and adds a detailed test for representative cases of a 2-d interval search problem. The 4 rectangles (2-d intervals) pictured here demonstrate three classes of relationships between multi-dimensional intervals (hyperslabs):
1. Nested (B is a proper subset of A)
2. Intersecting (A and C overlap but neither is a subset of the other)
3. Disjoint (A and D do not intersect)

The test points explore each region of the diagram and the returned interval under different tiebreaks. Circles are hits (annotated with correct rectangle) and Xs are misses. Dots are the vertices that define each rectangle. For tiebreaks, "first" means return the first matching rectangle, "smallest" means the one with the least area.

IMG_20220716_160206183

### DIFF
--- a/arkouda/alignment.py
+++ b/arkouda/alignment.py
@@ -278,20 +278,23 @@ def in1d_intervals(vals, intervals, symmetric=False):
         return found
 
 
-def search_intervals(vals, intervals):
+def search_intervals(vals, intervals, tiebreak=None):
     """
-    Given an array of query vals and non-overlapping, half-open (pythonic)
-    intervals, return the index of the interval containing each query value,
-    or -1 if not present in any interval.
+    Given an array of query vals and half-open (pythonic) intervals, return the index of the 
+    best (see tiebreak) interval containing each query value, or -1 if not present in any 
+    interval.
 
     Parameters
     ----------
     vals : (sequence of) pdarray(int, uint, float)
         Values to search for in intervals. If multiple arrays, each "row" is an item.
     intervals : 2-tuple of (sequences of) pdarrays
-        Non-overlapping, half-open intervals, as a tuple of
-        (lower_bounds_inclusive, upper_bounds_exclusive)
-        Must have same shape/dtype as vals.
+        Half-open intervals, as a tuple of (lower_bounds_inclusive, upper_bounds_exclusive).
+        Must have same dtype(s) as vals.
+    tiebreak : (optional) pdarray, numeric
+        When a value is present in more than one interval, the interval with the
+        lowest tiebreak value will be chosen. If no tiebreak is given, the
+        first containing interval will be chosen.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


arkouda/alignment.py
```diff
@@ -278,20 +278,23 @@ def in1d_intervals(vals, intervals, symmetric=False):
...
-        Must have same shape/dtype as vals.
+        Half-open intervals, as a tuple of (lower_bounds_inclusive, upper_bounds_exclusive).
+        Must have same dtype(s) as vals.
+    tiebreak : (optional) pdarray, numeric
+        When a value is present in more than one interval, the interval with the
+        lowest tiebreak value will be chosen. If no tiebreak is given, the
+        first containing interval will be chosen.
 
     Returns
     -------
```
pierce314159, line 300: would it be worthwhile to add an example demonstrating how this works with and without `tiebreak`? I think a slimmed down version of the test would be great

arkouda/alignment.py
```diff
@@ -363,10 +365,14 @@ def search_intervals(vals, intervals):
             matchintervalidx = broadcast(segs, hitidx, matches.size)
             # make sure not to include any of the bounds themselves
             validmatch = (matches >= low.size) & (matches < boundary)
-            # indices of unique values found (translated from concat keys)
-            uvalidx = matches[validmatch] - low.size
+            matchintervalidx = matchintervalidx[validmatch]
+            # indices of values found (translated from concat keys)
+            validx = matches[validmatch] - low.size
```
pierce314159, line 370: could we rename these to use `_` while we're in here. Most of them are fairly easy to parse out but my brain gets confused especially on `matchintervalidx`. I know it's meant to be read `match_interval_idx` but sometimes i read it as `match_inter_valid_x` because of `valid_match` right beside it

I just realized it could also be `match_inter_val_idx`. I never actually read it like that, but I could see someone getting it mixed up that way since `val_idx` is also pretty close by

joshmarshall1 (approved):

arkouda/alignment.py
```diff
@@ -425,43 +425,62 @@ def search_intervals(vals, intervals):
...
+            # a true hit happens when a value is contained in all of an interval's 1-d projections
+            isahit = byvalinterval.count()[1] == len(low)
+            # indices of the true hits and their containing intervals
+            valhits, intervalhits = [x[isahit] for x in byvalinterval.unique_keys]
+            # a value might be found in more than one interval, so we need to break ties
+            byval = GroupBy(valhits)
+            hitvalidx, tmpidx = byval.argmin(tiebreak[intervalhits])
+            winninginterval = intervalhits[tmpidx]
+            # set index of best containing interval for values that were found
+            containinginterval[hitvalidx] = winninginterval
```
pierce314159, line 459: this is a def nitpicky, but is there a reason we don't do this on one line?
```python
containinginterval[hitvalidx] = intervalhits[tmpidx]
```
This feels more symmetrical to the single pdarray case

pierce314159 (approved):
The core logic looks good to me!

I left a few comments, but none are really important

arkouda/alignment.py
Ethan-DeBandi99: I am in agreement with @pierce314159 on this.

Ethan-DeBandi99 (approved):
This looks good to me. I do agree with @pierce314159 on his comments and would be in support of those changes. The comments on this are great and make the code much easier to follow.

PR #1603, opened 2022-07-19
Closes #1376 - add Datetime and Timedelta accessors

### LINKED ISSUE #1376: add Datetime methods/accessors to extract components of datetime values
I know this is currently possible (for most components) by manipulating the floor method and various frequencies. 
Here are some thoughts:
(akdt is an arkouda Datetime array)

- akdt.dt.day : day of month, integer (1-31)
- akdt.dt.month: month, integer (1-12)
- akdt.dt.year: year, integer (e.g. 1999, 2021, 2022)
- akdt.dt.date: Datetime object, equivalent to akdt.floor('d')
- akdt.dt.hour: hour, integer (0-23)
- akdt.dt.minute: minute, integer (0-59)
- akdt.dt.second: second, integer (0-59)
- akdt.dt.dayofweek: integer, week starts on Monday, which is denoted by 0 and ends on Sunday which is denoted by 6. (aliases dt.day_of_week, dt.weekday)
- akdt.dt.dayofyear: integer, ordinal day of year (aliases dt.day_of_year)

Additionally, for Timedelta, 
(assume aktd is an arkouda Timedelta array)

- follow [pandas API](https://pandas.pydata.org/docs/reference/api/pandas.Series.dt.components.html) for aktd.dt.components
- aktd.dt.components returns a Dataframe of the components of the Timedeltas.
- components are:    days  hours  minutes  seconds  milliseconds  microseconds  nanoseconds

Finally, I'd love to see `aktd.dt.total_seconds()` which returns the value in seconds as a float.

jeichert60:
This PR (closes #1376): provides accessors for:

Datetime
  * day of monthe
  * month
  * year
  * date
  * hour
  * minute
  * second
  * day of week
  * day of year

Timedelta
  * components (dataframe)
    * days
    * hours
    * minutes
    * seconds
    * milliseconds
    * microseconds
    * nanoseconds

### DIFF
--- a/arkouda/timeclass.py
+++ b/arkouda/timeclass.py
@@ -9,11 +9,15 @@ from pandas import date_range as pd_date_range  # type: ignore
 from pandas import timedelta_range as pd_timedelta_range  # type: ignore
 from pandas import to_datetime, to_timedelta  # type: ignore
 
+# from arkouda import DataFrame
+import arkouda
+
+# from arkouda.dataframe import DataFrame
 from arkouda.dtypes import int64, int_scalars, intTypes, isSupportedInt
 from arkouda.numeric import abs as akabs
 from arkouda.numeric import cast
 from arkouda.pdarrayclass import pdarray
-from arkouda.pdarraycreation import from_series
+from arkouda.pdarraycreation import array, from_series
 
 _BASE_UNIT = "ns"
 
@@ -61,13 +65,15 @@ def _identity(x, **kwargs):
     return x
 
 
+def _isLeapYear(year: int) -> bool:
+    return ((year & 0x3) == 0) & (((year % 100) != 0) | ((year % 400) == 0))
+
+
 class _Timescalar:
     def __init__(self, scalar):
         if isinstance(scalar, np.datetime64) or isinstance(scalar, datetime.datetime):
             scalar = to_datetime(scalar).to_numpy()
-        elif isinstance(scalar, np.timedelta64) or isinstance(
-            scalar, datetime.timedelta
-        ):
+        elif isinstance(scalar, np.timedelta64) or isinstance(scalar, datetime.timedelta):
             scalar = to_timedelta(scalar).to_numpy()
         self.unit = np.datetime_data(scalar.dtype)[0]
         self._factor = _get_factor(self.unit)
@@ -84,54 +90,50 @@ class _AbstractBaseTime(pdarray):
     so that all resulting operations are transparent.
     """
 
-    def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
+    def __init__(self, pda, unit: str = _BASE_UNIT):  # type: ignore
 
-        if isinstance(array, Datetime) or isinstance(array, Timedelta):
-            self.unit: str = array.unit
-            self._factor: int = array._factor
+        if isinstance(pda, Datetime) or isinstance(pda, Timedelta):
+            self.unit: str = pda.unit
+            self._factor: int = pda._factor
             # Make a copy to avoid unknown symbol errors
-            self.values: pdarray = cast(array.values, int64)
+            self.values: pdarray = cast(pda.values, int64)
         # Convert the input to int64 pdarray of nanoseconds
-        elif isinstance(array, pdarray):
-            if array.dtype not in intTypes:
-                raise TypeError(
-                    f"{self.__class__.__name__} array must have int64 dtype"
-                )
+        elif isinstance(pda, pdarray):
+            if pda.dtype not in intTypes:
+                raise TypeError(f"{self.__class__.__name__} pda must have int64 dtype")
             # Already int64 pdarray, just scale
             self.unit = unit
             self._factor = _get_factor(self.unit)
             # This makes a copy of the input array, to leave input unchanged
-            self.values = cast(
-                self._factor * array, int64
-            )  # Mimics a datetime64[ns] array
-        elif hasattr(array, "dtype"):
+            self.values = cast(self._factor * pda, int64)  # Mimics a datetime64[ns] array
+        elif hasattr(pda, "dtype"):
             # Handles all pandas and numpy datetime/timedelta arrays
-            if array.dtype.kind not in ("M", "m"):
+            if pda.dtype.kind not in ("M", "m"):
                 # M = datetime64, m = timedelta64
-                raise TypeError(f"Invalid dtype: {array.dtype.name}")
-            if isinstance(array, Series):
+                raise TypeError(f"Invalid dtype: {pda.dtype.name}")
+            if isinstance(pda, Series):
                 # Pandas Datetime and Timedelta
                 # Get units of underlying numpy datetime64 array
-                self.unit = np.datetime_data(array.values.dtype)[0]
+                self.unit = np.datetime_data(pda.values.dtype)[0]
                 self._factor = _get_factor(self.unit)
                 # Create pdarray
-                self.values = from_series(array)
+                self.values = from_series(pda)
                 # Scale if necessary
                 # This is futureproofing; it will not be used unless pandas
                 # changes its Datetime implementation
                 if self._factor != 1:
                     # Scale inplace because we already created a copy
                     self.values *= self._factor
-            elif isinstance(array, np.ndarray):
+            elif isinstance(pda, np.ndarray):
                 # Numpy datetime64 and timedelta64
                 # Force through pandas.Series
-                self.__init__(to_datetime(array).to_series())  # type: ignore
-            elif hasattr(array, "to_series"):
+                self.__init__(to_datetime(pda).to_series())  # type: ignore
+            elif hasattr(pda, "to_series"):
                 # Pandas DatetimeIndex
                 # Force through pandas.Series
-                self.__init__(array.to_series())  # type: ignore
+                self.__init__(pda.to_series())  # type: ignore
         else:
-            raise TypeError(f"Unsupported type: {type(array)}")
+            raise TypeError(f"Unsupported type: {type(pda)}")
         # Now that self.values is correct, init self with same metadata
         super().__init__(
             self.values.name,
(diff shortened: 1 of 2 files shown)


arkouda/timeclass.py
```diff
@@ -144,6 +139,131 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
         # Deprecated
         self._data = self.values
 
+        if len(self.values):
+            # Year, hour, minute, second
+            days = self.values // _get_factor("d")
+            seconds = self.values // _get_factor("s")
+            days = cast((seconds // 86400), int64) + 1
+            self._year = cast((days // 365.25 + 1970), int64)
```
Ethan-DeBandi99, line 147: this is being returned as a `pdarray`. In pandas, `year` is returned as an Index. @reuster986, do you have a preference here for pdarray vs index?

arkouda/timeclass.py
```diff
@@ -144,6 +139,131 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+    @property
+    def month(self):
+        return self._month
+
+    @property
+    def day(self):
+        return self._day
+
+    @property
+    def week(self):
```
Ethan-DeBandi99, line 229: `._week` is never defined.

arkouda/timeclass.py
```diff
@@ -144,6 +139,131 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+        return self._week
+
+    @property
+    def dayofyear(self):
+        return self._dayofyear
```
Ethan-DeBandi99, line 234: The return here is shifted
```python
start_time = "2022-07-18 12:59:00"
dt = ak.date_range(start=start_time, periods=5, freq="d")
x = dt.dayofyear
```
results in 
```
[198 199 200 201 202]
```

pandas equivalent results in
```
Int64Index([199, 200, 201, 202, 203], dtype='int64')
```

arkouda/timeclass.py
```diff
@@ -144,6 +139,131 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+            year = self._year.to_ndarray()
+            lyear = self._year[:].to_ndarray()
+            lyear = [np.logical_and(year % 4 == 0, year % 100 != 0)]
+            lyear = np.array([np.logical_or(lyear, year % 400 == 0)]).astype(int)
+            days = self._day[:].to_ndarray()
+            self._date = []
+            for i in range(days.size):
+                days[i] += month_days[lyear[0][0][i]][self._month[i]]
+                self._date.append(f"{self._year[i]}-{self._month[i]:02d}-{self._day[i]:02d}")
+            self._dayofyear = ak.array(days)
```
Ethan-DeBandi99, line 205: This will be an issue with larger dataset here because `to_ndarray` pulls the data local. Thus, all of the computation is done on the client, which is not what we want.

We have support on the server to handle this using `&` and `|` as the operators on 2 pdarrays.
```python
a = ak.array([1,3])
b = ak.array([1, 1])
a&b
array([1 1])

a|b
array([1 3])
```

tests/datetime_test.py
```diff
@@ -34,6 +48,33 @@ def setUp(self):
         self.tdvec2 = ak.Timedelta(ak.ones(100, dtype=ak.int64), unit="s")
         self.onesecond = pd.Timedelta(1, unit="s")
 
+    def test_timedelta(self):
+        ak_df = build_ak_df()
```
Ethan-DeBandi99, line 52: I think we should probably just build this DataFrame using pandas since we are testing against it. Otherwise, we are still testing the functionality against itself.

tests/datetime_test.py
```diff
@@ -34,6 +48,33 @@ def setUp(self):
...
+        self.assertTrue(((tdvec3.month.to_ndarray() == exp.month).all()).all())
+        self.assertTrue(((tdvec3.day.to_ndarray() == exp.day).all()).all())
+        self.assertTrue(((tdvec3.hour.to_ndarray() == exp.hour).all()).all())
+        self.assertTrue(((tdvec3.minute.to_ndarray() == exp.minute).all()).all())
+        self.assertTrue(((tdvec3.second.to_ndarray() == exp.second).all()).all())
+        self.assertTrue(((tdvec3.dayofweek.to_ndarray() == exp.dayofweek).all()).all())
+        self.assertTrue(((tdvec3.dayofyear.to_ndarray() == exp.dayofyear).all()).all())
+        actual_total_seconds = list(tdvec3.total_seconds.to_ndarray())
+        exp_total_seconds = [np.datetime64(x,'s').astype(float) for x in np.array(exp)]
+        self.assertTrue(actual_total_seconds == exp_total_seconds)
```
Ethan-DeBandi99, line 76: This is testing that to_pandas works because you are converting the ak object to pandas and then testing that against the result. Build exp directly in pandas, not from the arkouda object.

Ethan-DeBandi99 (changes requested):
Still needs a few things to be ironed out. 

I would like to get @reuster986 input because we return pdarray for everything in arkouda and pandas returns and index for just about everything that is not `.date`. Should we match pandas and return an `Index` or continue returning a pdarray?

arkouda/timeclass.py
reuster986: @Ethan-DeBandi99 good question. I am actually not sure why pandas uses `Index` so much. I don't have a strong opinion either way -- I don't see a compelling need to conform to pandas on this, but if it helps simplify our design philosophy with fewer deviations from pandas, then I'm ok with it. Maybe for now we should stick with `pdarray` and start an issue to explore this question.

tests/datetime_test.py
jeichert60: Now testing against a pandas data_range

tests/datetime_test.py
jeichert60: Now using pandas dataframe

arkouda/timeclass.py
jeichert60: Now using the pdarray operators

arkouda/timeclass.py
jeichert60: Fixed algorithm

arkouda/timeclass.py
jeichert60: Removed

arkouda/timeclass.py
Ethan-DeBandi99: Sounds good. #1641 added to track the conversation around this.

Ethan-DeBandi99 (changes requested):
Pandas supports `pd.datetime.week`, so we should be too. My previous comment may not have been clear, but we need to calculate `._week` so we can return the property.

**EDIT**
I also noticed that this PR needs to be rebased over master. I pulled it locally to do some testing and it does not have recently merged features included. Hopefully, it will flag merge conflicts, but if not you will end up unintentionally overwriting a bunch of recent changes.

arkouda/timeclass.py
Ethan-DeBandi99: This should not have been removed. Instead, we need to compute `._week` to be the week of the year and ensure consistency with pandas.

Ethan-DeBandi99 (changes requested):
There are some calculations that are not working properly
`.dayofyear`
```
start_time = "2015-12-31 00:00:00"
dt = ak.date_range(start=start_time, periods=7, freq="y")
dt.dayofyear
[365 1 1 365 365 1 1]

check_dt = pd.date_range(start=start_time, periods=7, freq="y")
check_dt.dayofyear
Int64Index([365, 366, 365, 365, 365, 366, 365], dtype='int64')
```

Doing some research, it looks like pandas now relies heavily on the Python `datetime` module which clearly also has some unexpected things going on (as made obvious by the above example). We at least need to be matching the pandas output for all situations.

Ethan-DeBandi99 (commented):
The issue from my previous comment has been resolved. I do want to note, that I came up with an algorithm for `.week` this morning. Admittedly, it does not perfectly handle all the edges cases (if would need to be fine tuned some), but here is the pseudocode to do the majority of the work. 

```python
# assuming that dt is created using `ak.date_range`

first_of_year = dt.values - (ak.timeclass._get_factor("d")*(dt.dayofyear))

# compute the day of the week of the first of the year
fdw = ak.cast(((first_of_year/ ak.timeclass._get_factor("d")) + 3) + 1, ak.int64)
fdw %= 7

# compute days in first week
daysinfirstweek = (6 - fdw)

# compute things needed by iteration
day_of_yr = dt.dayofyear
days_in_first_week_of_last_yr = # similar computation as above
days_in_first_week_of_next_yr = # similar computation
prev_yr_is_leap = False # this needs to be calculated

# this example converts to a list to highlight the iteration and each case. This should most likely be done on the server.
for (i, days) in enumerate(daysinfirstweek.to_list()):
    if day_of_yr[i] <= days:
        if days > 4:
            weeks.append(1)
        elif (6 - days_in_first_week_of_last_yr[i]) == 4 or prev_is_leap and (6 - days_in_first_week_of_last_yr[i]) == 5:
            weeks.append(53)
        else:
            weeks.append(52)
    else:
        w = math.ceil((doy[i] - daysinfirstweek[i]) / 7)
        if daysinfirstweek[i] >= 4:
            w += 1
        elif w == 53 and days_in_first_week_of_next_yr[i]>=4:
            w = 1
        weeks.append(w)
return weeks
```
Again, some of the edge cases are a little off with the above (I did not perform every calculation), but overall this is close. We should look into cleaning it up and implementing weeks, though we should probably do this on the server due to the number of computations.

pierce314159:
> The issue from my previous comment has been resolved. I do want to note, that I came up with an algorithm for `.week` this morning. Admittedly, it does not perfectly handle all the edges cases (if would need to be fine tuned some), but here is the pseudocode to do the majority of the work.

Digging into this a bit, i think this should be reasonable to add week of year in this PR, I found where pandas does this calculation and (at a glance) it doesn't seem terrible to implement. I believe we are trying to return `iso_week` from here:
https://github.com/pandas-dev/pandas/blob/6e1a0401420bffd282f3cf8b57e653a39e373d3b/pandas/_libs/tslibs/ccalendar.pyx#L182-L198

We should ensure whatever our solution is passes pandas week of year edge cases test (` test_woy_boundary`):
https://github.com/pandas-dev/pandas/blob/6e1a0401420bffd282f3cf8b57e653a39e373d3b/pandas/tests/scalar/timestamp/test_timestamp.py#L204

arkouda/timeclass.py
```diff
@@ -144,6 +143,158 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+                    days[i] += month_days[lyear[i]][self._month[i]]
+                self._date.append(f"{self._year[i]}-{self._month[i]:02d}-{self._day[i]:02d}")
+            self._dayofyear = ak.array(days)
+
+            # Day of week
+            self._dayofweek = cast(((self.values / _get_factor("d")) + 3), int64)
+            self._dayofweek %= 7
+
+            # Week of year
+            week = (self._dayofyear[:] - 1) - self._dayofweek + 3
```
Ethan-DeBandi99, line 224: `self._dayofyear[:]` is making an unnecessary copy of `self._dayofyear`. On large arrays, this will be an issue. Update this to `week = (self._dayofyear - 1) - self._dayofweek + 3`.

arkouda/timeclass.py
```diff
@@ -144,6 +143,158 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+            for i in range(days.size):
+                if week[i] < 0:
+                    year = int(self._year[i] - 1)
+                    if week[i] > -2 or (week[i] == -2 and _isLeapYear(year)):
+                        week[i] = 53
+                    else:
+                        week[i] = 52
+                elif week[i] == 53:
+                    if 31 - self._day[i] + self.dayofweek[i] < 3:
+                        week[i] = 1
```
Ethan-DeBandi99, line 237: This is something we are definitely going to need to move to the server. @pierce314159 do you have any strong opinions on if it goes no or in a latter PR?

Ethan-DeBandi99 (changes requested):
The computation is calculating correctly. There is a small thing with grabbing a copy that we want to remove. I also left a question for Pierce because we may be ok to hold off on moving things to the server until a later point.

arkouda/timeclass.py
pierce314159: yeah i think we should def move this and the other `for i in range(days.size)` loop to the server. I would be okay doing that in another PR, but I think this is still not performant enough to go in rn. My thinking is if we can come up with a way to do these loops with vector operations, then it should be fine to keep them client side for now. But if we can't come up with a vector op way, we should move it to the server in this PR

arkouda/timeclass.py
```diff
@@ -144,6 +143,158 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+            self._date = []
+            for i in range(days.size):
+                if self._month[i] == 2 and self._day[i] == 29 and lyear[i] == 0:
+                    self._month[i] += 1
+                    self._day[i] = 1
+                    days[i] += month_days[lyear[i]][self._month[i] - 1]
+                else:
+                    days[i] += month_days[lyear[i]][self._month[i]]
+                self._date.append(f"{self._year[i]}-{self._month[i]:02d}-{self._day[i]:02d}")
+            self._dayofyear = cast(ak.array(days), int64)
```
pierce314159, line 217: this seems to the only place we use the `ak` import. I get why u tried to solve it this way, there was an existing design decision that was unfortunate

in `def __init__(self, array, unit: str = _BASE_UNIT)`, the `array` argument overrides the `array` function from `pdarraycreation`. I think the best solution is to update this argument to have another name (like `pda`) and update the appropriate occurrences, so you free up the namespace to do `array()` here

Also we don't have to create the array and then cast, we can just use dtype to do
```python
array(days, int64)
```

arkouda/timeclass.py
```diff
@@ -144,6 +143,158 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+    @property
+    def total_seconds(self):
+        ts = cast(self.values, float)
+        return ts / _get_factor("s")
+
+    # Timedelta values
+    @property
+    def components(self):
+        dt = self.to_pandas()
+        return dt.components
```
pierce314159, line 296: should we be doing `self.to_pandas` here? I feel like we should just modify this function to return our components in a similar format as pandas

pierce314159 (changes requested):
I left a few notes. I haven't looked at the tests too closely but will once these get addressed

arkouda/timeclass.py
```diff
@@ -144,6 +143,158 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+            d = self._day[:]
+            self._day[d >= 153] -= 153
+            self._month[d >= 153] = 8
+
+            d = self._day[:]
+            self._day[d >= 122] -= 122
+            self._month[d >= 122] += 4
+            d = self._day[:]
+            self._day[d >= 61] -= 61
+            self._month[d >= 61] += 2
```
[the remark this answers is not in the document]
pierce314159: or we could store the boolean pdarray to avoid recomputation since we use it more than once
```python
self._month = self._day - self._day + 3
ge_306 = self._day >= 306
self._month[ge_306] = 1
self._day[ge_306] -= 306
```

reuster986 (changes requested):
In addition to @pierce314159 and @Ethan-DeBandi99 's good feedback, I have a general concern about the performance implications of pre-computing all these attributes in `__init__`. I count something like 100 new server messages in this PR, and even if the computations were free, every server message has something like 50 ms of overhead. Currently, creating a Datetime/Timedelta feels like an instantaneous operation, but after this PR I predict it would take over 5 seconds, which is pretty jarring. At large scales, the extra memory usage from pre-computing these attributes might also become problematic.

To avoid degrading the current user experience, I think we should move these computations inside the `@property` methods so that they only get evaluated when the user asks for the corresponding attribute. It's probably fine to cache attributes that have already been computed to avoid re-computation, but I think pre-computing all of them is too much.

arkouda/timeclass.py
```diff
@@ -9,11 +9,15 @@
 from pandas import timedelta_range as pd_timedelta_range  # type: ignore
 from pandas import to_datetime, to_timedelta  # type: ignore
 
+# from arkouda import DataFrame
+import arkouda
+
+# from arkouda.dataframe import DataFrame
```
pierce314159, line 15: I don't think we need the `import arkouda` or either of the 2 commented out imports

arkouda/timeclass.py
```diff
@@ -144,6 +146,155 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+    def date(self):
+        return self._date
+
+    @property
+    def year(self):
+        return self._year
+
+    @property
+    def month(self):
+        return self._month
```
pierce314159, line 242: we need to move the code for calculating these attributes into the property functions to avoid the unnecessary overhead like reuster said in his review. Something like the follow would work:

```python
def __init__(self, pda, unit: str = _BASE_UNIT):
  ...
self._date = None
self._year = None
  ...

@property
def date(self):
  if self._date is None:
    self._date = # do code to calculate date
  return self._date

@property
def year(self):
  if self._date is None:
    self._year = cast((days // 365.24 + 1970), int64)
  return  self._year
```
This way we only do the computation when a user asks for that attribute. But we save our work, so we don't have to recompute if they ask for it again

arkouda/timeclass.py
pierce314159: This unnecessary copy is still present

pierce314159 (changes requested):
Big things are we still need to implement reusters suggestion of lazy initialization and caching of the property attributes and still need to vectorize the `for i in range(pdarray.size)` loops or move them serverside

arkouda/timeclass.py
```diff
@@ -144,6 +143,158 @@ def __init__(self, array, unit: str = _BASE_UNIT):  # type: ignore
...
+            month_days = [
+                [0, 0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334],
+                [0, 0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335],
+            ]
+
+            lyear1 = self._year % 4 == 0
+            lyear2 = self._year % 100 != 0
+            lyear3 = self._year % 400 == 0
+            lyear = lyear1 & lyear2
+            lyear = cast((lyear | lyear3), int64)
```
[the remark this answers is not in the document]
pierce314159: This is has been updated to apply the `_isLeapYear` function to every element of the pdarray which isn't the same thing as what i provided here which performs the computation on the entire vector in one operation

PR #1606, opened 2022-07-19
Closes #1572: Update modular build process to initialize modules on command line

### LINKED ISSUE #1572: Proposal to update modular build once 1.26 is required by Arkouda
## Summary
In https://github.com/chapel-lang/chapel/pull/18724 (added in 1.26), modules specified on the command-line of a build are initialized, even if (seemingly) unused. This means that module-level code will be run from all modules on the command line (e.g., `chpl A.chpl B.chpl`, both modules initialized regardless of code). 

Using this feature in the Arkouda build would allow us to simplify the current modular build layout by removing the need for the `registerMe()` function and stop generating Chapel code that is magically handling the includes.

## Change proposed
1. move contents of each modules `registerMe()` function to module-level code
2. update generation script to add modules to the command line build rather than generate a new file

That is it!

Branch: https://github.com/bmcdonald3/arkouda/tree/cmd-line-modular

bmcdonald3:
This PR changes from the current approach of generating a
`ServerRegistration.chpl` file that calls the `registerMe()`
function required in all modules to instead move the contents
of the `registerMe()` to module-level scope that will add the
functions to the command map upon module initialization. All
modules that are to be modularly included in the server are now
directly specified on the command line.

This change is enabled by the work from https://github.com/chapel-lang/chapel/pull/18724,
which initializes all modules specified on the command line of a
`chpl` call, even if the modules aren't explicitly `use`d. This
means that Arkouda will require at least Chapel 1.26 after this
change.

Closes https://github.com/Bears-R-Us/arkouda/issues/1572


src/serverModuleGen.py
```diff
@@ -7,20 +7,22 @@ def getModules(filename):
...
+    res = ""
     for mod in getModules(config_filename):
-        serverfile.write(f"  import {mod};\n  {mod}.registerMe();\n")
+        if mod[0] != '/':
+            res += f" {src_dir}/{mod}.chpl"
+        else:
+            res += f"{mod}.chpl"
 
-    serverfile.write("}\n")
+    print(res)
```
pierce314159, line 24: I noticed the `if` starts with a space and the `else` doesn't, just wanted to check that was intentional and not overlooked

okay also i couldn't help but try and code golf this
```python
print("".join([f" {src_dir}/{mod}.chpl" if mod[0] != "/" else f"{mod}.chpl" for mod in getModules(config_filename)]))
```

src/serverModuleGen.py
```diff
@@ -7,20 +7,22 @@ def getModules(filename):
         modules = configfile.readlines()
         ret = []
         for module in modules:
-            module = module.split("#")[0].split("/")[-1].strip()
+            module = module.split("#")[0].strip()
             if module:
                 ret.append(module)
         return ret
```
pierce314159, line 13: if we actually like the code golf style we can also shorten this one a bit. With the added benefit of being first time I've ever actually used the walrus operator!
```python
return [m for module in modules if (m:=module.split("#")[0].strip())]
```

pierce314159: and if we really want to get crazy, i think the entire file might be able to get down to just the below, but i haven't tested it
```python
import sys

if __name__ == "__main__":
    with open(sys.argv[1]) as configfile:
        get_modules = [m for module in configfile.readlines() if (m := module.split("#")[0].strip())]
    print(" ".join([f"{sys.argv[2]}/{mod}.chpl" if mod[0] != "/" else f"{mod}.chpl" for mod in get_modules]))
```

src/serverModuleGen.py
bmcdonald3: Ooh, good catch, that was overlooked by me, it should have a space.

While I am continually impressed by the code gold syntax, I have to say I am not the biggest fan. In my opinion, the code just becomes a little bit harder to read/maintain (it is cool though).

pierce314159: That's fair! I certainly can't argue it's more readable lol. But it is more fun to write!

I know dict/list comprehensions can have slightly better performance but that hardly matters here

pierce314159 (approved):

Ethan-DeBandi99 (approved):
Looks good

bmcdonald3:
Forgot to update the docs -- marking as draft until I get that finished up

mhmerrill (approved):
i think this looks good.

MODULAR.md
```diff
@@ -12,17 +12,13 @@ New config files must follow the same format, listing one module name per line w
...
+Adding a module from outside of the Arkouda `src/` directory can be done in one of two ways: (1) adding the absolute pat of the module as a new line to the `ServerModules.cfg` file or (2) setting the `ARKOUDA_SERVER_USER_MODULES` environment variable to a string of the absolute path to the Chapel module. If multiple user modules are included, the paths must be separated by a space. For example, Arkouda could be built with a custom module `/path/toTestMsg.chpl` by adding `/path/to/TestMsg` to the `ServerModules.cfg` file and running `make` or by running the command `ARKOUDA_SERVER_USER_MODULES='/path/to/TestMsg.chpl' make` in the Arkouda home directory. Note that for this to work, the absolute path to the `.chpl` file must be specified in the `ARKOUDA_SERVER_USER_MODULES` environment variable.
```
pierce314159, line 15: missing `h` in `path`

pierce314159 (approved):
small typo in docs, looks good otherwise!

PR #1596, opened 2022-07-15
Closes #1531 and #1605: Update tests to use `assertListEqual` and add `.to_list` methods

pierce314159:
This PR (Closes #1531 and Closes #1605):
- Updates tests to use `assertListEqual` instead of `assertTrue(( == ).all())` to give more info in debugging (see issue for example)
- Adds `to_list` method to arkouda objects which have a `to_ndarray` method and updates tests to use it

PR #1612, opened 2022-07-20
Closes #1600 - `SegArray.to_ndarray()` returns `list`

### LINKED ISSUE #1600: `SegArray.to_ndarray` returns a list
```python
>>> segarr = ak.SegArray(ak.array([0, 4, 7]), ak.arange(12))
>>> segarr.to_ndarray()
[array([0, 1, 2, 3]), array([4, 5, 6]), array([ 7,  8,  9, 10, 11])]

>>> type(segarr.to_ndarray())
list
```

I'm not sure if this is an issue or not, because it kinda makes sense for it to return an iterable of `ndarray`s. But in that case the name seems a bit misleading and at the very least the doc string needs to be updated

https://github.com/Bears-R-Us/arkouda/blob/5b0fdd8150bfe38f7495b9ba02684a6feae8cc84/arkouda/segarray.py#L681-L685

Ethan-DeBandi99:
Closes #1600 

Updated `ak.SegArray.to_ndarray()` to return an `ndarray`. Updated `ak.DataFrame` to convert to a list when running `to_pandas` because columns in a df need to be 1d, otherwise some operations fail.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -510,9 +510,9 @@ class DataFrame(UserDict):
         # Get units that make the most sense.
         if self._bytes < 1024:
             mem = self.memory_usage(unit="B")
-        elif self._bytes < 1024 ** 2:
+        elif self._bytes < 1024**2:
             mem = self.memory_usage(unit="KB")
-        elif self._bytes < 1024 ** 3:
+        elif self._bytes < 1024**3:
             mem = self.memory_usage(unit="MB")
         else:
             mem = self.memory_usage(unit="GB")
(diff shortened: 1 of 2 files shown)


arkouda/dataframe.py
```diff
@@ -1397,12 +1395,16 @@ def to_pandas(self, datalimit=maxTransferBytes, retain_index=False):
...
         pandas_data = {}
         for key in self._columns:
             val = self[key]
             try:
-                pandas_data[key] = val.to_ndarray()
+                # in order for proper pandas functionality, SegArrays must be seen as 1d
+                # and therefore need to be converted to list
+                pandas_data[key] = (
+                    val.to_ndarray() if not isinstance(val, SegArray) else val.to_ndarray().tolist()
+                )
```
pierce314159, line 1407: just wanted to note that i think this will be simplified to `val.to_list()` once #1596 is merged. So whichever is merged second will need to investigate this a bit

pierce314159 (approved):
Looks good!

I want to note that the `to_pandas` method can likely be simplified by #1596, so whichever is merged second should investigate that a bit

jeichert60 (approved):

Issue #1622, opened 2022-07-22
Value mismatches with numpy in operations test


pierce314159:
We have some values mismatches with numpy when running `testAllOperators`:
```
  Value mismatches:         5 / 569
int64(scalar) // float64(array): np: [inf 22. 11.  7.  5.  4.  3.  3.  2.  2.]
ak: [nan 22. 11.  7.  5.  4.  3.  3.  2.  2.]
uint64(scalar) // float64(array): np: [inf 22. 11.  7.  5.  4.  3.  3.  2.  2.]
ak: [nan 22. 11.  7.  5.  4.  3.  3.  2.  2.]
float64(scalar) // int64(array): np: [inf  3.  1.  1.  0.  0.  0.  0.  0.  0.]
ak: [nan  3.  1.  1.  0.  0.  0.  0.  0.  0.]
float64(scalar) // float64(array): np: [inf 14.  7.  4.  3.  2.  2.  2.  1.  1.]
ak: [nan 14.  7.  4.  3.  2.  2.  2.  1.  1.]
bool(scalar) // float64(array): np: [inf  4.  2.  1.  1.  0.  0.  0.  0.  0.]
ak: [nan  4.  2.  1.  1.  0.  0.  0.  0.  0.]
```

These were not present in PR #1388

Ethan-DeBandi99:
@pierce314159 and I bounced a few ideas around on the cause of this, which ended up being correct. We verified with against numpy as well, but what is happening is that numpy is treating division by `0` on `bool` as `inf` and not `nan`.

We need to update arkouda to mimic this behavior.

PR #1388, opened 2022-05-12
Fixes #1387: Boolean comparison binops

pierce314159:
This PR (Fixes #1387):
- Adds the `BoolOps` comparison code for bool array compared with `int`, `uint`, and `float` scalars/arrays
- Adds the code for `float binop uint` and `bool binop uint`
- Adds other comparison functions for bools (ex `bool < bool`) by treating them as `int`

Edit:
@bmcdonald3 gave me a suggestion that fixed my issues with `BoolOps` comparions between `bool` and `float`!!! Thanks Ben!

the condensed output of ` testAllOperators` after this PR:
```
# ops not implemented by numpy or arkouda: 114
# ops implemented by numpy but not arkouda: 182
# ops implemented by arkouda but not numpy: 4
# ops implemented by both: 564
  Matching results:         564 / 564
  Arkouda execution errors: 0 / 564

  Dtype mismatches:         0 / 564

  Value mismatches:         0 / 564
```
vs before this PR
```
# ops not implemented by numpy or arkouda: 114
# ops implemented by numpy but not arkouda: 290
# ops implemented by arkouda but not numpy: 4
# ops implemented by both: 456
  Matching results:         456 / 456
  Arkouda execution errors: 0 / 456

  Dtype mismatches:         0 / 456

  Value mismatches:         0 / 456
```

<details>
  <summary>Full output of testAllOperators</summary>

```
Operators:  frozenset({'-', '%', '<', '|', '<<<', '>=', '*', '+', '**', '<<', '//', '!=', '<=', '^', '>', '&', '/', '==', '>>', '>>>'})
Dtypes:  dict_keys(['int64', 'uint64', 'float64', 'bool'])
pdarrays: 
int64 :  [0 1 2 3 4 5 6 7 8 9]
uint64 :  [0 1 2 3 4 5 6 7 8 9]
float64 :  [0.000000e+00 2.222222e-01 4.444444e-01 6.666667e-01 8.888889e-01 1.111111e+00 1.333333e+00 1.555556e+00 1.777778e+00 2.000000e+00]
bool :  [True False True False True False True False True False]
ndarrays: 
int64 :  [0 1 2 3 4 5 6 7 8 9]
uint64 :  [0 1 2 3 4 5 6 7 8 9]
float64 :  [0.         0.22222222 0.44444444 0.66666667 0.88888889 1.11111111
 1.33333333 1.55555556 1.77777778 2.        ]
bool :  [ True False  True False  True False  True False  True False]
scalars: 
int64 :  5
[...]

PR #1623, opened 2022-07-22
Fixes #1620: Fix uint and float binops

### LINKED ISSUE #1620: __sub__ BINOP in pdarrayclass not implemented for uint64 - float64
This issue may apply to other binops involving ak.uint64 as well. 

Executing the operation: 

```
X = ak.cast(ak.array([5]), dt = ak.uint64)
Y = ak.cast(ak.array([3.01]), dt = ak.float64)
X-Y
```

Results in ```RuntimeError: Error: binopvvMsg: uint64 - float64 not implemented```

pierce314159:
This PR (Fixes #1620):
- Adds code in `binopvv`, `binopvs`, and `binopsv` to enable support for `uint` binop `float`
- Adds test of this functionality against numpy

the condensed output of testAllOperators after this PR:
```
# ops not implemented by numpy or arkouda: 114
# ops implemented by numpy but not arkouda: 158
# ops implemented by arkouda but not numpy: 4
# ops implemented by both: 594
  Matching results:         588 / 594
  Arkouda execution errors: 0 / 594
  Dtype mismatches:         0 / 594
  Value mismatches:         6 / 594
int64(scalar) // float64(array): np: [inf 22. 11.  7.  5.  4.  3.  3.  2.  2.]
ak: [nan 22. 11.  7.  5.  4.  3.  3.  2.  2.]
uint64(scalar) // float64(array): np: [inf 22. 11.  7.  5.  4.  3.  3.  2.  2.]
ak: [nan 22. 11.  7.  5.  4.  3.  3.  2.  2.]
float64(scalar) // int64(array): np: [inf  3.  1.  1.  0.  0.  0.  0.  0.  0.]
ak: [nan  3.  1.  1.  0.  0.  0.  0.  0.  0.]
float64(scalar) // uint64(array): np: [inf  3.  1.  1.  0.  0.  0.  0.  0.  0.]
ak: [nan  3.  1.  1.  0.  0.  0.  0.  0.  0.]
float64(scalar) // float64(array): np: [inf 14.  7.  4.  3.  2.  2.  2.  1.  1.]
ak: [nan 14.  7.  4.  3.  2.  2.  2.  1.  1.]
bool(scalar) // float64(array): np: [inf  4.  2.  1.  1.  0.  0.  0.  0.  0.]
ak: [nan  4.  2.  1.  1.  0.  0.  0.  0.  0.]
```

vs before this PR:
```
# ops not implemented by numpy or arkouda: 114
# ops implemented by numpy but not arkouda: 182
# ops implemented by arkouda but not numpy: 4
# ops implemented by both: 569
  Matching results:         564 / 569
  Arkouda execution errors: 0 / 569

  Dtype mismatches:         0 / 569

  Value mismatches:         5 / 569
int64(scalar) // float64(array): np: [inf 22. 11.  7.  5.  4.  3.  3.  2.  2.]
ak: [nan 22. 11.  7.  5.  4.  3.  3.  2.  2.]
uint64(scalar) // float64(array): np: [inf 22. 11.  7.  5.  4.  3.  3.  2.  2.]
ak: [nan 22. 11.  7.  5.  4.  3.  3.  2.  2.]
float64(scalar) // int64(array): np: [inf  3.  1.  1.  0.  0.  0.  0.  0.  0.]
ak: [nan  3.  1.  1.  0.  0.  0.  0.  0.  0.]
float64(scalar) // float64(array): np: [inf 14.  7.  4.  3.  2.  2.  2.  1.  1.]
ak: [nan 14.  7.  4.  3.  2.  2.  2.  1.  1.]
bool(scalar) // float64(array): np: [inf  4.  2.  1.  1.  0.  0.  0.  0.  0.]
ak: [nan  4.  2.  1.  1.  0.  0.  0.  0.  0.]
```

NOTES:
- the value mismatches are surprising because they were not present in #1388. I have opened issue #1622, to investigate these further
- I wrote the test rather quickly so please let me know if there are any mistakes or if there's a better way

### DIFF
--- a/tests/operator_tests.py
+++ b/tests/operator_tests.py
@@ -280,6 +280,51 @@ class OperatorsTest(ArkoudaTest):
             (ak.array([True, False, True, False, True, True]) == ak.concatenate([pdaOne, pdaTwo])).all()
         )
 
+    def test_float_uint_binops(self):
+        # Test fix for issue #1620
+        ak_uint = ak.array([5], dtype=ak.uint64)
+        np_uint = np.array([5], dtype=np.uint64)
+        scalar_uint = np.uint64(5)
+
+        ak_float = ak.array([3.01], dtype=ak.float64)
+        np_float = np.array([3.01], dtype=np.float_)
+        scalar_float = 3.01
+
+        ak_uints = [ak_uint, scalar_uint]
+        np_uints = [np_uint, scalar_uint]
+        ak_floats = [ak_float, scalar_float]
+        np_floats = [np_float, scalar_float]
+        for aku, akf, npu, npf in zip(ak_uints, ak_floats, np_uints, np_floats):
+            self.assertEqual(ak_uint + akf, np_uint + npf)
+            self.assertEqual(akf + ak_uint, npf + np_uint)
+            self.assertEqual(ak_float + aku, np_float + npu)
+            self.assertEqual(aku + ak_float, npu + np_float)
+
+            self.assertEqual(ak_uint - akf, np_uint - npf)
+            self.assertEqual(akf - ak_uint, npf - np_uint)
+            self.assertEqual(ak_float - aku, np_float - npu)
+            self.assertEqual(aku - ak_float, npu - np_float)
+
+            self.assertEqual(ak_uint * akf, np_uint * npf)
+            self.assertEqual(akf * ak_uint, npf * np_uint)
+            self.assertEqual(ak_float * aku, np_float * npu)
+            self.assertEqual(aku * ak_float, npu * np_float)
+
+            self.assertEqual(ak_uint / akf, np_uint / npf)
+            self.assertEqual(akf / ak_uint, npf / np_uint)
+            self.assertEqual(ak_float / aku, np_float / npu)
+            self.assertEqual(aku / ak_float, npu / np_float)
+
+            self.assertEqual(ak_uint // akf, np_uint // npf)
+            self.assertEqual(akf // ak_uint, npf // np_uint)
+            self.assertEqual(ak_float // aku, np_float // npu)
+            self.assertEqual(aku // ak_float, npu // np_float)
+
+            self.assertEqual(ak_uint ** akf, np_uint ** npf)
+            self.assertEqual(akf ** ak_uint, npf ** np_uint)
+            self.assertEqual(ak_float ** aku, np_float ** npu)
+            self.assertEqual(aku ** ak_float, npu ** np_float)
+
     def test_concatenate_type_preservation(self):
         # Test that concatenate preserves special pdarray types (IPv4, Datetime, BitVector, ...)
         from arkouda.util import generic_concat as akuconcat
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -296,6 +296,37 @@ module BinOp
         }
       var repMsg = "created %s".format(st.attrib(rname));
       return new MsgTuple(repMsg, MsgType.NORMAL);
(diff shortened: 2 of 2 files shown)


Ethan-DeBandi99 (commented):
Logic and code looks good overall. I do have one question that is more for consistency. I noticed in the `BinOp` code that it does not appear consistent for when we are casting for the operation. I wanted to make sure there is not a reason I am unaware of for explicitly casting values for operations in some scenarios and not others. I believe the impact of this is minimal, but would it be worth making this consistent?

joshmarshall1 (approved):
As long as Ethan's comment doesn't pose change to be made, this looks good

pierce314159:
> I do have one question that is more for consistency. I noticed in the `BinOp` code that it does not appear consistent for when we are casting for the operation. I wanted to make sure there is not a reason I am unaware of for explicitly casting values for operations in some scenarios and not others. I believe the impact of this is minimal, but would it be worth making this consistent?
> 
> If this is something we wanted to do, I would probably recommend updating the changes here to match the decision and add a new issue to update the remaining operations that were not touched here.

This is a great question! But the answer prob won't be super satisfying. Is there a good reason or is it arbitrary?... kinda both. I think the right way to distinguish between these cases is to add comments. I think adding an issue to go through and see which spots need to be explicitly cast to get correct behavior and commenting those is a good idea
____
## When it's arbitrary

To be completely honest, the reason I cast to `real` in this case is because i know the return type is supposed to be `real`. And i didn't want to have to think to hard about if there would be any corner cases i'm missing when the top bit is being used. I don't think this should be strictly necessary because chapel handles these ops natively, but I decided to be explicit. I think this is likely the case in other places as well

## When there's a good reason

I think there was a discussion over a video chat about this when `uint` support was first added because I can't seem to find it documented in any issues/PRs. But there definitely are cases where the casting is necessary. There are some cases that arkouda supports (because numpy supports them) that chapel doesn't natively. The easiest example is `uint+int`, it isn't clear what the return type should be in this case

Numpy handles this by return a float:
```python
>>> np.uint(5) + 3
8.0
```

Chapel throws an error:
```Chapel
module PlayingAround {

  proc main () throws {
    var u: uint = 5;
    var i: int = 3;
    writeln(u+i);
  }
}
```
```
$ chpl -o playing PlayingAround.chpl
PlayingAround.chpl:3: In function 'main':
PlayingAround.chpl:6: error: illegal use of '+' on operands of type uint(64) and signed integer
```

So the way to be consistent with numpy is to cast to `real` before the operation
https://github.com/Bears-R-Us/arkouda/blob/14d98cc479af4ea456021fdeb8642502f83e5434/src/BinOp.chpl#L247-L252

tests/operator_tests.py
```diff
@@ -280,6 +280,51 @@ def testConcatenate(self):
...
+
+        ak_float = ak.array([3.01], dtype=ak.float64)
+        np_float = np.array([3.01], dtype=np.float_)
+        scalar_float = 3.01
+
+        ak_uints = [ak_uint, scalar_uint]
+        np_uints = [np_uint, scalar_uint]
+        ak_floats = [ak_float, scalar_float]
+        np_floats = [np_float, scalar_float]
+        for aku, akf, npu, npf in zip(ak_uints, ak_floats, np_uints, np_floats):
```
pierce314159, line 297: looking back at this, i realize this is only testing `ak_uint` against `ak_float` and never comparing either against the scalar versions... This can be solved by a nested `for` but their might be a better way, either way this needs to be updated

pierce314159: actually im wrong, ignore me.... this code is fine lmao

Ethan-DeBandi99 (approved):
@pierce314159 thanks for the background! I think this is good to go.

Issue #1404, opened 2022-05-18
Reduce sort overheads


ronawho:
The core RadixSortLSD routine has some start up overheads that can be pretty dramatic, especially at scale. At 75K cores we have seen 20-25 second of start up time. For large sorts this is typically amortized, but this start up cost can be painful for small sorts and when doing strong scaling (running a fixed problem size that you want to get faster with more nodes.) In this issue, we'd like to explore some of the overheads and see what we can do to reduce them.

There are a couple of known start up costs for RadixSortLSD including algorithmic overhead, aggregator creation time, and dynamic memory registration costs (at least under gasnet-ibv). Going into this, I was assuming most of the overhead would be algorithmic, but that doesn't appear to be the whole story. The following table has results from some initial explorations with dynamic and static registration at varying number of tasks. This is with sorting `2**15` elements per node on a 32 node InfiniBand system with 128 cores per node.

| Tasks | Dynamic  | Static |
| ----: | -------: | -----: |
|    1  |  0.26s   |  0.15s |
|  128  |  2.99s   |  0.59s |
| 2048  | 60.79s   |  6.81s |

Run details (with subtimers from https://github.com/Bears-R-Us/arkouda/compare/master...ronawho:sort-subtimers-2):

<details>

gasnet-ibv-large (dynamic registration):
```
CHPL_FLAGS="--gasnet-segment=large -sperfOnlyCompile -sperfValRange='uint(32)'" make TEST_BINARY_DIR=test-bin-large test-bin-large/UnitTestSort

./test-bin-large/UnitTestSort -nl 32 --elemsPerLocale=$((2**15)) --RSLSD_numTasks=1
checkSortedT:       0.00
arrayCreateT:       0.01
countDigitsT:       0.02
globalCountsT:      0.17
calcPosAndPermuteT: 0.05
arrayCopyT:         0.00
ranksAssignT:       0.00
 radixSortLSD_ranks(A:[] int(64)) -- (rand uint(32) vals) -- 0.98 MiB/s per node (0.26s)

./test-bin-large/UnitTestSort -nl 32 --elemsPerLocale=$((2**15)) --RSLSD_numTasks=128
checkSortedT:       0.00
arrayCreateT:       0.01
countDigitsT:       0.28
globalCountsT:      0.16
calcPosAndPermuteT: 2.53
arrayCopyT:         0.00
ranksAssignT:       0.00
[...]

PR #1635, opened 2022-07-26
Reduce sort bucket exchange overhead

ronawho:
Previously, each task would write its local bucket counts directly to
the global counts in transposed order. Each task has 64K buckets with
64K/numLocales going to each locale. With 8K aggregation buffers this
meant that at over 8 locales we'd have single-use aggregators that would
get very small at higher scales. This meant that we were paying the
non-trivial aggregation creation cost and not amortizing it with
multiple uses. And worse, each task was flushing in lockstep to locale
0, then 1, and so on. So there were many single-use aggregators being
used in lockstep, which created bottlenecks at scale.

Improve that here by combining all the buckets from local tasks on a
node and chunking up the write. This allows us to send fewer and
fuller aggregation buffers and write to all locales simultaneously
instead of doing so in lockstep. This reduces sort startup costs,
particularly at scale.

At 40 nodes on a Cray CS with HDR InfiniBand this takes a trivial sized
sort on 32-bit uints from 0.60s down to 0.15s. This is the largest
InfiniBand system I have access to, but I also simulated 240 nodes by
running 6x oversubscribed on these 40 nodes. For the same trivial
problem size this takes us from 90.0s down to 1.7s. This isn't really a
fair timing since overheads will be exaggerated but if we're at 1.7s
oversubscribed we'll be even lower on real hardware.

Back when I had access to a large InfiniBand system I saw a trivial sort
take 20-25s at 512 nodes. Simulating 512 nodes on this 40 node system
the trivial sort takes 7s. That's with over 12x oversubscription, so I
would guess on real hardware it'd be down around a second or two.

I believe we can further improve the bucket exchange by doing RDMA
writes of the buffers in non transposed order and then transposing
before/after scanning, but that's a larger algorithmic change and this
has a substantial improvement so I think is still a worthwhile stepping
stone with a simpler implementation.

Perf results in table form:

| Config                     | Before | Now   |
| -------------------------- | -----: | ----: |
|  40 node IB                |   0.6s | 0.15s |
| 240 node IB (  6x oversub) |  90.0s | 1.70s |
| 512 node IB (~12x oversub) | 560.0s | 7.10s |
| 512 node XC                |   3.0s | 0.50s |

Part of #1404

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -19,6 +19,7 @@ module RadixSortLSD
     use IO;
     use CTypes;
     use Reflection;
+    use RangeChunk;
     use Logging;
     use ServerConfig;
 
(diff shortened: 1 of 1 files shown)


ronawho:
Drafting only because I'm not around if there's any fallout from this. I feel pretty confident about the change if you want to try it out, but if you want to wait I'll be back on the 8th.

src/RadixSortLSD.chpl
```diff
@@ -78,28 +79,30 @@ module RadixSortLSD
...
+                    // write counts in to global counts in transposed order
+                    coforall tid in Tasks {
                         var aggregator = newDstAggregator(int);
-                        for bucket in bD {
-                            aggregator.copy(globalCounts[calcGlobalIndex(bucket, loc.id, task)], 
-                                                         taskBucketCounts[bucket]);
+                        for task in Tasks {
+                            ref taskBucketCounts = tasksBucketCounts[task];
+                            for bucket in chunk(0..#numBuckets, numTasks, tid) {
+                                aggregator.copy(globalCounts[calcGlobalIndex(bucket, loc.id, task)],
```
mhmerrill, line 103: this seems ok to me. although my brain goes a little numb following the index reordering.

mhmerrill:
> Drafting only because I'm not around if there's any fallout from this. I feel pretty confident about the change if you want to try it out, but if you want to wait I'll be back on the 8th.

i will also be out until Aug 8th.

src/RadixSortLSD.chpl
ronawho: The indexing is actually the same as before. The change here is that previously each task would scatter/gather all it's own buckets and now tasks gather/scatter chunks of all the buckets on a node.

reuster986 (approved):
This is great!

pierce314159 (approved):
This is awesome! What a great speedup

PR #1636, opened 2022-07-26
Test to see if 1.26.0 causes OOM Error

Ethan-DeBandi99:
This PR is only for testing the CI step that is failing with an OOM error from the docker container.


hokiegeek2 (approved):
looks good

Ethan-DeBandi99:
We have not seen this issue pop up again this week. Based on this, I believe the OOM error we were seeing was being caused by the host machines for our CI being overtasked at times. 

I am going to close this PR as a result, since we can easily set up another one for testing if needed.

PR #1638, opened 2022-07-26
Closes #1639: Adds aggregations for first, mode, and unique

### LINKED ISSUE #1639: Aggregations for first, mode, and unique
Would like three new `GroupBy` aggregation methods:
- `first` returns the first value of each group
- `mode` returns the most common value in each group (first in the case of ties)
- `unique` returns a `SegArray` of all the unique values per group

reuster986:
This PR adds three new `GroupBy` aggregation methods and corresponding unit tests:
- `first` returns the first value of each group
- `mode` returns the most common value in each group (first in the case of ties)
- `unique` returns a `SegArray` of all the unique values per group

Closes #1639

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -398,6 +398,12 @@ class GroupBy:
         # TO DO: remove once logic is ported over to Chapel
         if operator == "nunique":
             return self.nunique(values)
+        if operator == "first":
+            return self.first(values)
+        if operator == "mode":
+            return self.mode(values)
+        if operator == "unique":
+            return self.unique(values)
 
         # All other aggregations operate on pdarray
         if cast(pdarray, values).size != self.length:
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -10,9 +10,9 @@ import numpy as np  # type: ignore
 from typeguard import typechecked
 
 from arkouda.client import generic_msg
+from arkouda.dtypes import float64 as akfloat64
 from arkouda.dtypes import int64 as akint64
 from arkouda.dtypes import uint64 as akuint64
-from arkouda.dtypes import float64 as akfloat64
 from arkouda.infoclass import list_registry
 from arkouda.logger import getArkoudaLogger
 from arkouda.pdarrayclass import (
@@ -112,7 +112,11 @@ def unique(
     repMsg = generic_msg(
         cmd="unique",
         args="{} {} {:n} {} {}".format(
-            return_groups, assume_sorted, effectiveKeys, " ".join(keynames), " ".join(keytypes),
+            return_groups,
+            assume_sorted,
+            effectiveKeys,
+            " ".join(keynames),
+            " ".join(keytypes),
         ),
     )
     if return_groups:
@@ -1007,7 +1011,7 @@ class GroupBy:
         ----------
         values : pdarray-like
             The values from which to take the first of each group
-        
+
         Returns
         -------
         unique_keys : (list of) pdarray-like
@@ -1028,7 +1032,7 @@ class GroupBy:
         ----------
         values : (list of) pdarray-like
             The values from which to take the mode of each group
-        
+
         Returns
         -------
         unique_keys : (list of) pdarray-like
@@ -1065,12 +1069,12 @@ class GroupBy:
     def unique(self, values: groupable) -> Tuple[groupable, groupable]:
         """
         Return the set of unique values in each group, as a SegArray.
-        
+
         Parameters
         ----------
         values : (list of) pdarray-like
             The values to unique
-        
+
         Returns
         -------
         unique_keys : (list of) pdarray-like
@@ -1079,6 +1083,7 @@ class GroupBy:
             The unique values of each group
         """
         from arkouda.segarray import SegArray
+
         togroup = self._nested_grouping_helper(values)
         # Group to unique (key, value) pairs
         g = GroupBy(togroup)
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -12,6 +12,7 @@ from typeguard import typechecked
 from arkouda.client import generic_msg
 from arkouda.dtypes import int64 as akint64
 from arkouda.dtypes import uint64 as akuint64
+from arkouda.dtypes import float64 as akfloat64
 from arkouda.infoclass import list_registry
 from arkouda.logger import getArkoudaLogger
 from arkouda.pdarrayclass import (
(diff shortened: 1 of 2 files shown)


tests/groupby_test.py
```diff
@@ -440,6 +441,41 @@ def test_zero_length_groupby(self):
         g = ak.GroupBy(ak.zeros(0, dtype=ak.int64))
         str(g.segments)  # passing condition, if this was deleted it will cause the test to fail
 
+    def test_first_aggregation(self):
+        keys = ak.array([0, 1, 0, 1, 0, 1])
+        vals = ak.array([9, 8, 7, 6, 5, 4])
+        ans = [9, 8]
+        g = ak.GroupBy(keys)
+        _, res = g.first(vals)
+        self.assertListEqual(ans, res.to_ndarray().tolist())
```
Ethan-DeBandi99, line 450: you should be able to use the new `to_list()` functionality that runs `.to_ndarray().tolist()` under the hood. @pierce314159 added this recently to make tests easier to read.

tests/groupby_test.py
```diff
@@ -440,6 +441,41 @@ def test_zero_length_groupby(self):
...
+
+    def test_mode_aggregation(self):
+        keys = ak.array([0, 1, 0, 1, 0, 1, 0, 1])
+        vals = ak.array([4, 3, 5, 3, 5, 2, 6, 2])
+        ans = [5, 3]
+        g = ak.GroupBy(keys)
+        _, res = g.mode(vals)
+        self.assertListEqual(ans, res.to_ndarray().tolist())
```
Ethan-DeBandi99, line 458: Again, `to_ndarray().tolist()` can be replaced with `to_list()`

tests/groupby_test.py
```diff
@@ -440,6 +441,41 @@ def test_zero_length_groupby(self):
...
+        # Test with multi-array values
+        _, res2 = g.mode([vals, vals])
+        self.assertListEqual(ans, res2[0].to_ndarray().tolist())
+        self.assertListEqual(ans, res2[1].to_ndarray().tolist())
```
Ethan-DeBandi99, line 462: `to_ndarray().tolist()` can be replaced with `to_list()`

tests/groupby_test.py
```diff
@@ -440,6 +441,41 @@ def test_zero_length_groupby(self):
...
+        g = ak.GroupBy(keys)
+        _, res = g.unique(vals)
+        for a, r in zip(ans, res.to_ndarray().tolist()):
+            self.assertListEqual(a, r.tolist())
+        # Test with multi-array values
+        _, res2 = g.unique([vals, vals])
+        for a, r in zip(ans, res2[0].to_ndarray().tolist()):
+            self.assertListEqual(a, r.tolist())
+        for a, r in zip(ans, res2[1].to_ndarray().tolist()):
+            self.assertListEqual(a, r.tolist())
```
Ethan-DeBandi99, line 477: `to_ndarray().tolist()` can be replaced with `to_list()`

Ethan-DeBandi99 (changes requested):
Code looks good overall. One request to update `to_ndarray().tolist()` to use the new `to_list()` functionality to conform to all other tests. @pierce314159 added this recently for convenience. I marked each location with a comment to make them easy to find.

arkouda/groupbyclass.py
```diff
@@ -742,7 +763,27 @@ def argmax(self, values: pdarray) -> Tuple[groupable, pdarray]:
...
-        return self.aggregate(values, "argmax")
+        k, v = self.aggregate(values, "argmax")
+        return k, cast(pdarray, v)
+
+    def _nested_grouping_helper(self, values: groupable) -> groupable:
+        unique_key_idx = self.broadcast(arange(self.ngroups), permute=True)
+        if hasattr(values, "_get_grouping_keys"):
+            # All single-array groupable types must have a _get_grouping_keys method
+            if isinstance(values, pdarray) and cast(pdarray, values).dtype == akfloat64:
+                raise TypeError("grouping/uniquing unsupported for float64 arrays")
```
pierce314159, line 774: python's `and` does short circuiting, so the second clause is only ever evaluated if the first part is true, so we don't need the explicit cast since we know `values` is an instance of `pdarray`

So we can simplify a bit to just 
```python
if isinstance(values, pdarray) and values.dtype == akfloat64:
```

arkouda/groupbyclass.py
```diff
@@ -742,7 +763,27 @@ def argmax(self, values: pdarray) -> Tuple[groupable, pdarray]:
...
+            togroup = [unique_key_idx, values]
+        else:
+            # Treat as a sequence of groupable arrays
+            for v in values:
+                if (
+                    isinstance(values, pdarray)
+                    and cast(pdarray, values).dtype != akint64
+                    and cast(pdarray, values).dtype != akuint64
+                ):
```
pierce314159, line 783: two things here
- we can apply the same short circuiting logic to avoid an explicit cast
- `v` never seems to be used here. I think `values` is supposed to `v` here

I think we should change this to 
```python
for v in values:
    if isinstance(v, pdarray) and v.dtype not in [akint64, akuint64]:
```

arkouda/groupbyclass.py
```diff
@@ -742,7 +763,27 @@ def argmax(self, values: pdarray) -> Tuple[groupable, pdarray]:
...
+                    raise TypeError("nunique unsupported for this dtype")
```
pierce314159, line 784: it seems `_nested_grouping_helper` is used by more than just `nunique`, so we should probably update this error message to be like the one above `"grouping/uniquing unsupported for this dtype"`

tests/groupby_test.py
```diff
@@ -326,8 +327,8 @@ def test_error_handling(self):
         with self.assertRaises(TypeError):
             gb.broadcast([])
 
-        with self.assertRaises(TypeError):
-            self.igb.nunique(ak.randint(0, 1, 10, dtype=bool))
+        # with self.assertRaises(TypeError):
+        #     self.igb.nunique(ak.randint(0, 1, 10, dtype=bool))
```
pierce314159, line 331: we should remove this if `bool` is no longer supposed to raise a type error

tests/groupby_test.py
pierce314159: I think that this may eliminate the need for the `r.tolist()` too!

pierce314159 (changes requested):
Looks good overall! Just some things to look at in `_nested_grouping_helper`

pierce314159 (approved):

Ethan-DeBandi99 (approved):
Looks good to go

PR #1647, opened 2022-07-29
Closes #1630 - Use `Enum` for `ObjType`

### LINKED ISSUE #1630: Use `Enum` for `objType` in JSON messages
@hokiegeek2 suggested to look into using and `Enum` for the `objType` to avoid developers setting the wrong value.

This would be updated on the Python and Chapel side .

Examples for how this is configured is in `arkouda/message.py`. Create a class that inherits `Enum` and sets the values we need.

Requires PR #1619 to be merged.

Ethan-DeBandi99:
Closes #1630 

- Adds `ObjectType` and `ParameterObject` classes to the client
- Uses `Enum` for `objType` assignment
- Adds `enum ObjectType` to the server
- Updates JSON parsing to create `ParameterObject` and writes to json using the `.dict` property

*Functionally, this does not change the JSON message arguments, but provides a more formal structure that will be easier to maintain*

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -8,7 +8,7 @@ import zmq  # type: ignore
 
 from arkouda import __version__, io_util, security
 from arkouda.logger import getArkoudaLogger
-from arkouda.message import MessageFormat, MessageType, ReplyMessage, RequestMessage
+from arkouda.message import MessageFormat, MessageType, ReplyMessage, RequestMessage, ParameterObject, ObjectType
 
 __all__ = [
     "connect",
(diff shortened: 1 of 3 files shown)


arkouda/client.py
```diff
@@ -570,30 +577,23 @@ def _json_args_to_str(json_obj: Dict) -> Tuple[int, str]:
...
-            param["dtype"] = dtypes.pop()
-            param["val"] = json.dumps(val)
+            param = ParameterObject(key, ObjectType.LIST, dtypes.pop(), json.dumps(val))
         else:
-            param["objType"] = type(val).__name__
-            param["dtype"] = type(val).__name__
-            param["val"] = val if isinstance(val, str) else str(val)
-        j.append(json.dumps(param))
+            v = val if isinstance(val, str) else str(val)
+            param = ParameterObject(key, ObjectType.VALUE, type(val).__name__, v)
```
pierce314159, line 595: I might be wrong but I think now that the we use `ObjectType.VALUE` here instead of `type(val).__name__`, we need to update `getValueAsType` to check that `objType != "value"` instead of `objType != dtype`

https://github.com/Bears-R-Us/arkouda/blob/afd073556545ab32ffabdf7e72511c0e09405bf0/src/Message.chpl#L158-L160

pierce314159 (commented):
Just wanna check that our objtype comparisons on the server are still correct in the `ObjectType.VALUE` case

arkouda/client.py
Ethan-DeBandi99: You are 100% correct. I update this (and the similar check in `getList`).

pierce314159 (approved):
Looks good!

hokiegeek2 (approved):
Totally cool!

PR #1652, opened 2022-08-01
Closes #1631 - Message Arg Generation Using Dispatch Table

### LINKED ISSUE #1631: Dispatch Table to replace `if else` for JSON Parsing
The function converting JSON into a format that is readable by arkouda has an `if elif else` block with 4 cases currently. @hokiegeek2 recommended that if this expands, it would be a good candidate to update to a dispatch table.

Update this block to utilize a dispatch table preemptively.

PR #1619 changes required for this change.

Ethan-DeBandi99:
Closes #1631 

This PR updates the `if/else` logic to generate the appropriate `ParameterObject` based on the type to use a dispatch table. This is for better performance down the road when more object formats are moved to the server.

- Adds `factory()` static method to `ParameterObject` to pass a `key, val` pair and return the appropriate `ParameterObject`. This is using the dispatch table.
- Updates the `_json_args_to_str()` to call `ParameterObject.factory(key, val)` instead of running the logic itself.

### DIFF
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -115,7 +115,7 @@ class ParameterObject:
         -------
         ParameterObject
         """
-        dtypes = set([p.dtype for p in val])
+        dtypes = set([p.dtype if hasattr(p, "dtype") else type(p).__name__ for p in val])
         if len(dtypes) > 1:
             t_str = ", ".join(dtypes)
             raise TypeError(f"List values must be of the same type. Found {t_str}")
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -3,6 +3,9 @@ from __future__ import annotations
 from dataclasses import dataclass
 from enum import Enum
 from typing import Dict
+from typeguard import typechecked
+
+import json
 
 
 class ObjectType(Enum):
@@ -32,6 +35,7 @@ class ObjectType(Enum):
 
 
 class ParameterObject:
+
     __slots = ("key", "objType", "dtype", "val")
 
     key: str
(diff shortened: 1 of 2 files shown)


arkouda/message.py
```diff
@@ -54,6 +58,129 @@ def dict(self):
...
+        key : str
+            key from the dictionary object
+        val
+            the value corresponding to the provided key from the dictionary
+
+        Returns
+        --------
+        ParameterObject - The parameter object formatted to be parsed by the chapel server
+        """
+        dispatch = ParameterObject.generate_dispatch()
```
pierce314159, line 178: is it possible to pull `dispatch` out, so we can cache the table instead of rebuilding? I first I thought we might be able to do just define it since it doesn't appear to change, but then i realized their might be circular dependency issues

Ethan-DeBandi99:
This requires the change from PR #1655, so I am converting to draft for now.

arkouda/message.py
Ethan-DeBandi99: We probably can, I will look into this.

Ethan-DeBandi99: In order to do this (due to the circular references) we would have to update the current return from `generate_dispatch()` to 
```
return {
           "pdarray": ParameterObject._build_pdarray_param,
            "Strings": ParameterObject._build_strings_param,
            list.__name__: ParameterObject._build_list_param,
        }
```
This is not really an issue to do, but will rely on devs providing the correct value for arkouda class names because it would need to be the equivalent to `class.__name__`.

Ethan-DeBandi99: We would also need to change the structure because making it a class property of `ParameterObject` is not possible due to the fact that the dispatch is mapping to static methods within the class which are not yet defined. 

The other option would be to create the functions at the file level and have `arkouda.client.PARAM_DISPATCH`. This would again require the changes above unless we defined inside a function to prevent issues with the circular import. It would end up looking something like this
```
def generate_dispatch() -> Dict:
    """
    Builds and returns the dispatch table used to build parameter object.

    Returns
    -------
    Dictionary - mapping the parameter type to the build function
    """
    from arkouda.pdarrayclass import pdarray
    from arkouda.strings import Strings

    return {
        pdarray.__name__: _build_pdarray_param,
        Strings.__name__: _build_strings_param,
        list.__name__: _build_list_param,
    }
PARAM_DISPATCH = generate_dispatch()
```
This would also require the `_build_*` functions to be moved out of the `ParameterObject` class

Ethan-DeBandi99: Attempted configuring as above and we still get stuck with the circular import issues.

hokiegeek2: @pierce314159 definitely makes an excellent point. However, given the circular dependency stuff here, I recommend keeping this as-is for now and adding a tech debt issue to resolve the circular dependency

hokiegeek2 (approved):
overall looks good to me

pierce314159 (approved):

PR #1655, opened 2022-08-01
Closes #1654 - Fix JSON bug when param does not have `.dtype`

### LINKED ISSUE #1654: Bug in JSON message parsing with list of strings
The JSON message parsing assumes that `List`s will contain arkouda objects, It would be nice if it could handle `List[str]` as well

```python
args = {
    "size": len(self.columns),
    "columns": self.columns,
    "data_names": [numeric(self[c]).name for c in self.columns],
}

stuff = json.loads(
    generic_msg(
        cmd="corrMatrix",
        args=args,
    )
)
```
Here `self.columns = ['x', 'y']`. `str` doesn't have a dtype so we get the following error
```
In [5]: ak.DataFrame({'x': x, 'y': y}).corr()

File ~/proj/arkouda/arkouda/client.py:588, in <listcomp>(.0)
    586     param = ParameterObject(key, ObjectType.STRINGS, "str", name)
    587 elif isinstance(val, list):
--> 588     dtypes = set([p.dtype for p in val])
    589     if len(dtypes) > 1:
    590         t_str = ", ".join(dtypes)

AttributeError: 'str' object has no attribute 'dtype'
```
encountered while working #1643

Ethan-DeBandi99:
Closes #1654 

Updates type check to handle when `dtype` attribute does not exist.

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -585,7 +585,7 @@ def _json_args_to_str(json_obj: Dict) -> Tuple[int, str]:
             name = val.name if val.name else ""
             param = ParameterObject(key, ObjectType.STRINGS, "str", name)
         elif isinstance(val, list):
-            dtypes = set([p.dtype for p in val])
+            dtypes = set([p.dtype if hasattr(p, "dtype") else type(p) for p in val])
             if len(dtypes) > 1:
                 t_str = ", ".join(dtypes)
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
looks good!

joshmarshall1 (approved):

PR #1657, opened 2022-08-03
Closes #1589 - Adds test to verify randomness for `ak.randint`

### LINKED ISSUE #1589: Add checks to ensure `randint` values produced by chapel are random enough
Related to #1220, suggested in PR #1542 

We should add checks to ensure that non-seeded random array value generation methods are producing arrays with confidently random enough values. If the values are not determined random enough, produce and return a new set of values.

joshmarshall1:
This PR closes #1589 

This test uses a series of 20 trials with `randint` pdarrays of 250 entries with a min of 0 and a max of 2**32 and performs a 'runs test' on each trial to verify randomness to 95% confidence. The test will pass as long as 80% of trials (16 or more) are verified random. 

In my testing this has passed 25 times with 0 failures. I did increase the requirement to pass to 20/20 trials to force failures in order to verify functionality. Most of these failures were either 18 or 19 successes, with some tests passing all 20 trials.

### DIFF
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -1,4 +1,6 @@
 import datetime as dt
+import math
+import statistics
 from collections import deque
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
This looks good to me.

I am not sure we need more than one test here, but I am thinking it may be beneficial at some point to add a second randomness test. Maybe the Poker Test? This link has some good info on it, https://towardsdatascience.com/poker-test-pattern-based-random-number-detection-in-python-ec2dba4955bc. 

There is also a Python package that will do this, [randtest](https://pypi.org/project/randtest/). Might be worth adding that and using it to check randomness. 

Again, this PR looks good and I think we can merge this, the other comments are just for awareness and and maybe future issues.

pierce314159 (approved):
I don't know enough about stats to verify the logic, but passing a 95% confidence test 80% of the time seems good enough for me! I also like ethan suggestion of using `randtest` i think this combined with being validated by an external random test would be great

Ethan-DeBandi99:
I added issue #1665 to integrate `randtest` as well. I don't think that is a pressing priority, but wanted to have it tracked so this could be merged.

PR #1661, opened 2022-08-04
Closes #1643: `Covariance` and `Correlation`

### LINKED ISSUE #1643: Functions to calculate covariance, pearson r between two pdarrays and generate correlation matrix on dataframe
The below code snippets allow for the calculation of the covariance between two ak.pdarrays, the Pearson Correlation Coefficient (Pearson R) between two ak.pdarrays and the correlation matrix (using above) of a DataFrame. 

The correlation matrix is similar to pandas pd.DataFrame.corr and can be set to return either an ak.DataFrame or pd.DataFrame. When returning a pd.DataFrame allows row index to be set to same as column names. 

The below are possibly not the best implementations. Maybe cov and pearson_r (or similar) go into ak.numeric and corr into ak.DataFrame. Up for discussion. 

Attempts to handle type issues. 
```python
def cov(X,Y):
     '''
     Calculates the covariance of the sample
     Params: X, Y are ak.pdarray
     Returns: covariance value (float)
     '''
     covariance = None
     muX = ak.mean(X)
     muY = ak.mean(Y)
     covariance = ak.sum((X-muX)*(Y-muY)) / (X.size-1)
     return covariance

def pearson_r(X,Y):
     '''
     Calculates the Pearson Correlation Coefficient (Pearson R)
     Params: X, Y are ak.pdarray
     Returns: the value of Pearson R as float from -1 to 1
     '''
     covXY = cov(X,Y)
     sigmaX = X.std()
     sigmaY = Y.std()
     corr_p = covXY / (sigmaX * sigmaY)
     return corr_p

def corr(df, return_pandas = True):
     '''
     Generates the correlation matrix using Pearson R for all columns of a Dataframe with each other. Attempts to convert to numeric
     values where possible for inclusion in the matrix. 
     Does not handle uint as current arkouda does not support the required binop
     Params: df:ak.DataFrame
                   return_pandas : boolean 
     Returns: pandas dataframe if return_pandas is True. Otherwise return ak.DataFrame
     '''
     numeric_cols = []
     pd_df = pd.DataFrame()
     #copy the dataframe to handle the type issues. There may be a better way. 
     tmp_df = df.copy(deep = True)
     for c in tmp_df.columns:
           if type(tmp_df[c]) is ak.pdarray:
                 #temp until fixed
                 if tmp_df[c] is ak.uint64:
                      continue
                 else:
                      numeric_cols.append(c)
           elif type(tmp_df[c]) is ak.Categorical:
                tmp_df[c] = tmp_df[c].codes
                numeric_cols.append(c)
           elif type(tmp_df[c]) is ak.Strings:
                tmp_df[c] = ak.Categorical(tmp_df[c]).codes
                numeric_cols.append(c)

     for c in numeric_cols:
          _ = []
          for c2 in numeric_cols:
               _.append(pearson_r(tmp_df[c], tmp_df[c2]))
          pd_df[c] = _
     tmp_df = {}
     if return_pandas:
          pd_df.index = pd_df.columns
          return pd_df
     else:
          return ak.DataFrame.from_pandas(pd_df)

```

pierce314159:
This PR (Closes https://github.com/Bears-R-Us/arkouda/issues/1643):
- Adds covariance `ak.cov(pda1, pda2)`, correlation for pdarrays `ak.corr(pda1, pda2)`, and correlation matrix for dataframes `df.corr()`
- Moves `mean`, `var`, `std` to server
- Adds testing for these functions and their type combos
- Utilizes the new JSON message arguments infrastructure

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1802,6 +1802,49 @@ class DataFrame(UserDict):
 
         return DataFrame(df_def, index=self.index)
 
+    def corr(self) -> DataFrame:
+        """
+        Return new DataFrame with pairwise correlation of columns
+
+        Returns
+        -------
+        DataFrame
+            Arkouda DataFrame containing correlation matrix of all columns
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there's a server-side error thrown
+
+        See Also
+        --------
+        pdarray.corr
+
+        Notes
+        -----
+        Generates the correlation matrix using Pearson R for all columns
+
+        Attempts to convert to numeric values where possible for inclusion in the matrix.
+        """
+        # TODO converting Strings to Categorical goes out of scope, i don't love registering it
+        unregister_names: List = []
+
+        def numeric_help(d):
+            if isinstance(d, Strings):
+                d = Categorical(d).register(f"keep_{len(unregister_names)}")
+                unregister_names.append(d.name)
+            return d if isinstance(d, pdarray) else d.codes
+
+        args = {
+            "size": len(self.columns),
+            "columns": self.columns,
+            "data_names": [numeric_help(self[c]).name for c in self.columns],
+        }
+
+        ret_dict = json.loads(generic_msg(cmd="corrMatrix", args=args))
+        [Categorical.unregister_categorical_by_name(name) for name in unregister_names]
+        return DataFrame({c: create_pdarray(ret_dict[c]) for c in self.columns})
+
 
 def sorted(df, column=False):
     """
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1832,11 +1832,10 @@ class DataFrame(UserDict):
                 d = Categorical(d)
             return d if isinstance(d, pdarray) else d.codes
 
-        numeric_cols = [numeric_help(self[c]) for c in self.columns]
         args = {
             "size": len(self.columns),
             "columns": self.columns,
-            "data_names": [col.name for col in numeric_cols],
+            "data_names": [numeric_help(self[c]) for c in self.columns],
         }
 
         ret_dict = json.loads(generic_msg(cmd="corrMatrix", args=args))
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1826,23 +1826,20 @@ class DataFrame(UserDict):
 
         Attempts to convert to numeric values where possible for inclusion in the matrix.
         """
-        # TODO converting Strings to Categorical goes out of scope, i don't love registering it
-        unregister_names: List = []
 
         def numeric_help(d):
             if isinstance(d, Strings):
-                d = Categorical(d).register(f"keep_{len(unregister_names)}")
-                unregister_names.append(d.name)
+                d = Categorical(d)
             return d if isinstance(d, pdarray) else d.codes
 
+        numeric_cols = [numeric_help(self[c]) for c in self.columns]
         args = {
             "size": len(self.columns),
             "columns": self.columns,
-            "data_names": [numeric_help(self[c]).name for c in self.columns],
+            "data_names": [col.name for col in numeric_cols],
         }
 
         ret_dict = json.loads(generic_msg(cmd="corrMatrix", args=args))
-        [Categorical.unregister_categorical_by_name(name) for name in unregister_names]
         return DataFrame({c: create_pdarray(ret_dict[c]) for c in self.columns})
 
 
--- a/src/Stats.chpl
+++ b/src/Stats.chpl
@@ -9,7 +9,7 @@ module Stats {
         return (+ reduce ((ar:real - mean(ar)) ** 2)) / (aD.size - ddof):real;
     }
 
-    proc std(ar: [?aD] ?t1, ddof: int): real throws {
+    proc std(ar: [?aD] ?t, ddof: int): real throws {
         return sqrt(variance(ar, ddof));
     }
 
--- a/src/Stats.chpl
+++ b/src/Stats.chpl
@@ -1,0 +1,23 @@
+module Stats {
+    use AryUtil;
+
+    proc mean(ar1: [?aD1] ?t): real throws {
+        return (+ reduce ar1:real) / aD1.size:real;
+    }
+
+    proc variance(ar1: [?aD1] ?t, ddof: int): real throws {
+        return (+ reduce ((ar1:real - mean(ar1)) ** 2)) / (aD1.size - ddof):real;
+    }
+
+    proc std(ar1: [?aD1] ?t, ddof: int): real throws {
+        return sqrt(variance(ar1, ddof));
+    }
+
+    proc cov(ar1: [?aD1] ?t, ar2: [?aD2] ?t2): real throws {
+        return (+ reduce ((ar1:real - mean(ar1)) * (ar2:real - mean(ar2)))) / (aD1.size - 1):real;
+    }
+
+    proc corr(ar1: [?aD1] ?t, ar2: [?aD2] ?t2): real throws {
+        return cov(ar1, ar2) / (std(ar1, 1) * std(ar2, 1));
+    }
+}
--- a/src/Stats.chpl
+++ b/src/Stats.chpl
@@ -1,23 +1,23 @@
 module Stats {
     use AryUtil;
 
-    proc mean(ar1: [?aD1] ?t): real throws {
-        return (+ reduce ar1:real) / aD1.size:real;
+    proc mean(ar: [?aD] ?t): real throws {
+        return (+ reduce ar:real) / aD.size:real;
     }
 
-    proc variance(ar1: [?aD1] ?t, ddof: int): real throws {
-        return (+ reduce ((ar1:real - mean(ar1)) ** 2)) / (aD1.size - ddof):real;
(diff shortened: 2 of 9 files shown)


src/Stats.chpl
```diff
@@ -0,0 +1,23 @@
...
+        return (+ reduce ar1:real) / aD1.size:real;
+    }
+
+    proc variance(ar1: [?aD1] ?t, ddof: int): real throws {
+        return (+ reduce ((ar1:real - mean(ar1)) ** 2)) / (aD1.size - ddof):real;
+    }
+
+    proc std(ar1: [?aD1] ?t, ddof: int): real throws {
+        return sqrt(variance(ar1, ddof));
+    }
```
Ethan-DeBandi99, line 14: This is not an issue just me being picky. In both these functions, my brain wanted to look for `ar2` and `aD2` because of the naming. Since there are only 1 array being passed to these functions, could we change `ar1` and `aD1` to `ar` and `aD`.  I think this will read a little easier, but is not a deal breaker.

src/Stats.chpl
```diff
@@ -0,0 +1,23 @@
...
+
+    proc cov(ar1: [?aD1] ?t, ar2: [?aD2] ?t2): real throws {
+        return (+ reduce ((ar1:real - mean(ar1)) * (ar2:real - mean(ar2)))) / (aD1.size - 1):real;
+    }
+
+    proc corr(ar1: [?aD1] ?t, ar2: [?aD2] ?t2): real throws {
+        return cov(ar1, ar2) / (std(ar1, 1) * std(ar2, 1));
+    }
```
Ethan-DeBandi99, line 22: Again, being picky... can we make `?t` be `?t1`. Not really an issue, but provides consistency in the naming.

tests/stats_test.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_mean(self):
+        self.assertEqual(self.x.mean(), self.pdx.mean())
+        self.assertEqual(self.y.mean(), self.pdy.mean())
+        self.assertEqual(self.u.mean(), self.pdu.mean())
+        self.assertEqual(self.b.mean(), self.pdb.mean())
+        self.assertEqual(self.f.mean(), self.pdf.mean())
+
+    def test_var(self):
+        # Note the numpy and pandas var/std differ, we follow numpy by default
+        # The pandas version requires ddof = 1
```
Ethan-DeBandi99, line 32: Can we add this as a `Note` to the `var` method docstring? I am just worried this will get missed otherwise.

tests/stats_test.py
```diff
@@ -0,0 +1,108 @@
...
+        self.c = ak.Categorical(self.s)
+
+    def test_mean(self):
+        self.assertEqual(self.x.mean(), self.pdx.mean())
+        self.assertEqual(self.y.mean(), self.pdy.mean())
+        self.assertEqual(self.u.mean(), self.pdu.mean())
+        self.assertEqual(self.b.mean(), self.pdb.mean())
+        self.assertEqual(self.f.mean(), self.pdf.mean())
+
+    def test_var(self):
```
Ethan-DeBandi99, line 30: Should this also test against numpy? It looks like we are only testing pandas.

arkouda/dataframe.py
```diff
@@ -1802,6 +1802,49 @@ def isin(self, values: Union[pdarray, Dict, Series, DataFrame]) -> DataFrame:
...
+        --------
+        pdarray.corr
+
+        Notes
+        -----
+        Generates the correlation matrix using Pearson R for all columns
+
+        Attempts to convert to numeric values where possible for inclusion in the matrix.
+        """
+        # TODO converting Strings to Categorical goes out of scope, i don't love registering it
```
Ethan-DeBandi99, line 1829: Can we remove this and add an issue for looking into this further? I am not a fan of pushing `TODO`s in master.

Ethan-DeBandi99 (changes requested):
A few minor things, but all the logic looks good. Just me being picky on most the things.

I also wanted to note that @pierce314159 and I discussed putting in an issue with Chapel to allow generic functions to be passed as params. I think that would really help scale down the code with the type checking.

src/Stats.chpl
pierce314159: yeah that's a good idea!

tests/stats_test.py
pierce314159: it's kinda already there. This is saying that 0 is the maximum likelihood estimate and 1 is the unbiased estimator. It might be worthwhile to give a more explicit call out that numpy uses maximum likelihood estimate by default and pandas uses unbiased estimator. and that we default to numpy
https://github.com/Bears-R-Us/arkouda/blob/c8ce6c449562ab6ce67363a347043f1b8d86a76b/arkouda/pdarrayclass.py#L2005-2015

Ethan-DeBandi99: Yeah you are right. I only looked at the docstring of the `var` function inside of the class, not the file level function.

src/StatsMsg.chpl
```diff
@@ -0,0 +1,316 @@
...
+        param pn = Reflection.getRoutineName();
+        var args = parseMessageArgs(payload, 3);
+
+        var size = args.get("size").getIntValue();
+        var columns = args.get("columns").getList(size);
+        var dataNames = args.get("data_names").getList(size);
+
+        var corrDict = new map(keyType=string, valType=string);
+        // TODO CORRELATION MATRIX IS SYMMETRIC, WE CAN USE THAT!
+        // would this be easy to do with a 2D chapel matrix?
```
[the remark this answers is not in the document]
pierce314159: Issue created to track this optimization #1663

arkouda/dataframe.py
pierce314159: Removed! And while I was writing up the issue, I actually came up with a workaround, if u don't mind checking it out and letting me know what u think

```python
>>> s = ak.array([f"string {i}" for i in range(10)])
>>> c = ak.Categorical(s)
>>> s2 = ak.array([f"{i} string {i}" for i in range(10)])
>>> c2 = ak.Categorical(ak.array([f'my_cat {i}' for i in range(10)]))
>>> ak.DataFrame({"s": s, "c": c, "s2": s2, "c2": c2}).corr()
          s         c        s2        c2
0  1.000000  1.000000 -0.127273  0.090909
1  1.000000  1.000000 -0.127273  0.090909
2 -0.127273 -0.127273  1.000000  0.272727
3  0.090909  0.090909  0.272727  1.000000 (4 rows x 4 columns)
```

tests/stats_test.py
pierce314159: done!

arkouda/dataframe.py
Ethan-DeBandi99: @pierce314159 if this work around is functional I would much prefer using that over having to register and deregister things. Though that is a matter of preference on my part. I think this looks simple enough as well that we could update this as part of the current PR and close the subsequent issue you created. Unless you believe there is still a better way to do this. 

*Leaving this thread open for now. The original concern was addressed., but I think there is more discussion to have at this time.*

Ethan-DeBandi99: I also wanted to note, I am not currently seeing the register/unregister code here anymore, did you remove it in anticipation of making the above update?

src/Stats.chpl
Ethan-DeBandi99: Just being picky again, you updated 
`proc std(ar1: [?aD1] ?t, ddof: int): real throws {`
to 
`proc std(ar: [?aD] ?t1, ddof: int): real throws {`
Can we remove the `1` from `?t1` so it matches the definitions of `mean` and `variance`?

Ethan-DeBandi99 (commented):
I updated a few of the comments from my previous review instead of adding new as everything was related to those. Nothing major.

arkouda/dataframe.py
pierce314159: Oh no the code you're seeing is the fix for this. Previously we had
```python
args = {
    "size": len(self.columns),
    "columns": self.columns,
    "data_names": [numeric_help(self[c]).name for c in self.columns],
}
```
which would go out of scope because `Categorical`s created by `numeric_help` go out of scope.

To fix this i created a list to store the `Categorical`s created by `numeric_help` which is gonna stick around for a bit
```python
numeric_cols = [numeric_help(self[c]) for c in self.columns]
args = {
    "size": len(self.columns),
    "columns": self.columns,
    "data_names": [col.name for col in numeric_cols],
}
```
So this takes care of the need to register/unregister

Once your PR #1662, is merged this will just be
```python
numeric_cols = [numeric_help(self[c]) for c in self.columns]
args = {
    "size": len(self.columns),
    "columns": self.columns,
    "data_names": numeric_cols,
}
```
which is why i call this a workaround cause it feels a little silly to create a var for this when it's only used once... but it's wayyyyy better than the reg/unreg stuff

pierce314159: actually, after merging 1662, the way we want to do this just works (i'm guessing because we're storing the actual pdarray in `args` now and not just the name?). So this has been updated to 
```python
args = {
    "size": len(self.columns),
    "columns": self.columns,
    "data_names": [numeric_help(self[c]) for c in self.columns],
}
```

Ethan-DeBandi99 (approved):
Looks good!

joshmarshall1 (approved):

reuster986 (approved):
very cool!

PR #1662, opened 2022-08-05
Closes #1632 - Multiple `dtype` support and improved list processing

### LINKED ISSUE #1632: Multiple `dtype` support for list in JSON formatted message arguments
Expand `list` case to support multiple `dtypes`.

Initial thought is to build a list of the `dtypes` with the `dtype` at index corresponding to the `dtype` of the value at index.

Ethan-DeBandi99:
Closes #1632 

- Adds proper support for lists. The `dtype` is now set to the type of the value within the list. This allows for a list of `pdarray`s that contain different `dtypes` to be in the same list. This does still require the list to be made up of 1 type. 
- Adds testing to ensure that all json arguments are parsed properly and set to allow for parsing on the server.

*Note - The regexes in the test are not using the escape sequences `\w` or `\d` because pytest produces deprecation warnings when using them*

### DIFF
--- a/tests/message_test.py
+++ b/tests/message_test.py
@@ -1,7 +1,11 @@
 import json
 import unittest
 
+from base_test import ArkoudaTest
+
+from arkouda.client import _json_args_to_str
 from arkouda.message import MessageFormat, MessageType, ReplyMessage, RequestMessage
+from arkouda.pdarraycreation import arange, array
 
 
 class MessageTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


joshmarshall1 (approved):

tests/message_test.py
```diff
@@ -98,3 +102,62 @@ def testReplyMessage(self):
...
+        str1 = array(["abc", "def"])
+        str2 = array(["Test", "Test2"])
+        size, args = _json_args_to_str({"str1": str1, "str2": str2})
+        self.assertEqual(size, 2)
+        for a in json.loads(args):
+            p = json.loads(a)
+            self.assertRegex(p["key"], "^str(1|2)$")
+            self.assertEqual(p["objType"], "SEGSTRING")
+            self.assertEqual(p["dtype"], "str")
+            self.assertRegex(p["val"], "^id_[a-zA-Z0-9]{7}_[0-9]$")
```
pierce314159, line 163: doing `\\w` and `\\d` will work here without the deprecation warning, so `"^id_\\w{7}_\\d$"`

pierce314159 (commented):
Logic looks good! I just have a note on addressing the deprecation warning u were seeing with `\w` and `\d`

tests/message_test.py
Ethan-DeBandi99: Oh, sweet! Thanks @pierce314159. I updated it.

pierce314159 (approved):

PR #1673, opened 2022-08-10
Closes #1672: Temporarily change test_randint_randomness not to run

### LINKED ISSUE #1672: Temporarily change `test_randint_randomness` not to run
`test_randint_randomness` has been failing in CI runs a bit too frequently. We should lower the threshold for now and decide how we want to proceed (either relying solely on `randtest` after #1665 or if we are okay with the lower confidence test)

pierce314159:
This PR (Closes #1672):
- Changes the name of `test_randint_randomness` to `randint_randomness` which will cause the test to not run

Note:
Originally i was going to lower the threshold for passing, but did some back of the envelope math which lead me to believe the frequency of failures that we are seeing is surprising if `randint` is truly random. So I decided to remove the test until we can investigate this further

<details>
  <summary>Possibly incorrect probability math</summary>

This is a 95% confidence. So I think that if `randint` is truly random, then every trial has a 95% chance of passing and 5% chance of failing. Since the test should be 20 (i think independent) trials, this is a [Bernoulli experiment](https://en.wikipedia.org/wiki/Bernoulli_trial) with $n = 20$ and $p = 0.95$

So the probability of getting $k$ successes is
$$P(X= k) = { n \choose k}p^k q^{n-k}$$

Since the overall test only fails if we have >4 failures, to calculate the odds of overall success we can sum the probabilities of passing 16, 17, 18, 19, and 20 trials (since these are all independent events)
$$\sum_{i=16}^{20}{20 \choose i}0.95^{i} 0.05^{20-i} = 0.997426$$

so the odds of failure is $1-$ this and should be $0.0025739$. The overall test should only fail $0.25$% of the time... So I'm thinking there is something wrong with `randint` given the frequency of failures
</details>

### DIFF
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -814,7 +814,9 @@ class PdarrayCreationTest(ArkoudaTest):
             self.assertTrue(np.all(a == i + 1))
             self.assertTrue(np.all(npa == i + 1))
 
-    def test_randint_randomness(self):
+    def randint_randomness(self):
+        # THIS TEST DOES NOT RUN, see Issue #1672
+        # To run rename to `test_randint_randomness`
         minVal = 0
         maxVal = 2**32
         size = 250


joshmarshall1 (approved):

Ethan-DeBandi99 (approved):
Looks good

PR #1674, opened 2022-08-10
Closes #1670: Remove `in-place` from `DataFrame.drop` docstring

### LINKED ISSUE #1670: df.drop docstring is misleading
https://github.com/Bears-R-Us/arkouda/blob/c89c1a40bacc012e7e7ce6a667e4282fbb6a034d/arkouda/dataframe.py#L676

Kind of a minor point, but if the main docstring says "in-place" a user would reasonably expect the default to be in place. Instead, the default value of the `inplace` argument is False.   I think either the docstring or default value ought to be changed, and I don't have an opinion on which one. 

Perhaps follow the pandas convention in this case if there's no reason not to:

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html

pierce314159:
Removes `in-place` from `DataFrame.drop` docstring (Closes #1670)

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -680,7 +680,7 @@ class DataFrame(UserDict):
         inplace: bool = False,
     ) -> Union[None, DataFrame]:
         """
-        Drop column/s or row/s from the dataframe, in-place.
+        Drop column/s or row/s from the dataframe.
 
         Parameters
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
Looks good to me

PR #1675, opened 2022-08-11
Closes #1618 - Nested JSON Message Argument Support

### LINKED ISSUE #1618: Add support for nested JSON in Messages Arguments
Future messages may require nested JSON elements. This will require additional parsing support when configuring the message on the python side and additional handling for reading the elements on the server.

Additionally, we should add size data for lists as part of the `ParameterObj` because this is also required for the nested JSON. This prevents the dev from being required to know the size or send it as a separate parameter.

Ethan-DeBandi99:
Closes #1618 

Adds support for nested json arguments to the client and server message parsing.

# Example for use
## Python Code
```python
# one layer nested. 'a' contains JSON
x = {
    'a': {
        'test':1,
        'JSON':2
    }
}

rep_msg = ak.generic_msg(
    cmd="<yourCommandHere>", 
    args=x
)
```

## Chapel Code
```chapel
// This is just an example function I used for initial testing to show how to access sub-json
proc nestedTestMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
    // Parse the top level json into message args
    var msgArgs = parseMessageArgs(payload, 1);

    // Access the json formatted string value 
    var json_str = msgArgs.get('a');

    writeln("JSON STRING: %s".format(json_str.getValue()));

    // Parse the sub-json. This requires you provide the number of parameters like lists.
    var sub_json_args = json_str.getJSON(2);

    writeln("SUB JSON KEYS: %jt".format(sub_json_args.keys()));
    return new MsgTuple("JSON Read Success", MsgType.NORMAL);
}
```

### DIFF
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -137,7 +137,7 @@ class ParameterObject:
     def _build_dict_param(key: str, val: Dict) -> ParameterObject:
         j = []
         for k, v in val.items():
-            if not isinstance(key, str):
+            if not isinstance(k, str):
                 raise TypeError(f"Argument keys are required to be str. Found {type(key)}")
             param = ParameterObject.factory(k, v)
             j.append(json.dumps(param.dict))
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -138,7 +138,7 @@ class ParameterObject:
         j = []
         for k, v in val.items():
             if not isinstance(k, str):
-                raise TypeError(f"Argument keys are required to be str. Found {type(key)}")
+                raise TypeError(f"Argument keys are required to be str. Found {type(k)}")
             param = ParameterObject.factory(k, v)
             j.append(json.dumps(param.dict))
         return ParameterObject(key, ObjectType.DICT, str(dict.__name__), json.dumps(j))
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -17,6 +17,7 @@ class ObjectType(Enum):
     PDARRAY = "PDARRAY"
     STRINGS = "SEGSTRING"
     LIST = "LIST"
+    DICT = "DICT"
     VALUE = "VALUE"
 
     def __str__(self) -> str:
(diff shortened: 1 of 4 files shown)


arkouda/message.py
```diff
@@ -131,6 +132,17 @@ def _build_list_param(key: str, val: list) -> ParameterObject:
             v = val if t == str.__name__ else [str(x) for x in val]
             return ParameterObject(key, ObjectType.LIST, t, json.dumps(v))
 
+    @staticmethod
+    @typechecked
+    def _build_dict_param(key: str, val: Dict) -> ParameterObject:
+        j = []
+        for k, v in val.items():
+            if not isinstance(key, str):
+                raise TypeError(f"Argument keys are required to be str. Found {type(key)}")
```
pierce314159, line 141: should this be checking if `k` is an instance of `str` instead?

pierce314159 (commented):
Looks great other than one small thing

arkouda/message.py
Ethan-DeBandi99: Yes, good catch thank you!

hokiegeek2 (approved):
looks good!

arkouda/message.py
```diff
@@ -131,6 +132,17 @@ def _build_list_param(key: str, val: list) -> ParameterObject:
             v = val if t == str.__name__ else [str(x) for x in val]
             return ParameterObject(key, ObjectType.LIST, t, json.dumps(v))
 
+    @staticmethod
+    @typechecked
+    def _build_dict_param(key: str, val: Dict) -> ParameterObject:
+        j = []
+        for k, v in val.items():
+            if not isinstance(k, str):
+                raise TypeError(f"Argument keys are required to be str. Found {type(key)}")
```
pierce314159, line 141: i think the error message was missed here in the update, i believe it should be `type(k)`

Ethan-DeBandi99: My bad. Fixed.

pierce314159 (approved):

PR #1677, opened 2022-08-12
Closes #1613: Remove empty string regex workaround

### LINKED ISSUE #1613: Remove empty string regex workaround from #959
Now that we've dropped support for chapel <1.25, we no longer need the workaround code from Issue #959 / PR #962

https://github.com/Bears-R-Us/arkouda/blob/00bc0fba6025071f4377602d597b6889716df4a9/arkouda/strings.py#L947-L952

pierce314159:
This PR (Closes #1613):
- Removes the workaround to raise a ValueError for patterns that match the empty string
- Updates `sub` chapel code to correctly handle patterns which match the empty string
- Updates tests for this functionality

There are a few exceptions:
- We don't currently support `pattern="$"` because of [Chapel issue #20431](https://github.com/chapel-lang/chapel/issues/20431)
  - Might be able to come up with a workaround for this
- We don't support `pattern=""` if Strings also contains a `""` because of [Chapel issue #20441](https://github.com/chapel-lang/chapel/issues/20441)
- We don't allow patterns that match the empty string for `peel`, `split`, `flatten`
  - `re.split` and `non-regex flatten` don't work with empty separator, so this lines up
  - `peel` hangs with single character strings and regex which matches the empty string. This needs to be looked into further.

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -562,11 +562,7 @@ class Strings:
         """
         if isinstance(chars, bytes):
             chars = chars.decode()
-        args = {
-            "objType": self.objtype,
-            "name": self.entry.name,
-            "chars": chars
-        }
+        args = {"objType": self.objtype, "name": self.entry.name, "chars": chars}
         rep_msg = generic_msg(cmd="segmentedStrip", args=args)
         return Strings.from_return_msg(cast(str, rep_msg))
 
@@ -596,6 +592,12 @@ class Strings:
             re.compile(pattern)
         except Exception as e:
             raise ValueError(e)
+        if pattern == "$" or (pattern == "" and (self == "").any()):  # type: ignore
+            # TODO remove once changes from chapel issue #20431 and #20441 are in arkouda
+            raise ValueError(
+                "regex operations not currently supported with a pattern=$ or pattern='' when "
+                "the empty string is contained in Strings"
+            )
         matcher = None
         if pattern in self._regex_dict:
             matcher = self._regex_dict[pattern]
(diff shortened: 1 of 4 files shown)


arkouda/strings.py
```diff
@@ -596,6 +592,12 @@ def _get_matcher(self, pattern: Union[bytes, str_scalars], create: bool = True):
             re.compile(pattern)
         except Exception as e:
             raise ValueError(e)
+        if pattern == "$" or (pattern == "" and (self == "").any()):  # type: ignore
+            # TODO remove once changes from chapel issue #20431 and #20441 are in arkouda
+            raise ValueError(
+                "regex operations not currently supported with a pattern=$ or pattern='' when "
+                "the empty string is contained in Strings"
+            )
```
Ethan-DeBandi99, line 600: Can we pull this out into a subfunction since we repeat the same code for this in 3 places?

arkouda/strings.py
```diff
@@ -1071,6 +1073,13 @@ def endswith(self, substr: Union[bytes, str_scalars], regex: bool = False) -> pd
             substr = substr.decode()
         if not regex:
             substr = re.escape(substr)
+        if substr == "$":
+            # TODO remove once changes from chapel issue #20431 and #20441 are in arkouda
+            # no need to add "" case because it will be caught in contains
+            raise ValueError(
+                "regex operations not currently supported with a pattern=$ or pattern='' when "
+                "the empty string is contained in Strings"
+            )
```
Ethan-DeBandi99, line 1082: I know this condition is different, but I would like to see this incorporated to a sub-function as well. Maybe we can pass a `case` parameter to indicate which `if` to use?

Ethan-DeBandi99 (changes requested):
Nothing major, just some reorganization to limit repeated code.

I did also want to ask, you mentioned in the PR `peel` needs to be looked into further. Is this by us or the Chapel side?

pierce314159:
> I did also want to ask, you mentioned in the PR `peel` needs to be looked into further. Is this by us or the Chapel side?

I think the further looking into needs to be done by us, since chapel correctly handles `matches` with a single character string and a pattern that matches the new line. I created issue #1678 to track this, thanks for the reminder!

arkouda/strings.py
pierce314159: yeah this is actually the same case, it's just the `pattern=''` condition wasn't necessary, because it will be handled by the call to `self.contains(substr + "$", regex=True)`. So we can use the same helper across the board

Ethan-DeBandi99 (approved):
Looks good

joshmarshall1 (approved):

PR #1683, opened 2022-08-17
Closes #1571 - Moves SegArray to Server

### LINKED ISSUE #1571: Move `SegArray` to Chapel Module
Part of #1570 

`SegArray` is currently a python class that needs to be configured as a Chapel module.

`SegArray` should also be added to the symbol table as a data type.

This will be configured very similar to `Strings` objects on the server.

Ethan-DeBandi99:
Closes #1571 

- Adds `SegArraySymEntry` as a `CompositeEntry` to types that can be held in the symbol table.
- Adds a `SegArray` class to the server.
- Updates the client constructor to send a server message to add the `SegArray` object to the symbol table
- Adds accessors for `SegArray` properties to be pulled from the server. These are cached so they are only pulled once. This is temporary until client side functionality no longer needs access to the `segments`/`values` independently. Supporting message processing is added to the server to return these properties.

*`SegArray` functionality/computations remain the same. Only the object has been moved to the server in this PR.*

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -4,6 +4,7 @@ from warnings import warn
 
 import numpy as np  # type: ignore
 
+from arkouda.client import generic_msg
 from arkouda.dtypes import bool as akbool
 from arkouda.dtypes import int64 as akint64
 from arkouda.dtypes import isSupportedInt, str_
@@ -13,6 +14,7 @@ from arkouda.pdarrayclass import attach_pdarray, create_pdarray, is_sorted, pdar
 from arkouda.pdarraycreation import arange, array, ones, zeros
 from arkouda.pdarrayIO import load
 from arkouda.pdarraysetops import concatenate
+from arkouda.strings import Strings
 
 
 def gen_ranges(starts, ends, stride=1):
@@ -84,8 +86,9 @@ class SegArray:
             Start index of each sub-array in the flattened values array
         values : pdarray
             The flattened values of all sub-arrays
-        copy : bool
+        copy : bool (Deprecated)
             If True, make a copy of the input arrays; otherwise, just store a reference.
+            Moving the object to the server has resulted in this being deprecated.
 
         Returns
         -------
@@ -97,6 +100,7 @@ class SegArray:
         Keyword args 'lengths' and 'grouping' are not user-facing. They are used by the
         attach method.
         """
+        # validate inputs
         if not isinstance(segments, pdarray) or segments.dtype != akint64:
             raise TypeError("Segments must be int64 pdarray")
         if not is_sorted(segments):
@@ -106,23 +110,47 @@ class SegArray:
                 raise ValueError("Segments must start at zero and be less than values.size")
         elif values.size > 0:
             raise ValueError("Cannot have non-empty values with empty segments")
-        if copy:
-            self.segments = segments[:]
-            self.values = values[:]
-        else:
-            self.segments = segments
-            self.values = values
-        self.size = segments.size
-        self.valsize = values.size
+
+        # create the server message
+        rep_msg = generic_msg(
+            cmd="segArr-assemble",
+            args={
+                "segments": segments,
+                "values": values,
+            },
+        )
+
+        try:
+            # parse the return message from building the server object
+            fields = rep_msg.split()
+            self.name = fields[1]
+            self.dtype = fields[2]
+            self.size = int(fields[3])
+            self.ndim = int(fields[4])
+
+            # remove comma from 1 tuple with trailing comma
+            if fields[5][len(fields[5]) - 2] == ",":
+                fields[5] = fields[5].replace(",", "")
+            self.shape = [int(el) for el in fields[5][1:-1].split(",")]
+            self.itemsize = int(fields[6])
+        except Exception as e:
+            raise ValueError(e)
+
+        # Caching vars until all functionality is moved to server - these probably won't be
+        # needed once all functionality is moved to the server
+        self._values = None  # cache - use .values to access/set
+        self._segments = None  # cache - use .segments to access/set
+        self._valsize = None  # cache - use .valsize to access/set
+
+        # Compute the lengths/empty segments (This will need to be moved to the server)
         if lengths is None:
             self.lengths = self._get_lengths()
         else:
             self.lengths = lengths
-        self.dtype = values.dtype
-
         self._non_empty = self.lengths > 0
         self._non_empty_count = self._non_empty.sum()
 
+        # grouping object computation. (This will need to be moved to the server)
         if grouping is None:
             if self.size == 0 or self._non_empty_count == 0:
                 self.grouping = GroupBy(zeros(0, dtype=akint64))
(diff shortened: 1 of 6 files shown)


arkouda/segarray.py
```diff
@@ -106,23 +110,47 @@ def __init__(self, segments, values, copy=False, lengths=None, grouping=None):
...
+            fields = rep_msg.split()
+            self.name = fields[1]
+            self.dtype = fields[2]
+            self.size = int(fields[3])
+            self.ndim = int(fields[4])
+
+            # remove comma from 1 tuple with trailing comma
+            if fields[5][len(fields[5]) - 2] == ",":
+                fields[5] = fields[5].replace(",", "")
+            self.shape = [int(el) for el in fields[5][1:-1].split(",")]
```
joshmarshall1, line 134: I'm curious if this `if` hits and the commas are replaced with empty string, are you still expecting there to be something to split on?

joshmarshall1 (approved):
Just one question, but otherwise looks good

src/SegmentedArray.chpl
```diff
@@ -0,0 +1,59 @@
...
+    */
+    proc getSegArray(segments: [] int, values: [] ?t, st: borrowed SymTab): owned SegArray throws {
+        var segmentsEntry = new shared SymEntry(segments);
+        var valuesEntry = new shared SymEntry(values);
+        var segEntry = new shared SegArraySymEntry(segmentsEntry, valuesEntry, t);
+        var name = st.nextName();
+        st.addEntry(name, segEntry);
+        return getSegArray(name, st, segEntry.etype);
+    }
+
```
pierce314159, line 38: this looks good! 

We don't need this for this PR, but we should look into making `assembleSegArrayFromParts` methods. Maybe something like
```chapel
proc assembleSegArrayFromParts(offsets:GenSymEntry, values:GenSymEntry, type etype, st:borrowed SymTab): owned SegArray throws {
    var offs = toSymEntry(offsets, int);
    var vals = toSymEntry(values, etype);
    // This probably invokes a copy, but it's a temporary legacy bridge until we can pass
    // array components as a single message.
    return getSegArray(offs.a, vals.a, st);
}

proc assembleSegArrayFromParts(offsets:SymEntry, values:SymEntry, type etype, st:borrowed SymTab): owned SegArray throws {
    var segArrayEntry = new shared SegArraySymEntry(segmentsEntry, valuesEntry, etype);
    var name = st.nextName();
    st.addEntry(name, segArrayEntry);
    return getSegArray(name, st, segEntry.etype);
}
```
There might need some typechecking boilerplate and a case for segstrings, but that's left as an exercise for the reader :)

src/SegmentedMsg.chpl
```diff
@@ -12,10 +13,167 @@ module SegmentedMsg {
...
+    smLogger.debug(getModuleName(), getRoutineName(), getLineNumber(), repMsg);
+    return new MsgTuple(repMsg, MsgType.NORMAL);
+  }
+
+  /**
+  * Procedure to access the Segments of a Segmented Array
+  * This may not be needed once all functionality is server side
+  **/
+  proc getSASegmentsMsg(cmd: string, payload: string, st: borrowed SymTab): MsgTuple throws {
+    var repMsg: string = "";
```
pierce314159, line 82: i hate how we gotta create a new `SymEntry` for `getSASegments/Values` when we already have one, we just don't have the name... but like you say, these functions won't be needed once the segarray functionality is severside

src/MultiTypeSymEntry.chpl
```diff
@@ -393,6 +425,11 @@ module MultiTypeSymEntry
         return (entry: borrowed SegStringSymEntry);
     }
 
+    proc toCompositeSymEntry(entry: borrowed AbstractSymEntry) throws {
+        // TODO - need a way to manipulate this to give the type
+        return (entry: borrowed CompositeSymEntry);
+    }
+
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I don't think the parens are required, just probably left over from something. 

And no because the return type of the function cannot be figured out by the compiler due to the SegArraySymEntry being a generic type.

src/MultiTypeSymbolTable.chpl
```diff
@@ -390,6 +390,12 @@ module MultiTypeSymbolTable
                 var g:SegStringSymEntry = toSegStringSymEntry(entry);
                 return "%s %s %t %t %t %t".format(name, dtype2str(g.dtype), g.size, g.ndim, g.shape, g.itemsize);
             }
+            else if entry.isAssignableTo(SymbolEntryType.SegArraySymEntry) {
+                var g:CompositeSymEntry = toCompositeSymEntry(entry);
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Currently this is not necessary because we can access all the data needed from the CompositeSymEntry. It may be required later on, but currently it is not.

And again, `toSegArraySymEntry` cannot be created because the return type cannot be determined due the `SegArraySymEntry` being a generic type.

src/SegmentedArray.chpl
Ethan-DeBandi99: This is not possible. I built the function you recommend and due to the type checking (in order to create the proper SegArray), the return type is not able to be determined by the compiler and therefore it will not compile properly.

src/SegmentedMsg.chpl
Ethan-DeBandi99: I may have an idea to work around this in the meantime that would be useful in other areas as well, but wanted to get this up. I will look into implementation options and see if we can get around this, but for now we won't have a choice.

arkouda/segarray.py
Ethan-DeBandi99: This logic is to remove a trailing comma, so in the event of `(n,)`. However, in cases of `(n,m)` the if will not hit. This allows the logic to be used in both instances.

src/SegmentedArray.chpl
```diff
@@ -0,0 +1,59 @@
...
+            var errorMsg = "Error: Unhandled SymbolEntryType %s".format(abstractEntry.entryType);
+            saLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+            throw new Error(errorMsg);
+        }
+        var entry: SegArraySymEntry = abstractEntry: borrowed SegArraySymEntry(eltType);
+        return new owned SegArray(name, entry, eltType);
+    }
+
+    /*
+    * This version of the getSegString method takes segments and values arrays as
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Updated to `getSegArray`

arkouda/segarray.py
```diff
@@ -106,23 +110,47 @@ def __init__(self, segments, values, copy=False, lengths=None, grouping=None):
...
+            self.itemsize = int(fields[6])
+        except Exception as e:
+            raise ValueError(e)
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Issue #1708 has been added to address this.

PR #1688, opened 2022-08-18
Closes #1611 - DataFrame Register and Attach Functionality

### LINKED ISSUE #1611: DataFrame lacks Register / Attach functionality
ak.DataFrame objects lack the ability to register / attach to an Arkouda server. This can be useful for moving between analytic environments or when wanting to share dataframes between analysts both operating from a single server. While the code allows for registering / attaching of the underlying pdarray components, this can be cumbersome for analysts who are primarily working in the DataFrame model. 

The attached functions provide a working example of registering / attaching ak.DataFrames. The registration function creates three arrays and registers them on the server: 

The first tracks the column names of the dataframe, the second tracks the client side datatypes, and the third is a checksum of the columns to verify the dataframe before and after registration. Then each column pdarray is registered with an appropriate registration function. The name of the dataframe to be registered is user selected and prepended to each of the registered arrays. 

Attaching is the reverse of this process. Much of the code is to recast types and give back the original datatypes as they existed when registered. 

These might be best served to exist within the DataFrame class and .register_dataframe() could be called on self. 

**Note:** This code was retyped. Apologies if there are typos. Will try to subsequently create a feature branch, but wanted to get this off to the Arkouda team for awareness / alternative implementation methods. 

[dataframe_registration_functions.txt](https://github.com/Bears-R-Us/arkouda/files/9149708/dataframe_registration_functions.txt)

joshmarshall1:
This PR closes #1611 

Adds the basic `register` and `attach` capabilities to DataFrame:
`register(name)`
`attach(name)`
`unregister()`
`is_registered()`
`unregister_dataframe_by_name(name)`

Also adds `.objtype` property to `SegArray` to support using the objtype to distinguish between potential column types in the DataFrame to ensure proper rebuilding.

Most of this should be fairly straight forward, the only little "trick" that I had to do was in the attach function. In order to preserve the order of columns within the DataFrame I register the `dataframe.columns` property with the DataFrame and use that list to create a dictionary with the proper keys and empty values. The values are then parsed through the names of the registered column data and assigned to their respective location in the dictionary. This prevents index misalignment issues and allows the original DataFrame and newly attached DataFrame to evaluate to equal.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1809,7 +1809,6 @@ class DataFrame(UserDict):
 
         return DataFrame(df_def, index=self.index)
 
-
     def corr(self) -> DataFrame:
         """
         Return new DataFrame with pairwise correlation of columns
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1969,9 +1969,8 @@ class DataFrame(UserDict):
                 f"WARNING: DataFrame {self.name} expected {total} components to be registered,"
                 f" but only located {registered}."
             )
-            return False
-        else:
-            return registered == total
+
+        return registered == total
 
     @staticmethod
     def attach(user_defined_name: str) -> DataFrame:
@@ -1999,18 +1998,15 @@ class DataFrame(UserDict):
         register, is_registered, unregister, unregister_groupby_by_name
         """
 
-        registry = list_registry()
-
-        col_resp = cast(str, generic_msg(cmd="attach", args=f"df_columns_{user_defined_name}"))
-        cols = Strings.from_return_msg(col_resp)
-        columns = dict.fromkeys(cols.to_list())
+        col_resp = cast(str, generic_msg(cmd="stringsToJSON", args=f"df_columns_{user_defined_name}"))
+        columns = dict.fromkeys(json.loads(col_resp))
         matches: List[str] = []
         regEx = compile(
             f"^df_data_[a-zA-Z0-9]+_({pdarray.objtype}|{Strings.objtype}|"
             f"{Categorical.objtype}|{SegArray.objtype})_{user_defined_name}"
         )
         # Using the regex, cycle through the registered items and find all the columns in the DataFrame
-        for name in registry:
+        for name in list_registry():
             x = match(regEx, name)
             if x is not None:
                 matches.append(x.group())
@@ -2020,7 +2016,7 @@ class DataFrame(UserDict):
             raise RegistrationError(f"No registered elements with name '{user_defined_name}'")
 
         # Remove duplicates caused by multiple components in Categorical or SegArray
-        dedupe = list(dict.fromkeys(matches))
+        dedupe = set(matches)
 
         for name in dedupe:
             colName = name.split("_")[2]
@@ -2073,14 +2069,13 @@ class DataFrame(UserDict):
         register, unregister, attach, is_registered
         """
 
-        registry = list_registry()
         matches = []
         regEx = compile(
             f"^df_data_[a-zA-Z0-9]+_({pdarray.objtype}|{Strings.objtype}|"
             f"{Categorical.objtype}|{SegArray.objtype})_{user_defined_name}"
         )
         # Using the regex, cycle through the registered items and find all the columns in the DataFrame
-        for name in registry:
+        for name in list_registry():
             x = match(regEx, name)
             if x is not None:
                 matches.append(x.group())
@@ -2090,16 +2085,13 @@ class DataFrame(UserDict):
             raise RegistrationError(f"No registered elements with name '{user_defined_name}'")
 
         # Remove duplicates caused by multiple components in categorical or segarray
-        dedupe = list(dict.fromkeys(matches))
+        dedupe = set(matches)
 
         for name in dedupe:
-            if f"_{Strings.objtype}_" in name or f"_{pdarray.objtype}_" in name:
-                cols_resp = cast(str, generic_msg(cmd="attach", args=name))
-                dtype = cols_resp.split()[2]
-                if dtype == Strings.objtype:
-                    Strings.unregister_strings_by_name(name)
-                else:  # pdarray
-                    unregister_pdarray_by_name(name)
+            if f"_{Strings.objtype}_" in name:
+                Strings.unregister_strings_by_name(name)
+            elif f"_{pdarray.objtype}_" in name:
+                unregister_pdarray_by_name(name)
             elif f"_{Categorical.objtype}_" in name:
                 Categorical.unregister_categorical_by_name(name)
             elif f"_{SegArray.objtype}_" in name:
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1882,6 +1882,9 @@ class DataFrame(UserDict):
         -----
         Objects registered with the server are immune to deletion until
         they are unregistered.
+
+        Any changes made to a DataFrame object after registering with the server may not be reflected
+        in attached copies.
         """
 
         self.index.register(f"df_index_{user_defined_name}")
@@ -1953,11 +1956,7 @@ class DataFrame(UserDict):
         # Total number of registered parts should equal number of columns plus an entry
         # for the index and the column order
         total = len(self.data) + 2
-        registered = 0
-
-        for col, data in self.data.items():
-            if data.is_registered():
-                registered += 1
+        registered = sum([data.is_registered() for col, data in self.data.items()])
 
         if self.index.values.is_registered():
             registered += 1
@@ -2000,7 +1999,7 @@ class DataFrame(UserDict):
 
         col_resp = cast(str, generic_msg(cmd="stringsToJSON", args=f"df_columns_{user_defined_name}"))
         columns = dict.fromkeys(json.loads(col_resp))
-        matches: List[str] = []
+        matches = []
         regEx = compile(
             f"^df_data_[a-zA-Z0-9]+_({pdarray.objtype}|{Strings.objtype}|"
             f"{Categorical.objtype}|{SegArray.objtype})_{user_defined_name}"
@@ -2010,15 +2009,13 @@ class DataFrame(UserDict):
             x = match(regEx, name)
             if x is not None:
                 matches.append(x.group())
-        matches.sort()
 
         if len(matches) == 0:
             raise RegistrationError(f"No registered elements with name '{user_defined_name}'")
 
-        # Remove duplicates caused by multiple components in Categorical or SegArray
-        dedupe = set(matches)
-
-        for name in dedupe:
+        # Remove duplicates caused by multiple components in Categorical or SegArray and
+        # loop through
+        for name in set(matches):
             colName = name.split("_")[2]
             if f"_{Strings.objtype}_" in name or f"_{pdarray.objtype}_" in name:
                 cols_resp = cast(str, generic_msg(cmd="attach", args=name))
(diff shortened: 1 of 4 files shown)


arkouda/dataframe.py
```diff
@@ -1841,6 +1844,269 @@ def numeric_help(d):
...
+        See Also
+        --------
+        register, is_registered, unregister, unregister_groupby_by_name
+        """
+
+        registry = list_registry()
+
+        col_resp = cast(str, generic_msg(cmd="attach", args=f"df_columns_{user_defined_name}"))
+        cols = Strings.from_return_msg(col_resp)
+        columns = dict.fromkeys(cols.to_list())
```
Ethan-DeBandi99, line 2002: Can we not just return a list of columns from chapel? Since we do not need the strings object here, we are doing more than is required. Something along the lines of 
```chapel
var x: [0..#3] string = ["abc", "123", "def"];
writeln("%jt".format(x));
```
This way the python can just run `json.loads` instead of creating a strings object and then calling `to_list`.

arkouda/dataframe.py
```diff
@@ -1841,6 +1844,269 @@ def numeric_help(d):
...
+        """
+
+        registry = list_registry()
+        matches = []
+        regEx = compile(
+            f"^df_data_[a-zA-Z0-9]+_({pdarray.objtype}|{Strings.objtype}|"
+            f"{Categorical.objtype}|{SegArray.objtype})_{user_defined_name}"
+        )
+        # Using the regex, cycle through the registered items and find all the columns in the DataFrame
+        for name in registry:
```
Ethan-DeBandi99, line 2079: could we just do `for name in registry` and then remove ln 2072?

arkouda/dataframe.py
```diff
@@ -1841,6 +1844,269 @@ def numeric_help(d):
...
+            x = match(regEx, name)
+            if x is not None:
+                matches.append(x.group())
+        matches.sort()
+
+        if len(matches) == 0:
+            raise RegistrationError(f"No registered elements with name '{user_defined_name}'")
+
+        # Remove duplicates caused by multiple components in categorical or segarray
+        dedupe = list(dict.fromkeys(matches))
```
Ethan-DeBandi99, line 2089: Should this be a unique set if we are performing deduplication?

arkouda/dataframe.py
```diff
@@ -1841,6 +1844,269 @@ def numeric_help(d):
...
+                    Strings.unregister_strings_by_name(name)
+                else:  # pdarray
+                    unregister_pdarray_by_name(name)
+            elif f"_{Categorical.objtype}_" in name:
+                Categorical.unregister_categorical_by_name(name)
+            elif f"_{SegArray.objtype}_" in name:
+                # SegArray does not have an unregister_by_name function. By attaching and using
+                # unregister() it prevents unregistering the components individually here
+                segarray = SegArray.attach(name)
+                segarray.unregister()
```
Ethan-DeBandi99, line 2105: My brain does not like the idea of attaching for the purpose of unregistering, which seems to happen in a few of the cases. I believe the first block could be reorganized to something like
```python
if f"_{Strings.objtype}_" in name:
    Strings.unregister_strings_by_name(name)
elif f"_{pdarray.objtype}_" in name:
    unregister_pdarray_by_name(name)
```
Since we are in the process of moving SegArray to the server, this is totally fine for now because it will most likely require updates for that case in the near future.

Ethan-DeBandi99 (changes requested):
Just a few organizational things to hopefully eliminate some unnecessary processing. I may be mistaken on some things, but if that is the case, clarification would be great!

arkouda/dataframe.py
```diff
@@ -1841,6 +1844,269 @@ def numeric_help(d):
...
+            registered += 1
+
+        if 0 < registered < total:
+            warn(
+                f"WARNING: DataFrame {self.name} expected {total} components to be registered,"
+                f" but only located {registered}."
+            )
+            return False
+        else:
+            return registered == total
```
pierce314159, line 1970: we can just return `registered == total`

arkouda/dataframe.py
```diff
@@ -1841,6 +1844,269 @@ def numeric_help(d):
...
+        cols = Strings.from_return_msg(col_resp)
+        columns = dict.fromkeys(cols.to_list())
+        matches: List[str] = []
+        regEx = compile(
+            f"^df_data_[a-zA-Z0-9]+_({pdarray.objtype}|{Strings.objtype}|"
+            f"{Categorical.objtype}|{SegArray.objtype})_{user_defined_name}"
+        )
+        # Using the regex, cycle through the registered items and find all the columns in the DataFrame
+        for name in registry:
+            x = match(regEx, name)
```
pierce314159, line 2010: `registry` is only used once, we can get away with just doing `for name in list_registry()` here

arkouda/dataframe.py
joshmarshall1: Implemented `stringsToJSONMsg` return to enable this functionality

arkouda/dataframe.py
joshmarshall1: Updated

arkouda/dataframe.py
joshmarshall1: Switched this to use `set()`

arkouda/dataframe.py
joshmarshall1: This was reused logic from the attach case that isn't needed here - good catch

arkouda/dataframe.py
joshmarshall1: Updated

arkouda/dataframe.py
joshmarshall1: Updated

arkouda/dataframe.py
Ethan-DeBandi99: Do we have an issue in to add an unregister by name to segarray?

Ethan-DeBandi99 (approved):
Looks good. Do we already have an issue in to add an `unregister_by_name` function to `SegArray`? If not, can we add one?

joshmarshall1:
If not, I'll add an issue for it

arkouda/dataframe.py
```diff
@@ -1845,6 +1848,261 @@ def numeric_help(d):
...
+            return False
+
+        # Total number of registered parts should equal number of columns plus an entry
+        # for the index and the column order
+        total = len(self.data) + 2
+        registered = 0
+
+        for col, data in self.data.items():
+            if data.is_registered():
+                registered += 1
```
pierce314159, line 1960: this way is totally fine but i wanted to throw out this alternative
```python
registered = sum([data.is_registered() for col, data in self.data.items()])
```

arkouda/dataframe.py
```diff
@@ -1845,6 +1848,261 @@ def numeric_help(d):
...
+        regEx = compile(
+            f"^df_data_[a-zA-Z0-9]+_({pdarray.objtype}|{Strings.objtype}|"
+            f"{Categorical.objtype}|{SegArray.objtype})_{user_defined_name}"
+        )
+        # Using the regex, cycle through the registered items and find all the columns in the DataFrame
+        for name in list_registry():
+            x = match(regEx, name)
+            if x is not None:
+                matches.append(x.group())
+        matches.sort()
```
pierce314159, line 2013: we could do something like this
```python
matches = builtins.sorted([x.group() for name in list_registry() if (x := match(regEx, name)) is not None])
```
which would eliminate the need for explicitly typing `matches` above

Note this requires us to `import builtins` because at some point in the past we overrode `sorted` from `builtins`... which I don't love

arkouda/dataframe.py
```diff
@@ -1845,6 +1848,261 @@ def numeric_help(d):
...
+
+        if len(matches) == 0:
+            raise RegistrationError(f"No registered elements with name '{user_defined_name}'")
+
+        # Remove duplicates caused by multiple components in Categorical or SegArray
+        dedupe = set(matches)
+
+        for name in dedupe:
```
pierce314159, line 2021: we only use `dedupe` once, so we can prob just do `for name in set(matches):`

arkouda/dataframe.py
```diff
@@ -1845,6 +1848,261 @@ def numeric_help(d):
...
+                matches.append(x.group())
+        matches.sort()
+
+        if len(matches) == 0:
+            raise RegistrationError(f"No registered elements with name '{user_defined_name}'")
+
+        # Remove duplicates caused by multiple components in categorical or segarray
+        dedupe = set(matches)
+
+        for name in dedupe:
```
pierce314159, line 2090: we can prob just do `for name in set(matches):`

src/SegmentedMsg.chpl
```diff
@@ -1178,6 +1178,19 @@ module SegmentedMsg {
...
   }
+
+  proc stringsToJSONMsg(cmd: string, name: string, st: borrowed SymTab): MsgTuple throws {
+    var strings = getSegString(name, st);
+    var size = strings.size;
+    var rtn: [0..#size] string;
+
+    for i in 0..#size {
+      rtn[i] = strings[i];
+    }
```
pierce314159, line 1189: is there a reason not to do a `forall` here?

arkouda/dataframe.py
```diff
@@ -1845,6 +1848,261 @@ def numeric_help(d):
         ret_dict = json.loads(generic_msg(cmd="corrMatrix", args=args))
         return DataFrame({c: create_pdarray(ret_dict[c]) for c in self.columns})
 
+    @typechecked
+    def register(self, user_defined_name: str) -> DataFrame:
+        """
+        Register this DataFrame object and underlying components with the Arkouda server
+
+        Parameters
```
pierce314159, line 1856: This is more a quirk having to do with dataframe having some `inplace` functionality than anything to do with register functionality, but we should prob let the user know that any `inplace` changes made to a dataframe once it's been registered will not appear in an attached copy. I'm not sure how to note this, @Ethan-DeBandi99 what do u think

pierce314159 (commented):
This functionality looks really good! The only reason I'm not approving rn is potentially adding docs about `inplace` changes to a registered dataframe

src/SegmentedMsg.chpl
joshmarshall1: Order matters. As far as I could find, there's no way to guarantee the order of the list matches the occurrence of the strings in the Strings object when using a forall loop.

pierce314159: I'm pretty sure that shouldn't matter here because we're explicitly specifying the position in both arrays and every iteration is independent of the others and doesn't rely on the result from those iterations.

If we had something like:
```chapel
forall i in 0..#size {
  rtn += strings[i];
}
```
then order would certainly matter because then the `rtn` array depends on the order we visit the indices in the range `0..#size` (i.e. if the first iteration of our parallel loop is `i=27`, then the first index of `rtn` will be `strings[27]`

but since we are specifying `rtn[i]` this won't matter. Because if the first iteration of parallel loop is `i=27` then `rtn[27] = strings[27]`, so every index still has the right thing written into it. We don't need to know the result of any other loop iterations, so the order that we iterate through the loop shouldn't matter

joshmarshall1: You're right, updated to use `forall`

arkouda/dataframe.py
joshmarshall1: Updated

arkouda/dataframe.py
joshmarshall1: Updated

arkouda/dataframe.py
joshmarshall1: Updated

arkouda/dataframe.py
pierce314159: Sounds good! The overriding then `__builtin__.sorted` has been captured in https://github.com/Bears-R-Us/arkouda/issues/1734

pierce314159 (approved):
Looks good!

PR #1693, opened 2022-08-19
Closes #1692: Use assertAlmostEqual in stats test

### LINKED ISSUE #1692: Stats test sporadically failing
The stats test uses `assertEqual` on float equality, which causes sporadic failures, so we should switch to `assertAlmostEqual`.

bmcdonald3:
Different functions from the stats test have sporadically been
failing for us on nightly testing due to small errors with
float equality. Switching everything to `assertAlmostEqual`
instead would resolve this without sacrificing much, so that
is what this PR does.

Closes: #1692

### DIFF
--- a/tests/stats_test.py
+++ b/tests/stats_test.py
@@ -26,26 +26,26 @@ class StatsTest(ArkoudaTest):
         self.c = ak.Categorical(self.s)
 
     def test_mean(self):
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Awesome! Thanks Ben

Issue #1616, opened 2022-07-22
Convert Remaining Messages to JSON


Ethan-DeBandi99:
Once #1587 is completed, the remaining messages will need to be reformatted to send arguments as JSON. 

Some messages are being updated with the initial work. Once those are done, this issue should be updated with a list of messages to update.

# Files in `src` to Modify
*Only tracking `src` because multiple Python functions may require modification per `src` file*
- [x] ArgSortMsg.chpl
- [x] BroadcastMsg.chpl 
- [x] CastMsg.chpl
- [x] ConcatenateMsg.chpl
- [x] DataFrameIndexingMsg.chpl
- [x] EfuncMsg.chpl
- [x] FindSegmentsMsg.chpl
- [x] FlattenMsg.chpl
- [x] HDF5Msg.chpl
- [x] In1dMsg.chpl
- [x] IndexingMsg.chpl
- [x] JoinEqWithDTMsg.chpl
- [x] KEtremeMsg.chpl
- [x] MetricsMsg.chpl
- [x] OperatorMsg.chpl
- [x] ParquetMsg.chpl
- [x] RandMsg.chpl
- [x] ReductionMsg.chpl
- [x] RegistrationMsg.chpl
- [x] SegmentedMsg.chpl
- [x] SequenceMsg.chpl
- [x] SortMsg.chpl
- [x] StatsMsg.chpl
- [x] UniqueMsg.chpl

Ethan-DeBandi99:
Closing since all supporting issues have been merged.

PR #1712, opened 2022-08-23
Closes #1711 - `FlattenMsg.chpl` JSON Message Arguments

### LINKED ISSUE #1711: `FlattenMsg.chpl` JSON Message Arguments
Part of #1616 

Update `FlattenMsg.chpl` to use JSON message arguments in place of string arguments.

Ethan-DeBandi99:
Closes #1711 
Part of #1616 

Updates `segmentedFlatten` and `segmentedSplit` messages to use JSON message arguments.

### DIFF
--- a/arkouda/matcher.py
+++ b/arkouda/matcher.py
@@ -113,10 +113,19 @@ class Matcher:
         if re.search(self.pattern, ""):
             raise ValueError("Cannot split or flatten with a pattern that matches the empty string")
         cmd = "segmentedSplit"
-        args = "{} {} {} {} {}".format(
-            self.objtype, self.parent_entry_name, maxsplit, return_segments, json.dumps([self.pattern])
+        repMsg = cast(
+            str,
+            generic_msg(
+                cmd=cmd,
+                args={
+                    "parent_name": self.parent_entry_name,
+                    "objtype": self.objtype,
+                    "max": maxsplit,
+                    "return_segs": return_segments,
+                    "pattern": [self.pattern],
+                },
+            ),
         )
-        repMsg = cast(str, generic_msg(cmd=cmd, args=args))
         if return_segments:
             arrays = repMsg.split("+", maxsplit=2)
             return Strings.from_return_msg("+".join(arrays[0:2])), create_pdarray(arrays[2])
--- a/arkouda/matcher.py
+++ b/arkouda/matcher.py
@@ -122,7 +122,7 @@ class Matcher:
                     "objtype": self.objtype,
                     "max": maxsplit,
                     "return_segs": return_segments,
-                    "pattern": [self.pattern],
+                    "pattern": self.pattern,
                 },
             ),
         )
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1121,10 +1121,19 @@ class Strings:
             return self.split(delimiter, return_segments=return_segments)
         else:
             cmd = "segmentedFlatten"
-            args = "{} {} {} {} {}".format(
-                self.entry.name, self.objtype, return_segments, regex, json.dumps([delimiter])
+            repMsg = cast(
+                str,
+                generic_msg(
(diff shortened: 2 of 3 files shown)


arkouda/strings.py
```diff
@@ -1121,10 +1121,19 @@ def flatten(
...
+            repMsg = cast(
+                str,
+                generic_msg(
+                    cmd=cmd,
+                    args={
+                        "values": self.entry,
+                        "objtype": self.objtype,
+                        "return_segs": return_segments,
+                        "regex": regex,
+                        "delim": [delimiter],
```
pierce314159, line 1133: did a deep dive with git blame on why we were doing a JSON dumps on `[delimiter]`. I thought I did this but it predates the regex code and has been like this at least two years

I think that was probably because the delimiter can contain spaces which would break the old message passing. I *think* with the new message passing, we can just do `"delim": delimiter` here since everything is json

arkouda/matcher.py
```diff
@@ -113,10 +113,19 @@ def split(self, maxsplit: int = 0, return_segments: bool = False):
...
+            str,
+            generic_msg(
+                cmd=cmd,
+                args={
+                    "parent_name": self.parent_entry_name,
+                    "objtype": self.objtype,
+                    "max": maxsplit,
+                    "return_segs": return_segments,
+                    "pattern": [self.pattern],
+                },
```
pierce314159, line 126: I was def just modeling after `flatten`. I think we can also just do `"pattern": self.pattern` here

src/FlattenMsg.chpl
```diff
@@ -13,17 +13,18 @@ module FlattenMsg {
...
-    var (name, objtype, returnSegsStr, regexStr, delimJson) = payload.splitMsgToTuple(5);
-    const returnSegs: bool = returnSegsStr.toLower() == "true";
-    const regex: bool = regexStr.toLower() == "true";
-    const arr = jsonToPdArray(delimJson, 1);
+    var msgArgs = parseMessageArgs(payload, 5);
+    const objtype = msgArgs.getValueOf("objtype");
+    const returnSegs: bool = msgArgs.get("return_segs").getBoolValue();
+    const regex: bool = msgArgs.get("regex").getBoolValue();
+    const arr = msgArgs.get("delim").getList(1);
     const delim: string = arr[arr.domain.low];
```
pierce314159, line 21: if we can get away with just passing `delimiter`, I think this can be simplified to just be 
```chapel
const delim: string = msgArgs.getValueOf("delim");
```

src/FlattenMsg.chpl
```diff
@@ -48,22 +49,17 @@ module FlattenMsg {
...
+    const name = msgArgs.getValueOf("parent_name");
+    const returnSegs = msgArgs.get("return_segs").getBoolValue();
+    var maxsplit: int = msgArgs.get("max").getIntValue();
 
     // check to make sure symbols defined
     st.checkTable(name);
 
-    const json = jsonToPdArray(patternJson, 1);
+    const json = msgArgs.get("pattern").getList(1);
     const pattern: string = json[json.domain.low];
```
pierce314159, line 63: Same deal here, i think we can just do
```chapel
const pattern = msgArgs.getValueOf("pattern");
```

pierce314159 (commented):
I think we don't need the `[delimiter]` anymore now that we have the JSON message arg processing

arkouda/strings.py
Ethan-DeBandi99: I was just matching the existing. Updated this.

src/FlattenMsg.chpl
Ethan-DeBandi99: Done

src/FlattenMsg.chpl
Ethan-DeBandi99: Done

pierce314159 (approved):

PR #1714, opened 2022-08-24
Closes #1709 - `FindSegmentsMsg.chpl` JSON Message Arguments

### LINKED ISSUE #1709: `FindSegmentsMsg.chpl` JSON Message Arguments
Part of #1616 

Update `FindSegmentsMsg.chpl to use JSON message arguments in place of the current string arguments.

Ethan-DeBandi99:
Closes #1709 
Part of #1616 

Updating `FindSegments.chpl` to use JSON message arguments in place of string arguments.


src/FindSegmentsMsg.chpl
```diff
@@ -37,20 +37,13 @@ module FindSegmentsMsg
         var repMsg: string; // response message
         // split request into fields
         var (pname, nkeysStr, rest) = payload.splitMsgToTuple(3);
```
pierce314159, line 39: we don't need the `splitMsgToTuple` line anymore

pierce314159 (approved):
I don't know where we ever use this since there doesn't seem to be a client side `"findSegments"` command. I think there's one extra line that can be removed, but i don't see any issue with the JSONification of the arguments

joshmarshall1 (approved):
Aside from what Pierce already pointed out, looks good

src/FindSegmentsMsg.chpl
Ethan-DeBandi99: Removed. I believe we could probably get rid of this file though... Maybe?

pierce314159: yep! it looks like this used to be used by `GroupBy` but `findSegments` was removed in https://github.com/Bears-R-Us/arkouda/pull/1330. I think it should be safe to remove

pierce314159: I also built the server with `FindSegmentsMsg` commented out in `ServerModules.cfg` and verified all the tests pass

Ethan-DeBandi99: I will remove the file and push that to this PR.

Ethan-DeBandi99: Updated this PR to remove the File

PR #1726, opened 2022-08-29
Closes #1725 - `sqrt` and `power` functions on pdarrays

### LINKED ISSUE #1725: Add sqrt and power functions to operate on pdarray
Square root and power operations are useful; this enhancement would add them as functions to the `arkouda` module alongside `exp` etc.

`arkouda.sqrt` would be a simple element-wise operation (like `exp`). 
`arkouda.power` would have a similar interface to [`numpy.power`](https://numpy.org/doc/stable/reference/generated/numpy.power.html), accepting as the second argument either a scalar value to which all elements of the first array are raised, or an array of powers to which corresponding elements of the first array are raised.

Ethan-DeBandi99:
Closes #1725 

Added `ak.power` and `ak.sqrt` functions. These functions are just aliases for `**` and `**0.5`, respectively. 
Adds testing to ensure arkouda functionality is in alignment with `Numpy`.

### DIFF
--- a/tests/operator_tests.py
+++ b/tests/operator_tests.py
@@ -454,7 +454,7 @@ class OperatorsTest(ArkoudaTest):
             equality_helper(a_vect // ak_edge_cases, n_vect // np_edge_cases)
 
     def test_pda_sqrt(self):
-        n = np.array([4, 16, -1, 0])
+        n = np.array([4, 16, -1, 0, np.inf])
         a = ak.array(n)
 
         s = ak.sqrt(a)
@@ -479,6 +479,15 @@ class OperatorsTest(ArkoudaTest):
         p = ak.power(a, ak.array([2, 3, 4]))
         ex = np.power(n, [2, 3, 4])
         self.assertListEqual(p.to_list(), ex.tolist())
+
+        n = np.array([-1.0, -3.0])
+        a = ak.array(n)
+
+        p = ak.power(a, 0.5)
+        ex = np.power(n, 0.5)
+        nan = ak.isnan(p)
+        npnan = np.isnan(ex)
+        self.assertListEqual(nan.to_list(), npnan.tolist())
 
     def testAllOperators(self):
         run_tests(verbose)
--- a/tests/operator_tests.py
+++ b/tests/operator_tests.py
@@ -453,6 +453,33 @@ class OperatorsTest(ArkoudaTest):
             equality_helper(ak_edge_cases // a_vect, np_edge_cases // n_vect)
             equality_helper(a_vect // ak_edge_cases, n_vect // np_edge_cases)
 
+    def test_pda_sqrt(self):
+        n = np.array([4, 16, -1, 0])
+        a = ak.array(n)
+
+        s = ak.sqrt(a)
+        ex = np.sqrt(n)
+
+        # id nans
+        nan = ak.isnan(s)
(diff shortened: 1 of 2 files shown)


tests/operator_tests.py
```diff
@@ -453,6 +453,33 @@ def equality_helper(a, n):
             equality_helper(ak_edge_cases // a_vect, np_edge_cases // n_vect)
             equality_helper(a_vect // ak_edge_cases, n_vect // np_edge_cases)
 
+    def test_pda_sqrt(self):
+        n = np.array([4, 16, -1, 0])
```
pierce314159, line 457: we should add `np.inf` here to make sure it remains infinity. Perhaps `-np.inf`, `-0.0` too to ensure we're covering our edge cases well

tests/operator_tests.py
```diff
@@ -453,6 +453,33 @@ def equality_helper(a, n):
...
+        # id nans
+        nan = ak.isnan(s)
+        npnan = np.isnan(ex)
+        self.assertListEqual(nan.to_list(), npnan.tolist())
+
+        # check results are equal remove nans
+        self.assertListEqual(s[~nan].to_list(), ex[~npnan].tolist())
+
+    def test_pda_power(self):
+        n = np.array([10, 5, 2])
```
pierce314159, line 472: yeah i think testing some of the weirder edge cases would be good, maybe `inf`, `-inf`, `0.0`, and `-0.0`

pierce314159 (commented):
A couple things to look into

Also numpy has a `where` argument for both of these functions, but there's some design questions i have about it's implementation, so i think adding it is outside the scope of this PR. I've captured this in issue #1727

tests/operator_tests.py
Ethan-DeBandi99: Added

tests/operator_tests.py
pierce314159: Note chapel doesn't handle `(-INFINITY)**0.5` and `(-0.0)**0.5` correctly. These cases are correctly handled by the chapel `sqrt` function. I put up https://github.com/chapel-lang/chapel/issues/20542 to track this.

We should be able to get around this on the client in the meantime with a bit of hacky code... something like
```python
ret_array = # normal stuff
ret_array[pda == -np.inf] = np.nan
ret_array[pda== -0.0] = -0.0
```

pierce314159: digging in a bit deeper, it seems numpy also has variation here...
```python
>>> n = np.array([-0.0, -7, -np.inf])
>>> n**0.5
array([-0., nan, nan])

>>> np.power(n, 0.5)
array([ 0., nan, inf])

>>> np.sqrt(n)
array([-0., nan, nan])

>>> n**(1/3)
array([ 0., nan, inf])

>>> np.power(n, 1/3)
array([ 0., nan, inf])

>>> n**(1/4)
array([ 0., nan, inf])

>>> np.power(n, 1/4)
array([ 0., nan, inf])
```

Ethan-DeBandi99: I will be adding a ticket to track this once I push the other requested changes.

Ethan-DeBandi99: Added issue #1730

pierce314159 (approved):

joshmarshall1 (approved):

Issue #1729, opened 2022-08-29
enable client UI and API modes


hokiegeek2:
In situations where the Arkouda is used in API mode (not via shell or Jupyter notebook), the splash message can cause errors when the result of an Arkouda API prints a result. An example is in my arkouda_contrib fork [here](https://github.com/hokiegeek2/arkouda-contrib/blob/17c8039f0ea2d3ce46e536eb8f8f909b601ff035/arkouda_integration/client/scripts/pods.py#L14) where the logic pretty prints ip addresses of arkouda locales on kubernetes:

![image](https://user-images.githubusercontent.com/10785153/187402054-02a84017-696b-4d80-9c9e-9e711bd097ae.png)

If the splash message is printed on the arkouda import statement, the pretty print picks up the pyfiglet-generated banner which conflicts with the list of IP addresses. To address this, there should be a UI client mode (default) or API mode where the Arkouda client is access via an API call.

I implemented this and this is how it works:

![image](https://user-images.githubusercontent.com/10785153/187302994-09ccb779-5d23-4057-b765-2855702987f2.png)

hokiegeek2:
PR merged into master, issue resolved

PR #1732, opened 2022-08-30
Enable UI and API client mode

hokiegeek2:
This PR addresses issue #1729 via the following:

1. ClientMode enum--UI or API mode
2. logic to retrieve ClientMode value from env variable, defaulting to UI
3. logic to display pyfiglet banner if ClientMode is UI (again, the default), suppress if in API mode, which I have a use case for as described in #1729

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -74,7 +74,7 @@ class ClientMode(Enum):
 
 
 # Get ClientMode, defaulting to UI
-mode = ClientMode(os.getenv('ARKOUDA_CLIENT_MODE', 'UI'))
+mode = ClientMode(os.getenv('ARKOUDA_CLIENT_MODE', 'UI').upper())
 
 # Print splash message if in UI mode
 if mode == ClientMode.UI:
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -1,6 +1,7 @@
 import json
 import os
 import warnings
+from enum import Enum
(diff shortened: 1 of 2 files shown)


arkouda/client.py
```diff
@@ -49,9 +50,36 @@
...
+
+    def __repr__(self) -> str:
+        """
+        Overridden method returns value.
+        """
+        return self.value
+
+
+# Get ClientMode, defaulting to UI
+mode = ClientMode(os.getenv('ARKOUDA_CLIENT_MODE', 'UI'))
```
pierce314159, line 77: would it be a good idea to change this to
```python
mode = ClientMode(os.getenv('ARKOUDA_CLIENT_MODE', 'UI').upper())
```
or something like that so it case doesn't matter when setting the env var? it prob doesn't matter too much

pierce314159 (commented):
Logic looks good! Should we update the `ENVRONMENT.md` or whatever the relevant documentation is for this?

arkouda/client.py
hokiegeek2: @pierce314159 good points on both, will implement

hokiegeek2: @pierce314159 done :)

joshmarshall1 (approved):
Looks good

hokiegeek2:
@pierce314159 @joshmarshall1 thanks guys!!!

PR #1733, opened 2022-08-30
#1728 Fix nunique

### LINKED ISSUE #1728: `assume_sorted=True` no longer valid in `GroupBy.nunique()` and `.most_common()`
My giant refactor of `GroupBy` and `unique` broke the assumptions about ordering of the output used by `GroupBy.nunique()` and `.most_common()`, so that `assume_sorted=True` is no longer valid there. Here's the offending section of `.nunique`:

https://github.com/Bears-R-Us/arkouda/blob/db5fcb7562dab489d7edd93707448be397382e93/arkouda/groupbyclass.py#L837-L841

This requires two fixes:
 - remove the `assume_sorted=True`
 - check whether `g2.unique_keys` is sorted and if not, permute `nuniq` accordingly

Here's a reproducer and future unit test:

```python
a = ak.array(['1' for _ in range(8)] + ['2' for _ in range(3)])
b = ak.array([str(i) for i in range(8)] + [str(i) for i in range(3)])
g = ak.GroupBy(a)
keys, nuniq = g.nunique(b)
expected_keys = ak.array(['2', '1'])
expected_nuniq = ak.array([3, 8])
assert (nuniq == expected_nuniq).all()

# Currently raises ValueError for size mismatch
```

reuster986:
Closes #1728 

- Fixes logic of `GroupBy.nunique()`, which relied on the old sorting behavior of multi-level GroupBy that has since changed.
- Adds a unit test for the ordering bug fix
- Deprecates `GroupBy.most_common()` and makes it an alias to `GroupBy.mode()`

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -1606,32 +1606,9 @@ class GroupBy:
 
     def most_common(self, values):
         """
-        Find the most common value for each segment of a GroupBy object. This method only supports
-        array-like GroupBy key types.
-
-        Parameters
-        ----------
-        values : array-like
-            Values in which to find most common based on the GroupBy's segment indexes.
-            values.size must equal GroupBy.keys[0].size
-
-        Returns
-        -------
-        most_common_values : array-like
-            The most common value for each segment of the GroupBy
+        (Deprecated) See `GroupBy.mode()`.
         """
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good

PR #1735, opened 2022-08-30
[draft] Lisp code for collaboration

bmcdonald3:
This is a draft implementation of the lisp code being opened for easy sharing. This branch has the `ObjectPool` optimization as a work in progress with values hardcoded for the `my_axpy` function and works for the stream benchmark, but will not work generally.

To run the benchmark on this branch:

1. Make with the `ServerModules.cfg` file included (this is the minimum build for the benchmark)
2. Start the Arkouda server
3. Run `./benchmarks/stream.py <hostname> <port> --size=100000 --lisp`
    - small size since lisp code doesn't perform very well at the moment
    - `--lisp` flag runs the lisp benchmark

For a fully-functional branch that doesn't perform as well, see: https://github.com/bmcdonald3/arkouda/tree/lisp-code

Corresponding arkouda-contrib PR: https://github.com/Bears-R-Us/arkouda-contrib/pull/47


Ethan-DeBandi99:
This work is no longer being pursued. Spoke to @bmcdonald3 and confirmed this can be closed.

PR #1739, opened 2022-09-01
Closes #1681 - `SegArray.__init__` to Server

### LINKED ISSUE #1681: Move `SegArray.__init__` fully to server
Part of #1570 

`SegArray` has been moved to the server as an object. `SegArray` now exists in the symbol table. However, not all of the `__init__` has been moved. As a result, some properties are still computed on the client (additional server calls). 

Update `__init__` to move the remaining property calculation to the server to be stored within a given `SegArray` object. This will require accessors as currently exist for `.values`, `.segments` and `.valsize` until all functionality is moved.

Ethan-DeBandi99:
Closes #1681 

This PR moves the remaining `init` functionality for SegArray to the server. This includes creating messages to access these values from the server. `non_empty`, `non_empty_count` and `lengths` have been moved to the server. `grouping` was not moved because `GroupBy` objects are not server side. In the future, we may want to move GroupBy to the server and update this since most of the functionality is there with the `UniqueMsg`. In the future, we may need to move the `getLengths`, `getNonEmpty` and `getNonEmptyCount` function logic to compute properties in the init in chapel in order to prevent recomputing the values unnecessarily. 

Currently, if a `lengths` is passed to the init in python, it is used and computes `_non_empty` and `_non_empty_count` from it. A few questions here for moving forward (since this is an iteration and should not impact anything negatively at this point).
@pierce314159, @joshmarshall1, @mhmerrill, @reuster986 - if you have any thoughts, they would be greatly appreciated.
1) Should we pass the `name` of the `lengths` pdarray to the server and add the `SymEntry` as part of the `SegArraySymEntry`? 
2) Should we deprecate the `lengths` parameter in python and just recompute in the `init` on the server? (This would be my suggestion). The recompute should be minimal and this way we do not have a `SymEntry` always hanging around taking up space.

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -143,9 +143,18 @@ class SegArray:
         self._values = None  # cache - use .values to access/set
         self._segments = None  # cache - use .segments to access/set
         self._valsize = None  # cache - use .valsize to access/set
-        self._lengths = None  # cache - use .lengths to access/set
-        self._non_empty = None  # cache - use .non_empty to access/set
-        self._non_empty_count = None  # cache - use .non_empty_count to access/set
+        """
+        Note - if lengths is provided, it will need to be sent to the server in the future (or deprecated).
+        Since no computation is currently done on the server, we will not be sending lengths right now
+        """
+        self._lengths = lengths  # cache - use .lengths to access/set. If passed, do not recompute
+        # the following is to maintain support for lengths being passed in (since not currently passed to server)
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99:
I forgot to add the initialization if lengths or one of the other properties is provided. I will add this and resubmit the PR.

pierce314159 (approved):
Looks good!

And the compute in the init on server seems reasonable to me

PR #1754, opened 2022-09-06
Closes #1752: Update `README` about

### LINKED ISSUE #1752: Update README with new about
Update the `README` to match the new arkouda about info

pierce314159:
This PR (Closes #1752):
- Updates the `README` to match the github about
- Centers the logo
- Adds badges for CI, docs, license, and code style
- Changes workflow name from `gh-pages-docs` to `docs` since we only support github pages

Note in these pictures the documentation badge isn't showing up, this is because the name change in this PR would need to merged first. It will look like the `CI` badge but say `docs`

Before this PR:
image

After this PR:
image

<details>
  <summary>Other options if people prefer:</summary>

image

image
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 <p align="center">
   <img src="pictures/arkouda_wide_marker1.png"
 />
-<h1 align="center">Arkouda (αρκούδα)</h1>
-<h2 align="center">Interactive Data Analytics at Supercomputing Scale -- </br>a Python API powered by Chapel :bear:</h2>
+<h1 align="center">Arkouda (αρκούδα) :bear:</h1>
+<h2 align="center">Interactive Data Analytics at Supercomputing Scale</br>a Python API powered by Chapel</h2>
 
 <p align="center">
 <a href="https://github.com/Bears-R-Us/arkouda/actions/workflows/CI.yml"><img alt="Actions Status" src="https://github.com/Bears-R-Us/arkouda/workflows/CI/badge.svg"></a>
--- a/README.md
+++ b/README.md
@@ -1,7 +1,15 @@
-![Arkouda logo](pictures/arkouda_wide_marker1.png)
+<p align="center">
+  <img src="pictures/arkouda_wide_marker1.png"
+/>
+<h1 align="center">Arkouda (αρκούδα)</h1>
+<h2 align="center">Interactive Data Analytics at Supercomputing Scale -- </br>a Python API powered by Chapel :bear:</h2>
 
-# Arkouda (αρκούδα): NumPy-like arrays at massive scale backed by Chapel.
-## _NOTE_: Arkouda is under the MIT license.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,7 +1,14 @@
-![Arkouda logo](pictures/arkouda_wide_marker1.png)
-
-# Arkouda (αρκούδα): NumPy-like arrays at massive scale backed by Chapel.
-## _NOTE_: Arkouda is under the MIT license.
+<p align="center">
+  <img src="pictures/arkouda_wide_marker1.png"
+/>
+<h2 align="center">Arkouda (αρκούδα) :bear:</br>Interactive Data Analytics at Supercomputing Scale</br>a Python API powered by Chapel</h2>
```
Ethan-DeBandi99, line 4: Since you start another `p` tag on line 6, this one needs a closing tag. Or we can eliminate the 2nd `<p align="center">`. Either one is fine, but we should definitely have each tag open match a close.

Ethan-DeBandi99 (changes requested):
This looks great! Amazing how much a small change impacts the overall appearance. Just one small change to correct tags, but otherwise this is great.

I would like to note that I like the formatting of the 2nd or 3rd better (though I would make `Arkouda` and `h1` tag and the blurb an h2. However, I would like to see it without the dividing line between the 2 portions of the header. If that is not possible the current implementation works great. It may be something GitHubs formatting is adding because my editor does not. To be clear this is not why I am requesting changes. Just noting my thoughts.

README.md
pierce314159: good catch!

pierce314159: Removing the second `<p align="center">` removed the centering of the badges, but adding the closing tag worked like a charm! Thanks

pierce314159:
> I would like to note that I like the formatting of the 2nd or 3rd better (though I would make `Arkouda` and `h1` tag and the blurb an h2. However, I would like to see it without the dividing line between the 2 portions of the header. If that is not possible the current implementation works great. It may be something GitHubs formatting is adding because my editor does not. To be clear this is not why I am requesting changes. Just noting my thoughts.

Yeah I did a lot of digging and could not find a good way to remove the header after the `h1` tag with github markdown

Ethan-DeBandi99:
> > I would like to note that I like the formatting of the 2nd or 3rd better (though I would make `Arkouda` and `h1` tag and the blurb an h2. However, I would like to see it without the dividing line between the 2 portions of the header. If that is not possible the current implementation works great. It may be something GitHubs formatting is adding because my editor does not. To be clear this is not why I am requesting changes. Just noting my thoughts.
> 
> Yeah I did a lot of digging and could not find a good way to remove the header after the `h1` tag with github markdown

No worries. I think what you have here is way better than what it was and should be good to go.

joshmarshall1 (approved):
This looks great!

PR #1758, opened 2022-09-07
1747 add connect url to server connection info

### LINKED ISSUE #1747: add connect_url to serverConnectionInfo
The current serverConnectionInfo file contains the Arkouda connect host and port, which is good, but when authentication is enabled, the complete connect_url value is needed. Accordingly, the connect_url should be written out as well.

hokiegeek2:
This PR adds the connect_url value to the serverConnectionInfo file, which is especially handy when authentication is enabled, which adds a token param to the connect url. The test/util module is also updated accordingly

Closes #1747


Ethan-DeBandi99 (approved):
This looks good to me

pierce314159 (approved):

PR #1764, opened 2022-09-08
Closes #1763 - `SegmentedMsg.chpl` JSON Message Arguments

### LINKED ISSUE #1763: `SegmentedMsg.chpl` JSON Message Arguments
Part of #1616 

Update `SegmentedMsg.chpl` to use JSON message arguments.

Ethan-DeBandi99:
Closes #1763 

Updates `SegmentedMsg.chpl` to use JSON message arguments.

### DIFF
--- a/arkouda/matcher.py
+++ b/arkouda/matcher.py
@@ -54,14 +54,18 @@ class Matcher:
         if not self.populated or any(
             [getattr(self, pda).name not in sym_tab for pda in self.LocationsInfo]
         ):
-            cmd = "segmentedFindLoc"
-            args = "{} {} {} {}".format(
-                self.objtype,
-                self.parent_entry_name,
-                0,  # groupNum is 0 for regular matches
-                json.dumps([self.pattern]),
+            repMsg = cast(
+                str,
+                generic_msg(
+                    cmd="segmentedFindLoc",
+                    args={
+                        "objType": self.objtype,
+                        "parent_name": self.parent_entry_name,
+                        "groupNum": 0,  # groupNum is 0 for regular matches
+                        "pattern": self.pattern,
+                    },
+                ),
             )
-            repMsg = cast(str, generic_msg(cmd=cmd, args=args))
             created_map = json.loads(repMsg)
             self.num_matches = create_pdarray(created_map["NumMatches"])
             self.starts = create_pdarray(created_map["Starts"])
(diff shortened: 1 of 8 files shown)


joshmarshall1 (approved):

arkouda/pdarraycreation.py
```diff
@@ -1009,13 +1009,13 @@ def random_strings_lognormal(
...
-            NUMBER_FORMAT_STRINGS["float64"].format(logmean),
-            NUMBER_FORMAT_STRINGS["float64"].format(logstd),
-            seed,
-        ),
+        args={
+            "size": NUMBER_FORMAT_STRINGS["int64"].format(size),
+            "dist": "lognormal",
+            "chars": characters,
+            "min": NUMBER_FORMAT_STRINGS["float64"].format(logmean),
+            "max": NUMBER_FORMAT_STRINGS["float64"].format(logstd),
```
pierce314159, line 1017: it's unfortunate that these arguments mean different things depending on the distribution, maybe we could name them `min/mean` and `max/std`? Or just generic `argstr1/2` like they were previously in the chapel code? I'm also fine with them as is

arkouda/strings.py
```diff
@@ -562,8 +590,9 @@ def strip(self, chars: Optional[Union[bytes, str_scalars]] = "") -> Strings:
         """
         if isinstance(chars, bytes):
             chars = chars.decode()
-        args = {"objType": self.objtype, "name": self.entry.name, "chars": chars}
-        rep_msg = generic_msg(cmd="segmentedStrip", args=args)
+        rep_msg = generic_msg(
+            cmd="segmentedStrip", args={"objType": self.objtype, "name": self.entry.name, "chars": chars}
```
pierce314159, line 594: the `.name` on `self.entry.name` in not needed right?

arkouda/strings.py
```diff
@@ -1227,20 +1259,21 @@ def peel(
...
-            NUMBER_FORMAT_STRINGS["bool"].format(regex),
-            json.dumps([delimiter]),
+        repMsg = generic_msg(
+            cmd="segmentedPeel",
+            args={
+                "subcmd": "peel",
+                "objType": self.objtype,
+                "obj": self.entry,
+                "valType": "str",
+                "val": NUMBER_FORMAT_STRINGS["int64"].format(times),
```
pierce314159, line 1269: I know it was already `val` in the chapel code, but I think we should rename this to `times`. It immediately gets turned into `times` and something else becomes `val`, so I think making the name change sooner is clearer

pierce314159 (approved):
Looks good! I left some small comments but nothing important

arkouda/matcher.py
```diff
@@ -165,16 +173,20 @@ def sub(self, repl: str, count: int = 0, return_num_subs: bool = False):
...
-            count,
-            return_num_subs,
-            json.dumps([self.pattern]),
+        repMsg = cast(
+            str,
+            generic_msg(
+                cmd="segmentedSub",
+                args={
+                    "objType": self.objtype,
+                    "obj": self.parent_entry_name,
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I kept this as obj to match the schema I used for segmented messages. I can change it, but thought it made more sense to have `objType` and `obj`. Let me know what is preferred and I can update if needed.

arkouda/pdarraycreation.py
Ethan-DeBandi99: True, I will update it. I am leaning towards the more generic `arg1`, `arg2`.

arkouda/strings.py
Ethan-DeBandi99: Correct. I will remove that.

arkouda/strings.py
Ethan-DeBandi99: Yeah, good idea.

Ethan-DeBandi99: Updated to `times`

arkouda/strings.py
Ethan-DeBandi99: Removed.

arkouda/pdarraycreation.py
Ethan-DeBandi99: Updated from `min` and `max` to `arg1` and `arg2`.

PR #1770, opened 2022-09-09
Closes #1742 - Symbol Table Structure Update

### LINKED ISSUE #1742: Update Symbol Table Organization
This issue stems from moving `SegArray` to the server.

In moving `SegArray` to the server, the `CompositeSymEntry` required several updates. This made `GenSymEntry` and `CompositeSymEntry` functionally equivalent with the exception that `CompositeSymEntry` indicates that the entry is made up of 2 or more `SymEntry` objects. Looking at the current structure, it appears to make more sense if `CompositeSymEntry` and `SymEntry` are on the same level and both inherit `GenSymEntry`. This would only require minor adjustments to the current Symbol Table structure. Additionally, the `attrib` function could then be simplified because there is no longer a need to differentiate between `GenSymEntry` and `CompositeSymEntry`.

This should have no impact on functionality, but rather is an organizational simplification for long term ease of maintenance.

Ethan-DeBandi99:
Closes #1742

Updated `CompositeSymEntry` to inherit `GenSymEntry` instead of `AbstractSymEntry` since the properties of `CompositeSymEntry` and `GenSymEntry` were identical. By doing this, the `attrib` function now only requires a single `if`.

Overall, this update was made to slim down the code and remove duplicate code for easier long term maintenance.

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -302,23 +302,13 @@ module MultiTypeSymEntry
      * The base parent class for SymbolEntries that are composites
      * of other entry types.
      */
-    class CompositeSymEntry:AbstractSymEntry {
-        var dtype: DType; // answer to numpy dtype
-        var itemsize: int; // answer to numpy itemsize = num bytes per elt
-        var size: int = 0; // answer to numpy size == num elts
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
Looks good!

joshmarshall1 (approved):
This looks much cleaner!

PR #1773, opened 2022-09-12
Closes #1731 - JSON Message Arguments in Chapel

### LINKED ISSUE #1731: Message Argument/JSON Formatting in Chapel
Currently, if we need to call a message from another message with differing arguments from those passed to the first message, we need to manually build them. We need to add a helper function to generate a map of that can be used to generate the message.

This is not currently done enough to make the need for helpers immediate, but it could be useful. This issue is to look into/configure these helpers.

**Example**
```chapel
var arrayMap = new map(string, string);
arrayMap.add("key", "array");
arrayMap.add("val", pdaName);
arrayMap.add("objType", "PDARRAY");
arrayMap.add("dtype", "");

var valMap = new map(string, string);
valMap.add("key", "value");
valMap.add("val", value);
valMap.add("objType", "VALUE");
valMap.add("dtype", dtypeStr);

var json = ["%jt".format(arrayMap), "%jt".format(valMap)];
var new_payload = "%jt".format(json);

### PR DESCRIPTION
**Pull request by Ethan-DeBandi99**
Closes #1731 

Updates Chapel side of message argument interaction to be better positioned to handle creation of JSON message args. This was previously fully supported on the client, but any subsequent calls on the Chapel side would have required substantially more code to generate message args if the arguments were not identical to the top level message.

Updates instances where JSON arguments were being generated by Chapel.

### DIFF
--- a/src/Message.chpl
+++ b/src/Message.chpl
@@ -5,6 +5,7 @@ module Message {
     use ServerErrors;
     use NumPyDType;
     use Map;
+    use List;
 
     enum MsgType {NORMAL,WARNING,ERROR}
     enum MsgFormat {STRING,BINARY}
@@ -55,7 +56,7 @@ module Message {
 
         proc init() {}
 
-        proc init(key: string, val: string, objType: string, dtype: string) {
+        proc init(key: string, val: string, objType: ObjectType, dtype: string) {
             this.key = key;
             this.val = val;
             this.objType = objType;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Message.chpl [hunk 1]
```diff
@@ -245,6 +262,22 @@ module Message {
             this.size = param_list.size;
         }
 
+        proc getJSON(keys: list(string) = list(string)): string throws {
+            var s: int = if keys.isEmpty() then this.size else keys.size;
+            var json: [0..#s] string;
+            var idx: int = 0;
+            for p in this.param_list {
+                if (keys.isEmpty() || keys.contains(p.key)) {
```
pierce314159, line 270: saving `keys.isEmpty` into a boolean const would save the recompute

pierce314159 (approved):
This is great!!

src/Message.chpl
Ethan-DeBandi99: Updated.

pierce314159 (approved):

PR #1777, opened 2022-09-13
Closes #1774 - Message Function Definition Update

### LINKED ISSUE #1774: Message Function Definition Update
Related to #1617 

Because of Chapel's handling of generic functions, the message functions cannot be updated to take in a `MessageArgs` object as a parameter. 

We can however update the `akMsgSign` to add a `size: int` parameter and update all of the message function with the same. This will allow use to update `parseMessageArgs(payload, size)`. This will allow for dynamic message sizes and prevent the need for the developer to provide the number.

Ethan-DeBandi99:
Closes #1774 

Due to limitations with `generic` functions when dealing with the message parsing, we are not currently able to send `MessageArgs` objects directly into the functions. This prevents us from pre-parsing the JSON. This PR is intended to simplify parsing of the arguments for the time being.

This updates function in the command map to take the following parameters:
```chapel
proc exampleMsg(cmd: string, payload: string, argSize: int, st: borrowed SymTab): MsgTuple throws {}
```
Notice the addition of the `argSize` parameter. This parameter indicates the number of parameters in the JSON object. It is sent as part of the message by the client. This prevents the developer from having to manually enter the size. Additionally, this may provide the opportunity for messages with varying parameter numbers in the same message.

This is purely a convenience feature for developers, but simplifies the messaging. 

I know this is a lot of files, but we had to do everything in one shot to avoid issues with message parsing.

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -57,8 +57,9 @@ class ClientMode(Enum):
     Arkouda client is in UI mode or API mode. If in API mode, it is assumed the
     Arkouda client is being used via an API call instead of a Python shell or notebook.
     """
-    UI = 'UI'
-    API = 'API'
+
+    UI = "UI"
+    API = "API"
 
     def __str__(self) -> str:
         """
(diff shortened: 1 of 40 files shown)


joshmarshall1 (approved):

pierce314159 (approved):
Looks good! This will be a nice quality of life improvement when adding arguments to existing methods!

Issue #1685, opened 2022-08-18
Update `list.extend` to `list.append` in `MetricMsg.chpl`


Ethan-DeBandi99:
After building Arkouda following the metrics integration, I noticed that we are getting the warning below

```
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:191: In function 'exportAllMetrics':
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:194: warning: list.extend is deprecated, please use list.append
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:195: warning: list.extend is deprecated, please use list.append
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:196: warning: list.extend is deprecated, please use list.append
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:197: warning: list.extend is deprecated, please use list.append
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:91: In method 'getPerUserNumRequestsPerCommandForAllUsersMetrics':
/Users/ethandebandi/Documents/git/arkouda//src/MetricsMsg.chpl:94: warning: list.extend is deprecated, please use list.append
```

We should update this to remove the deprecation warning.

Ethan-DeBandi99:
This cannot be updated until `Chapel 1.27.0` is the lowest supported version by Arkouda.

ronawho:
FWIW this can be worked around by just explicitly over iterating over the list and calling append on individual elements (e.g. instead of appending the entire list at once, just iterate over the list and append element at a time.)

This is done in https://github.com/Bears-R-Us/arkouda/pull/1782

PR #1782, opened 2022-09-13
Closes #1781 - Quiet deprecation warnings in prep for Chapel 1.28

### LINKED ISSUE #1781: Quiet deprecation warnings for chapel 1.28
There are several new deprecation warnings with the 1.28 release. Quiet them in preparation for upgrading to 1.28 when it's released.

ronawho:
There are a number of new deprecations in Chapel 1.28. Quiet them here while maintaining compatibly (and quiet builds) for 1.26/1.27. The specific changes are:
- Handle DateTime deprecation: The `DateTime` module has been deprecated in favor of `Time`. Extend the existing `ArkoudaDateTimeCompat` module to paper over the transition period.
- Handle SysError deprecation: The `SysError` module has been deprecated, but it turns out Arkouda didn't actually need it anymore, so remove the remaining `use`s.
- Handle isFloatType deprecation: The `isFloatType` function is deprecated in favor of a more explict `isRealType || isImagType` query, but Arkouda only supports `real`s anyways, so update the checks to just look for real's.
- Handle owned->shared class assignment deprecation: Assignment from owned to shared has been deprecated, so update Arkouda's use to explicitly create shared objects in the first place instead of getting it through assignment (implicit in the function return)
- Stop using `start` for file.reader: `file.reader` has changed from explicit ints for `start`/`stop` to a range to describe the `region`. The few uses in arkouda were specifying a start of 0, which is the default so we can just stop specifying it use the default.
- Handle file.writer start -> region change: `file.writer` has changed from explicit ints for `start`/`stop` to a range to describe the `region`. Use explicit version checks to select which version to use in order to compatible across versions.
- Stop relying on `<~>` operator: The `<~>` operator has been deprecated, so stop using it an just rely on explicit `write` calls.
- Avoid using the deprecated `list.extend`: `list.extend` was deprecated in favor of a `list.append` overload, but that requires at least chapel 1.27, so to maintain compatibility to 1.26 just explicitly loop over and do individual `append` calls.

Resolves https://github.com/Bears-R-Us/arkouda/issues/1685
Resolves https://github.com/Bears-R-Us/arkouda/issues/1781

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -266,9 +266,14 @@ ifeq ($(shell expr $(CHPL_MINOR) \< 27),1)
 	ARKOUDA_COMPAT_MODULES += -M $(ARKOUDA_SOURCE_DIR)/compat/lt-127
 endif
 
-ifeq ($(shell expr $(CHPL_MINOR) \>= 27),1)
-	ARKOUDA_COMPAT_MODULES += -M $(ARKOUDA_SOURCE_DIR)/compat/ge-127
+ifeq ($(shell expr $(CHPL_MINOR) \= 27),1)
+	ARKOUDA_COMPAT_MODULES += -M $(ARKOUDA_SOURCE_DIR)/compat/e-127
 endif
+
+ifeq ($(shell expr $(CHPL_MINOR) \>= 28),1)
+	ARKOUDA_COMPAT_MODULES += -M $(ARKOUDA_SOURCE_DIR)/compat/ge-128
+endif
+
 
 MODULE_GENERATION_SCRIPT=$(ARKOUDA_SOURCE_DIR)/serverModuleGen.py
 # This is the main compilation statement section
--- a/src/CastMsg.chpl
+++ b/src/CastMsg.chpl
@@ -6,7 +6,6 @@ module CastMsg {
   use ServerErrors;
   use Logging;
   use Message;
-  use SysError;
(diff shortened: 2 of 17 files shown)


Ethan-DeBandi99 (approved):
Changes all look good to me.

I do have one question though, I believe the plan was for Arkouda to support only 1 previous version moving forward and if so, we don't need support for `1.26` once `1.28` is released. If that is still true, I believe we will be able to update the `list.extend`/`list.append` stuff to replace `extend` with `append` instead of needing the work around of looping and appending individually. This can obviously be done after the fact. @mhmerrill - are we still planning to support only one previous version at this time? Or has that changed?

ronawho:
> I believe we will be able to update the list.extend/list.append stuff to replace extend with append instead of needing the work around of looping and appending individually.

Yeah, that's probably true, although under the covers this is what `list.append(list)` will do anyways so there's no real perf hit, and it's not too much more code to maintain so if you drop 1.26 would could revert, but I wouldn't personally push to drop 1.26 just because of this.

pierce314159 (approved):
Looks good to me! 

It touches a bit of @hokiegeek2's stuff that i don't know too much about (namely using `shared` vs `:ArkoudaServerDaemon`). So once he looks over it, I think it's good to merge

PR #1798, opened 2022-09-21
Closes #1689 - Adds support for DataFrame to generic attach

### LINKED ISSUE #1689: Add support for DataFrame to generic attach
With the addition of Register and Attach functionality to DataFrame we should explore if it's possible to integrate DataFrame into the Generic Attach workflow

joshmarshall1:
This PR closes #1689

On the chapel side, the symbol table proc `findAll` has been modified to return the matching portion of the regex instead of the entire string that was searched. This allows for more flexibility in searches. The DataFrame attach proc uses the existing attach processes for Strings, pdarray, Categorical, and SegArray to build it's return message.

On the python side the function from_return_msg has been added to DataFrame as well as a helper function that is used to parse the column name out of the registry entry name.


arkouda/dataframe.py
```diff
@@ -2102,6 +2102,101 @@ def unregister_dataframe_by_name(user_defined_name: str) -> None:
...
+        DataFrame
+            A DataFrame representing a set of DataFrame components on the server
+
+        Raises
+        ------
+        RuntimeError
+            Raised if a server-side error is thrown in the process of creating
+            the DataFrame instance
+        """
+        parts = repMsg.split("+")
```
Ethan-DeBandi99, line 2156: Is there any way we could make the return from the a JSON format here? Everything you have looks to check out, but it seems like reading things out of a keyed dictionary would make it much more simple and easier to read.

src/RegistrationMsg.chpl
```diff
@@ -278,6 +280,116 @@ module RegistrationMsg
...
+        const name = msgArgs.getValueOf("name"); 
+        var colName = "df_columns_%s".format(name);
+        var repMsg = "dataframe+%s".format(name);
+
+        regLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                            "%s: Collecting DataFrame components for '%s'".format(cmd, name));
+
+        var jsonParam = msgArgs.get("name");
+        jsonParam.setVal(colName);
+        var json: [0..#1] string = [jsonParam.getJSON()];
```
Ethan-DeBandi99, line 312: `getJSON` requires a size be provided. I would suggest updating this to 
```chapel
var keys: List(string) = new List(["name"]);
var json = msgArgs.getJSON(keys);
```

src/RegistrationMsg.chpl
```diff
@@ -278,6 +280,116 @@ module RegistrationMsg
...
+
+        // Add columns as a json list
+        var cols = stringsToJSONMsg(cmd, "%jt".format(json), json.size, st).msg;
+        repMsg += "+json %s".format(cols);
+
+        // Get index 
+        var indParam = new ParameterObj("name", "df_index_%s_key".format(name), ObjectType.VALUE, "");
+        var indJSON: [0..#1] string = [indParam.getJSON()];
```
Ethan-DeBandi99, line 320: I think you need `asMap` here, not `getJSON`.
```chapel
var indJSON: [0..#1] map(string, string) = [indParam.asMap()];
```

Ethan-DeBandi99 (changes requested):
Nothing major, just a few suggested updates to interacting with message arguments. Also, a suggestion to move the return to the client to JSON formatting in order to make things more readable, but if this is not doable, not a big deal.

src/RegistrationMsg.chpl
Ethan-DeBandi99: @joshmarshall1  - this is my fault. I forgot that `getJSON` is overloaded to support this. In this instance, you are 100% correct because you are creating a new ParameterObj. Resolving this comment.

src/RegistrationMsg.chpl
Ethan-DeBandi99: you would still need to update the value with my previous suggestion. I would actually suggest updating to something similar to the code on 318-320.
```chapel
var colParam = new ParameterObj("name", colName, ObjectType.VALUE, "str")
var json: [0..#1] string = [colParam.getJSON()];
```

arkouda/dataframe.py
joshmarshall1: Possible - Yes. 

However that would require changing how all of the types handle the return and not just DataFrame (Since DataFrame calls all of the other types using the repMsg parts) so I think this should be a separate issue

Ethan-DeBandi99: 10-4, let's just put in a separate issue for that then and work it when time permits.

src/RegistrationMsg.chpl
joshmarshall1: Updated

Ethan-DeBandi99 (approved):

arkouda/dataframe.py
```diff
@@ -2102,6 +2102,101 @@ def unregister_dataframe_by_name(user_defined_name: str) -> None:
...
+        Tuple (columnName, columnType)
+        """
+        regName = entryName.split(" ")[1]
+        colParts = regName.split("_")
+        colType = colParts[2]
+
+        # Case of '_' in the column or dataframe name
+        if len(colParts) > 5:
+            nameInd = regName.rindex(dfName) - 1
+            startInd = len(colType) + 9
```
pierce314159, line 2130: what purpose does the `+ 9` serve?

pierce314159 (approved):
I don't see any issues with this, nice job!

arkouda/dataframe.py
joshmarshall1: It pads the start index so you're only looking at the column name instead of having to do any additional parsing of the string. An example formatting and break down of this for this case would be `df_data_column_name_dfName`. This takes a slice after the `df_data_` (which is always there, and can't vary in length) until the `_dfName` which results in `column_name` regardless of special characters in either the column or dataframe name

PR #1807, opened 2022-09-27
Recommend Chapel 1.28.0 and use it for CI testing

ronawho:
As far as impacts to Arkouda, Chapel 1.28 includes some minor scan optimizations and improvements to instantiation limits.

More details about the 1.28 release can be found at: https://chapel.discourse.group/t/announcing-chapel-1-28-0/16129

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
   mypy:
     runs-on: ubuntu-latest
     container:
-      image: chapel/chapel:1.27.0
+      image: chapel/chapel:1.28.0
     steps:
     - uses: actions/checkout@v2
     - name: Install dependencies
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        chpl-version: ['1.26.0', '1.27.0']
+        chpl-version: ['1.26.0', '1.27.0', '1.28.0']
```
Ethan-DeBandi99, line 97: I am not sure we need 1.26 still since we are only supporting one previous release of chapel. With that said, I don't think having it here is an issue at all.

Ethan-DeBandi99 (approved):
Looks good

.github/workflows/CI.yml
ronawho: I'm personally not sure what the policy is for which versions to support. If somebody wants to create an issue to remove 1.26 support, I'll be happy to remove it in a followup PR.

pierce314159 (approved):

PR #1811, opened 2022-09-28
Closes #1810 add `get_prefixes` and `get_suffixes` to `Strings`

### LINKED ISSUE #1810: Add `get_suffixes` and `get_prefixes` to `Strings`
Similar to `SegArray`, `Strings` should have the ability to get `n`-long prefixes or suffixes. Unlike `SegArray`, the return should be a `Strings`, not a table of character arrays.

reuster986:
Closes #1810 . This PR extends the `SegArray.get_prefixes()` and `.get_suffixes()` semantics to `Strings`. It includes a new unit test.

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1475,7 +1475,7 @@ class Strings:
 
     def get_prefixes(
         self, n: int_scalars, return_origins: bool = True, proper: bool = True
-    ) -> Union[Strings, Tuple[pdarray, Strings]]:
+    ) -> Union[Strings, Tuple[Strings, pdarray]]:
         """
         Return the n-long prefix of each string, where possible
 
@@ -1500,7 +1500,7 @@ class Strings:
             Boolean array that is True where the string was long enough to return
             an n-character prefix, False otherwise.
         """
-        repMsg = generic_msg(
+        repMsg = cast(str, generic_msg(
             cmd="segmentedSubstring",
             args={
                 "objType": self.objtype,
@@ -1510,18 +1510,18 @@ class Strings:
                 "kind": "prefixes",
                 "proper": str(proper).lower(),
             },
-        )
+        ))
         if return_origins:
             parts = repMsg.split("+")
             prefixes = Strings.from_return_msg("+".join(parts[:2]))
             longenough = create_pdarray(parts[2])
-            return prefixes, longenough
+            return prefixes, cast(pdarray, longenough)
         else:
             return Strings.from_return_msg(repMsg)
 
     def get_suffixes(
         self, n: int_scalars, return_origins: bool = True, proper: bool = True
-    ) -> Union[Strings, Tuple[pdarray, Strings]]:
+    ) -> Union[Strings, Tuple[Strings, pdarray]]:
         """
         Return the n-long suffix of each string, where possible
 
@@ -1546,7 +1546,7 @@ class Strings:
             Boolean array that is True where the string was long enough to return
             an n-character suffix, False otherwise.
         """
-        repMsg = generic_msg(
+        repMsg = cast(str, generic_msg(
             cmd="segmentedSubstring",
             args={
                 "objType": self.objtype,
@@ -1556,12 +1556,12 @@ class Strings:
                 "kind": "suffixes",
                 "proper": str(proper).lower(),
             },
-        )
+        ))
         if return_origins:
             parts = repMsg.split("+")
             suffixes = Strings.from_return_msg("+".join(parts[:2]))
             longenough = create_pdarray(parts[2])
-            return suffixes, longenough
+            return suffixes, cast(pdarray, longenough)
         else:
             return Strings.from_return_msg(repMsg)
 
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1504,11 +1504,11 @@ class Strings:
             cmd="segmentedSubstring",
             args={
                 "objType": self.objtype,
-                "name": self.name,
+                "name": self,
                 "nChars": n,
                 "returnOrigins": return_origins,
                 "kind": "prefixes",
-                "proper": str(proper).lower(),
+                "proper": proper,
             },
         ))
         if return_origins:
@@ -1550,11 +1550,11 @@ class Strings:
             cmd="segmentedSubstring",
             args={
                 "objType": self.objtype,
-                "name": self.name,
+                "name": self,
                 "nChars": n,
                 "returnOrigins": return_origins,
                 "kind": "suffixes",
-                "proper": str(proper).lower(),
+                "proper": proper,
             },
         ))
         if return_origins:
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1473,6 +1473,84 @@ class Strings:
     def __radd__(self, other: Strings) -> Strings:
         return self.lstick(other)
 
+    def get_prefixes(self, n: integral, return_origins:bool=True, proper:bool=True) -> Union[Strings,Tuple[pdarray,Strings]]:
+        """
+        Return the n-long prefix of each string, where possible
+
+        Parameters
+        ----------
+        n : int
+            Length of prefix
+        return_origins : bool
+            If True, return a logical index indicating which strings
+            were long enough to return an n-prefix
+        proper : bool
+            If True, only return proper prefixes, i.e. from strings
+            that are at least n+1 long. If False, allow the entire
+            string to be returned as a prefix.
+
+        Returns
+        -------
+        prefixes : Strings
+            The array of n-character prefixes; the number of elements is the number of 
+            True values in the returned mask.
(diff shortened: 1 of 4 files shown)


src/SegmentedMsg.chpl
```diff
@@ -1368,6 +1368,48 @@ module SegmentedMsg {
...
+    st.checkTable(name);
+
+    select (objtype) {
+      when ("str") {
+        var strings = getSegString(name, st);
+        var returnOrigins = msgArgs.getValueOf("returnOrigins").toLower(): bool;
+        if returnOrigins {
+          var (off, byt, longEnough) = strings.getFixes(msgArgs.getValueOf("nChars"): int,
+                                                        msgArgs.getValueOf("kind"): Fixes,
+                                                        msgArgs.getValueOf("proper").toLower(): bool,
```
pierce314159, line 1390: similarly we should able to do 
```chapel
msgArgs.get("nChars").getIntValue(),
...
msgArgs.get("returnOrigins").getBoolValue(),
```

arkouda/strings.py
```diff
@@ -1473,6 +1473,98 @@ def lstick(self, other: Strings, delimiter: Union[bytes, str_scalars] = "") -> S
...
+        """
+        repMsg = cast(str, generic_msg(
+            cmd="segmentedSubstring",
+            args={
+                "objType": self.objtype,
+                "name": self.name,
+                "nChars": n,
+                "returnOrigins": return_origins,
+                "kind": "prefixes",
+                "proper": str(proper).lower(),
```
pierce314159, line 1511: I think you should be able to just pass `self` and `proper` here. Ethans message processing code should take care of the rest. So
```python
 args={
      "objType": self.objtype,
      "name": self,
      "nChars": n,
      "returnOrigins": return_origins,
      "kind": "prefixes",
      "proper": proper,
}
```

arkouda/strings.py
```diff
@@ -1473,6 +1473,98 @@ def lstick(self, other: Strings, delimiter: Union[bytes, str_scalars] = "") -> S
...
+        repMsg = cast(str, generic_msg(
+            cmd="segmentedSubstring",
+            args={
+                "objType": self.objtype,
+                "name": self.name,
+                "nChars": n,
+                "returnOrigins": return_origins,
+                "kind": "suffixes",
+                "proper": str(proper).lower(),
+            },
```
pierce314159, line 1558: same as above, we can just pass `self` and `proper`

src/SegmentedMsg.chpl
```diff
@@ -1368,6 +1368,48 @@ module SegmentedMsg {
...
+          st.addEntry(leName, new shared SymEntry(longEnough));
+          repMsg = "created " + st.attrib(retString.name) + "+created bytes.size %t".format(retString.nBytes) + "+created " + st.attrib(leName);
+        } else {
+          var (off, byt) = strings.getFixes(msgArgs.getValueOf("nChars"): int,
+                                            msgArgs.getValueOf("kind"): Fixes,
+                                            msgArgs.getValueOf("proper").toLower(): bool,
+                                            false);
+          var retString = getSegString(off, byt, st);
+          repMsg = "created " + st.attrib(retString.name) + "+created bytes.size %t".format(retString.nBytes);
+        }
```
pierce314159, line 1403: I think we simplify things a bit if we wait til the end to do the `if/else` on `returnOrigins`

```Chapel
var (off, byt, longEnough) = strings.getFixes(msgArgs.get("nChars").getIntValue(),
                                              msgArgs.getValueOf("kind"): Fixes,
                                              msgArgs.get("proper").getBoolValue());
var retString = getSegString(off, byt, st);
repMsg = "created " + st.attrib(retString.name) + "+created bytes.size %t".format(retString.nBytes);
if returnOrigins {
  var leName = st.nextName();
  st.addEntry(leName, new shared SymEntry(longEnough));
  repMsg += "+created " + st.attrib(leName);
}
```

This  also allows us to eliminate the 2 versions of `getFixes` based on `returnOrigins` since `getFixesHelper` always returns what we want now

src/SegmentedString.chpl
```diff
@@ -1112,6 +1117,59 @@ module SegmentedString {
...
+    }
+
+    proc getFixes(n: int, kind: Fixes, proper: bool, param returnOrigins) where returnOrigins {
+      return getFixesHelper(n, kind, proper);
+    }
+
+    proc getFixes(n: int, kind: Fixes, proper: bool, param returnOrigins) where !returnOrigins {
+      var (off, byt, le) = getFixesHelper(n, kind, proper);
+      return (off, byt);
+    }
```
pierce314159, line 1171: if we go with the delayed action on `returnOrigins`, we should need to do this anymore since we can just call `getFixesHelper` directly

src/SegmentedString.chpl
```diff
@@ -1112,6 +1117,59 @@ module SegmentedString {
...
+      } else {
+        longEnough = (lengths >= n);
+      }
+      var nFound = + reduce longEnough;
+      var retOffsets = makeDistArray(nFound, int);
+      forall (i, o) in zip(retOffsets.domain, retOffsets) {
+        o = i * (n + 1);
+      }
+      var retBytes = makeDistArray(nFound * (n + 1), uint(8));
+      var srcInds = makeDistArray(nFound * (n + 1), int);
```
pierce314159, line 1134: Are we guaranteed that `retBytes` and `srcInds` will have the same domain if we do it this way? And does that even matter? If that is something to worry about, maybe we do
```chapel
const retDom = makeDistDom(nFound * (n + 1));
var retBytes = [retDom] uint(8);
var srcInds = [retDom] int;
```

src/SegmentedString.chpl
```diff
@@ -1112,6 +1117,59 @@ module SegmentedString {
...
+            for j in 0..#(n+1) {
+              agg.copy(srcInds[srcIndStart + j], byteStart + j);
+            }
+          }
+        }
+      }
+      ref va = values.a;
+      forall (b, i) in zip(retBytes, srcInds) with (var agg = newSrcAggregator(uint(8))) {
+        agg.copy(b, va[i]);
+      }
```
pierce314159, line 1160: I think this is the write way to do it with gathering the indices first

At first, I felt like we should be able to write directly into `retBytes`, but I convinced myself that this would cause remote-to-remote issues because it essentially boils down to
```chapel
agg.copy(retBytes[srcIndStart + j], va[byteStart + j]);
```

pierce314159 (approved):
The logic all looks good to me!

Only real suggestion I have is I think delaying action on `returnOrigins` simplifies things a bit, but everything seems to work to me!

src/SegmentedMsg.chpl
```diff
@@ -1368,6 +1368,48 @@ module SegmentedMsg {
...
+    var objtype = msgArgs.getValueOf("objType");
+    var name = msgArgs.getValueOf("name");
+    
+    // check to make sure symbols defined
+    st.checkTable(name);
+
+    select (objtype) {
+      when ("str") {
+        var strings = getSegString(name, st);
+        var returnOrigins = msgArgs.getValueOf("returnOrigins").toLower(): bool;
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Just to add more context here `getBoolValue` will do exactly what you have under the hood.

Ethan-DeBandi99 (approved):
Other than what Pierce has already brought up, I do not see anything that needs to be addressed.

src/SegmentedMsg.chpl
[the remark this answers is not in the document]
reuster986: Thanks, this is cool!

src/SegmentedMsg.chpl
reuster986: good to know!

src/SegmentedString.chpl
reuster986: done!

src/SegmentedString.chpl
reuster986: Your way is safer and more maintainable, so I'll go with it, although I don't think the way it's written carries much risk. Unless something fundamental changes in `makeDistArray`, those two arrays are guaranteed to have equivalent domains (although not the same domain object in memory, I don't think). Having equivalent domains is important, because I don't think you can `zip` different domains together in a `forall`.

src/SegmentedString.chpl
reuster986: Yeah, it feels like two `forall`s shouldn't be necessary, but I don't see a way to do it in one `forall` without remote-to-remote. This is similar to how we had to do it for the string gather implementation, `SegString.this(iv: [] int)`.

arkouda/strings.py
reuster986: Thanks, this is cool!

arkouda/strings.py
reuster986: fixed

PR #1813, opened 2022-09-29
Raise instantiation limit for 1.26/1.27

ronawho:
Chapel will error if a generic function is instantiated too many times in order to avoid infinite loops with recursive instantiations. Recently, we've hit the default limit of 256 instantiations for some internal functions used for I/O, but this isn't actually an issue so just raise the default to 512 instantiations. Note that Chapel 1.28 improved the situation by ignoring the specific IO function and raising the default to 512 so we only need this for older versions.


Ethan-DeBandi99 (approved):
Looks good

pierce314159 (approved):

PR #1815, opened 2022-09-30
Closes #1814 Change interval arguments to closed

### LINKED ISSUE #1814: Support closed intervals in interval functions
Currently, in `interval_lookup()` and `search_intervals()`, the interval arguments are assumed to be half-open. This was done to conform to Python's slicing convention, but it actually creates problems for some use cases of these functions. For example, if the (closed) upper bound of an interval is `2**63 - 1`, then there is no way to express this as an open upper bound in a `uint64`. Also, the current logic converts open upper bounds to closed by decrementing, but this is incorrect for `float64` bounds. If all intervals were closed, `float64` would work nicely.

reuster986:
Closes #1814 . This is a backward-breaking change, but it was requested by users, and it resolves some tricky issues, so I think it's ok. We can add a warning if people think it's necessary.

### DIFF
--- a/arkouda/alignment.py
+++ b/arkouda/alignment.py
@@ -280,7 +280,7 @@ def in1d_intervals(vals, intervals, symmetric=False):
 
 def search_intervals(vals, intervals, tiebreak=None):
     """
-    Given an array of query vals and half-open (pythonic) intervals, return the index of the
+    Given an array of query vals and closed intervals, return the index of the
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good. I don't see a need to add a warning since the docstrings have been updated. However, this would probably be something that we want to highlight explicitly in the release notes for awareness when we cut the next release.

pierce314159 (approved):
Looks good

PR #1821, opened 2022-09-30
Closes #1795: Chapel Intro Docs

### LINKED ISSUE #1795: Intro to Chapel docs
Part of #1793

Chapel examples and guided tutorial on writing code in chapel to help familiarize new developers with concepts used regularly within arkouda

pierce314159:
This PR (Closes #1795):
- Adds `CHAPEL_TUTORIAL.md` to the `training` folder introducing a lot of common chapel functionality used by arkouda devs
- Adds `Tutorial.chpl` containing most code examples and space to try concepts out for themselves

I recommend viewing the rendered version:
https://github.com/pierce314159/arkouda/blob/1795_chapel_intro_docs/training/CHAPEL_TUTORIAL.md

I'm curious what some of the chapel folks think of this. Especially if there's anything I've missed or misexplained (because this also a good opportunity to fill any gaps in my knowledge) @bmcdonald3 @ronawho @bradcray 

I'm also very interested on what @cdhollis94 thinks! Seeing as they're the newest to the team

### DIFF
--- a/training/CHAPEL_TUTORIAL.md
+++ b/training/CHAPEL_TUTORIAL.md
@@ -17,7 +17,7 @@ This guide will familiarize new developers with Chapel concepts regularly used i
 5. [`forall` Loops](#forall)
    1. [Example 6: Factorial with must-parallel `forall` and Reduction](#ex6)
    2. [Example 7: may-parallel `forall`](#ex7)
-   3. [Example 8: may-parallel `forall` Expression](#ex8)
+   3. [Example 8: `forall` Expressions](#ex8)
    4. [Example 9: Factorial with may-parallel `forall` Expression and Reduction](#ex9)
    5. [Try It Yourself 1: Perfect Squares <=25](#TIY1)
 6. [Ternary](#ternary)
@@ -56,12 +56,12 @@ The command to run the executable is:
 ```console
 ./tutorial
 ```
-To follow along, comment out all but the relevant example in `Tutorial.chpl`, compile, run, and verify the output matches the guide!
+To follow along, uncomment the relevant example in `Tutorial.chpl`, compile, run, and verify the output matches the guide!
 
 <a id="ranges_and_domains"></a>
 ## Ranges and Domains
 This will only cover enough to provide context for the other examples (the very basics). More information can be found in the Chapel docs for
-[ranges](https://chapel-lang.org/docs/primers/ranges.html) and [domains](https://chapel-lang.org/docs/language/spec/domains.html).
+[ranges](https://chapel-lang.org/docs/primers/ranges.html) and [domains](https://chapel-lang.org/docs/primers/domains.html).
 
 <a id="ranges"></a>
 ### Ranges
@@ -115,7 +115,7 @@ you are guaranteed that index `i` of one array will be local to index `i` of the
 
 In Arkouda, pdarrays (which stand for parallel, distributed arrays), are Chapel arrays stored in
 [block-distributed domains](https://chapel-lang.org/docs/primers/distributions.html#block-and-distribution-basics),
-meaning that the elements are split evenly across all locales.
+meaning that the elements are split as evenly as possible across all locales.
 
 The syntax for declaring an array with domain `D` and type `t` is:
 ```Chapel 
@@ -142,7 +142,7 @@ Note: `makeDistDom` is a helper function in Arkouda and not part of Chapel.
 
 <a id="init_fact"></a>
 ## Initial Factorial
-Let's start our chapel journey by writing a function to calculate the factorial of a given integer `n`.
+Let's start our Chapel journey by writing a function to calculate the factorial of a given integer `n`.
 Where factorial is
 
 $$ n! = \prod_{i=1}^n i = 1 \cdot 2 \cdot\ldots\cdot (n-1) \cdot n$$
@@ -194,18 +194,19 @@ Tutorial.chpl:43: note: The shadow variable 'fact' is constant due to forall int
 That's not what we want to see! There was an error during compilation.
 The issue is different iterations of the parallel loop could be modifying the outer variable `fact` simultaneously.
 
-Chapel handles this by giving each task its own copy of `fact` known as a [shadow variable](https://chapel-lang.org/docs/primers/forallLoops.html?highlight=intents#task-intents-and-shadow-variables).
+Chapel handles this by giving each task its own copy of `fact` known as a [shadow variable](https://chapel-lang.org/docs/primers/forallLoops.html#task-intents-and-shadow-variables).
 To combine the shadow variables and write the result back into `fact`, we'll need to learn about `reduction` operations.
 
 <a id="reduce_and_scan"></a>
 ## `reduce` and `scan`
-[Reductions and scans](https://chapel-lang.org/docs/language/spec/data-parallelism.html?highlight=scan#reductions-and-scans)
+[Reductions and scans](https://chapel-lang.org/docs/language/spec/data-parallelism.html#reductions-and-scans)
 cumulatively apply an operation over the elements of an array (or any iterable) in parallel.
 
 - `op scan array`
   - scans over `array` and cumulatively applies `op` to every element
   - returns an array
   - `+ scan a` behaves like [`np.cumsum(a)`](https://numpy.org/doc/stable/reference/generated/numpy.cumsum.html) in numpy
+  - `scan` is an [inclusive scan](https://en.wikipedia.org/wiki/Prefix_sum#Inclusive_and_exclusive_scans)
 - `op reduce array`
   - reduces the result of a scan to a single summary value
   - returns a single value
@@ -240,22 +241,22 @@ min reduce a = 1
 max scan a = 1 2 3 4 5
 max reduce a = 5
 ```
-Notice `op reduce array` is often the last element of `op scan array`. 
+Notice `op reduce array` is always the last element of `op scan array`.
 
-Reductions and scans are defined for [many operations](https://chapel-lang.org/docs/language/spec/data-parallelism.html?highlight=scan#reduction-expressions)
-, and you can even [write your own](https://chapel-lang.org/docs/technotes/reduceIntents.html#user-defined-reduction-example)!
+Reductions and scans are defined for [many operations](https://chapel-lang.org/docs/language/spec/data-parallelism.html#reduction-expressions),
+and you can even [write your own](https://chapel-lang.org/docs/technotes/reduceIntents.html#user-defined-reduction-example)!
 
 <a id="forall"></a>
 ## `forall` Loops
 <a id="ex6"></a>
 #### Example 6: Factorial with must-parallel `forall` and Reduction
 Back to parallel factorial, let's use `reduce` to combine the results of each task.
-This a [task intent](https://chapel-lang.org/docs/primers/forallLoops.html#task-intents-and-shadow-variables) as signified by the `with` keyword.
+This is a [task intent](https://chapel-lang.org/docs/primers/forallLoops.html#task-intents-and-shadow-variables) as signified by the `with` keyword.
 ```Chapel
 proc factorial(n: int) {
   var fact: int = 1;
   forall i in 1..n with (* reduce fact) {
-    fact = i;
+    fact *= i;
   }
   return fact;
 }
@@ -265,7 +266,7 @@ writeln(factorial(5));
 120
 ```
 That's more like it!
-Every iteration assigned its shadow variable of `fact` to `i`, so a `* reduce` of the shadow variables gives the product of all `i`.
+Every task multiplies its shadow variable of `fact` by `i`, so a `* reduce` of the shadow variables gives the product of all `i`.
 
 This is an example of a must-parallel `forall`.
 
@@ -299,14 +300,17 @@ Let's look at an example of a may-parallel `forall`:
 5
 ```
 As we can see this loop is not executing serially.
+This is because core Chapel types like ranges, domains, and arrays support parallel iterators, so they will be invoked by the may-parallel form.
 Your output will likely be in a different order than the above and will differ between runs.
 
 <a id="ex8"></a>
-#### Example 8: may-parallel `forall` Expression
+#### Example 8: `forall` Expressions
 `forall`s can also be used in expressions, for example:
 ```Chapel
-var tens = [i in 1..10] i*10;
+// must-parallel forall expression
+var tens = forall i in 1..10 do i*10;
 writeln(tens);
+// may-parallel forall expression
 var negativeTens = [i in tens] -i;
 writeln(negativeTens);
 ```
@@ -327,7 +331,7 @@ writeln(factorial(5));
 120
 ```
 
-For a specific `n`, we  can do this in one line!
+For a specified `n`, we  can do this in one line!
 ```Chapel
 writeln(* reduce [i in 1..#5] i);
 ```
@@ -335,7 +339,7 @@ writeln(* reduce [i in 1..#5] i);
 <a id="TIY1"></a>
 #### Try It Yourself 1: Perfect Squares <=25
 Problem:
-Generate and print an array of all perfect squares less than or equal to `25`
+Compute and print out all perfect squares less than or equal to `25`
 
 Bonus points if you can do it in one line using a `forall` expression!
 <details>
@@ -399,8 +403,7 @@ Problem:
 Write a `proc` using a ternary to take an `int array`, `A`, 
 and return an array where index `i` is the absolute value of `A[i]`
 
-Input: 
-`[-3, 7, 0, -4, 12]`
+Call: `arrayAbsVal([-3, 7, 0, -4, 12]);`
 
 <details>
   <summary>Potential Solution</summary>
@@ -426,10 +429,10 @@ Using introspection will result in a [generic](https://chapel-lang.org/docs/lang
 
 The syntax for this is:
 ```Chapel
-proc foo(array: [?D] ?t, value: t, array2: [?D2] ?t2) {
-  // using `D` and `t` in this proc will refer to `array`'s domain and type respectively
-  // `value` has the same type as `array` so there's no need for the `?`
-  // `D2` and `t2` refer to the domain and type of `array2`
+proc foo(arr1: [?D] ?t, value: t, arr2: [?D2] ?t2) {
+  // using `D` and `t` in this proc will refer to `arr1`'s domain and type respectively
+  // since value is declared to have type `t`, it must be passed a value that is compatible with `arr1`'s element type
+  // `D2` and `t2` refer to the domain and type of `arr2`
 }
 ```
 <a id="ex12"></a>
@@ -474,16 +477,23 @@ writeln(factorial([1, 2, 3, 4, 5]));
 1 2 6 24 120
 ```
 
+Promotion is equivalent to a `forall` loop. For this example the equivalent loop would be:
+```chapel
+[x in [1, 2, 3, 4, 5]] factorial(x);
+```
+
 <a id="zip_and_agg"></a>
 ## Zippered Iteration and Aggregation
 
 <a id="zippered_iteration"></a>
 ### Zippered Iteration
 [Zippered Iteration](https://chapel-lang.org/docs/users-guide/base/zip.html#zippered-iteration) is simultaneously
-iterating over multiple iterables (usually arrays and ranges) with compatible domains.
+iterating over multiple iterables (usually arrays and ranges) with compatible shape and size.
+For arrays, this means that their domains have the same number of elements in each dimension
 
 The syntax:
 ```chapel
+var A1, A2, A3, ..., An: [D] int;
 forall (v1, v2, v3, ..., vn) in zip(A1, A2, A3, ..., An) {
   // for loop iteration `j`, `vi` will refer to the `j`th element of `Ai`
 }
@@ -504,7 +514,7 @@ writeln(A);
 1.0 1.0 3.0 16.0 625.0
 ```
 Notice we have an [unbounded range](https://chapel-lang.org/docs/primers/ranges.html#variations-on-basic-ranges), `1..`,
-so the end bound is determined by the size of the iterables.
+so the end bound is determined by the size of the other iterables.
 Since in this case the other iterables are length 5, `1..` is equivalent to `1..5`.
 
 <a id="ex15"></a>
@@ -526,7 +536,8 @@ On the server, a SegString is made up of two distributed array components:
 - offsets: `int`
   - the start index of each individual string in the `values`
 
-For the sake of simplicity, we'll treat values as a string array. So for our `s` above, these components look something like
+For the sake of simplicity, we'll treat values as a string array.
+So for our `s` above, these components look something like:
 ```
 values = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00']
 offsets = [0, 5, 9]
@@ -534,9 +545,9 @@ offsets = [0, 5, 9]
 For `getLengths`, we want to calculate the length of each individual string including the null terminator.
 
 ```chapel
-const values: [0..#12] string = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00'];
-const offsets = [0, 5, 9];
-const size = 3;
+const values: [0..#12] string = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00'],
+      offsets = [0, 5, 9],
+      size = offsets.size;  // size refers to the number of stings in the Segstring, this is always equivalent to offsets.size
 
 /* Return lengths of all strings, including null terminator. */
 proc getLengths() {
@@ -556,7 +567,7 @@ proc getLengths() {
     }
     else {
       // not the last string
-      // len = start poistion of next string - start position of this string
+      // len = start position of next string - start position of this string
       l = offsets[i+1] - o;
     }
   }
@@ -590,6 +601,9 @@ forall (i, v) in zip(inds, vals) with (var agg = new DstAggregator(int)) {
   agg.copy(remoteArr[i], v);
 }
 ```
+The `with` keyword here declares `agg` as a [Task-Private variable](https://chapel-lang.org/docs/language/spec/data-parallelism.html#task-private-variables),
+meaning each task will get its own shadow variable.
+
 It's important to note aggregation will only work if at least one side is local.
 Both sides being remote (remote-to-remote aggregation) is not yet supported. 
 
@@ -605,7 +619,7 @@ use CopyAggregation;
   Given a SegString, return a new SegString with all lowercase characters from the original replaced with their uppercase equivalent
   :returns: Strings – Substrings with lowercase characters replaced with uppercase equivalent
 */
-proc upper() throws {
+proc upper() {
   var upperVals: [values.domain] string;
   const lengths = getLengths();
   forall (off, len) in zip(offsets, lengths) with (var valAgg = new DstAggregator(string)) {
@@ -626,9 +640,9 @@ writeln("New Vals:", newVals);
 Old Vals:s s s 0  s s 1  s 2
 New Vals:S S S 0  S S 1  S 2
 ```
-This function uses our previous `getLengths` function as well as the chapel builtins
-[`toUpper`](https://chapel-lang.org/docs/language/spec/strings.html?highlight=totitle#String.string.toUpper)
-and [`join`](https://chapel-lang.org/docs/language/spec/strings.html?highlight=totitle#String.string.join)!
+This function uses our previous `getLengths` function as well as the Chapel builtins
+[`toUpper`](https://chapel-lang.org/docs/language/spec/strings.html#String.string.toUpper)
+and [`join`](https://chapel-lang.org/docs/language/spec/strings.html#String.string.join)!
 
 
 <a id="TIY3"></a>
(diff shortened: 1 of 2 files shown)


bradcray:
@pierce314159 : Just acknowledging that I've seen this.  I'm currently hip-deep in editing release notes for our 1.28 release, but will hope to use this as a study break next week.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+
+* scan a = 1 2 6 24 120
+* reduce a = 120
+
+min scan a = 1 1 1 1 1
+min reduce a = 1
+
+max scan a = 1 2 3 4 5
+max reduce a = 5
+```
```
Ethan-DeBandi99, line 234: Personally, I find it easier to read/reference when the output is provided under the command. However, this is entirely preference and since we have the `tutorial.chpl` for a reader to work along with this guide, I would support leaving as is. Just something to think about.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+  }
+  return fact;
+}
+writeln(factorial(5));
+```
+```console
+120
+```
+That's more like it!
+Every iteration assigned its shadow variable of `fact` to `i`, so a `* reduce` of the shadow variables gives the product of all `i`.
```
Ethan-DeBandi99, line 259: I would explicitly call out that this is done by the `with` statement.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+There are [2 types of `forall`](https://chapel-lang.org/docs/primers/forallLoops.html#forall-loops) loops:
+- must-parallel `forall`
+  - Is written using the `forall` keyword i.e. `forall i in D`
+  - Requires a parallel iterator
+- may-parallel `forall`
+  - Is written using bracket notation i.e. `[i in D]`
+  - Will use a parallel iterator if present, otherwise will iterate serially
+
+<a id="ex7"></a>
+#### Example 7: may-parallel `forall`
```
Ethan-DeBandi99, line 272: Would it be worth also demonstrating when a parallel iterator does not exist? I am not entirely sure it will add much value

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+
+```console
+3 7 0 4 12
+```
+
+<a id="introspection"></a>
+## Introspection
+Introspection is determining properties of a function argument at runtime.
+This is often used to determine the type and/or domain of a function argument. 
+Using introspection can avoid duplicating a `proc` for multiple types when none of the logic has changed.
```
Ethan-DeBandi99, line 415: I would add a note about this making the function `generic` since we have had to debug multiple things due to generic function recently. I think a simple mention with a link to the Chapel documentation would be good.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+var i: [0..#5] int = [-3, 7, 0, -4, 12];
+writeln(absoluteVal(i));
+```
+```console
+3.14 7.0 0.0 inf inf
+3 7 0 4 12
+```
+
+<a id="promotion"></a>
+## Promotion
```
Ethan-DeBandi99, line 446: Glad you highlighted this! I constantly have to remind myself about it!

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+}
+```
+Since all `Ai` have compatible domains, we know all `vi` will be local to one another on any given loop iteration.
+
+<a id="ex14"></a>
+#### Example 14: Simple Zippered Iteration
+Let's start with a simple example involving zippered iteration:
+```Chapel
+var A: [1..5] real;
+for (a, i, j) in zip(A, 1.., [3, 0, 1, 2, 4]) {
```
Ethan-DeBandi99, line 488: We have not yet mention `n..` it would be good to explicitly explain this functionality

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+```console
+5 4 3
+```
+Nice! We'll use `getLengths`, `values`, and `offsets` in the next section as well.
+
+<a id="aggregation"></a>
+### Aggregation
+
+[Aggregation](https://chapel-lang.org/docs/modules/packages/CopyAggregation.html#copyaggregation) is used to copy a
+local value into a remote array or copy a value from a remote array into a local variable. This provides a significant speed up when doing batch assignments.
```
Ethan-DeBandi99, line 564: Since we are only demoing the DstAggegrator, I think with would be useful to  to update `is used to copy a
local value into a remote array or copy a value from a remote array into a local variable` to `is used to copy a
local value into a remote array (DstAggregator) or copy a value from a remote array into a local variable (SrcAggegator)

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+  return (offsets, upperVals);
+}
+
+writeln("Old Vals:", values);
+var (newOffs, newVals) = upper();
+writeln("New Vals:", newVals);
+```
+```console
+Old Vals: s s s 0  s s 1  s 2
+New Vals: S S S 0  S S 1  S 2
```
Ethan-DeBandi99, line 613: I don't think the original strings had spaces.

Ethan-DeBandi99 (changes requested):
This looks great overall. You did a great job highlighting the key concepts we use most often and the examples for each are great. A few minor updates (some of which are optional/up for discussion) and this will be ready to go.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+
+On the server, a SegString is made up of two distributed array components:
+- values: `uint(8)`
+  - the flattened array of bytes in all the strings (null byte delimited)
+- offsets: `int`
+  - the start index of each individual string in the `values`
+
+For the sake of simplicity, we'll treat values as a string array. So for our `s` above, these components look something like
+```
+values = ['\x00', 's', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2']
```
bmcdonald3, line 518: I guess I thought that the null terminator came at the end of the strings (if it doesn't then I think the Parquet code needs some reworking!). Am I just misremembering or should this instead be `values = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00']`?

training/CHAPEL_TUTORIAL.md
bmcdonald3: Another thing here is maybe we should say values plural, rather than singular, since aggregation is typically used for portions of arrays and not just single elements? (I feel like this is implied by the note about batch assignments)

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+iv = 0 0 0 1 2 2 3 3 4 4
+iv[truth] = 0 1 2 3
+
++ reduce truth = 4
+0..#(+ reduce truth) = 0..3
+
+Y = 5 5 5 5
+```
+Awesome! We used most of the information from this guide to reproduce useful numpy functionality!
+Now let's tackle the very similar challenge of expansion indexing.
```
bmcdonald3, line 748: Maybe call out explicitly that the reproduced numpy functionality is in parallel and could be distributed? I feel like this gives some context to why we would use Chapel instead of numpy, but probably just nit picking.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+```
+- `j..#n` is equivalent to `j..j+(n-1)`.
+
+<a id="domains"></a>
+### Domains
+
+A domain is an index set used to specify iteration spaces and define the size and shape of arrays.
+A domain’s indices may be distributed across multiple locales.
+When iterating through two arrays with the same domain,
+you are guaranteed that index `i` of one array will be local to index `i` of the other.
```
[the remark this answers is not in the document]
pierce314159: Great idea! I'll update to ben's suggestion!

cdhollis94:
I don't know enough about Chapel to have many opinions about the content or if anything is missing, but I will say that the format is exactly what I would want from documentation like this. Having lots of examples with the expected output given is a huge plus, and whenever I ran into a concept I wasn't familiar with (like shadow variables) it always had a link to additional reading. I'll definitely be using this as I learn and review the content 👍

training/CHAPEL_TUTORIAL.md
pierce314159: Yeah I don't think it's worth it. The [example ](https://chapel-lang.org/docs/primers/forallLoops.html#may-parallel-forall-statement) in the chapel docs (copied below) on this seems unlikely to be encountered by a new arkouda dev

```Chapel
iter onlySerial(m: int) {
  for j in 1..m do
    yield j;
}

[i in onlySerial(n)] {
  writeln("in onlySerial iteration #", i);
}
```
```Chapel
forall i in onlySerial(n) { // error: a parallel iterator is not found
  writeln("in iteration #", i);
}
```

training/CHAPEL_TUTORIAL.md
pierce314159: oh yeah good point!

training/CHAPEL_TUTORIAL.md
pierce314159: great suggestions! I'll be sure to update this!

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+
+<a id="ex8"></a>
+#### Example 8: may-parallel `forall` Expression
+`forall`s can also be used in expressions, for example:
+```Chapel
+var tens = [i in 1..10] i*10;
+writeln(tens);
+var negativeTens = [i in tens] -i;
+writeln(negativeTens);
+```
```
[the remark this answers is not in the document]
pierce314159: It certainly does pop up! I only looked in `SegmentedString` and there are 4 cases with 3 of them being variable assignments to a may-parallel `forall` expression. I feel like it's useful to have for awareness so they're not completely lost (or at least have a reference) in case they come across it in existing code/code reviews

training/CHAPEL_TUTORIAL.md
pierce314159: oooo I always have trouble remembering. I'll run an example real quick and see

training/CHAPEL_TUTORIAL.md
pierce314159: that's a great idea!

training/CHAPEL_TUTORIAL.md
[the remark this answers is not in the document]
pierce314159: Updated!

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,846 @@
...
+writeln(factorial(5));
+```
+```console
+$ chpl Tutorial.chpl -o tutorial
+$ ./tutorial
+120
+```
+
+Excellent! Now let's write this loop in parallel.
+Our parallelism comes from the data parallel [`forall`](https://chapel-lang.org/docs/users-guide/datapar/forall.html) loop.
```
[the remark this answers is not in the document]
pierce314159: Added!

training/CHAPEL_TUTORIAL.md
pierce314159: Added

training/CHAPEL_TUTORIAL.md
pierce314159: Added

training/CHAPEL_TUTORIAL.md
pierce314159: You're completely right! I'll get this updated

training/CHAPEL_TUTORIAL.md
bradcray: Related to my previous comment, "Every iteration assigned" should change to "Every task updated" here.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+0
+9
+2
+6
+3
+1
+7
+5
+```
+As we can see this loop is not executing serially.
```
bradcray, line 301: Could add: "This is because core Chapel types like ranges, domains, and arrays support parallel iterators, so they will be invoked by the may-parallel form."

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+1 2 6 24 120
+```
+
+<a id="zip_and_agg"></a>
+## Zippered Iteration and Aggregation
+
+<a id="zippered_iteration"></a>
+### Zippered Iteration
+[Zippered Iteration](https://chapel-lang.org/docs/users-guide/base/zip.html#zippered-iteration) is simultaneously
+iterating over multiple iterables (usually arrays and ranges) with compatible domains.
```
bradcray, line 483: More generally, "with compatible shape and size" (since ranges don't have domains).  You could then add something like "for arrays, this means that their domains have the same number of elements in each dimension."

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+
+The syntax:
+```chapel
+forall (v1, v2, v3, ..., vn) in zip(A1, A2, A3, ..., An) {
+  // for loop iteration `j`, `vi` will refer to the `j`th element of `Ai`
+}
+```
+Since all `Ai` have compatible domains, we know all `vi` will be local to one another on any given loop iteration.
```
bradcray, line 491: Wait, do we know that they do?  We haven't seen the declarations of these arrays, have we?  This could be remedied by adding:

```chapel
var A1, A2, A3, ..., An: [something-something] int;
```

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+```
+values = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00']
+offsets = [0, 5, 9]
+```
+For `getLengths`, we want to calculate the length of each individual string including the null terminator.
+
+```chapel
+const values: [0..#12] string = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00'];
+const offsets = [0, 5, 9];
+const size = 3;
```
bradcray, line 539: What is the role of `size` here?  Is it the number of offsets?  (in which case, writing `const size = offsets.size;` feels more bulletproof and descriptive).

Totally optional note, but note that each `const` doesn't require its own keyword if you'd rather write this as:

```chapel
const values: [0..#12] string = [...],
      offsets = [0, 5, 9],
      size = 3;
```

(or any combination thereof).  Mostly just good to know that every declaration need not get its own keyword for those who get tired of typing it.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+  const high = offsets.domain.high;
+  forall (i, o, l) in zip(offsets.domain, offsets, lengths) {
+    if i == high {
+      // the last string
+      // len = values.size - start position of this string
+      l = values.size - o;
+    }
+    else {
+      // not the last string
+      // len = start poistion of next string - start position of this string
```
bradcray, line 559: Typo: "poistion"

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+```
+For `getLengths`, we want to calculate the length of each individual string including the null terminator.
+
+```chapel
+const values: [0..#12] string = ['s', 's', 's', '0', '\x00', 's', 's', '1', '\x00', 's', '2', '\x00'];
+const offsets = [0, 5, 9];
+const size = 3;
+
+/* Return lengths of all strings, including null terminator. */
+proc getLengths() {
```
bradcray, line 542: Should this take the arrays like `offsets`, `values` in as arguments rather than referring to (seeming) globals?  Doing so would let you replace `offsets.domain` with a queried domain as introduced in the previous examples if desired.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+  // initialize lengths with the same domain as offsets
+  var lengths: [offsets.domain] int;
+  if size == 0 {
+    // if the segstring is empty, the lengths are also empty
+    return lengths;
+  }
+  // save off the last index of offsets
+  const high = offsets.domain.high;
+  forall (i, o, l) in zip(offsets.domain, offsets, lengths) {
```
bradcray, line 551: Style choice, but we could also equivalently just iterate over `offsets.domain` and index into `offsets[i]` and `lengths[i]` rather than using zippered iteration.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+
+We will use `getLengths` and the `SegString s` from [example 15](#ex15):
+```chapel
+use CopyAggregation;
+
+/*
+  Given a SegString, return a new SegString with all lowercase characters from the original replaced with their uppercase equivalent
+  :returns: Strings – Substrings with lowercase characters replaced with uppercase equivalent
+*/
+proc upper() throws {
```
bradcray, line 608: Just noting that `throws` is used here without explanation or rationale.  (I think the rationale is that you're calling throwing functions and don't want to handle errors, so are throwing them upward to not have the compiler complain at you?  And that's true within Arkouda proper.  I _believe_ though that for a standalone toy example like this, you could probably leave off the `throws` and the program would halt if an error were thrown, complaining about it being unhandled.  The difference being whether you're within an explicit `module` declaration, or just sketching out some file-scope code like this (IIRC).

If I've got all that straight, I might just remove the `throws` from these examples.  If I'm wrong, you should probably give an explanation of why it's here. Or possibly introduce `throws` much earlier, like when `proc` is introduced?

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+Syntax:
+```chapel
+// without aggregation
+forall (i, v) in zip(inds, vals) {
+  remoteArr[i] = v;
+}
+
+// with aggregation
+use CopyAggregation;
+forall (i, v) in zip(inds, vals) with (var agg = new DstAggregator(int)) {
```
bradcray, line 589: This starts using `with (var ...)` somewhat out of the blue.  It could be useful to introduce this earlier, when the forall loops were first introduced, as introducing a task-local variable.  That'd make it a nice companion to `ref` and/or `reduce` intents, each of which are also introducing task-private variables of one form or another.  Alternatively, could try to describe it briefly here.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+writeln("iv[truth] = ", [(t, v) in zip(truth, iv)] if t then v, "\n");
+
+// we then create the return array `Y`
+// it contains all the elements where the condition is met
+// so its size is the number of `True`s i.e. `+ reduce truth`
+var Y: [0..#(+ reduce truth)] int;
+writeln("+ reduce truth = ", + reduce truth);
+writeln("0..#(+ reduce truth) = ", 0..#(+ reduce truth), "\n");
+
+// now that we have the set up, it's time for the actual indexing
```
bradcray, line 736: "setup"?

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+var X = [1, 2, 5, 5, 1, 5, 2, 5, 3, 1];
+writeln("X = ", X, "\n");
+
+// we begin by creating a boolean array, `truth`, indicating where the condition is met
+var truth = X == 5;
+writeln("truth = ", truth);
+
+// we use `truth` to create the indices, `iv`, into the compressed array
+// note: `iv[truth] = [0, 1, 2, 3]`
+var iv = + scan truth - truth;
```
bradcray, line 725: I was wondering, when you were introducing scans, whether you should define inclusive vs. exclusive scans and point out that Chapel's are inclusive.  This made me wonder again, since I think the `- truth;` is essentially changing the inclusive result into an exclusive scan.  Note that we'd like for Chapel to support exclusive scans as well, but it's never been prioritized...

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+var Y: [0..#(+ reduce truth)] int;
+writeln("+ reduce truth = ", + reduce truth);
+writeln("0..#(+ reduce truth) = ", 0..#(+ reduce truth), "\n");
+
+// now that we have the set up, it's time for the actual indexing
+// we do a may-parallel `forall` to iterate over the indices of `X`
+// we filter on `truth[i]`, so we only act if the condition is met
+// we use the compressed indices `iv[i]` to write into `Y`
+// while using the original indices `i` to get the correct value from `X`
+[i in X.domain] if truth[i] {Y[iv[i]] = X[i];}
```
bradcray, line 741: Note that you could write this using `then` rather than curly brackets if desired...  Though you may be trying to establish a form where conditional expressions use then/else and conditional statements use curly brackets (?).

bradcray (commented):
Impressive!  I've made a number of stylistic suggestions or potential opportunities for clarification.  The one show-stopper I saw related to the reduce intent being misused, but in a way that didn't generate a bug for your program.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+The command to run the executable is:
+```console
+./tutorial
+```
+To follow along, comment out all but the relevant example in `Tutorial.chpl`, compile, run, and verify the output matches the guide!
+
+<a id="ranges_and_domains"></a>
+## Ranges and Domains
+This will only cover enough to provide context for the other examples (the very basics). More information can be found in the Chapel docs for
+[ranges](https://chapel-lang.org/docs/primers/ranges.html) and [domains](https://chapel-lang.org/docs/language/spec/domains.html).
```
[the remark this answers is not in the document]
pierce314159: Ah yeah this was intentional albeit a bit arbitrary. I really liked the examples in the ranges primer and the explanation in the domains spec, so I just kinda rolled with that hahaha.

Upon revisit, the domains primer is probably better suited for this introduction (and it also links the domain spec at the bottom for anyone who wants to dive deeper!) so I'll go ahead and update that

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+
+// you can access an existing array's domain using `.domain`
+var boolArray: [makeDistDom(10)] bool;
+var uintArray: [boolArray.domain] uint;
+```
+Note: `makeDistDom` is a helper function in Arkouda and not part of Chapel.
+
+<a id="init_fact"></a>
+## Initial Factorial
+Let's start our chapel journey by writing a function to calculate the factorial of a given integer `n`.
```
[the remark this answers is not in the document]
pierce314159: Thanks! Yeah I'm glad u pointed this out! I def prefer to be consistent

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+## `forall` Loops
+<a id="ex6"></a>
+#### Example 6: Factorial with must-parallel `forall` and Reduction
+Back to parallel factorial, let's use `reduce` to combine the results of each task.
+This a [task intent](https://chapel-lang.org/docs/primers/forallLoops.html#task-intents-and-shadow-variables) as signified by the `with` keyword.
+```Chapel
+proc factorial(n: int) {
+  var fact: int = 1;
+  forall i in 1..n with (* reduce fact) {
+    fact = i;
```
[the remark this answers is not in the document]
pierce314159: Thanks so much for catching this!! I was thinking about this incorrectly as my explanation shows :)

This makes a lot of sense! I appreciate u taking the time to review and ensuring this misunderstanding didn't get passed along to new arkouda devs

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+$ chpl Tutorial.chpl -o tutorial
+Tutorial.chpl:41: In function 'factorial':
+Tutorial.chpl:44: error: cannot assign to const variable
+Tutorial.chpl:43: note: The shadow variable 'fact' is constant due to forall intents in this loop
+```
+That's not what we want to see! There was an error during compilation.
+The issue is different iterations of the parallel loop could be modifying the outer variable `fact` simultaneously.
+
+Chapel handles this by giving each task its own copy of `fact` known as a [shadow variable](https://chapel-lang.org/docs/primers/forallLoops.html?highlight=intents#task-intents-and-shadow-variables).
+To combine the shadow variables and write the result back into `fact`, we'll need to learn about `reduction` operations.
```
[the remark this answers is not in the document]
pierce314159: that's interesting! But yeah i think ill shy away from it for beginners

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+Introspection is determining properties of a function argument at runtime.
+This is often used to determine the type and/or domain of a function argument.
+Introspection can avoid duplicating a `proc` for multiple types when none of the logic has changed.
+Using introspection will result in a [generic](https://chapel-lang.org/docs/language/spec/generics.html#generics) function.
+
+The syntax for this is:
+```Chapel
+proc foo(array: [?D] ?t, value: t, array2: [?D2] ?t2) {
+  // using `D` and `t` in this proc will refer to `array`'s domain and type respectively
+  // `value` has the same type as `array` so there's no need for the `?`
```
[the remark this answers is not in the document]
pierce314159: good points! Ill update this with these suggestions

training/CHAPEL_TUTORIAL.md
pierce314159: `size` is the number of strings in the `SegString` which will always be equivalent to `offsets.size`. I'll add a comment for this, thanks!

and the single keyword for all the declarations is really neat! I don't know how i've gotten this far without seeing that, but I'll def be using it now!

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+}
+writeln(factorial(5));
+```
+```console
+120
+```
+
+For a specific `n`, we  can do this in one line!
+```Chapel
+writeln(* reduce [i in 1..#5] i);
```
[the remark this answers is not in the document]
bradcray: I think my proposed re-flow could still accomplish a logical ordering.  I'm imagining:
* forall loop + problem with doing the naive thing approach, as it currently stands
* then introduce reduce expressions
* then introduce foreach loops (or do this prior to reduce expressions) and show how the compact versions of factorial can be written
* then go back to the forall-loop based version and introduce the reduce intent pointing out that "sometimes you want to write big loops like this, and trying to stuff it into a foreach expression just makes things less clear; or sometimes you will want to reduce multiple things in these loops", so the reduce task intents can handle such cases."  (could even note that reduce expressions are essentially rewritten by the compiler into this style of forall loop.

In saying that, though, I don't want to sound at all insistent on making this change.  Just trying to point out that I think there is a flow that still starts with things that are most familiar and simple, pushing the reduce intents to the end.  But I also realize it's a big change and that what you have is probably good enough.

training/CHAPEL_TUTORIAL.md
pierce314159: I went back and forth on this... I decided to go with this implementation to stay similar to class method `getLengths` in `SegmentedString` since the `offsets`, `values`, and `size` are attributes of a `SegString` they are declared outside the scope of the function. It prob doesn't matter that much though

training/CHAPEL_TUTORIAL.md
pierce314159: yeah I just stuck with stylistic decision in the original `getLengths`

training/CHAPEL_TUTORIAL.md
pierce314159: good point! I was following the reference function in arkouda a bit too closely here!

training/CHAPEL_TUTORIAL.md
pierce314159: good point!

training/CHAPEL_TUTORIAL.md
pierce314159: Yeah this is the general rule of thumb of how i've seen things structured in arkouda, so i tried to stick to examples that are most syntactically similar.

training/CHAPEL_TUTORIAL.md
bradcray: Rather than a comment, maybe just change to `const size = offsets.size;`?  Makes it bulletproof and self-descriptive.

training/CHAPEL_TUTORIAL.md
```diff
@@ -0,0 +1,861 @@
...
+  - `+ scan a` behaves like [`np.cumsum(a)`](https://numpy.org/doc/stable/reference/generated/numpy.cumsum.html) in numpy
+- `op reduce array`
+  - reduces the result of a scan to a single summary value
+  - returns a single value
+  - `+ reduce a` behaves like [`sum(a)`](https://docs.python.org/3/library/functions.html#sum) in python
+
+<a id="ex5"></a>
+#### Example 5: `reduce` and `scan`
+```Chapel
+var a: [0..#5] int = [1, 2, 3, 4, 5];
```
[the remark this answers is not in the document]
pierce314159: The `0..#n` syntax is much more common in arkouda, but I do really like `0..<n`. I might start using that enough that it will be needed in the next update of these docs!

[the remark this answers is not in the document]
bradcray: Arkouda development started very close to the time that `..<` was added to the language, which probably explains its choice.  If the Arkouda devs as a whole agreed with you/us, I think it'd be worth filing a PR to just change all `0..#` occurrences to `0..<`.  Note that `#` is still very useful in other cases like `lo..#segSize`, so I'm not trying to throw the operator under the bus... just not  so wild about it in the `0..n-1` context anymore.

training/CHAPEL_TUTORIAL.md
pierce314159: sorry I actually did both, I just didn't make that clear in the reply!

Ethan-DeBandi99 (approved):
I think this is at a good point now.

PR #1822, opened 2022-10-04
Fix errors with large Parquet files

reuster986:
This PR fixes issues that occur when reading or writing "large" Parquet columns:
* When writing string columns, Parquet limits the chunk size to `2**31 - 2`, but arkouda currently puts everything in one chunk per file, so if a `Strings` has more than 2 GB of data per locale, it will cause an error. This patch detects when a chunk is full and starts a new one. I don't know if this same problem exists for other dtypes, but we might need to apply a similar fix in the future.
* When reading Parquet files, arkouda had been returning prematurely after reading only one row group. This patch ensures that all row groups are read.

These changes have been tested on real data at scale to ensure equivalent results between Parquet and HDF5.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -273,7 +273,7 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
         auto chpl_ptr = (int64_t*)chpl_arr;
         parquet::Int64Reader* reader =
           static_cast<parquet::Int64Reader*>(column_reader.get());
-        reader->Skip(startIdx);
+        startIdx -= reader->Skip(startIdx);
 
         while (reader->HasNext() && i < numElems) {
           if((numElems - i) < batchSize)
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
I don't see any issues

src/ArrowFunctions.cpp
```diff
@@ -273,7 +273,7 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
         auto chpl_ptr = (int64_t*)chpl_arr;
         parquet::Int64Reader* reader =
           static_cast<parquet::Int64Reader*>(column_reader.get());
-        reader->Skip(startIdx);
+        startIdx -= reader->Skip(startIdx);
```
bmcdonald3, line 276: I think that I am missing something with these subtractions, is there a reason for subtracting the result of the `Skip` call from `startIdx`? It doesn't look like the value is used again after this and I think that the `Skip()` function will loop until it has skipped the correct number of values.

bmcdonald3: Ooh, my mistake, it is in that loop, of course...

bmcdonald3 (approved):
This makes me wonder how we could be testing this Parquet code better, given most of our tests are relying on either small files or ones that we have written in Arkouda.

I don't think that we would want to store a huge file that was written externally in the repository, but this seems like something that I should have caught in testing... Thanks for fixing this Bill.

pierce314159 (approved):

PR #1824, opened 2022-10-05
Closes #1825: Fix aggregator use in median reduction

### LINKED ISSUE #1825: Fix aggregator use in median reduction
The newly added median reduction was trying to aggregate a temporary value, which meant that when the aggregator went to flush and copy the values over, the value had gone out of scope and the reference was to garbage memory, which may or may not still have held the correct value.

This was causing some sporadic correctness issues that seemed to increase in frequency when running with oversubscription.

This failure was caught by our nightly testing.

bmcdonald3:
The newly added median reduction was trying to aggregate a temporary value, which meant that when the aggregator went to flush and copy the values over, the value had gone out of scope and the reference was to garbage memory, which may or may not still have held the correct value.

This PR introduces intermediary arrays that remove the temporary value from the aggregator and also aggregates the `GET`s to the `sortedVal`s array.

Closes: https://github.com/Bears-R-Us/arkouda/issues/1825

### DIFF
--- a/src/ReductionMsg.chpl
+++ b/src/ReductionMsg.chpl
@@ -832,17 +832,28 @@ module ReductionMsg
         valsAgg.copy(sv, noNanVals[idx]);
       }
 
-      // Now that we have sorted segments, grab the middle element
-      forall (s, c, r) in zip(segments, counts, res) with (var resAgg = newSrcAggregator(real)) {
+      var tmp1: [D] t;
+      var tmp2: [D] t;
+      forall (s, c, r, t1, t2) in zip(segments, counts, res, tmp1, tmp2) with (var resAgg = newSrcAggregator(t)) {
         if c % 2 != 0 {
           // odd case: grab middle of sorted values
-          resAgg.copy(r, sortedVals[s + (c + 1)/2 - 1]:real);
+          resAgg.copy(t1, sortedVals[s + (c + 1)/2 - 1]);
         }
         else {
           // even case: average the 2 middles of the sorted values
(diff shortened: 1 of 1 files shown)


src/ReductionMsg.chpl
```diff
@@ -832,17 +832,28 @@ module ReductionMsg
...
+      var tmp2: [D] t;
+      forall (s, c, r, t1, t2) in zip(segments, counts, res, tmp1, tmp2) with (var resAgg = newSrcAggregator(t)) {
         if c % 2 != 0 {
           // odd case: grab middle of sorted values
-          resAgg.copy(r, sortedVals[s + (c + 1)/2 - 1]:real);
+          resAgg.copy(t1, sortedVals[s + (c + 1)/2 - 1]);
         }
         else {
           // even case: average the 2 middles of the sorted values
-          resAgg.copy(r, (sortedVals[s + c/2 - 1]:real + sortedVals[s + c/2]:real) / 2.0 );
```
ronawho, line 843: In case it isn't obvious to others, this was the problem line. Copy aggregators can only aggregate copies today and then the thing they aggregate must outlive the aggregators. The old code turned into something like:

```chpl
var tmp = (sortedVals[s + c/2 - 1]:real + sortedVals[s + c/2]:real) / 2.0;
resAgg.copy(r, tmp);
```

and `tmp` doesn't outlive the aggregator and it's value changes between iterations. And it's also preventing aggregation because the `tmp` assignment would GET the values. 

The aggregators effectively aggregate addresses locally and when flushing will actually do the comm/memcpy to figure out values and assign.

We have longer term goals to add support for arbitrary user-defined aggregators, but today we really just have the copy ones so have to use a workaround like copying into tmps and then doing the calculations.

ronawho (approved):

Issue #1828, opened 2022-10-06
`Strings.to_ndarray()` UTF-8 Bug


Ethan-DeBandi99:
While writing the test for issue #1799, I discovered an issue with `Strings.to_ndarray` handle of `Non-ASCII` characters.

When running `Strings.to_ndarray()` with non-ascii characters, we are seeing unexpected results. For example, 
```
vals = ['Mötley', 'Crüe']
s1 = ak.array(vals)
s1.to_ndarray()
['MÃ¶tley CrÃ¼e']
```

Upon inspection, the issue has to do with how the values are stored vs how they are read in `to_ndarray`. UTF-8 can have multiple `uint(8)` values to represent a single character if the character is outside the `ASCII` range. However, when we read the `uint(8)` values in `to_ndarray`, we treat that as if each individual number represents a character which may not be true.

We need to update this to handle the `uint(8)` values as a whole string instead of individual characters.

Issue #556, opened 2020-11-03
io_test failures after #553 


ronawho:
https://github.com/mhmerrill/arkouda/pull/553 changed a [spawn call](https://github.com/mhmerrill/arkouda/pull/553/files#diff-394b8a344150500f3214611a8f8716ecb550bf84cb725c8ff705fec3437ff952R197) to use `stderr=PIPE,stdout=PIPE`. Unfortunately, piping stderr/stdout is not supported under Aries (ugni) or InfiniBand (gasnet-ibv):
 - https://github.com/chapel-lang/chapel/issues/7550
 - https://github.com/chapel-lang/chapel/issues/13387

The ugni issue is documented in the [spawn documentation](https://chapel-lang.org/docs/modules/standard/Spawn.html), though there is a similar issue for InfiniBand. Under ugni we get:

```
RuntimeError: Error: failed to spawn process and read output "SystemError: Invalid argument (spawn with more than 1 locale for CHPL_COMM=ugni with hugepages currently requires stdin, stdout, stderr=FORWARD)"
```

And under InfiniBand we're just hitting a segfault since there currently isn't a chapel level error being thrown, though we (the Chapel team) should add one.

---

Not being able to spawn things like this is general problem in HPC beyond just Chapel. `fork` doesn't work in configurations where memory has to be registered with the NIC. When not piping, we're able to use something other than fork, but when piping fork is the only option.

As a simple fix you could just capture the h5ls stderr to a different file and read that in separately (`h5ls filename > tmpFile 2> tmpFileStderr`), but a more principled solution might be to stop shelling out to `h5ls` and use the C api directly. That would also resolve https://github.com/mhmerrill/arkouda/issues/452

If better spawn support is important, we may be able to do something. Under IB you can set `IBV_FORK_SAFE=1` but that has performance implications. I'm not sure what the path forward would be for ugni though.

ronawho:
FYI @hokiegeek2

hokiegeek2:
@ronawho @mhmerrill @reuster986 I am on it now.

hokiegeek2:
@ronawho i vote for better spawn support that enables direct capture of logging messages. In the meantime, I will work on getting the logging output to tmp file

ronawho:
[...]

PR #1829, opened 2022-10-06
Closes #1799 - idna `encoding`/`decoding`

### LINKED ISSUE #1799: idna encoding/decoding
Request implementation of idna codec in arkouda.

in python for ascii string x, x.encode().decode('idna') will fail if x is not a valid encoding of a unicode string. Rather than error, request the ability to choose returned unicode string value (.eg. "","INVALID STRING", etc.)

Ethan-DeBandi99:
Closes #1799

This PR is a work around until Chapel implements codec support. It uses Chapel's `Subprocess` module to call out to Python files that convert each individual string.  The Python files are configured to return `''` in the event of a failed conversion.

`idna_encode` and `idna_decode` methods are added to the client `Strings` class.

An `EncodingMsg` module to the Arkouda server. This module takes in the cmd `encode`/`decode` and the encoding type. Currently, only `idna` is supported, but other encodings can be added in the future if needed. 

Testing has been added to ensure proper functionality. Currently, the test for `idna_decode` uses string comparison against `Strings.__repr__()` due to issues with `Strings.to_ndarray()`. Issue #1828 has been added to correct the issues with non-ascii characters in UTF-8 strings.

### DIFF
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -528,9 +528,6 @@ module SegmentedString {
         // var str_entry: string;
         const filename: string = basePath+"/src/exec/%i_tmp.txt".format(i);
         var str_entry: string = interpretAsString(origVals, off..#len);
-        // for b in interpretAsString(origVals, off..#len, borrow=true) {
-        //   str_entry = str_entry + b;
-        // }
         // use subprocessing to make a call to a python file for the encoding
         var sub = spawn(["python3", procFile, "-v", str_entry, "-f", filename]);
         sub.wait();
@@ -547,15 +544,8 @@ module SegmentedString {
       // calculate offsets and lengths
       encodeLengths = [e in encodeArr] e.numBytes;
       encodeOffsets = (+ scan encodeLengths) - encodeLengths + [i in 0..<encodeLengths.size] i;
-      // encodeOffsets = (+ scan encodeLengths) - encodeLengths;
-      // forall (i, o) in zip(0..#encodeOffsets.size, encodeOffsets){
-      //   if i != 0 {
-      //     o = o + i;
-      //   }
-      // }
       
       //calculate values for the segmentedstring
-      // var finalValues: [0..#((+ reduce encodeLengths)+encodeLengths.size)] uint(8);
       var finalValues = makeDistArray((+ reduce encodeLengths)+encodeLengths.size, uint(8));
       forall (s, o) in zip(encodeArr, encodeOffsets) with (var agg = newDstAggregator(uint(8))) {
         for (j, c) in zip(0.., s.chpl_bytes()) {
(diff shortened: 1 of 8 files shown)


src/SegmentedString.chpl
```diff
@@ -488,6 +492,69 @@ module SegmentedString {
...
+      var encodeLengths: [this.offsets.aD] int;
+
+      const lengths = this.getLengths();
+      forall (i, off, len) in zip(0..#this.size, offs, lengths) {
+        var str_entry: string;
+        for b in interpretAsString(origVals, off..#len, borrow=true) {
+          str_entry = str_entry + b;
+        }
+        // use subprocessing to make a call to a python file for the encoding
+        var sub = spawn(["python3", procFile, "-v", str_entry], stdout=pipeStyle.pipe);
```
ronawho, line 533: Unfortunately, piping does not work on HPC systems that have to register memory with the network (cray networks, infiniband.) https://github.com/Bears-R-Us/arkouda/issues/556 has more details.

I can never keep the details of this fresh in mind, but I think the workaround is to write your output to a file and then read that file in?

Ethan-DeBandi99: Good to know. Thanks. I will look into it.

src/SegmentedString.chpl
```diff
@@ -488,6 +492,69 @@ module SegmentedString {
...
+        encodeArr[i] = line.strip();
+        sub.wait();
+      }
+      // calculate offsets and lengths
+      encodeLengths = [e in encodeArr] e.numBytes;
+      encodeOffsets = (+ scan encodeLengths) - encodeLengths;
+      forall (i, o) in zip(0..#encodeOffsets.size, encodeOffsets){
+        if i != 0 {
+          o = o + i;
+        }
```
pierce314159, line 545: I feel like this should be easier, instinctively i wanted to do something like
```chapel
encodeOffsets = (+ scan encodeLengths) - encodeLengths + 0..;
```
but that doesn't seem to work :(

This *seems* to work at least in the simple example i ran, but idk if it's more or less performant
```chapel
encodeOffsets = (+ scan encodeLengths) - encodeLengths + [i in 0..<encodeLengths.size] i;
```

src/SegmentedString.chpl
```diff
@@ -525,28 +525,28 @@ module SegmentedString {
...
-        sub.stdout.readLine(line);
-        encodeArr[i] = line.strip();
+        var sub = spawn(["python3", procFile, "-v", str_entry, "-f", filename]);
         sub.wait();
+
+        // read file python wrote to
+        var encodedFile = open(filename, iomode.r);
+        var encodedStr: string;
+        var reader = encodedFile.reader();
+        var readSomething = reader.read(encodedStr);
```
pierce314159, line 539: I don't think this `readSomething` var is ever used

pierce314159 (approved):
Looks good!

PR #1830, opened 2022-10-07
Closes #1828 - `Strings.to_ndarray()` UTF-8 Bug Fix

### LINKED ISSUE #1828: `Strings.to_ndarray()` UTF-8 Bug

Ethan-DeBandi99:
Closes #1828 

This PR updates the `Strings.to_ndarray()` to use the Python codecs to properly decode `uint(8)` arrays representing Strings objs when the values represent non-ASCII characters. 

Adds testing to ensure that `Strings.to_ndarray()` properly returns the strings used to create the Arkouda `Strings` object.

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1690,7 +1690,7 @@ class Strings:
         res = np.empty(self.size, dtype=dt)
         # Form a string from each segment and store in numpy array
         for i, (o, l) in enumerate(zip(npoffsets, lengths)):
-            res[i] = np.str_(codecs.decode(npvalues[o : o + l].tolist()))
+            res[i] = np.str_(codecs.decode(b"".join(npvalues[o:o + l])))
         return res
 
     def to_list(self) -> list:
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import codecs
 import itertools
 import re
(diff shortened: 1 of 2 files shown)


hokiegeek2 (approved):
looks good! we both learned a little bit about the codecs library 👍

pierce314159 (approved):
Nice job!

Ethan-DeBandi99:
using `.tolist` is causing issues with tests, I need to dig into this and will push to review again once its fixed.

Ethan-DeBandi99:
Corrected the issues with the codecs library being used. Should be all set now.

PR #1833, opened 2022-10-12
Closes #1832 - Revert smallIdx optimization for DF indexing

bmcdonald3:
Previously, in https://github.com/Bears-R-Us/arkouda/pull/1537, an optimization for the DF _get_head_tail_server function was made. That optimization now has less value, since Elliot has optimized some of the overhead that was being avoided there. This code also introduced a bug reported in https://github.com/Bears-R-Us/arkouda/issues/1832, so this PR reverts the optimization altogether.

closes #1832

PR #1834, opened 2022-10-13
Closes #1831 - `DataFrame._get_head_tail_server` return typing bug

### LINKED ISSUE #1831: _get_head_tail_server in DataFrame dropping types when DataFrame exceeds pandas max_rows
Categorical, SegArray, and Strings are preserved, but at a minimum Fields, IPv4, and Datetime were not preserved. This happened when an arkouda DataFrame had those types, exceeded pandas max_rows, and then the corresponding arrays in the DataFrame would be either uint64 or int64 (I don't recall which) when displayed in a notebook. Incidentally, it also seems like there is an issue with Strings or perhaps only when there are multiple Strings arrays where the offsets are potentially incorrect. This causes an element in a Strings array to only display part of the string and/or the string from an incorrect row. I can try to generate a simpler exemplar if needed, but this occurred in a relatively small arkouda DataFrame with multiple Strings arrays.

Ethan-DeBandi99:
Closes #1831 

Updates the `ak.DataFrame._get_head_tail_server()` method used by the `ak.DataFrame.__repr__()` method to properly preserve the typing for the following client dtypes:
- BitVector
- Fields
- Datetime
- IPv4

Adds testing that the client types are preserved properly.

*Note for Developers - This updated the structure of Chapel code to use and `if` instead of a `select` since multiple cases became virtually identical. However, we needed to know the class of the object to properly preserve it on return.*

The issue reported with strings was resolved by PR #1833.

### DIFF
--- a/tests/dataframe_test.py
+++ b/tests/dataframe_test.py
@@ -102,7 +102,6 @@ class DataFrameTest(ArkoudaTest):
         self.assertTrue(ref_df.equals(df.to_pandas()))
 
     def test_client_type_creation(self):
-        import datetime as dt
         f = ak.Fields(ak.arange(10), ["A", "B", "c"])
         ip = ak.ip_address(ak.arange(10))
         d = ak.Datetime(ak.arange(10))
--- a/tests/dataframe_test.py
+++ b/tests/dataframe_test.py
@@ -101,6 +101,38 @@ class DataFrameTest(ArkoudaTest):
         self.assertEqual(len(df), 6)
         self.assertTrue(ref_df.equals(df.to_pandas()))
 
+    def test_client_type_creation(self):
+        import datetime as dt
+        f = ak.Fields(ak.arange(10), ["A", "B", "c"])
+        ip = ak.ip_address(ak.arange(10))
+        d = ak.Datetime(ak.arange(10))
+
+        df_dict = {
+            "fields": f,
+            "ip": ip,
+            "date": d
+        }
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99:
@pierce314159 @joshmarshall1 - This is a high priority and needs to be merged today. Please code review ASAP and let me know if I overlooked something. Thanks!

jaketrookman (approved):
Looks good

joshmarshall1 (approved):
All looks good. My only question would be can we add a `BitVector` column to the test case with the other special cases? Or is this already covered in another test?

tests/dataframe_test.py
```diff
@@ -101,6 +101,37 @@ def test_dataframe_creation(self):
...
+            "ip": ip,
+            "date": d
+        }
+        df = ak.DataFrame(df_dict)
+        pd_d = [pd.to_datetime(x, unit='ns') for x in d.to_list()]
+        pddf = pd.DataFrame({
+            "fields": f.to_list(),
+            "ip": ip.to_list(),
+            "date": pd_d
+        })
```
pierce314159, line 120: it seems `BitVector` is missing here. I'm not sure if there's a good candidate to compare this against and it wasn't specifically requested, so I'm okay leaving it out. But I wanted to make sure it was intentionally excluded

pierce314159 (approved):
Looks good!!

PR #1838, opened 2022-10-13
Closes #1837: Intermittent failures in groupby prod aggregate test

### LINKED ISSUE #1837: Intermittent failures in groupby prod aggregate test
A [github workflow in my fork](https://github.com/pierce314159/arkouda/actions/runs/3238079925/jobs/5305892891) saw a failure in the groupby prod aggregation. 
```
Doing aggregate(uint64, prod)
Different values (abs diff = 1.8446744073709564e+19)
```
I believe this a similar issue to #1529 where we need to `np.allclose` for float equality

pierce314159:
This PR (Closes #1837)
- Updates code using `equality_helper` to use `np.allclose(equal_nan=True)` instead

A [github workflow in my fork](https://github.com/pierce314159/arkouda/actions/runs/3238079925/jobs/5305892891) saw a failure in the groupby prod aggregation in `test_groupby_on_one_level`
```
Doing aggregate(uint64, prod)
Different values (abs diff = 1.8446744073709564e+19)
```
I'm not sure if I'm just misinterpreting, but I think this should be within the tolerance of [`np.allclose`](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html). So I'm not sure how this happened. Maybe there's something wrong with the way I calculate if the arrays are `NAN` in the same places in `equality_helper`?

While looking into this, I noticed the `equality_helper` is not necessary if we use the `equal_nan` flag. So I went ahead and updated this (and spots in `operator_test`) to use the functionality built into numpy instead. Hopefully this take care of the problem

### DIFF
--- a/tests/groupby_test.py
+++ b/tests/groupby_test.py
@@ -57,12 +57,7 @@ def compare_keys(pdkeys, akkeys, levels, pdvals, akvals) -> int:
                 print("Different keys")
                 return 1
 
-    def equality_helper(a, n):
-        # verify NAN in the same locations and equal in the locations they are not NAN
-        # Note: we have to do this becasue NAN != NAN
-        return ((nin := np.isnan(n)) == (ain := np.isnan(a))).all() and np.allclose(n[~nin], a[~ain])
-
-    if not equality_helper(pdvals, akvals):
+    if not np.allclose(pdvals, akvals, equal_nan=True):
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Based on the background you provided, this change makes sense to me.

joshmarshall1 (approved):

PR #1845, opened 2022-10-21
Closes #1559, #1558, & #487 - HDF5 Upgrades

### LINKED ISSUE #487: enable writing to single HDF5 file
Multiple users have expressed a moderate desire to be able to save arrays to a single HDF5 file, as opposed to one file per locale. I think this is possible but perhaps requires a special version of HDF5?

### LINKED ISSUE #1558: Integrate Existing HDF5 workflows into MultiDim workflows
Currently, the multidimensional support for HDF5 is in its own module. That was done as an initial iteration to get things working for multidimensional dataset without impacting the existing workflows.

- This should remain a separate module as it will not yet be configured to write a file per locale or be fully configured to efficiency. 
- This should add the additional functionality for file management as well. 
- Configure the attributes for array datasets.
- Configure the ability to read datasets that were written by arkouda before updates.
  - Add to the documentation for the schema of HDF5

### LINKED ISSUE #1559: HDF5 Multi-Locale Support
This is to address the updated HDF5 module that was not initially configured to work properly on multilocale systems.

HDF5 has a separate module configured for MultiDimensional objects. This needs to be configured to work for multi-locale systems to read and write from multiple files. Additionally, we will need to update the functionality to work per locale and be efficient.

Ethan-DeBandi99:
Closes #487 
Closes #1558
Closes #1559

I know this is a large PR, but it made the most sense to work all of this at once and slim down the code base, rather than work things individually and not end with an optimal result. 

- Adds Distributed file storage for `ArrayView` Objects
- Adds single file storage for `pdarray` and `strings` objects
- Moves `ArrayView` HDF5 save/load code into the main module
- Updates the `HDF5Msg` module to reduce the amount of code needed and prevent repetition.
- Maintains the previous `HDF5Msg` module as `HDF5Msg_LEGACY`. This allows for any files written using v2022.10.13 or earlier to be read. In these instances, the files should be rewritten with the updated `HDF5Msg` code. The `HDF5MultiDim` and `HDF5Msg_LEGACY` modules only support reading of old file formats and will be removed in a future release.
- Added a flag to `read`/`load` functionality to use the old HDF5 code for reading old files.
- Updated client code to properly configure messages to the format now required by the server. Added `ArrayView` handling to the main IO `write`/`read` functionality.
- Updated `ArrayView` to always store as a `flattened` array. This simplifies the read and write processes.
- Updated testing to ensure proper functionality in several tests. These changes were minor, but ensure the test files have the proper information. 
- The calls now require different parameters than `Parquet` to write. We will need to update our `Parquet` code to be in line with HDF5.
- Corrects and issue in the previous version where files would not be truncated properly.
- `HDF5Msg_LEGACY` was previously `HDF5Msg` and contains no new code. This is maintained for legacy read functionality.

**PLEASE NOTE: The majority of the code in `HDF5Msg` module has been refactored and reorganized. Viewing it as a comparison may not be the best option because most of the file appears to change. I would recommend this be reviewed as a new file.**


src/HDF5Msg.chpl
```diff
@@ -24,1759 +26,1622 @@ module HDF5Msg {
...
+    proc getDataType(type t) : C_HDF5.hid_t {
+        if t == bool {
+            return C_HDF5.H5T_NATIVE_HBOOL;
         }
-        
-        try {
+        else {
+            return getHDF5Type(t);
+        }
+    }
```
pierce314159, line 85: let's double check that we're actually using this helper

arkouda/array_view.py
```diff
@@ -378,21 +378,23 @@ def save(self, filepath: str, dset: str, mode: str = "truncate", storage: str =
             Default: truncate
             Mode to write the dataset in. Truncate will overwrite any existing files.
             Append will add the dataset to an existing file.
-        storage: str (Flat | Multi)
-            Default: Flat
-            Method to use when storing the dataset.
-            Flat - flatten the multi-dimensional object into a 1-D array of values
-            Multi - Store the object in the multidimensional presentation.
+        file_type: str (single|distribute)
+            efault: distribute
```
bmcdonald3, line 382: "efault" -> Default

src/HDF5Msg.chpl
```diff
@@ -24,1759 +26,1622 @@ module HDF5Msg {
...
-                                'adjusted locale %i for left shuffle to %i'.format(idx,idx-1)); 
-                    } else {
-                        h5Logger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-                                'no left shuffle adjustment for locale %i'.format(idx));
+                var dtype: C_HDF5.hid_t;
+                
+                select entryDtype {
+                    when DType.Int64 {
+                        var flat = toSymEntry(toGenSymEntry(entry), int);
+                        var localFlat = new lowLevelLocalizingSlice(flat.a, 0..#flat.size);
```
bmcdonald3, line 491: Not suggesting we make any changes now, but I think with very large arrays that can't all be loaded into memory on a single node, we may want to just serially loop over the locales and then append to a single file so that no nodes ever have to hold the entire array.

Also, not sure what I would expect to perform better given the extra communication needed here. I think it would just depend on what the intended use case is of this function (e.g., are we writing only relatively small pdarrays to single files?)

bmcdonald3 (commented):
Changes look good to me, just a couple of comments.

src/HDF5Msg.chpl
Ethan-DeBandi99: @bmcdonald3 - I believe the intended case here is for small arrays. However, I will verify that and if future modifications are required for this to perform more optimally, I will add issues to handle those changes.

src/HDF5Msg.chpl
Ethan-DeBandi99: This is being used on ln 390. This usage in `writeLocalDset` mimic the call to `getDataType` from ln 65 that takes in an array when writing to a distributed file set. We could add it to a few spots where we `var dtype = C_HDF5.H5T_NATIVE_HBOOL` but that seems like overkill in situations where we know that value is to be set.

arkouda/util.py
```diff
@@ -6,6 +6,7 @@
 import h5py  # type: ignore
 import numpy as np  # type: ignore
 
+import arkouda
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Fixed. This is supposed to be `import arkouda.array_view` so that references to `arkouda.array_view.ArrayView` function properly.

arkouda/pdarrayclass.py
```diff
@@ -1539,7 +1568,7 @@ def register(self, user_defined_name: str) -> pdarray:
         RegistrationError
             If the server was unable to register the pdarray with the user_defined_name
             If the user is attempting to register more than one pdarray with the same name,
-            the former should be unregistered first to free up the registration name.
+            the former should be unregisterzed first to free up the registration name.
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Fixed

arkouda/array_view.py
```diff
@@ -378,21 +378,23 @@ def save(self, filepath: str, dset: str, mode: str = "truncate", storage: str =
...
         """
-        write_hdf5_multi_dim(self, filepath, dset, mode=mode, storage=storage)
+        write_hdf5_multi_dim(self, filepath, dset, mode=mode, file_type=file_type)
 
     @staticmethod
     def load(filepath: str, dset: str) -> ArrayView:
         """
+        DEPRECATED
+        This function is being mantained to allow reading from files written in Arkouda v2022.10.13
+        or earlier. If used, save the object to update formatting.
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Added deprecation warning.

I also added one to `read_hdf5_multi_dim`

arkouda/array_view.py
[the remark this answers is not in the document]
Ethan-DeBandi99: fixed

Ethan-DeBandi99: fixed

bmcdonald3 (approved):

pierce314159 (approved):
This looks good to me! Nice work!

src/HDF5Msg.chpl
```diff
@@ -24,1759 +26,1621 @@ module HDF5Msg {
...
+        where (MyDmap == Dmap.blockDist || MyDmap == Dmap.defaultRectangular)
+    {
+        h5Logger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                "entry.a.targetLocales() = %t".format(A.targetLocales()));
+        h5Logger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                "Filedomains: %t".format(filedomains));
+        h5Logger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                "skips: %t".format(skips));
+
+        coforall loc in A.targetLocales() do on loc {
```
bmcdonald3, line 1112: This is where one of the missing localize calls was lost. The `coforall` over `filenames.domain` and then the subsequent indexing into filenames was preserving the locality. This is definitely the preferred method over what I've opened in https://github.com/Bears-R-Us/arkouda/pull/1879, but I won't have a chance to do a more thorough look through this to see if there are other similar things that need to be updated.

PR #1851, opened 2022-10-21
Closes #1776: GroupBy Strings Optimization

### LINKED ISSUE #1776: Optimizations for GroupBy on Strings
The current `GroupBy` logic for strings is optimized for grouping a single array of long (or variable-length) strings. However, there are at least two other cases that could benefit from different logic:
- A single array of short strings. Currently, all strings get hashed to 128-bit values, which are then sorted to achieve grouping. However, if the maximum length of the strings is 16 bytes or less, then directly sorting the strings would use the same or fewer radix digits as sorting the hashes, and you wouldn't have to compute the hashes. In practice, it might even make sense to set the length threshold a little higher, e.g. directly sort strings when the max length is 20 bytes or less.
- Multiple arrays, at least one of which is strings. Currently, when grouping by multiple arrays, arkouda will first hash any strings arrays (via `Strings._get_grouping_keys()`) so that all arrays are represented by integers, and then it will hash all the integer representations (inside `UniqueMsg.chpl`) and accumulate those into a single array of 128-bit hashes for sorting. In this situation, strings effectively get hashed twice, which is wasteful. I think we can skip the first hash and rework `UniqueMsg` to just hash each strings array once.

pierce314159:
This PR (Closes #1776):
- Removes duplicated string hash in `Strings._get_grouping_keys()` since this will be performed on all arrays during `unique`
- When perfroming groupby on a single strings array with max length < 16 bytes, we now concat the bytes into 1 or 2 uint arrays and then use the regular numeric unique logic. This is hopefully more efficient than uniquing/sorting the hashed arrays (since 16 bytes <= 128 bits from the hash and we don't have to do the computation of the hash)
- Combines the logic in `assumeSortedShortcut` and `uniqueAndCount` 

The performance of these changes will be tracked by the `String Groupby Performance` benchmark and the `Small String Groupby Performance` benchmark

NOTE:
This takes advantage of `computeOnSegments`, but I had to duplicate the logic in `twoReturnComputeOnSegments` because I need to return 2 arrays. I really hate doing that so I would love suggestions on a better way to accomplish this

### DIFF
--- a/src/UniqueMsg.chpl
+++ b/src/UniqueMsg.chpl
@@ -33,38 +33,6 @@ module UniqueMsg
     private config const logLevel = ServerConfig.logLevel;
     const umLogger = new Logger(logLevel);
 
-    proc assumeSortedShortcut(n, namesList: [] string, typesList: [] string, st) throws {
-      // very similar to uniqueAndCount but skips sort
-      var (size, hasStr, names, types) = validateArraysSameLength(n, namesList, typesList, st);
-      if (size == 0) {
-        return (new shared SymEntry(0, int), new shared SymEntry(0, int));
-      }
-      proc skipSortHelper(type t, keys: [?D] t) throws {
-        // skip sorting, set permutation to 0..#size and go directly to finding segment boundaries.
-        var permutation = new shared SymEntry(keys.size, int);
-        permutation.a = permutation.aD;
-        var (uniqueKeys, counts) = uniqueFromSorted(keys);
-        var segments = new shared SymEntry(counts.size, int);
-        segments.a = (+ scan counts) - counts;
-        return (permutation, segments);
-      }
-
-      // If no strings are present and row values can fit in 128 bits (8 digits),
-      // then pack into tuples of uint(16) for sorting keys.
-      if !hasStr {
-        var (totalDigits, bitWidths, negs) = getNumDigitsNumericArrays(names, st);
-        if totalDigits <= 2 { return skipSortHelper(2*uint(bitsPerDigit), mergeNumericArrays(2, size, totalDigits, bitWidths, negs, names, st)); }
-        if totalDigits <= 4 { return skipSortHelper(4*uint(bitsPerDigit), mergeNumericArrays(4, size, totalDigits, bitWidths, negs, names, st)); }
-        if totalDigits <= 6 { return skipSortHelper(6*uint(bitsPerDigit), mergeNumericArrays(6, size, totalDigits, bitWidths, negs, names, st)); }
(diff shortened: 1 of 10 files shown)


src/UniqueMsg.chpl
```diff
@@ -48,7 +48,64 @@ module UniqueMsg
...
+            st.addEntry(concat1Name, new shared SymEntry(numeric1));
+            st.addEntry(concat2Name, new shared SymEntry(numeric2));
+            var (totalDigits, bitWidths, negs) = getNumDigitsNumericArrays(names, st);
+            if totalDigits <= 2 { return skipSortHelper(2*uint(bitsPerDigit), mergeNumericArrays(2, size, totalDigits, bitWidths, negs, names, st)); }
+            if totalDigits <= 4 { return skipSortHelper(4*uint(bitsPerDigit), mergeNumericArrays(4, size, totalDigits, bitWidths, negs, names, st)); }
+            if totalDigits <= 6 { return skipSortHelper(6*uint(bitsPerDigit), mergeNumericArrays(6, size, totalDigits, bitWidths, negs, names, st)); }
+            if totalDigits <= 8 { return skipSortHelper(8*uint(bitsPerDigit), mergeNumericArrays(8, size, totalDigits, bitWidths, negs, names, st)); }
+          }
+        }
+      }
```
Ethan-DeBandi99, line 108: Could we pull this logic out into a function since it is reused below to reduce duplicated code?

Ethan-DeBandi99 (commented):
Overall looks good. Just one question on reducing some of the code.

src/UniqueMsg.chpl
pierce314159: You're exactly right! It's duplicated in the `assumeSortedShortcut` and `uniqueAndCount` functions. Both of these functions do basically the same thing (except `assumeSortedShortcut` skips the sort). I implemented these for an important bug a while back so I went with the approach with the quickest turn around. I am planning on combining the two functions into one during the next iteration which will eliminate the duplicated code

Ethan-DeBandi99 (approved):

pierce314159:
I figure I might as well do the next iteration now and give the nightly a few more runs. I went ahead and combined the logic from the `assumeSortedShortcut` into `uniqueAndCount`. Once I rewrite the `bytesToUintArr` to take advantage of `computeOnSegments`, I think this will be in a good state

Ethan-DeBandi99 (approved):
This looks good to me. Chatted with @pierce314159 about a few things offline, but they were already on his radar for future considerations.

joshmarshall1 (approved):

PR #1848, opened 2022-10-21
Closes #1847: Small String GroupBy benchmark

pierce314159:
This PR (closes #1847):
- Adds benchmark comparing strings containing < 8 bytes, < 16 bytes, and > 16 bytes. This will track the single array of small strings optimization in #1776

PR #1853, opened 2022-10-24
Closes #1852: Add small string groupby perf to `run_benchmarks`

### LINKED ISSUE #1852: Add small string groupby performance to `run_benchmarks`
Add small string groupby performance to `run_benchmarks`. In response to comments left in #1848

pierce314159:
This PR (closes #1852) adds small string groupby perf to `run_benchmarks` so it will get tested in the nightly.

EDIT:
I was misinterpreting the feedback and have moved the graph to `arkouda-string.graph`. My original note is in the quote below
> The PR feedback on #1848 mentioned `arkouda-string.graph` might be a more appropriate place for this benchmark rather than it's own graph. I considered having it there originally, but I felt like it wasn't really comparable to the other keys since they are comparing how groupby scales with multiple arrays 1,2,8,16 and this is tracking the performance based on the size of the strings within the SegString. That being said, I don't have terribly strong opinions on this and would be happy to change it if others feel that's the way to go

### DIFF
--- a/benchmarks/graph_infra/arkouda-string.graph
+++ b/benchmarks/graph_infra/arkouda-string.graph
@@ -16,6 +16,12 @@ files: str-groupby.dat, str-groupby.dat, str-groupby.dat, str-groupby.dat
 graphtitle: String Groupby Performance
 ylabel: Performance (GiB/S)
 
+perfkeys: small str array Average rate =, medium str array Average rate =, big str array Average rate =
+graphkeys: small str array GiB/s, medium str array GiB/s, big str array GiB/s
+files: small-str-groupby.dat, small-str-groupby.dat, small-str-groupby.dat
+graphtitle: Small String Groupby Performance
+ylabel: Performance (GiB/s)
+
 perfkeys: Average rate =
 graphkeys: Gather GiB/s
 files: str-gather.dat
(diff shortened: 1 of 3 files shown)


Ethan-DeBandi99 (approved):

ronawho:
> The PR feedback on https://github.com/Bears-R-Us/arkouda/pull/1848 mentioned arkouda-string.graph might be a more appropriate place for this benchmark rather than it's own graph. I considered having it there originally, but I felt like it wasn't really comparable to the other keys since they are comparing how groupby scales with multiple arrays 1,2,8,16 and this is tracking the performance based on the size of the strings within the SegString.

I think it also includes normal string sort and string gathers and other string operations, so seems like a good spot to me. Ultimately, the only thing it really controls is the suite the graphs appear in, so it's not a big deal either way.

pierce314159:
> > The PR feedback on #1848 mentioned arkouda-string.graph might be a more appropriate place for this benchmark rather than it's own graph. I considered having it there originally, but I felt like it wasn't really comparable to the other keys since they are comparing how groupby scales with multiple arrays 1,2,8,16 and this is tracking the performance based on the size of the strings within the SegString.
> 
> I think it also includes normal string sort and string gathers and other string operations, so seems like a good spot to me. Ultimately, the only thing it really controls is the suite the graphs appear in, so it's not a big deal either way.

Oh I see!! sorry i thought you were suggesting we add this to the `String Groupby Performance` chart, my bad! I think you're right, I'll move it over there

PR #1861, opened 2022-10-26
Closes #1801 - Reorganizes Symbol Table

### LINKED ISSUE #1801: Add Layer to Symbol Table & Remove `CompositeSymEntry`
In its current state, `GenSymEntry` requires a Chapel type be provided for the purpose of setting the `dtype`. However, while moving the `GroupBy` object to the server, it has become apparent that we will have situations where `dtype` does not make sense as a property. For example, in `GroupBy` the keys can be made up of various objects of various dtypes. Additionally, each can be accessed as needed from each individual key. Because of this, an untyped version of `GenSymEntry` is needed. The current `GenSymEntry` would inherit from the untyped in order to minimize duplication of code. This will not impact current functionality but will provide more flexibility in cases like `GroupBy`.

It has also become apparent that `CompositeSymEntry` is unnecessary. It does not add any information currently. It only inherits from `GenSymEntry` and the constructor only calls `super.init`. As a result, this can be removed. Any SymEntries inheriting from composite should switch to inherit to `GenSymEntry` without issue.

Overall, this should provide us with more flexibility and a structure that is easier to follow moving forward.

Ethan-DeBandi99:
Closes #1801

Reorganizes the symbol table so that the structure is more intuitive. `GenSymEntry` now allows for entires that contain multiple symbol table entries. In these cases the multiple entries are required to represent a singular object. For example, `SegArraySymEntry` contains `values` and `offsets` `SymEntries`. However, the each independently is required to conceptualize the `SegArray`. These objects also have a defined type. `CompositeSymEntry` is to be used for `SymEntries` that contain multiple entries, but those entries have varying types and are not necessarily required to conceptualize the full object.

**NOTE - At this time only `GroupBySymEntry` should inherit from `CompositeSymEntry`.**

- Updates `SegStringSymEntry` and `SegArraySymEntry` to inherit from `GenSymEntry`
- Updates `GroupBySymEntry` to inherit from `CompositeSymEntry`
- Adds `attrib` case for `CompositeSymEntry`
- Adds `toCompositeSymEntry`

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -25,12 +25,14 @@ module MultiTypeSymEntry
             TypedArraySymEntry, // Parent type for Arrays with a dtype, legacy->GenSymEntry
                 PrimitiveTypedArraySymEntry, // int, uint8, bool, etc.
                 ComplexTypedArraySymEntry,   // DateTime, TimeDelta, IP Address, etc.
-                GroupBySymEntry,      // GroupBy
         
-            CompositeSymEntry, // Parent type for things which are composites of arrays
+            GenSymEntry,
                 SegStringSymEntry,    // SegString composed of offset-int[], bytes->uint(8)
                 CategoricalSymEntry,  // Categorical
                 SegArraySymEntry,     // Segmented Array
+
+            CompositeSymEntry,        // Entries that consist of multiple SymEntries of varying type
+                GroupBySymEntry,      // GroupBy
 
             AnythingSymEntry, // Placeholder to stick aritrary things in the map
             UnknownSymEntry,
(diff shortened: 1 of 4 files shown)


hokiegeek2 (approved):
I like this design change, well done.

src/MultiTypeSymEntry.chpl
```diff
@@ -439,6 +451,13 @@ module MultiTypeSymEntry
     proc toGenSymEntry(entry: borrowed AbstractSymEntry) throws {
         return (entry: borrowed GenSymEntry);
     }
+
+    /**
+     * Helper proc to cast AbstrcatSymEntry to CompositeSymEntry
```
joshmarshall1, line 456: AbstrcatSymEntry -> AbstractSymEntry

src/MultiTypeSymEntry.chpl
```diff
@@ -439,6 +451,13 @@ module MultiTypeSymEntry
...
+     */
+    proc toCompositeSymEntry(entry: borrowed AbstractSymEntry) throws {
+        return (entry: borrowed CompositeSymEntry);
+    }
     
     /**
      * Helper proc to cast AbstrcatSymEntry to SegStringSymEntry
```
joshmarshall1, line 463: I know this isn't part of the changes, but since it's the same typo as above I figured I would mention it here as well

Ethan-DeBandi99:
> I like this design change, well done.

Thanks!

joshmarshall1 (approved):
All looks good. Noticed a typo in a comment so no real issues, but figured I'd mention it

src/MultiTypeSymEntry.chpl
Ethan-DeBandi99: Fixed

src/MultiTypeSymEntry.chpl
Ethan-DeBandi99: fixed

PR #1870, opened 2022-10-28
Closes #1868 - `AbstractSymEntry` Name Property

### LINKED ISSUE #1868: `AbstractSymEntry` Name Property
`CompositeSymEntries` will usually return their component `SymEntries` for easy access from the client without additional messages. Each return adds a new entry for each component to the Symbol Table. Instead, we should track a `name` property in `AbstractSymEntry` that is set when `addEntry` is called. This way the entry is self aware of its name and can call `attrib` without adding a new entry each time.

Currently, only `SegArray` and `GroupBy` will require this update.

Ethan-DeBandi99:
Closes #1868 

Adds `name` property to `AbstractSymEntry`. This is not 100% necessary at the current time. However, this should alleviate a few potential pain points with `register`/`attach` and allow us to support `copy` functionality. Currently, we ALWAYS create a new `SymEntry` to avoid changes in the object used for creation to impact any object it is used to create. This is something we are discussing changing and this change would make that easier because we get the name of the Symbol Table entry if it already exists. Otherwise, we create the entry and return that name.

This PR also updates the `SegArraySymEntry` initialization to support the `length` component to be stored as a SymEntry instead of just being an array. This also updates the `fillReturnMap` function to call a single function passing in the entry you want to return the name of. This eliminates the need for a function to return the name of each component.

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -412,7 +412,6 @@ module MultiTypeSymEntry
             this.valuesEntry = valuesSymEntry;
 
             ref sa = segmentsSymEntry.a;
-            const low = segmentsSymEntry.aD.low;
             const high = segmentsSymEntry.aD.high;
             var lengths = [(i, s) in zip (segmentsSymEntry.aD, sa)] if i == high then valuesSymEntry.size - s else sa[i+1] - s;
             
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -396,18 +396,6 @@ module MultiTypeSymEntry
         }
     }
 
-    proc createSegArrayEntry(segmentsSymEntry: shared SymEntry, valuesSymEntry: shared SymEntry, type etype, st: borrowed SymTab): shared SegArraySymEntry throws {
-        ref sa = segmentsSymEntry.a;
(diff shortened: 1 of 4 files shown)


src/MultiTypeSymEntry.chpl
```diff
@@ -393,15 +401,23 @@ module MultiTypeSymEntry
...
+        proc init(segmentsSymEntry: shared SymEntry, valuesSymEntry: shared SymEntry, type etype) {
             super.init(etype, valuesSymEntry.size);
             this.entryType = SymbolEntryType.SegArraySymEntry;
             assignableTypes.add(this.entryType);
             this.etype = etype;
             this.segmentsEntry = segmentsSymEntry;
             this.valuesEntry = valuesSymEntry;
 
+            ref sa = segmentsSymEntry.a;
+            const low = segmentsSymEntry.aD.low;
```
pierce314159, line 415: I don't think `low` is ever used

pierce314159 (approved):
Looks good!

hokiegeek2 (approved):
looks good, well done!

src/MultiTypeSymEntry.chpl
Ethan-DeBandi99: You are right. I removed it.

joshmarshall1 (approved):

PR #1871, opened 2022-10-31
Closes #1727: Add `where` argument to `sqrt` and `power`

### LINKED ISSUE #1727: Add `where` argument to `sqrt` and `power`
Came up in PR #1726

The numpy `sqrt` and `power` functions have a `where` argument, so we should to in order to match their API. There is some question on how it should be implemented since there is some weirdness here (it doesn't behave in a way that feels intuitive to me)

From numpy's docs:
> where array_like, optional
    This condition is broadcast over the input. At locations where the condition is True, the out array will be set to the ufunc result. Elsewhere, the out array will retain its original value. Note that if an uninitialized out array is created via the default out=None, locations within it where the condition is False will remain uninitialized.

So in my head this means that the operation (sqrt or power) occurs where `where` is True and otherwise it should default to the original value of the input array. But this is not how it works at all
```python
# how it should work in my head
# >>> np.sqrt(np.array([4,9,16]), where=[True, False, True])
# array([2., 9., 4.])
# how it actually works
>>> np.sqrt(np.array([4,9,16]), where=[True, False, True])
array([2., 0., 4.])
```
This is because numpy the out array is uninitialized by default and locations within it where the condition is False will remain uninitialized. This makes the same function call inconsistent between runs
```python
>>> n = np.array([i**2 for i in range(10)])
>>> n
array([ 0,  1,  4,  9, 16, 25, 36, 49, 64, 81])

# uninitialized so the output is garbage
>>> np.sqrt(n, where=False)
array([4.65907677e-310, 0.00000000e+000, 4.35287203e-114, 6.59025685e+246,
       1.51948310e+132, 3.52161575e+180, 3.99190445e+252, 7.10534150e+159,
       1.56649479e+208, 1.71872047e+161])

>>> np.sqrt(n, where=True)
array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])

# same function call with different output...
>>> np.sqrt(n, where=False)
array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])

>>> np.sqrt(n, where= n%2 == 1)
array([0., 1., 0., 3., 0., 5., 0., 7., 0., 9.])

# same function call with different output again...
>>> np.sqrt(n, where=False)
array([0., 1., 0., 3., 0., 5., 0., 7., 0., 9.])
```

It's possible i'm misunderstanding but i feel like a much more useful design choice to return `n**(1/2)` where True and `n` where False. I would absolutely love feedback from @reuster986 here and any other users since this would be a deviation from numpy

This implementation with my suggested design choice is pretty easy:
```python
def power(pda, pwr, where = True):
    if where:
        return pda**pwr
    else:
        return where(where, pda**pwr, pda)

def sqrt(pda, where = True):
    self.power(pda, 0.5, where)
```

jaketrookman:
This PR (Closes #1727)

- Implements an updated version of the Power and Sqrt functions
- Correctly accept boolean arguments, more closely matches Numpy and returns original value if False condition is input
- Updated and added additional testing and documentation

### DIFF
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -2638,8 +2638,6 @@ def rotr(x, rot) -> pdarray:
     >>> A = ak.arange(10)
     >>> ak.rotr(1024 * A, A)
     array([0, 512, 512, 384, 256, 160, 96, 56, 32, 18])
-
-
     """
     if isinstance(x, pdarray) and x.dtype in [akint64, akuint64]:
         if (isinstance(rot, pdarray) and rot.dtype in [akint64, akuint64]) or isSupportedInt(rot):
@@ -2658,18 +2656,19 @@ def power(pda: pdarray, pwr: Union[int, float, pdarray], where: Union[bool, pdar
     Raises an array to a power. If where is given, the operation will only take place in the positions
     where the where condition is True.
 
-    Notes:
-    This function was modified to deviate from the previous implementation that matched numpy.
-    If a False condition was given, numpy would return 0. This implementation will now return
-    the original value.
+    Note:
+    Our implementation of the where argument deviates from numpy. The difference in behavior occurs
+    at positions where the where argument contains a False. In numpy, these position will have
+    uninitialized memory (which can contain anything and will vary between runs). We have chosen to
+    instead return the value of the original array in these positions.
 
     Parameters
     ----------
     pda : pdarray
-        A pdarry of values that will be raised to a power (pwr)
+        A pdarray of values that will be raised to a power (pwr)
     pwr : integer, float, or pdarray
         The power(s) that pda is raised to
-    where : Boolean or pdaarray
+    where : Boolean or pdarray
         This condition is broadcast over the input. At locations where the condition is True, the
         corresponding value will be raised to the respective power. Elsewhere, it will retain its
         original value. Default set to True.
@@ -2681,7 +2680,10 @@ def power(pda: pdarray, pwr: Union[int, float, pdarray], where: Union[bool, pdar
 
     Examples
     --------
-    >>> ak.power(ak.arange(5), 3, truth)
+    >>> a = ak.arange(5)
+    >>> ak.power(a, 3)
+    array([0, 1, 8, 27, 64])
+    >>> ak.power(a), 3, a % 2 == 0)
     array([0, 1, 8, 3, 64])
     """
     from arkouda.numeric import cast as akcast
@@ -2717,9 +2719,10 @@ def sqrt(pda: pdarray, where: Union[bool, pdarray] = True) -> pdarray:
         Returns a pdarray of square rooted values, under the boolean where condition.
 
     Examples:
-    >>> A = ak.arange(5)
-    >>> truth = ak.sqrt([True, True, False, False, True])
-    >>> ak.sqrt(A, truth)
+    >>> a = ak.arange(5)
+    >>> ak.sqrt(a)
+    array([0 1 1.4142135623730951 1.7320508075688772 2])
+    >>> ak.sqrt(a, ak.sqrt([True, True, False, False, True]))
     array([0, 1, 2, 3, 2])
     """
     return power(pda, 0.5, where)
--- a/arkouda/pdarrayclass.py
+++ b/arkouda/pdarrayclass.py
@@ -2638,6 +2638,8 @@ def rotr(x, rot) -> pdarray:
     >>> A = ak.arange(10)
     >>> ak.rotr(1024 * A, A)
     array([0, 512, 512, 384, 256, 160, 96, 56, 32, 18])
+
+
     """
     if isinstance(x, pdarray) and x.dtype in [akint64, akuint64]:
         if (isinstance(rot, pdarray) and rot.dtype in [akint64, akuint64]) or isSupportedInt(rot):
@@ -2651,13 +2653,76 @@ def rotr(x, rot) -> pdarray:
 
 
 @typechecked
-def power(pda: pdarray, pwr: Union[int, float, pdarray]) -> pdarray:
-    return pda**pwr
+def power(pda: pdarray, pwr: Union[int, float, pdarray], where: Union[bool, pdarray] = True) -> pdarray:
+    """
+    Raises an array to a power. If where is given, the operation will only take place in the positions
+    where the where condition is True.
+
+    Notes:
+    This function was modified to deviate from the previous implementation that matched numpy.
+    If a False condition was given, numpy would return 0. This implementation will now return
+    the original value.
+
+    Parameters
+    ----------
+    pda : pdarray
+        A pdarry of values that will be raised to a power (pwr)
+    pwr : integer, float, or pdarray
+        The power(s) that pda is raised to
+    where : Boolean or pdaarray
+        This condition is broadcast over the input. At locations where the condition is True, the
+        corresponding value will be raised to the respective power. Elsewhere, it will retain its
+        original value. Default set to True.
+
+    Returns
+    -------
+        pdarray
+        Returns a pdarray of values raised to a power, under the boolean where condition.
+
+    Examples
+    --------
+    >>> ak.power(ak.arange(5), 3, truth)
+    array([0, 1, 8, 3, 64])
+    """
+    from arkouda.numeric import cast as akcast
+    from arkouda.numeric import where as akwhere
+
+    if where is True:
+        return pda**pwr
+    elif where is False:
+        return pda
+    else:
+        exp = pda**pwr
+        return akwhere(where, exp, akcast(pda, dtype(exp)))
 
 
 @typechecked
-def sqrt(pda: pdarray) -> pdarray:
-    return power(pda, 0.5)
+def sqrt(pda: pdarray, where: Union[bool, pdarray] = True) -> pdarray:
+    """
+    Takes the square root of array. If where is given, the operation will only take place in
+    the positions where the where condition is True.
+
+    Parameters
+    ----------
+    pda : pdarray
+        A pdarry of values that will be square rooted
+    where : Boolean or pdaarray
+        This condition is broadcast over the input. At locations where the condition is True, the
+        corresponding value will be square rooted. Elsewhere, it will retain its original value.
+        Default set to True.
+
+    Returns
+    -------
+        pdarray
+        Returns a pdarray of square rooted values, under the boolean where condition.
+
+    Examples:
+    >>> A = ak.arange(5)
+    >>> truth = ak.sqrt([True, True, False, False, True])
+    >>> ak.sqrt(A, truth)
+    array([0, 1, 2, 3, 2])
+    """
+    return power(pda, 0.5, where)
 
 
 @typechecked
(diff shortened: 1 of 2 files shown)


arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
 
 
 @typechecked
-def power(pda: pdarray, pwr: Union[int, float, pdarray]) -> pdarray:
-    return pda**pwr
+def power(pda: pdarray, pwr: Union[int, float, pdarray], where: Union[bool, pdarray] = True) -> pdarray:
+    """
+    Raises an input value or array to a power, can be subjected to a condition that if True will raise value to desiered
```
pierce314159, line 2627: - `power` only takes an array, so I think we should remove the `input value` from this description
- typo `desiered` -> `desired`

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+    power and if False will return the original value.
+
+    Notes:
+    This function was modified to more closely match the implementation of numpy's power function. Also fixes
+    an error to make a False condition return the original value, as expected.
```
pierce314159, line 2632: I wouldn't say it `fixes an error`, but rather deviates from numpys implementation of `where`

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
 
+    Parameters
+    ----------
+    pda : pdarray
+        A pdarry of values that will be raise to a power (pwr)
```
pierce314159, line 2637: `raise` -> `raised`

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+        This condition is broadcast over the input. At locations where the condition is True, the corresponding value
+        will be raise to the respective power. Elsewhere, it will retain its original value. Default set to True.
+
+    Returns
+    -------
+        pdarray
+        Returns a pdarray of values raised to a power, under the boolean where condition.
+
+    Examples
+    --------
```
pierce314159, line 2650: we should also include an example with the `where` argument used

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+    else:
+        exp = pda**pwr
+        return akwhere(where, exp, akcast(pda, dtype(exp)))
 
 @typechecked
-def sqrt(pda: pdarray) -> pdarray:
-    return power(pda, 0.5)
+def sqrt(pda: pdarray, where: Union[bool, pdarray] = True) -> pdarray:
+    """
+    Raises an input value or array to a power, can be subjected to a condition that if True will raise value to desired
```
pierce314159, line 2669: We should reword this to say something along the lines of: `Takes the square root of array. If where is given, the operation will only take place in the positions where where is True`

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+    power and if False will return the original value.
+
+    Notes:
+    This function is just a call to the power function, with a power of .5. See above for more documentation and examples.
```
pierce314159, line 2673: we don't really need the user to know that this is implemented just using the `power` method

tests/operator_tests.py
```diff
@@ -476,13 +467,79 @@ def test_pda_power(self):
...
 
+        # Test a singleton with and without a Boolean argument
+        n = np.array([7])
+        a = ak.array(n)
+
+        p_True = ak.power(a, 3, True)
+        p_Imply = ak.power(a, 3)
+        p_False = ak.power(a, 3, False)
+        self.assertListEqual(p_True.to_list(), p_Imply.to_list())
+        self.assertListEqual(p_False.to_list(), a.to_list())
```
pierce314159, line 478: since we don't use the variables again, it seems cleaner to me to put the calls in line
```python
self.assertListEqual(ak.power(a, 3, True).to_list(), ak.power(a, 3).to_list())
self.assertListEqual(ak.power(a, 3, False).to_list(), a.to_list())
```

tests/operator_tests.py
```diff
@@ -476,13 +467,79 @@ def test_pda_power(self):
...
+
+        # Test an with and without a Boolean argument, all the same
+        n = np.array([0, 0.0, 1, 7.0, 10])
+        a = ak.array(n)
+
+        truth_True = ak.array([True, True, True, True, True])
+        truth_False = ak.array([False, False, False, False, False])
```
pierce314159, line 485: same comment about not needing to declare a var, but I also wanted to highlight the `ak.ones` and `ak.zeros` functionality which i think will simplify this a bit

```python
>>> ak.ones(5, bool)
array([True True True True True])

>>> ak.zeros(5, bool)
array([False False False False False])
```

tests/operator_tests.py
```diff
@@ -476,13 +467,79 @@ def test_pda_power(self):
...
+
+        # Test a singleton with and without a Boolean argument, mixed
+        n = np.arange(10)
+        a = ak.array(n)
+        n_Calculated = np.array([0, 1, 4, 3, 16, 5, 36, 7, 64, 9])
+        a_Calculated = ak.array(n_Calculated)
+
+        truth = ak.array([True, False, True, False, True, False, True, False, True, False])
+        p = ak.power(a, 2, truth)
+        self.assertListEqual(p.to_list(), a_Calculated.to_list())
```
pierce314159, line 500: instead of hardcoding the arrays, i think we could test the same functionality by doing
```python
a = ak.arange(10)
self.assertListEqual([i if i%2 else i**2 for i in range(10)], ak.power(a, 2, a%2==0).to_list())
```

tests/operator_tests.py
```diff
@@ -476,13 +467,79 @@ def test_pda_power(self):
...
         n = np.array([-1.0, -3.0])
         a = ak.array(n)
 
         p = ak.power(a, 0.5)
         ex = np.power(n, 0.5)
         self.assertTrue(np.allclose(p.to_ndarray(), ex, equal_nan=True))
 
+        # Test invalid input, INF
+        I = np.Infinity
+        n = np.array([I, -I])
```
pierce314159, line 512: what makes `INF` invalid input? also I think instead of defining `I` we should use `np.inf`

tests/operator_tests.py
```diff
@@ -476,13 +467,79 @@ def test_pda_power(self):
...
+        a = ak.array(n)
+        n_Calculated = np.array([I, I])
+        a_Calculated = ak.array(n_Calculated)
+
+        p = ak.power(a, 2)
+        self.assertListEqual(p.to_list(), a_Calculated.to_list())
```
pierce314159, line 518: we should compare directly against the numpy array here instead of converting to a pdarray first to verify we're handling `inf` the same way they are

```python
n = np.array([np.inf, -np.inf])
a = ak.array([np.inf, -np.inf])
self.assertListEqual(np.power(n, 2).tolist(), ak.power(a, 2).to_list())
```

pierce314159 (commented):
A few things to look into

arkouda/pdarrayclass.py
Ethan-DeBandi99: I definitely agree with @pierce314159 here. That was intended `NumPy` behavior, but something that we want to operate differently.

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+    This function was modified to more closely match the implementation of numpy's power function. Also fixes
+    an error to make a False condition return the original value, as expected.
```
Ethan-DeBandi99, line 2675: Again, just highlighting that this is not actually an error. We should indicate that we deviate from the `NumPy` behavior here instead of stating that we are fixing an error.

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+
+    Parameters
+    ----------
+    pda : pdarray
+        A pdarry of values that will be raise to the 1/2 power
```
Ethan-DeBandi99, line 2680: Piggybacking off pierce a bit here, but I think it would be clearer to state something like `A pdarray of values that will be square rooted.` The reason being the user doesn't need to know underneath it just leverages the `power` functionality. I would also mention that this could cause some confusion if someone is in a rush.

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+    where : Boolean or pdaarray
+        This condition is broadcast over the input. At locations where the condition is True, the corresponding value
+        will be raise to the 1/2 power. Elsewhere, it will retain its original value. Default set to True.
```
Ethan-DeBandi99, line 2683: Again, I would change `raise to the 1/2 power` to `square rooted`.

arkouda/pdarrayclass.py
```diff
@@ -2622,13 +2622,78 @@ def rotr(x, rot) -> pdarray:
...
+
+    Returns
+    -------
+        pdarray
+        Returns a pdarray of values raised to a power, under the boolean where condition.
```
Ethan-DeBandi99, line 2688: Update the return here to indicate that this is the square root being returned.

tests/operator_tests.py
Ethan-DeBandi99: I agree with using `np.inf` here. That is generally what we do in other places.

Ethan-DeBandi99 (changes requested):
Just a few small things in addition to @pierce314159 comments.

arkouda/pdarrayclass.py
```diff
@@ -2638,6 +2638,8 @@ def rotr(x, rot) -> pdarray:
     >>> A = ak.arange(10)
     >>> ak.rotr(1024 * A, A)
     array([0, 512, 512, 384, 256, 160, 96, 56, 32, 18])
+
+
```
Ethan-DeBandi99, line 2642: Not really a big deal, but I don't think we need the extra lines here.

tests/operator_tests.py
```diff
@@ -476,13 +467,56 @@ def test_pda_power(self):
...
+    def test_pda_sqrt(self):
+        # Base cases and edge cases
+        # Most cases are taken care of in the test_pda_power tests
+        n = np.array([4, 16.0, -1, 0, np.inf])
+        a = ak.array(n)
+
+        s = ak.sqrt(a)
+        ex = np.sqrt(n)
+
+        self.assertTrue(np.allclose(s.to_ndarray(), ex, equal_nan=True))
```
pierce314159, line 514: I think we can save ourselves some variable declarations and do something like
```python
n = np.array([4, 16.0, -1, 0, np.inf])
a = ak.array(n)
self.assertTrue(np.allclose(ak.sqrt(a).to_ndarray(), np.sqrt(n), equal_nan=True))
```

pierce314159 (commented):
couple small things

pierce314159 (approved):
Looks good!!

PR #1872, opened 2022-11-01
Part of #1404: Add 10/03/22 sort scalability graphs from SGI 8600

ronawho:
Add some scalability results from an older SGI 8600 w/ EDR InfiniBand that show the impact from some recent sort optimizations. This includes results for various sort problem sizes up to 240 nodes to show the impact from scan optimizations, improving sort bucket exchange, and offsetting aggregator flushing.

Part of #1404


pierce314159 (approved):
Looks good! And the performance increase we've seen in argsort has been great!

PR #1873, opened 2022-11-01
Closes #1800 - Adding Symbol Table Overview documentation

### LINKED ISSUE #1800: Overview of SymbolTable and Messaging Documentation
Part of #1793 

This issue covers the creation of a .md file to give an overview of how to use the symbol table and our messaging between client and server.

joshmarshall1:
This PR closes #1800 

I've been having a hard time settling on a way to outline the available properties within the "Accessing Symbol Table Entry Data" section at the end. Any input there is highly appreciated. Also any additional information you think I should add for the structure section and the brief explanation of each entry type would be great.


training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+    - All properties of `AbstractSymEntry`
+    - `ndim`
+    - `size`
+    - `GroupBySymEntry`
+      - All properties of `CompositeSymEntry`
+      - `keyNamesEntry`
+      - `keyTypesEntry`
+      - `segmentsEntry`
+      - `permEntry`
+      - `ukIndEntry`
```
Ethan-DeBandi99, line 143: I would break this out instead of also having the entries tiered:

- AbstractSymEntry
    - Properties listed
- GenSymEntry
    - Properties listed
etc.

I do like how you reference the properties of the parent.

Ethan-DeBandi99 (changes requested):
Overall, I think the bulk of the information needed is there. I also think this is easy to follow. Most of the changes I am requesting are thinking of understanding for someone who has never been exposed to this type of structure and ensure we provide a clear picture.

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+1. [Symbol Table Overview](#overview)
+2. [Symbol Table Structure](#symStruct)
+3. [Creating a New Entry From an Array](#addSym)
+4. [Finding an Entry in the Symbol Table](#findSym)
+5. [Accessing Symbol Table Entry Data](#accessSym)
+
+<a id="overview"></a>
+## Symbol Table Overview
+
+The Symbol Table is Arkouda's "database" that tracks the components every Arkouda object is composed of. At a fundamental level it functions similar to a relational database, however the design and access methods are quite different.
```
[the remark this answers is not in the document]
pierce314159: I second this because I have no clue what a relational database and that would make me feel like it was a complex object that I couldn't understand. I think lookup table feels more approachable

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+
+- `AbstractSymEntry`
+  - `GenSymEntry`
+    - `SymEntry`
+    - `SegArraySymEntry`
+    - `SegStringSymEntry`
+  - `CompositeSymEntry`
+    - `GroupBySymEntry`
+  
+The root-level entry type is `AbstractSymEntry`. All other entry types inherit from `AbstractSymEntry` or one of its ancestors. `AbstractSymEntry` does not store any data, only metadata information, such as name and entry type.
```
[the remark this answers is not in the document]
pierce314159: I was thinking rewording it to `All other entry types inherit from AbstractSymEntry either directly or through one of its children`, but I'm not sure if that's too wordy

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+
+`SymEntry` is the entry type used for basic one-dimensional arrays of numeric values and their metadata. This is the most commonly used entry type.
+
+`SegArraySymEntry` tracks the component pieces of a `SegmentedArray` object. These components are all instances of `SymEntry` types.
+
+`SegStringSymEntry`tracks the component pieces of a `SegmentedStrings` object. The values of a `SegStringSymEntry` are stored in a `uint8` `SymEntry`. These components are all instances of `SymEntry` types.
+
+`CompositeSymEntry` is used to indicate that child `Entries` inheriting from this entry type contain multiple `data-adding` objects. This means that types under `CompositeSymEntry` contain multiple objects of varying types. For example, `SegArraySymEntry` contains three `SymEntry` entries, but all three are of a single type and relate to a single SegArray Object. `GroupBySymEntry` contains five `SymEntry` entries, but each `SymEntry` could be of a different type from the others, i.e. `keyNamesEntry` would be of type `str` while `permEntry` would be of type `int64`.
```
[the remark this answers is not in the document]
pierce314159: also they aren't all of the same type right? the `values` has a different type (can be anything) than `offsets` (has to be int)

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+You can then loop through the returned array of names to access the data and perform operations on each entry in the set.
+
+<a id="accessSym"></a>
+## Accessing Symbol Table Entry Data
+
+Depending on your functionality requirements, there are a couple different ways to access a symbol table entry.
+
+### Passing the Entry to the Client
+
+If you need to return an entry to the python for computations, the `SymTab.attrib(name)` method will return the entry's attributes, which are required by the arkouda client to reference the entry and use the underlying data. You can then use the return message to `attach` to the entry and access the data as you would with any other Python object.
```
pierce314159, line 93: `to the python` -> `to python`

training/SYMBOL_TABLE_OVERVIEW.md
pierce314159: I agree, the current way makes it hard to tell where the properties end and the child class begins once rendered
- `AbstractSymEntry`
  - `entryType`
  - `assignableTypes`
  - `name`
  - `GenSymEntry`
    - All properties of `AbstractSymEntry`
    - `dtype`
    - `itemsize`
    - `size`
    - `ndim`
    - `shape`
    - `SymEntry`
      - All properties of `GenSymEntry`

training/SYMBOL_TABLE_OVERVIEW.md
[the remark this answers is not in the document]
joshmarshall1: I reworded this to use `lookup table` and mention that it uses a Chapel `Map`

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+
+<a id="overview"></a>
+## Symbol Table Overview
+
+The Symbol Table is Arkouda's "database" that tracks the components every Arkouda object is composed of. At a fundamental level it functions similar to a relational database, however the design and access methods are quite different.
+
+<a id="symStruct"></a>
+## Symbol Table Structure
+
+The Symbol Table stores `Entries`. The Entry types are designed in a simple hierarchical structure as follows:
```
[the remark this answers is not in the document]
joshmarshall1: Updated to your suggestion

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+
+- `AbstractSymEntry`
+  - `GenSymEntry`
+    - `SymEntry`
+    - `SegArraySymEntry`
+    - `SegStringSymEntry`
+  - `CompositeSymEntry`
+    - `GroupBySymEntry`
```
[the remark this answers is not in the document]
joshmarshall1: I agree that the `enum` object isn't necessary for this document

training/SYMBOL_TABLE_OVERVIEW.md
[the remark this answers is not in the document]
joshmarshall1: Reworded this with a mix of both suggestions

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+  
+The root-level entry type is `AbstractSymEntry`. All other entry types inherit from `AbstractSymEntry` or one of its ancestors. `AbstractSymEntry` does not store any data, only metadata information, such as name and entry type.
+
+`GenSymEntry` is the first type directly inheriting from `AbstractSymEntry`. `GenSymEntry` is used to define the main metadata necessary for entry types that contain a single set of values of a single data type.
```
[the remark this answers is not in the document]
joshmarshall1: I reworded this based on your comment, hopefully it makes more sense now

training/SYMBOL_TABLE_OVERVIEW.md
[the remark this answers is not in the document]
joshmarshall1: `SegArraySymEntry` does contain three `SymEntry` values. It used to be two until `lengths` was added

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+
+`SegArraySymEntry` tracks the component pieces of a `SegmentedArray` object. These components are all instances of `SymEntry` types.
+
+`SegStringSymEntry`tracks the component pieces of a `SegmentedStrings` object. The values of a `SegStringSymEntry` are stored in a `uint8` `SymEntry`. These components are all instances of `SymEntry` types.
+
+`CompositeSymEntry` is used to indicate that child `Entries` inheriting from this entry type contain multiple `data-adding` objects. This means that types under `CompositeSymEntry` contain multiple objects of varying types. For example, `SegArraySymEntry` contains three `SymEntry` entries, but all three are of a single type and relate to a single SegArray Object. `GroupBySymEntry` contains five `SymEntry` entries, but each `SymEntry` could be of a different type from the others, i.e. `keyNamesEntry` would be of type `str` while `permEntry` would be of type `int64`.
+
+`GroupBySymEntry` is used to track component pieces of a `GroupBy` object. It accomplishes this task by tracking five separate `SymEntry` instances that each relate to one of the components.
```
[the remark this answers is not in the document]
joshmarshall1: Updated

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+
+<a id="addSym"></a>
+## Creating a New Entry From an Array
+
+If your Chapel module creates an array of data that you then want to either persist or return a reference to back to the Client, then you would need to add that array to the symbol table. This is pretty simple and only requires a single line of code to accomplish.
```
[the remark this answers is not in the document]
joshmarshall1: Added

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+There are two ways of finding if a given string matches the name of a Symbol Table entry:
+
+```chapel
+  st.contains(string); // Returns bool
+  st.findAll(regex string); // Returns [] string
+```
+
+`st.contains(string)` is used to verify the existence of an entry that matches the provided string. This function will return `True` if it exists within the Symbol Table and `False` if it does not.
+
+`st.findAll(regex string)` is used when you have a partial name that can map to multiple entries in a set. An example of when this is used is when `Categorical` objects are registered to the symbol table. These objects register 4 separate entries, one for each piece of the `Categorical` needed to rebuild it on the client.
```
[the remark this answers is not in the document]
joshmarshall1: I added a blurb for `checkTable` and highlighted that it is the most commonly used. I'm not sure how I missed this the first time around

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,150 @@
...
+```chapel
+  st.contains(string); // Returns bool
+  st.findAll(regex string); // Returns [] string
+```
+
+`st.contains(string)` is used to verify the existence of an entry that matches the provided string. This function will return `True` if it exists within the Symbol Table and `False` if it does not.
+
+`st.findAll(regex string)` is used when you have a partial name that can map to multiple entries in a set. An example of when this is used is when `Categorical` objects are registered to the symbol table. These objects register 4 separate entries, one for each piece of the `Categorical` needed to rebuild it on the client. 
+
+If you have a set of data with entries `dataset_entry_0`, `dataset_entry_1`, and `dataset_entry_2`, you could manually find each entry if you know how many entries are in your dataset. However, this function is most useful for when you may not know exactly how many entries you have. By passing the portion of the name these three entries have in common, `dataset_entry_`, followed by regex notation for "Any number, 1 or more times", `/d+`. We can then get a list of all the matching names in the symbol table and operate on all the dataset pieces.
```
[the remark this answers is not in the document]
joshmarshall1: We don't need to know the number of entries ahead of time. I reworded this so it's hopefully clearer

training/SYMBOL_TABLE_OVERVIEW.md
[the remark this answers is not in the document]
joshmarshall1: Updated to highlight this point

training/SYMBOL_TABLE_OVERVIEW.md
joshmarshall1: Reformatted to be on one level with spacing. I think it's clearer now, but you can be the judge

training/SYMBOL_TABLE_OVERVIEW.md
joshmarshall1: Fixed

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,169 @@
...
+1. [Symbol Table Overview](#overview)
+2. [Symbol Table Structure](#symStruct)
+3. [Creating a New Entry From an Array](#addSym)
+4. [Finding an Entry in the Symbol Table](#findSym)
+5. [Accessing Symbol Table Entry Data](#accessSym)
+
+<a id="overview"></a>
+## Symbol Table Overview
+
+The Symbol Table is Arkouda's "database" that tracks the components every Arkouda object is composed of. It serves as a `lookup table` that uses a Chapel `Map` which allows Arkouda to persist objects until the server is shutdown or the object is deleted.
```
Ethan-DeBandi99, line 18:
```suggestion
The Symbol Table is implemented using a Chapel Map and serves as a `lookup table` that uses a Chapel `Map` which allows Arkouda to persist objects until the server is shutdown or the object is deleted.

```

I think we need to totally remove any reference to a database

training/SYMBOL_TABLE_OVERVIEW.md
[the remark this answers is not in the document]
Ethan-DeBandi99: @joshmarshall1 - you are right... I was referencing the old code. My fault.

Ethan-DeBandi99 (commented):
Looks good. Just one final suggestion.

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,170 @@
...
+1. [Symbol Table Overview](#overview)
+2. [Symbol Table Structure](#symStruct)
+3. [Creating a New Entry From an Array](#addSym)
+4. [Finding an Entry in the Symbol Table](#findSym)
+5. [Accessing Symbol Table Entry Data](#accessSym)
+
+<a id="overview"></a>
+## Symbol Table Overview
+
+The Symbol Table is implemented using a Chapel Map and serves as a `lookup table` that uses a Chapel `Map` which allows Arkouda to persist objects until the server is shutdown or the object is deleted.
```
pierce314159, line 18:
```suggestion
The Symbol Table is implemented using a Chapel Map and serves as a `lookup table` which allows Arkouda to persist objects until the server is shutdown or the object is deleted.

```

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,170 @@
...
+
+`SymEntry` is the entry type used for basic one-dimensional arrays of numeric values and their metadata. This is the most commonly used entry type.
+
+`SegArraySymEntry` tracks the component pieces of a `SegmentedArray` object. These components are all instances of `SymEntry` types.
+
+`SegStringSymEntry`tracks the component pieces of a `SegmentedStrings` object. The values of a `SegStringSymEntry` are stored in a `uint8` `SymEntry`. These components are all instances of `SymEntry` types.
+
+`CompositeSymEntry` is used to indicate that child `Entries` inheriting from this entry type contain multiple `data-adding` objects. This means that types under `CompositeSymEntry` contain multiple objects of varying types, which may not all be needed to conceptualize the Object. For example, `GroupBySymEntry` contains five `SymEntry` entries. One of these, `keyNamesEntry` could contain `n` number of references to additional data values. These could be of multiple types, and it is possible to build a GroupBy object with all or only part of the stored values.
```
pierce314159, line 44: `This means that types under CompositeSymEntry contain multiple objects of varying types`

I get what this is saying but I feel like it could be confusing since types has two meanings here. Maybe we could reword it to something like

`This means that entry types under CompositeSymEntry contain multiple objects of varying types`

Also I feel like `Object` doesn't need to be capitalized here

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,170 @@
...
+
+More information on passing data between the client and server can be found [here](https://github.com/Bears-R-Us/arkouda/blob/master/training/MESSAGING_OVERVIEW.md).
+
+### Working with Entries in Chapel
+
+When working with a basic `SymEntry` entry type, you can easily access the data values within the entry by accessing the property `a`. To access the domain information, access the property `aD`. For example, assume we have a SymTab entry variable called `entry`.
+
+```chapel
+  var values = entry.a;
+  var domain = entry.aD;
```
pierce314159, line 111: we should name our `domain` variable since that is a reserved keyword in chapel

training/SYMBOL_TABLE_OVERVIEW.md
```diff
@@ -0,0 +1,170 @@
...
+
+If you need to return an entry to python, the `SymTab.attrib(name)` method will return the entry's metadata, which is required by the arkouda client to reference the entry when operations are sent to the server to be performed on the object. You can then use the return message to `attach` to the entry and view the data as you would with any other Python object.
+
+It is important to note that on the client after `attaching` to a Symbol Table Entry, the client does not store the data or perform computations on the data. The client is only storing the metadata and any computations on the object are performed on the server.
+
+More information on passing data between the client and server can be found [here](https://github.com/Bears-R-Us/arkouda/blob/master/training/MESSAGING_OVERVIEW.md).
+
+### Working with Entries in Chapel
+
+When working with a basic `SymEntry` entry type, you can easily access the data values within the entry by accessing the property `a`. To access the domain information, access the property `aD`. For example, assume we have a SymTab entry variable called `entry`.
```
pierce314159, line 107: maybe drop `information` here, since we're getting the domain itself rather than info about it

pierce314159 (commented):
few small things

training/SYMBOL_TABLE_OVERVIEW.md
Ethan-DeBandi99: I think @pierce314159 suggestion makes sense here to avoid repeating that a Chapel Map is used.

training/SYMBOL_TABLE_OVERVIEW.md
Ethan-DeBandi99: You could also say `To access the domain field` or `To access the domain object`. Not sure if this is necessary or if @pierce314159's suggestion to just remove information is sufficient.

training/SYMBOL_TABLE_OVERVIEW.md
joshmarshall1: Updated to use this suggestion

training/SYMBOL_TABLE_OVERVIEW.md
joshmarshall1: Updated

training/SYMBOL_TABLE_OVERVIEW.md
joshmarshall1: Renamed to `entryDomain`

training/SYMBOL_TABLE_OVERVIEW.md
joshmarshall1: Dropped `information`. I think this is sufficient to just drop information

Ethan-DeBandi99 (approved):

PR #1879, opened 2022-11-03
Add missing localize call to dsetname

bmcdonald3:


Ethan-DeBandi99 (approved):
Thanks for catching this. Looks good to go.

PR #1880, opened 2022-11-03
Swap order of zippered for-loops in which an unbounded range was the leader

bradcray:
In Chapel, we recently fixed an inconsistency between serial and parallel loops such that 'for (i,j) in zip(1.., whatever)' now results in a (conceptually) infinite execution because we make the first iterand determine the number of iterations of the loop, and subsequent iterands "follow" the first's lead.  Prior to this change, if the first iterand was an unbounded loop, it would just stop whenever the other iterators were complete.  This PR flips such cases around in the source to 'for (j,i) in zip(whatever, 1..)' to preserve the old behavior in the face of the new interpretation.

For further information and context, see https://github.com/chapel-lang/chapel/issues/20051 and https://github.com/chapel-lang/chapel/pull/20891

### DIFF
--- a/src/Cast.chpl
+++ b/src/Cast.chpl
@@ -64,7 +64,7 @@ module Cast {
     var values = st.addEntry(vname, totBytes, uint(8));
     ref va = values.a;
     forall (o, s) in zip(segments.a, strings) with (var agg = newDstAggregator(uint(8))) {
-      for (i, b) in zip(0.., s.bytes()) {
+      for (b, i) in zip(s.bytes(), 0..) {
         agg.copy(va[o+i], b);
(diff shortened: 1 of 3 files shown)


pierce314159 (approved):
Looks good!

If you rebase/merge with master, you should get ronawho's changes from this morning which should take care of the CI

Ethan-DeBandi99 (approved):
Looks good. Once you rebase with master and CI clears, this should be ready to go.

bradcray:
@pierce314159 / @Ethan-DeBandi99 : Thanks for the review!  Looks like it's good to go now.

PR #1883, opened 2022-11-04
Temporarily drop python 3.11 testing from CI

### LINKED ISSUE #1881: Drop python 3.11 from CI testing until pyarrow supports it
Python 3.11 was released recently and our CI just picked up it, but builds with it are failing because pyarrow doesn't have 3.11 wheels yet and the build from source is failing. 

See https://github.com/Bears-R-Us/arkouda/actions/runs/3389266626/jobs/5632176455 for a build failure and https://issues.apache.org/jira/browse/ARROW-17487 for pyarrow adding 3.11 wheels. Until that issue is closed, will disable 3.11 testing.

ronawho:
Python 3.11 was recently released and CI testing just picked it up with the 3.X (latest supported python 3) config, but pyarrow doesn't yet support 3.11 so the build was failing. Until pyarrow supports 3.11, just stop testing 3.11.

Note that when we're on top of version updates the latest python we specify is the same as 3.X, so in the steady state there is some testing duplication. However, this is useful when new versions come out since we (or at least I) don't track releases very closely and this alerts us to potential incompatibilities like this.

Closes #1881

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -63,7 +63,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.x']
+        python-version: ['3.8', '3.9', '3.10']
     container:
       image: chapel/chapel:1.28.0
     steps:


Ethan-DeBandi99 (approved):
Looks good. Thanks for catching this.

PR #1890, opened 2022-11-09
Closes #1889 - MessageArgs Parameter for CommandMap Functions

### LINKED ISSUE #1889: MessageArgs Parameter for CommandMap Functions
After discussing this with @bmcdonald3, we should be able to update our CommandMap functions to take a MessageArgs object in as a parameter. This would replace the `payload` and `size` parameters and prevent the need to call `parseMessageArgs` in every message processing function. 

In order to make this work, we need to define the memory management to be used on the object. Currently, the memory management scheme is not provided and thus being seen as generic. This is resulting in the typing information being unavailable. If we update the parameters for CommandMap functions to be `(cmd: string, msgArgs: shared MessageArgs, st: borrowed SymTab)`, it should work and achieve the outcome initially desired when the update was initially done. 

This update will require all CommandMap functions to be updated in a single PR.

Ethan-DeBandi99:
Closes #1889 

Updates CommandMap function to take `msgArgs: borrowed MessageArgs` as a parameter in place of `payload: string, argSize: int`. This prevents the need to call `var msgArgs = parseMessageArgs(payload, argSize)` in each function. 

Adds the necessary processing to pass `msgArgs` to each function.

### DIFF
--- a/src/Message.chpl
+++ b/src/Message.chpl
@@ -371,8 +371,8 @@ module Message {
     */
     proc parseMessageArgs(json_str: string, size: int) throws {
         var pArr = jsonToPdArray(json_str, size);
-        var param_list: list(ParameterObj) = new list(ParameterObj);
-        for (i, j_str) in zip(0..#size, pArr) {
+        var param_list = new list(ParameterObj, parSafe=true);
+        forall j_str in pArr with (ref param_list) {
             param_list.append(parseParameter(j_str));
         }
         return new owned MessageArgs(param_list);
--- a/src/Message.chpl
+++ b/src/Message.chpl
@@ -255,9 +255,15 @@ module Message {
     :size: int - number of parameters contained in list
     */
     class MessageArgs {
-        var param_list;
+        var param_list: list(ParameterObj);
         var size: int;
-        proc init(param_list: [?aD] ?t) {
+
+        proc init() {
+            this.param_list = new list(ParameterObj);
+            this.size = 0;
+        }
+
+        proc init(param_list: list(ParameterObj)) {
             this.param_list = param_list;
             this.size = param_list.size;
         }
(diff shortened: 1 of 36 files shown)


bmcdonald3 (approved):
Nice, glad this worked out. I didn't give a thorough review, but looks good to me. Nice we don't have to change anything on the client side.

joshmarshall1 (approved):
Looks good to me

src/Message.chpl
```diff
@@ -365,11 +371,11 @@ module Message {
     */
     proc parseMessageArgs(json_str: string, size: int) throws {
         var pArr = jsonToPdArray(json_str, size);
-        var paramArr: [0..#size] ParameterObj;
-        forall (i, j_str) in zip(0..#size, pArr) {
-            paramArr[i] = parseParameter(j_str);
+        var param_list: list(ParameterObj) = new list(ParameterObj);
+        for (i, j_str) in zip(0..#size, pArr) {
+            param_list.append(parseParameter(j_str));
```
pierce314159, line 376: the `i` is no longer used and (provided order of the `ParameterObj`s within `param_list` doesn't matter) i think we should be able to still do this in parallel
```chapel
var param_list = new list(ParameterObj, parSafe=true);
forall j_str in pArr with (ref param_list) {
  param_list.append(parseParameter(j_str));
}
```

pierce314159 (approved):
This all looks good to me!

 I left a note about a loop potentially being parallelized, but it should be small so I'm okay with us making an issue to look into that later. We should be sure to update the `MESSAGING_OVERVIEW.md` docs to reflect this change

I also wanted to note that it looks like the legacy HDF5 code still uses the old way, but those functions aren't being added to the `commandMap`, so I don't think it'll affect anything.

src/Message.chpl
Ethan-DeBandi99: I intentionally ignored the HDF5 Legacy code that is not in the command map. That code should be removed fully after the next release cycle.

I am going to test the parallel structure with gasnet and 2 locals, but that should work.

Ethan-DeBandi99: Updated to use parallel loop

PR #1892, opened 2022-11-09
Closes #1891 - IDNA Bug

### LINKED ISSUE #1891: `IDNA` Bug
Bugs reported by @21771 with the `idna` functionality. These were brought up after #1875 was resolved.

- Empty strings cause SIGTERM(15) for idna_encode() or idna_decode(). Should return empty strings since we can't control whether they exist or not.
- decode of ascii strings returns empty string instead of self. example ak.array(['example.com']).idna_decode() should return copy of original array.

Ethan-DeBandi99:
Closes #1891 

Updates processing to check if a string is empty before running the Python scripts. If it is, empty string is just returned.

Updates decode processing to check if a value is `idna` formatted before decode. If it is and the value was not decoded, empty string is returned per the original requirements. If the string was not `idna` then the original string is returned.

Added testing for the cases provided in the bug report to ensure proper functionality moving forward.

### DIFF
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -540,24 +540,28 @@ module SegmentedString {
       var encodeArr: [0..#this.size] string; 
       var encodeOffsets: [this.offsets.aD] int;
       var encodeLengths: [this.offsets.aD] int;
-
       const lengths = this.getLengths();
       forall (i, off, len) in zip(0..#this.size, offs, lengths) {
-        // var str_entry: string;
         const filename: string = basePath+"/src/exec/%i_tmp.txt".format(i);
         var str_entry: string = interpretAsString(origVals, off..#len);
-        // use subprocessing to make a call to a python file for the encoding
-        var sub = spawn(["python3", procFile, "-v", str_entry, "-f", filename]);
-        sub.wait();
+        // only run the encoding if the string value is not empty string to avoid segfaults
+        if str_entry != "" {
+          // use subprocessing to make a call to a python file for the encoding
+          var sub = spawn(["python3", procFile, "-v", str_entry, "-f", filename]);
+          sub.wait();
 
-        // read file python wrote to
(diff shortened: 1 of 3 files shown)


src/SegmentedString.chpl
```diff
@@ -540,24 +540,28 @@ module SegmentedString {
       var encodeArr: [0..#this.size] string; 
       var encodeOffsets: [this.offsets.aD] int;
       var encodeLengths: [this.offsets.aD] int;
-
       const lengths = this.getLengths();
       forall (i, off, len) in zip(0..#this.size, offs, lengths) {
-        // var str_entry: string;
         const filename: string = basePath+"/src/exec/%i_tmp.txt".format(i);
```
joshmarshall1, line 545: Not a big deal, but since the filename is only used in the event `str_entry` isn't empty, we could move this into the `if` block

joshmarshall1 (approved):

tests/string_test.py
```diff
@@ -644,11 +644,17 @@ def test_encoding(self):
...
+        ak_e = ak.array(e)
+        result = ak_e.idna_encode()
+        self.assertListEqual([i.encode("idna").decode("ascii") for i in e], result.to_list())
+
+        a2 = ['xn--mnchen-3ya', 'xn--zrich-kva', ' xn--zrich-boguscode', 'xn--!!', 'example.com']
         s2 = ak.array(a2)
         result = s2.idna_decode()
         # using the below assertion due to a bug in `Strings.to_ndarray`. See issue #1828
-        self.assertListEqual(["münchen", "zürich", "", ""], result.to_list())
+        self.assertListEqual(["münchen", "zürich", "", "", "example.com"], result.to_list())
```
pierce314159, line 657: I don't know much about encoding, but should we be thinking about the situation where we get an ascii string that happens to start with `xn--`? I think either way this covers everything our users have asked for so i think it's good

pierce314159 (approved):

PR #1893, opened 2022-11-09
Closes #1722 - `unregister_by_name` added to SegArray

### LINKED ISSUE #1722: Add `unregister_by_name` to `SegArray`
Add `unregister_by_name` functionality to `SegArray`

### LINKED ISSUE #1796: `SegArray.unregister()` does not remove all components of an attached SegArray from registry
When registering a SegArray on the server, calling `.unregister()` on the original will remove all components from the symbol table. However, attaching to a registered SegArray then calling `.unregister()` on the attached version will only remove the `lengths` component from the symbol table, leaving `values` and `segments` registered.

joshmarshall1:
This PR Closes #1722 

Due to the way registration is done within segarray, the best solution for the current state is to implement the `unregister_by_name` functionality and call that instead of calling `segarray.unregister()`. Also due to the registration of segarray, an additional property `serverName` was added to maintain the connection between the `segarray` object and the server-generated name used to access it.

Issue #1894 was created to address this

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -241,7 +241,7 @@ class SegArray:
         values = create_pdarray(parts[2])
         lengths = create_pdarray(parts[3])
 
-        return cls.from_parts(segments, values, lengths=lengths)\
+        return cls.from_parts(segments, values, lengths=lengths)
 
     @classmethod
     def from_multi_array(cls, m):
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99:
Spoke with @joshmarshall1 about this some. We are thinking that the updates to SegArray register should be minimal and is the correct way to resolve the issue instead of adding a work around. This PR will be updated to ONLY cover `unregister_by_name`

joshmarshall1:
With the updates being made for #1894, this functionality will be immediately changed. The only overlap will be the test case. I will be opening a new PR with the combined functionality.

PR #1897, opened 2022-11-14
Closes #1904: Add Chapel-native encoding/decoding functionality

### LINKED ISSUE #1904: Add Chapel-native encoding
Today, the encoding routine calls out to Python for the encoding and the performance could be improved by switching to a Chapel-native approach.

bmcdonald3:
This PR adds support for the iconv and idn2 libraries, which enable Chapel-native support for encoding and decoding strings from IDNA and a number of other encodings. This results in a significant performance improvement (~1000x) over the current implementation that calls out to Python to handle the encoding.

The additional idn2 and iconv libraries are now required Arkouda dependencies and can be installed with `make install-iconv` and `make install-idn2`.

Performance numbers collected on a 16-node-cs-hdr using 1,000,000 strings using strings normally distributed between lengths 1 to 16:

Locales | Original | Native Chapel | Optimized Native
------- | -------- | ------------- | ----------------
1       |  ~5,000s |            6s |           0.12s
2       | ~10,000s |          110s |           0.21s
4       |  --      |            -- |           0.24s
8       |  --      |            -- |           0.26s

Where "Optimized Native" is the current version of this PR and "Native Chapel" was the performance of inserting Chapel-native calls into the code from master prior to this PR.

closes #1904


Ethan-DeBandi99 (commented):
One thing to think about, not sure how big of a deal this would be. I am not sure we want to default `toEncoding` in `encode` or `fromEncoding` in `decode`. My thought is that since we are supporting more than just `idna`, we should force input. I do agree with the other values defaulting to `UTF-8`. 

I think `idna` was the initial use case, but we should probably just require the user to provide their desired encoding. 
@pierce314159 - do you have any thoughts on this?

src/Codecs.chpl
```diff
@@ -0,0 +1,110 @@
...
+        return ret;
+      }
+      var cRes: c_string;
+      var rc = idn2_to_ascii_lz(obj:c_string, cRes, 0);
+      if (rc != IDNA_SUCCESS) {
+        throw new Error("Encode failed");
+      }
+      var tmp = cRes: c_ptr(uint(8));
+      var ret: [0..#outBufSize] uint(8);
+      for i in ret.domain do ret[i] = (tmp+i).deref();
```
pierce314159, line 21: is there a reason this is a `for` loop instead of a `forall`? each loop iteration seems independent of the others to me, but I've never used `deref()` before

src/Codecs.chpl
```diff
@@ -0,0 +1,110 @@
...
+      idn2_free(cRes: c_void_ptr);
+      return ret;
+    } else if fromEncoding == "IDNA" {
+      if outBufSize == 1 {
+        var ret: [0..0] uint(8) = 0x00;
+        return ret;
+      }
+      var cRes: c_string;
+      var rc = idn2_to_unicode_8z8z(obj:c_string, cRes, 0);
```
pierce314159, line 30: It looks like if `fromEncoding` is `"IDNA"`, we convert to UTF-8 regardless of what `toEncoding` is. Is that the right thing to do if the `toEncoding` is something other than `"UTF-8"`? Similarly the situation where `toEncoding` is `"IDNA"`, we don't check what `fromEncoding` is set to.

It's fine if we don't support idna encode/decode with anything other than UTF-8, but I feel like we should check what encoding is passed and error out when not UTF-8.

src/EncodingMsg.chpl
```diff
@@ -9,38 +9,61 @@ module EncodingMsg {
...
+
+      // TODO: We only need one of these, they have the same info
+      var encodeLengths: [stringsObj.offsets.a.domain] int;
+      var encodeOffsets: [stringsObj.offsets.a.domain] int;
+      
+      // First get lengths
+      forall (i, off, len) in zip(0..#stringsObj.size, offs, lengths) {
+        var slice = new lowLevelLocalizingSlice(origVals, off..#len);
+        encodeLengths[i] = getBufLength(slice.ptr: c_ptr(uint(8)), len, toEncoding, fromEncoding);
+      }
```
pierce314159, line 52: if we try to optimize more, we might be able to take advantage of the `computeOnSegments` for this loop. I'm not sure if it will be more or less efficient but might be worth looking into

src/iconv.chpl
```diff
@@ -0,0 +1,382 @@
...
+
+extern type iconv_wide_char_hook = c_fn_ptr;
+
+extern type libiconv_t = c_void_ptr;
+
+extern type mbstate_t = c_int;
+
+// c2chapel thinks these typedefs are from the fake headers:
+/*
+extern type FILE = c_int;
```
pierce314159, line 70: why do we need all these commented out `extern`s here and in `idna.chpl`?

pierce314159 (commented):
A couple of inline comments. Also I want to double check if we should go update these other places as well to track the new dependencies

dependency locations - `arkouda-env-dev.yaml`, `arkouda-env.yaml`, `setup.py`, `requirements.txt`
documentation - `REQUIREMENTS.md`, `BUILD.md`, `ENVIRONMENT.md`, `prerequisites.rst`

src/iconv.chpl
bmcdonald3: Fragment of using c2chapel, this code was auto-generated. I could remove if you'd like, but ya, it was just auto-generated

src/EncodingMsg.chpl
bmcdonald3: Ya, I have some optimizations that are in the works and getting rid of this `lowLevelLocalizing` slice is the first one on the list. Thanks for pointing out `computeOnSegments`, that is not something I'm familiar with and I'll investigate.

src/Codecs.chpl
bmcdonald3: Ya, this is something that I had thought about as well. I think that in the short-term, maybe we should error out, like you said, but we also could make it a two-pass implementation where we convert from idna -> utf-8 -> requested encoding, but that could already be achieved through a slightly less convenient client-side interaction.

Does that sound good to just error for this initial effort?

src/Codecs.chpl
bmcdonald3: No, I think that can probably be a forall; good catch.

bmcdonald3:
@Ethan-DeBandi99 that sounds good to me WRT defaulting to UTF-8 for encode fromEncoding and then forcing input for toEncoding.

src/iconv.chpl
pierce314159: Ah okay! That makes sense. I'm fine with either way. I was just making sure I wasn't missing anything and chapel wasn't somehow looking into the comments for information

src/Codecs.chpl
pierce314159: oh yeah the two pass is clever! I'm happy with the error for the initial impl

bmcdonald3:
Found a bug in distributed settings, marking as draft until resolved.

pierce314159 (commented):
Other than a questions about the correctness-only benchmark, i think this is good to go!

benchmarks/encode.py
```diff
@@ -0,0 +1,108 @@
...
+        print("Average {} decode rate = {:.4f} GiB/sec".format(encoding, bytes_per_sec / 2**30))
+
+        
+def check_correctness(encoding, seed):
+    N = 10
+    a = ak.random_strings_uniform(1, 16, N, seed=seed)
+
+    # Do round trip encode/decode
+    encoded = a.encode(encoding)
+    decoded = a.decode(encoding)
```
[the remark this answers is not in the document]
bmcdonald3: Good catch; fixed.

pierce314159 (approved):

Ethan-DeBandi99 (approved):

PR #1900, opened 2022-11-14
Closes #1796, #1722, and #1894 - Update `SegArray` to only register a single object

### LINKED ISSUE #1722: Add `unregister_by_name` to `SegArray`

### LINKED ISSUE #1796: `SegArray.unregister()` does not remove all components of an attached SegArray from registry

### LINKED ISSUE #1894: Update `SegArray` to register as a single object
This was found while working #1796 

Due to the current implementation of `segarray.register()`, each component is being registered individually. I believe this is causing the name of the components to be updated on the server, but not updating the name of the server-side `segarray` object, causing a desync in the object/component relationship. We should update this now that there's a server-side segarray object to register as a `segarray` instead of registering each component individually.

joshmarshall1:
This PR Closes #1722 
This PR Closes #1796 
This PR Closes #1894 

With the addition of the `SegArray` server object, we are now able to register `SegArray`s as single Symbol Table entries instead of having to register each segarray component individually. 

This PR addresses the single-object registration and updates the functionallity that touched segarray register/attach. This includes `SegArray` unregister_by_name, the generic `attach` method, and `DataFrame.attach` among others.


src/RegistrationMsg.chpl
```diff
@@ -542,7 +542,7 @@ module RegistrationMsg
         }
 
         // type possibilities for pdarray and strings
-        var simpleTypes: list(string) = ["pdarray","int64", "uint8", "uint64", "float64", "bool", "strings", "string", "str"];
+        var simpleTypes: list(string) = ["pdarray","int64", "uint8", "uint64", "float64", "bool", "strings", "string", "str", "segarray"];
```
Ethan-DeBandi99, line 545: Since we have used this list in multiple places in the code, make it a Global so we don't redefine it multiple times.

src/Message.chpl
```diff
@@ -344,6 +344,14 @@ module Message {
             }
             return val_list;
         }
+
+        /*
+        Return bool if param_list contains given key name
+        */
+        proc contains(key: string): bool {
```
Ethan-DeBandi99, line 351: Noting for myself that it may be worth updating all of this functionality to use `List` objects instead of arrays since we moved to that for the main args.

arkouda/util.py
```diff
@@ -303,12 +302,15 @@ def attach(name: str, dtype: str = "infer"):
 
         return DataFrame.from_return_msg(repMsg)
     else:
-        dtype = repMsg.split()[2]
-
-        if dtype == "str":
-            return Strings.from_return_msg(repMsg)
-        else:
-            return create_pdarray(repMsg)
+        try:
```
Ethan-DeBandi99, line 305: I really want there to be a better way to handle adding `SegArray` here than using a try except. Especially because in the majority of cases it will just be a `pdarray` and will hit that failure case.

I think it may be worth still adding the response type of to each of the returns because it would eliminate the `try/except` here.

tests/segarray_test.py
```diff
@@ -605,12 +605,25 @@ def test_register_attach(self):
...
-        segarr.values.unregister()
-        with self.assertWarns(UserWarning):
-            segarr.is_registered()
-        # Unregister all components
         segarr.unregister()
         self.assertFalse(segarr.is_registered())
 
         self.assertEqual(len(ak.list_registry()), 0)
+
+    def test_unregister_by_name(self):
```
Ethan-DeBandi99, line 613: Do we test all the register/attach stuff in the individual object tests? I am wondering if we should centralize this at some point to test all register/attach features in one place?

Ethan-DeBandi99 (changes requested):
Nothing major, just a few small things to think about. A few things are out of scope here, but we should add issues for them so we can address later.

arkouda/dataframe.py
```diff
@@ -2045,14 +2050,11 @@ def attach(user_defined_name: str) -> DataFrame:
...
-                dtype = cols_resp.split()[2]
-                if dtype == Strings.objtype:
-                    columns[colName] = Strings.from_return_msg(cols_resp)
-                else:  # pdarray
-                    columns[colName] = create_pdarray(cols_resp)
+            colName = DataFrame._parse_col_name(name, user_defined_name)[0]
+            if f"_{Strings.objtype}_" in name:
+                columns[colName] = Strings.attach(name)
+            elif f"_{pdarray.objtype}_" in name:
+                columns[colName] = attach_pdarray(name)
```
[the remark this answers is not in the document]
joshmarshall1: We probably should add the `pdarray.attach` method and have it alias the existing `attach_pdarray` function so we don't break anything

arkouda/dataframe.py
```diff
@@ -2121,10 +2123,7 @@ def unregister_dataframe_by_name(user_defined_name: str) -> None:
             elif f"_{Categorical.objtype}_" in name:
                 Categorical.unregister_categorical_by_name(name)
             elif f"_{SegArray.objtype}_" in name:
-                # SegArray does not have an unregister_by_name function. By attaching and using
-                # unregister() it prevents unregistering the components individually here
-                segarray = SegArray.attach(name)
-                segarray.unregister()
+                SegArray.unregister_segarray_by_name(name)
```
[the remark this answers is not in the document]
joshmarshall1: Updated the comment

arkouda/dataframe.py
```diff
@@ -2199,7 +2200,7 @@ def from_return_msg(repMsg):
                 colName, colType = DataFrame._parse_col_name(parts[i], dfName)
                 if colType == "pdarray":
                     cols[colName] = create_pdarray(parts[i])
-                else:
+                elif colType == "str":
```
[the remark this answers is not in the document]
joshmarshall1: Added error handling

src/RegistrationMsg.chpl
```diff
@@ -90,10 +97,38 @@ module RegistrationMsg
...
+                    }
+                    when (DType.Bool) {
+                        var seg = getSegArray(name, st, bool);
+                        seg.fillReturnMap(rtnmap, st);
+                    }
+                }
+
+                repMsg = "%jt".format(rtnmap);
+            }
+            else {
```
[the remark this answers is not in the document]
joshmarshall1: Did some checking and error handling here

src/RegistrationMsg.chpl
```diff
@@ -402,20 +395,23 @@ module RegistrationMsg
...
+            } else {
+                // pdarray or Strings
+                objtype = "simple";
+            }
         } else if st.contains("%s.categories".format(name)) && st.contains("%s.codes".format(name)) {
-            dtype = "categorical";
-        } else if st.contains("%s_segments".format(name)) && st.contains("%s_values".format(name)) {
-            // Important to note that categorical has a .segments while segarray uses _segments
-            dtype = "segarray";
+            objtype = "categorical";
```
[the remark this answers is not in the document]
joshmarshall1: I agree - I'll add a ticket to do this

src/RegistrationMsg.chpl
joshmarshall1: Done

arkouda/util.py
joshmarshall1: Added your recommendation, removed `try/except`

tests/segarray_test.py
joshmarshall1: Moved this into `registration_tests.py`

src/Message.chpl
joshmarshall1: Noting for you #1903

Ethan-DeBandi99: Thanks!

arkouda/dataframe.py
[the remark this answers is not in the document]
Ethan-DeBandi99: Are we going to handle that in this PR or add an issue to take care of it?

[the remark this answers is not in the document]
joshmarshall1: I did it in this one since it was super quick. Forgot to mention that change in a reply here

pierce314159 (approved):
looks good

arkouda/dataframe.py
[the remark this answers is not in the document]
joshmarshall1: Updated

arkouda/dataframe.py
```diff
@@ -2045,14 +2050,11 @@ def attach(user_defined_name: str) -> DataFrame:
...
-                dtype = cols_resp.split()[2]
-                if dtype == Strings.objtype:
-                    columns[colName] = Strings.from_return_msg(cols_resp)
-                else:  # pdarray
-                    columns[colName] = create_pdarray(cols_resp)
+            colName = DataFrame._parse_col_name(name, user_defined_name)[0]
+            if f"_{Strings.objtype}_" in name:
+                columns[colName] = Strings.attach(name)
+            elif f"_{pdarray.objtype}_" in name:
+                columns[colName] = attach(name)
```
Ethan-DeBandi99, line 2057: I am a little worried that people may confuse this for a generic method. I see why you are only calling attach, but I do think it'd be helpful to do something like 
```
from arkouda.pdarrayclass import attach as pd_attach
```
to indicate this is specific to pdarray. Not really a sticking point, but figured I would mention.

Ethan-DeBandi99 (commented):
Just one thing I wanted to check to potentially avoid confusion, but otherwise looks good.

arkouda/dataframe.py
joshmarshall1: Updated

Ethan-DeBandi99 (approved):

PR #1907, opened 2022-11-16
Closes #1906 - HDF5 Assume `pdarray`/`Strings` Object when no ObjType Attribute

### LINKED ISSUE #1906: HDF5 Assume `pdarray`/`Strings` Object when ObjectType not Set
After speaking with some users, they are using some datasets that they may not have control of the format of the HDF5 file. This presents some issues moving forward if more complex objects are presented, but for the time being we can assume the data is a single Array. Thus, we can assume it is a pdarray or Strings object based on the datatype.

The HDF5 processing needs to be further updated to accomodate this while decisions are made on a path forward.
- If the `ObjType` attribute is not set, we should assume `pdarray`.
- Update the `pdarray` processing case to add `uint(8)` handling.

These updates should also allow us to remove the `legacyHDF5` flags as the old formats will be natively supported in the codebase. This means we can also remove the legacy processing code

Ethan-DeBandi99:
Closes #1906

Updates the processing when reading HDF5 files to assume the object is `pdarray` or `Strings` when the `ObjType` attribute is not present. This allows for reading of HDF5 in the old format. 

Removes the `HDF5_Legacy` module as it is no longer required to read these files.

Updates client `load`/`read` to remove the `legacyHDF5` parameter since it is no longer used to read legacy files.


pierce314159 (approved):
Logic looks good! We should also update references to `legacyHDF5` in `FILEIO.md`, but other than that i think it's good to go

Ethan-DeBandi99:
> Logic looks good! We should also update references to `legacyHDF5` in `FILEIO.md`, but other than that i think it's good to go

Updated `FILEIO.md`

joshmarshall1 (approved):
All looks good to me

jaketrookman (commented):
looks good to me, maybe just add a new comment to the FILEIO.md

Issue #1901, opened 2022-11-15
Reduce SymEntry creation overheads


ronawho:
`SymEntry` is Arkouda's class-wrapper around chapel arrays. While investigating https://github.com/Bears-R-Us/arkouda/issues/1398 we found that there are some additional overheads with creating a SymEntry over just the array creation. We've explored these overheads a little upstream in https://github.com/chapel-lang/chapel/issues/20164, and believe that we can eliminate most of them by only storing the chapel array instead of storing the domain and array and by allowing the SymEntry initializer to steal arrays that are passed in.

Today SymEntry looks like something like:

```chpl
class SymEntry {
  type etype;
  const size: int;
  const aD: makeDistDom(size).type;
  var a: [aD] etype;

  proc init(size: int, type etype) {
    this.etype = etype;
    this.size = size;
    this.aD = makeDistDom(size);
    // this.a uses default initialization
  }

  proc init(a: [?D] ?etype) {
    this.etype = etype;
    this.size = a.size;
    this.aD = D;
    this.a = a;
  }
}
```

Where storing both a domain and array results in two domains being created (one to store and one actually backing the array) and when creating a SymEntry from an existing array it results in creating a new array and copying. I think both issues can be improved with something like:

```chpl
class SymEntry {
  type etype;
  const size: int;
  var a = makeDistArray(size, etype);

  proc init(size: int, type etype) {
    this.etype = etype;
    this.size = size;
    this.a = makeDistArray(size, etype);
  }

  proc init(in a: [?D] ?etype) {
    this.etype = etype;
    this.size = a.size;
    this.a = a;
  }
}
```

Where by not storing `aD` we can avoid the extra domain creation. This does mean we can no longer do `symEntry.aD` so I'm proposing we update callsites to `symEntry.a.domain` (https://github.com/chapel-lang/chapel/issues/20389 prevents creating an `aD` proc wrapper). And passing `a` by `in` intent allows the compiler to steal the array if it's not used after the callsite which avoids an array allocation and copy when creating a `SymEntry` from an existing array.

ronawho:
The overhead really comes from the way things are declared:

```chpl
const aD: makeDistDom(size).type;

proc init(size: int, type etype) {
  this.aD = makeDistDom(size);
}
```

actually results in creating 2 domains. We construct one just to figure out the .type, and then we actually create the one that gets assigned to `this.aD`. Another way to write this would be:

```chpl
const aD = makeDistDom(size);

[...]

PR #1908, opened 2022-11-16
Part of #1901: Replace `SymEntry.Ad` with `SymEntry.a.domain`

ronawho:
Motivated by #1901, I'd like to stop storing the domain in the `SymEntry` wrapper, so we now need to go through the array to get the domain (e.g. `symEntry.aD` is now `symEntry.a.domain`). You could imagine creating a wrapper to do this automatically with a wrapper like:

```chpl
proc aD const ref {
  return a.domain;
}
```

But unfortunately chapel-lang/chapel#20389 prevents using `const ref`, which is required to avoid a full domain copy on each call.

### DIFF
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -143,7 +143,7 @@ module SegmentedString {
 
     /* Retrieve one string from the array */
     proc this(idx: ?t): string throws where t == int || t == uint {
-      if (idx < offsets.aD.low) || (idx > offsets.aD.high) {
+      if (idx < offsets.a.domain.low) || (idx > offsets.a.domain.high) {
         throw new owned OutOfBoundsError();
       }
       // Start index of the string
@@ -164,7 +164,7 @@ module SegmentedString {
        Chapel range, i.e. low..high by stride, not a Python slice.
        Returns arrays for the segment offsets and bytes of the slice.*/
     proc this(const slice: range(stridable=false)) throws {
-      if (slice.low < offsets.aD.low) || (slice.high > offsets.aD.high) {
+      if (slice.low < offsets.a.domain.low) || (slice.high > offsets.a.domain.high) {
           ssLogger.error(getModuleName(),getRoutineName(),getLineNumber(),
           "Array is out of bounds");
           throw new owned OutOfBoundsError();
@@ -177,9 +177,9 @@ module SegmentedString {
       var start = offsets.a[slice.low];
       // End of bytearray slice
       var end: int;
-      if (slice.high == offsets.aD.high) {
+      if (slice.high == offsets.a.domain.high) {
         // if slice includes the last string, go to the end of values
-        end = values.aD.high;
+        end = values.a.domain.high;
       } else {
         end = offsets.a[slice.high+1] - 1;
       }
@@ -223,7 +223,7 @@ module SegmentedString {
                                               "Computing lengths and offsets");
       var t1 = getCurrentTime();
       ref oa = offsets.a;
-      const low = offsets.aD.low, high = offsets.aD.high;
+      const low = offsets.a.domain.low, high = offsets.a.domain.high;
       // Gather the right and left boundaries of the indexed strings
       // NOTE: cannot compute lengths inside forall because agg.copy will
       // experience race condition with loop-private variable
@@ -301,7 +301,7 @@ module SegmentedString {
     /* Logical indexing (compress) of strings. */
     proc this(iv: [?D] bool) throws {
       // Index vector must be same domain as array
-      if (D != offsets.aD) {
+      if (D != offsets.a.domain) {
           ssLogger.info(getModuleName(),getRoutineName(),getLineNumber(),
                                                            "Array out of bounds");
           throw new owned OutOfBoundsError();
@@ -310,7 +310,7 @@ module SegmentedString {
                                                  "Computing lengths and offsets");
       var t1 = getCurrentTime();
       ref oa = offsets.a;
-      const low = offsets.aD.low, high = offsets.aD.high;
+      const low = offsets.a.domain.low, high = offsets.a.domain.high;
       // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit
       overMemLimit(numBytes(int) * iv.size);
       // Calculate the destination indices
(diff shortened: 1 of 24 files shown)


Ethan-DeBandi99:
I would like to get clarification on one thing before approving. It looks like your changes address calling `someEntry.aD` being updated to `someEntry.a.domain`. However, we are still storing `someEntry.aD`. Are you intending to update `SymEntry` in a future PR to remove `.aD`? I think we may want to add that change to this PR to avoid people writing code that still call `someEntry.aD` That will then need to be updated in the future.

Also, it looks like there are merge conflicts. I believe these are because I made an update the deleted `src/HDF5Msg_LEGACY.chpl`.

Ethan-DeBandi99 (commented):
I would like to get clarification on one thing before approving. It looks like your changes address calling `someEntry.aD` being updated to `someEntry.a.domain`. However, we are still storing `someEntry.aD`. Are you intending to update `SymEntry` in a future PR to remove `.aD`? I think we may want to add that change to this PR to avoid people writing code that still call `someEntry.aD` That will then need to be updated in the future.

Also, it looks like there are merge conflicts. I believe these are because I made an update the deleted `src/HDF5Msg_LEGACY.chpl`.

ronawho:
> I would like to get clarification on one thing before approving. It looks like your changes address calling someEntry.aD being updated to someEntry.a.domain. However, we are still storing someEntry.aD. Are you intending to update SymEntry in a future PR to remove .aD

Yeah, I have a branch ready to go that removes it (wanted to separate name change from performance chang to make sure there's no unintended perf changes from this.) I can open that PR now and rebase after this is merged, or I could rename `aD` to `_aD` in this PR to avoid uses getting added between the PRs.

> Also, it looks like there are merge conflicts. I believe these are because I made an update the deleted src/HDF5Msg_LEGACY.chpl.

Rebased

ronawho:
> I think we may want to add that change to this PR to avoid people writing code that still call someEntry.aD That will then need to be updated in the future.

Devs are probably in the habit of calling `SymEntry.aD` so to help with the transition, we could add an overload that gives a clean error message. Something like:

```chpl
proc aD {
  compilerError("SymEntry.aD has been removed, use SymEntry.a.domain instead");
}
```

> @ronawho - I am fine with separating things and leaving this as is. I just wanted to make sure that was the intended workflow. Thanks!

Sounds good, if you do want me to rename, the diff is pretty small:

```diff
diff --git a/src/MultiTypeSymEntry.chpl b/src/MultiTypeSymEntry.chpl
index 0e869e78..49f387d2 100644
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -197,11 +197,12 @@ module MultiTypeSymEntry
         type etype;
 
         /*
-        'aD' is the distributed domain for 'a' whose value and type
+        '_aD' is the distributed domain for 'a' whose value and type
         are defined by makeDistDom() to support varying distributions
         */
-        const aD: makeDistDom(size).type;
-        var a: [aD] etype;
+        const _aD: makeDistDom(size).type;
+        proc aD { compilerError("SymEntry.aD has been removed, use SymEntry.a.domain instead"); }
+        var a: [_aD] etype;
         
         /*
         This init takes length and element type
@@ -218,7 +219,7 @@ module MultiTypeSymEntry
             assignableTypes.add(this.entryType);
 
             this.etype = etype;
-            this.aD = makeDistDom(len);
+            this._aD = makeDistDom(len);
             // this.a uses default initialization
         }
 
@@ -233,7 +234,7 @@ module MultiTypeSymEntry
             assignableTypes.add(this.entryType);
 
             this.etype = etype;
-            this.aD = D;
+            this._aD = D;
             this.a = a;
         }
 
```

src/SegmentedString.chpl
```diff
@@ -1301,7 +1301,7 @@ module SegmentedString {
       const rh = rss.siphash();
       return if polarity then (lh == rh) else (lh != rh);
     }
-    ref oD = lss.offsets.aD;
+    ref oD = lss.offsets.a.domain;
```
pierce314159, line 1304: should we make this a `const ref` as well?

pierce314159 (commented):
looks good to me! Just one question about using a `ref` for a domain instead of a `const ref`. To be honest im also not sure when it's appropriate to use `const ref` for a domain vs just `const`

src/SegmentedString.chpl
ronawho: Yup, good catch.`array.domain` returns an immutable thing, but using `const ref` allows us to avoid making a copy since `const aD = array.domain` would result in a domain copy, which we want to avoid because copying distributed-domains isn't free).

> To be honest im also not sure when it's appropriate to use const ref for a domain vs just const

My understanding is:

```chpl
    const aD = a.domain; // creates an immutable copy of a.domain
      var aD = a.domain; // creates a mutable copy of a.domain
const ref aD = a.domain; // captures an immutable reference to a.domain
      ref aD = a.domain; // should be illegal since `a.domain` cannot be modified.
```

For arkouda I don't think there's every a reason to prefer `const aD = a.domain;`. The only case I could imagine this being useful is if you have a non-distributed domain and you want to localize it to a new node. Something like:

```chpl
const d = {1..10};
on Locales[1] {
  const localD = d;
}
```

to avoid having to do comm when accessing localD

I'm also surprised the compiler didn't error about this `ref` since that should be illegal. I wonder if this code isn't called or there's a missing const check in the compiler somewhere.

---

And not done in Arkouda, but just for context/background on why `.domain` returns an immutable thing -- the way to resize arrays in chapel is to resize the domain it's declared over. So if you want to grow an array you'd do something like:

```chpl
var d = {1..10};
var A, B: [d] int;
writeln(A.size); // 10
d = {1..20};
writeln(A.size); // 20
```

But we don't want to support resizing through `.domain` for performance reasons and because we'd think it'd be confusing for something like `A.domain = {1..20};` to resize other arrays as well. https://chapel-lang.org/docs/language/spec/arrays.html#association-of-arrays-to-domains has some more info on this.

pierce314159: Thanks for the thorough explanation!

pierce314159 (approved):

PR #1916, opened 2022-11-18
Closes #1914 - Adding `autoShutdown` flag to server to shutdown on client disconnect

### LINKED ISSUE #1914: Add flag to arkouda-server to automatically shutdown the server when the client disconnects

joshmarshall1:
This PR Closes #1914 

Adds optional `--autoShutdown=<true/false>` flag to server startup command to automatically shutdown the server when the client disconnects. This flag is defaulted to false.

I tested this manually through `ipython`, with a script that uses `util.tests.util.start_arkouda_server(1, server_args=["--autoShutdown=true"])`, and through a jupyter notebook.

### DIFF
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -58,6 +58,11 @@ module ServerConfig
     config const serverConnectionInfo: string = getEnv("ARKOUDA_SERVER_CONNECTION_INFO", "");
 
     /*
+    Flag to shut down the arkouda server automatically when the client disconnects
+    */
+    config const autoShutdown = false;
+
+    /*
     Hostname where I am running
     */
     var serverHostname: string = try! get_hostname();
@@ -146,6 +151,7 @@ module ServerConfig
             const logLevel: LogLevel;
             const regexMaxCaptures: int;
             const byteorder: string;
+            const autoShutdown: bool;
         }
 
         var (Zmajor, Zminor, Zmicro) = ZMQ.version;
@@ -168,7 +174,8 @@ module ServerConfig
             authenticate = authenticate,
             logLevel = logLevel,
             regexMaxCaptures = regexMaxCaptures,
-            byteorder = try! getByteorder()
+            byteorder = try! getByteorder(),
+            autoShutdown = autoShutdown
         );
         return try! "%jt".format(cfg);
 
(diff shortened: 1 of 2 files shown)


hokiegeek2 (approved):
looks good

Ethan-DeBandi99 (commented):
This looks good for the auto shutdown feature. I do want to check, wasn't there a component for connection that we wanted to add as well? I believe we want to add a component to return the connection info to standard out. I didn't see an issue for this, so I wanted to check that we are planning to handle that portion separately. Though, I think we could also add it here if we wanted to go ahead an implement the full solution.

joshmarshall1:
> This looks good for the auto shutdown feature. I do want to check, wasn't there a component for connection that we wanted to add as well? I believe we want to add a component to return the connection info to standard out. I didn't see an issue for this, so I wanted to check that we are planning to handle that portion separately. Though, I think we could also add it here if we wanted to go ahead an implement the full solution.

I was under the impression that was going to be handled separately. I can add an issue and start working it or if anyone feels strongly about it, I can add it to this one

Ethan-DeBandi99:
> > This looks good for the auto shutdown feature. I do want to check, wasn't there a component for connection that we wanted to add as well? I believe we want to add a component to return the connection info to standard out. I didn't see an issue for this, so I wanted to check that we are planning to handle that portion separately. Though, I think we could also add it here if we wanted to go ahead an implement the full solution.
> 
> I was under the impression that was going to be handled separately. I can add an issue and start working it or if anyone feels strongly about it, I can add it to this one

10-4. Let's add the issue and get this pushed in. Thanks!

Ethan-DeBandi99 (approved):

joshmarshall1:
> > > This looks good for the auto shutdown feature. I do want to check, wasn't there a component for connection that we wanted to add as well? I believe we want to add a component to return the connection info to standard out. I didn't see an issue for this, so I wanted to check that we are planning to handle that portion separately. Though, I think we could also add it here if we wanted to go ahead an implement the full solution.
> > 
> > 
> > I was under the impression that was going to be handled separately. I can add an issue and start working it or if anyone feels strongly about it, I can add it to this one
> 
> 10-4. Let's add the issue and get this pushed in. Thanks!

Issue #1917

pierce314159 (approved):

PR #1920, opened 2022-11-18
Closes #1917 - Adds server config json to stdout

### LINKED ISSUE #1917: Add connection info json to stdout when server starts
It was requested that the server info be added in place of the usual output so it could be parsed by a script. I'll also add the option to have this returned to the client so it can be directly accessed as a response from the server connection.

joshmarshall1:
This PR closes #1917 

Adds a server startup command flag `--serverInfoNoSplash=<true/false>` that replaces the normal server splash message with a json string of server information.

This also adds that the server info string gets returned to the client through the `ak.connect()` call, so it's easily accessible from the client.

With `--serverInfoNoSplash=false` (default setting)

image

With `--serverInfoNoSplash=true`

image

**note: `akserver` is an alias I have setup. To remove confusion, this is what it does: 
```
alias arkouda="cd /home/josh/git/arkouda"
alias akserver="arkouda;./arkouda_server -nl 1"
```

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -207,8 +207,6 @@ def connect(
     regexMaxCaptures = serverConfig["regexMaxCaptures"]  # type:ignore
     clientLogger.info(return_message)
 
-    return cast(str, generic_msg(cmd="getconfig", args={}))
-
 
 def _parse_url(url: str) -> Tuple[str, int, Optional[str]]:
     """
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -106,7 +106,7 @@ def connect(
     timeout: int = 0,
     access_token: str = None,
     connect_url=None,
-) -> str:
+) -> None:
     """
     Connect to a running arkouda server.
 
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -129,7 +129,7 @@ def connect(
 
     Returns
     -------
-    JSON string of connection configuration information
+    None
 
     Raises
     ------
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -106,7 +106,7 @@ def connect(
     timeout: int = 0,
     access_token: str = None,
     connect_url=None,
-) -> None:
+) -> str:
     """
     Connect to a running arkouda server.
 
@@ -129,7 +129,7 @@ def connect(
 
     Returns
     -------
-    None
+    JSON string of connection configuration information
 
     Raises
     ------
@@ -206,6 +206,8 @@ def connect(
         )
     regexMaxCaptures = serverConfig["regexMaxCaptures"]  # type:ignore
     clientLogger.info(return_message)
+
+    return cast(str, generic_msg(cmd="getconfig", args={}))
 
 
 def _parse_url(url: str) -> Tuple[str, int, Optional[str]]:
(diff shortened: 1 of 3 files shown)


src/ServerDaemon.chpl
```diff
@@ -401,7 +401,14 @@ module ServerDaemon {
     
             this.connectUrl = this.getConnectUrl(this.serverToken);
             this.createServerConnectionInfo();
-            this.printServerSplashMessage(this.serverToken,this.arkDirectory);
+            if serverInfoNoSplash {
+                writeln("Server Information:");
+                writeln(getConfig());
```
pierce314159, line 406: I'm not sure if we want to write `"Server Information:"` first or just write `getConfig()` since this mode should be primarily used by users who want easy access to the config info, maybe it would be easier to parse? I don't know much about this stuff though, so i'll defer to other reviewers expertise

joshmarshall1: You make a good point, this will rarely be used out of context I wouldn't think, so I'll make this change for ease of parsing

joshmarshall1: Updated

src/ServerDaemon.chpl
```diff
@@ -401,7 +401,11 @@ module ServerDaemon {
     
             this.connectUrl = this.getConnectUrl(this.serverToken);
             this.createServerConnectionInfo();
-            this.printServerSplashMessage(this.serverToken,this.arkDirectory);
+            if serverInfoNoSplash {
+                writeln(getConfig());
```
Ethan-DeBandi99, line 405: After this `writeln`, you need a `stdout.flush()`. Without this you cannot read the output JSON on the Python side. Since the process is non-terminating when running Arkouda, it appears to just hang in the current configuration.

Ethan-DeBandi99 (changes requested):
One update is needed for this to be functional for the use case we were provided.

arkouda/client.py
```diff
@@ -207,6 +207,8 @@ def connect(
     regexMaxCaptures = serverConfig["regexMaxCaptures"]  # type:ignore
     clientLogger.info(return_message)
 
+    return cast(str, generic_msg(cmd="getconfig", args={}))
```
Ethan-DeBandi99, line 210: I don't think we need to return the config info here. The use case provided, this will be read from `stdout` when the server is started.

src/ServerDaemon.chpl
joshmarshall1: Fixed!

arkouda/client.py
joshmarshall1: Removed - I added this initially as a free convenience bit

arkouda/client.py
```diff
@@ -129,7 +129,7 @@ def connect(
 
     Returns
     -------
-    None
```
pierce314159, line 132: I don't see a `return` in this function, so I don't think it returns anything. I might be missing something

arkouda/client.py
```diff
@@ -129,7 +129,7 @@ def connect(
...
+    JSON string of connection configuration information
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Looks like this may be left over from the return on the previous commit. Just needs to be reverted.

Ethan-DeBandi99 (commented):
Other than updating the docstring this should be all set

arkouda/client.py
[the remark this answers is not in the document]
joshmarshall1: Missed this when I removed the return, fixed now

Ethan-DeBandi99 (approved):
Looks good!

pierce314159 (approved):
Looks good!! Nice work

PR #1921, opened 2022-11-18
Closes #1901: Reduce SymEntry creation overheads

### LINKED ISSUE #1901: Reduce SymEntry creation overheads

ronawho:
Eliminate overheads in `SymEntry` creation. Previously, creating a `SymEntry` was more expensive than just creating the underlying Chapel array. This is because `SymEntry` used to store the domain and array, and the way it was declared resulted in initializing the domain and then assigning to it, which has unnecessary overhead. We don't actually need to separately store the domain (we can access it through `a.domain`) so just stop storing it. This also optimizes the case where a new array is created and put into a SymEntry, but never used again. By adding the `in` intent to the `SymEntry` initializer the compiler is able to optimize this case and steal the dead array instead of creating a new array and assigning.

Here are 16-node-cs-hdr timings for the included `SymEntrySpeedTest` test, which measures Chapel array creation time compared to `SymEntry` creation time:

Before:
```
250 arrs from scratch  : 0.27s
250 syms from scratch  : 0.59s
250 arrs from dom      : 0.08s
250 arrs from sym      : 0.08s
250 arrs from arr      : 0.11s
250 syms from arr      : 0.43s
250 arrs from dead arr : 0.04s
250 syms from dead arr : 0.44s
```

Now:
```
250 arrs from scratch  : 0.27s
250 syms from scratch  : 0.27s
250 arrs from dom      : 0.08s
250 arrs from sym      : 0.08s
250 arrs from arr      : 0.11s
250 syms from arr      : 0.11s
250 arrs from dead arr : 0.04s
250 syms from dead arr : 0.04s
```

Where the important thing to note is that before there were various overheads for constructing a `SymEntry` compared to a plain Chapel array and now there aren't. Also note that these timings were with Chapel main, which has some array creation optimizations added so these timings are faster than the 1.28 release. The unchanged result (creating a array from a `SymEntry`) is just to highlight that `SymEntry.a.domain` doesn't add overhead, but adding a `proc aD` wrapper would have.

A side-effect of not storing the domain is that we need an explicit initializer for creating a SymEntry from a non-distributed array. There are a few cases in Arkouda that were relying on this and I want to preserve that functionality for now (previously this worked because we'd create the distributed array in the `SymEntry` and then assign the local array to it, but now we're directing creating or stealing from the passed in array in the main initializer)

Closes #1901

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -197,11 +197,12 @@ module MultiTypeSymEntry
         type etype;
 
         /*
-        'aD' is the distributed domain for 'a' whose value and type
-        are defined by makeDistDom() to support varying distributions
+        'a' is the distributed array whose value and type are defined by
+        makeDist{Dom,Array}() to support varying distributions
         */
-        const aD: makeDistDom(size).type;
-        var a: [aD] etype;
+        var a = makeDistArray(size, etype);
+        /* Removed domain accessor, use `a.domain` instead */
+        proc aD { compilerError("SymEntry.aD has been removed, use SymEntry.a.domain instead"); }
         
         /*
         This init takes length and element type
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
Looks good to me!

PR #1923, opened 2022-11-21
Closes #1922 - Running Arkouda from a script example documentation

### LINKED ISSUE #1922: Add Documentation on Running Arkouda from a Script
Users have presented a case where they are starting an Arkouda server from a python script and want to connect to it, perform some operations and then shutdown on exit. We need to provide documentation on this process with the flags added in PRs #1920 and #1916.

I put together some example code to be used in the documentation below
```python
import arkouda as ak

import subprocess
import json

cmd = "/Users/ethandebandi/Documents/git/arkouda/./arkouda_server -nl 1 --serverInfoNoSplash=true --autoShutdown=true"
p = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True)
# get the output as a string
connect_str = p.stdout.readline()

server_data = json.loads(connect_str)
server, port = server_data['serverHostname'], server_data['ServerPort']

ak.connect(server=server, port=int(port))

# Perform whatever Arkouda actions you need

ak.disconnect()
```

joshmarshall1:
This PR closes #1922 

This is a document that outlines an example implementation (provided by @Ethan-DeBandi99 ) for running Arkouda using a script with the new server config flags implemented in #1916 and #1920

I'm not sure what other information could be useful to include here, so any feedback is welcome

### DIFF
--- a/training/RUNNING_FROM_SCRIPT.md
+++ b/training/RUNNING_FROM_SCRIPT.md
@@ -1,0 +1,47 @@
+# Running Arkouda from a Script
+
+<a id="toc"></a>
+## Table of Contents
+1. [Overview](#overview)
+2. [New Server Startup Flags](#flags)
+3. [Example Implementation](#example)
+
+<a id="overview"></a>
+## Overview
+The purpose of this document is to provide an example process for using a script to automatically run the Arkouda server, connect to the server, perform required actions, then disconnect from the server without any user interaction required.
+
+<a id="flags"></a>
+## New Server Startup Flags
+Two new flags were added to the arkouda_server startup command to help streamline this type of Arkouda usage. These flags are both boolean flags that are `false` by default.
+
+- `--serverInfoNoSplash`
+  - This flag replaces the usual splash message in the output of the server startup with the server configuration JSON. This JSON can then be parsed to read the server host, server port, and any other configuration info you may need.
+- `--autoShutdown`
+  - This flag toggles whether the arkouda_server will shut itself down when the client disconnects. When set to `true`, `ak.disconnect()` will trigger the server shutdown process.
+
+<a id="example"></a>
+## Example Implementation
+
+```python
+import arkouda as ak
+
+import subprocess
+import json
+
+# Update the below path to point to your arkouda_server
+cmd = "/Users/<username>/Documents/git/arkouda/arkouda_server -nl 1 --serverInfoNoSplash=true --autoShutdown=true"
+p = subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True)
+# get the output as a string
+connect_str = p.stdout.readline()
+
+server_data = json.loads(connect_str)
+server, port = server_data['serverHostname'], server_data['ServerPort']
+
+ak.connect(server=server, port=int(port))
+
+# Perform whatever Arkouda actions you need
+a = ak.arange(15)
+b = ak.array(["This", "is", "an", "example"])
+
+ak.disconnect()
+```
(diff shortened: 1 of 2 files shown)


training/RUNNING_FROM_SCRIPT.md
```diff
@@ -0,0 +1,72 @@
...
+
+<a id="toc"></a>
+## Table of Contents
+1. [Overview](#overview)
+2. [New Server Startup Flags](#flags)
+3. [Example Implementation](#example)
+
+<a id="overview"></a>
+## Overview
+The purpose of this document is to provide an example process for using a script to automatically run the Arkouda server, connect to the server, perform required actions, then disconnect from the server without any user interaction required.
```
Ethan-DeBandi99, line 11:
```suggestion
The purpose of this document is to provide an example process for using a script to automatically run the Arkouda server, connect to the server, perform required actions, then shutdown the server without any user interaction required.

```

Ethan-DeBandi99 (commented):
Few suggestions on wording, but otherwise I think this is good.

Ethan-DeBandi99 (approved):

pierce314159 (commented):
This looks great to me! Only one request, can we add a quick write up of these flage in [`ENVIRONMENT.md` ](https://github.com/Bears-R-Us/arkouda/blob/master/ENVIRONMENT.md#running), which also points to this example?

joshmarshall1:
> This looks great to me! Only one request, can we add a quick write up of these flage in [`ENVIRONMENT.md` ](https://github.com/Bears-R-Us/arkouda/blob/master/ENVIRONMENT.md#running), which also points to this example?

I don't think ENVIRONMENT.md is the right place for this request since that file outlines environment variables and doesn't cover any of the available flags for the arkouda_server

Ethan-DeBandi99:
@pierce314159 - did you have a suggestion based on Josh's comment? 

I was looking at our Docs and `README.md` has a section on running the server. My suggestion would be to add a note there that Arkouda is able to be started from a started in a script and link to the document added with this PR for now if any further updates are to be made.

If we have anywhere that lists flags for the server, it is not obvious. My overall recommendation moving forward would be to have a document on running the server that removes the section from `README.md` and adds the flags that are available. I think this would be a separate issue down the road though.

joshmarshall1:
> @pierce314159 - did you have a suggestion based on Josh's comment?
> 
> I was looking at our Docs and `README.md` has a section on running the server. My suggestion would be to add a note there that Arkouda is able to be started from a started in a script and link to the document added with this PR for now if any further updates are to be made.
> 
> If we have anywhere that lists flags for the server, it is not obvious. My overall recommendation moving forward would be to have a document on running the server that removes the section from `README.md` and adds the flags that are available. I think this would be a separate issue down the road though.

I would agree with this. It seems like a good idea to create a document down the road outlining running the server and the available methods/flags for doing so. 

For now, I will add a section to the `README.md` that mentions running from a script on points to this document.

pierce314159:
> > @pierce314159 - did you have a suggestion based on Josh's comment?
> > I was looking at our Docs and `README.md` has a section on running the server. My suggestion would be to add a note there that Arkouda is able to be started from a started in a script and link to the document added with this PR for now if any further updates are to be made.
> > If we have anywhere that lists flags for the server, it is not obvious. My overall recommendation moving forward would be to have a document on running the server that removes the section from `README.md` and adds the flags that are available. I think this would be a separate issue down the road though.
> 
> I would agree with this. It seems like a good idea to create a document down the road outlining running the server and the available methods/flags for doing so.
> 
> For now, I will add a section to the `README.md` that mentions running from a script on points to this document.

yeah  i did a bit of digging for where this should fit in our documentation yesterday and came up short. I'm good with adding a section to the `README` and eventually adding a file outlining server flags

pierce314159 (approved):

PR #1930, opened 2022-11-22
Closes #1926: Fix `+=` for `uint` arrays

### LINKED ISSUE #1926: Fix `+=` for `uint` arrays
We are missing support for operation equals for `uint` arrays
```python
>>> u = ak.array([13], dtype=ak.uint64)
>>> u += 1
RuntimeError                             
Traceback (most recent call last)

RuntimeError: Error: opeqvsMsg: unrecognized type: (uint64,uint64)
```

pierce314159:
This PR (closes #1926):
- Adds support for operation equals with uint arrays and scalars
- Adds a test of this functionality
- Updates `//=` with reals to use `floorDivisionHelper` to ensure proper behavior for corner cases

NOTE:
Currently for `arr opequals scalar` we cast the scalar to the return type if possible, I left this the same for this PR. But I'm not sure if this is the behavior users would expect, so I wanted to get input from others.
```python
>>> u_arr = ak.arange(10, dtype=ak.uint64)
>>> u_arr
array([0 1 2 3 4 5 6 7 8 9])

>>> u_arr += 3.14
>>> u_arr
array([3 4 5 6 7 8 9 10 11 12])

>>> u_arr += -7
>>> u_arr
array([18446744073709551612 18446744073709551613 18446744073709551614 18446744073709551615 0 1 2 3 4 5])
```

Co-authored-by: jaketrookman <jaketrookman@users.noreply.github.com>

### DIFF
--- a/src/OperatorMsg.chpl
+++ b/src/OperatorMsg.chpl
@@ -753,6 +753,7 @@ module OperatorMsg
                 }
             }
             when (DType.Int64, DType.UInt64) {
+                // The result of operations between int and uint are float by default which doesn't fit in either type
                 var errorMsg = notImplementedError(pn,left.dtype,op,right.dtype);
                 omLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
                 return new MsgTuple(errorMsg, MsgType.ERROR);
@@ -777,6 +778,7 @@ module OperatorMsg
                 }
             }
             when (DType.UInt64, DType.Int64) {
+                // The result of operations between int and uint are float by default which doesn't fit in either type
                 var errorMsg = notImplementedError(pn,left.dtype,op,right.dtype);
                 omLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
                 return new MsgTuple(errorMsg, MsgType.ERROR);
@@ -787,14 +789,7 @@ module OperatorMsg
                 select op {
                     when "+=" { l.a += r.a; }
                     when "-=" {
-                        if || reduce (l.a < r.a) {
-                            // will result in a negative which won't work in a uint
-                            var errorMsg = "Subtracting by an amount larger than self results in a negative";
-                            return new MsgTuple(errorMsg, MsgType.ERROR);
-                        }
-                        else {
-                            l.a -= r.a;
-                        }
+                        l.a -= r.a;
                     }
                     when "*=" { l.a *= r.a; }
                     when "//=" {
@@ -1005,6 +1000,7 @@ module OperatorMsg
                 }
             }
             when (DType.Int64, DType.UInt64) {
+                // The result of operations between int and uint are float by default which doesn't fit in either type
                 var errorMsg = notImplementedError(pn,left.dtype,op,dtype);
                 omLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
                 return new MsgTuple(errorMsg, MsgType.ERROR);
@@ -1029,6 +1025,7 @@ module OperatorMsg
                 }
             }
             when (DType.UInt64, DType.Int64) {
+                // The result of operations between int and uint are float by default which doesn't fit in either type
                 var errorMsg = notImplementedError(pn,left.dtype,op,dtype);
                 omLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
                 return new MsgTuple(errorMsg, MsgType.ERROR);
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
This looks good. It may be beneficial to add a comment to indicate why `UInt64, Int64` and `Int64, UInt64` operations are not supported. I am thinking this may make things more clear for anyone looking at it in the future.

pierce314159:
> This looks good. It may be beneficial to add a comment to indicate why `UInt64, Int64` and `Int64, UInt64` operations are not supported. I am thinking this may make things more clear for anyone looking at it in the future.

yeah it's not obvious, i guess the best reasoning is that it's not obvious what the return type should be when mixing `int` and `uint` (numpy by default converts their output to a float which won't fit into either of these types), but the real reason is probably because it matches what numpy does best
```python
>>> u = np.array([2], dtype=np.uint)
>>> u -= np.array([7])
----> 1 u -= np.array([7])
UFuncTypeError: Cannot cast ufunc 'subtract' output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'

>>> u -= 7
>>> u
array([18446744073709551611], dtype=uint64)

>>> i = np.array([2])
>>> i -= np.array([7], dtype=np.uint)
----> 1 i -= np.array([7], dtype=np.uint)
UFuncTypeError: Cannot cast ufunc 'subtract' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

>>> i -= np.uint(7)
>>> i
array([-5])
```
However numpy does handle opequals with a float a bit more intuitively than we do
```python
>>> u = np.array([2], dtype=np.uint)
>>> u += 3.14
----> 1 u += 3.14
UFuncTypeError: Cannot cast ufunc 'add' output from dtype('float64') to dtype('uint64') with casting rule 'same_kind'
```

joshmarshall1 (approved):
Looks good

pierce314159:
I added the comment that @Ethan-DeBandi99 requested. I also removed the guardrails around uint subtractions which result in a negative to maintain behavior like numpy where these wrap around to the appropriate `uint` value. 

I added Issue #1931 to track the weirdness I noted in the example and offer a potential solution if we decide we want to change this to behave like numpy

src/OperatorMsg.chpl
```diff
@@ -926,62 +999,142 @@ module OperatorMsg
...
+                select op {
+                    when "+=" { l.a += val; }
+                    when "-=" { l.a -= val; }
+                    when "*=" { l.a *= val; }
+                    when "//=" {
+                        ref la = l.a;
+                        [li in la] li = floorDivisionHelper(li, val);
+                    }//floordiv
+                    when "%=" {
+                        if val != 0 {l.a %= val;} else {l.a = 0;}
```
bradcray, line 1109: @Ethan-DeBandi99 : With a quick look, this PR seems to have broken our smoke testing with:

```
$CHPL_HOME/test/studies/arkouda/arkouda//src/OperatorMsg.chpl:954: In function 'opeqvsMsg':
$CHPL_HOME/test/studies/arkouda/arkouda//src/OperatorMsg.chpl:1109: error: unresolved call '%([domain(1,int(64),false)] real(64), uint(64))'
$CHPL_HOME/modules/internal/ChapelBase.chpl:2211: note: this candidate did not match: %(a: uint(64), param b: uint(64))
$CHPL_HOME/modules/internal/ChapelBase.chpl:2012: note: because actual argument #1 with type '[domain(1,int(64),false)] real(64)'
$CHPL_HOME/modules/internal/ChapelBase.chpl:2211: note: is passed to formal 'a: uint(64)'
$CHPL_HOME/test/studies/arkouda/arkouda//src/OperatorMsg.chpl:1109: note: other candidates are:
$CHPL_HOME/modules/internal/ChapelBase.chpl:2222: note:   %(a: int(64), param b: int(64))
$CHPL_HOME/modules/internal/ChapelBase.chpl:400: note:   %(a: int(8), b: int(8))
note: and 23 other candidates, use --print-all-candidates to see them
Makefile:321: recipe for target 'arkouda_server' failed
```

are you not seeing the same behavior?

ronawho: CI testing with Chapel 1.28 was clean, so I suspect this is from a change to chapel main, which is what our smoke testing uses (guessing something related to preferring instantiations over converting or something)

bmcdonald3: The Arkouda CI only tests Chapel 1.28.0 and 1.29.0 currently, where I am able to reproduce this on `main`, but not one of those older releases.

ronawho: @pierce314159 internally we're trying to figure out if Chapel intends to support `%`/`%=` on reals. According to our spec, we don't yet, but previously the implementation supported `%=` (perhaps accidentally)

However, I'm wondering if you actually need/want to support it yourself here. It looks like you added support for `Float64 %= UInt64`, but I don't think this is supported for any other float parings (`Float64 %= Int64` or `Float64 %= Float64`) so I'm wondering if this was an accidental addition and not needed.

bradcray: FWIW, it looks like I may have added the %= overload on reals accidentally some time ago—at least, I don't see any real justification for it.  Also, I checked and C doesn't support `%` or `%=` on doubles (i.e., I'm not just forgetting some advanced math).  So I don't currently think Chapel should be supporting % on reals, which would suggest avoiding reliance on it in Arkouda (i.e., I'm also interested in the answer to Elliot's question about whether its use here was intentional).

I'm doing a check to see whether any other Chapel code is relying on that `%=` overload on reals...

PR #1936, opened 2022-11-30
Closes #1935: Increase encode benchmark size

### LINKED ISSUE #1935: Increase codec benchmark default problem size
The codec default problem size is very small today because the unoptimized code was much less performant, so it needed to be small for the old code to finish in a reasonable amount of time. Now that it is more optimized, increasing the problem size will give us a better read of the performance.

bmcdonald3:
The codec default problem size is very small today because the unoptimized code was much less performant, so it needed to be small for the old code to finish in a reasonable amount of time. Now that it is more optimized, increasing the problem size will give us a better read of the performance.

### DIFF
--- a/benchmarks/encode.py
+++ b/benchmarks/encode.py
@@ -72,7 +72,7 @@ def create_parser():
     parser.add_argument("hostname", help="Hostname of arkouda server")
     parser.add_argument("port", type=int, help="Port of arkouda server")
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):

Ethan-DeBandi99 (approved):
Looks good. Thanks again for knocking all the encoding support out!

PR #1937, opened 2022-11-30
Create local copies of remote strings for encoding

bmcdonald3:
The original encoding code was referencing a remote string, causing some extra communication, so creating a local copy on each locale helps cut that down.


pierce314159 (approved):

Ethan-DeBandi99 (approved):
Looks good!

PR #1942, opened 2022-11-30
Closes #1941 - Proper pattern search to get the correct error

### LINKED ISSUE #1941: Error on patterns that match `""` when Strings contains an `""`
This is related to PR #1677. I meant to block regex operations for patterns that match the empty string when the strings object contains the empty string, but I mistyped. I believe the only thing that needs to be changed is
https://github.com/Bears-R-Us/arkouda/blob/ab31e6a5a4d58a7d06dc629930d98a46ff9707ea/arkouda/strings.py#L670-L673

to be something like:
```python
def _empty_pattern_verification(self, pattern):
    if pattern == "$" or (re.search(self.pattern, "") and (self == "").any()):  # type: ignore
```

reproducer for this:
```python
>>> ak.array(['abc', '', 'def']).findall('.*')
```

jaketrookman:
This PR (Closes https://github.com/Bears-R-Us/arkouda/issues/1941):

With the goal of stopping an infinite loop bug, the empty string check now raises an error.

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -668,7 +668,7 @@ class Strings:
         self._regex_dict = dict()
 
     def _empty_pattern_verification(self, pattern):
-        if pattern == "$" or (pattern == "" and (self == "").any()):  # type: ignore
+        if pattern == "$" or (re.search(pattern, "") and (self == "").any()):  # type: ignore
             # TODO remove once changes from chapel issue #20431 and #20441 are in arkouda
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (changes requested):
The unit test need to be updated to match this change.

Ethan-DeBandi99 (approved):

pierce314159 (approved):
Looks good!!

PR #1946, opened 2022-12-01
Closes 1945 - Remove `src/exec`

### LINKED ISSUE #1945: Remove `src/exec`
Now that the Chapel native support for codecs has been implemented, we will no longer need the python files used in the original work around.

Ethan-DeBandi99:
Closes #1945 

With the updates made to support codecs in Chapel, we no longer need the python files from the work around.


pierce314159 (approved):

jaketrookman (approved):
Looks good.

PR #1948, opened 2022-12-02
Closes #1947 - Conversion Error when IPv4 is Index

### LINKED ISSUE #1947: Conversion Error when `IPv4` is Index
Reproducing Code:

```python
ak.DataFrame({'a': ak.IPv4(ak.array([1,2,3,4]))}).groupby('a').count()
```

The above code will run fine and produce the expected results. However, it errors when attempting to display when dealing with `IPv4`. The issue is that the IPv4 is not being converted to Pandas formatting properly.

Ethan-DeBandi99:
Closes #1947 

Updates `ak.Index.to_pandas` method to assign the `dtype` of the pandas object based on the `dtype` of the NumPy object instead of the Arkouda object. Previously if an index contained an `IPv4`, it would attempt to assign the `dtype` to `uint64` because that was the Arkouda `dtype`. This resulted in an attempt to cast the `IPv4` string to `uint64`, causing the error.

Adds testing to ensure proper functionality moving forward based on the example provided in the issue.

### DIFF
--- a/tests/dataframe_test.py
+++ b/tests/dataframe_test.py
@@ -387,6 +387,11 @@ class DataFrameTest(ArkoudaTest):
         self.assertListEqual(c.index.to_list(), ["Bob", "Alice", "Carol"])
         self.assertListEqual(c.values.to_list(), [2, 3, 1])
 
+        # testing counts with IPv4 column
+        s = ak.DataFrame({'a': ak.IPv4(ak.array([1, 2, 3, 4]))}).groupby('a').count()
+        pds = pd.Series(data=np.array([1, 1, 1, 1]), index=pd.Index(data=np.array(['0.0.0.1', '0.0.0.2', '0.0.0.3', '0.0.0.4'], dtype="<U7")))
+        self.assertTrue(s.to_pandas().equals(other=pds))
+
     def test_to_pandas(self):
         df = build_ak_df()
         pd_df = build_pd_df()
--- a/tests/dataframe_test.py
+++ b/tests/dataframe_test.py
@@ -388,8 +388,10 @@ class DataFrameTest(ArkoudaTest):
         self.assertListEqual(c.values.to_list(), [2, 3, 1])
 
         # testing counts with IPv4 column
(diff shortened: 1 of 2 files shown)


tests/dataframe_test.py
```diff
@@ -387,6 +387,11 @@ def test_gb_series(self):
         self.assertListEqual(c.index.to_list(), ["Bob", "Alice", "Carol"])
         self.assertListEqual(c.values.to_list(), [2, 3, 1])
 
+        # testing counts with IPv4 column
+        s = ak.DataFrame({'a': ak.IPv4(ak.array([1, 2, 3, 4]))}).groupby('a').count()
```
jaketrookman, line 391: Small style suggestion, change to ak.arange(1, 4)

tests/dataframe_test.py
```diff
@@ -387,6 +387,11 @@ def test_gb_series(self):
...
+        pds = pd.Series(data=np.array([1, 1, 1, 1]), index=pd.Index(data=np.array(['0.0.0.1', '0.0.0.2', '0.0.0.3', '0.0.0.4'], dtype="<U7")))
```
jaketrookman, line 392: Small style suggestion, change to ak.ones(4)

jaketrookman (commented):
It looks good, just a few stylistic suggestions.

tests/dataframe_test.py
jaketrookman: sorry, *np.ones(4)

pierce314159 (approved):

tests/dataframe_test.py
Ethan-DeBandi99: @jaketrookman - I agree with the use of `ak.arange` here. I just copied over the user provided reproducer originally. However, when using `ak.arange`, the stop parameter is exclusive, so we would need `ak.arange(1,5)`.

tests/dataframe_test.py
Ethan-DeBandi99: @jaketrookman - I updated this to `np.ones(4, dtype=np.int64)`. `np.ones` defaults the `dtype` to `float64`.

joshmarshall1 (approved):

PR #1950, opened 2022-12-02
Closes #1949: Allow conversions with IDNA and non-UTF-8 encodings

### LINKED ISSUE #1949: IDNA can only convert to/from UTF-8
IDNA currently can only convert to/from UTF-8, but if we loop through iconv and first convert to UTF-8, then we can simulate the conversion actually working.

bmcdonald3:
The `idn2` library only supports encoding/decoding IDNA to UTF-8,
so, in order to get to other encodings from IDNA, we can first
convert to UTF-8 through `idn2` and then convert to the desired
encoding through `iconv` (with the process being flipped for
decoding).

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -396,7 +396,7 @@ class Strings:
                 },
             )
             return Strings.from_return_msg(rep_msg)
-            
+
         rep_msg = generic_msg(
             cmd="encode",
             args={
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -374,6 +374,29 @@ class Strings:
         RuntimeError
             Raised if there is a server-side error thrown
         """
+        if (toEncoding.upper() == "IDNA" and fromEncoding.upper() != "UTF-8") or \
+           (toEncoding.upper() != "UTF-8" and fromEncoding.upper() == "IDNA"):
+            # first convert to UTF-8
+            rep_msg = generic_msg(
+                cmd="encode",
+                args={
+                    "toEncoding": "UTF-8",
+                    "fromEncoding": fromEncoding,
+                    "obj": self.entry,
+                },
+            )
+            intermediate = Strings.from_return_msg(rep_msg)
+            # now go to idna
+            rep_msg = generic_msg(
+                cmd="encode",
+                args={
+                    "toEncoding": toEncoding,
+                    "fromEncoding": "UTF-8",
+                    "obj": intermediate,
+                },
+            )
+            return Strings.from_return_msg(rep_msg)
+            
         rep_msg = generic_msg(
             cmd="encode",
             args={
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (commented):
@bmcdonald3 - Would it be possible to handle this on the server side? We specifically call out during the IDNA, that if we are working with something that is not UTF-8 , it should error. Could we just update this to convert to UTF-8 first? This could be a future iteration, I am just thinking that we don't really need to make 2 server calls for this.

bmcdonald3:
> Would it be possible to handle this on the server side? 

It would be possible, though I don't know if the performance gains would really be worth the effort in this case, since it is essentially doing the same thing, but would also require adding another helper function to handle the different argument types. Because of this, we wouldn't be saving much by converting to server-side. I think it is fine if we don't want to merge this, but, if we do, I think the client-side approach is much simpler for something that probably won't see much use and also helps not adding code-bloat to the server side.

> We specifically call out during the IDNA, that if we are working with something that is not UTF-8 , it should error.

I'm not sure I'm following this; are you saying that it is desired behavior to error if a non UTF-8 conversion is specified for IDNA? I don't recall hearing that, but I may have missed something (or forgot on vacation 🌴).

Ethan-DeBandi99:
@bmcdonald3 - I am fine with leaving it on the client, but wanted to ask about handling on the server. I think the issue of adding code to the server is a separate topic entirely at this point (mostly surrounding compile times). My initial thought was just that it would be easier to have everything in one place for maintenance. 

Also, the server side workflow has a check that if you are not working with UTF-8 to convert to IDNA, it should error here, https://github.com/Bears-R-Us/arkouda/blob/91501696b60186b56bd902c360f1079bac1c3493/src/Codecs.chpl#L10. I am not sure that we want to allow IDNA encoding from other encodings due to the fact that IDNA is designed to convert UTF-8 into an ASCII format.

bmcdonald3:
That check should never be hit, since we are never passing the `UTF-8` string to the server in combination with IDNA, so it would be an error if we did and I think we still want that check.

> I am not sure that we want to allow IDNA encoding from other encodings due to the fact that IDNA is designed to convert UTF-8 into an ASCII format.

I think that's a good point, but I also don't really see that anything is lost by adding it. It really is just for consistency with the rest of the API (and is in conformity with the Codecs API: https://docs.python.org/3/library/codecs.html).

It's fine if I am overruled here, but I don't really see much downside with adding the call and I don't think there is much of a maintenance burden from it; if you have further qualms, I am fine with just closing this and won't die on a hill to get this PR in.

Ethan-DeBandi99:
@bmcdonald3 - I was hoping a user would chime in and provide their preferred functionality. I agree, I does not hurt. I am fine with merging, but would like to check with users just to be sure they don't get anything unexpected.

@21771 - what is your preference since you were the original requestor of this functionality?

bmcdonald3:
@Ethan-DeBandi99 Did we want to go ahead and merge this or hold off for now? (just trying to get some housecleaning done)

Ethan-DeBandi99 (approved):

pierce314159 (approved):
Looks good to me

PR #1954, opened 2022-12-06
Closes #1953 CI Testing Error Fix

### LINKED ISSUE #1953: CI Testing Error
Several of the CI tests were no longer passing. Comments were provided by Ben and Elliot for fixes on slack.

Changes made to setup-python and ubuntu-latest.

jaketrookman:
Closes #1953 

Changes to the version numbers on the setup, actions, and ubuntu in the CI workflow. These changes were done with the suggestions made from Ben, Elliot, and Ethan.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,19 +7,19 @@ env:
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Check for tabs
       run: |
         ! git --no-pager grep -n $'\t' -- '*.chpl'
         
   mypy:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     container:
       image: chapel/chapel:1.28.0
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Install dependencies
       run: |
         apt-get update && apt-get install -y python3-pip libhdf5-dev hdf5-tools libzmq3-dev libcurl4-openssl-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -60,16 +60,16 @@ jobs:
           flake8 arkouda          
 
   arkouda_python_portability:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
ronawho, line 63: I think you only need/want `20.04` for the python portability job. And I'd add a comment with something like "use older ubuntu so setup-python uses python binaries with glibc requirements chpl docker can support"

.github/workflows/CI.yml
```diff
@@ -60,16 +60,16 @@ jobs:
...
     strategy:
       matrix:
         python-version: ['3.8', '3.9', '3.10']
     container:
       image: chapel/chapel:1.28.0
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
ronawho, line 70: Upgrade other jobs to `checkout@v3` as well?

ronawho:
Just capturing a slack comment here in case it's useful history (I'm imagining trying to figure out / remember why we pin to a specific version down the road and free slack only has 90 day history):

> I don't fully understand how `setup-python` works, but my rough understanding is basically that they build and cache python binaries for all the supported github actions environments. But we're doing something a little different where we're not using setup-python in a github actions environment, but instead inside the chapel docker image. So it's possible we're using it incorrectly and only now just getting bit by it. For instance, if they only built python for ubuntu 22.04 or something then it won't run on systems with an older glibc and that probably has a newer glibc than our docker image.
> 
> I think the difference is that `ubuntu-latest` (the github action image we use) is upgrading from 20.04 to 22.04, so probably setup-python is installing a newer python that has gibc deps our  docker image can't support.
>
> Changing the portability job to use `ubuntu-20.04` should workaround the issue, because we'd be back to using python binaries with the older glibc dependency

ronawho (approved):
👍

Ethan-DeBandi99 (approved):
Thanks for tracking this down @ronawho.

PR #1964, opened 2022-12-12
Closes #1963 - Message Arg failure when List contains `Strings` and `pdarray`

### LINKED ISSUE #1963: Messages with List Containing `Strings` and `pdarray`
When a list argument contains both `pdarray` and `strings`, the validation does not properly pass the `SymEntry` name to the server for processing. We need to update the logic to ensure the following an properly send the data to the Chapel side for processing.

- Lists only contain 1 type unless they contain `pdarray` and `strings`.
- If `pdarray`, `Strings` or a combination in the list, the type in the parameter is always `pdarray` we may want to change this in the future.
- `strings` and `pdarray` send the SymEntry name to be accessed on the server.

Ethan-DeBandi99:
Closes #1963 

Updates the logic for parsing message arguments that are lists. Expects lists to contain a single type with the exception of `pdarray` and `Strings`. A list can contain `pdarray` and `Strings` objects. In these cases, the type is set to `pdarray`. This does not have an impact on processing and is simply metadata.

### DIFF
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -130,7 +130,7 @@ class ParameterObject:
                 if t not in [pdarray.__name__, Strings.__name__]:
                     t_str = ", ".join(dtypes)
                     raise TypeError(f"Lists of multiple types can only "
-                                    f"conatin strings and pdarray. Found {t_str}")
+                                    f"contain strings and pdarray. Found {t_str}")
             # using pdarray for now. May change in future. This does not impact functionality
             t = pdarray.__name__
         if any(x == t for x in [pdarray.__name__, Strings.__name__]):
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -123,10 +123,15 @@ class ParameterObject:
 
         # want the object type. If pdarray the content dtypes can vary
         dtypes = {type(p).__name__ for p in val}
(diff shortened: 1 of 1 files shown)


arkouda/message.py
```diff
@@ -123,10 +123,16 @@ def _build_list_param(key: str, val: list) -> ParameterObject:
...
-            raise TypeError(f"List values must be of the same type. Found {t_str}")
-        t = dtypes.pop()
+        if len(dtypes) == 1:
+            t = dtypes.pop()
+        else:
+            for t in dtypes:
+                if t not in [pdarray.__name__, Strings.__name__]:
+                    t_str = ", ".join(dtypes)
+                    raise TypeError(f"Lists of multiple types can only "
+                                    f"conatin strings and pdarray. Found {t_str}")
```
jaketrookman, line 133: The logic looks right to me, just a small spelling mistake with "contain".

pierce314159 (approved):
Looks good

arkouda/message.py
Ethan-DeBandi99: fixed

jaketrookman (commented):
Looks good.

jaketrookman (approved):

Issue #1968, opened 2022-12-16
Parquet Benchmark - Append vs Multi-Column Workflows


Ethan-DeBandi99:
Add benchmarking to compare writing mulitple columns to Parquet files using the old `append` workflow and the new multi-column write workflow. Once this has been documented, the append workflow should be removed.

Ethan-DeBandi99:
Performance has been observed to be increased. Currently, I do not think there is a benefit to comparing these.

Issue #1967, opened 2022-12-16
Remove Append Option from Parquet


Ethan-DeBandi99:
Parquet is not designed to have columns appended to an existing file. In order to work around this before multi-column parquet write functionality was implemented, it was designed to allow for append. Unfortunately this process is extremely inefficient. 

Once Multi-Column writes are implemented, we will need to remove the append functionality.

Ethan-DeBandi99:
@pierce314159, @bmcdonald3 - I have everything set up to require a user to explicitly set append for that functionality. I know this is extremely inefficient, but I think we may want to allow this. However, I would like to handle it a bit more efficiently. I am thinking that we can read the entire file out and then use the multi-column workflow to write everything back in. I don't believe this is a high priority but wanted to get some thoughts from others.

pierce314159:
i don't have any issue with this approach. I think it's well documented that parquet append isn't efficient and should ve avoided if possible

PR #1969, opened 2022-12-16
Closes #1876 - Multi-Column Write Parquet

### LINKED ISSUE #1876: Multi-column Parquet write inefficiencies
Passing along a comment from another developer. " I've noticed that Parquet output from Arkouoda is currently very slow. The Arkouda client library saves multi-column dataframes by sending a separate writeParquet request to the server for each column, adding them to the file one-by-one. The ccp_appendColumnToParquet function doesn't add the column to the file in-place, though; instead it loads the whole existing file into memory, adds the new column to the table object in memory, and overwrites the whole file with the new version. That means it's loading and re-saving the content of all the old columns each time it adds a new one, which adds a lot of overhead since this happens for every column."

Ethan-DeBandi99:
Closes #1876 

This PR adds a separate workflow for writing multiple columns to Parquet. We now write all columns at once using the Arrow Parquet low level API. `append` write modes is maintained for Parquet for the time being, but is only used if `append` is specifically set as the write mode.  Since the `append` workflow is so inefficient, #1967 has been added to remove this. However, I think it would be good to get some benchmark data comparing the two write methods for multiple columns (#1968).

This implementation does require accessing `SymEntries` on each locale to get the local elements. This may not be the most efficient workflow, but I am unaware of another way to achieve the desired functionality. If anyone knows a better way to handle this, please let me know. However, for an initial implementation, this should be ready to go. If there is another option I would think we would handle the update under a separate PR.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -862,15 +862,32 @@ def to_parquet(
     >>> # Save using names instead of mapping
     >>> ak.to_parquet([a, b], 'path/name_prefix', names=['a', 'b'])
     """
-    if mode.lower() not in ["append", "truncate"]:
-        raise ValueError("Allowed modes are 'truncate' and 'append'")
-
-    datasetNames, pdarrays = _bulk_write_prep(columns, names)
-
-    for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
-        arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compressed)
-        if mode.lower() == "truncate":
-            mode = "append"
+    # if mode.lower() not in ["append", "truncate"]:
+    #     raise ValueError("Allowed modes are 'truncate' and 'append'")
+    #
+    # datasetNames, pdarrays = _bulk_write_prep(columns, names)
+    #
+    # for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
+    #     arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compressed)
+    #     if mode.lower() == "truncate":
+    #         mode = "append"
+    print(list(columns.values()))
+    print(list(columns.keys()))
+    print(len(columns.values()))
+    repmsg = cast(
+        str,
+        generic_msg(
+            cmd="toParquet_multi",
+            args={
+                "columns": list(columns.values()),
+                "col_names": list(columns.keys()),
+                "filename": prefix_path,
+                "num_cols": len(columns.values()),
+                "compressed": compressed,
+            },
+        ),
+    )
+    print(repmsg)
 
 
 def to_hdf(
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -883,7 +883,7 @@ def to_parquet(
                     "columns": pdarrays,
                     "col_names": datasetNames,
                     "filename": prefix_path,
-                    "num_cols": len(columns.values()),
+                    "num_cols": len(pdarrays),
                     "compressed": compressed,
                 },
             )
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -1,7 +1,6 @@
 import glob
 import json
 import os
-import warnings
 from typing import Dict, List, Mapping, Optional, Union, cast
 from warnings import warn
 
@@ -334,7 +333,7 @@ def _parse_errors(rep_msg, allow_errors: bool = False):
     file_errors = rep_msg["file_errors"] if "file_errors" in rep_msg else []
     if allow_errors and file_errors:
         file_error_count = rep_msg["file_error_count"] if "file_error_count" in rep_msg else -1
-        warnings.warn(
+        warn(
             f"There were {file_error_count} errors reading files on the server. "
             + f"Sample error messages {file_errors}",
             RuntimeWarning,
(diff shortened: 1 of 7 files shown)


joshmarshall1 (approved):

bmcdonald3:
Here are some numbers collected on 16-node-cs-hdr using DataFrames with 5 columns and 100,000,000 int64 elements per column per locale:

Locales | Append | This PR
-- | -- | --
1 | 14.9965 s | 3.3233 s
2 | 15.1250 s | 6.0503 s
4 | 14.6877 s | 6.0120 s
8 | 15.0165 s | 6.1107 s
16 | 15.8496 s | 6.1967 s

So it seems this results in about a 2.5x speedup over the old code.

I had to make some modifications to enable the old code path, but here is the branch that I was using to collect these numbers: https://github.com/bmcdonald3/arkouda/tree/multi-col-perf-test with the benchmark being run with 
```
$ python3 benchmarks/run_benchmarks.py df --size=100_000_000 -nl <numlocs>
$ python3 benchmarks/run_benchmarks.py df --size=100_000_000 -nl <numlocs> --oldcode
```

bmcdonald3 (approved):
Looks good to me; nothing jumped out from the C++ code 👍

Ethan-DeBandi99:
@bmcdonald3 - Thanks for running the performance tests!

jaketrookman (commented):
Looks good to me.

src/ArrowFunctions.h
```diff
@@ -3,13 +3,18 @@
 
 // Wrap functions in C extern if compiling C++ object file
 #ifdef __cplusplus
+#include <cassert>
+#include <fstream>
 #include <iostream>
+#include <memory>
 #include <arrow/api.h>
 #include <arrow/io/api.h>
+#include <arrow/table.h>
```
pierce314159, line 12: are we using `<arrow/table.h>` anywhere? It would be nice to double check all the new imports to make sure they're still needed

src/ParquetMsg.chpl
```diff
@@ -769,6 +770,245 @@ module ParquetMsg {
...
+            ptrList[i] = c_ptrTo(e.a[locDom]): c_void_ptr;
+            datatypes[i] = ARROWINT64;
+            sizeList[i] = locDom.size;
+          }
+          when DType.UInt64 {
+            var e = toSymEntry(toGenSymEntry(entry), uint);
+            var locDom = e.a.localSubdomain();
+            // set the pointer to the entry array in the list of Pointers
+            ptrList[i] = c_ptrTo(e.a[locDom]): c_void_ptr;
+            datatypes[i] = ARROWINT64;
```
pierce314159, line 860: should this be `ARROWUINT64`?

src/ParquetMsg.chpl
```diff
@@ -769,6 +770,245 @@ module ParquetMsg {
...
+            );
+          }
+        }
+      }
+      
+      // validate all elements same size
+      var numelems: int = sizeList[0];
+      if !(|| reduce (sizeList==numelems)) {
+        throw getErrorWithContext(
+              msg="Parquet columns must be the same size",
```
pierce314159, line 911: shouldn't this be `!(&& reduce (sizeList==numelems))`? because we want all columns to be the same size?

I think the current code will never fail since `numelems = sizeList[0]`, the first element of `sizeList == numelems` will always be true. So `|| reduce (sizeList == numelems)` is always true

pierce314159 (commented):
a few small things to check out-- clever handling of the strings case!

arkouda/io.py
```diff
@@ -865,12 +862,32 @@ def to_parquet(
...
-        arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compressed)
-        if mode.lower() == "truncate":
-            mode = "append"
+    datasetNames, pdarrays = _bulk_write_prep(columns, names)
+    # append or single column use the old logic
+    if mode.lower() == "append" or len(pdarrays) == 1:
+        for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
+            arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compressed)
+    else:
+        print(cast(
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Yeah it just returns a success message. This will also correctly handle error messages. Without the print, there is not record of success, but I would also be fine with removing it.

src/ArrowFunctions.h
Ethan-DeBandi99: That's odd, table should not be new, that was being used by the append code (which was not removed). Let me look into this.

src/ParquetMsg.chpl
Ethan-DeBandi99: Yes it should, thanks! Updated.

src/ArrowFunctions.h
Ethan-DeBandi99: I think I added some of these when trying out a few alternative methods, if memory serves correctly only `queue` is needed. I will remove and retest to be positive.

Ethan-DeBandi99: Verified only queue is needed. Updated

src/ParquetMsg.chpl
Ethan-DeBandi99: You are 100% correct idk how I missed that. Thanks for catching!

arkouda/io.py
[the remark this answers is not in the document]
pierce314159: sounds good! I was mostly making sure it wasn't leftover from debugging

src/ArrowFunctions.h
pierce314159: sweet!

src/ParquetMsg.chpl
pierce314159: Awesome! :)

pierce314159 (approved):
Nice work!

PR #1971, opened 2022-12-16
Closes #1970: Basic bigint pdarray

### LINKED ISSUE #1970: basic bigint pdarray
Plain jane bigint pdarray with no functionality

pierce314159:
This PR (closes #1970):
- Adds `ak.bigint` dtype
- Adds bigint pdarray with `max_bits`
- Adds pdarraycreation methods for bigint
- Adds casting to and from bigint pdarray
- Adds `ak.bigint_from_uint_arrays`

<details>
  <summary>Examples</summary>

```python
>>> a = ak.array([2410407066388485413312943138511743903783304490674189252952038, 4820814132776970826625886277023487807566608981348378505904076, 9641628265553941653251772554046975615133217962696757011808152, 19283256531107883306503545108093951230266435925393514023616304, 38566513062215766613007090216187902460532871850787028047232608, 77133026124431533226014180432375804921065743701574056094465216, 154266052248863066452028360864751609842131487403148112188930432, 308532104497726132904056721729503219684262974806296224377860864, 617064208995452265808113443459006439368525949612592448755721728, 1234128417990904531616226886918012878737051899225184897511443456], max_bits=200)

>>> a
array(["803469022129495137770981046170581301261101496891396417650662" "1606938044258990275541962092341162602522202993782792835301324" "1606938044258990275541962092341162602522202993782792835301272" "1606938044258990275541962092341162602522202993782792835301168" "1606938044258990275541962092341162602522202993782792835300960" "1606938044258990275541962092341162602522202993782792835300544" "1606938044258990275541962092341162602522202993782792835299712" "1606938044258990275541962092341162602522202993782792835298048" "1606938044258990275541962092341162602522202993782792835294720" "1606938044258990275541962092341162602522202993782792835288064"])

>>> a.dtype
dtype(bigint)

>>> a.dtype.name
'bigint'

>>> ak.arange(2**200, 2**200+5, dtype=ak.bigint)
array(["1606938044258990275541962092341162602522202993782792835301376" "1606938044258990275541962092341162602522202993782792835301377" "1606938044258990275541962092341162602522202993782792835301378" "1606938044258990275541962092341162602522202993782792835301379" "1606938044258990275541962092341162602522202993782792835301380"])

>>> s = ak.array(["18446744073709551616", "18446744073709551617", "18446744073709551618", "18446744073709551619", "18446744073709551620"])

>>> ak.cast(s, ak.bigint)
array(["18446744073709551616" "18446744073709551617" "18446744073709551618" "18446744073709551619" "18446744073709551620"])

>>> ak.arange(2**200, 2**200+5)
array(["1606938044258990275541962092341162602522202993782792835301376" "1606938044258990275541962092341162602522202993782792835301377" "1606938044258990275541962092341162602522202993782792835301378" "1606938044258990275541962092341162602522202993782792835301379" "1606938044258990275541962092341162602522202993782792835301380"])

>>> ak.bigint_from_uint_arrays([ak.ones(5, dtype=ak.uint64), ak.arange(5, dtype=ak.uint64)])
array(["18446744073709551616" "18446744073709551617" "18446744073709551618" "18446744073709551619" "18446744073709551620"])

>>> 18446744073709551616 == 2**64
True
```
</details>

Notes:
- running this requires `CHPL_GMP` be set. Either to `bundled` or `system`, whichever `CHPL_LLVM` is currently set to
- I already have most of indexing and binop code as well as `bigint_to_uint_arrays` and `slice_bits` methods, but we decided to break this into multiple PRs to make reviews more manageable

- I had to do some workarounds because the `bigint` chapel types can't be passed to the standard module `numBytes`. So I couldn't create symEntries and populate them after the fact i.e.
```chapel
// this doesn't work
var e = st.addEntry(rname, l.size, bigint);
e.a = l.a + r.a;
// this does work
var e = st.addEntry(rname, new shared SymEntry(l.a + r.a));
```
and for a seemingly similar reason, I can't do aggregation because `unorderedCopy` isn't supported for `bigint`.

Because of the above, I'm having to write bigint specific version of a lot of these functions to pass back a distributed array instead of populating a predefined symentry and to do straight assignment instead of aggregation. I'd like @ronawho to verify that these limitations make sense and there's not an easier way around this. Also at some point I'm probably gonna be hitting you up for help with `to_list/ndarray`

- And I'd like @hokiegeek2 to verify what I'm doing with `dtypes` isn't completely dumb


arkouda/dtypes.py
```diff
@@ -40,15 +41,38 @@
...
+    "bigint": "{:n}",
 }
 
-dtype = np.dtype
+
+def dtype(x):
+    if (isinstance(x, str) and x == "bigint") or isinstance(x, BigInt):
+        return bigint
+    else:
+        return np.dtype(x)
```
Ethan-DeBandi99, line 52: Just verifying my memory here. We are doing this because `BigInt` has a `dtype` of `object` in NumPy, correct? If that is the case, can we add a comment indicating that for clarity?

arkouda/pdarraycreation.py
```diff
@@ -196,13 +207,23 @@ def array(
 
     # If a is already a pdarray, do nothing
     if isinstance(a, pdarray):
+        # TODO update if max_bits is set
```
Ethan-DeBandi99, line 210: Just making sure the max bit functionality was not supposed to be included in this PR.

src/BigInt.chpl
```diff
@@ -0,0 +1,88 @@
+module BigInt {
```
Ethan-DeBandi99, line 1: Don't really think it matters, but should we name this `BigIntMsg.chpl`? I am just thinking that we do that with most of our files handling message parsing.

src/CastMsg.chpl
```diff
@@ -109,13 +116,26 @@ module CastMsg {
...
             when (DType.Bool, "str") {
                 return new MsgTuple(castGenSymEntryToString(gse, st, bool), MsgType.NORMAL);
             }
+            when (DType.Bool, "bigint") {
+                return new MsgTuple(castGenSymEntryToBigInt(gse, st, bool), MsgType.NORMAL);
+            }
+            // TODO add bigint to "str"
+            // when (DType.BigInt, "bigint") {
+            //     return new MsgTuple(castGenSymEntry(gse, st, bigint, bigint), MsgType.NORMAL);
+            // }
```
Ethan-DeBandi99: Is there a reason this is marked TODO? Are we waiting on something else?

src/MsgProcessing.chpl
```diff
@@ -339,11 +340,38 @@ module MsgProcessing
                 repMsg = "set %s to %t".format(name, val);
             }
             when (DType.UInt64, DType.UInt64) {
+                // TODO create a .getBigIntValue()
```
Ethan-DeBandi99: Is this referring to a different `.getBigIntValue` than the one created in `src/Message.chpl`?

src/MsgProcessing.chpl
```diff
@@ -339,11 +340,38 @@ module MsgProcessing
...
+                var e = toSymEntry(gEnt,bigint);
+                var val: int = value.getIntValue();
+                e.a = val:bigint;
+                repMsg = "set %s to %t".format(name, val);
+            }
+            when (DType.BigInt, DType.Bool) {
+                var e = toSymEntry(gEnt,bigint);
+                var val: bool = value.getBoolValue();
+                // can't cast from a bool to a bigint, so first cast to int
+                e.a = val:int:bigint;
```
Ethan-DeBandi99, line 370: I would recommend marking this for update in the future like you do at other locations so we do not miss it here.

src/SequenceMsg.chpl
```diff
@@ -55,7 +56,38 @@ module SequenceMsg {
...
+    }
+
+    proc bigIntArangeMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
+        var repMsg: string; // response message
+        var start = msgArgs.get("start").getBigIntValue();
+        var stop = msgArgs.get("stop").getBigIntValue();
+        var stride = msgArgs.get("stride").getBigIntValue();
+        // compute length
+        var len = (stop - start + stride - 1) / stride;
+        // overMemLimit(8*len);
```
Ethan-DeBandi99, line 68: Just noting here that we really need to determine a method for computing memory allocation for `BigInt`. I know that is not the easiest thing to do with how it functions, but we should definitely put in an issue to look into this.

Ethan-DeBandi99 (commented):
Just a few small things to confirm/think about. I would like to do a more thorough review next week, but wanted to get eyes on this before the weekend.

arkouda/dtypes.py
pierce314159: that is right, every other arkouda dtype has an equivalent `np.dtype`. `Bigint` doesn't because numpy gives them dtype object. So we had to roll our own `bigint` type. I'll add a comment

arkouda/pdarraycreation.py
pierce314159: Yeah I am planning on doing this in a future PR

src/BigInt.chpl
pierce314159: Sure, I can make that change

src/CastMsg.chpl
pierce314159: updated

src/MsgProcessing.chpl
pierce314159: updated

src/SequenceMsg.chpl
pierce314159: yeah for sure, this is something I would lean on one of the chapel guys for @ronawho and @bmcdonald3 since it required knowing how many bits are in the type `bigint` which may vary unlike other types

ServerModules.cfg
```diff
@@ -29,6 +29,7 @@ HDF5MultiDim
 ParquetMsg
 EncodingMsg
 GroupByMsg
+BigInt
```
bmcdonald3, line 32: Just noting that this isn't really an "optional" module (e.g., commenting out this will still compile the code since the module is used in a couple of modules that are always built into the server), so it might not make sense to put in the `ServerModules.cfg` file.

src/SequenceMsg.chpl
bmcdonald3: It seems that there is a `sizeInBase` method (https://chapel-lang.org/docs/modules/standard/BigInteger.html#BigInteger.bigint.sizeInBase) for `bigint` that could be used to compute this, but I think it would be pretty expensive to run that over every element in the pdarray every time the size is queryed... Maybe we could calculate it once upon creation and cache that value? I am not enough of a `bigint` expert to know exactly what that size calculation would look like.

pierce314159: I created an issue to track this (#1977)

ServerModules.cfg
bmcdonald3: (FWIW, I might remove the `BigInt` module from this list and then just add a `use BigInt;` in the `ServerDaemon.chpl` file, like we do with other non-optional modules)

pierce314159: Sounds good!

pierce314159: done!

Ethan-DeBandi99 (approved):

arkouda/dtypes.py
```diff
@@ -177,7 +204,7 @@ def isSupportedNumber(num):
 
 def _as_dtype(dt) -> np.dtype:
     if not isinstance(dt, np.dtype):
-        return np.dtype(dt)
+        return dtype(dt)
```
jaketrookman, line 207: Everything looked good, this was the one detail that stood out to me, just making sure you don't need the np. here.

pierce314159: we don't I'm changing it to call the `ak.dtype` now which is defined as:
```python
def dtype(x):
    # we had to create our own bigint type since numpy
    # gives them dtype=object there's no np equivalent
    if (isinstance(x, str) and x == "bigint") or isinstance(x, BigInt):
        return bigint
    else:
        return np.dtype(x)
```
this way we correctly handled `ak.bigint` but everything else calls `np.dtype` like before

jaketrookman: That makes sense, thank you.

jaketrookman (approved):

joshmarshall1 (approved):

ronawho:
> and for a seemingly similar reason, I can't do aggregation because unorderedCopy isn't supported for bigint.

> Because of the above, I'm having to write bigint specific version of a lot of these functions to pass back a distributed array instead of populating a predefined symentry and to do straight assignment instead of aggregation. I'd like @ronawho to verify that these limitations make sense and there's not an easier way around this. Also at some point I'm probably gonna be hitting you up for help with to_list/ndarray

Hmm, this is interesting. unorderedCopy and aggregation is only supported for "trivially copyable types", basically types that can be copied by just copying the bits (i.e. copied with a `memcpy`.) Offhand I'm not sure if the gmp types underlying bigint support being bit-copied.

arkouda/pdarraycreation.py
```diff
@@ -196,13 +207,23 @@ def array(
...
+            if dtype == bigint:
+                tmp = np.array(a)
+                # numpy converts to float for large int (i.e. 2**64 - 1)
+                # avoid this so we don't lose precision
+                if tmp.dtype != np.float_:
+                    a = tmp
+                else:
+                    a = np.array(a, dtype=np.uint)
+            else:
+                a = np.array(a, dtype=dtype)
```
pierce314159, line 226: we should hold off on merging this until after #1984, if people agree with that approach since this is a workaround for the same issue. Once that is merged this section should just be
```python
if dtype is not None and dtype != bigint:
```

pierce314159: I updated this block to account for 1984, it should be good once CI clears

PR #1973, opened 2022-12-19
Closes #1972: Switch to allclose for float comparison in operator test

### LINKED ISSUE #1972: Operator test using `assertEqual` on float comparison
The operator test is currently using `assertEqual` to do float comparison and this can cause some issues when Chapel generates a floating point number slightly different than numpy.

bmcdonald3:
The operator test is currently using `assertEqual` to do float comparison and this can cause some issues when Chapel generates a floating point number slightly different than numpy. This PR switches it to instead use `np.allclose` to avoid this problem.

Additionally, this PR fixes another CI issue that was occurring due to the removal of `np.str()` from the numpy library (see https://numpy.org/devdocs/release/1.24.0-notes.html)

Closes: https://github.com/Bears-R-Us/arkouda/issues/1974

### DIFF
--- a/benchmarks/gather.py
+++ b/benchmarks/gather.py
@@ -100,7 +100,7 @@ def check_correctness(dtype, random, seed):
         elif dtype == "bool":
             npv = np.random.randint(0, 1, Nv, dtype=np.bool)
         elif dtype == "str":
-            npv = np.array([np.str(x) for x in np.random.randint(0, 2**32, Nv)])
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good.

pierce314159 (approved):
Looks good to me! Should we also add `closes #1972` to the PR description?

PR #1982, opened 2022-12-20
Closes #1932: binopvv support for uint and bool types

### LINKED ISSUE #1932: Fix `binopvv` with `uint` and `bool`
While working a different issue, I found that `uint_array + bool_array` is not supported. We should go in and add this support in `binopvv`
```python
>>> u_arr = ak.arange(10, dtype=ak.uint64)
>>> u_arr
array([0 1 2 3 4 5 6 7 8 9])

>>> b_arr = u_arr % 2 == 0
>>> b_arr
array([True False True False True False True False True False])

>>> u_arr + b_arr
RuntimeError: Error: binopvvMsg: uint64 + bool not implemented

>>> ak.arange(10) + b_arr
array([1 1 3 3 5 5 7 7 9 9])
```

jaketrookman:
This PR (closes #1932)

- Adds support to allow for proper binopvv compatibility for uint64 and boolean types
- Fixes resulting type bug
- Adds testing to ensure proper implementation

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -52,7 +52,7 @@ module BinOp
   :returns: (MsgTuple) 
   :throws: `UndefinedSymbolError(name)`
   */
-  proc doBinOpvv (l, r, e, op: string, rname, pn, st) throws {
+  proc doBinOpvv(l, r, e, op: string, rname, pn, st) throws {
     if e.etype == bool {
       // Since we know that the result type is a boolean, we know
       // that it either (1) is an operation between bools or (2) uses
@@ -367,13 +367,13 @@ module BinOp
     } else if ((l.etype == uint && r.etype == bool) || (l.etype == bool && r.etype == uint)) {
       select op {
           when "+" {
-            e.a = l.a:real + r.a:real;
+            e.a = l.a:uint + r.a:uint;
           }
           when "-" {
-            e.a = l.a:real - r.a:real;
+            e.a = l.a:uint - r.a:uint;
           }
           when "*" {
-            e.a = l.a:real * r.a:real;
+            e.a = l.a:uint * r.a:uint;
           }
           otherwise {
             var errorMsg = notImplementedError(pn,l.dtype,op,r.dtype);
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -52,7 +52,7 @@ module BinOp
   :returns: (MsgTuple) 
   :throws: `UndefinedSymbolError(name)`
   */
-  proc doBinOpvv(l, r, e, op: string, rname, pn, st) throws {
+  proc doBinOpvv (l, r, e, op: string, rname, pn, st) throws {
     if e.etype == bool {
       // Since we know that the result type is a boolean, we know
       // that it either (1) is an operation between bools or (2) uses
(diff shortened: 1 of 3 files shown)


src/BinOp.chpl
```diff
@@ -364,6 +364,25 @@ module BinOp
...
+      select op {
+          when "+" {
+            e.a = l.a:real + r.a:real;
+          }
+          when "-" {
+            e.a = l.a:real - r.a:real;
+          }
+          when "*" {
+            e.a = l.a:real * r.a:real;
+          }
```
pierce314159, line 377: why are we opting to cast to `real` instead of `uint`?

src/BinOp.chpl
```diff
@@ -52,7 +52,7 @@ module BinOp
   :returns: (MsgTuple) 
   :throws: `UndefinedSymbolError(name)`
   */
-  proc doBinOpvv(l, r, e, op: string, rname, pn, st) throws {
+  proc doBinOpvv (l, r, e, op: string, rname, pn, st) throws {
```
pierce314159, line 55: seems like we added an extra space here

src/OperatorMsg.chpl
```diff
@@ -187,7 +187,7 @@ module OperatorMsg
               var e = st.addEntry(rname, l.size, bool);
               return doBinOpvv(l, r, e, op, rname, pn, st);
             }
-            var e = st.addEntry(rname, l.size, uint);
```
pierce314159, line 190: same question about having the return type be `real` rather than `uint`

pierce314159 (commented):
I'd like some clarification on why we chose to have the return type be `real` rather than `uint`, but it looks good other than that

src/OperatorMsg.chpl
```diff
@@ -194,10 +194,10 @@ module OperatorMsg
             var l = toSymEntry(left,uint);
             var r = toSymEntry(right,bool);
             if boolOps.contains(op) {
-              var e = st.addEntry(rname, l.size, bool);
+              var e = st.addEntry(rname, l.size, uint);
               return doBinOpvv(l, r, e, op, rname, pn, st);
             }
-            var e = st.addEntry(rname, l.size, uint);
+            var e = st.addEntry(rname, l.size, real);
```
Ethan-DeBandi99, line 200: We should be returning a `uint` here not a `real`.

Ethan-DeBandi99 (changes requested):

src/OperatorMsg.chpl
jaketrookman: Yup, I changed the wrong one. Let me fix it rq

Ethan-DeBandi99 (approved):
Looks good

tests/operator_tests.py
```diff
@@ -300,6 +300,14 @@ def test_invert(self):
         np_uint_inv = ~np.arange(10, dtype=np.uint)
         self.assertListEqual(np_uint_inv.tolist(), inverted.to_list())
 
+    def test_uint_bool_binops(self):
+        # Test fix for issue #1932
+        # Adding support to binopvv to correctly handle uint and bool types
+        ak_uint = ak.arange(10, dtype=ak.uint64)
+        ak_bool = ak_uint % 2 == 0
+        self.assertTrue(np.allclose((ak_uint + ak_bool).to_ndarray(),
+                                    (ak.arange(10) + ak_bool).to_ndarray(), equal_nan=True))
```
pierce314159, line 309: we really only use the `assertTrue(np.allclose` style when things return floats since there's a chance they might differ slightly

For things that return uint/int, we prefer something like
```python
self.assertListEqual((ak_uint + ak_bool).tolist(), (ak.arange(10) + ak_bool).to_list())
```

pierce314159 (commented):
one small comment on the tests, but good otherwise

pierce314159 (approved):
Looks good!

PR #1984, opened 2022-12-20
Closes #1297 and #1983: Change `ak.array` to prefer `uint` over `float` when `>2**63`

### LINKED ISSUE #1297: Python positive Integer list becomes Float if values both GTE and LT 2^63
Using ak.array(x), python list x containing positive integer values <2^64  is converted to pdarray of float if list contains integer values >= 2^63 and also contains integer values < 2^63. 

Not a big issue because using numpy uint64 arrays instead of python list eliminates the problem.

### LINKED ISSUE #1983: ak.DataFrame.from_pandas modifies type from uint to float
When creating an ak.DataFrame using .from_pandas() and with a source column containing large uint values and zero at the same time, the resulting Arkouda column becomes a float. 

```
df = pd.DataFrame({'Test':[2**64-1,0]})
df['Test'].dtype
#results in dtype('uint64')
df2 = ak.DataFrame.from_pandas(df)
df2['Test'].dtype
#results in dtype('float64')
```

Creating the arkouda dataframe explicitly by specifying: 

```
df3 = ak.DataFrame({'Test':ak.array(df['Test'].values)})
df3['Test'].dtype
#results in dtype('uint64')
```

does not cause the conversion. 

Type preservation of the uint is preferred.

pierce314159:
This PR (closes #1297 and closes #1983):
- Changes `ak.array` to default to `uint` when all values are supportedInts and at least one is `> 2**63`
- Adds tests for the reproducers from the two issues

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -202,7 +202,17 @@ def array(
     # If a is not already a numpy.ndarray, convert it
     if not isinstance(a, np.ndarray):
         try:
-            a = np.array(a, dtype=dtype)
+            if dtype is not None:
+                # if the user specified dtype, use that dtype
+                a = np.array(a, dtype=dtype)
+            elif all(isSupportedInt(i) for i in a) and any(i > 2 ** 63 for i in a):
+                # all supportedInt values but some >2**63, default to uint (see #1297)
+                # iterating shouldn't be too expensive since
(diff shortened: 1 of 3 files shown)


Ethan-DeBandi99 (approved):
This looks good to me.

joshmarshall1 (approved):

jaketrookman (approved):
Looks good.

Issue #1919, opened 2022-11-18
Add BigInt support


pierce314159:
Add `BigInt` support in arkouda. We already have use cases that require more than 64 bit that we use multiple `uint64` arrays to simulate, namely `IPv6` support.

Chapel has a [BigInteger](https://chapel-lang.org/docs/modules/standard/BigInteger.html), so hopefully we can leverage that to get most of this functionality for free. We'll need to look into this a bit more to see if it needs to be it's own object or if we can make a datatype for pdarray

Ideally we could specify the number of bits for the pdarray (that might not align with 64 bit boundaries for instance 200 bits instead of 256) to ensure bit shifts work as expected.

We need this to be a native object that can be used in other objects (dataframes, segarray, IPv6, ...)

pierce314159:
I think we can have an optional `max_bits` field and do some masking if it is set to get the correct behavior with bit shifts. This is more of a proof of concept than an implementation suggestion because i think we could prob adapt our bit mask based on the shift amount and perhaps save some work

```chapel
use BigInteger;

var bi = "1205203533194242706656471569255871951891652245337094626476019": bigint;
var max_bits = 200;
var bit_mask = new bigint(1);
bit_mask <<= max_bits;
bit_mask -= 1;
writeln("our num:\n", bi, "\n");
writeln("our mask:\n", bit_mask, "\n");

writeln("shift by 1 w/o mask:\n", bi<<1, "\n");
writeln("shift by 1 w/ mask:\n", (bi<<1) & bit_mask, "\n");
writeln("shift by 2 w/o mask:\n", bi<<2, "\n");
writeln("shift by 2 w/ mask:\n", (bi<<2) & bit_mask);
```
```bash
our num:
1205203533194242706656471569255871951891652245337094626476019

our mask:
1606938044258990275541962092341162602522202993782792835301375

shift by 1 w/o mask:
2410407066388485413312943138511743903783304490674189252952038

shift by 1 w/ mask:
803469022129495137770981046170581301261101496891396417650662

shift by 2 w/o mask:
4820814132776970826625886277023487807566608981348378505904076

shift by 2 w/ mask:
1606938044258990275541962092341162602522202993782792835301324
```

comparing this with python (with the help of `bin`) we see this is the behavior we want
```python
>>> bi = 1205203533194242706656471569255871951891652245337094626476019
>>>  bin(bi)
'0b10111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
[...]

PR #1990, opened 2022-12-21
Closes #1989: `bigint` indexing

### LINKED ISSUE #1989: bigint indexing
Add indexing to `bigint` pdarrays

pierce314159:
This PR (closes #1989) adds indexing for `bigint` pdarrays and tests

### DIFF
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -12,6 +12,7 @@ module IndexingMsg
     use MultiTypeSymbolTable;
 
     use CommAggregation;
+    use BigInteger;
 
     use FileIO;
     use List;
@@ -248,6 +249,12 @@ module IndexingMsg
                  imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
                  return new MsgTuple(repMsg, MsgType.NORMAL); 
              }
+             when (DType.BigInt) {
+                 var e = toSymEntry(gEnt,bigint);
+                 repMsg = "item %s %t".format(dtype2str(e.dtype),e.a[idx]);
+                 imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
+                 return new MsgTuple(repMsg, MsgType.NORMAL);
+             }
              otherwise {
                  var errorMsg = notImplementedError(pn,dtype2str(gEnt.dtype));
                  imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
@@ -311,6 +318,21 @@ module IndexingMsg
             }
             when (DType.Bool) {
                 return sliceHelper(bool);
+            }
+            when (DType.BigInt) {
+                var e = toSymEntry(gEnt,bigint);
+                var tmp = makeDistArray(slice.size, bigint);
+                ref ea = e.a;
+                // TODO figure out a way to copy more efficiently attempting
+                // to use an aggregator runs into issues with unordered copy
+                forall (elt,j) in zip(tmp, slice) {
+                    elt = ea[j];
+                }
+                var a = st.addEntry(rname, new shared SymEntry(tmp));
+                var repMsg = "created " + st.attrib(rname);
+                imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
+                return new MsgTuple(repMsg, MsgType.NORMAL);
+
             }
             otherwise {
                 var errorMsg = notImplementedError(pn,dtype2str(gEnt.dtype));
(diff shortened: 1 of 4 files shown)


tests/pdarray_creation_test.py
```diff
@@ -62,23 +61,34 @@ def test_bigint_creation(self):
             self.assertIsInstance(pda, ak.pdarray)
             self.assertEqual(10, len(pda))
             self.assertEqual(ak.bigint, pda.dtype)
+            self.assertEqual(pda[-1], bi + 9)
 
         # test array and arange infer dtype
         arr = ak.array([bi, bi + 1, bi + 2, bi + 3, bi + 4])
         ara = ak.arange(bi, bi + 5)
+        # TODO add .to_list and .to_ndarray functionality
```
Ethan-DeBandi99, line 69: Do we have an issue for this? I am guessing it is not supported for `dtype=bigint` since NumPy sets the type as object.

src/IndexingMsg.chpl
```diff
@@ -312,6 +319,21 @@ module IndexingMsg
...
             }
+            when (DType.BigInt) {
+                var e = toSymEntry(gEnt,bigint);
+                var tmp = makeDistArray(slice.size, bigint);
+                ref ea = e.a;
+                // TODO figure out a way to copy more efficiently attempting
+                // to use an aggregator runs into issues with unordered copy
+                forall (elt,j) in zip(tmp, slice) {
+                    elt = ea[j];
+                }
```
Ethan-DeBandi99, line 330: Based on @ronawho mentioning that array slicing should be pretty efficient at this point, could we just do something like `tmp = e.a[slice]`? Or is there something with the type BigInt that makes this not efficient?

src/IndexingMsg.chpl
```diff
@@ -480,6 +502,107 @@ module IndexingMsg
...
+                if ivMax >= e.size {
+                    var errorMsg = "Error: %s: OOBindex %i > %i".format(pn,ivMin,e.size-1);
+                    imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+                    return new MsgTuple(errorMsg,MsgType.ERROR);
+                }
+                var tmp = makeDistArray(iv.size, bigint);
+                ref a2 = e.a;
+                ref iva = iv.a;
+                forall (t,idx) in zip(tmp,iva) {
+                    // TODO find a way to use aggregators
```
Ethan-DeBandi99, line 530: Should we add an issue to research this?

Ethan-DeBandi99 (commented):
Looks good overall, just a few minor questions/ideas.

src/IndexingMsg.chpl
ronawho: This isn't just straight slice assignment like`tmp = slice`, but random indexing into `ea` based on values in slice. I think aggregators would be appropriate here. @pierce314159 performance without aggregation is going to be quite poor and I'm really not sure if bigint can support aggregation (or really being bit-copied). Can you open an issue for me to look into this after the new year?

Ethan-DeBandi99: Makes sense. That is why I wanted to check. My memory is a bit crowded these days.

tests/pdarray_creation_test.py
pierce314159: it's being tracked in the `Add BigInt support` parent issue #1919

What you're saying is true, but I think we can special case it and just check for `dtype == bigint` and then select object. What really makes this difficult is `to_ndarray` uses `bytesToSymEntry` and it's not obvious to me that this would play nice with chapel `bigint` type. I'm planning on looping in elliot when I work this

https://github.com/Bears-R-Us/arkouda/blob/master/src/GenSymIO.chpl#L53-L60

src/IndexingMsg.chpl
pierce314159: > This isn't just straight slice assignment like`tmp = slice`, but random indexing into `ea` based on values in slice. I think aggregators would be appropriate here. @pierce314159 performance without aggregation is going to be quite poor and I'm really not sure if bigint can support aggregation (or really being bit-copied). Can you open an issue for me to look into this after the new year?

for sure! I'll open the issue now, thanks!

src/IndexingMsg.chpl
pierce314159: Added https://github.com/Bears-R-Us/arkouda/issues/1994

tests/pdarray_creation_test.py
Ethan-DeBandi99: Sweet thanks!

Ethan-DeBandi99 (approved):

tests/pdarray_creation_test.py
```diff
@@ -62,23 +61,34 @@ def test_bigint_creation(self):
...
+        for i in range(5):
+            self.assertEqual(arr[i], bi + i)
+            self.assertEqual(ara[i], bi + i)
 
         # test that max_bits being set results in a mod
         pda = ak.array([bi, bi + 1, bi + 2, bi + 3, bi + 4], max_bits=200)
+        for i in range(5):
+            self.assertEqual(i, pda[i])
```
jaketrookman, line 77: Should this be self.asserEqual(bi + i, pda[i])? Does bi roll back down to 0 in this case?

pierce314159: ah so the reason why it rolls over in this case is because `max_bits = 200`, so  since `bi = 2**200` is equivalent to `0`

jaketrookman: Perfect, thank you

jaketrookman (approved):
Looks good to me.

joshmarshall1 (approved):

PR #1993, opened 2022-12-22
Closes #1925 - Remove Deprecated IO Functions

### LINKED ISSUE #1925: Remove deprecated IO functions
Following the separation of HDF5 and Parquet operations, several generic functions were deprecated. These deprecated functions will need to be removed after at least one release cycle.
- pdarrayclass.py
    - `save`
    - `save_parquet`
    - `save_hdf`
- array_view.py
    - `save`
- strings.py
    - `save`
    - `save_parquet`
    - `save_hdf`
- categorical.py
    - `save`
- index.py
    - `save`
- segarray.py
    - `save`
- io.py
    - `save_all`
    - `write_hdf5_multi_dim`
    - `read_hdf5_multi_dim`
    - `read`
- deprecated_io_test.py

Ethan-DeBandi99:
Closes #1925 

- Removes Client IO functions that were deprecated due to the separation of Parquet and HDF5. 
- Removes Server Code supporting reads of files where file type is inferred, `readAny`. 
- Updates code that was overlooked during deprecation.
- Updates `ak.read` to check the filetype and then call the appropriate function. This function is useful if you are unsure of the file type.


Ethan-DeBandi99:
Converting to draft to make some alterations with reads based on conversation with @pierce314159

jaketrookman (approved):
Looks good.

arkouda/io.py
```diff
@@ -1229,194 +1215,30 @@ def read(
     Read with file Extension
     >>> x = ak.read('path/name_prefix.h5') # load HDF5 - processing determines file type not extension
     Read without file Extension
```
pierce314159, line 1217: not super important, but maybe we should remove the line `Read without file extension`, since that's no longer supported

pierce314159 (approved):
Looks good

joshmarshall1 (approved):

arkouda/io.py
pierce314159: I'm gonna go ahead and merge since the actual code in the example has been updated, I don't think this one line comment is worth holding this up

PR #2004, opened 2023-01-03
Closes #1991 - Additional Compression Support for Parquet

### LINKED ISSUE #1991: Additional Compression Support for Parquet
Look into and update our Parquet workflow to allow for varying compression types on write. Currently, we are only supporting writing with Snappy compression.

Additionally, validate that we are able to read all compression forms. This includes when the columns themselves are compressed not the entire file. **EDIT** - This has been verified as fully functional for any compression type for the columns.

Ethan-DeBandi99:
Closes #1991

Adds support for GZip, Brotli, ZSTD, and LZ4 compression for Parquet files. Snappy compression is still supported. Reading all compression types was supported, but now Arkouda supports writing Parquet files with all compressions supported by Parquet. 

Replaces the `compressed` parameter in the `to_parquet` functions with `compression`. This optional parameter defaults to None, but can be set to `snappy`, `gzip`, `brotli`, `zstd`, or `lz4` depending upon the desired compression. 

This PR also adds testing to ensure read/write functionality with all compression types. 

**Example Usage**
```python
>> a = ak.arange(150)
>> a
array([0 1 2 ... 147 148 149])

>> // write the file using Brotli Compression
>> a.to_parquet('file_path/filename.parquet', dataset='dset_name', compression='brotli')

//read the file
>> rd_a = a.read_parquet('file_path/filename.parquet', 'dset_name')
>> rda_a
array([0 1 2 ... 147 148 149])
```

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -876,7 +876,7 @@ def to_parquet(
     # append or single column use the old logic
     if mode.lower() == "append" or len(pdarrays) == 1:
         for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
-            arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compression)
+            arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compression=compression)
     else:
         print(
             cast(
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -697,7 +697,9 @@ def import_data(read_path: str, write_file: str = None, return_obj: bool = True,
     df = DataFrame(df_def)
 
     if write_file:
-        df.to_hdf(write_file, index=index) if filetype == "HDF5" else df.to_parquet(write_file, index=index)
+        df.to_hdf(write_file, index=index) if filetype == "HDF5" else df.to_parquet(
+            write_file, index=index
+        )
 
     if return_obj:
         return df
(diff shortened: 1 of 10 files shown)


arkouda/io.py
```diff
@@ -873,7 +876,7 @@ def to_parquet(
     # append or single column use the old logic
     if mode.lower() == "append" or len(pdarrays) == 1:
         for arr, name in zip(pdarrays, cast(List[str], datasetNames)):
-            arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compressed)
+            arr.to_parquet(prefix_path=prefix_path, dataset=name, mode=mode, compressed=compression)
```
pierce314159, line 879: should this be `compression=compression`? I'm a bit confused why it's not complaining about using a keyword arg that shouldn't exist anymore

src/ArrowFunctions.cpp
```diff
@@ -685,7 +737,7 @@ int cpp_createEmptyParquetFile(const char* filename, const char* dsetname, int64
 
 int cpp_appendColumnToParquet(const char* filename, void* chpl_arr,
                               const char* dsetname, int64_t numelems,
-                              int64_t dtype, bool compressed,
+                              int64_t dtype, int64_t compression,
```
pierce314159, line 740: does this function ever use the compression argument?

pierce314159 (commented):
everything looks good outside of 2 small comments

src/ArrowFunctions.cpp
Ethan-DeBandi99: This is due to how it is called from the message parsing. I believe we are deprecating in the future, so did not want to alter too much.

arkouda/io.py
Ethan-DeBandi99: Updated.

pierce314159 (approved):

joshmarshall1 (approved):
Looks good

PR #2008, opened 2023-01-03
enable asynchronous Arkouda client requests

hokiegeek2:
Hey guys,

This is a draft PR that covers the initial proof-of-concept for enabling asynchronous client-server comms in Arkouda utilizing asyncio.

There is currently an issue w/ pytest failures that I am investigating, but the code itself is working.

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -1,5 +1,8 @@
 import json
 import os
+import asyncio
+from asyncio.exceptions import CancelledError
+import threading
 import warnings
 from enum import Enum
 from typing import Dict, List, Mapping, Optional, Tuple, Union, cast
@@ -83,6 +86,27 @@ def _mem_get_factor(unit: str) -> int:
 logger = getArkoudaLogger(name="Arkouda Client")
 clientLogger = getArkoudaLogger(name="Arkouda User Logger", logFormat="%(message)s")
 
+class RequestMode(Enum):
+    """
+    The RequestMode enum provides controlled vocabulary indicating whether the
+    Arkouda client is submitting a request via asyncio (ASYNC) or via standard, 
+    synchronous (SYNC) flow 
+    """
+
+    ASYNC = "ASYNC"
+    SYNC = "SYNC"
+
+    def __str__(self) -> str:
+        """
+        Overridden method returns value.
+        """
+        return self.value
+
+    def __repr__(self) -> str:
+        """
+        Overridden method returns value.
+        """
+        return self.value
 
 class ClientMode(Enum):
     """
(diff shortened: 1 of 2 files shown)


hokiegeek2:
example of async connect w/ delay:

https://user-images.githubusercontent.com/10785153/210417083-6fa37817-286c-43a3-ab24-a7d9c8f01280.mp4

example of async ones request w/ delay:

https://user-images.githubusercontent.com/10785153/210417152-705698c6-c463-4e83-84bb-88dc1b8f212c.mp4

hokiegeek2:
just have to fix flake8 errors:

image

hokiegeek2:
all tests are now working: 
image

pierce314159 (commented):
This super neat!! I don't know much about async io but nothing jumped out at me during my once over. Nice work!!! 🎉

arkouda/client.py
```diff
@@ -438,7 +646,16 @@ def _send_string_message(
 
     logger.debug(f"sending message {message}")
 
-    socket.send_string(json.dumps(message.asdict()))
+    loop = get_event_loop()
+
+    def asyncEligible(cmd: str) -> bool:
+        return cmd not in {'delete', 'connect', 'getconfig'}
```
Ethan-DeBandi99, line 652: Moving forward, I am assuming we will have more commands if not all supported. Would we need to list every command here or would this be removed if we are supporting all commands?

Ethan-DeBandi99 (commented):
Overall this looks really good. Just one note on documentation and one question regarding future maintenance.

arkouda/client.py
pierce314159: I might be wrong but I think this is saying every command except `delete`, `connect`, and `getconfig` is async eligible (i.e. will return True)

README.md
```diff
@@ -402,6 +403,29 @@ Integrating Arkouda with cloud environments enables users to access Arkouda from
...
+```
+
+### Exiting the Python shell
+
+As of 01022023, exiting the Python shell in ASYNC request mode requires the following command:
+
+```
+ak.exit()
+```
+
```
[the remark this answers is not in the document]
hokiegeek2: @Ethan-DeBandi99 yeah, good point, should be at top-level

arkouda/client.py
hokiegeek2: @Ethan-DeBandi99 ah, sorry, as @pierce314159 this checks to see if a method can be run in async mode. The reasons for the three are as follows:

1. delete--since delete invokes __del__ on pdarray, this cannot be run async b/c __del__ is also invoked on Arkouda shell exit() and fails due to the resulting asyncio failing due to modules being deleted
2. connect--this is because Python client connect method is itself async-enabled
3. getconfig--this is invoked within the async Python client connect method, which, again, is itself async-enabled

hokiegeek2:
Here's an example of the "fire-and-forget" approach where the ak.connect() method is invoked and returns immediately. Once the request is actually returned, the user is prompted with the return message. This approach is different and handy in that the user can submit a request, continue working on something else, and then return to the workflow that spawned the request as shown below:


https://user-images.githubusercontent.com/10785153/210675379-cc7cc351-b4bf-47a5-8685-f025289b3383.mp4

jaketrookman (approved):
Looks good

PR #2012, opened 2023-01-04
Closes #2005 - Parquet Read Error

### LINKED ISSUE #2005: Parquet Read Error
Notes on reproducing the issue:

- gzipped column
- Column contains only `NaN` values. Suspected to be `numpy.nan`
- dtype when the dataset is read with pandas is `float64`

Ethan-DeBandi99:
Closes #2005 

Updates reading of Parquet files to allow for `NaN` values in `double` and `float` columns. Adds `cmath` to allow setting `NaN` from C++. 

Adds testing to ensure that Arkouda is able to read Parquet files written by pandas containing columns that:
- Contain all `NaN` values
- Contain some `NaN` values


src/ArrowFunctions.cpp
```diff
@@ -333,29 +333,35 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
...
         while (reader->HasNext() && i < numElems) {
-          if((numElems - i) < batchSize)
-            batchSize = numElems - i;
+          float value;
           // Can't read directly into chpl_ptr because it is a double
-          (void)reader->ReadBatch(batchSize, nullptr, nullptr, tmpArr, &values_read);
-          for (int64_t j = 0; j < values_read; j++)
-            chpl_ptr[i+j] = (double)tmpArr[j];
-          i+=values_read;
+          (void)reader->ReadBatch(1, &definition_level, nullptr, &value, &values_read);
```
bmcdonald3, line 340: Haven't had a chance to do a full review, but just noting at a quick glance that this is going to result in a significant performance hit to only read one value at a time as opposed to `batchSize` at a time.

Ethan-DeBandi99: Discussed this with @bmcdonald3. After reviewing our current options, there is not an obvious alternative for this to work properly. We are going to continue with this resolution and add an additional issue to pursue this in a more efficient manor.

joshmarshall1 (approved):

pierce314159 (approved):

PR #2018, opened 2023-01-04
Closes #2013: Create local copies of block size for bigint modulus

### LINKED ISSUE #2013: Gasnet error in 1999
PR #1999 is resulting a gasnet error, specifically `var low = tmp % block_size;` in `bigint_to_uint_arrays`.

bmcdonald3:
A Chapel bug in the bigint module was causing a crash in distributed settings due to an attempt to pass remote values to C. This should be fixed by https://github.com/chapel-lang/chapel/pull/21321 in the future (it is still a work in progress), but we still need a workaround for Arkouda until that fix is included in the oldest supported Chapel.

closes #2013

### DIFF
--- a/src/BigIntMsg.chpl
+++ b/src/BigIntMsg.chpl
@@ -63,17 +63,17 @@ module BigIntMsg {
                 var block_size = 1:bigint;
                 block_size <<= 64;
                 while || reduce (tmp!=0) {
-                  on tmp.locale {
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Looks good! Thanks for getting this workaround in so quickly!

joshmarshall1 (approved):

PR #2022, opened 2023-01-05
Closes #2021 - Compression Fixes for Benchmarks

### LINKED ISSUE #2021: Compression Fixes for Benchmarks
When adding selectable compression types to Arkouda's Parquet support, the benchmarks were not updated.

Update the benchmarks to work with the new configuration.

Ethan-DeBandi99:
Closes #2021 

Updates benchmarks to set `compression='snappy'` if `compressed=True`. Otherwise, it is set to `None`. This will allow the benchmark to handle compression as it was before it was made selectable. 

A separate issue/PR will be added to update the benchmark to run against all available compression types.


pierce314159 (approved):
Looks good!

PR #2055, opened 2023-01-11
Closes #2054: Fix iconv encoding for multi-byte characters and reenable tests

### LINKED ISSUE #2054: Fix failing IDNA tests
The IDNA tests have been disabled due to a failure and should be fixed and reenabled.

bmcdonald3:
In the initial encoding implementation, the output buffer for
calculating the buffer length was a `bytes` buffer and that
was causing issues when an output character was a multi-byte
UTF-8 character. To fix this, the `bytes` buffer has been
switched to a `uint-8` array.


Ethan-DeBandi99 (approved):
Looks good

pierce314159 (approved):

Issue #1994, opened 2022-12-22
Aggregation on `bigint`


pierce314159:
When adding functionality for `bigint` pdarrays, we are running into cases (see #1990) where we'd like to do aggregation. This doesn't work out of the box and it's unclear if it's possible to aggregate or even bit-copy `bigint`. This issue will track the investigation of adding this functionality

ronawho:
Haven't thought about this deeply, but there are two main assumptions unorderedCopy/Aggregation currently make:
 - The underlying type is trivially copyable (bits can be `memcpy`'d without any additional fixups)
 - The datatype is uniform in size. This is not generally true for bigints since something like `2**64` will require more limbs/bits than `2**1`, but if you're using bigints as a fixed with integer, maybe we can ensure the underlying gmp type has the same number of limbs.

pierce314159:
> * The datatype is uniform in size. This is not generally true for bigints since something like `2**64` will require more limbs/bits than `2**1`, but if you're using bigints as a fixed with integer, maybe we can ensure the underlying gmp type has the same number of limbs.

I don't think the users are expecting us to treat bigints as fixed width integers, so if we are able to find a way to make this more efficient without making that restriction, that would be ideal. I can run it by them and see if fixed width would suit their needs

PR #2064, opened 2023-01-19
Add aggregator support for `bigint`

ronawho:
Add aggregators that support bigint.

Plain old integer DstAggregators locally aggregate the target address and the inline value by storing a tuple of `(c_ptr(int), int)` but it's not as simple with variable sized bigints. It's also not as easy to extract the raw value since bigints have metadata stored inline, but the values stored off in the heap. To enable aggregation, the buffers are no longer typed and are just raw bytes that we'll interpret. We now locally aggregate the target address, the 8-byte sign+size (int64 that stores the number of limbs and the sign of the limbs), and the variables sized limbs. This uses internal functions to expose the internal sign+size and limbs. With this we have our target address and "value" stored inline and can get data to the target side like normal. On the target side we then extract the address and size, and then resize the target bigint, copy the limbs over, and update the target sign+size.

SrcAggregator is a little more complicated since we don't know how big our src values are until we're on the remote node. Because of this we can't pre-allocate space for all our elements, so instead we just aggregate addresses locally and when those buffers are full we fill up value buffers as much as we can and just loop until we've processed all values.

Testing with single limb (64-bit value) bigints, I see aggregation performance that's within 2x of plain old int(64) aggregation. This is around what I'd expect since we have to aggregate an extra 8 bytes for the sign_size and there's other overheads with real location and extracting limbs. I don't think there's much more performance we can squeeze out of this, but there is certainly a lot of code cleanup we can do and a lot more testing I'd like to do. This is a prototype currently, but I wanted to capture what I have so far.

Implementation TODOs:
 - Fix support for bigints with more limbs than we can store
 - Fix support for mismatched bigint record and mpz locale.id
 - Tune buffer size (currently sized to fit the same number of elements as plain old int64 aggregation to do comm comparisons)
 - See if we can use public functions (`mpz_limbs_write`, `mpz_limbs_finish`) added in GMP 6 -- https://gmplib.org/manual/Integer-Special-Functions

Testing TODOs:
 - check negative numbers (gmp stores sign as part of size)
 - check massive numbers (more limbs than we have buffer space for, broken currently)
 - check variable sized limbs (only tested with single limb so far)
 - check performance with larger limbs (single limb is ~2x slower)
 - check with bigint record wrapper and heap-allocated mpz on different nodes (broken currently)
 - check with limb sizes that will grow and shrink target bigint
 - check with aggregating 0 values (no limbs), should work with current strategy, but would require special case with `mpz_limbs_write`
 
 Closes https://github.com/Bears-R-Us/arkouda/issues/1994

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -559,7 +559,7 @@ module CommAggregation {
     record SrcAggregatorBigint {
       type aggType = c_ptr(bigint);
       const bufferSize = srcBuffSize;
-      const uintBufferSize = srcBuffSize * minInlineBigintSize();
+      const uintBufferSize = srcBuffSize * (c_sizeof(mp_size_t) + c_sizeof(mp_limb_t)): int;
       const myLocaleSpace = 0..<numLocales;
       var lastLocale: int;
       var opsUntilYield = yieldFrequency;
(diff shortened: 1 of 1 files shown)


ronawho:
@pierce314159 I'd like some time to clean this up and add some testing. My assumption is that you're not in any particular rush for this and that having the interface added is enough to help you make progress in the short term. I'd probably naturally get back to this in ~2 weeks or so, but let me know if you'd like to see this done sooner.

PR #2071, opened 2023-01-23
Reformat 1-element array literals to use a trailing comma

bradcray:
Chapel has recently changed 1-element array literals to require a trailing comma, deprecating the previous syntax (due to user confusion, and to unify syntax with 1-tuples—see https://github.com/chapel-lang/chapel/issues/21092).  Here, I've updated 1-element arrays in Arkouda to use this trailing comma syntax.  Because trailing commas were supported prior to this deprecation, this is not a breaking change for older versions of Chapel (since Arkouda's inception).

I also did a second version of this PR that attempts to remove some 1-element arrays that were not necessary, and felt like overkill in terms of compilation/execution complexity.  See #2072 for details.  Only one of the two should be merged.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -113,7 +113,7 @@ module HDF5Msg {
         var extension: string;
         (prefix,extension) = getFileMetadata(filename);
 
-        var filenames: [0..#1] string = ["%s%s".format(prefix, extension)];
+        var filenames: [0..#1] string = ["%s%s".format(prefix, extension), ];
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
This looks good to me! So does the other version of this PR but that's modifying HDF5 code, which is @Ethan-DeBandi99 wheelhouse. So ill defer to him on which version of this PR to accept/merge

Ethan-DeBandi99:
Converting to draft to avoid accidental merge of competing PRs.

PR #2072, opened 2023-01-23
Reformat 1-element arrays to avoid trailing commas and avoid some unnecessary arrays

bradcray:
Chapel has recently changed 1-element array literals to require a trailing comma, deprecating the previous syntax (due to user confusion, and to unify syntax with 1-tuples—see https://github.com/chapel-lang/chapel/issues/21092).  Here, I've updated 1-element arrays in Arkouda to use this trailing comma syntax.  Because trailing commas were supported prior to this deprecation, this is not a breaking change for older versions of Chapel (since Arkouda's inception).

This is a variation on PR #2071, where I realized that some of the 1-element arrays in the HDF5 code are unnecessary and overkill, so rewrote the code to avoid them.  Creating the arrays only adds (a very modest amount) of compile-time and execution-time overhead.  Only one of these two PRs should be merged.  I have not tested these code changes beyond building the server, and am hoping to rely on the CI for that.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -113,7 +113,7 @@ module HDF5Msg {
         var extension: string;
         (prefix,extension) = getFileMetadata(filename);
 
-        var filenames: [0..#1] string = ["%s%s".format(prefix, extension)];
+        var filenames: [0..#1] string = ["%s%s".format(prefix, extension), ];
         var matchingFilenames = glob("%s*%s".format(prefix, extension));
 
         const f = filenames[0];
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
This looks good to me! 

I don't see any issue with the removal of the arrays, but that's modifying HDF5 code, which I'm def not an expert on (that is @Ethan-DeBandi99 wheelhouse). So I'll defer to him on which version of this PR to accept/merge

Ethan-DeBandi99:
Converting to draft to avoid accidental merge of competing PRs.

Ethan-DeBandi99 (approved):
This looks good to me. I had modeled this off existing code and did not think to refactor the single element arrays.

bradcray:
@pierce314159 and @Ethan-DeBandi99 : Since merging this PR, we've gotten cold feet about requiring the trailing comma for one-element arrays, so have backed out the warning that led to this PR.  I think the changes to src/HDF5Msg.chpl made here are still worthwhile, in that they seem more straightforward and less roundabout to me than the old style (i.e., had I noticed them earlier, I'd probably have made the same changes, and it was just the new warning that alerted me to them). 
 
 
 The changes made in src/RegistrationMsg.chpl are completely unnecessary now, though also fine (which is to say, trailing commas were optional before and will remain optional going forward, so nothing is worse).  That said, if you prefer the comma-less form, stylistically (or for any other reason at all), let me know and I will set up a PR to back out those changes.

Thanks!

PR #2074, opened 2023-01-24
Add some function calls so that iconv check works properly

bmcdonald3:
Since the existing `checkIconv.chpl` file only uses an extern variable, this was causing it to pass unexpectedly in some environments when the header files were included, but not linked. This PR adds some function calls so that it should always fail in the case of missing iconv.

### DIFF
--- a/dep/checkIconv.chpl
+++ b/dep/checkIconv.chpl
@@ -1,5 +1,8 @@
 use iconv;
 
 proc main() {
+  var enc = "UTF-8";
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
@bmcdonald3 - good catch. I agree this should catch any outliers.

joshmarshall1 (approved):

pierce314159 (approved):

PR #2075, opened 2023-01-24
Closes #2070: Improve Makefile errors when dependencies aren't found

### LINKED ISSUE #2070: Display a nice message when an Arkouda dependency is missing
Recently, when more Chapel devs were ramping up with Arkouda, the dependency missing errors were a bit confusing, displaying walls of text about the C code that can be pretty indecipherable.

It seems like we could improve that to check the error message in the `Makefile` or something and then display a more helpful message instead like `Looks like you are missing <this-dependency>, see <doc-link> for more info` or something like that.

bmcdonald3:
This PR adds a message when the Arkouda `make check-deps` steps fails pointing users to documentation for how to install the dependencies properly. Also, the compilation commands are suppressed here so that there won't be so much clutter on the screen during the compilation.

closes #2070

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,21 +206,21 @@ endif
 ZMQ_CHECK = $(DEP_INSTALL_DIR)/checkZMQ.chpl
 check-zmq: $(ZMQ_CHECK)
 	@echo "Checking for ZMQ"
-	$(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@
+	@$(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@ && ([ $$? -eq 0 ] && echo "Success compiling program") || echo "\nERROR: Please ensure that dependencies have been installed correctly (see -> https://github.com/Bears-R-Us/arkouda/blob/master/pydoc/setup/BUILD.md)\n"
 	$(DEP_INSTALL_DIR)/$@ -nl 1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -206,43 +206,43 @@ endif
 ZMQ_CHECK = $(DEP_INSTALL_DIR)/checkZMQ.chpl
 check-zmq: $(ZMQ_CHECK)
 	@echo "Checking for ZMQ"
-	$(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@
+	@$(CHPL) $(CHPL_FLAGS) $(ARKOUDA_COMPAT_MODULES) $< -o $(DEP_INSTALL_DIR)/$@ && ([ $$? -eq 0 ] && echo "Success compiling program") || echo "\nERROR: Please ensure that dependencies have been installed correctly (see -> https://github.com/Bears-R-Us/arkouda/blob/master/pydoc/setup/BUILD.md)\n"
```
bradcray, line 209: Thanks for taking this on, @bmcdonald3 !

I'm not at all a bash expert, but I probably would've just written this `|| echo "\nERROR: Please ensure that dependencies have been installed correctly (see -> https://github.com/Bears-R-Us/arkouda/blob/master/pydoc/setup/BUILD.md)\n"`  Specifically, I don't think a success message is necessary and the `[ $$? -eq 0]` feels redundant with what the && / || are doing inherently (?).

Ethan-DeBandi99: I do agree with @bradcray here, I do not know that we need the `Success compiling program` message. However, I do not believe that it will hurt anything if we decide to keep it.

Ethan-DeBandi99 (approved):
While I agree the success message is not necessary, I do not have any issue with maintaining it. I think this will be incredibly useful especially for anyone new to Arkouda.

PR #2078, opened 2023-01-25
Separate setting optimization level and enabling runtime checks

### LINKED ISSUE #2066: Figure out right set of quick vs debug vs optimized build flags
Arkouda builds fully optimized with `--fast` (no runtime checks, all chpl optimizations, all backend optimization) by default and I think that makes sense for end users. However, for developers that results in really long compilation times.

Because of that `ARKOUDA_DEVELOPER` was added in #172 and #306 to enable runtime checks and disable some backend optimizations. This improves compilation time somewhat and adds useful runtime checks, but it can still be somewhat slow.

`ARKOUDA_QUICK_COMPILE` was then added in #358 to compile as quick as possible. This disables runtime checks, some chpl optimizations and all backend optimizations.

I tend to use `ARKOUDA_QUICK_COMPILE` by default and switch to `ARKOUDA_DEVELOPER` when I want runtime checks, but others prefer having the runtime checks on by default. This makes me think we should split runtime checks out from optimization level so we can get checks with faster compilers.

Maybe something like `ARKOUDA_RUNTIME_CHECKS` to select whether runtime checks are enabled and sticking with 
`ARKOUDA_QUICK_COMPILE` to select between optimized and unoptimized builds.

Another middle ground that I sometimes want is full chpl optimizations and whatever backend optimizations won't mess with using a debugger (this was originally what `ARKOUDA_QUICK_COMPILE` was meant for with enabling checks and only `-O1` for the backend, but probably I'd want `-Og -g` anymore.) It's pretty rare that I actually want that and when I do I tend to just modify the Makefile when I do, but this may be a useful target to consider for others.

Anyways, left to my own devices for now I'd add `ARKOUDA_RUNTIME_CHECKS` and either remove `ARKOUDA_DEVELOPER` or have it be a meta-flag that maps to `ARKOUDA_QUICK_COMPILE=true ARKOUDA_RUNTIME_CHECKS=true`.

ronawho:
Previously, `ARKOUDA_QUICK_COMPILE` disabled all optimizations and disable runtime checks and `ARKOUDA_DEVELOPER` disabled some optimizations and enabled runtime checks.

However, there are times where you might want runtime checks without optimizations or maybe want runtime checks with full optimizations. This creates a new `ARKOUDA_RUNTIME_CHECKS` var that can independently toggle runtime checks to support that. It also changes `ARKOUDA_DEVELOPER` to map to `ARKOUDA_QUICK_COMPILE=true ARKOUDA_RUNTIME_CHECKS=true` since that is a more desired default for developers (speeds up compilation from previous default but still leaves runtime checks useful for debugging on.)

Some timings that Ben captured for me with Chapel 1.29 and Arkouda v2023.01.11 (~2018 x86 macbook):

| config      | default | Old AK_DEV | AK_QUICK | New AK_DEV |
| ----------- | ------: | ---------: | -------: | ---------: |
| COMM=none   |  761s   |  540s      |  411s    |  461s      |
| COMM=gasnet | 2245s   | 2019s      | 1255s    | 1423s      |

Resolves #2066

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,13 +18,22 @@ CHPL := chpl
 CHPL_FLAGS += --ccflags="-DH5_USE_110_API"
 
 CHPL_DEBUG_FLAGS += --print-passes
+
 ifdef ARKOUDA_DEVELOPER
-CHPL_FLAGS += --ccflags="-O1"
-else ifdef ARKOUDA_QUICK_COMPILE
+ARKOUDA_QUICK_COMPILE = true
+ARKOUDA_RUNTIME_CHECKS = true
+endif
+
+ifdef ARKOUDA_QUICK_COMPILE
 CHPL_FLAGS += --no-checks --no-loop-invariant-code-motion --no-fast-followers --ccflags="-O0"
 else
 CHPL_FLAGS += --fast
 endif
+
+ifdef ARKOUDA_RUNTIME_CHECKS
+CHPL_FLAGS += --checks
+endif
+
 CHPL_FLAGS += -smemTrack=true -smemThreshold=1048576
 
 # We have seen segfaults with cache remote at some node counts


Ethan-DeBandi99 (approved):
This looks good to me based on our discussions earlier.

pierce314159 (approved):
Looks good

PR #2079, opened 2023-01-25
Fixes #2076: Fixes integer overflow in `groupby.mean`

### LINKED ISSUE #2076: Integer overflow in `groupby.mean`
Rewording of the issue described in #2010 by @21771:
`groupby.mean` can have incorrect results due to integer overflow in the sum portion. We should be able to do the sum portion as float for `.mean` calls

User description:
> incorrect results for aggregations of many very large int64s. The sums appear to
overflow 64 bits. A test would be to use a column containing values = `2**63-1` as a target
for the .mean() of a GroupBy. This will return values of less than `2**63-1` as a result
of an overflow. I saw the same behavior with .sum() where the sum was less than the value of individual elements.

Edit:
Adding reproducer for this issue:
```python
>>> b = ak.arange(10) % 2 == 0
>>> g = ak.GroupBy(b)
>>> g.count()
(array([False True]), array([5 5]))

>>> a = ak.full(10, 2**63 -1, dtype=ak.int64)
>>> a
array([9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807])

# since all the values are the same, the mean should be 2**63 - 1, we see that's not the case
>>> g.mean(a)
(array([False True]), array([1.8446744073709553e+18 1.8446744073709553e+18]))
>>> float(a[0])
9.223372036854776e+18
```

pierce314159:
This PR (fixes #2076) casts to `real` early in `groupby.mean` to avoid integer overflow when aggregating over large values

<details>
  <summary>Output on reproducer</summary>

set up:
```python
>>> b = ak.arange(10) % 2 == 0
>>> g = ak.GroupBy(b)
>>> g.count()
(array([False True]), array([5 5]))

>>> a = ak.full(10, 2**63 - 1, dtype=ak.int64)
>>> a
array([9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807 9223372036854775807])
```

before this PR:
```python
# since all the values are the same, the means should be 2**63 - 1, we see that's not the case
>>> g.mean(a)
(array([False True]), array([1.8446744073709553e+18 1.8446744073709553e+18]))
>>> float(a[0])
9.223372036854776e+18
```

after this PR:
```python
# now the means are 2**63 - 1
>>> g.mean(a)
(array([False True]), array([9.2233720368547758e+18 9.2233720368547758e+18]))
>>> float(a[0])
9.223372036854776e+18
```
</details>

### DIFF
--- a/src/ReductionMsg.chpl
+++ b/src/ReductionMsg.chpl
@@ -777,18 +777,21 @@ module ReductionMsg
     proc segMean(values:[] ?t, segments:[?D] int, skipNan=false): [D] real throws {
       var res: [D] real;
       if (D.size == 0) { return res; }
+      // convert to real early to avoid int overflow
+      overMemLimit(numBytes(real) * values.size);
+      var real_values = values: real;
       var sums;
       var counts;
-      if isRealType(t) && skipNan {
-        // first verify that we can make a copy of values
-        overMemLimit(numBytes(t) * values.size);
-        // calculate sum and counts with nan values replaced with 0.0
-        var arrCopy = [elem in values] if isnan(elem) then 0.0 else elem;
+      if skipNan {
+        // first verify that we can make a copy of real_values
+        overMemLimit(numBytes(real) * real_values.size);
+        // calculate sum and counts with nan real_values replaced with 0.0
+        var arrCopy = [elem in real_values] if isnan(elem) then 0.0 else elem;
         sums = segSum(arrCopy, segments);
-        counts = segCount(segments, values.size) - nanCounts(values, segments);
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good.

joshmarshall1 (approved):

jaketrookman (approved):
Looks good

PR #2081, opened 2023-01-25
Closes #2077 Remove duplicated `bigint` logic in `IndexingMsg.chpl`

### LINKED ISSUE #2077: Remove duplicated `bigint` logic in `IndexingMsg.chpl`
Now that we have an aggregation interface for `bigint`, we can remove a lot of duplicated logic. An example is something like:
https://github.com/Bears-R-Us/arkouda/blob/e309d0e2ab324aac835bdc4bbb4f2055da6c8c85/src/IndexingMsg.chpl#L320-L337

This can now be updated to match the others, so just
```chapel
when (DType.BigInt) {
    return sliceHelper(bigint);
}
```

jaketrookman:
This PR (closes #2077) cleans up `IndexingMsg.chpl` by removing the duplicated logic for `bigint` implementation, as we now have an aggregation interface for `bigint`.

### DIFF
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -963,7 +963,7 @@ module IndexingMsg
             }
             else {
                 var e = toSymEntry(gX,t);
-                // var truth = toSymEntry(gIV,bool);
+                var truth = toSymEntry(gIV,bool);
                 const ref ead = e.a.domain;
                 ref ea = e.a;
                 ref trutha = truth.a;
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -970,22 +970,23 @@ module IndexingMsg
             //         }
             //     }
             // }
-            if t == bigint {
-                var e = toSymEntry(gX, bigint);
-                var truth = toSymEntry(gIV,bool);
-                // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit
-                overMemLimit(numBytes(int) * truth.size);
-                var iv: [truth.a.domain] int = (+ scan truth.a);
-                var pop = iv[iv.size-1];
-                imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
-                                            "pop = %t last-scan = %t".format(pop,iv[iv.size-1]));
-                var y = toSymEntry(gY,t);
-                if (y.size != pop) {
-                    var errorMsg = "Error: %s: pop size mismatch %i %i".format(pn,pop,y.size);
-                    imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
-                    return new MsgTuple(errorMsg,MsgType.ERROR);
-                }
-                ref ya = y.a;
+            
+            var e = toSymEntry(gX, bigint);
+            var truth = toSymEntry(gIV,bool);
+            // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit
+            overMemLimit(numBytes(int) * truth.size);
+            var iv: [truth.a.domain] int = (+ scan truth.a);
+            var pop = iv[iv.size-1];
+            imLogger.debug(getModuleName(),getRoutineName(),getLineNumber(), 
+                                        "pop = %t last-scan = %t".format(pop,iv[iv.size-1]));
+            var y = toSymEntry(gY,t);
+            if (y.size != pop) {
+                var errorMsg = "Error: %s: pop size mismatch %i %i".format(pn,pop,y.size);
+                imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+                return new MsgTuple(errorMsg,MsgType.ERROR);
+            }
+            ref ya = y.a;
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
Looks good

joshmarshall1 (approved):
Looks good

Ethan-DeBandi99 (approved):
Looks good.

src/IndexingMsg.chpl
```diff
@@ -1328,31 +980,31 @@ module IndexingMsg
...
+                return ivInt64Helper(bigint);
             }
             when (DType.BigInt, DType.Int64, DType.Int64) {
-              return bigintIvInt64Helper(int);
+                return ivInt64Helper(int);
             }
             when (DType.BigInt, DType.Int64, DType.UInt64) {
-              return bigintIvInt64Helper(uint);
+                return ivInt64Helper(uint);
             }
```
pierce314159, line 990: there's an issue here. I don't know why the compiler doesn't complain, but we have the same aggregation into a bigint array from a non-bigint type. I just pulled down this PR and verified that the below block of code causes the server to hang

```python
>>> a = ak.arange(2**80, 2**80+ 10)
>>> a[ak.arange(5,10)] = ak.arange(5,10)
```
On master this gives us the expected result
```python
>>> a = ak.arange(2**80, 2**80+ 10)
>>> a[ak.arange(5,10)] = ak.arange(5,10)
>>> a
array(["1208925819614629174706176" "1208925819614629174706177" "1208925819614629174706178" "1208925819614629174706179" "1208925819614629174706180" "5" "6" "7" "8" "9"])
```


We prob want to do something similar to `setPdarrayIndexToValueMsg` where we have a dtype const and set it to something like
```chapel
const dtype = if gX.dtype == DType.BigInt then DType.BigInt else gY.dtype;
```
then we'll need to update the select statement to be `select(gX.dtype, gIV.dtype, dtype)`. And change the helper functions to cast the type to dtype in the `agg.copy`

Once we do that we can remove the cases that aren't `when (DType.BigInt, indexType, DType.BigInt)`

pierce314159 (changes requested):
One issue in `setPdarrayIndexToPdarrayMsg`

src/IndexingMsg.chpl
Ethan-DeBandi99: Good catch @pierce314159.

Ethan-DeBandi99 (approved):
Looks good to me, but definitely want @pierce314159 to take another look just to be sure.

src/IndexingMsg.chpl
```diff
@@ -1078,27 +843,53 @@ module IndexingMsg
...
+                    var errorMsg = "Error: %s: OOBindex %i > %i".format(pn,ivMax,e.size-1);
+                    imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);           
+                    return new MsgTuple(errorMsg,MsgType.ERROR);
+                }
+                ref iva = iv.a;
+                var ya = y.a;
+                ref ea = e.a;
+                forall (i,v) in zip(iva,ya) with (var agg = newDstAggregator(t)) {
+                    agg.copy(ea[i],v);
+                }
```
pierce314159, line 892: I think we should try and pull as much common logic out of the if block as possible, something like
```chapel
var iv = toSymEntry(gIV,int);
var ivMin = min reduce iv.a;
var ivMax = max reduce iv.a;
var y = toSymEntry(gY,t);
if ivMin < 0 {
    var errorMsg = "Error: %s: OOBindex %i < 0".format(pn,ivMin);
    imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg); 
    return new MsgTuple(errorMsg,MsgType.ERROR);
}
if ivMax >= gX.size {
    var errorMsg = "Error: %s: OOBindex %i > %i".format(pn,ivMax,gX.size-1);
    imLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);           
    return new MsgTuple(errorMsg,MsgType.ERROR);
}
ref iva = iv.a;
ref ya = y.a;
if gX.dtype == DType.BigInt {
    var e = toSymEntry(gX, bigint);
    // NOTE y.etype will never be real when gX.dtype is bigint, but the compiler doesn't know that
    var tmp = if y.etype == bigint then ya else if (y.etype == bool || y.etype == real) then ya:int:bigint else ya:bigint;
    ref ea = e.a;
    forall (i,v) in zip(iva,tmp) with (var agg = newDstAggregator(bigint)) {
        agg.copy(ea[i],v);
    }
}
else {
    var e = toSymEntry(gX,t);
    ref ea = e.a;
    forall (i,v) in zip(iva,ya) with (var agg = newDstAggregator(t)) {
        agg.copy(ea[i],v);
    }
}
```
And the same type of deal in the other helper functions

pierce314159 (changes requested):
Really close but I think for cleanup we should pull out of the if block as much of the common logic as we can

pierce314159 (approved):
This looks great! Really cuts down on the amount of code! Nice work!

Issue #2046, opened 2023-01-11
Rework Usage Section of Sphinx Docs


Ethan-DeBandi99:
The `Usage` section of our documentation generated by Sphinx on GitHub pages seems to be outdated. Additionally, a lot of the sections regenerate function documentation that is available in the API Reference that is auto-generated. At a minimum, we should update this regeneration to simply link into the API Reference. 

I would recommend either removing or refreshing this section entirely to make sure it is as up-to-date as possible.

Ethan-DeBandi99:
Users have reported missing functionality from the docs.

Ethan-DeBandi99:
@jaketrookman - I would add a Getting Started file with some of the basic information I am about to remove from quickstart.  I am providing it below:

```rst
******************************
Simple Computations
******************************

Create and sum an array
=========================

The following code creates an arkouda ``pdarray`` that resides on the arkouda server and performs a server-side computation, returning the result to the Python client.

.. code-block:: python

   # Create a server-side array with integers from 1 to N inclusive
   # This syntax is from NumPy
   >>> N = 10**6
   >>> A = ak.arange(1, N+1, 1)
   # Sum the array, returning the result to Python
   >>> print(A.sum())
   # Check the result
   >>> assert A.sum() == (N * (N + 1)) // 2

Array arithmetic
=========================
   
Now, we will perform an operation on two ``pdarray`` objects to create a new ``pdarray``. This time, the result will not be returned to the Python client, but will be stored on the server. In general, only scalar results are automatically returned to Python; ``pdarray`` results remain on the server unless explicitly transferred by the user (see :py:meth:`arkouda.pdarray.to_ndarray`).

.. code-block:: python

   # Generate two (server-side) arrays of random integers 0-9
   >>> B = ak.randint(0, 10, N)
   >>> C = ak.randint(0, 10, N)
   # Multiply them (server-side)
   >>> D = B * C
   # Print a small representation of the array
   # This does NOT move the array to the client
   >>> print(D)
   # Get the min and max values
   # Because these are scalars, they live in Python
   >>> minVal = D.min()
   >>> maxVal = D.max()
   >>> print(minVal, maxVal)

Indexing
=========================

Arkouda ``pdarray`` objects support most of the same indexing and assignment syntax of 1-dimensional NumPy ``ndarray``s (arkouda currently only supports 1-D arrays). This code shows two ways to get the even elements of ``A`` from above: with a slice, and with logical indexing.

.. code-block:: python

[...]

PR #2091, opened 2023-01-27
Closes #1988 Update `pydoc/quickstart.rst`

### LINKED ISSUE #1988: Update `pydoc/quickstart.rst`
Complete after PR #1985 has been merged.

Updates can be completed as `.rst` or `.md`

Update the file to follow the workflow listed below:

- Overview indicating that this is intended for users to install, not developers
- Link to the Chapel quickstart guide to download and install Chapel, [here](https://chapel-lang.org/docs/usingchapel/QUICKSTART.html)
- Add instructions similar to the Chapel quickstart guide
    - Download compressed Arkouda file
    - Instructions to uncompress and install Arkouda
    - Run `make test` to validate install

Ethan-DeBandi99:
closes #1988

Updates our `Quick Start Guide` to provide a simplified user install workflow. This provides a download link. Some of the basic information has been removed, but I have added a comment to #2046 to add this into a basic `Getting Started` doc.

**This will require updating the link after each release, but I believe it significantly lowers the barrier to entry for someone who may not want to pull the repo or has no need to.**

### DIFF
--- a/pydoc/quickstart.rst
+++ b/pydoc/quickstart.rst
@@ -87,7 +87,7 @@ In another terminal window, launch an interactive Python 3 session, such as ``ip
 .. code-block:: python
 
    >>> import arkouda as ak
-   >>> default way to connect is
+   # default way to connect is
    >>> ak.connect(connect_url='tcp://node01:5555')
    ...
    connected to tcp://node01:5555
--- a/pydoc/quickstart.rst
+++ b/pydoc/quickstart.rst
@@ -4,10 +4,51 @@
 Quickstart
 #######################
 
-This guide assumes you have satisfied the `Requirements <setup/REQUIREMENTS.html>`_ and followed the :ref:`installation-label` to build the arkouda server. Also, both your ``PATH`` and ``PYTHONPATH`` environment variables should contain the arkouda root directory.
+This guide is intended to instruct users on the installation Arkouda. It will also walk through launching and shutting down the server. For usage information, visit our :ref:`usage-label`
 
 **********************
-Launch Arkouda Server
+Install Dependencies
+**********************
+
+1. Follow the `Chapel Quickstart Guide <https://chapel-lang.org/docs/usingchapel/QUICKSTART.html>`_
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good to me

pydoc/quickstart.rst
```diff
@@ -4,10 +4,51 @@
...
+
+   .. substitution-code-block:: bash
+
+       tar xzf arkouda-2023.01.11.tar.gz
+
+3. Change to the arkouda directory
+
+   .. substitution-code-block:: bash
+
+       cd arkouda-2023.01.11
```
pierce314159, line 36: is this something we should try to update each release?

pydoc/quickstart.rst
```diff
@@ -49,98 +92,28 @@ In another terminal window, launch an interactive Python 3 session, such as ``ip
    ...
    connected to tcp://node01:5555
```
pierce314159, line 93: in this block we do
```python
   >>> import arkouda as ak
   >>> default way to connect is
   >>> ak.connect(connect_url='tcp://node01:5555')
   ...
   connected to tcp://node01:5555
```

For the second line,  the `>>>` makes it seem like a command. I think it would look better if we changed the line to a comment instead 
```python
   >>> import arkouda as ak
   # default way to connect is
   >>> ak.connect(connect_url='tcp://node01:5555')
   ...
   connected to tcp://node01:5555
```

pierce314159 (approved):
Looks good to me. Only suggestion is in an existing section not added in this PR

pydoc/quickstart.rst
Ethan-DeBandi99: @pierce314159 - I mentioned in the description that after each release, we will need to update the download link and page to match the latest release. 

Maybe we should add the release process to the developer section and include doing this in the workflow?

pydoc/quickstart.rst
Ethan-DeBandi99: Definitely agree. Updated.

PR #2095, opened 2023-01-30
Closes #2089 - Update MacOS Install

### LINKED ISSUE #2089: Update MacOS Install
I was doing some testing and found that `anaconda3` now functions properly on Apple silicon. I would recommend that we update our docs to install recommend installing `anaconda3` and remove `miniforge`. Either will still work, but anaconda3 seems to be quicker with updates and support for new python version. The workflow is listed below:

```bash
brew install anaconda
/opt/homebrew/Caskroom/anaconda/2022.10/Anaconda3-2022.10-MacOSX-arm64.sh
```
This workflow will install anaconda.

Ethan-DeBandi99:
Closes #2089 

Updates the install documentation for MacOS to only recommend `anaconda3`. Previously, there were issues on `m1` chipset machines prevent it from being used. These issues have been resolved. Since, `anaconda3` provides more features and it now works on all platforms,  we will now only be recommending it for environment management. 

Adds additional steps to ensure `conda` gets installed and is configured properly.


pierce314159 (approved):
Looks good to me

jaketrookman (approved):
Looks good

PR #2101, opened 2023-02-01
Closes #2096: numpy overflow deprecations

### LINKED ISSUE #2096: Address NumPy Deprecations
NumPy version 1.24.1

When running make test produces several deprecation warning. All warnings appear related. Warning message: 
```
DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays. The conversion of -2 to uint64 will fail in the future.
For the old behavior, usually:
      np.array(value).astype(dtype)`
  will give the desired result (the cast overflows).
    pda[:10] = np.uint(-2)
```
The message above is one example. To view all and to verify all have been resolved run `make test`. Based on the locations where these errors are occuring, we will need to update testing and some array interactions.

pierce314159:
This PR (closes #2096) updates our tests to avoid the numpy overflow deprecation warning

### DIFF
--- a/tests/array_view_test.py
+++ b/tests/array_view_test.py
@@ -67,7 +67,7 @@ class ArrayViewTest(ArkoudaTest):
         inav[nind] = -9999
         unav[nind] = 2**64 - 9999
         iav[uind] = -9999
-        uav[iind] = 2**64 - 9999
+        uav[iind] = -9999
         self.assertEqual(iav[uind], inav[nind])
         self.assertEqual(iav[iind], iav[uind])
         self.assertEqual(uav[uind], unav[nind])
--- a/tests/array_view_test.py
+++ b/tests/array_view_test.py
@@ -65,9 +65,9 @@ class ArrayViewTest(ArkoudaTest):
         uind = ak.cast(iind, ak.uint64)
 
         inav[nind] = -9999
-        unav[nind] = -9999
+        unav[nind] = 2**64 - 9999
         iav[uind] = -9999
-        uav[iind] = -9999
+        uav[iind] = 2**64 - 9999
         self.assertEqual(iav[uind], inav[nind])
         self.assertEqual(iav[iind], iav[uind])
         self.assertEqual(uav[uind], unav[nind])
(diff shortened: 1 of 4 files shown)


tests/array_view_test.py
```diff
@@ -65,9 +65,9 @@ def test_set_index(self):
         uind = ak.cast(iind, ak.uint64)
 
         inav[nind] = -9999
-        unav[nind] = -9999
+        unav[nind] = 2**64 - 9999
         iav[uind] = -9999
-        uav[iind] = -9999
+        uav[iind] = 2**64 - 9999
```
Ethan-DeBandi99, line 70: I want to discuss this a bit more, specifically for setting a value in a `pdarray`. I think overall this change is going to have larger implications. Especially, because a user will still be able to call `uav[iind] = -9999`.

I think we need to document this at a minimum. Especially, seen as these updates do not address the areas where the warning can be produced within the main code, they are updating the format to not produce the warning. However, it can still happen within our code. Not sure if there is a good way to address, but think it warrants some more discussion.

Ethan-DeBandi99 (commented):
The changes here address the `NumPy` update, but I am not sure our code cannot still have issues with this. I added most of my notes in one comment, but I think it applies overall.

tests/array_view_test.py
pierce314159: Thanks for the feedback! I updated `format_other` to allow us to keep using vals that are out of bounds without having the numpy deprecation warning
```python
>>> ak.array([2**64 - 1, 0, -1], dtype=ak.uint64)
array([18446744073709551615 0 18446744073709551615])

>>> pda = ak.arange(10, dtype=ak.uint64)
>>> pda[5:10] = -2
>>> pda
array([0 1 2 3 4 18446744073709551614 18446744073709551614 18446744073709551614 18446744073709551614 18446744073709551614])
```

Ethan-DeBandi99 (approved):
Looks good to me. Thanks for taking care of this!

joshmarshall1 (approved):

PR #2106, opened 2023-02-02
Fixes #2105: bug when loading a `dataframe` containing a `segarray` with an `_` in the column name

### LINKED ISSUE #2105: bug loading `dataframe` containing a `segarray` with an `_` in column name
**Feature Description**
When saving dataframes containing segarrays I have not found a straight forward way to read the data back into a dataframe. Is there a command or flags to enable automatic recreation of the segarrays from the _segments and _values arrays? This may exist but I have not found it.

```
**Example Behavior**
example:    df = ak.DataFrame(ak.recreate_segarrays(ak.read('foo.hdf')))
or               df = ak.DataFrame(ak.read('foo.hdf',restore_segarrays=True))
or               df = ak.DataFrame(ak.read('foo.hdf'),restore_segarrays=True)
```

**Describe Workarounds**
currently have a function to do this based on dataset names and array lengths, but am looking for a more robust capability.

pierce314159:
This PR (fixes #2105) there was a bug preventing `ak.DataFrame.load` from handling segarrays when their column name contains an `_`

I think it's easiest to see this by looking at an example
```python
# this works when there's no `_` in the column name
>>> col = 'NO-UNDERSCORES-EVER_segments'
>>> col.split("_")[0]
'NO-UNDERSCORES-EVER'

>>> '_'.join(col.split("_")[:-1])
'NO-UNDERSCORES-EVER'

# but breaks when there is
>>> col = 'c_3_segments'
>>> col.split("_")[0]
'c'

>>> '_'.join(col.split("_")[:-1])
'c_3'
```

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1599,9 +1599,11 @@ class DataFrame(UserDict):
 
         # this assumes segments will always have corresponding values.
         # This should happen due to save config
-        seg_cols = [col.split("_")[0] for col in df_dict.keys() if col.endswith("_segments")]
+        seg_cols = ["_".join(col.split("_")[:-1]) for col in df_dict.keys() if col.endswith("_segments")]
         df_dict_keys = [
-            col.split("_")[0] if col.endswith("_segments") or col.endswith("_values") else col
+            "_".join(col.split("_")[:-1])
+            if col.endswith("_segments") or col.endswith("_values")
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good

Ethan-DeBandi99 (approved):
Looks good

Issue #2088, opened 2023-01-26
Improve bigint creation performance


pierce314159:
There's been feedback that performance for creating bigint pdarrays is really poor especially when the `max_bits` flag is set. I'll work on speeding this up a bit and create a benchmark to track the perf changes.

It is definitely possible that just creating a large array of bigints is too slow for the common use cases. If that's the case, we might want to investigate a different implementation that keeps up with `2/4` uint arrays when `max bits <= 128/256` respectively and use the current implementation only when `max_bits` is not set or is `>256`. Maybe @ronawho will have a better idea of this

Edit:
Sorry I don't think i made it clear, this is specifically creating `bigint` pdarray using the `bigint_from_uint_arrays` function

ronawho:
I haven't seen bigint array creation be too expensive. It is more expensive than creating int arrays since we have to do a small heap allocation for each element, but I wouldn't expect it to be massively behind. I think I was seeing maybe a 4x perf hit compared to `int(64)` array creation when I checked a standalone chapel test. 

I was seeing much slower array deinit, which I noted in https://github.com/chapel-lang/chapel/issues/15215, but I think that'll have a pretty easy fix.

---

Here's a quick perf test:

```chpl
use TestBase;
use BigInteger;

config const N = 10**8;
proc testit(type elemType) {
  writeln(elemType:string);
  const D = makeDistDom(N*numLocales);

  var d: Diags;
  {
    d.start();
    var A: [D] elemType;
    d.stop("array create");

    d.start();
    forall a in A do a += 4;
    d.stop("array add");

    d.start();
  }
  d.stop("array delete");
  writeln();
}

testit(int);
testit(bigint);
```

16-node-cs-hdr results:
```
 make test-bin/BigintCreation
./test-bin/BigintCreation -nl 16 --N=$((2**30))

int(64)
array create: 0.13s
array add: 0.15s
array delete: 0.00s

bigint
array create: 0.38s
array add: 0.57s
array delete: 25.62s
```

So bigint creation is ~3x slower than int64 creation, operators are maybe ~4x (though will get more expensive with more digits/limbs), and deletion is a lot slower. That's because deinit is currently serialized (which doesn't matter for POD types since there's nothing to deinit.) Parallelizing deinit with some upstream chapel changes brings this much closer to creation times:

```
./test-bin-par-deinit/BigintCreation -nl 16 --N=$((2**30))

int(64)
array create: 0.13s
array add: 0.15s
[...]

PR #2131, opened 2023-02-09
Part of #2088: Speedup bigint_from_uint_arrays

ronawho:
Optimize the uint->bigint copy by limiting temporary array creation and avoiding what was effectively an array-of-dist-arrays reduction. For max_bits, optimize modding by localizing the `modBy` bigint.

On 16-node-cs-hdr, `bigint_from_uint_arrays` is over 100x faster when `max_bits` isn't set, and nearly 1000x faster when it is:

```python

import argparse, math, time
import arkouda as ak

parser = argparse.ArgumentParser("bigint_from_uint_arrays perf test")
parser.add_argument("hostname")
parser.add_argument("port", type=int)
parser.add_argument("--size", type=int, default=2**20)
parser.add_argument("--max-bits", type=int, default=-1)
args = parser.parse_args()

ak.connect(args.hostname, args.port)

a = ak.randint(0, 2**32, args.size, dtype=ak.uint64)
b = ak.randint(0, 2**32, args.size, dtype=ak.uint64)

start = time.time()
ba = ak.bigint_from_uint_arrays([a, b], max_bits=args.max_bits)
end = time.time()
print("time: {:.2f} sec".format(end-start))
```

```
runClient -nl 16 ./uint_to_bigint_perf.py --max-bits={-1,64} --size=$((2**{26,30}))
```

For `--size=$((2**26))` (1G, 64M/node)

| config      | before   | after |
| ----------- | -------: | ----: |
| max_bits=-1 |   3.23s  | 0.02s |
| max_bits=64 | 235.80s  | 0.03s |

For `--size=$((2**30))` (16G, 1G/node)

| config      | before   | after |
| ----------- | -------: | ----: |
| max_bits=-1 |  57.78s  | 0.12s |
| max_bits=64 |     N/A  | 0.19s |

As future work, this could be further optimized by copying the uint arrays directly into the bigint limbs and masking instead of modding.

Part of #2088

### DIFF
--- a/src/BigIntMsg.chpl
+++ b/src/BigIntMsg.chpl
@@ -26,23 +26,25 @@ module BigIntMsg {
         var max_bits = msgArgs.get("max_bits").getIntValue();
 
         var bigIntArray = makeDistArray(len, bigint);
-        // block size is 2**64
-        var block_size = 1:bigint;
-        block_size <<= 64;
-        forall (name, i) in zip(arrayNames, 0..<num_arrays by -1) with (+ reduce bigIntArray) {
-            var tmp = toSymEntry(getGenericTypedArrayEntry(name, st), uint).a:bigint;
-            tmp <<= (64*i);
-            bigIntArray += tmp;
+        for (name, i) in zip(arrayNames, 0..<num_arrays by -1) {
+            ref uintA = toSymEntry(getGenericTypedArrayEntry(name, st), uint).a;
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
Looks good. The speed up is awesome.

pierce314159 (approved):
This is fantastic!! Thanks so much for looking into this, the performance increase is amazing! 🚀

It wouldn't have expected the `+ reduce` to be such a problem, it'll be really good to keep this optimization in mind for future bigint functionality

PR #2135, opened 2023-02-09
Close #1522 - Add `Index`&`MultiIndex` Support for `key` in `Series.locate()`

### LINKED ISSUE #1522: Add support for `key` to be `Index` in `Series.locate(key)`
Related to a question from #1477, should we add the ability for `key` to be an `Index` in `Series.locate`? It should be a simple fix if we want it, i believe all that's required is 
```python
def locate(self, key: Union[int, pdarray, Index, Series, List, Tuple]) -> Series:
  ...
  elif isinstance(key, Index):
    idx = self.index.lookup(key.index)
  ...
```
but we should make sure to test single and multi index to verify correct functionality

Ethan-DeBandi99:
Closes #1522

Adds support for `MultiIndex` and `Index` in `Series.locate()`. 

Updates `MultiIndex.lookup` and `Index.lookup` to allow for single value objects to be passed for proper functionality when called from `Series.locate()`.

Adds testing to validate functionality.

### DIFF
--- a/arkouda/index.py
+++ b/arkouda/index.py
@@ -193,7 +193,11 @@ class Index:
 
     def lookup(self, key):
         if not isinstance(key, pdarray):
-            raise TypeError("Lookup must be on an arkouda array")
+            # try to handle single value
+            try:
+                key = array([key])
+            except Exception:
+                raise TypeError("Lookup must be on an arkouda array")
 
         return in1d(self.values, key)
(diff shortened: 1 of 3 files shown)


pierce314159 (approved):
Looks good to me

jaketrookman (approved):
Looks good

PR #2144, opened 2023-02-14
Closes #1961: `bigint` bug hotfix

### LINKED ISSUE #1961: Figure out hotfixes for bigint bugs
Pierce has reported a few bigint bugs upstream:
 - https://github.com/chapel-lang/chapel/issues/21206
 - https://github.com/chapel-lang/chapel/issues/21194

Once those are fixed (and assuming we don't want to wait for Chapel 1.30) we'll want to figure out some way to bring these fixes in.  This has been done in the past for things like random module fixes in https://github.com/Bears-R-Us/arkouda/issues/1220, so don't think will be too hard, just wanted to capture that we'll want to do something here.

pierce314159:
This PR adds the last `bigint` bug hotfix needed to close #1961

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -1222,7 +1222,7 @@ module BinOp
             [(ei,li,ri) in zip(tmp,la,ra)] ei.powMod(li, ri, max_size);
           }
           else {
-            tmp = la ** ra;
+            tmp = la ** ra:int;
           }
           visted = true;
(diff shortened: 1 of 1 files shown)


bmcdonald3 (approved):
Looks good, thanks Pierce! I think the uint -> int case is the only thing I really thought about, but I don't really think that's a problem realistically.

PR #2151, opened 2023-02-15
Closes #2150: Update path in IO benchmarks

### LINKED ISSUE #2150: glob error with path in `IO/multi_IO.py` benchmark
Someone was having a glob expression error in the benchmarks. The only benchmark that uses glob seems to be `IO.py`

I printed out the `args.path` and see this: `/Users/pierce/proj/arkoudaak-io-test`, so I updated this line to default to `os.getcwd() + "/ak-io-test"`
https://github.com/Bears-R-Us/arkouda/blob/1b827c0b00524a2d734758186b90086ccf0dfb25/benchmarks/IO.py#L126

and get a more reasonable looking path: `/Users/pierce/proj/arkouda/ak-io-test`. It's worth noting that both versions work for me locally running `./benchmarks/run_benchmarks.py --size=1000 IO`. I'm hoping @bmcdonald3 could verify this is sensible.

Also, I noticed that we have a place in `check_correctness` where we could use `remove_files`
https://github.com/Bears-R-Us/arkouda/blob/1b827c0b00524a2d734758186b90086ccf0dfb25/benchmarks/IO.py#L99-L100
https://github.com/Bears-R-Us/arkouda/blob/1b827c0b00524a2d734758186b90086ccf0dfb25/benchmarks/IO.py#L73-L75

pierce314159:
This PR (closes #2150):
- Updates default path in IO benchmarks to do `os.path.join(os.getcwd(), "ak-io-test")` instead
- Changes one spot to use `remove_files` helper function

### DIFF
--- a/benchmarks/IO.py
+++ b/benchmarks/IO.py
@@ -96,8 +96,7 @@ def check_correctness(dtype, path, seed, parquet, multifile=False):
 
     c = ak.read_hdf(path + "*") if not parquet else ak.read_parquet(path + "*")
 
-    for f in glob(path + "*"):
-        os.remove(f)
+    remove_files(path)
     if not multifile:
         assert (a == c).all()
     else:
@@ -123,7 +122,7 @@ def create_parser():
     parser.add_argument(
         "-p",
         "--path",
-        default=os.getcwd() + "ak-io-test",
+        default=os.path.join(os.getcwd(), "ak-io-test"),
         help="Target path for measuring read/write rates",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


bmcdonald3 (approved):
This looks good and makes sense to me. Were you hoping I'd run any specific tests or anything or just wanted to make sure I was on board with the change?

I think this issue stems from my relative lack of inexperience with Python, but reading some docs now I think this is correct.

jaketrookman (approved):
Looks good

pierce314159:
> This looks good and makes sense to me. Were you hoping I'd run any specific tests or anything or just wanted to make sure I was on board with the change?
> 
> I think this issue stems from my relative lack of inexperience with Python, but reading some docs now I think this is correct.

Just doublechecking there wasn't a good reason I was missing to keep it as is! Thanks for taking a look!

Ethan-DeBandi99 (approved):
Looks good

PR #2157, opened 2023-02-17
Closes #2156: Bigint stream benchmark

### LINKED ISSUE #2156: bigint stream benchmark
Add a bigint stream benchmark performing `c = a + b * alpha`

### LINKED ISSUE #2165: bigint bitwise binops benchmark
Add benchmark for bigint bitwise binops

pierce314159:
This PR (closes #2156 and closes #2165) adds a bigint stream benchmark. The bigint stream does not scale well. I couldn't find any obvious optimizations

`bigint_stream` after this PR:
```
% ./benchmarks/run_benchmarks.py bigint_stream -t 1
array size = 100,000,000
number of trials =  1
>>> arkouda bigint stream
numLocales = 1, N = 100,000,000
Average bigint stream time = 22.1851 sec
Average bigint stream rate = 0.10 GiB/sec

% ./benchmarks/run_benchmarks.py bigint_stream -t 1 --max-bits=64
array size = 100,000,000
number of trials =  1
>>> arkouda bigint stream
numLocales = 1, N = 100,000,000
Average bigint stream time = 20.6594 sec
Average bigint stream rate = 0.11 GiB/sec
```
`bigint_stream` before this PR:
```
% ./benchmarks/run_benchmarks.py bigint_stream -t 1
array size = 100,000,000
number of trials =  1
>>> arkouda bigint stream
numLocales = 1, N = 100,000,000
Average bigint stream time = 57.1895 sec
Average bigint stream rate = 0.04 GiB/sec

% ./benchmarks/run_benchmarks.py bigint_stream -t 1 --max-bits=64
array size = 100,000,000
number of trials =  1
>>> arkouda bigint stream
numLocales = 1, N = 100,000,000
Average bigint stream time = 58.1247 sec
Average bigint stream rate = 0.04 GiB/sec
```

Compared to non-bigint `stream`:
```
% ./benchmarks/run_benchmarks.py stream                    
array size = 100,000,000
number of trials =  6
>>> arkouda float64 stream
numLocales = 1, N = 100,000,000
Average time = 0.2566 sec
Average rate = 8.71 GiB/sec
```

bigint bitwise binops after this PR:
```
% ./benchmarks/run_benchmarks.py bigint_bitwise_binops -t 1
array size = 100,000,000
number of trials =  1
>>> arkouda bigint bitwise binops
numLocales = 1, N = 100,000,000
Average bigint AND time = 9.9440 sec
Average bigint AND rate = 0.30 GiB/sec

Average bigint OR time = 14.2256 sec
Average bigint OR rate = 0.21 GiB/sec

Average bigint SHIFT time = 23.6632 sec
Average bigint SHIFT rate = 0.06 GiB/sec

% ./benchmarks/run_benchmarks.py bigint_bitwise_binops -t 1 --max-bits=64
array size = 100,000,000
number of trials =  1
>>> arkouda bigint bitwise binops
numLocales = 1, N = 100,000,000
Average bigint AND time = 9.2267 sec
Average bigint AND rate = 0.32 GiB/sec

Average bigint OR time = 14.2185 sec
Average bigint OR rate = 0.21 GiB/sec

Average bigint SHIFT time = 22.9780 sec
Average bigint SHIFT rate = 0.06 GiB/sec
```

bigint bitwise binops before this PR:
```
% ./benchmarks/run_benchmarks.py bigint_bitwise_binops -t 1
array size = 100,000,000
number of trials =  1
>>> arkouda bigint bitwise binops
numLocales = 1, N = 100,000,000
Average bigint AND time = 21.7934 sec
Average bigint AND rate = 0.14 GiB/sec

Average bigint OR time = 26.4953 sec
Average bigint OR rate = 0.11 GiB/sec

Average bigint SHIFT time = 39.9405 sec
Average bigint SHIFT rate = 0.04 GiB/sec

 ./benchmarks/run_benchmarks.py bigint_bitwise_binops -t 1 --max-bits=64
array size = 100,000,000
number of trials =  1
>>> arkouda bigint bitwise binops
numLocales = 1, N = 100,000,000
Average bigint AND time = 22.2118 sec
Average bigint AND rate = 0.13 GiB/sec

Average bigint OR time = 25.9856 sec
Average bigint OR rate = 0.11 GiB/sec

Average bigint SHIFT time = 39.0686 sec
Average bigint SHIFT rate = 0.04 GiB/sec
```

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -1099,10 +1099,11 @@ module BinOp
     var has_max_bits = max_bits != -1;
     if has_max_bits {
       max_size <<= max_bits;
+      max_size -= 1;
     }
     ref la = l.a;
     ref ra = r.a;
-    var tmp = makeDistArray(la.size, bigint);
+    var tmp = if l.etype == bigint then la else if l.etype == bool then la:int:bigint else la:bigint;
     // these cases are not mutually exclusive,
     // so we have a flag to track if tmp is ever populated
     var visted = false;
@@ -1114,19 +1115,59 @@ module BinOp
       // both being bigint
       select op {
         when "&" {
-          tmp = la & ra;
+          if has_max_bits {
+            forall (t, ri) in zip(tmp, ra) with (var local_max_size = max_size) {
+              t &= ri;
+              t &= local_max_size;
+            }
+          }
+          else {
+            forall (t, ri) in zip(tmp, ra) {
+              t &= ri;
+            }
+          }
           visted = true;
         }
         when "|" {
-          tmp = la | ra;
+          if has_max_bits {
+            forall (t, ri) in zip(tmp, ra) with (var local_max_size = max_size) {
+              t |= ri;
+              t &= local_max_size;
+            }
+          }
+          else {
+            forall (t, ri) in zip(tmp, ra) {
+              t |= ri;
+            }
+          }
           visted = true;
         }
         when "^" {
-          tmp = la ^ ra;
+          if has_max_bits {
+            forall (t, ri) in zip(tmp, ra) with (var local_max_size = max_size) {
+              t ^= ri;
+              t &= local_max_size;
+            }
+          }
+          else {
+            forall (t, ri) in zip(tmp, ra) {
+              t ^= ri;
+            }
+          }
           visted = true;
         }
         when "/" {
-          tmp = la / ra;
+          if has_max_bits {
+            forall (t, ri) in zip(tmp, ra) with (var local_max_size = max_size) {
+              t /= ri;
+              t &= local_max_size;
+            }
+          }
+          else {
+            forall (t, ri) in zip(tmp, ra) {
+              t /= ri;
+            }
+          }
           visted = true;
         }
       }
@@ -1138,7 +1179,17 @@ module BinOp
         // can't shift a bigint by a bigint
         select op {
           when "<<" {
-            tmp = la << ra;
+            if has_max_bits {
+              forall (t, ri) in zip(tmp, ra) with (var local_max_size = max_size) {
+                t <<= ri;
+                t &= local_max_size;
+              }
+            }
+            else {
+              forall (t, ri) in zip(tmp, ra) {
+                t <<= ri;
+              }
+            }
             visted = true;
           }
           when ">>" {
@@ -1148,7 +1199,17 @@ module BinOp
             var divideBy = makeDistArray(la.size, bigint);
             divideBy = 1:bigint;
             divideBy <<= ra;
-            tmp = la / divideBy;
+            if has_max_bits {
+              forall (t, dB) in zip(tmp, divideBy) with (var local_max_size = max_size) {
+                t /= dB;
+                t &= local_max_size;
+              }
+            }
+            else {
+              forall (t, dB) in zip(tmp, divideBy) {
+                t /= dB;
+              }
+            }
             visted = true;
           }
           when "<<<" {
@@ -1157,21 +1218,27 @@ module BinOp
             }
             // should be as simple as the below, see issue #2006
             // return (la << ra) | (la >> (max_bits - ra));
-            tmp = la << ra;
             var botBits = la;
             if r.etype == int {
-              var shift_amt = max_bits - ra;
               // cant just do botBits >>= shift_amt;
-              var divideBy = makeDistArray(la.size, bigint);
-              divideBy = 1:bigint;
-              divideBy <<= shift_amt;
-              botBits = botBits / divideBy;
-              tmp += botBits;
+              forall (t, ri, bot_bits) in zip(tmp, ra, botBits) with (var local_max_size = max_size) {
+                t <<= ri;
+                var div_by = 1:bigint;
+                var shift_amt = max_bits - ri;
+                div_by <<= shift_amt;
+                bot_bits /= div_by;
+                t += bot_bits;
+                t &= local_max_size;
+              }
             }
             else {
-              var shift_amt = max_bits:uint - ra;
-              botBits >>= shift_amt;
-              tmp += botBits;
+              forall (t, ri, bot_bits) in zip(tmp, ra, botBits) with (var local_max_size = max_size) {
+                t <<= ri;
+                var shift_amt = max_bits:uint - ri;
+                bot_bits >>= shift_amt;
+                t += bot_bits;
+                t &= local_max_size;
+              }
             }
             visted = true;
           }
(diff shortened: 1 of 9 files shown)


src/BinOp.chpl
```diff
@@ -1425,15 +1427,21 @@ module BinOp
        (val.type == bigint && (l.etype == int || l.etype == uint || l.etype == bool)) {
       select op {
         when "+" {
-          tmp = la + val;
+          forall (t, li) in zip(tmp, la) with (var local_val = val) {
+            t = li + local_val;
+          }
```
pierce314159, line 1432: @ronawho will this result in any extra overhead in the `val.type != bigint` cases? I was thinking it probably wouldn't be much if any since `int/uint` would be trivially copyable. If I'm wrong, do you think updating to something like the below would be a good idea
```chapel
if val.type != bigint {
  tmp = la + val;
}
else {
  forall (t, li) in zip(tmp, la) with (var local_val = val) {
    t = li + local_val;
  }
}
```

pierce314159:
I found some optimizations by refactoring code to try and favor inplace ops. So instead of 
```chapel
// instead of
tmp = la * ra;
// do
tmp = la;
tmp *= ra;
```
The performance is better but still not great. I still need to update `doBigIntBinOpsv` and the bool return methods. I also think I'm gonna go ahead and add the logical/arithmetic benchmark in this PR to get a better idea of the perf boost these changes provide. Converting to a draft until I wrap that up

src/BinOp.chpl
pierce314159: I have another question, I have two versions of the same chunk of code
```chapel
if has_max_bits {
  forall t in tmp with (var local_val = val, var local_max_size = max_size) {
    t *= local_val;
    t &= local_max_size;
  }
}
else {
  forall t in tmp with (var local_val = val) {
    t *= local_val;
  }
}
```
and
```chapel
forall t in tmp with (var local_val = val, var local_max_size = max_size) {
  t *= local_val;
  if has_max_bits {
    t &= local_max_size;
  }
}
```
I like the second one better visually, but I felt like the first one would be faster (just checking the condition once instead of every iteration). However the performance seems to be the same locally. Maybe it's just negligible compared to the bottle neck of the bigint operation?

ronawho:
```chpl
forall (t, li) in zip(tmp, la) with (var local_val = val) {
  t = li + local_val;
}
```

Shouldn't have any overhead compared to:

```chpl
tmp = la + val;
```

for plain old numeric types. 


The promoted version is equivalent to (and transformed by the compiler into):

```chpl
forall (t, li) in zip(tmp, la) {
  t = li + val;
}
```

In this case the compiler will also detect that `val` is const and trivially-copyable, so it'll "remote value forward" it, which is basically means copy it's value into the argument bundled that's used to implement on-statements under the covers and avoid having to issue a `GET` for it. By introducing the task local var, there's an extra local copy at task startup, but this is so very trivial that you'll never notice the overhead.

ronawho:
```chpl
forall t in tmp with (var local_val = val, var local_max_size = max_size) {
  t *= local_val;
  if has_max_bits {
    t &= local_max_size;
  }
}
```

This version will have some overhead, but probably a fairly small amount. It's adding a branch to every iteration, but the branch will be correctly predicted since it doesn't change over the course of the loop. So that effectively means just a few extra memory reads each iteration, which should be fast, but so are the other operations in the loop (at least for small bigints).


Here's a perf test to check:

```chpl
use TestBase, BigInteger;

config const size=10**8;
config const has_max_bits = false;

proc testit(type elemType) {
  writeln("Testing '", elemType:string, "'");
  const D = makeDistDom(size*numLocales);
  var tmp: [D] elemType;
  var val = 5: elemType;
  var max_size = 64: elemType;

  var d: Diags;

  d.start();
  if has_max_bits {
    forall t in tmp with (var local_val = val, var local_max_size = max_size) {
      t *= local_val;
      t &= local_max_size;
    }
  } else {
    forall t in tmp with (var local_val = val) {
      t *= local_val;
    }
  }
  d.stop("Outer Branch");

  d.start();
  forall t in tmp with (var local_val = val, var local_max_size = max_size) {
    t *= local_val;
    if has_max_bits {
      t &= local_max_size;
    }
  }
  d.stop("Inner Branch");
}

testit(int); writeln();
testit(bigint);
```


```
16-node-cs-hdr: arkouda (master) $ make test-bin/BigintOpSpeedTest
16-node-cs-hdr: arkouda (master) $ ./test-bin/BigintOpSpeedTest -nl 16 --size=$((10**8))
Testing 'int(64)'
Outer Branch: 0.01s
Inner Branch: 0.01s

Testing 'bigint'
Outer Branch: 0.05s
Inner Branch: 0.06s

16-node-cs-hdr: arkouda (master) $ ./test-bin/BigintOpSpeedTest -nl 16 --size=$((2**30))
Testing 'int(64)'
Outer Branch: 0.11s
Inner Branch: 0.12s

Testing 'bigint'
Outer Branch: 0.61s
Inner Branch: 0.64s
```

So there is a minor perf hit from the inner branch, but I'd probably still go with that to avoid code duplication (increased compilation times and more code to maintain)

ronawho:
I also see fairly poor performance with 16-node-cs-hdr, but I see much better performance when enabling parallel array deinit (https://github.com/chapel-lang/chapel/pull/21670), which we saw benefit other bigint operations:

chapel 1.29:
```
>>> arkouda float64 stream
Average time = 0.0597 sec
Average rate = 599.50 GiB/sec

>>> arkouda bigint stream
Average bigint stream time = 14.5178 sec
Average bigint stream rate = 2.46 GiB/sec
```

chapel main w/ https://github.com/chapel-lang/chapel/pull/21670 (parallel deinit):
```
>>> arkouda float64 stream
Average time = 0.0589 sec
Average rate = 607.65 GiB/sec

>>> arkouda bigint stream
Average bigint stream time = 0.8836 sec
Average bigint stream rate = 40.47 GiB/sec
```

So ~20x improvement from parallel deinit. Still ~15x off from int/float stream, which is a little higher than I might expect, but not terrible.

pierce314159:
> So ~20x improvement from parallel deinit. Still ~15x off from int/float stream, which is a little higher than I might expect, but not terrible.

This is great news!!! So the bigint code will be much more performant in 1.30! Thanks for looking into this @ronawho!

ronawho:
> So the bigint code will be much more performant in 1.30!

Yeah, I think we'll highly highly recommend 1.30 for anybody using bigints when it's released (given the performance improvements, bug fixes, and implementation cleanup)

pierce314159:
This code should be ready to review. I will put up my most updated perf comparison either tonight or tomorrow but it's too pretty not to be outside rn

jaketrookman (approved):

joshmarshall1 (approved):
Looks good to me

ronawho (approved):

src/BinOp.chpl
```diff
@@ -1148,7 +1174,12 @@ module BinOp
             var divideBy = makeDistArray(la.size, bigint);
             divideBy = 1:bigint;
```
ronawho, line 1175: @pierce314159 we're seeing `bigint_bitwise_binops.py` timeout during nightly testing for `>>`. I think the problem is probably here with assigning all elements to a bigint `1` that lives on locale 0. I _think_ you could get rid of the tmp array and just do:

```chpl
forall t in tmp with (var dB = (1:bigint) << val, var local_max_size = max_size) {
```

ronawho: Whoops, I meant this as a comment for the binopVS case. For the binopVV cases you should be able to do something like:

```chpl
forall (t, ri) in zip(tmp, ra) with (var local_max_size = max_size) {
  var dB = 1:bigint;
  dB <<= ri;
  t /= dB;
```

ronawho: Opened https://github.com/Bears-R-Us/arkouda/pull/2178

PR #2158, opened 2023-02-21
fixed bug in pct memory used calculation

### LINKED ISSUE #2152: enhance memory management, logging, and metrics
**Feature Description**

1. Document how to set --memMax to set a specific, per-locale memory limit
2. Add logic to set per-locale memory limit per --memMax or host memory
3. Calculate pct memory used based upon configurable per-locale memory limit
4. Log/report remaining memory

**Additional context**
Related to #2056

hokiegeek2:
Closes #2152 

There is a bug in ServerDaemon.chpl where the pctMemUsed is calculated:

image

the divisor was set to memMax for testing purposes (test setting max memory via --memMax), but should have been replaced with getMemLimit, which generates the max memory from either memMax or the lone, previous method of 90% of physical memory.

The new version is correct, invoking the getMemLimit() function to return either the limit set via --memMax or calculated from 90% of max physical memory:

image

Testing shows that it is working correctly:

memMax set:

image

image

memMax not set:

image

image

### DIFF
--- a/src/ServerDaemon.chpl
+++ b/src/ServerDaemon.chpl
@@ -646,10 +646,11 @@ module ServerDaemon {
                 }
                 if (trace && memTrack) {
                     var memUsed = getMemUsed():uint * numLocales:uint;
-                    var pctMemUsed = ((memUsed:real/memMax:real)*100):int;
+                    var pctMemUsed = ((memUsed:real/getMemLimit():real)*100):int;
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):

joshmarshall1 (approved):

pierce314159 (approved):
looks good!

PR #2160, opened 2023-02-22
Closes #2145 - Enable Parquet Read SegArray w/ Empty Segments

### LINKED ISSUE #2145: Parquet Read SegArray with Empty Segments
The current implementation does not account for empty segments in the indexing.  The workflow for reading needs to be updated to properly increment past the empty segment.

Ethan-DeBandi99:
Closes #2145 

- Updates Parquet SegArray read workflow to properly account for and handle SegArrays with empty segments.
- Adds multiple test cases to ensure functionality with different locations of empty segments.
- Minor updates to documentation for Parquet to correct formatting.

**Important Note**
This code does not functions for SegArrays saved to Parquet with ALL segments empty. This is because most toolsets save this as a list of `nulls`. Currently, Arkouda does not have any support for Parquet `nulls` to be read. This will need to be added as a handled datatype and will be handled separately.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -265,10 +265,10 @@ int64_t cpp_getStringColumnNumBytes(const char* filename, const char* colname, v
   }
 }
 
-int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* chpl_offsets, int64_t numElems, int64_t startIdx, char** errMsg) {
+int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* chpl_seg_sizes, int64_t numElems, int64_t startIdx, char** errMsg) {
   try {
     int64_t ty = cpp_getType(filename, colname, errMsg);
-    auto offsets = (int64_t*)chpl_offsets;
+    auto seg_sizes = (int64_t*)chpl_seg_sizes;
     int64_t listSize = 0;
     
     if (ty == ARROWLIST){
@@ -313,7 +313,7 @@ int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* c
             int64_t value;
             (void)int_reader->ReadBatch(1, &definition_level, &rep_lvl, &value, &values_read);
             if (values_read == 0 || (!first && rep_lvl == 0)) {
-              offsets[i] = seg_size;
+              seg_sizes[i] = seg_size;
               i++;
               seg_size = 0;
             }
@@ -325,7 +325,7 @@ int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* c
               }
             }
             if (!int_reader->HasNext()){
-              offsets[i] = seg_size;
+              seg_sizes[i] = seg_size;
             }
           }
         } else if(lty == ARROWINT32 || lty == ARROWUINT32) {
@@ -336,7 +336,7 @@ int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* c
             int32_t value;
             (void)int_reader->ReadBatch(1, &definition_level, &rep_lvl, &value, &values_read);
             if (values_read == 0 || (!first && rep_lvl == 0)) {
-              offsets[i] = seg_size;
+              seg_sizes[i] = seg_size;
               i++;
               seg_size = 0;
             }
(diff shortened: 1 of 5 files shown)


src/ArrowFunctions.cpp
```diff
@@ -311,61 +312,44 @@ int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* c
...
-            } else {
-              // new segment starts. This has to be before increment to avoid false empty on first segment.
-              if (rep_lvl == 0 && seg_size >0) {
-                offsets[i] = seg_size;
-                i++;
-                seg_size = 0;
-              }
-              // increment counters
+            if (values_read == 0 || (!first && rep_lvl == 0)) {
+              offsets[i] = seg_size;
```
pierce314159, line 316: should we rename `offsets` in this function since it's really the `seg_sizes`

tests/parquet_test.py
```diff
@@ -309,25 +309,109 @@ def test_segarray_integration(self):
...
+        with tempfile.TemporaryDirectory(dir=ParquetTest.par_test_base_tmp) as tmp_dirname:
+            pq.write_table(table, f"{tmp_dirname}/empty_segments_LOCALE0000")
+            pq.write_table(table2, f"{tmp_dirname}/empty_segments_LOCALE0001")
+
+            ak_data = ak.read_parquet(f"{tmp_dirname}/empty_segments*")
+            print(ak_data.to_list())
+            self.assertIsInstance(ak_data, ak.SegArray)
+            self.assertEqual(ak_data.size, 9)
+            for i in range(5):
+                self.assertListEqual(df["ListCol"][i], ak_data[i].tolist())
```
pierce314159, line 381: I think this should be `i in range(9)` and i think the print is leftover from testing

tests/parquet_test.py
```diff
@@ -309,25 +309,109 @@ def test_segarray_integration(self):
...
```
pierce314159, line 412: Same as above:
I think this should be `i in range(9)` and i think the print is leftover from testing

pierce314159 (commented):
All the logic looks great!

The only thing I saw that really needs updating is a spot in the tests

src/ArrowFunctions.cpp
```diff
@@ -312,22 +313,21 @@ int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* c
...
                 i++;
                 seg_size = 0;
               }
               // increment counters
               seg_size++;
               vct++;
+              first = false;
               
               // if this is the last value in the iterator, assign the value in order to set last offset
               if (!int_reader->HasNext()){
```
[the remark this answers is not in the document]
pierce314159: oops this got changed before I reviewed and I forgot to remove it

jaketrookman (approved):
The changes look good!

src/ParquetMsg.chpl
```diff
@@ -178,7 +178,31 @@ module ParquetMsg {
     }
   }
 
-  proc readListFilesByName(A: [] ?t, filenames: [] string, sizes: [] int, dsetname: string, ty) throws {
+  proc computeIdx(offsets: [] int, val: int, lower_bound: bool = false): int throws {
+    // compute the index of the segment containing val
+    if lower_bound {
+      var (v, idx) = maxloc reduce zip(offsets >= val, offsets.domain);
+      return if v then idx else 0;
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Agreed. I forgot to update after we talked about this.

src/ArrowFunctions.cpp
Ethan-DeBandi99: We could, but we do the same thing in the strings case. I think it makes more sense though, so I will update it.

Ethan-DeBandi99: I don't think it really matters since we do the same thing in the Strings case. I can update it though so it is clearer.

Ethan-DeBandi99: I updated the entire workflow instead of just this function.

tests/parquet_test.py
Ethan-DeBandi99: corrected. Needed to update the assertion as well to check the merged dataframes.

tests/parquet_test.py
Ethan-DeBandi99: corrected. Needed to update the assertion as well to check the merged dataframes.

pierce314159 (approved):
Looks good to me!!

PR #2162, opened 2023-02-22
Add some error handling code for upcoming change to map module

bmcdonald3:
For the upcoming Chapel 1.30 release, we have made map methods throw instead of halt in https://github.com/chapel-lang/chapel/pull/21641. This means that, in order for Arkouda to build with the latest changes, we need to add `throws` in places where it was missing where the map accessors were being used in functions and add a `try!` when map accessors are used in module-level code (this is happening in `RandArray.chpl`). This does not change the behavior of the code in `RandArray`, but makes it so that invalid accesses will throw errors rather than crashing the server in `MetricsMsg`.

### DIFF
--- a/src/MetricsMsg.chpl
+++ b/src/MetricsMsg.chpl
@@ -194,7 +194,7 @@ module MetricsMsg {
         // total value of measurements to be averaged for each metric measured.s
         var measurementTotals = new map(string, real);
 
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (approved):
Looks good.

pierce314159 (approved):

PR #2169, opened 2023-02-24
Closes #2168 - Remove push from on clause

### LINKED ISSUE #2168: Remove Push from CI
Since we are using merge queue we should no longer need push in our `on` clause of the CI.yml

Ethan-DeBandi99:
Closes #2168 

Removes `push` from `on` clause in `CI.yml` since we are not using Merge Queue, we don't need to rerun CI on master after running on the Queue.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request, merge_group]
+on: [pull_request, merge_group]
 
 env:
   ARKOUDA_QUICK_COMPILE: true


pierce314159 (approved):

ronawho:
Note that this means you also lose CI checks when pushing to branches on your fork (which I sometimes use to test things out)

pierce314159:
Oh that's good to keep in mind! I've def done that before too. But I don't think the CI buys us much more than running `make test` and stuff locally other than verifying things work with past versions.

PR #2170, opened 2023-02-24
Closes #2163 - Fixes SymEntry Destruction Issue

### LINKED ISSUE #2163: SymEntry Destruction Issue
It appears that there are situations where Symbol Table entries are not being cleaned up properly. Upon creation, there is an expected number of entries. However, when setting `z=None` in the code below, 2 entries remain. We believe that this is due to garbage collection not properly triggering the objects to be destroyed. If `ak.clear()` is run, the symbol table is then empty. We need to run down why these entries are not being destroyed as expected.

The below code can be used to reproduce.
```python
z = ak.Categorical(ak.array(["a", "b", "c"]))
z = None
ak.list_symbol_table()
```

Ethan-DeBandi99:
Closes #2163

Updates the `GroupBy` client class to send a `delete` command to the Arkouda Server to delete the associated `GroupBySymEntry` if one exists when the client destruction is called. 

Updates a workflow on the server to remove `SymEntries` from the Symbol Table after they are used. The code is calling a function that requires the entries be added to the symbol table, so we just remove them after they are used.

Adds testing to ensure that when the Categorical object is deleted, the associated components in the Symbol Table are removed.

### DIFF
--- a/src/UniqueMsg.chpl
+++ b/src/UniqueMsg.chpl
@@ -164,6 +164,12 @@ module UniqueMsg
         return (permutation, segments);
       }
 
+      inline proc cleanup(str_names: [] string, st) throws {
+        forall name in str_names {
+          st.deleteEntry(name);
+        }
+      }
+
       if hasStr && n == 1 {
         // Only one array which is a strings
         var (myNames, _) = namesList[0].splitMsgToTuple("+", 2);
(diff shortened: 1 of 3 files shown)


pierce314159 (approved):
Looks good!

src/UniqueMsg.chpl
```diff
@@ -164,6 +164,12 @@ module UniqueMsg
         return (permutation, segments);
       }
 
+      inline proc cleanup(str_names: [] string, st) throws {
+        forall name in str_names {
```
ronawho, line 168: I don't think the symbol table is parallel safe, so probably want this to be serial.

Ethan-DeBandi99: @ronawho you are totally right. I have it written as a for, but did not save the file before pushing that is my bad. Should be updated now.

jaketrookman (approved):
looks good

joshmarshall1 (approved):

PR #2176, opened 2023-02-27
Closes #2175: Add check for local domain size before passing as C pointer

### LINKED ISSUE #2175: Writing a Parquet dataframe results in OOB access when number of locales exceed column numbers
The code `ptrList[i] = c_ptrTo(e.a[locDom]): c_void_ptr;` causes an OOB access when `locDom` is an empty slice, so we need to be checking if `locDom` is empty prior to trying to pass it to a `c_ptr`.

bmcdonald3:
In the case where an Arkouda server is running with more locales than the number of columns in a dataframe, Arkouda is attempting to pass a C pointer to an empty slice, which results in an OOB error in Chapel.


Ethan-DeBandi99 (approved):
Looks good.

PR #2178, opened 2023-02-27
Improve bigint `>>` performance

ronawho:
Avoid assigning bigint arrays from a single bigint value, which results in lot of communication since that value lives on a single locale.

16-node-cs-hdr results for a small problem size:

Before:
```
./benchmarks/run_benchmarks.py bigint_bitwise_binops -nl 16 --size=1000000
Average bigint SHIFT time = 3.8234 sec
Average bigint SHIFT rate = 0.06 GiB/sec
```

After:
```
./benchmarks/run_benchmarks.py bigint_bitwise_binops -nl 16 --size=1000000
Average bigint SHIFT time = 0.0950 sec
Average bigint SHIFT rate = 2.51 GiB/sec
```

And with a larger (default) problem size after:
```
./benchmarks/run_benchmarks.py bigint_bitwise_binops -nl 16 --size=$((10**8))
Average bigint SHIFT time = 7.4676 sec
Average bigint SHIFT rate = 3.19 GiB/sec
```

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -1171,10 +1171,9 @@ module BinOp
             // workaround for right shift until chapel issue #21206
             // makes it into a release, eventually we can just do
             // tmp = la >> ra;
-            var divideBy = makeDistArray(la.size, bigint);
-            divideBy = 1:bigint;
-            divideBy <<= ra;
-            forall (t, dB) in zip(tmp, divideBy) with (var local_max_size = max_size) {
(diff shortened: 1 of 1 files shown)


ronawho:
Note that this should resolve a timeout we're seeing in nightly perf testing

pierce314159 (approved):
This looks great to me!

jaketrookman (approved):
Looks good

PR #2188, opened 2023-03-03
Closes #2110 - Updates `_buildReadAllJSON` to use Map

### LINKED ISSUE #2110: `_buildReadAllJSON` to use Map
When it was written `GenSymIO._buildReadAllJSON` was configured to build a json string manually. This was done because using `%jt` to write a map out to a JSON string was not support. `%jt` is now supported for writing a Map to a JSON string. 

This function is currently a maintenance headache because it is cumbersome to add additional types to be written out properly. This should be updated to generate a Map and call `"%jt".format(map)` to generate the JSON string returned to the client. This will make maintainability moving forward much easier.

Ethan-DeBandi99:
Closes #2110

Updates the workflow for `_buildReadAllJSON` to use Chapel Maps and Lists instead of manually building the JSON string. This is the first step in moving away from `string` based returns and towards fully JSON formatted messaging.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -460,7 +460,7 @@ def _build_objects(
     RuntimeError
         - If no objects were returned
     """
-    items = rep_msg["items"] if "items" in rep_msg else []
+    items = json.loads(rep_msg["items"]) if "items" in rep_msg else []
     # We have a couple possible return conditions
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
This looks way better!

jaketrookman (approved):
Looks good

PR #2190, opened 2023-03-03
Closes #2083 - Adds `get_filetype` support for CSV

### LINKED ISSUE #2083: CSV supported by `get_filetype`
For the initial implementation, CSV did not support the `get_filetype` method. This issue is to track adding in that functionality. Doing so will allow us to integrate CSV into some of the generic file io functionality.

Currently, I am think this will require the Arkouda generated header to be most efficient.

Ethan-DeBandi99:
Closes #2083

- Adds support to `get_filetype` for CSV files. This will only work if when the file was written by Arkouda and contains the Arkouda Header used in CSV files.
- Integrates CSV support into `ak.io.load`, `ak.io.load_all`, `ak.io.read`, `ak.io.ls`, and `ak.io.get_datasets`. This adds the `column delimiter` as a parameter, which is only used when the file is a CSV.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -1612,6 +1612,8 @@ def read(
             allow_errors=allow_errors,
         )
     elif ftype.lower() == "csv":
-        return read_csv(filenames, datasets=datasets, column_delim=column_delim, allow_errors=allow_errors)
+        return read_csv(
+            filenames, datasets=datasets, column_delim=column_delim, allow_errors=allow_errors
+        )
     else:
         raise RuntimeError(f"Invalid File Type detected, {ftype}")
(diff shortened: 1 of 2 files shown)


arkouda/io.py
```diff
@@ -1565,5 +1611,7 @@ def read(
             strict_types=strictTypes,
             allow_errors=allow_errors,
         )
+    elif ftype.lower() == "csv":
+        return read_csv(filenames, datasets=datasets, column_delim=",", allow_errors=allow_errors)
```
pierce314159, line 1615: this should be `column_delim=column_delim` right?

pierce314159 (changes requested):
Looks good other than one spot where we hard code column_delim instead of using value passed in

arkouda/io.py
Ethan-DeBandi99: Yes. Initially, I was going to have these use the default and not add the parameter to the functions, but changed my mind and forgot to update here.

pierce314159 (approved):

jaketrookman (approved):
Looks good

PR #2192, opened 2023-03-06
Closes #2086 - Cleans up extra `SymEntry` creation calls in `SegmentedMsg.chpl`

### LINKED ISSUE #2086: Clean up additional `SymEntry` creations in `SegmentedMsg.chpl`
In `SegmentedMsg.chpl` there are a couple instances where we are creating new `SymEntries` for processing where we could likely get away with not having to create new ones. Found in PR #2080 

First in `assembleSegArrayMsg()`: 
Per @Ethan-DeBandi99 "I did notice that we could use `assembleSegStringFromParts(segs, vals)` and just use the `SymEntry` instead of creating additional ones"

Second in `getSANonEmptyMsg()`
Per @Ethan-DeBandi99 "I think we could do this without creating a new `SymEntry` ... something like:"
```chapel
var segArr = getSegArray(name, st, type);
repMsg = "created " + st.attrib(name) + "+" + segArr.getNonEmptyCount():string;
```

joshmarshall1:
This PR Closes #2086 

Small changes to remove extra `SymEntry` creation calls and update a deprecated method in the `segarray_test` file.

The first of the two recommendations outlined in #2086 was found in testing to be not beneficial and would add additional complexity instead of removing complexity.

### DIFF
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -91,33 +91,27 @@ module SegmentedMsg {
     select genEntry.dtype {
       when (DType.Int64) {
         var segArr = getSegArray(name, st, int);
-        st.addEntry(neName, new shared SymEntry(segArr.getNonEmpty()));
-        repMsg = "created " + st.attrib(neName) + "+" + segArr.getNonEmptyCount():string;
+        repMsg = "created " + st.attrib(name) + "+" + segArr.getNonEmptyCount():string;
       }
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
I dont see any issues with this

jaketrookman (approved):
Looks good

Ethan-DeBandi99 (approved):
Looks good

PR #2197, opened 2023-03-06
Closes #2179 - Fixes OOB Error when writing SegArray to HDF5

### LINKED ISSUE #2179: OOB error when writing to HDF5 when number of locales exceeds number of columns in dataframe
[`var startValIdx = localSegs[locDom.low];`](https://github.com/Bears-R-Us/arkouda/blob/6ada664ab7ffc29b920ff5cba59d77a9325750d0/src/HDF5Msg.chpl#LL925C13-L925C53) is resulting in an OOB exception when the local subdomain is empty.

Ethan-DeBandi99:
Closes #2179

- Updates the SegArray write workflow for HDF5 to generate an empty file in the event that a locale contains no segments.
- Renames `writeNilStringsGroupToHdf` to `writeNilSegmentedGroupToHdf` and adds a ctype parameter to pass the value returned from `getDataType` or `getHDF5Type`.
- Applies a fix to `fixupSegBoundaries` to ensure that when a locale has no segments, values at index 0 are not overwritten.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -696,13 +696,15 @@ module HDF5Msg {
      * :arg writeOffsets: boolean switch for whether or not to write offsets/segements to file
      * :type writeOffsets: bool
      */
-    private proc writeNilStringsGroupToHdf(fileId: int, group: string, writeOffsets: bool) throws {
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Looks good!

jaketrookman (approved):
Looks good to me

Issue #2201, opened 2023-03-08
Quiet deprecation warnings in prep for Chapel 1.30


bmcdonald3:
There are several new deprecation warnings with the 1.30 release. Quiet them in preparation for upgrading to 1.30 when it's released.

PR #2202, opened 2023-03-08
Closes 2201: Quiet deprecations in prep for Chapel 1.30

bmcdonald3:
There are a number of new deprecations in Chapel 1.30. Quiet them here while maintaining compatibly (and quiet builds) for 1.28/1.29. The specific changes are:

- Handle `getCurrentTime` deprecation. `getCurrentTime` is deprecated in favor of `timeSinceEpoch().totalSeconds()`
- Handle `iomode` deprecation. `iomode` has been deprecated in favor of `ioMode`
- Handle `openmem` deprecation. `openmem` has been deprecated in favor of `openMemFile`
- Handle `map.items` deprecation. `map.items` has been deprecated in favor of zipping over the keys and values
- Handle `map.get*` deprecations. `map.get*` accessors have been deprecated in favor of `proc this`
- Handle `newBlockDom` deprecations. `newBlockDom` has been deprecated in favor of `Block.createDomain`
- Handle `Regex.compile` deprecation. `Regex.compile` has been deprecated in favor of `new regex()`
- Handle `array.find` deprecation. `array.find` no longer returns a tuple and instead takes a reference variable that will store the index if found
- Handle `file.lines` deprecation. `file.lines` has been deprecated in favor of `fileReader.lines`, which can be gotten from `file.reader().lines()`
- Handle `readbytes` deprecation. `readbytes` has been deprecated in favor of `readBytes`
- Handle throwing `bigint` initializers compilation flag. `bigint` initializers that don't throw have been deprecated in favor of the new throwing initializers, which must be opted into with -sbigintInitThrows=true`

Closes https://github.com/Bears-R-Us/arkouda/issues/2201


src/compat/e-129/ArkoudaRegexCompat.chpl
```diff
@@ -0,0 +1,30 @@
+module ArkoudaRegexCompat {
+  import Regex.regex as chapelRegex;
+  import Regex.regexMatch;
+  import Regex.compile;
+
+  record regex {
```
ronawho, line 6: We don't want to compile the regex for each search/match. I think since we couldn't throw in <=1.29 initializers, let's just stick with the `compile` interface for now (and add a comment here about why it's needed)

ronawho (commented):
Left some inline comments, other thing that popped out is that we'll want to use the 130 shims for >=1.30 (since main will be named 1.31 soon) and it'd be nice if we could combine 1.28/1.29 shims and use them for <= 1.29

ronawho (approved):

pierce314159 (approved):

ronawho:
Looks like merging is waiting for 1.27 testing, but that's removed in this PR. Might have to manually merge?

bmcdonald3:
I don't think there is a need to open up an issue for removing the compatibility modules. I plan to continue doing these for each Chapel release (and then, once Chapel 2.0 is out, we shouldn't need them anymore) and, like in this PR, I'll remove the old ones as I go. An issue wouldn't trigger any more remembrance on my part.

Ethan-DeBandi99:
> I don't think there is a need to open up an issue for removing the compatibility modules. I plan to continue doing these for each Chapel release (and then, once Chapel 2.0 is out, we shouldn't need them anymore) and, like in this PR, I'll remove the old ones as I go. An issue wouldn't trigger any more remembrance on my part.

Sounds good to me. Just wanted to check.

bmcdonald3:
I am not sure what's going on with the failing CI, seems to be a failure with the Arrow dependency (unrelated to this PR)... I'll try to figure that out.

PR #2204, opened 2023-03-08
Closes #2203: Fix bigint shifting negative values for 1.30 release

### LINKED ISSUE #2203: Fix bigint shifting negative values for 1.30 release
In the 1.30 release, there was a bigint bug that was uncovered surrounding shifting bigints of/with negative values. The current behavior that was working in Arkouda was relying on the buggy behavior and needs to be updated.

bmcdonald3:
In the 1.30 release, there was a bigint bug that was uncovered surrounding shifting bigints of/with negative values. The current behavior that was working in Arkouda was relying on the buggy behavior, so this fixes that (works on both 1.29 and 1.30).

Closes: #2203 

See: https://github.com/chapel-lang/chapel/pull/21784

### DIFF
--- a/src/BigIntMsg.chpl
+++ b/src/BigIntMsg.chpl
@@ -77,7 +77,7 @@ module BigIntMsg {
                   all_zero = true;
                   forall t in tmp with (&& reduce all_zero) {
                     t >>= ushift;
-                    all_zero &&= (t == 0);
+                    all_zero &&= (t <= 0);
                   }
                 }
                 var repMsg = "%jt".format(retList);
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Thanks for this!

`bigint_to_uint_arrays` when the bigint array contains negatives does seem like a weird case. I think this will return the absolute value of the bigint array represented as a list of uint arrays which seems sensible to me. I might update the doc string to indicate this

bmcdonald3:
> I might update the doc string to indicate this

I don't think this impacts the behavior of Arkouda at all, everything should still return the same as it did before. The behavior shouldn't be changing from previously at all, this is just something that was changed in Chapel (i.e., over right shifting bigint used to return `0`, but it returns `-1` now)

Ethan-DeBandi99 (approved):

PR #2205, opened 2023-03-08
Closes #2181: Bigint array transfer benchmark

### LINKED ISSUE #2181: bigint client <-> server transfer benchmark
Add a benchmark to measure performance of bigint client <-> server transfers

pierce314159:
This PR (closes #2181) adds bigint array transfer benchmark. The default problem size is set to 10**6 since I think performance is going to have difficulty scaling and I don't want to timeout the nightly

```
% ./benchmarks/run_benchmarks.py bigint_array_transfer -t 1 --size=100000000
array size = 100,000,000
number of trials =  1
>>> arkouda bigint array creation
numLocales = 1, N = 100,000,000
to_ndarray Average time = 26.5110 sec
ak.array Average time = 28.8146 sec
to_ndarray Average rate = 0.0562 GiB/sec
ak.array Average rate = 0.0517 GiB/sec

% ./benchmarks/run_benchmarks.py bigint_array_transfer -t 1 --size=100000000 --max-bits=64
array size = 100,000,000
number of trials =  1
>>> arkouda bigint array creation
numLocales = 1, N = 100,000,000
to_ndarray Average time = 10.9486 sec
ak.array Average time = 8.3930 sec
to_ndarray Average rate = 0.0681 GiB/sec
ak.array Average rate = 0.0888 GiB/sec

```


ronawho:
> I think performance is going to have difficulty scaling

There could certainly be some, but I'm not expecting any offhand. Between client and server it's transferring uints, which should be optimized/scalable. The conversion between bigint and uint on the server side has been optimized, so that shouldn't be too bad. The client side conversion between bigint and uint might not be optimal, but that's only done by the client and won't be impacted by scaling the number of nodes (it's speed should just be based on the total number of bytes, not how many server nodes are used)

PR #2208, opened 2023-03-10
Closes #2207: Create task private copies of values for pdarray=value

### LINKED ISSUE #2207: Create task private copies of values for pdarray=value
Now that bigint is using the shared code with the other types for
indexing implemented in #2081, there was a bug uncovered in the
bigint module failing with remote assignments that will be fixed
in 1.30. For 1.29, using task-private copies is a suitable
workaround to avoid the remote assignments.

Also, this will result in a performance improvement for bigint
indexing, as discussed in
https://github.com/Bears-R-Us/arkouda/pull/2157#discussion_r1110440741.

bmcdonald3:
Now that bigint is using the shared code with the other types for indexing implemented in #2081, there was a bug uncovered in the bigint module failing with remote assignments that will be fixed in 1.30. For 1.29, using task-private copies is a suitable workaround to avoid the remote assignmnets.

Also, this will result in a performance improvement for bigint indexing, as discussed in
https://github.com/Bears-R-Us/arkouda/pull/2157#discussion_r1110440741.

Closes: #2207

### DIFF
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -690,8 +690,8 @@ module IndexingMsg
             // [i in iv.a] e.a[i] = val;
             ref iva = iv.a;
             ref ea = e.a;
-            forall i in iva with (var agg = newDstAggregator(dtype)) {
-              agg.copy(ea[i],val);
+            forall i in iva with (var agg = newDstAggregator(dtype), var locVal = val) {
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Looks great! Thanks ben

jaketrookman (approved):
Looks good to me

Issue #2213, opened 2023-03-14
Configure benchmarks to use pytest-benchmark


joshmarshall1:
Through exploring options for benchmark setups, we found pytest-benchmark and decided this gives us a level of granularity in the data output we're happy with. This issue will track the progress of updating the existing benchmarks to use the pytest-benchmark approach

Ethan-DeBandi99:
For each benchmark we want to be sure to include the following informations in `benchmark.extra_info`:
- problem size
- Description of benchmark
- data transfer rate

Ethan-DeBandi99:
Adding a note to remind us to update/add documentation on benchmarks. I don't think we want to remove our existing docs while we still have the legacy benchmark platform. However, we definitely should document and provide examples of utilization of the `pytest-benchmark` format. I have come up with a list below to hopefully guide us in the right direction.

- [ ] How to run all benchmarks.
- [ ] How to run individual benchmarks
- [ ] How to use custom commandline parameters
- [ ] What custom commandline parameters are used for all tests and which only apply to a subset. If they only apply to a subset list the benchmarks they are used for.
- [ ] Any useful tips & tricks for developers who may be adding new benchmarks.
- [ ] How to read the JSON output and what information will be most important for our purposes

Ethan-DeBandi99:
Closing since additional work in comments is being tracked in separate issues.

PR #2218, opened 2023-03-15
Part of #2213 - Configuration & Example of PyTest Benchmark

Ethan-DeBandi99:
Part of #2213 

Sets the configuration for using PyTest Benchmark for benchmarking. An example benchmark is included.

This PR configures the `make benchmark` command and configures components needed for benchmarking. New benchmarks should be added to `benchmark.ini` to be executed. Problem size can be set with the `ARKOUDA_PROBLEM_SIZE` env var and accessed from benchmarks using `pytest.problem_size`.


benchmark_v2/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+)
+
+
+@pytest.fixture(scope="module", autouse=True)
+def startup_teardown():
+    port = int(os.getenv("ARKOUDA_SERVER_PORT", 5555))
+    server = os.getenv("ARKOUDA_SERVER_HOST", "localhost")
+    test_running_mode = TestRunningMode(os.getenv("ARKOUDA_RUNNING_MODE", "GLOBAL_SERVER"))
+
+    if not importlib.util.find_spec("pytest") or not importlib.util.find_spec("pytest_env"):
```
joshmarshall1, line 21: Should we add `pytest-benchmark` to this import check?

joshmarshall1 (approved):
This all looks good. Only one question

pierce314159 (approved):
looks good

PR #2224, opened 2023-03-16
Closes #2220 - PyTest Benchmark for `ak.argsort`

### LINKED ISSUE #2220: PyTest Benchmark `argsort.py`
Configure the pytest benchmark for `benchmarks/argsort.py`

Ethan-DeBandi99:
Closes #2220 

- Adds pytest benchmark for `ak.argsort`.
- Adds command line options for benchmarks. These are added to `tests/conftest.py` in order to function due to the way pytest looks for and utilizes conftest.py files. This should not be an issue as the eventual plan is to move to a uniform system only having a single `conftest.py` file to maintain once our tests can be updated.
- Allows for setting more than 1 dtype. Our old benchmarks required you to run for a single dtype. This allows for running with as many of the supported dtypes as needed.
- All command line arguments are set as pytest variables. For example, using `--size x` will trigger the `pytest.prob_size` variable to be set to `x`. All variables configured this way can be accessed by importing `pytest`.
- The `--check-correctness` option was not moved over because these checks are going to be moved into the main testing suite. Again, once this is being done, we will most likely make use of the options supplied for benchmarks.
- All command line args are pulled as strings, so the proper parsing is added to ensure correct values for usage.
- The `--numpy` arg is added to allow for running numpy comparisons. However, the comparison in `argsort` is not running anything, so it was not ported over.

**Example Call**
This example demonstrates a call setting all flags.
```bash
python3 -m pytest -c benchmark.ini --size="10**6" --trials=10 --seed="2**20" --dtype=int64,float64
```

**Output from Example Call**
```bash
benchmark_v2/argsort_benchmark.py ....                                                                                                                                             [100%]

==================================================================================== warnings summary ====================================================================================
benchmark_v2/argsort_benchmark.py::bench_argsort[uint64]
benchmark_v2/argsort_benchmark.py::bench_argsort[str]
  benchmark_v2/argsort_benchmark.py:6: PytestBenchmarkWarning: Benchmark fixture was not used at all in this test!
    @pytest.mark.parametrize("dtype", TYPES)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html

-------------------------------------------------------------------------------------- benchmark: 2 tests -------------------------------------------------------------------------------------
Name (time in ms)               Min                 Max                Mean             StdDev              Median               IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_argsort[int64]        72.3735 (1.0)      111.7408 (1.0)       78.1717 (1.0)      12.0035 (2.29)      74.6366 (1.0)      4.3496 (1.66)          1;1  12.7924 (1.0)          10           1
bench_argsort[float64]     130.5417 (1.80)     149.0939 (1.33)     140.1040 (1.79)      5.2396 (1.0)      138.8819 (1.86)     2.6208 (1.0)           3;3   7.1376 (0.56)         10           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================================================================= 4 passed, 2 warnings in 4.55s ==============================================================================
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,6 +10,31 @@ def pytest_addoption(parser):
         "--optional-parquet", action="store_true", default=False, help="run optional parquet tests"
     )
 
+    # due to the order pytest looks for conftest in, options for our benchmarks are added here
+    # this has no impact on testing, but allows quick and easy access to settings without the need
+    # for env vars. Everything below is for workflows in arkouda/benchmark_v2
+    parser.addoption(
+        "--size", action="store", default=10**8, type=int,
+        help="Benchmark only option. Problem size: length of array to use for benchmarks."
+    )
+    parser.addoption(
+        "--trials", action="store", default=5, type=int,
+        help="Benchmark only option. Problem size: length of array to use for benchmarks."
+    )
+    parser.addoption(
+        "--seed", action="store", default=None, type=int,
+        help="Benchmark only option. Value to initialize random number generator."
+    )
+    parser.addoption(
+        "--dtype", action="store", default=None, type=str,
+        help="Benchmark only option. Dtypes to run benchmarks against. Comma separated list "
+             "(NO SPACES) allowing for multiple. Accepted values: int64, uint64, float64, bool, and str."
+    )
+    parser.addoption(
+        "--numpy", action="store_true", default=False, type=bool,
+        help="Benchmark only option. When set, runs numpy comparison benchmarks."
+    )
+
 
 def pytest_collection_modifyitems(config, items):
     if config.getoption("--optional-parquet"):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,11 +22,11 @@ def pytest_addoption(parser):
         help="Benchmark only option. Problem size: length of array to use for benchmarks."
     )
     parser.addoption(
-        "--seed", action="store", default="None",
+        "--seed", action="store", default="",
         help="Benchmark only option. Value to initialize random number generator."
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -10,6 +10,31 @@ def pytest_addoption(parser):
...
+    )
+    parser.addoption(
+        "--dtype", action="store", default="",
+        help="Benchmark only option. Dtypes to run benchmarks against. Comma separated list "
+             "(NO SPACES) allowing for multiple. Accepted values: int64, uint64, float64, bool, and str."
+    )
+    parser.addoption(
+        "--numpy", action="store_true", default=False,
+        help="Benchmark only option. When set, runs numpy comparison benchmarks."
+    )
```
pierce314159, line 36: these are for all benchmarks, not just argsort right? I think that is the case and if so this should be good

pierce314159 (approved):
Looks good to me

tests/conftest.py
Ethan-DeBandi99: Yes the options are for all benchmarks. This way we don't define a parser in each benchmark like with the old setup

joshmarshall1: Yes these are for all benchmarks

PR #2238, opened 2023-03-16
Closes #2235 - PyTest Benchmark for stream

### LINKED ISSUE #2235: PyTest Benchmark for Stream
Configure pytest benchmark for bigint_stream.py and stream.py

joshmarshall1:
This PR closes #2235 

- Adds benchmark `stream_benchmark.py`
- Combines `bigint_stream.py` and `stream.py` benchmarks into a single benchmark


Ethan-DeBandi99 (changes requested):
It does not look like the `--randomize` option was pulled over here. We should maintain that in the event that you want a random set, but don't set the `seed`.

joshmarshall1:
> It does not look like the `--randomize` option was pulled over here. We should maintain that in the event that you want a random set, but don't set the `seed`.

Added in the --randomize flag

Ethan-DeBandi99 (approved):

tests/conftest.py
```diff
@@ -39,6 +39,14 @@ def pytest_addoption(parser):
         help="Benchmark only option. Only applies to bigint testing."
              "Maximum number of bits, so values > 2**max_bits will wraparound. -1 is interpreted as no maximum."
     )
+    parser.addoption(
+        "--alpha", action="store", default="1",
```
pierce314159, line 43: in the original alpha is a real value, i just wanna make sure we'll still be multiplying by `1.0` here

joshmarshall1: Added the `1.0` multiplication to the non-bigint cases - verified with @pierce314159 that bigint can't accept a `real` multiplication so it has to remain `int` in that case

pierce314159 (approved):

jaketrookman (approved):

PR #2249, opened 2023-03-17
Closes #2243 - Initial implementation of dataframe benchmark updates

### LINKED ISSUE #2243: PyTest Benchmark for `dataframe.py`
Configure pytest benchmark for dataframe.py

joshmarshall1:
This PR closes #2243 

Below issue has been resolved, maintaining description here for record.

Potential issue with performance when setting segarray generation to 

`ak.arange(0, N * 5, 5), ak.array(np.random.randint(0, 2**32, N * 5))`

as it is in the original benchmark. removing the `* 5` appears to work, but it's worth exploring some more

### DIFF
--- a/benchmark_v2/dataframe_indexing_benchmark.py
+++ b/benchmark_v2/dataframe_indexing_benchmark.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+import numpy as np
+import pandas as pd
+import pytest
+
+import arkouda as ak
+
+OPS = ["_get_head_tail_server", "_get_head_tail"]
+
+
+def generate_dataframe():
+    cfg = ak.get_config()
+    N = pytest.prob_size * cfg["numLocales"]
+    types = [ak.Categorical, ak.pdarray, ak.Strings, ak.SegArray]
+
+    # generate random columns to build dataframe
+    df_dict = {}
+    np.random.seed(pytest.seed)
+    for x in range(20):  # loop to create 20 random columns
+        key = f"c_{x}"
+        d = types[x % 4]
+        if d == ak.Categorical:
+            str_arr = ak.random_strings_uniform(minlen=5, maxlen=6, size=N, seed=pytest.seed)
+            df_dict[key] = ak.Categorical(str_arr)
+        elif d == ak.pdarray:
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99 (commented):
@joshmarshall1 - looking at this I think it really comes down to the problem size blowing up with the `*5`. I am able to consistently run the full benchmark without it. Also, I do not think that we need the `*5`. This is similar to the idea of not multiplying the by the number of locales for numpy benchmarks. I think if we remove the `*5` from the legacy benchmark and fully configure this we should be good.

I wanted to also add that we should be adding the segments size into the data transfer rate record because that is a factor here.

jaketrookman (approved):

benchmark_v2/dataframe_indexing_benchmark.py
```diff
@@ -0,0 +1,69 @@
...
+        if d == ak.Categorical:
+            str_arr = ak.random_strings_uniform(minlen=5, maxlen=6, size=N, seed=pytest.seed)
+            df_dict[key] = ak.Categorical(str_arr)
+        elif d == ak.pdarray:
+            df_dict[key] = ak.array(np.random.randint(0, 2**32, N))
+        elif d == ak.Strings:
+            df_dict[key] = ak.random_strings_uniform(minlen=5, maxlen=6, size=N, seed=pytest.seed)
+        elif d == ak.SegArray:
+            df_dict[key] = ak.segarray(
+                ak.arange(0, N), ak.array(np.random.randint(0, 2**32, N))
```
pierce314159, line 31: Should we keep this `ak.arange(0, N, 5)`? That way we have more than one segment in our segarray

joshmarshall1: I'm not opposed to adding the step of 5 back. But just to clarify what this is doing, it is increasing the size of each segment to 5 values per segment instead of 1 value per segment.

pierce314159: Ah yes! you're right, my bad

joshmarshall1: After further discussion with @Ethan-DeBandi99 and @pierce314159, we've decided to leave this as it is for now

PR #2259, opened 2023-03-20
Closes #2248 - Pytest Benchmark for Gather

### LINKED ISSUE #2248: Pytest Benchmark for Gather
Configure the pytest-benchmark for gather performance corresponding to `benchmarks/gather.py`.

Ethan-DeBandi99:
Closes #2248 

- Adds pytest-benchmark for gather performance. Benchmarks for Arkouda and NumPy performance included.
- Adds `--idx_size` and `--val_size` command line parameters to set the size of the values and index individually. If they are not provided, defaults to `pytest.prob_size`.

### DIFF
--- a/benchmark_v2/gather_benchmark.py
+++ b/benchmark_v2/gather_benchmark.py
@@ -1,0 +1,97 @@
+import arkouda as ak
+import pytest
+import numpy as np
+
+TYPES = ("int64", "float64", "bool", "str")
+
+
+def _ak_gather(a, i):
+    """
+    Helper function allowing for the gather to be benchmarked
+    """
+    return a[i]
+
+
+@pytest.mark.benchmark(group="AK_Gather")
+@pytest.mark.parametrize("dtype", TYPES)
+def bench_ak_gather(benchmark, dtype):
+    cfg = ak.get_config()
+    isize = pytest.prob_size if pytest.idx_size is None else pytest.idx_size
+    vsize = pytest.prob_size if pytest.val_size is None else pytest.val_size
+    Ni = isize * cfg["numLocales"]
+    Nv = vsize * cfg["numLocales"]
+
+    if dtype in pytest.dtype:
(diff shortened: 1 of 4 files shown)


benchmark_v2/gather_benchmark.py
```diff
@@ -0,0 +1,97 @@
...
+        c = benchmark.pedantic(_ak_gather, args=[v, i], rounds=pytest.trials)
+
+        if dtype == "str":
+            offsets_transferred = 3 * c.size * 8
+            bytes_transferred = (c.size * 8) + (2 * c.nbytes)
+            bytes_per_sec = (offsets_transferred + bytes_transferred) / benchmark.stats["mean"]
+        else:
+            bytes_per_sec = (c.size * c.itemsize * 3) / benchmark.stats["mean"]
+
+        bytes_per_sec / 2 ** 30
```
joshmarshall1, line 52: I don't think we need this here since we're doing the same operation in the transfer rate comment and this isn't being used

joshmarshall1 (dismissed):

benchmark_v2/gather_benchmark.py
pierce314159, line 52: does this line do anything? The result isn't being stored anywhere

benchmark_v2/gather_benchmark.py
```diff
@@ -0,0 +1,97 @@
...
+        benchmark.extra_info["value_size"] = vsize
+        benchmark.extra_info["transfer_rate"] = "{:.4f} GiB/sec".format(
+            (bytes_per_sec / 2 ** 30))
+
+
+def _np_gather(a, i):
+    """
+    Helper function allowing for the gather to be benchmarked
+    """
+    return a[i]
```
pierce314159, line 65: I feel like we could use the same function for both `_ak_gather` and `_np_gather`

pierce314159 (approved):

jaketrookman (approved):
Looks good

PR #2267, opened 2023-03-21
Closes #2263 - Fixes Locality of Segment Compution

### LINKED ISSUE #2263: v.2023.03.01 issue with large segarray
**Describe the bug**
Some large dataframes which include segarrays cannot be read in if saved with v.2023.03.01. If files containing the same internal data are created using v.2023.02.08 then the files can be read into either of these versions.

Reading with v.2023.03.01
-- results in server appearing to hang. No error messages are produced, Just no new lines are written to log and command doesn't complete. Server must be killed manually. It won't accept connections. (Normal "busy server" behavior)

Reading with v.2023.02.08 
-- results in the following error lines
RuntimeError
/proj/HPDA/software/arkouda/v2023.02.08/arkouda/io.py in load(path_prefix, file_format, dataset, calc_string_offsets)  line 1366

RuntimeError: Other error in accessing file <filepath>: Error 1661 HDF5Msg:get_info in get_info illegal argument 'bad cast from int '3' to enum 'ObjType': '

Related specifically to segarrays
Files can be read using ak.read(<filepath*>,datasets=<column_list_excluding_segarrays>) using either version.

Also, appears to be size specific. Writing out data frames in 1M record slices with v.2023.03.01 results in files that can each be read back in individually, but writing out 50M record slice creates file that cannot be read back in.

**To Reproduce**
have not been able to reproduce the behavior in simple fashion

**Expected behavior**

**Error Message**
If you are receiving any error messages, please provide them here.

**Is this a Blocking Issue**
Falling back to v.2023.02.08 is the only known workaround

**`ak.get_config()` Output**

**Additional context**

Ethan-DeBandi99:
Closes #2263

Modifies the computation of the adjustment of the segment indexes into the values array to have better locality. This results in much better performance when loading SegArray and Strings objects on distributed systems.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -1463,9 +1463,16 @@ module HDF5Msg {
         
         // compute amount to adjust 
         var adjustments = (+ scan diffs) - diffs;
-        forall(sd, adj) in zip(segSubdoms, adjustments) {
-            // adjust offset of the segment based on the sizes of the segments preceeding it
-            a[sd] += adj;
+        coforall loc in a.targetLocales() do on loc {
+            forall(sd, adj) in zip(segSubdoms, adjustments) {
+                for locdom in a.localSubdomains() {
+                    const intersection = domain_intersection(locdom, sd);
+                    if intersection.size > 0 {
+                        // adjust offset of the segment based on the sizes of the segments preceeding it
+                        a[intersection] += adj;
+                    }
+                }
+            }
         }
     }
 


pierce314159 (approved):
This looks good to me! If @ronawho doesn't mind taking a peek to verify that would be awesome!

Reading a segstring,`ak.random_strings_uniform(1, 16, 10**5)`, with 4 locales. I'm seeing these numbers, which looks way better. Way less time and comm

Before this PR:
```
fixup time = 18

| locale |   get |   put | execute_on_nb |
| -----: | ----: | ----: | ------------: |
|      0 | 75021 | 75000 |            24 |
|      1 |     0 |     0 |             0 |
|      2 |     0 |     0 |             0 |
|      3 |     0 |     0 |             0 |
```

After this PR:
```
fixup time = 0

|      0 |   0 |   3 |          0 |             9 |
|      1 |  49 |   0 |          1 |             6 |
|      2 |  49 |   0 |          1 |             6 |
|      3 |  49 |   0 |          1 |             6 |
```

To measure this I just updated `_buildEntryLoadOffsets()` to have
```chapel
use Time;
use CommDiagnostics;
startCommDiagnostics();
var t1 = Time.getCurrentTime();
fixupSegBoundaries(offsetsEntry.a, segSubdoms, subdoms);
var t2 = Time.getCurrentTime() - t1;
stopCommDiagnostics();
writeln("fixup time = %i".format(t2));
writeln();
printCommDiagnosticsTable();
```

pierce314159:
One potential alternative is to add aggregation (use a `forall` to expand the adjustments to the same size as `a` using aggregation and then add them into `a`). This seems to perform similarly to what you have (time is 0), and I'm not versed enough in communication diagnostics to know which is better. So it would actually be pretty informative to get @ronawho's insight into this

```chapel
var expanded_adj: [D] int;
forall (sd, adj) in zip(segSubdoms, adjustments) with (var agg = newDstAggregator(int)) {
    for i in sd {
        agg.copy(expanded_adj[i], adj);
    }
}
a += expanded_adj;
```

```
fixup time = 0

|      0 |  21 |  12 |         15 |            15 |
|      1 |  21 |   2 |          0 |             0 |
|      2 |  21 |   2 |          0 |             0 |
|      3 |  21 |   2 |          0 |             0 |
```

My knee jerk reaction is that this will be worse since we have to create an entire new array

I think it would really useful for future perf testing which of these metrics are most important? Like are GET/PUT more or less expensive than `excute_on`/`excute_on_nb`

ronawho:
Purely from a comm diagnostics perspective I couldn't guess which one will be faster. gets/puts are cheaper than ons (for systems like InfiniBand that have RDMA) but both the "improved diags" reports have so little comm that the comm is probably within the noise.

I'm not familiar with this code, but would be happy to go over this and make suggestions during dev meeting if that works.

joshmarshall1 (approved):

21771:
Seeing same behavior with 2023.03.24 release. Should this fix be in later release or might it be a different issue?

Ethan-DeBandi99:
@21771 - The should be in `v2023.03.24`. The only other thing that I can think of is that there is the possibility that reading large amounts of data out may have a hiccup somewhere after this. I will put in a new issue and track updates there. I suspect at this point it may be caused elsewhere in the code, but will have to verify.

PR #2269, opened 2023-03-21
Closes #2265 - Fixes `ak.DataFrame.to_parquet` with empty `Strings` Column

### LINKED ISSUE #2265: Server crashes when writing empty dataframe with string column
**Describe the bug**
Arkouda server crashes when trying to write an empty (zero rows) dataframe that has at least one string column

**To Reproduce**
ak.DataFrame({ "empty": ak.array(pd.array([], dtype=str))  }).to_hdf("empty.hdf")   # or to_parquet(); both fail

**Expected behavior**
Mirror pandas behavior and write out file with column names but no data
example: pd.DataFrame({'empty':np.array([]).astype('str')}).to_csv('empty.csv')

**Error Message**
From server logs:
*** Caught a fatal signal (proc 0): SIGSEGV(11)

**Is this a Blocking Issue**
row count can be tested before writing to save the server from crashing, but no known way to produce the desired files. Sometimes the presence of a file is the indication that process ran so it would be helpful to have even zero length files.

**`ak.get_config()` Output**

**Additional context**
This issue has come up while working on ETL stuff (sometimes filtering produces empty output). The presence of the files is often the indication that the process ran and that it does not need to be run again.

Ethan-DeBandi99:
Closes #2265 

Adds handling for when an empty `Strings` object is written to Parquet. Updates access of the `lastOffset` to default to `0` if there are no offsets. This allows the workflow to properly assign attributes in the file without hitting the index error.

Adds the same handling to Parquet workflows for `SegArray` as the same logic is used. The workflow for HDF5 is slightly different and does not require this change.

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -813,7 +813,7 @@ module HDF5Msg {
 
                 ref ss = segString;
                 var A = ss.offsets.a;
-                const lastOffset = A[A.domain.high];
+                const lastOffset = if A.size == 0 then 0 else A[A.domain.high]; // prevent index error when empty
                 const lastValIdx = ss.values.a.domain.high;
 
                 // For each locale gather the string bytes corresponding to the offsets in its local domain
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -560,7 +560,7 @@ module ParquetMsg {
     }
     
     const extraOffset = ss.values.size;
-    const lastOffset = A[A.domain.high];
+    const lastOffset = if A.size == 0 then 0 else A[A.domain.high]; // prevent index error when empty
     const lastValIdx = ss.values.a.domain.high;
     // For each locale gather the string bytes corresponding to the offsets in its local domain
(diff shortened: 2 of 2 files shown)


pierce314159 (changes requested):
Do we need to do the same kinda thing in `writeMultiColParquet`? We seem to have similar logic in all the `select` cases there

https://github.com/Bears-R-Us/arkouda/blob/a1bf9e7f501be65a1840ddd1d2e0078f405b4877/src/ParquetMsg.chpl#L1298

Ethan-DeBandi99:
> Do we need to do the same kinda thing in `writeMultiColParquet`? We seem to have similar logic in all the `select` cases there
> 
> https://github.com/Bears-R-Us/arkouda/blob/a1bf9e7f501be65a1840ddd1d2e0078f405b4877/src/ParquetMsg.chpl#L1298

Oh good catch! I will get it added to those cases as well and update.

joshmarshall1 (approved):

pierce314159 (approved):

PR #2275, opened 2023-03-22
Closes #2274: Separate bigintInitThrows flag from CHPL_FLAGS for deps

### LINKED ISSUE #2274: Separate bigintInitThrows flag from CHPL_FLAGS for deps
The `-sbigintInitThrows` flag is being appended to `CHPL_FLAGS` in the Makefile, meaning that it is being used when checking for C dependencies, which will not compile since those modules do not include the `BigInteger` module, which exposes the flag to be set in the command line.

bmcdonald3:
The `-sbigintInitThrows` flag was being appended to `CHPL_FLAGS` in the Makefile, meaning that it was being used when checking for C dependencies, which would not compile since those modules did not include the `BigInteger` module, which contains the flag.

Closes https://github.com/Bears-R-Us/arkouda/issues/2274


Makefile
```diff
@@ -315,7 +315,7 @@ ARKOUDA_SOURCES = $(shell find $(ARKOUDA_SOURCE_DIR)/ -type f -name '*.chpl')
 ARKOUDA_MAIN_SOURCE := $(ARKOUDA_SOURCE_DIR)/$(ARKOUDA_MAIN_MODULE).chpl
 
 ifeq ($(shell expr $(CHPL_MINOR) \>= 30),1)
-	CHPL_FLAGS += -sbigintInitThrows=true
+	CHPL_COMPAT_FLAGS += -sbigintInitThrows=true
```
ronawho, line 318: Could/should this use `OPTIONAL_SERVER_FLAGS`?

bmcdonald3: So you are saying, rather than `CHPL_COMPAT_FLAGS`, we'd use `OPTIONAL_SERVER_FLAGS`? 

I don't know if I would go that route, since it isn't really an "optional" flag (you need it with 1.30). Though, I guess the definition of "need" here is debatable, since it is just quieting a deprecation warning.

pierce314159 (approved):
This looks good to me! I don't have strong feelings on if it should be added to `CHPL_COMPAT_FLAGS` vs `OPTIONAL_SERVER_FLAGS`

ronawho (approved):

Ethan-DeBandi99 (approved):

jaketrookman (approved):

PR #2276, opened 2023-03-22
Closes #2272 - sort-cases PyTest benchmark

### LINKED ISSUE #2272: PyTest Benchmark for `sort-cases.py`
Configure pytest benchmark for `sort-cases.py`

joshmarshall1:
This PR closes #2272 

Adds the benchmarks for various sort configurations and generator settings

### DIFF
--- a/benchmark_v2/sort_cases_benchmark.py
+++ b/benchmark_v2/sort_cases_benchmark.py
@@ -1,0 +1,253 @@
+import arkouda as ak
+import numpy as np
+import pytest
+from arkouda.sorting import SortingAlgorithm
+
+TYPES = ("int64", "float64")
+POWERLAW_DATA = None
+
+
+def get_nbytes(data):
+    if isinstance(data, ak.pdarray):
+        return data.size * data.itemsize
+    elif isinstance(data, ak.Strings):
+        return data.size * 8 + data.nbytes
+    else:
+        return sum(get_nbytes(x) for x in data)
+
+
+def do_argsort(data, algo):
+    if isinstance(data, (ak.pdarray, ak.Strings)):
+        return ak.argsort(data, algo)
+    else:
+        return ak.coargsort(data, algo)
+
+
+@pytest.mark.benchmark(group="AK_Sort_Cases")
+@pytest.mark.parametrize("algo", SortingAlgorithm)
+@pytest.mark.parametrize("dtype", TYPES)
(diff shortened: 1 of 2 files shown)


benchmark_v2/sort_cases_benchmark.py
```diff
@@ -0,0 +1,179 @@
...
+@pytest.mark.benchmark(group="AK_Sort_Cases")
+@pytest.mark.parametrize("gen", GENERATORS)
+@pytest.mark.parametrize("algo", SortingAlgorithm)
+def bench_sort_cases(benchmark, gen, algo):
+    """
+    Measure both performance of various sort cases
+    """
+    res = gen(pytest.prob_size)
+    data = res.send(None)[1]
+    benchmark.pedantic(do_argsort, args=(data, algo), rounds=pytest.trials)
```
pierce314159, line 173: This won't work since these are generators, they could have more than one thing yielded (like in `block_sorted`, `power_law`, `random_uniform`)

At first I thought you were gonna have to do something like this:
```python
for _, data in gen(pytest.prob_size):
  benchmark.pedantic(do_argsort, args=(data, algo), rounds=pytest.trials)
```
But im realizing now that the parameterization would prob get messed up. Because it would record them all under the same generator (it doesn't know that `"block-sorted concat"` is different from `"block-sorted interleaved"`, it just knows they were both done using the parameter `block_sorted` so it will average their times)

I think the best way to do this in the new framework is to have a separate benchmark for each generator and then parameterize based on that, so the end result is more similar to the [graphs we have today](https://chapel-lang.org/perf/arkouda/16-node-cs-hdr/?startdate=2023/02/14&enddate=2023/03/21&configs=release,nightly&suite=sortcases). It's gonna be a bit tough because we want to use the same data in these cases (like in `power_law`, we don't wanna change the randomly generated data between the `"power-law float64"` and `"power-law int64"` cases. We want the only variable changed to be the type)

pierce314159 (changes requested):
This is gonna need a lot of reworking because the current generator structure doesn't place nice with the pytest framework. I think the best way forward is a different benchmark for each data generator and taking care for the data not to change between paramters

pierce314159 (approved):
This looks good! The only thing is if possible it'd be nice to preserve the data generated in the power law case

Ethan-DeBandi99 (approved):
Overall looks good.

jaketrookman (approved):

pierce314159 (approved):
Looks good!

PR #2279, opened 2023-03-23
Closes #2277 - Pytest Benchmark for String Locality

### LINKED ISSUE #2277: Pytest Benchmark for String Locality
Implement the pytest-benchmark configuration for string locality corresponding to `benchmarks/str-locality`.

Ethan-DeBandi99:
Closes #2277

Adds pytest-benchmark for Strings Locality corresponding to `benchmarks/str-locality.py`. 

Configures the operations as a dispatch table with lambda functions to easily test all operations with good and poor locality.


joshmarshall1 (approved):

jaketrookman (approved):
Looks good

benchmark_v2/str_locality_benchmark.py
```diff
@@ -0,0 +1,43 @@
...
+}
+
+# Good - generates random Strings object with "good" locality
+# poor - generates a sorted Strings object with "poor" locality
+LOCALITY = {"Good", "Poor"}
+
+
+def _generate_data(loc):
+    N = pytest.prob_size * ak.get_config()["numLocales"]
+    prefix = ak.random_strings_uniform(minlen=1, maxlen=16, size=N, seed=pytest.seed, characters="numeric")
```
pierce314159, line 18: It would be nice to use the same random data between parameters instead of generating new each time. I'm not sure if it's possible in this case

pierce314159 (approved):
This looks good! The only thing is if possible it'd be nice to preserve the data generated to be maintained between different `op` parameters

benchmark_v2/str_locality_benchmark.py
Ethan-DeBandi99: @pierce314159 - I created a global for random data and another for sorted data to handle both cases. I ran the benchmark and save the json with the name of the Strings object used. I verified that the name was identical for all poor locality tests and for all good locality tests. 

The only change I may want to make is to set both objects at once and only return the one needed. This would ensure that the data is the same just sorted or not. I think this would be good to do so I will look at that as well.

Ethan-DeBandi99: Decided to make the further test I mentioned previously. The code will now return the value desired if the objects were already created. Otherwise, it creates both objects so that the same data is random and sorted. This should ensure that the current benchmark is as close to the previous as possible.

pierce314159 (approved):

Issue #2288, opened 2023-03-23
Bigint aggregation next steps


ronawho:
Initial aggregation support is done, but there's some follow up improvements I'd like to explore

Potential implementation TODOs:
- [ ] Fix support for bigints with more limbs than we can store in aggregation buffers
- [ ] Explore support for mismatched bigint record/mpz locality
- [ ] Tune buffer size (currently sized to fit the same number of elements as plain old int64 aggregation to do comm comparisons)
- [ ] Improve serialization API (work through with Chapel team)
- [ ] See if we can/should use public functions (`mpz_limbs_write`, `mpz_limbs_finish`) added in GMP 6 -- https://gmplib.org/manual/Integer-Special-Functions

Potential testing TODOs:
- [ ] check negative numbers (gmp stores sign as part of size)
- [ ] check massive numbers (more limbs than we have buffer space for, broken currently)
- [ ] check additional variable sized limbs
- [ ] check with bigint record wrapper and heap-allocated mpz on different nodes (broken currently)
- [ ] check with limb sizes that will grow and shrink target bigint
- [ ] check with aggregating 0 values (no limbs), should work with current strategy, but needs special case with `mpz_limbs_write`
- [ ] check memleaks

ronawho:
Some of the testing TODOs are done with https://github.com/Bears-R-Us/arkouda/pull/2287, but I wanted to capture the original list and verify we covered everything when I get a chance.

PR #2290, opened 2023-03-23
Add initial aggregation support for `bigint`

### LINKED ISSUE #1994: Aggregation on `bigint`

### LINKED ISSUE #2212: Investigate perf overheads for bigint groupby
bigint groupby performance isn't looking great and parallel deinit doesn't seem to help:
https://chapel-lang.org/perf/arkouda/16-node-cs-hdr/?startdate=2023/03/04&enddate=2023/03/14&configs=release,nightly&graphs=bigintgroupbyperformance

I'm gonna tag team this with @ronawho or @bmcdonald3 to get better at discovering and addressing perf bottlenecks

ronawho:
Plain old integer DstAggregators locally aggregate the target address and the inline value by storing a tuple of `(c_ptr(int), int)` but it's not as simple with variable sized bigints. It's also not as easy to extract the raw value since bigints have metadata stored inline, but the values stored off in the heap. To enable aggregation, the buffers are no longer typed and are just raw bytes that we'll interpret. We now locally aggregate the target address, the 8-byte sign+size (int64 that stores the number of limbs and the sign of the limbs), and the variable sized limbs. This uses internal functions to expose the internal sign+size and limbs. With this we have our target address and "value" stored inline and can get data to the target side like normal. On the target side we then extract the address and size, and then resize the target bigint, copy the limbs over, and update the target sign+size.

SrcAggregator is a little more complicated since we don't know how big our src values are until we're on the remote node. Because of this we can't pre-allocate space for all our elements, so instead we just aggregate addresses locally and when those buffers are full we fill up value buffers as much as we can and just loop until we've processed all values.

This significantly improves the performance of operations that were using the aggregation interface. Testing with single limb (64-bit value) bigints, I see aggregation performance that's within 2x of plain old int(64) aggregation. This is around what I'd expect since we have to aggregate an extra 8 bytes for the sign_size and there's other overheads with real location and extracting limbs. Here's perforamnce for bigint groupby (which does a bigint gather) on 16-node-cs-hdr:

```
>>> arkouda bigint groupby
numLocales = 16, N = 16,000,000
2-array Average time = 16.3907 sec
2-array Average rate = 0.0145 GiB/sec

>>> arkouda bigint groupby
numLocales = 16, N = 16,000,000
2-array Average time = 0.9008 sec
2-array Average rate = 0.2647 GiB/sec

>>> arkouda bigint groupby
2-array Average time = 4.0616 sec
2-array Average rate = 5.8700 GiB/sec
```

I don't think there's much more performance we can squeeze out of this, but there is certainly some code cleanup and more testing I'd like to add in the future. That's captured in #2288

Resolves #1994
Resolves #2212

### DIFF
--- a/src/CommAggregation.chpl
+++ b/src/CommAggregation.chpl
@@ -403,10 +403,58 @@ module CommAggregation {
     use CommAggregation;
     use BigInteger, GMP;
 
+    proc bigint._serializedSize() {
+      extern proc chpl_gmp_mpz_struct_sign_size(from: __mpz_struct) : mp_size_t;
+
+      var sign_size = chpl_gmp_mpz_struct_sign_size(this.getImpl());
+
+      var size_bytes = c_sizeof(mp_size_t):int;
+      var limb_bytes = AutoMath.abs(sign_size:int) * c_sizeof(mp_limb_t):int;
+
+      return size_bytes + limb_bytes;
+    }
+
+    proc bigint._serializeInto(x: c_ptr(uint(8))) {
+      extern proc chpl_gmp_mpz_struct_sign_size(from: __mpz_struct) : mp_size_t;
+      extern proc chpl_gmp_mpz_struct_limbs(from: __mpz_struct) : c_ptr(mp_limb_t);
+
+      var sign_size = chpl_gmp_mpz_struct_sign_size(this.getImpl());
+
+      var size_bytes = c_sizeof(mp_size_t):int;
+      var limb_bytes = AutoMath.abs(sign_size:int) * c_sizeof(mp_limb_t):int;
+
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
I hardly feel qualified to comment of the correctness of this code, but nothing pops out to me as wrong

I'm excited for the performance increase! Thanks @ronawho!!

Ethan-DeBandi99 (approved):
Looks good. Thanks for knocking this out!

joshmarshall1 (approved):
Looks good

jaketrookman (approved):

PR #2294, opened 2023-03-24
Closes #2292 - Write Log from Client

### LINKED ISSUE #2292: Write Log from Client
It has been requested that the ability to send a log message to the server and have it written to the log be implemented.

The client side function should allow for the Log Message and Log Type to be specified.

Ethan-DeBandi99:
Closes #2292

Adds the client side `write_log` function. Updates also expose the `LogLevel` enum to allow setting the log level using `ak.LogLevel.<desired_lvl>`.

Adds the `LogMsg` module. 

This workflow allows the user to supply a custom message to be written to the logs and a log level `(DEBUG, INFO, WARN, ERROR, CRITICAL)`. An example of usage is below.

```python
import arkouda as ak
ak.connect()

# Note that the below is an example if setting log_lvl to something that is not the default.
# The default value for log_lvl is ak.LogLevel.INFO
ak.write_log("Example Log Message.", tag="MyCustomLog", log_lvl=ak.LogLevel.INFO)

ak.disconnect()
```

The above snippet of code will result in the output below being added to the log. Please note - the log output location will vary depending upon the log channel setting.

```
2023-03-24:12:28:22 [ClientGeneratedLog] MyCustomLog INFO [Chapel] Example Log Message.
```

This updates the Logging module to not print the `Line #` when client supplied logs are provided. All Logs generated using this method will show the module as `ClientGeneratedLog`. The log can also be tagged in place of the function name normally found in logs.

### DIFF
--- a/arkouda/logger.py
+++ b/arkouda/logger.py
@@ -340,10 +340,26 @@ def disableVerbose(logLevel: LogLevel = LogLevel.INFO) -> None:
 
 
 @typechecked
-def write_log(log_msg: str, log_lvl: LogLevel = LogLevel.INFO):
+def write_log(log_msg: str, tag: str= "ClientGeneratedLog", log_lvl: LogLevel = LogLevel.INFO):
     """
     Allows the user to write custom logs.
+
+    Parameters
+    -----------
+    log_msg: str
+        The message to be added to the server log
+    tag: str
+        The tag to use in the log. This takes the place of the server function name.
+        Allows for easy identification of custom logs.
+        Defaults to "ClientGeneratedLog"
+    log_lvl: LogLevel
+        The type of log to be written
+        Defaults to LogLevel.INFO
+
+    See Also
+    ---------
+    LogLevel
     """
     from arkouda.client import generic_msg
 
-    generic_msg(cmd="clientlog", args={"log_msg": log_msg, "log_lvl": log_lvl.name})
+    generic_msg(cmd="clientlog", args={"log_msg": log_msg, "log_lvl": log_lvl.name, "tag": tag})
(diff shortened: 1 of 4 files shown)


arkouda/logger.py
```diff
@@ -340,3 +337,13 @@ def disableVerbose(logLevel: LogLevel = LogLevel.INFO) -> None:
     """
     for logger in loggers.values():
         logger.disableVerbose(logLevel)
+
+
+@typechecked
+def write_log(log_msg: str, log_lvl: LogLevel = LogLevel.INFO):
+    """
+    Allows the user to write custom logs.
+    """
```
pierce314159, line 346: could we add some of the usual docstrings stuff (`Parameters`, `Returns`, `Examples`)?

pierce314159 (commented):
This all looks good! My only suggestion is maybe flesh out the docstring a bit more

jaketrookman (approved):
Looks good

src/LogMsg.chpl
```diff
@@ -18,22 +18,25 @@ module LogMsg
...
             }
             when Logging.LogLevel.INFO {
-                clLogger.info(getModuleName(),getRoutineName(),getLineNumber(),logMsg);
+                clLogger.info(mod_replace,tag,0,logMsg);
             }
             when Logging.LogLevel.WARN {
-                clLogger.warn(getModuleName(),getRoutineName(),getLineNumber(),logMsg);
+                clLogger.warn(mod_replace,tag,getLineNumber(),logMsg);
             }
             when Logging.LogLevel.ERROR {
```
pierce314159: It looks like `.INFO` is the only case that got the line number = 0 update

pierce314159 (approved):

PR #2302, opened 2023-03-24
Closes #2023 - Added all compression types to legacy benchmarks

### LINKED ISSUE #2023: IO Benchmark Updates for Compression
Arkouda's Parquet writing has been updated to allow for all compression types supported by Parquet (list below). The Benchmarks for IO need to be updated to benchmark all compression types, not just Snappy compression. The types to be supported are listed below.

- Snappy
- GZip
- Brotli
- ZSTD
- LZ4

joshmarshall1:
This PR closes #2023 

Adds `"none", "snappy", "gzip", "brotli", "zstd", and "lz4"` as supported compression types in legacy IO benchmarks (Only supported in Parquet writes).

Also added dtype support for `float64`, and `str` where it wasn't already supported. This is in order to match more closely to the new IO benchmark's tests for proper comparison


benchmarks/IO.py
```diff
@@ -164,9 +177,10 @@ def create_parser():
     parser.add_argument(
         "-c",
         "--compressed",
-        default=False,
-        action="store_true",
-        help="Write with Snappy compression and RLE encoding",
+        default="none",
```
Ethan-DeBandi99, line 180: Marking this here since the line above did not change. Since we are now allowing the compression types to be set instead of just turning on and off, I think it would be better for the argument to be `--compression`. We could stick with compressed and have that just trigger all commpression types (including no compression) to run. I think that would simplify things a bit. However, I do like the ability to turn things on and off as we please.

I am wondering though if we want to default to no compression or running all. I think we are at a point where we would like to see the comparsion of the compression more than we don't. Can we configure this to default to `""` and have that trigger all the compression types?

benchmarks/IO.py
```diff
@@ -180,6 +194,8 @@ def create_parser():
         raise ValueError("Dtype must be {}, not {}".format("/".join(TYPES), args.dtype))
     ak.verbose = False
     ak.connect(args.hostname, args.port)
+    comp_str = args.compressed
+    comp_types = None if comp_str == "" else comp_str.lower().split(",")
```
Ethan-DeBandi99, line 198: Going off my previous comment, I think we want this to default to all compression types now.

benchmarks/IO.py
```diff
@@ -11,14 +11,19 @@
...
 
-def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, compressed=False):
+def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, comps=None):
+    if comps is None:
+        comps = ["none"]
+    supported_comps = [None, "none", "snappy", "gzip", "brotli", "zstd", "lz4"]
+
     if not parquet:
-        print(">>> arkouda {} HDF5 write with compressed={}".format(dtype, compressed))
+        print(">>> arkouda {} HDF5 write with compressed={}".format(dtype, comps))
```
Ethan-DeBandi99, line 24: Change `compressed` to `compression`

benchmarks/IO.py
```diff
@@ -28,16 +33,20 @@ def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, co
...
     for i in range(trials):
-        for j in range(numfiles):
-            start = time.time()
-            a.to_hdf(f"{path}{j:04}") if not parquet else a.to_parquet(
-                f"{path}{j:04}", compression="snappy" if compressed else None
-            )
-            end = time.time()
-            writetimes.append(end - start)
+        for comp in comps:
+            if comp in supported_comps:
```
Ethan-DeBandi99, line 42: Need this to check (assuming the change to global for the list)
```python
if comp.lower() in COMPRESSION
```
Otherwise, capitalization will impact this check.

benchmarks/parquetIO.py
Ethan-DeBandi99, line 1: We should make the same changes here as in IO.py

benchmarks/parquetMultiIO.py
Ethan-DeBandi99, line 1: Again, same changes required here as in `IO.py`

Ethan-DeBandi99 (changes requested):

benchmarks/IO.py
joshmarshall1: Updated to be `--compression` and default to `""`

benchmarks/IO.py
joshmarshall1: Changed this to default to `COMPRESSIONS`

benchmarks/IO.py
```diff
@@ -11,14 +11,19 @@
...
     "uint64",
+    "str"
 )
 
 
-def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, compressed=False):
+def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, comps=None):
+    if comps is None:
+        comps = ["none"]
+    supported_comps = [None, "none", "snappy", "gzip", "brotli", "zstd", "lz4"]
```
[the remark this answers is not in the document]
joshmarshall1: Done

benchmarks/IO.py
joshmarshall1: Done

benchmarks/parquetIO.py
joshmarshall1: Done

benchmarks/parquetMultiIO.py
joshmarshall1: Done

benchmarks/IO.py
joshmarshall1: Okay, I see now. Fixed

benchmarks/IO.py
```diff
@@ -11,14 +11,19 @@
...
+
     if not parquet:
-        print(">>> arkouda {} HDF5 write with compressed={}".format(dtype, compressed))
+        print(">>> arkouda {} HDF5 write with compressed={}".format(dtype, comps))
     else:
-        print(">>> arkouda {} Parquet write with compressed={}".format(dtype, compressed))
+        print(">>> arkouda {} Parquet write with compressed={}".format(dtype, comps))
```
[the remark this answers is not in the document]
joshmarshall1: Fixed

Ethan-DeBandi99 (approved):

benchmarks/IO.py
```diff
@@ -28,16 +40,20 @@ def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, co
...
-            writetimes.append(end - start)
+        for comp in comps:
+            if comp in COMPRESSIONS:
+                for j in range(numfiles):
+                    start = time.time()
+                    a.to_hdf(f"{path}{j:04}") if not parquet else a.to_parquet(
+                        f"{path}{comp}{j:04}", compression=None if comp == "none" else comp
+                    )
+                    end = time.time()
+                    writetimes.append(end - start)
```
pierce314159, line 56: It seems like we're averaging the times of all the compressions together? Would it make more sense to have separate timings for each compression type?

pierce314159 (changes requested):
I might be misunderstanding but wanted to double check

benchmarks/IO.py
joshmarshall1: Added averaging for each compression type

pierce314159 (approved):

Ethan-DeBandi99:
Just a thought after giving a final once over on this. Do we need to have all the compression types recorded for HDF5? HDF5 is not using the compressions, only Parquet and I believe (especially with this being the old benchmark system), this will bloat the results significantly. Additionally, it will add a bunch of "different" cases for the exact same thing and not add any value.

joshmarshall1:
> Just a thought after giving a final once over on this. Do we need to have all the compression types recorded for HDF5? HDF5 is not using the compressions, only Parquet and I believe (especially with this being the old benchmark system), this will bloat the results significantly. Additionally, it will add a bunch of "different" cases for the exact same thing and not add any value.
> 
> Re-requested review just so we can make sure to address this.

This makes sense. I'm leaning toward moving hdf5 out of the compressions loop so it's only ran once per trial

benchmarks/IO.py
```diff
@@ -28,22 +40,37 @@ def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, co
...
+        times["HDF5"] = sum(writetimes) / trials
 
     nb = a.size * a.itemsize * numfiles
-    print("write Average rate = {:.2f} GiB/sec".format(nb / 2**30 / avgwrite))
+    for key in times.keys():
+        print("write Average time {} = {:.4f} sec".format(key, times[key]))
+        print("write Average rate {} = {:.2f} GiB/sec".format(key, nb / 2**30 / times[key]))
 
 
 def time_ak_read(N_per_locale, numfiles, trials, dtype, path, seed, parquet):
```
pierce314159, line 76: Question for the reads, we don't do any iteration on the compression here. So I'm not familiar enough with this workflow to say definitively (maybe @Ethan-DeBandi99 can chime in), but it seems like it would read all the different file compressions and average them together maybe?

Before when doing the full read write workflow, you were either h5, pq with compression, or pq without compression. So the read was always measuring exactly what the write was doing. Now it feels like we could have a couple different write times broken out by compression, but only one read time which could vary quite a lot depending which compressions the files were written with

I'm not sure if this is worth changing but I def wanted to capture the discussion around it

joshmarshall1: This is now implemented

pierce314159 (approved):

benchmarks/IO.py
```diff
@@ -11,14 +11,26 @@
...
+
 
+def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, comps=None):
+    if comps is None or comps == [""]:
+        comps = COMPRESSIONS
 
-def time_ak_write(N_per_locale, numfiles, trials, dtype, path, seed, parquet, compressed=False):
     if not parquet:
-        print(">>> arkouda {} HDF5 write with compressed={}".format(dtype, compressed))
+        print(">>> arkouda {} HDF5 write with compression={}".format(dtype, comps))
```
bmcdonald3, line 31: Do we want to be printing out the compressions with HDF5 since it isn't supported?

PR #2308, opened 2023-03-28
Closes #2240: `ak.coargsort` empty String and Categorical bug fix

### LINKED ISSUE #2240: `ak.coargsort` unexpected behavior with empty Strings arrays
When `ak.coargsort` is run against a `Strings` array that is empty, it attempts to return an array of zeros. This is an issue because it passes the `dtype` of the strings array through, but `ak.zeros` expects a numeric dtype.

The chances of this actually occurring are slim (I found it running a benchmark with a problem size that was too small). 

This should be a fairly simple fix. Simply add a return case for `size==0 and dtype=="str"`, that does not error.

jaketrookman:
This PR (Closes #2240):

- Fixes error for empty arrays of String type and Categorical and handles them correctly.
- Adds a test to ensure that they properly produce an empty array.

### DIFF
--- a/arkouda/sorting.py
+++ b/arkouda/sorting.py
@@ -158,8 +158,9 @@ def coargsort(
             size = a.size
         elif size != a.size:
             raise ValueError("All pdarrays, Strings, or Categoricals must be of the same size")
+
     if size == 0:
-        return zeros(0, dtype=arrays[0].dtype)
+        return zeros(0, dtype=int if a.dtype=='str' else arrays[0].dtype)
 
     repMsg = generic_msg(
         cmd="coargsort",
--- a/arkouda/sorting.py
+++ b/arkouda/sorting.py
@@ -160,7 +160,7 @@ def coargsort(
             raise ValueError("All pdarrays, Strings, or Categoricals must be of the same size")
 
     if size == 0:
-        return zeros(0, dtype=int if a.dtype=='str' else arrays[0].dtype)
+        return zeros(0, dtype=int if a.dtype == 'str' else arrays[0].dtype)
 
     repMsg = generic_msg(
         cmd="coargsort",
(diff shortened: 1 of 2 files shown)


arkouda/sorting.py
```diff
@@ -158,8 +158,9 @@ def coargsort(
             size = a.size
         elif size != a.size:
             raise ValueError("All pdarrays, Strings, or Categoricals must be of the same size")
+
     if size == 0:
-        return zeros(0, dtype=arrays[0].dtype)
+        return zeros(0, dtype=int if a.dtype == 'str' else arrays[0].dtype)
```
pierce314159, line 163: We're referencing `a` which is defined in the `for` loop. That feels risky to me since it could be garbage collected. We should try and use `arrays[0]` instead

The other things is doing `== 'str'`, doing direct comparison against a hard coded value isn't best practices. I would recommend using `isinstance` instead for type comparisons

Overall I think changing this to something like
```python
return zeros(0, dtype=int if isinstance(arrays[0], (Strings, Categorical)) else arrays[0].dtype)
```
would be better

tests/coargsort_test.py
```diff
@@ -191,6 +191,14 @@ def test_coargsort_categorical(self):
             mixed_perm = ak.coargsort([cat, string, cat_from_codes], algo)
             self.assertListEqual(["a", "a", "b", "b", "c"], cat_from_codes[mixed_perm].to_list())
 
+    def test_coargsort_empty(self):
+        empty_str = ak.random_strings_uniform(1, 16, 0)
+        empty_cat = ak.Categorical(empty_str)
+        empty_pda = ak.array([], int)
+        self.assertEqual(0, len(ak.coargsort([empty_pda])))
```
pierce314159, line 198: We don't really need to test `empty_pda` that was included just to demonstrate the desired behavior

pierce314159 (approved):

joshmarshall1 (approved):

Ethan-DeBandi99 (approved):

PR #2310, opened 2023-03-28
Closes #2297 - Fixes HDF5 Single File Write Stall

### LINKED ISSUE #2297: HDF5 Single File Stalls
It was reported that when using the HDF5 single file capability on a large dataset appears to stall out. What appears to happen is the header will be written, it appears idle for ~1 hr and then will write a chunk of the file, sit idle and repeat. 

I have a feeling that this is due to how the file write is occurring. The data is localized and written into the file in large chunks. We will need to look into this and determine if there is a way to provide progress updates if this is the case or speed up the writes so that it does not appear idle for long periods.

Ethan-DeBandi99:
Closes #2297 

This PR updates the workflow for writing a single file to go to the locale where each chunk of data is and copy it to the root node using a `dstAggregator` instead of `lowLevelLocalizingSlice`. This results in significantly faster runtimes for the localization and better comms for numerics and strings. Based on the updates @ronawho suggested, I do not think we are going to squeeze any more performance out of this, but performance is ~25x better for numerics and ~96x better for strings at a minimum.

Before and after metrics posted below. Numerics using a pdarray of size `10**5` and strings using `ak.random_strings_uniform(1, 16, 10**5)`

Also noting that for review to ensure this performance make sure `-suseBulkTransfer=true` is set. 

**Numeric Localization Before PR**
```
Localize Time = 28

| locale |    get | put | execute_on |
| -----: | -----: | --: | ---------: |
|      0 | 320098 |  40 |        112 |
|      1 |      0 |  12 |          0 |
|      2 |      0 |  12 |          0 |
|      3 |      0 |  12 |          0 |
|      4 |      0 |  12 |          0 |
```

**Numeric Localization After PR**
```
Localize Time = 0

| locale | get | put | execute_on_nb |
| -----: | --: | --: | ------------: |
|      0 |   0 |   0 |             4 |
|      1 |   3 |   1 |             0 |
|      2 |   3 |   1 |             0 |
|      3 |   3 |   1 |             0 |
|      4 |   3 |   1 |             0 |

```

**Strings Localization Before PR**
```
Localize Values Time = 96

| locale |    get |
| -----: | -----: |
|      0 | 719086 |
|      1 |      0 |
|      2 |      0 |
|      3 |      0 |
|      4 |      0 |
```

**Strings Localization After PR**
```
Localize Time = 0

| locale | get | put | execute_on_nb |
| -----: | --: | --: | ------------: |
|      0 |   0 |   0 |             8 |
|      1 |   6 |   2 |             0 |
|      2 |   6 |   2 |             0 |
|      3 |   6 |   2 |             0 |
|      4 |   6 |   2 |             0 |
```

### DIFF
--- a/src/HDF5Msg.chpl
+++ b/src/HDF5Msg.chpl
@@ -384,6 +384,20 @@ module HDF5Msg {
     }
 
     /*
+        Takes distributed data and localizes it to the root.
+        This is used when writing data to a single file.
+    */
+    proc localizeData(localFlat: [?D] ?t, A) throws {
+        coforall loc in A.targetLocales() with (ref A) do on loc {
+            const locDom = A.localSubdomain();
+            forall (localVal, valIdx) in zip(localFlat[locDom.low..locDom.high], locDom.low..locDom.high) with (var agg = newDstAggregator(t), ref A) {
+                // Copy the value local to the current locale to the array on the root.
+                agg.copy(localVal, A[valIdx]);
+            }
+        }
+    }
+
+     /*
         writes 1D array to dataset in single file
     */
     proc writeLocalDset(file_id: C_HDF5.hid_t, dset_name: string, A, dimension: int, type t) throws{
@@ -394,7 +408,7 @@ module HDF5Msg {
 
         // always store multidimensional arrays as flattened array
         var dims = dimension:C_HDF5.hsize_t;
-        C_HDF5.H5LTmake_dataset(file_id, dset_name.c_str(), 1:c_int, dims, dtype_id, A.ptr);
+        C_HDF5.H5LTmake_dataset(file_id, dset_name.c_str(), 1:c_int, dims, dtype_id, A);
     }
 
     /*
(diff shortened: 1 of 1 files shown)


src/HDF5Msg.chpl
```diff
@@ -489,26 +503,38 @@ module HDF5Msg {
                 select entryDtype {
                     when DType.Int64 {
                         var flat = toSymEntry(toGenSymEntry(entry), int);
-                        var localFlat = new lowLevelLocalizingSlice(flat.a, 0..#flat.size);
-                        writeLocalDset(file_id, dset_name, localFlat, dims, int);
+                        var localFlat: [0..#flat.size] int;
+                        ref olda = flat.a;
+                        localizeData(localFlat, olda);
```
ronawho, line 508: I'd recommend trying to use chapel's bulk communication by just doing array assignment. This would be:

```chpl
var localFlat: [0..#flat.size] int = flat.a;
```

(note that the typing is needed to avoid creating an array with the same distribution as flat)

---

Background if it's helpful:
 - `lowLevelLocalizingSlice` was designed to optimize cases where most of your slices are local slices (so you can avoid allocation) and some slices live entirely on one remote node so you can optimize and use a single communication for the slice instead of a comm per element. It was not designed to optimize this case and just falls back to grabbing an element at a time so it makes sense to me that it performed poorly. You could do something like initiate one low-level slice per locale like we do for `in1dAr2PerLocAssoc`, but I think that's overkill here and was only done in that code because we were already parallel across nodes and I didn't want to create new tasks. 
 - The "go remote and have each node aggregate elements to locale 0" in this PR should be more efficient than GETing each element, but it will create pretty high load on locale 0, and there's a GET for the local array mettadata each time. I think it'd be better to have locale 0 use a `SrcAggregator`
 - The most optimal since we're communicating contiguous indices (normally aggregators are used for random indices) would just be to have a single GET from each locale or do the coforall on and have each node do a single PUT. You could do this manually, but I'm pretty confident the array assignment will do this under the covers. Our bulk assignment has some overhead for exchanging meta-data so it won't truly be a single GET/PUT, but it should be a near trivial amount.

ronawho: Discussed offline, but forgot when I suggested this -- quick compile disable bulk transfer with `-suseBulkTransfer=false` so that needs to be unset for this to perform well. I opened https://github.com/Bears-R-Us/arkouda/issues/2311 to revisit whether it's still worthwhile to disable bulk transfer.

ronawho (approved):
Code looks good to me, but probably want to update comm counts in PR description

pierce314159 (approved):
Looks good! And nice perf boost 🚀

jaketrookman (approved):
Looks good

PR #2314, opened 2023-03-28
Closes #2319 - PyTest Test Suite Restructure

### LINKED ISSUE #2319: Test Suite PyTest configuration & First test update
We need to create a configuration file in line with `benchmark_v2/conftest.py` to be used by version 2 of our test suite that will later be merged with the benchmark file. Should also include the first test update as an example of how the original and the new formats differ, `alignment_test.py`

joshmarshall1:
This PR closes #2319 

- Adds `pytest` folder similar to how we did `benchmark_v2` to separate new and old files without loosing current functionality (`tests_v2` causes issues with the old test suite since pytest looks for `test*` directories)
- Adds `pytest/conftest.py` that mirrors the implementation of `benchmark_v2/conftest.py` for easy merging of the two files to root level when ready
- Adds `exampleTest.py` that shows an implementation of `argsort.py` benchmark's correctness check using PyTest `prob_size` and `seed` variables (This is a temporary file that will be removed once more tests are upgraded)
- Adds a v2 version of `alignment_test.py` showing a basic implementation of the PyTest structure

This can be easily added to the makefile in place of (or alongside) `make test`

Currently ran using the following command with optional args `--size` and `--seed`:
```
python3 -m pytest -c test_v2.ini --size="10**2" --seed="255"
```


test.ini
Ethan-DeBandi99, line 1: Could we rename this `test_v2.ini`? I just want to make things as obvious as possible.

pytest/alignment_test.py
```diff
@@ -0,0 +1,182 @@
...
+    lb = [0, 10, 20, 30, 40, 50]
+    ub = [9, 19, 29, 39, 49, 59]
+    v = [22, 51, 44, 1, 38, 19, 40, 60, 100]
+
+    # test int64
+    lower_bound = ak.array(lb)
+    upper_bound = ak.array(ub)
+    vals = ak.array(v)
+    interval_idxs = ak.search_intervals(vals, (lower_bound, upper_bound))
+    assert expected_result == interval_idxs.to_list()
```
Ethan-DeBandi99, line 17: General question here - Do we lose the ability to run things like `assertListEqual` with this configuration of pytest? Recently, that has become more and more useful for identifying where problems are. If we are still able to use it, we should update this code to use things like that.

pytest/conftest.py
Ethan-DeBandi99, line 1: Just noting, because based on the PR description I am worried there may be some confusion. Ultimately, we want to have one `conftest.py` file. This will either need to go into the root of `arkouda` or `arkouda/tests` (which `pytest` will be renamed to, when complete.

Ethan-DeBandi99 (commented):
There are a few things concerning me with this. My concerns are not necessarily with the implementation itself, but with some potential limitations I am seeing. Depending on how we are able to handle some of these, I would have some reservations about transitioning. We should definitely have more discussion on this.

pytest/conftest.py
joshmarshall1: Maybe I wrote it strangely, but that's exactly what I meant by "merging of the two files to root level when ready"

pytest/alignment_test.py
joshmarshall1: This is something I've already been looking into. PyTest's failed assertions on lists do provide `diff` output of each list highlighting where problems are, but I have been looking into using custom asserts to handle more complex cases

joshmarshall1: As an example output I purposefully caused a test to fail to show the failure ouput
```
        vals = (ak.array([23, 13, 3]), ak.array([23, 13, 3]))
        ans = [2, 0, 1]
>       assert ans == ak.search_intervals(vals, (starts, ends), hierarchical=False).to_list()
E       assert [2, 0, 1] == [2, 1, 0]
E         At index 1 diff: 0 != 1
E         Use -v to get more diff
```
Adding `-v`
```
>       assert ans == ak.search_intervals(vals, (starts, ends), hierarchical=False).to_list()
E       assert [2, 0, 1] == [2, 1, 0]
E         At index 1 diff: 0 != 1
E         Full diff:
E         - [2, 1, 0]
E         + [2, 0, 1]
```

Ethan-DeBandi99: Sounds good. I just wanted to be sure that we weren't losing any information for debugging.

pytest/alignment_test.py
```diff
@@ -0,0 +1,182 @@
...
+        -1,
+        0,
+        0,
+        1,
+        -1,
+        1,
+        -1,
+    ]
+    search_intervals_hierarchical = ak.search_intervals(vals, (starts, ends)).to_list()
+    assert search_intervals_hierarchical == [0, 0, 0, 0, 1, 1, 1, -1]
```
pierce314159, line 69: if we could do something like `search_intervals_hierarchical` for the `non-hierarchical` case to avoid the black getting so wild with the formatting that would sweet

```python
search_int_non_hierarch = ak.search_intervals(vals, (starts, ends), hierarchical=False).to_list()
assert search_int_non_hierarch == [0, -1, 0, 0, 1, -1, 1, -1]
search_intervals_hierarchical = ak.search_intervals(vals, (starts, ends)).to_list()
assert search_intervals_hierarchical == [0, 0, 0, 0, 1, 1, 1, -1]
```

pytest/alignment_test.py
```diff
@@ -0,0 +1,182 @@
...
+    assert expected_result == interval_idxs.to_list()
+
+    # test uint64
+    lower_bound = ak.array(lb, dtype=ak.uint64)
+    upper_bound = ak.array(ub, dtype=ak.uint64)
+    vals = ak.array(v, dtype=ak.uint64)
+    interval_idxs = ak.search_intervals(vals, (lower_bound, upper_bound))
+    assert expected_result == interval_idxs.to_list()
+
+    # test uint64
```
pierce314159, line 98: this comment should say `float64`. It's also wrong in the existing tests, so I'm not sure if it's worth updating in this PR, but i wanted to note it

pytest/exampleTest.py
```diff
@@ -0,0 +1,21 @@
...
+@pytest.mark.parametrize("dtype", TYPES)
+def test_argsort_correctness(dtype):
+    N = pytest.prob_size
+    seed = pytest.seed
+    if dtype == "int64":
+        a = ak.randint(0, 2**32, N, seed=seed)
+    elif dtype == "uint64":
+        a = ak.randint(0, 2**32, N, dtype=ak.uint64, seed=seed)
+    elif dtype == "float64":
+        a = ak.randint(0, 1, N, dtype=ak.float64, seed=seed)
```
pierce314159, line 17: we should add the `"str"` case

pytest/alignment_test.py
```diff
@@ -0,0 +1,182 @@
+import pytest
+
+import arkouda as ak
+
+
+def test_search_interval():
```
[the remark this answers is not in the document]
Ethan-DeBandi99: @pierce314159 - yes it will be. You just will not need the class any longer for the entire file and use `-k "test_name"` for individual

pytest/alignment_test.py
joshmarshall1: Added your suggested edit

pytest/alignment_test.py
joshmarshall1: Fixed

pytest/exampleTest.py
joshmarshall1: Removed the example file since alignment_test.py can serve as the example now

pierce314159 (approved):

PR #2323, opened 2023-03-30
Closes #2050 - File of Origin on Read

### LINKED ISSUE #2050: File of Origin on Read
When reading data from HDF5/Parquet files that are distributed, add a flag to return a column indicating the `File of Origin` as a column in the dataset. When returned to the client, this should be a `Categorical` object.

The intention is to allow users to easily determine where the data came from.

Ethan-DeBandi99:
Closes #2050

This PR adds the ability to link the data returned during a file read to the file it was read from. This is done by first creating a Categorical and then returning a dataset that will always be named `Filename_Codes`. These codes can then be used to access the filename from the categorical object. 

Adds the function `ak.read_tagged_data`. This is functionally equivalent to `ak.read`, but triggers the addition of tagging with filename codes. This function also returns a tuple containing the data returned, where `Filename_Codes` will be included as a dataset and a `Categorical` object created from ALL filenames accessed.

As an example, running with 16 locales:
```python
import arkouda as ak

ak.connect()

fname = "path_to_file"

a = ak.SegArray.from_parts(ak.arange(0, 10, 2), ak.arange(10))
a.to_parquet(fname)
data, cat = ak.read_tagged_data(fname+"_*")
print(data)

{'Filname_Code': array([0 3 6 9 12]), 'segarray': SegArray([
[0 1]
[2 3]
[4 5]
[6 7]
[8 9]
])}

import pandas as pd
fcodes = data["Filname_Code"].to_list()
for fc in fcodes:
    fname = cat[fc]
    # verification that data is being properly linked to the file
    df = pd.read_parquet(fname, engine='pyarrow')
    print(f"File: {fname}\n{df}\n\n")

ak.disconnect()
```
**Output from the loop Print**
```python
File: /Users/ethandebandi/Documents/save_load_testing/tagged_hdf_LOCALE0000
  segarray
0   [0, 1]


File: /Users/ethandebandi/Documents/save_load_testing/tagged_hdf_LOCALE0003
  segarray
0   [2, 3]


File: /Users/ethandebandi/Documents/save_load_testing/tagged_hdf_LOCALE0006
  segarray
0   [4, 5]


File: /Users/ethandebandi/Documents/save_load_testing/tagged_hdf_LOCALE0009
  segarray
0   [6, 7]


File: /Users/ethandebandi/Documents/save_load_testing/tagged_hdf_LOCALE0012
  segarray
0   [8, 9]
```

This PR does add a message for handling of the glob expression. This is because we cannot guarantee that the client will have access to the file system containing data. This allows for access of the full file list in the case of a glob expression being provided and then allows creation of the `Categorical`.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -11,6 +11,7 @@ import arkouda.array_view
 from arkouda.categorical import Categorical
 from arkouda.client import generic_msg
 from arkouda.pdarrayclass import create_pdarray, pdarray
+from arkouda.pdarraycreation import array
 from arkouda.segarray import SegArray
 from arkouda.strings import Strings
 
@@ -25,6 +26,7 @@ __all__ = [
     "read_parquet",
     "read_csv",
     "read",
+    "read_tagged_data",
     "import_data",
     "export",
     "to_hdf",
@@ -512,6 +514,7 @@ def read_hdf(
     strict_types: bool = True,
     allow_errors: bool = False,
     calc_string_offsets: bool = False,
+    tag_data = False
 ) -> Union[
     pdarray,
     Strings,
@@ -545,6 +548,9 @@ def read_hdf(
         Default False, if True this will tell the server to calculate the
         offsets/segments array on the server versus loading them from HDF5 files.
         In the future this option may be set to True as the default.
+    tagData: bool
+        Default False, if True tag the data with the code associated with the filename
+        that the data was pulled from.
 
     Returns
     -------
@@ -580,6 +586,10 @@ def read_hdf(
     and read all of them. Use ``get_datasets`` to show the names of datasets
     to HDF5 files.
 
+    See Also
+    ---------
+    read_tagged_data
+
     Examples
     --------
     >>>
@@ -593,6 +603,8 @@ def read_hdf(
     datasets = _prep_datasets(filenames, datasets, allow_errors)
 
     if iterative:
+        if tag_data:
+            raise RuntimeError("Cannot tag data with iterative read.")
         return {
             dset: read_hdf(
                 filenames,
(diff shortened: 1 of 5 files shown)


joshmarshall1 (approved):

arkouda/io.py
```diff
@@ -1617,3 +1643,94 @@ def read(
...
+    j_str = generic_msg(
+        cmd="globExpansion",
+        args={"file_count": len(filenames), "filenames": filenames},
+    )
+    file_list = json.loads(j_str)
+    file_cat = Categorical(
+        array(file_list)
+    )  # create a categorical from the ak.Strings representation of the file list
+
+    # TODO - Add parameter sent to server that triggers the filename mapping be added
```
pierce314159, line 1706: Are these still TODOs?

arkouda/io.py
```diff
@@ -1617,3 +1643,94 @@ def read(
...
+    File Type is determined automatically.
+
+    Parameters
+    ----------
+    filenames : list or str
+        Either a list of filenames or shell expression
+    datasets : list or str or None
+        (List of) name(s) of dataset(s) to read (default: all available)
+    iterative : bool
+        Iterative (True) or Single (False) function call(s) to server
```
pierce314159, line 1667: Should we say `"Cannot tag data with iterative read."` and also add a `RuntimeError` like in `read_hdf5/parquet`? Or maybe it would make more sense not to let them pass in a bool for `iterative` for the `read_tagged_data` function?

pierce314159 (approved):
Just a few small things client side. Logic looks good and i verified locally this works how I would expect with 16 locales and segarray size=100 with parquet and hdf

arkouda/io.py
Ethan-DeBandi99: Agreed. I removed the iterative param from the `read_tagged_data` function and hard-coded the calls to false."

pierce314159 (approved):

jaketrookman (approved):

PR #2325, opened 2023-03-30
Closes #2295 - SegArray Filters

### LINKED ISSUE #2295: SegArray Filters
Ability to filter out values from a `SegArray`. The segments will need to be maintained. Thus, when values are removed the offsets will need to be adjusted appropriately. 

We will need to implement this with a `remove_empty` flag. This will default to `False`, but when set to `True` will remove any empty segments in the result of the filter.

Ethan-DeBandi99:
Closes #2295

Adds `ak.SegArray.filter`. This function allows for a `SegArray` object to be filtered by a single value, list/tuple of values, or a pdarray. The returned SegArray object contains only the  values not found in the filter. The user has the option to remove any empty segments in the result by setting `discard_empty=True` when calling the function.

Testing for `ak.SegArray.filter` added to `tests/segarray_tests.py`.

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -1411,6 +1411,43 @@ class SegArray:
             segments[truth] = segments[arange(self.size)[truth] + 1]
             return SegArray.from_parts(segments, new_values[g.permutation])
 
+    def filter(self, filter, discard_empty: bool = False):
+        """
+        Filter values out of the SegArray object
+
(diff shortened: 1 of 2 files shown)


pierce314159 (approved):
Looks good! I think this implementation is super slick

joshmarshall1 (approved):

PR #2333, opened 2023-04-03
Closes #2327 - Fixes Issue Loading 16-bit and 32-bit from Parquet

### LINKED ISSUE #2327: Issue loading Parquet files with integer columns narrower than 64 bits
**Describe the bug**
When loading Parquet files that have integer columns narrower than 64 bits, Arkouda corrupts uint32 values whose high bit is set and ignores 16-bit columns entirely. (submitting for colleague)

**To Reproduce**
```python
df_pd = pd.DataFrame({
    "int16":  pd.Series([2**15-1, -2**15], dtype=np.int16),
    "int32":  pd.Series([2**31-1, -2**31], dtype=np.int32),
    "uint16: pd.Series(['2**15-1,  2**15], dtype=np.uint16),
    "uint32: pd.Series(['2**31-1,  2**31], dtype=np.uint32),
})
df_pd.to_arquet("test.parquet")
df_ak = ak.DataFrame(ak.read("test.parquet"))
```

**Expected behavior**
df_ak values to match df_pd values

**Error Message**
No error message

**Is this a Blocking Issue**
Works fine reading hdf5 with smaller column sizes.

Ethan-DeBandi99:
Closes #2327

This PR fixes issues with reading `int16`, `uint16`, `int32` and `uint32` columns from Parquet files. `int16` and `uint16` are logical types in Parquet and are stored as `int32` or `uint32`. There was an issue with the handling of the high bit when converting from `uint32` to `uint64`. To resolve this, we shift the high bit to the correct position and apply a mask to remove the top 32 bits allowing `uint32` to be properly stored in a `uint64`.  `int16` and `uint16` were added as valid types to read from Parquet files. Because they are stored as 32-bit values, the same fix for 32-bit values apply to them.

Tested using the following python code:
```python
df_pd = pd.DataFrame({
"int16": pd.Series([2**15-1, -2**15], dtype=np.int16),
"int32": pd.Series([2**31-1, -2**31], dtype=np.int32),
"uint16": pd.Series([2**15-1, 2**15], dtype=np.uint16),
"uint32": pd.Series([2**31-1, 2**31], dtype=np.uint32),
})
print(df_pd)
df_pd.to_parquet(fname)
df_ak = ak.DataFrame(ak.read(fname+"*"))
print(df_ak.__repr__())

**Expected output from Pandas** 
```
 int16       int32  uint16      uint32
0  32767  2147483647   32767  2147483647
1 -32768 -2147483648   32768  2147483648
```

**Arkouda Read Output BEFORE this PR**
```
        int32                uint32
0  2147483647            2147483647
1 -2147483648  18446744071562067968 (2 rows x 2 columns)
```

**Arkouda Read Output AFTER this PR**
```
   int16       int32  uint16      uint32
0  32767  2147483647   32767  2147483647
1 -32768 -2147483648   32768  2147483648 (2 rows x 4 columns)
```

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -89,15 +89,16 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
       return ARROWERROR;
     }
     auto myType = sc -> field(idx) -> type();
+    std::cout << "\n\nCol: " << colname << " Type: " << myType->name() << "\n\n";
 
     if(myType->id() == arrow::Type::INT64)
       return ARROWINT64;
-    else if(myType->id() == arrow::Type::INT32)
-      return ARROWINT32;
+    else if(myType->id() == arrow::Type::INT32 || myType->id() == arrow::Type::INT16)
+      return ARROWINT32; // 16bit stored as 32bit so can treat both as 32bit
     else if(myType->id() == arrow::Type::UINT64)
       return ARROWUINT64;
-    else if(myType->id() == arrow::Type::UINT32)
-      return ARROWUINT32;
+    else if(myType->id() == arrow::Type::UINT32 || myType->id() == arrow::Type::UINT16)
+      return ARROWUINT32; // 16bit stored as 32bit so can treat both as 32bit
     else if(myType->id() == arrow::Type::TIMESTAMP)
       return ARROWTIMESTAMP;
     else if(myType->id() == arrow::Type::BOOL)
@@ -163,11 +164,11 @@ int cpp_getListType(const char* filename, const char* colname, char** errMsg) {
         auto f_type = field->type();
         if(f_type->id() == arrow::Type::INT64)
           return ARROWINT64;
-        else if(f_type->id() == arrow::Type::INT32)
+        else if(f_type->id() == arrow::Type::INT32 || f_type->id() == arrow::Type::INT16)
           return ARROWINT32;
         else if(f_type->id() == arrow::Type::UINT64)
           return ARROWUINT64;
-        else if(f_type->id() == arrow::Type::UINT32)
+        else if(f_type->id() == arrow::Type::UINT32 || f_type->id() == arrow::Type::UINT16)
           return ARROWUINT32;
         else if(f_type->id() == arrow::Type::TIMESTAMP)
           return ARROWTIMESTAMP;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/ArrowFunctions.cpp [hunk 1]
```diff
@@ -89,15 +89,17 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
       return ARROWERROR;
     }
     auto myType = sc -> field(idx) -> type();
+    std::cout << "\n\nCol: " << colname << " Type: " << myType->name() << "\n\n";
```
pierce314159, line 92: Is this left over from debugging?

src/ParquetMsg.chpl
```diff
@@ -830,6 +830,12 @@ module ParquetMsg {
         } else if ty == ArrowTypes.uint64 || ty == ArrowTypes.uint32 {
           var entryVal = new shared SymEntry(len, uint);
           readFilesByName(entryVal.a, filenames, sizes, dsetname, ty);
+          if (ty == ArrowTypes.uint32){ // correction for high bit 
+            ref ea = entryVal.a;
+            // Access the high bit (64th bit) and shift it into the high bit for uint32 (32nd bit)
+            // Apply 32 bit mask to drop top 64 bits and properly store uint32
```
pierce314159, line 836:
```suggestion
            // Apply 32 bit mask to drop top 32 bits and properly store uint32
```

pierce314159 (approved):
Looks good!

jaketrookman (approved):
Looks good to me

e-kayrakli:
Bit-wise logic added here is unnecessary and just masks over the actual issue.

- The template function [`readColumnIrregularBidwidth`](https://github.com/Bears-R-Us/arkouda/blob/master/src/parquet/ReadParquet.cpp#L221) is called with `int64_t` as the Chapel type and `int32_t` as the parquet type. That is the root cause of the issue that only arises at overflow boundaries.
- [This line](https://github.com/Bears-R-Us/arkouda/blob/master/src/parquet/ReadParquet.cpp#L78) allocates an array of `int32_t`s.
- We end up reading `2147483648` into this data. Note that that number is 1 in the MSB and 0 afterwards in 32 bits. That's the lower bound for a 32-bit integer. IOW, the most negative number as shown in the `int32` column (`-2147483648`).
- [This line](https://github.com/Bears-R-Us/arkouda/blob/master/src/parquet/ReadParquet.cpp#L86) casts that number, which is all 1s, but mathematically equivalent to `-2147483648` into `int64_t`. At that point, C has to keep the value same mathematically. In 2's complement that's achieved by filling the upper bits by 1s, resulting in `0b1111111111111111111111111111111110000000000000000000000000000000` which is equivalent to `18446744071562067968`

The solution here is to tighten up the usage of types in C. I have a branch where I am addressing some other Parquet issues. I will handle it there and revert the bit magic added here.

PR #2338, opened 2023-04-04
Closes #2320 - Adding Pytest structure to `array_view_test.py`

### LINKED ISSUE #2320: Remove class-based test structure for `array_view_test.py`
Remove class-based test structure for `array_view_test.py`

joshmarshall1:
This PR closes #2320 

Removes Unit test class-based test structure and implements PyTest structure and asserts

### DIFF
--- a/pytest/array_view_test.py
+++ b/pytest/array_view_test.py
@@ -1,0 +1,198 @@
+from itertools import product
+
+import numpy as np
+
+import arkouda as ak
+import pytest
+
+
+def test_mulitdimensional_array_creation():
+    n = np.array([[0, 0], [0, 1], [1, 1]])
+    a = ak.array([[0, 0], [0, 1], [1, 1]])
+    assert (n.tolist() == a.to_list())
+    n = np.arange(27).reshape((3, 3, 3))
+    a = ak.arange(27).reshape((3, 3, 3))
+    assert (n.tolist() == a.to_list())
+    n = np.arange(27).reshape(3, 3, 3)
+    a = ak.arange(27).reshape(3, 3, 3)
+    assert (n.tolist() == a.to_list())
+    n = np.arange(27, dtype=np.uint64).reshape(3, 3, 3)
+    a = ak.arange(27, dtype=ak.uint64).reshape(3, 3, 3)
+    assert (n.tolist() == a.to_list())
+
+
+def test_arrayview_int_indexing():
+    nd = np.arange(9).reshape(3, 3)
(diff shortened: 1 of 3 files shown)


pytest/conftest.py
Ethan-DeBandi99, line 1: could we just go ahead an make this a duplicate of `benchmarks_v2/conftest.py`? We will eventually be using these as 1 file, so I think it makes sense to just duplicate everything here. Obviously, if additional options are needed moving forward we can add and then reconcile once we merge them and most to the root `arkouda` directory.

pytest/array_view_test.py
```diff
@@ -0,0 +1,198 @@
...
+    nd_ind = [nd[i, j] for (i, j) in product(range(3), repeat=2)]
+    reshape_ind = [pd_reshape[i, j] for (i, j) in product(range(3), repeat=2)]
+    array_ind = [pd_array[i, j] for (i, j) in product(range(3), repeat=2)]
+    assert (nd_ind == reshape_ind)
+    assert (nd_ind == array_ind)
+
+    with pytest.raises(IndexError):
+        # index out bounds (>= dimension)
+        # index 3 is out of bounds for axis 0 with size 3
+        pd_reshape[3, 1]
```
Ethan-DeBandi99, line 38: Just for clarity and documentation of the changes, can you provide an example of the `pytest.raises` failing compared to the old test? Based on the `assert` functionality, I don't see it being an issue, but it would be good to document moving forward.

Ethan-DeBandi99 (commented):
No issues, just a few things to go through/document moving forward.

pytest/array_view_test.py
joshmarshall1: Since this is asserting that the error type was raised, on fails it really just says that the expected error wasn't raised
```
        with pytest.raises(IndexError):
            # index out bounds (>= dimension)
            # index 3 is out of bounds for axis 0 with size 3
>           pd_reshape[1, 1]
E           Failed: DID NOT RAISE <class 'IndexError'>
```

The previous implementation:
```
        with self.assertRaises(IndexError):
            # index out bounds (>= dimension)
            # index 3 is out of bounds for axis 0 with size 3
>           pd_reshape[1, 1]
E           AssertionError: IndexError not raised
```

pytest/conftest.py
joshmarshall1: Updated

Ethan-DeBandi99 (approved):

pierce314159 (approved):

jaketrookman (approved):

PR #2345, opened 2023-04-04
Closes #2343 - PyTest structure refactor for `bitops_test.py`

### LINKED ISSUE #2343: Remove class-based test structure for `bitops_test.py`
Remove class-based test structure for `bitops_test.py`

joshmarshall1:
This PR closes #2343 

Removes UnitTest class-based structure in `bitops_test.py` and refactors it to use PyTest formatting and assertions.

### DIFF
--- a/pytest/bitops_test.py
+++ b/pytest/bitops_test.py
@@ -1,0 +1,163 @@
+import numpy as np
+
+import arkouda as ak
+import pytest
+
+A = B = EDGE_CASES = EDGE_CASES_UINT = None
+
+
+def set_globals():
+    global A, B, EDGE_CASES, EDGE_CASES_UINT
+
+    A = ak.arange(10)
+    B = ak.cast(A, ak.uint64)
+    EDGE_CASES = ak.array([-(2**63), -1, 2**63 - 1])
+    EDGE_CASES_UINT = ak.cast(ak.array([-(2**63), -1, 2**63 - 1]), ak.uint64)
+
+
+def test_popcount():
+    if A is None:
+        set_globals()
+
+    # Method invocation
+    # Toy input
+    ans = [0, 1, 1, 2, 1, 2, 2, 3, 1, 2]
+    assert A.popcount().to_list() == ans
+
+    # Test uint case
+    ans = np.array(ans, ak.uint64)
+    assert B.popcount().to_list() == ans.tolist()
+
+    # Function invocation
+    # Edge case input
+    ans = [1, 64, 63]
+    assert ak.popcount(EDGE_CASES).to_list() == ans
+
+    # Test uint case
+    ans = np.array(ans, ak.uint64)
+    assert ak.popcount(EDGE_CASES_UINT).to_list() == ans.tolist()
+
+
+def test_parity():
+    if A is None:
+        set_globals()
+
+    ans = [0, 1, 1, 0, 1, 0, 0, 1, 1, 0]
+    assert A.parity().to_list() == ans
+
+    # Test uint case
+    ans = np.array(ans, ak.uint64)
+    assert B.parity().to_list() == ans.tolist()
+
+    ans = [1, 0, 1]
+    assert ak.parity(EDGE_CASES).to_list() == ans
+
+    # Test uint case
+    ans = np.array(ans, ak.uint64)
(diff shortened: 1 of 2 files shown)


pytest/bitops_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # Test uint case
+    ans = np.array(ans, ak.uint64)
+    assert ak.ctz(EDGE_CASES_UINT).to_list() == ans.tolist()
+
+
+def test_dtypes():
+    if A is None:
+        set_globals()
+
```
Ethan-DeBandi99, line 101: These values don't need to be set here.

pytest/bitops_test.py
```diff
@@ -0,0 +1,163 @@
...
+A = B = EDGE_CASES = EDGE_CASES_UINT = None
+
+
+def set_globals():
+    global A, B, EDGE_CASES, EDGE_CASES_UINT
+
+    A = ak.arange(10)
+    B = ak.cast(A, ak.uint64)
+    EDGE_CASES = ak.array([-(2**63), -1, 2**63 - 1])
+    EDGE_CASES_UINT = ak.cast(ak.array([-(2**63), -1, 2**63 - 1]), ak.uint64)
```
Ethan-DeBandi99, line 15: I am not sure we want to use globals here. I am thinking using a fixture may be better. Example below:

```python
@pytest.fixture
def bitops_data():
    return {
        'A': ak.arange(10),
        'B': ak.cast(A, ak.uint64),
        'EDGE_CASES': ak.array([-(2**63), -1, 2**63 - 1]),
        'EDGE_CASES_UINT': ak.cast(ak.array([-(2**63), -1, 2**63 - 1]), ak.uint64)
     }
```

Then in each test requiring the information you would pass `bitops_data` as a parameter and access each component like you would any other dictionary.

Ethan-DeBandi99 (changes requested):
Tests look good overall. We should replace the globals with a fixture though.

pytest/bitops_test.py
joshmarshall1: Removed

pytest/bitops_test.py
joshmarshall1: Added `@pytest.fixture(scope="function", autouse=True)` 

Just `@pytest.fixture` was causing a lot of logging statements in the run, bloating the console output since the fixture was being ran on every test without the file

Ethan-DeBandi99 (commented):
I don't have an issue with the current implementation. Since this is still in the prototyping phase I left some comments that I think it would be beneficial to have some discussion around. There are a few cases that I think moving to the class structure would be beneficial, but others it may be more cumbersome. I would like to have a bit of discussion around this before we continue moving forward on the updates to our testing.

pytest/bitops_test.py
```diff
@@ -0,0 +1,141 @@
+import numpy as np
+
+import arkouda as ak
+import pytest
+
+
+@pytest.fixture(scope="function", autouse=True)
```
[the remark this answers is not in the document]
joshmarshall1: Strangely, not setting the scope here causes info statements of array creation and deletion per function to be written into the output. Setting it here fixes that.

pytest/bitops_test.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@pytest.fixture(scope="function", autouse=True)
+def bitops_data():
+    pytest.bitops_data = {
+        'A': ak.arange(10),
+        'B': ak.cast(ak.arange(10), ak.uint64),
+        'EDGE_CASES': ak.array([-(2**63), -1, 2**63 - 1]),
+        'EDGE_CASES_UINT': ak.cast(ak.array([-(2**63), -1, 2**63 - 1]), ak.uint64)
+    }
```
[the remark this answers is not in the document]
joshmarshall1: The specific scope declaration was the result of the fixture output of `pdarray` deletion, not the var/return decision. Since the scope is `function`, recreation will be the case regardless of using `return` or this method. I thought this way was cleaner which is the only reason I did it this way. I'm not opposed to changing it if that's the path we want to take

joshmarshall1:
Closing this PR as per discussion with @Ethan-DeBandi99 @pierce314159 @jaketrookman about restructuring our test suite. We will need to look into scope and class structure more due to the questions raised in this PR

PR #2349, opened 2023-04-05
Closes #2293: `ak.where` support Strings and Categoricals

### LINKED ISSUE #2293: `where` support for strings and categoricals
Extend `ak.where` to support Strings and Categoricals

pierce314159:
This PR (closes #2293) adds `where` support Strings, Categoricals, and string literals. Adds testing for this functionality

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -526,12 +526,6 @@ def _str_cat_where(
     from arkouda.categorical import Categorical
     from arkouda.pdarraysetops import concatenate
 
-    if not isinstance(A, (str, Strings, Categorical)) or not isinstance(B, (str, Strings, Categorical)):
-        raise TypeError(
-            "both A and B must be an int, np.int64, float, np.float64, pdarray OR"
-            " both A and B must be an str, Strings, Categorical"
-        )
-
     if isinstance(A, str) and isinstance(B, (Categorical, Strings)):
         # This allows us to assume if a str is present it is B
         A, B, condition = B, A, ~condition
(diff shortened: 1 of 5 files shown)


arkouda/numeric.py
```diff
@@ -511,10 +515,89 @@ def _hash_single(pda: pdarray, full: bool = True):
...
+    # added @no_type_check because mypy can't handle Categorical not being declared
+    # sooner, but there are circular dependencies preventing that
+    from arkouda.categorical import Categorical
+    from arkouda.pdarraysetops import concatenate
+
+    if not isinstance(A, (str, Strings, Categorical)) or not isinstance(B, (str, Strings, Categorical)):
+        raise TypeError(
+            "both A and B must be an int, np.int64, float, np.float64, pdarray OR"
+            " both A and B must be an str, Strings, Categorical"
+        )
```
Ethan-DeBandi99, line 533: I think this error message is correct overall, but this function is only handling `str`, `Categorical` or `Strings` so I think it should just be the 2nd part of the error message here.

Ethan-DeBandi99 (approved):
Looks good. Just one error message that may need to be adjusted.

arkouda/numeric.py
pierce314159: This error message also captures the previous error message
```python
raise TypeError("both A and B must be an int, np.int64, float, np.float64, or pdarray")
```
got changed to this
```python
return _str_cat_where(condition, A, B)
```
So we only get here if it's not a regular where case or a str/cat where case

Ethan-DeBandi99: True, I am just wondering if we shouldn't have a check to ensure those cases in the `where` function instead.

pierce314159: im totally down to update this! In my head i was thinking "check if we're in the pdarray case. If we're not try the str/cat case. If not that either then error" but i don't mind moving that logic into `where`

I think the only thing that gave me pause was I didn't want to have do the `import Categorical` any further up the chain than I needed to because it can cause issues sometimes, but as long as that works im down with this flow. I'll try this out first thing tomorrow

Ethan-DeBandi99: Sounds good. I am not hard set on either direction, it just seemed counterintuitive at first and I wanted to raise the question.

joshmarshall1: I would agree with Ethan in that this wasn't initially intuitive, but I'm also not hard pressed either way

joshmarshall1 (approved):

arkouda/numeric.py
pierce314159: done

PR #2351, opened 2023-04-05
Closes #2015 - Efficient NaN read from Parquet

### LINKED ISSUE #2015: Parquet Read `NaN` Efficiently
PR #2012 was put up to resolve errors when reading float/double columns in Parquet that contained `NaN`. The current fix is less efficient than reading larger batches of `batchSize` (currently reading one row at a time).  

Need to research and determine if we can properly return `NaN` values when reading while also using larger batch sizes to read.

Ethan-DeBandi99:
Closes #2015 

Previously Parquet read workflows for `float` and `double` types were updated to allow reading of NaN values. This update only read value by value, which is not the most efficient. This PR updates these workflows to read in batches and identify the NaN values using the definition level and a temporary array. Because we are able to read the data in chunks and then process/set the return, performance is improved. It is also worth noting that List columns (used for SegArrays) do not require a check of the Definition Level to set NaN values.

The example runtimes below were run on a single locale with a DataFrame containing 2 columns, one with all NaN values and one with some NaN values. The DataFrame contained 10**8 rows.

**Before this PR**
```
Runtime: 4.234747886657715
```

**After this PR**
```
Runtime: 1.29227876663208
```

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -721,7 +721,8 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
           (void)reader->ReadBatch(batchSize, def_lvl, rep_lvl, tmpArr, &values_read);
           // copy values to Chapel array. Convert to double if not NaN
           for (int64_t j = 0; j < batchSize; j++){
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Looks good!

PR #2352, opened 2023-04-05
Closes #2337 - Fixes `ak.DataFrame.to_parquet` with IPv4 Columns

### LINKED ISSUE #2337: to_parquet() fails on IPv4 and multiple columns
**Describe the bug**
ak.DataFrame.to_parquet()  fails if a dataframe contains type IPv4 and there is more than one column

**To Reproduce**
```
df=ak.DataFrame({
    'a':ak.arange(10),
    'b':ak.IPv4(ak.arange(10))
})
```

**Expected behavior**
Expect the data to be written, however the IPv4 type will probably be lost and need to be restored after reading.

**Error Message**
"TypeError: Lists of multiple types can only contain strings and pdarray. Found IPv4, pdarray"
(Note: interesting error results if both "a" and "b" are IPv4 in df above.

**Is this a Blocking Issue**
Non blocking. Just need to specifically change IPv4 types to ak.uint64 before writing.

**`ak.get_config()` Output**

**Additional context**

Ethan-DeBandi99:
Closes #2337 

This PR adjusts the message parameter creation workflow to better support lists containing varying types. This allows IPv4 to properly be saved. 

As an additional note, this should make things much more flexible overall with list parameters for messages.

Adds testing to ensure 1 or more IPv4 columns can be saved in a dataframe.

### DIFF
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -7,8 +7,7 @@ from typing import Dict
 
 from typeguard import typechecked
 
-import arkouda
-from arkouda.dtypes import bigint
+from arkouda.dtypes import bigint, all_scalars
 
 
 class ObjectType(Enum):
@@ -135,7 +134,7 @@ class ParameterObject:
                     or isinstance(p, Strings)
                     or isinstance(p, SegArray)
                     or isinstance(p, str)
-                    or isinstance(p, arkouda.dtypes.all_scalars)
+                    or isinstance(p, all_scalars)
                 ):
                     raise TypeError(
                         f"List parameters must be pdarray, Strings, SegArray, str or a type "
@@ -143,7 +142,7 @@ class ParameterObject:
                         f"does not meet that criteria."
                     )
             t = "mixed"
-        data = [str(p) if isinstance(p, arkouda.dtypes.all_scalars) else p.name for p in val]
+        data = [str(p) if isinstance(p, all_scalars) else p.name for p in val]
         return ParameterObject(key, ObjectType.LIST, t, json.dumps(data))
 
     @staticmethod
--- a/arkouda/message.py
+++ b/arkouda/message.py
@@ -108,11 +108,8 @@ class ParameterObject:
         import numpy as np
 
         return (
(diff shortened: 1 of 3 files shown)


arkouda/message.py
```diff
@@ -103,6 +102,19 @@ def _build_strings_param(key: str, val) -> ParameterObject:
...
+        import builtins
+        import numpy as np
+
+        return (
+            isinstance(val, str)
+            or isinstance(val, np.str_)
+            or isSupportedNumber(val)
+            or isinstance(val, builtins.bool)
+            or isinstance(val, np.bool_)
+        )
```
pierce314159, line 116: i think we could make this a shorter by combining the `isinstance`s
```python
return isSupportedNumber(val) or isinstance(val, (str, np.str_, builtins.bool, np.bool_))
```

arkouda/message.py
```diff
@@ -129,19 +141,21 @@ def _build_list_param(key: str, val: list) -> ParameterObject:
...
-            # need all values to be str for chapel to read list properly
-            v = val if t == str.__name__ else [str(x) for x in val]
-            return ParameterObject(key, ObjectType.LIST, t, json.dumps(v))
+            for p in val:
+                if not (
+                    isinstance(p, pdarray)
+                    or isinstance(p, Strings)
+                    or isinstance(p, SegArray)
+                    or ParameterObject._is_supported_value(p)
+                ):
```
pierce314159, line 150: same `isinstance` suggestion here
```python
if not (
    isinstance(p, (pdarray, Strings, SegArray)) or ParameterObject._is_supported_value(p)
):
```

pierce314159 (approved):
Looks good!

jaketrookman (approved):

arkouda/message.py
Ethan-DeBandi99: Oh yeah good catch. This is updated.

arkouda/message.py
Ethan-DeBandi99: updated

PR #2356, opened 2023-04-05
Closes #1855: Implement `divmod`

### LINKED ISSUE #1855: Implement `divmod`
Implement an `ak.divmod(pda1,pda2)` which behaves like [`np.divmod`](https://numpy.org/doc/stable/reference/generated/numpy.divmod.html). For now don't worry about any of the flags other than the pdarrays.

So `ak.divmod(x,y)` should return `x//y, x%y`. This would be fairly easy to implement client-side, but we should write a server function to do this

jaketrookman:
This PR (closes #1855)

- Implements `divmod` on the client side.
- Adds testing to ensure proper implementation

Note: uints are unsupported, as we do not yet handle them in our current binops. Can be added in the future.


arkouda/pdarrayclass.py
```diff
@@ -2341,6 +2342,82 @@ def corr(x: pdarray, y: pdarray) -> np.float64:
...
+
+    if not isinstance(x, pdarray) and not isinstance(y, pdarray):
+        raise TypeError("At least one entry must be a pdarray.")
+
+    if isinstance(x, pdarray) and isinstance(y, pdarray):
+        if x.size != y.size:
+            raise ValueError(f"size mismatch {x.size} {y.size}")
+
+    if (not isinstance(y, pdarray) and y == 0) or (isinstance(y, pdarray) and not all(y)):
+        raise ZeroDivisionError("Can not divide by zero")
```
pierce314159, line 2403: This line is what is preventing you from doing `divmod` on `uint` arrays. There's no `ak.all` reduction defined for `uint`. If you change the `not all(y)` to `any(y == 0)` then there's no reason you can't do `divmod(uint_pda, uint_pda)`


I think the cleanest way to do this is something like
```python
equal_zero = y == 0
if equal_zero if isinstance(equal_zero, bool) else any(equal_zero):
```

```python
>>> x = ak.arange(1, 11, dtype=ak.uint64)
>>> y = ak.ones(10, dtype=ak.uint64)
>>> x
array([1 2 3 4 5 6 7 8 9 10])

>>> y
array([1 1 1 1 1 1 1 1 1 1])

>>> x.dtype
dtype('uint64')

>>> y.dtype
dtype('uint64')

>>> ak.divmod(x,y)
(array([1 2 3 4 5 6 7 8 9 10]), array([0 0 0 0 0 0 0 0 0 0]))

>>> ak.divmod(y,x)
(array([1 0 0 0 0 0 0 0 0 0]), array([0 1 1 1 1 1 1 1 1 1]))
```

arkouda/pdarrayclass.py
```diff
@@ -2341,6 +2342,82 @@ def corr(x: pdarray, y: pdarray) -> np.float64:
...
+
+    if where is True:
+        return x // y, x % y  # type: ignore
+    elif where is False:
+        if not isinstance(x, pdarray) and isinstance(y, pdarray):
+            x = full(y.size, x)
+            return x, x  # type: ignore
+        else:
+            return x, x  # type: ignore
```
pierce314159, line 2412: we don't need 2 returns here, we can do 
```python
if not isinstance(x, pdarray) and isinstance(y, pdarray):
    x = full(y.size, x)
return x, x  # type: ignore
```

arkouda/pdarrayclass.py
```diff
@@ -2341,6 +2342,82 @@ def corr(x: pdarray, y: pdarray) -> np.float64:
...
+    else:
+        div = x // y
+        mod = x % y
+        return (
+            akwhere(where, div, akcast(x, div.dtype)),  # type: ignore
+            akwhere(where, mod, akcast(x, mod.dtype)))  # type: ignore
```
pierce314159, line 2418: we really should run black on this file, but that will mess up the `# type: ignore`s. We could use `# fmt off/on` to avoid that but I think it's nicer to cast to pdarray and avoid the `# type: ignore` instead

```python
div = cast(pdarray, x // y)
mod = cast(pdarray, x % y)
return (
    akwhere(where, div, akcast(x, div.dtype)),
    akwhere(where, mod, akcast(x, mod.dtype)),
)
```

tests/stats_test.py
```diff
@@ -123,3 +123,106 @@ def test_corr_matrix(self):
...
+        np_div, np_mod = np.divmod(30.0, self.npy)
+        self.assertListEqual(ak_div.to_list(), np_div.tolist())
+        self.assertListEqual(ak_mod.to_list(), np_mod.tolist())
+
+        # float float
+        ak_div, ak_mod = ak.divmod(30.0, ak.cast(self.y, ak.float64))
+        np_div, np_mod = np.divmod(30.0, self.npy.astype(float))
+        self.assertListEqual(ak_div.to_list(), np_div.tolist())
+        self.assertListEqual(ak_mod.to_list(), np_mod.tolist())
+
```
pierce314159, line 177: for the float cases (at the very least both float) we should consider doing something like `self.f + 0.5` instead to have a non-round float

pierce314159 (approved):
Looks good!

joshmarshall1 (approved):

PR #2367, opened 2023-04-06
Closes #2366: Update graph files after IO benchmark changes

### LINKED ISSUE #2366: Update graph files after IO changes
Now that the IO benchmark is saving the write performance for all of the different compression formats, the graph file needs to be updated to reflect that.

bmcdonald3:
Now that the IO benchmark is saving all of the different compression performance data, the graph file needs to be updated so that we are also graphing those new data points.


ronawho (approved):

ronawho:
I approved since this fixes an existing bug, but FWIW I wonder how busy the graphs will be now. I also wonder how much time running all the compressions in nightly is taking (and if the results are interesting enough to justify that time)

Ethan-DeBandi99 (approved):

Ethan-DeBandi99:
The ability to benchmark the different compression types was a requirement we got for performance testing the systems running Arkouda. Maybe we can separate the graphs, but the comparison is needed regardless.

PR #2371, opened 2023-04-07
Closes #2370 - Fixes Deprecation Warnings during `make` 

### LINKED ISSUE #2370: Deprecation Warnings when Running Make
When running `make` for Arkouda, we are now getting deprecation warnings within the Metrics code. The code here should be updated to remove references to deprecated code.

**Example of Warnings**
```
arkouda//src/MetricsMsg.chpl:206: In method 'add':
arkouda//src/MetricsMsg.chpl:207: warning: 'Map.set' is deprecated. Please use 'Map.replace' instead.
```

Ethan-DeBandi99:
Closes #2370 

Updates call from `Map.set` to `Map.replace` to remove deprecation warnings being seen when running `make` on Arkouda.

### DIFF
--- a/src/MetricsMsg.chpl
+++ b/src/MetricsMsg.chpl
@@ -204,7 +204,7 @@ module MetricsMsg {
          * metric name, setting the value if the metric does not exist.
          */
         proc add(metric: string, measurement: real) throws {
-            this.measurements.set(metric,(this.get(metric) + measurement));
+            this.measurements.replace(metric,(this.get(metric) + measurement));
         }
 
         iter items() {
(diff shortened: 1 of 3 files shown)


pierce314159 (approved):

Ethan-DeBandi99:
Spoke with @hokiegeek2 on this. Making the change to prevent deprecation warnings will break if being used with Chapel 1.29. I believe we just need to add a compat for this... Or there may already be one. @bmcdonald3, @ronawho - can you advise on the best way to handle this

hokiegeek2 (approved):
nice work, all looks good to me. Seems kinda odd that the deprecation warning was just added in Chapel 1.30.0 and that map.reduce is a breaking change in 1.29.0

PR #2376, opened 2023-04-12
Closes #2304 - Innerjoin support for Strings and Categorical

### LINKED ISSUE #2304: Join on segstrings
Add support for performing `inner_join` on Strings. In particular between two columns of a dataframe

We should also look into verifying/adding this functionality for categorical

joshmarshall1:
This PR Closes #2304

Adds support for Strings and Categorical to the `ak.join.innerjoin` function. Includes `where` support against `pdarray, Strings, and Categorical` types using `pdarray` or `Strings` arguments in the where.

This was accomplished by converting `Strings` to `Categorical` and using `categorical.codes` as a numeric representation which could then follow the same workflow and processes as the original `pdarray` case. 

Adds test case for `Strings` and `Categorical` as well as a test using a `wherefunc` for each type.

### DIFF
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -1612,13 +1612,13 @@ class GroupBy:
                     "permName": self.permutation.name,
                     "segName": self.segments.name,
                     "valName": values.name,
-                    "objType": values.objtype,
+                    "objType": values.objType,
                     "permute": permute,
                     "size": self.length,
                 },
             ),
         )
-        if values.objtype == Strings.objtype:
+        if values.objType == Strings.objType:
             return Strings.from_return_msg(repMsg)
         else:
             return create_pdarray(repMsg)
@@ -2109,13 +2109,13 @@ def broadcast(
                 "permName": pname,
                 "segName": segments.name,
                 "valName": values.name,
-                "objType": values.objtype,
+                "objType": values.objType,
                 "permute": permute,
                 "size": size,
             },
         ),
     )
-    if values.objtype == Strings.objtype:
+    if values.objType == Strings.objType:
         return Strings.from_return_msg(repMsg)
     else:
         return create_pdarray(repMsg)
--- a/arkouda/groupbyclass.py
+++ b/arkouda/groupbyclass.py
@@ -1543,7 +1543,9 @@ class GroupBy:
         return self.unique_keys, ret  # type: ignore
 
     @typechecked
-    def broadcast(self, values: pdarray, permute: bool = True) -> pdarray:
+    def broadcast(
+        self, values: Union[pdarray, Strings], permute: bool = True
+    ) -> Union[pdarray, Strings]:
         """
         Fill each group's segment with a constant value.
 
@@ -1602,17 +1604,24 @@ class GroupBy:
         if values.size != self.segments.size:
             raise ValueError("Must have one value per segment")
         cmd = "broadcast"
-        repMsg = generic_msg(
-            cmd=cmd,
-            args={
-                "permName": self.permutation.name,
-                "segName": self.segments.name,
-                "valName": values.name,
-                "permute": permute,
-                "size": self.length,
-            },
+        repMsg = cast(
+            str,
+            generic_msg(
+                cmd=cmd,
+                args={
+                    "permName": self.permutation.name,
+                    "segName": self.segments.name,
+                    "valName": values.name,
+                    "objType": values.objtype,
+                    "permute": permute,
+                    "size": self.length,
+                },
+            ),
         )
-        return create_pdarray(repMsg)
+        if values.objtype == Strings.objtype:
+            return Strings.from_return_msg(repMsg)
+        else:
+            return create_pdarray(repMsg)
 
     @staticmethod
     def build_from_components(user_defined_name: str = None, **kwargs) -> GroupBy:
@@ -2092,14 +2101,21 @@ def broadcast(
     if size < 1:
         raise ValueError("result size must be greater than zero")
     cmd = "broadcast"
-    repMsg = generic_msg(
-        cmd=cmd,
-        args={
-            "permName": pname,
-            "segName": segments.name,
-            "valName": values.name,
-            "permute": permute,
-            "size": size,
-        },
+    repMsg = cast(
+        str,
+        generic_msg(
+            cmd=cmd,
+            args={
+                "permName": pname,
+                "segName": segments.name,
+                "valName": values.name,
+                "objType": values.objtype,
+                "permute": permute,
+                "size": size,
+            },
+        ),
     )
-    return create_pdarray(repMsg)
+    if values.objtype == Strings.objtype:
+        return Strings.from_return_msg(repMsg)
+    else:
+        return create_pdarray(repMsg)
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -3,7 +3,7 @@ from typing import Callable, Tuple, Union, cast
 import numpy as np  # type: ignore
 from typeguard import typechecked
 
-from arkouda.alignment import right_align
+from arkouda.alignment import right_align, align
 from arkouda.categorical import Categorical
 from arkouda.client import generic_msg
 from arkouda.dtypes import NUMBER_FORMAT_STRINGS
@@ -209,25 +209,16 @@ def inner_join(
     if type(left) != type(right):
         raise TypeError("Left and Right arrays must be of same type")
 
-    # Convert Strings and Categorical left & right values to Categorical.codes. This generates an
-    # index representation of the strings. This way they can be handled as numerics and follow the
-    # same process as int
-    t = cast(str, pdarray.objtype)
-    cats = None
+    whereLeft = whereRight = None
+    if whereargs is not None:
+        whereLeft, whereRight = whereargs[0:2]
 
-    if isinstance(left, Strings) and isinstance(right, Strings):
-        left = Categorical(left)
-        right = Categorical(right)
+    if (isinstance(left, Strings) and isinstance(right, Strings)) or \
+            (isinstance(left, Categorical) and isinstance(right, Categorical)):
 
-    if isinstance(left, Categorical) and isinstance(right, Categorical):
-        t = cast(str, Categorical.objtype)
-        left, right = Categorical.standardize_categories([left, right])
-        if isinstance(left, Categorical) and isinstance(right, Categorical):
-            cats = left.categories[left.codes]  # Get standardized categories without NAvalue
-            leftKeepCodes = left.isna() == 0  # Invert bool array
-            rightKeepCodes = right.isna() == 0  # Invert bool array
-            left = left.codes[leftKeepCodes]
-            right = right.codes[rightKeepCodes]
+        left, right = align(left, right)
+        if whereargs is not None:
+            whereLeft, whereRight = align(whereLeft, whereRight)
 
     sample = np.min((left.size, right.size, 5))  # type: ignore
     if wherefunc is not None:
@@ -235,14 +226,14 @@ def inner_join(
             raise ValueError("wherefunc must be a function that accepts exactly two arguments")
         if whereargs is None or len(whereargs) != 2:
             raise ValueError("whereargs must be a 2-tuple with left and right arg arrays")
-        if whereargs[0].size != left.size:
+        if whereLeft.size != left.size:
             raise ValueError("Left whereargs must be same size as left join values")
-        if whereargs[1].size != right.size:
+        if whereRight.size != right.size:
             raise ValueError("Right whereargs must be same size as right join values")
-        if isinstance(whereargs[0], Strings) and t == pdarray.objtype:
+        if isinstance(whereLeft, Strings) and isinstance(left, pdarray):
             raise TypeError("Strings whereargs are only supported for Strings left and right arg arrays")
         try:
-            _ = wherefunc(whereargs[0][:sample], whereargs[1][:sample])
+            _ = wherefunc(whereLeft[:sample], whereRight[:sample])
         except Exception as e:
             raise ValueError("Error evaluating wherefunc") from e
 
@@ -267,43 +258,10 @@ def inner_join(
         keep12 = keep
     else:
         if whereargs is not None:
-            if (
-                t == Categorical.objtype
-                and isinstance(cats, Strings)
-                and isinstance(whereargs[0], Strings)
-            ):
-                # Find the corresponding codes value for each term in the whereargs
-                repMsg = generic_msg(
-                    cmd="codeMapping",
-                    args={
-                        "categories": cats,
-                        "query": whereargs[0][keep],  # type: ignore
-                    },
-                )
-                lCodes = create_pdarray(repMsg)
-
-                if isinstance(lCodes, pdarray):
-                    lIdx = broadcast(fullSegs, lCodes, ranges.size)
-                else:
-                    raise TypeError(
-                        f"Invalid type returned from lCodes generation: {type(lCodes)}. Expected pdarray"
-                    )
-                leftWhere = whereargs[0][lIdx]
-                # Gather right whereargs
-                repMsg = generic_msg(
-                    cmd="codeMapping",
-                    args={
-                        "categories": cats,
-                        "query": whereargs[1][byRight.permutation][ranges],  # type: ignore
-                    },
-                )
-                rCodes = create_pdarray(repMsg)
-                rightWhere = whereargs[1][rCodes]
-            else:
-                # Expand left whereargs
-                leftWhere = broadcast(fullSegs, whereargs[0][keep], ranges.size)
-                # Gather right whereargs
-                rightWhere = whereargs[1][byRight.permutation][ranges]
+            # Expand left whereargs
+            leftWhere = broadcast(fullSegs, whereLeft[keep], ranges.size)
+            # Gather right whereargs
+            rightWhere = whereRight[byRight.permutation][ranges]
 
             # Evaluate wherefunc and filter ranges, recompute segments
             whereSatisfied = wherefunc(leftWhere, rightWhere)
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -13,6 +13,7 @@ from arkouda.numeric import cumsum
 from arkouda.pdarrayclass import create_pdarray, pdarray
 from arkouda.pdarraycreation import arange, array, ones, zeros
 from arkouda.pdarraysetops import concatenate, in1d
+from arkouda.strings import Strings
 
 __all__ = ["join_on_eq_with_dt", "gen_ranges", "compute_join_size"]
 
@@ -169,7 +170,8 @@ def compute_join_size(a: pdarray, b: pdarray) -> Tuple[int, int]:
 
 @typechecked
 def inner_join(
-    left: pdarray, right: pdarray, wherefunc: Callable = None, whereargs: Tuple[pdarray, pdarray] = None
+    left: Union[pdarray, Strings], right: Union[pdarray, Strings], wherefunc: Callable = None,
+        whereargs: Tuple[Union[pdarray, Strings], Union[pdarray, Strings]] = None
 ) -> Tuple[pdarray, pdarray]:
     """Perform inner join on values in <left> and <right>,
     using conditions defined by <wherefunc> evaluated on
@@ -177,9 +179,9 @@ def inner_join(
 
     Parameters
     ----------
-    left : pdarray(int64)
+    left : pdarray(int64), Strings
         The left values to join
-    right : pdarray(int64)
+    right : pdarray(int64), Strings
         The right values to join
     wherefunc : function, optional
         Function that takes two pdarray arguments and returns
@@ -204,6 +206,9 @@ def inner_join(
 
     """
     from inspect import signature
+
+    if type(left) != type(right):
+        raise TypeError("Left and Right arrays must be of same type")
 
     sample = np.min((left.size, right.size, 5))  # type: ignore
     if wherefunc is not None:
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -3,7 +3,7 @@ from typing import Callable, Tuple, Union, cast
 import numpy as np  # type: ignore
 from typeguard import typechecked
 
-from arkouda.alignment import right_align, align
+from arkouda.alignment import align, right_align
 from arkouda.categorical import Categorical
 from arkouda.client import generic_msg
 from arkouda.dtypes import NUMBER_FORMAT_STRINGS
@@ -213,8 +213,9 @@ def inner_join(
     if whereargs is not None:
         whereLeft, whereRight = whereargs[0:2]
 
-    if (isinstance(left, Strings) and isinstance(right, Strings)) or \
-            (isinstance(left, Categorical) and isinstance(right, Categorical)):
+    if (isinstance(left, Strings) and isinstance(right, Strings)) or (
+        isinstance(left, Categorical) and isinstance(right, Categorical)
+    ):
 
         left, right = align(left, right)
         if whereargs is not None:
@@ -226,16 +227,17 @@ def inner_join(
             raise ValueError("wherefunc must be a function that accepts exactly two arguments")
         if whereargs is None or len(whereargs) != 2:
             raise ValueError("whereargs must be a 2-tuple with left and right arg arrays")
-        if whereLeft.size != left.size:
+        if whereLeft is not None and whereLeft.size != left.size:
             raise ValueError("Left whereargs must be same size as left join values")
-        if whereRight.size != right.size:
+        if whereRight is not None and whereRight.size != right.size:
             raise ValueError("Right whereargs must be same size as right join values")
         if isinstance(whereLeft, Strings) and isinstance(left, pdarray):
             raise TypeError("Strings whereargs are only supported for Strings left and right arg arrays")
-        try:
-            _ = wherefunc(whereLeft[:sample], whereRight[:sample])
-        except Exception as e:
-            raise ValueError("Error evaluating wherefunc") from e
+        if whereLeft is not None and whereRight is not None:
+            try:
+                _ = wherefunc(whereLeft[:sample], whereRight[:sample])
+            except Exception as e:
+                raise ValueError("Error evaluating wherefunc") from e
 
     # Need dense 0-up right index, to filter out left not in right
     keep, (denseLeft, denseRight) = right_align(left, right)
@@ -257,7 +259,7 @@ def inner_join(
         filtSegs = fullSegs
         keep12 = keep
     else:
-        if whereargs is not None:
+        if whereLeft is not None and whereRight is not None:
             # Expand left whereargs
             leftWhere = broadcast(fullSegs, whereLeft[keep], ranges.size)
             # Gather right whereargs
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -3,7 +3,7 @@ from typing import Callable, Tuple, Union, cast
 import numpy as np  # type: ignore
 from typeguard import typechecked
 
-from arkouda.alignment import right_align
+from arkouda.alignment import right_align, align
 from arkouda.categorical import Categorical
 from arkouda.client import generic_msg
 from arkouda.dtypes import NUMBER_FORMAT_STRINGS
@@ -213,25 +213,16 @@ def inner_join(
     if type(left) != type(right):
         raise TypeError("Left and Right arrays must be of same type")
 
-    # Convert Strings and Categorical left & right values to Categorical.codes. This generates an
-    # index representation of the strings. This way they can be handled as numerics and follow the
-    # same process as int
-    t = cast(str, pdarray.objtype)
-    cats = None
+    whereLeft = whereRight = None
+    if whereargs is not None:
+        whereLeft, whereRight = whereargs[0:2]
 
-    if isinstance(left, Strings) and isinstance(right, Strings):
-        left = Categorical(left)
-        right = Categorical(right)
+    if (isinstance(left, Strings) and isinstance(right, Strings)) or \
+            (isinstance(left, Categorical) and isinstance(right, Categorical)):
 
-    if isinstance(left, Categorical) and isinstance(right, Categorical):
-        t = cast(str, Categorical.objtype)
-        left, right = Categorical.standardize_categories([left, right])
-        if isinstance(left, Categorical) and isinstance(right, Categorical):
-            cats = left.categories[left.codes]  # Get standardized categories without NAvalue
-            leftKeepCodes = left.isna() == 0  # Invert bool array
-            rightKeepCodes = right.isna() == 0  # Invert bool array
-            left = left.codes[leftKeepCodes]
-            right = right.codes[rightKeepCodes]
+        left, right = align(left, right)
+        if whereargs is not None:
+            whereLeft, whereRight = align(whereLeft, whereRight)
 
     sample = np.min((left.size, right.size, 5))  # type: ignore
     if wherefunc is not None:
@@ -239,14 +230,14 @@ def inner_join(
             raise ValueError("wherefunc must be a function that accepts exactly two arguments")
         if whereargs is None or len(whereargs) != 2:
             raise ValueError("whereargs must be a 2-tuple with left and right arg arrays")
-        if whereargs[0].size != left.size:
+        if whereLeft.size != left.size:
             raise ValueError("Left whereargs must be same size as left join values")
-        if whereargs[1].size != right.size:
+        if whereRight.size != right.size:
             raise ValueError("Right whereargs must be same size as right join values")
-        if isinstance(whereargs[0], Strings) and t == pdarray.objtype:
+        if isinstance(whereLeft, Strings) and isinstance(left, pdarray):
             raise TypeError("Strings whereargs are only supported for Strings left and right arg arrays")
         try:
-            _ = wherefunc(whereargs[0][:sample], whereargs[1][:sample])
+            _ = wherefunc(whereLeft[:sample], whereRight[:sample])
         except Exception as e:
             raise ValueError("Error evaluating wherefunc") from e
 
@@ -271,43 +262,10 @@ def inner_join(
         keep12 = keep
     else:
         if whereargs is not None:
-            if (
-                t == Categorical.objtype
-                and isinstance(cats, Strings)
-                and isinstance(whereargs[0], Strings)
-            ):
-                # Find the corresponding codes value for each term in the whereargs
-                repMsg = generic_msg(
-                    cmd="codeMapping",
-                    args={
-                        "categories": cats,
-                        "query": whereargs[0][keep],  # type: ignore
-                    },
-                )
-                lCodes = create_pdarray(repMsg)
-
-                if isinstance(lCodes, pdarray):
-                    lIdx = broadcast(fullSegs, lCodes, ranges.size)
-                else:
-                    raise TypeError(
-                        f"Invalid type returned from lCodes generation: {type(lCodes)}. Expected pdarray"
-                    )
-                leftWhere = whereargs[0][lIdx]
-                # Gather right whereargs
-                repMsg = generic_msg(
-                    cmd="codeMapping",
-                    args={
-                        "categories": cats,
-                        "query": whereargs[1][byRight.permutation][ranges],  # type: ignore
-                    },
-                )
-                rCodes = create_pdarray(repMsg)
-                rightWhere = whereargs[1][rCodes]
-            else:
-                # Expand left whereargs
-                leftWhere = broadcast(fullSegs, whereargs[0][keep], ranges.size)
-                # Gather right whereargs
-                rightWhere = whereargs[1][byRight.permutation][ranges]
+            # Expand left whereargs
+            leftWhere = broadcast(fullSegs, whereLeft[keep], ranges.size)
+            # Gather right whereargs
+            rightWhere = whereRight[byRight.permutation][ranges]
 
             # Evaluate wherefunc and filter ranges, recompute segments
             whereSatisfied = wherefunc(leftWhere, rightWhere)
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -213,21 +213,20 @@ def inner_join(
     # index representation of the strings. This way they can be handled as numerics and follow the
     # same process as int
     t = cast(str, pdarray.objtype)
-    lCats = rCats = None
+    cats = None
+
     if isinstance(left, Strings) and isinstance(right, Strings):
-        t = cast(str, Strings.objtype)
-        catLeft = Categorical(left)
-        catRight = Categorical(right)
-        lCats = catLeft.categories
-        rCats = catRight.categories
-        left = catLeft.codes
-        right = catRight.codes
-    elif isinstance(left, Categorical) and isinstance(right, Categorical):
+        left = Categorical(left)
+        right = Categorical(right)
+
+    if isinstance(left, Categorical) and isinstance(right, Categorical):
         t = cast(str, Categorical.objtype)
-        lCats = left.categories
-        rCats = right.categories
-        left = left.codes
-        right = right.codes
+        left, right = Categorical.standardize_categories([left, right])
+        cats = left.categories[left.codes]  # Get standardized categories without NAvalue
+        leftKeepCodes = array([x is not True for x in left.isna().to_ndarray()])  # Keep codes that aren't NAvalue
+        rightKeepCodes = array([x is not True for x in right.isna().to_ndarray()])  # Keep codes that aren't NAvalue
+        left = left.codes[leftKeepCodes]
+        right = right.codes[rightKeepCodes]
 
     sample = np.min((left.size, right.size, 5))  # type: ignore
     if wherefunc is not None:
@@ -265,28 +264,22 @@ def inner_join(
         keep12 = keep
     else:
         if whereargs is not None:
-            # Gather right whereargs
-            rightWhere = whereargs[1][byRight.permutation][ranges]
-            # Expand left whereargs
-            keep_where = whereargs[0][keep]
-            if isinstance(keep_where, (Strings, Categorical)):
-                leftWhereIdx = broadcast(fullSegs, arange(keep_where.size), ranges.size)
-                leftWhere = keep_where[leftWhereIdx]
-            else:
-                leftWhere = broadcast(fullSegs, keep_where, ranges.size)
-            if (
-                t in [Strings.objtype, Categorical.objtype]
-                and isinstance(lCats, Strings)
-                and isinstance(rCats, Strings)
-            ):
+            if t == Categorical.objtype and isinstance(cats, Strings):
                 # Find the corresponding codes value for each term in the whereargs
-                lCodes = array(
-                    [
-                        lCats.to_list().index(w)
-                        for w in whereargs[0][keep].to_ndarray()
-                        if w in lCats.to_ndarray()
-                    ]
-                )
+                repMsg = generic_msg(
+                    cmd="codeMapping",
+                    args={
+                        "categories": cats,
+                        "query": whereargs[0][keep],  # type: ignore
+                    })
+                lCodes = create_pdarray(repMsg)
+                # lCodes = array(
+                #     [
+                #         cats.to_list().index(w)
+                #         for w in whereargs[0][keep].to_ndarray()
+                #         if w in cats.to_ndarray()
+                #     ]
+                # )
                 if isinstance(lCodes, pdarray):
                     lIdx = broadcast(fullSegs, lCodes, ranges.size)
                 else:
@@ -295,13 +288,20 @@ def inner_join(
                     )
                 leftWhere = whereargs[0][lIdx]
                 # Gather right whereargs
-                rCodes = array(
-                    [
-                        rCats.to_list().index(w)
-                        for w in whereargs[1][byRight.permutation][ranges].to_ndarray()
-                        if w in rCats.to_ndarray()
-                    ]
-                )
+                repMsg = generic_msg(
+                    cmd="codeMapping",
+                    args={
+                        "categories": cats,
+                        "query": whereargs[1][byRight.permutation][ranges],  # type: ignore
+                    })
+                rCodes = create_pdarray(repMsg)
+                # rCodes = array(
+                #     [
+                #         cats.to_list().index(w)
+                #         for w in whereargs[1][byRight.permutation][ranges].to_ndarray()
+                #         if w in cats.to_ndarray()
+                #     ]
+                # )
                 rightWhere = whereargs[1][rCodes]
             else:
                 # Expand left whereargs
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -184,14 +184,14 @@ def inner_join(
     ----------
     left : pdarray(int64), Strings, Categorical
         The left values to join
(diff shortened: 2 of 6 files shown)


pierce314159 (changes requested):
I'm seeing a few incorrect results and an error (when using wherefunc). I'm also a bit concerned about the performance of a few parts namely converting segstrings to categorical and calling `to_ndarray/list`

Reproducer for incorrect results and error:
```python
# create comparable left and right for int and str
>>> int_left = ak.arange(10)
>>> int_right = ak.array([0, 5, 3, 3, 4, 6, 7, 9, 8, 1])
>>> str_left = ak.array([f'str {i}' for i in int_left.to_list()])
>>> str_right = ak.array([f'str {i}' for i in int_right.to_list()])

# output of these should be the same (it's not because non-standardized categories)
>>> ak.join.inner_join(int_left, int_right)
(array([0 1 3 3 4 5 6 7 8 9]), array([0 9 2 3 4 1 5 6 8 7]))

>>> ak.join.inner_join(str_left, str_right)
(array([0 1 2 2 3 4 5 6 7 8]), array([0 9 2 3 4 1 5 6 8 7]))

# upon indexing we see the int case meets the inner join condition
>>> sl, sr = ak.join.inner_join(str_left, str_right)
>>> il, ir = ak.join.inner_join(int_left, int_right)
>>> ak.all(int_left[il] == int_right[ir])
True

# but the str case does not
>>> ak.all(str_left[sl] == str_right[sr])
False

# if we declare a where function which will only operate on evens
>>> def where_func(x,y):
...    return x%2 == 0

>>> ak.join.inner_join(int_left, int_right, wherefunc = where_func, whereargs=(int_left,int_right))
(array([0 4 6 8]), array([0 4 5 8]))

# we see the str case breaks
>>> ak.join.inner_join(str_left, str_right, wherefunc = where_func, whereargs=(int_left,int_right))

ValueError: segments and values arrays must be same size
```

arkouda/join.py
```diff
@@ -205,6 +210,29 @@ def inner_join(
...
+
+    # Convert Strings and Categorical left & right values to Categorical.codes. This generates an
+    # index representation of the strings. This way they can be handled as numerics and follow the
+    # same process as int
+    t = cast(str, pdarray.objtype)
+    lCats = rCats = None
+    if isinstance(left, Strings) and isinstance(right, Strings):
+        t = cast(str, Strings.objtype)
+        catLeft = Categorical(left)
+        catRight = Categorical(right)
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I agree here. I do not believe this is going to scale well because there will be a lot of startup overhead.

Ethan-DeBandi99 (changes requested):
I agree with @pierce314159 on his evaluation. I would also add that I believe we need to significantly reinforce the testing of this because based on some of the issues I am seeing, it looks to be missing some things.

arkouda/join.py
```diff
@@ -205,6 +210,29 @@ def inner_join(
...
+        lCats = catLeft.categories
+        rCats = catRight.categories
+        left = catLeft.codes
+        right = catRight.codes
+    elif isinstance(left, Categorical) and isinstance(right, Categorical):
+        t = cast(str, Categorical.objtype)
+        lCats = left.categories
+        rCats = right.categories
+        left = left.codes
+        right = right.codes
```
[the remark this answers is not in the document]
joshmarshall1: Added standardize_categories

arkouda/join.py
```diff
@@ -205,6 +210,29 @@ def inner_join(
...
+    t = cast(str, pdarray.objtype)
+    lCats = rCats = None
+    if isinstance(left, Strings) and isinstance(right, Strings):
+        t = cast(str, Strings.objtype)
+        catLeft = Categorical(left)
+        catRight = Categorical(right)
+        lCats = catLeft.categories
+        rCats = catRight.categories
+        left = catLeft.codes
+        right = catRight.codes
```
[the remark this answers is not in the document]
joshmarshall1: Removed duplicate logic - added an additional Categorical check due to MyPy

arkouda/join.py
```diff
@@ -241,10 +269,41 @@ def inner_join(
...
+                and isinstance(rCats, Strings)
+            ):
+                # Find the corresponding codes value for each term in the whereargs
+                lCodes = array(
+                    [
+                        lCats.to_list().index(w)
+                        for w in whereargs[0][keep].to_ndarray()
+                        if w in lCats.to_ndarray()
+                    ]
+                )
```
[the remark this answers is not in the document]
joshmarshall1: Moved this and the `rCodes` equivalent to Chapel

arkouda/join.py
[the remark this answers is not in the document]
joshmarshall1: Performance testing on this proved less impactful than expected, the main performance hit comes from the categories -> codes remapping in the where case which I have moved to Chapel. There's likely more to be done to further improve performance, but it is 1000x better than it was when this comment was written.

For 10**4 test size, previous implementation took ~21 minutes to run with a where case
The new implementation runs 10**4 in 2.5 seconds, 10**5 in 8 seconds, and 10**6 in 55 seconds.

Obviously there's some exponential growth here that I'd like to find a way around, but for the first iteration of this functionality I think this is okay for now.

arkouda/join.py
```diff
@@ -241,10 +269,41 @@ def inner_join(
...
+                if isinstance(lCodes, pdarray):
+                    lIdx = broadcast(fullSegs, lCodes, ranges.size)
```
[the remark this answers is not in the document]
joshmarshall1: This was caused by the 'N/A' value of Categoricals not being taken into account after standardizing the categories. This is now resolved

joshmarshall1:
Performance testing results using new implementation:

10**4 previously ran on average in ~1,270 seconds (21 minutes)

`python -m pytest -c benchmark.ini --size=10**4 -k join_benchmark`
```
------------------------------------ benchmark 'Inner Join': 1 tests ------------------------------------
Name (time in s)        Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------
bench_inner_join     2.6474  2.9105  2.7826  0.0945  2.7727  0.0958       2;0  0.3594       5           1
---------------------------------------------------------------------------------------------------------
```

`python -m pytest -c benchmark.ini --size=10**5 -k join_benchmark`
```
------------------------------------ benchmark 'Inner Join': 1 tests ------------------------------------
Name (time in s)        Min     Max    Mean  StdDev  Median     IQR  Outliers     OPS  Rounds  Iterations
---------------------------------------------------------------------------------------------------------
bench_inner_join     7.8871  9.1806  8.5893  0.4869  8.5113  0.6267       2;0  0.1164       5           1
---------------------------------------------------------------------------------------------------------
```

`python -m pytest -c benchmark.ini --size=10**6 -k join_benchmark`
```
-------------------------------------- benchmark 'Inner Join': 1 tests --------------------------------------
Name (time in s)         Min      Max     Mean  StdDev   Median     IQR  Outliers     OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------
bench_inner_join     52.8868  54.6260  53.7980  0.7168  53.8436  1.2116       2;0  0.0186       5           1
-------------------------------------------------------------------------------------------------------------
```

The benchmark used to test:
```python
import arkouda as ak
import pytest


def join_helper(L, R):
    idx = []
    for i in range(L.size):
        if i % 2 == 0:
            idx.append(True)
        else:
            idx.append(False)

    return ak.array(idx)


@pytest.mark.benchmark(group="Inner Join")
def bench_inner_join(benchmark):
    N = pytest.prob_size
    leftInt = ak.randint(0, 2**16, N)
    rightInt = ak.arange(N)
    leftStr = ak.array([f"str {i}" for i in leftInt.to_ndarray()])
    rightStr = ak.array([f"str {i}" for i in rightInt.to_ndarray()])

    left, right = benchmark.pedantic(
        ak.join.inner_join,
        args=(leftStr, rightStr, join_helper, (leftStr, rightStr)),
        rounds=pytest.trials,
    )

    assert leftStr[left].to_list() == rightStr[right].to_list()
```

arkouda/join.py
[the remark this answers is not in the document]
pierce314159: @joshmarshall1 are these numbers with gasnet enabled or not?

pierce314159:
> Performance testing results using new implementation:
> 
> 10**4 previously ran on average in ~1,270 seconds (21 minutes)
> 
> `python -m pytest -c benchmark.ini --size=10**4 -k join_benchmark`
> 
I'd be curious to see these timings with gasnet enabled!

> ```python
> N = pytest.prob_size
> leftInt = ak.randint(0, 2**16, N)
> rightInt = ak.arange(N)
> leftStr = ak.array([f"str {i}" for i in leftInt.to_ndarray()])
> rightStr = ak.array([f"str {i}" for i in rightInt.to_ndarray()])
> ```

The setup seems to be the bottleneck here, i think changing it to something like this makes the benchmarking a bit more doable
```python
N = pytest.prob_size
leftStr = ak.cast(ak.randint(0, 2**16, N), str)
rightStr = ak.cast(ak.arange(N), str)
```

arkouda/join.py
```diff
@@ -241,10 +273,40 @@ def inner_join(
...
+            if t == Categorical.objtype and isinstance(cats, Strings):
+                # Find the corresponding codes value for each term in the whereargs
+                repMsg = generic_msg(
+                    cmd="codeMapping",
+                    args={
+                        "categories": cats,
+                        "query": whereargs[0][keep],  # type: ignore
+                    },
+                )
+                lCodes = create_pdarray(repMsg)
```
pierce314159, line 285: Running the same example i put in my original review gives me a different error on this call
```python
>>> ak.join.inner_join(str_left, str_right, wherefunc = where_func, whereargs=(int_left,int_right))
RuntimeError: Error: Unhandled SymbolEntryType PrimitiveTypedArraySymEntry
```

arkouda/join.py
```diff
@@ -205,6 +210,33 @@ def inner_join(
...
+        if isinstance(left, Categorical) and isinstance(right, Categorical):
+            cats = left.categories[left.codes]  # Get standardized categories without NAvalue
+            leftKeepCodes = array(
+                [x is not True for x in left.isna().to_ndarray()]
+            )  # Keep codes that aren't NAvalue
+            rightKeepCodes = array(
+                [x is not True for x in right.isna().to_ndarray()]
+            )  # Keep codes that aren't NAvalue
+            left = left.codes[leftKeepCodes]
+            right = right.codes[rightKeepCodes]
```
pierce314159, line 238: since we call `to_ndarray` it's still possible to exceed max_transfer bytes. In this example I'm artificially lower the max, but this could definitely happen for a reasonably sized strings array if every element is unique
```python
In [3]: int_left = ak.arange(100)
   ...: int_right = ak.arange(100)
   ...: str_left = ak.array([f'str {i}' for i in int_left.to_list()])
   ...: str_right = ak.array([f'str {i}' for i in int_right.to_list()])

In [4]: ak.client.maxTransferBytes = 10

In [5]: sl, sr = ak.join.inner_join(str_left, str_right)
Cell In[5], line 1
----> 1 sl, sr = ak.join.inner_join(str_left, str_right)

RuntimeError: Array exceeds allowed size for transfer. Increase client.maxTransferBytes to allow
```

pierce314159 (changes requested):

joshmarshall1:
> > Performance testing results using new implementation:
> > 10**4 previously ran on average in ~1,270 seconds (21 minutes)
> > `python -m pytest -c benchmark.ini --size=10**4 -k join_benchmark`
> 
> I'd be curious to see these timings with gasnet enabled!
> 
> > ```python
> > N = pytest.prob_size
> > leftInt = ak.randint(0, 2**16, N)
> > rightInt = ak.arange(N)
> > leftStr = ak.array([f"str {i}" for i in leftInt.to_ndarray()])
> > rightStr = ak.array([f"str {i}" for i in rightInt.to_ndarray()])
> > ```
> 
> The setup seems to be the bottleneck here, i think changing it to something like this makes the benchmarking a bit more doable
> 
> ```python
> N = pytest.prob_size
> leftStr = ak.cast(ak.randint(0, 2**16, N), str)
> rightStr = ak.cast(ak.arange(N), str)
> ```

The setup doesn't effect the timings. The only thing that's timed is this piece
```python
    left, right = benchmark.pedantic(
        ak.join.inner_join,
        args=(leftStr, rightStr, join_helper, (leftStr, rightStr)),
        rounds=pytest.trials,
    )
```

arkouda/join.py
joshmarshall1: Fixed

arkouda/join.py
joshmarshall1: Replaced the list interpretation doing `to_ndarray()`

src/JoinEqWithDTMsg.chpl
```diff
@@ -376,6 +377,43 @@ module JoinEqWithDTMsg
...
+
+        var (catOff, catVal) = catSym[0..#catSize];
+        var (queryOff, queryVal) = querySym[0..#querySize];
+        
+        // For all indexes in the query segString, find the matching index in the categories segString
+        // Exponential Growth - break once a match is found
+        coforall i in 0..#querySize {
+            for x in 0..#catSize {
+                if queryVal[i] == catVal[x]{
+                    codes[i] = x;
```
pierce314159, line 401: might need @Ethan-DeBandi99's input on the `coforall` since he's got more experience. My knee jerk reaction is this should be a `forall` but i don't know for sure

I'm not sure if it would be more efficient but one idea of a more vectorized way to do the inner loop is something like the below. This is just pseudocode i haven't tried running it
```chapel
forall (val, code) in zip(queryVal, codes) {
  // create bool array of locations where we match the queryval
  var equalsVal = val == catVal;
  // find the first place we match
  var (maxVal, maxLoc) = maxloc reduce zip(equalsVal, equalsVal.domain);
  if maxVal {
    code = maxLoc;
  }
}
```

arkouda/join.py
```diff
@@ -205,6 +210,29 @@ def inner_join(
...
+
+    if isinstance(left, Categorical) and isinstance(right, Categorical):
+        t = cast(str, Categorical.objtype)
+        left, right = Categorical.standardize_categories([left, right])
+        if isinstance(left, Categorical) and isinstance(right, Categorical):
+            cats = left.categories[left.codes]  # Get standardized categories without NAvalue
+            leftKeepCodes = left.isna() == 0  # Invert bool array
+            rightKeepCodes = right.isna() == 0  # Invert bool array
+            left = left.codes[leftKeepCodes]
+            right = right.codes[rightKeepCodes]
```
pierce314159, line 234: I'm curious how this compares to something like:
```python
left, right = Categorical.standardize_categories([left, right])
# maybe isinstance check for mypy
left.reset_categorizes()
right.reset_categorizes()
```

joshmarshall1: Some tests using `10**6` problem size showed this to be slower than what's currently there.
Current method:
```
-------------------------------------- benchmark 'Inner Join': 1 tests --------------------------------------
Name (time in s)         Min      Max     Mean  StdDev   Median     IQR  Outliers     OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------
bench_inner_join     42.1997  44.3982  42.7200  0.9449  42.2967  0.7485       1;1  0.0234       5           1
-------------------------------------------------------------------------------------------------------------
```

Suggested Method:
```
-------------------------------------- benchmark 'Inner Join': 1 tests --------------------------------------
Name (time in s)         Min      Max     Mean  StdDev   Median     IQR  Outliers     OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------
bench_inner_join     47.6216  49.4303  48.4438  0.7423  48.4373  1.2332       2;0  0.0206       5           1
-------------------------------------------------------------------------------------------------------------
```

src/JoinEqWithDTMsg.chpl
pierce314159: the `forall` is actually the outerloop (replacing the `coforall`). The inner loop/ for loop is being replaced by the `==` and `maxloc reduce`

Ethan-DeBandi99: The coforall here won't actually do much because you are iterating over a local object which won't allow for any parallelism. So this is effectively fully serial, breaking into a single `forall` will at least allow for task parallel operations.

arkouda/join.py
```diff
@@ -205,6 +210,29 @@ def inner_join(
...
+
+    # Convert Strings and Categorical left & right values to Categorical.codes. This generates an
+    # index representation of the strings. This way they can be handled as numerics and follow the
+    # same process as int
+    t = cast(str, pdarray.objtype)
+    cats = None
+
+    if isinstance(left, Strings) and isinstance(right, Strings):
+        left = Categorical(left)
+        right = Categorical(right)
```
Ethan-DeBandi99, line 224: This will result in more than double the data within strings being generated. With something that is already memory bound this is a bit concerning.

Ethan-DeBandi99 (changes requested):
After look through this, I think there are some core structure things that we need to address before moving forward. Ultimately, I think there is enough of a scaling concern here that we need to address that before this is ready to merge. I know Pierce mentioned some existing Arkouda functionality that may be equivalent and it was noted that it performs worse. Based on the numbers I am seeing with that, I am not sure the addition of code is warranted over using the existing and looking into improving the performance of that. I think some things can be restructured to reduce code complexity and improve performance.

src/JoinEqWithDTMsg.chpl
joshmarshall1: That was my fault for misreading it. 
`10**6` with a `forall` outer-loop
```
-------------------------------------- benchmark 'Inner Join': 1 tests --------------------------------------
Name (time in s)         Min      Max     Mean  StdDev   Median     IQR  Outliers     OPS  Rounds  Iterations
-------------------------------------------------------------------------------------------------------------
bench_inner_join     16.8480  17.4797  17.1619  0.2655  17.1079  0.4548       2;0  0.0583       5           1
-------------------------------------------------------------------------------------------------------------
```

arkouda/join.py
joshmarshall1: This is something I'm aware of and intend to handle through #2390 by moving a lot of the setup workflow into Chapel where I can more easily operate on the `uint8` versions of the segString. This should improve performance a lot, but will take a lot longer to implement

joshmarshall1: After touching base with @Ethan-DeBandi99 we decided to convert this PR to a draft for now and I'll go through the code with the intent to improve performance as much as possible instead of the initial plan of focusing on function here and performance in #2390. Once complete we'll look at the performance numbers again and close #2390 with this PR as well if we're happy with where they are

joshmarshall1:
After touching base with @Ethan-DeBandi99 we decided to convert this PR to a draft for now and I'll go through the code with the intent to improve performance as much as possible instead of the initial plan of focusing on function here and performance in https://github.com/Bears-R-Us/arkouda/issues/2390. Once complete we'll look at the performance numbers again and close https://github.com/Bears-R-Us/arkouda/issues/2390 with this PR as well if we're happy with where they are

joshmarshall1:
Current timings with `10**6` problem size after switching workflow to use align:
```
------------------------------------------------------------------------------- benchmark 'Inner Join': 3 tests --------------------------------------------------------------------------------
Name (time in s)                      Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_inner_join[int]              2.6168 (1.0)       2.8741 (1.0)       2.7232 (1.0)      0.1004 (1.0)       2.6900 (1.0)      0.1399 (1.0)           2;0  0.3672 (1.0)           5           1
bench_inner_join[str]              8.5680 (3.27)      9.7994 (3.41)      9.2853 (3.41)     0.4454 (4.43)      9.3469 (3.47)     0.3424 (2.45)          2;1  0.1077 (0.29)          5           1
bench_inner_join[categorical]     17.0942 (6.53)     18.6592 (6.49)     17.8937 (6.57)     0.7301 (7.27)     18.0534 (6.71)     1.3827 (9.88)          2;0  0.0559 (0.15)          5           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Categorical in particular is still taking longer than I would like, so I'm going to continue experimenting with alternate approaches

joshmarshall1:
Timings using @pierce314159's broadcast implementation and `10**6` problem size. The `[categorical-int-str]` in each test represents `[left/right type - left where type - right where type]`

The difference in timings between some of the tests is odd, one example being `[int-categorical-int]` and `[int-int-categorical]`. The workflow for the `left where` is faster by ~50% than the `right where`

```
------------------------------------------------------------------------------------------- benchmark 'Inner Join': 27 tests -------------------------------------------------------------------------------------------
Name (time in s)                                              Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_inner_join[int-int-int]                              2.6385 (1.0)       2.6614 (1.0)       2.6471 (1.0)      0.0125 (1.0)       2.6414 (1.0)      0.0172 (1.0)           1;0  0.3778 (1.0)           3           1
bench_inner_join[str-int-int]                              3.0094 (1.14)      3.1513 (1.18)      3.0706 (1.16)     0.0729 (5.85)      3.0511 (1.16)     0.1064 (6.20)          1;0  0.3257 (0.86)          3           1
bench_inner_join[int-str-int]                              3.0492 (1.16)      3.1351 (1.18)      3.0819 (1.16)     0.0465 (3.72)      3.0614 (1.16)     0.0644 (3.75)          1;0  0.3245 (0.86)          3           1
bench_inner_join[str-str-int]                              3.3485 (1.27)      3.5057 (1.32)      3.4141 (1.29)     0.0818 (6.55)      3.3881 (1.28)     0.1179 (6.87)          1;0  0.2929 (0.78)          3           1
bench_inner_join[int-categorical-int]                      3.3883 (1.28)      3.5746 (1.34)      3.4523 (1.30)     0.1060 (8.49)      3.3940 (1.28)     0.1397 (8.14)          1;0  0.2897 (0.77)          3           1
bench_inner_join[str-categorical-int]                      3.5518 (1.35)      3.7463 (1.41)      3.6548 (1.38)     0.0977 (7.83)      3.6662 (1.39)     0.1458 (8.49)          1;0  0.2736 (0.72)          3           1
bench_inner_join[categorical-int-int]                      3.8983 (1.48)      4.0591 (1.53)      3.9699 (1.50)     0.0818 (6.56)      3.9525 (1.50)     0.1206 (7.02)          1;0  0.2519 (0.67)          3           1
bench_inner_join[categorical-str-int]                      4.3263 (1.64)      4.4650 (1.68)      4.3744 (1.65)     0.0785 (6.29)      4.3318 (1.64)     0.1040 (6.06)          1;0  0.2286 (0.61)          3           1
bench_inner_join[categorical-categorical-int]              4.5968 (1.74)      4.6794 (1.76)      4.6430 (1.75)     0.0422 (3.38)      4.6528 (1.76)     0.0620 (3.61)          1;0  0.2154 (0.57)          3           1
bench_inner_join[int-int-str]                              7.6869 (2.91)      7.8451 (2.95)      7.7532 (2.93)     0.0822 (6.59)      7.7276 (2.93)     0.1187 (6.91)          1;0  0.1290 (0.34)          3           1
bench_inner_join[str-int-str]                              8.0939 (3.07)      8.4777 (3.19)      8.3015 (3.14)     0.1938 (15.53)     8.3330 (3.15)     0.2879 (16.76)         1;0  0.1205 (0.32)          3           1
bench_inner_join[int-categorical-str]                      8.1072 (3.07)      8.6779 (3.26)      8.3395 (3.15)     0.2998 (24.03)     8.2333 (3.12)     0.4280 (24.93)         1;0  0.1199 (0.32)          3           1
bench_inner_join[str-str-str]                              8.1751 (3.10)      8.5690 (3.22)      8.3842 (3.17)     0.1981 (15.88)     8.4086 (3.18)     0.2954 (17.21)         1;0  0.1193 (0.32)          3           1
bench_inner_join[str-categorical-str]                      8.4908 (3.22)      8.9430 (3.36)      8.7516 (3.31)     0.2339 (18.75)     8.8208 (3.34)     0.3392 (19.75)         1;0  0.1143 (0.30)          3           1
bench_inner_join[int-str-str]                              8.6340 (3.27)      8.9713 (3.37)      8.7561 (3.31)     0.1869 (14.98)     8.6629 (3.28)     0.2529 (14.73)         1;0  0.1142 (0.30)          3           1
bench_inner_join[categorical-int-str]                      8.9051 (3.38)      9.4540 (3.55)      9.1969 (3.47)     0.2761 (22.13)     9.2317 (3.50)     0.4117 (23.98)         1;0  0.1087 (0.29)          3           1
bench_inner_join[categorical-str-str]                      9.3883 (3.56)      9.6370 (3.62)      9.5327 (3.60)     0.1291 (10.35)     9.5729 (3.62)     0.1865 (10.86)         1;0  0.1049 (0.28)          3           1
bench_inner_join[categorical-categorical-str]              9.6535 (3.66)     10.3699 (3.90)     10.0223 (3.79)     0.3587 (28.75)    10.0435 (3.80)     0.5373 (31.29)         1;0  0.0998 (0.26)          3           1
bench_inner_join[int-int-categorical]                     15.1784 (5.75)     15.3644 (5.77)     15.2496 (5.76)     0.1004 (8.04)     15.2059 (5.76)     0.1395 (8.12)          1;0  0.0656 (0.17)          3           1
bench_inner_join[str-int-categorical]                     15.2229 (5.77)     15.7787 (5.93)     15.4582 (5.84)     0.2876 (23.05)    15.3730 (5.82)     0.4169 (24.28)         1;0  0.0647 (0.17)          3           1
bench_inner_join[int-str-categorical]                     15.8482 (6.01)     16.5925 (6.23)     16.1530 (6.10)     0.3900 (31.26)    16.0184 (6.06)     0.5582 (32.51)         1;0  0.0619 (0.16)          3           1
bench_inner_join[int-categorical-categorical]             16.0382 (6.08)     16.1755 (6.08)     16.1117 (6.09)     0.0691 (5.54)     16.1214 (6.10)     0.1029 (5.99)          1;0  0.0621 (0.16)          3           1
bench_inner_join[str-categorical-categorical]             16.1162 (6.11)     18.1617 (6.82)     17.2566 (6.52)     1.0428 (83.58)    17.4920 (6.62)     1.5341 (89.35)         1;0  0.0579 (0.15)          3           1
bench_inner_join[str-str-categorical]                     16.2922 (6.17)     16.4657 (6.19)     16.3781 (6.19)     0.0868 (6.96)     16.3763 (6.20)     0.1301 (7.58)          1;0  0.0611 (0.16)          3           1
bench_inner_join[categorical-int-categorical]             16.3565 (6.20)     17.2736 (6.49)     16.7321 (6.32)     0.4805 (38.51)    16.5662 (6.27)     0.6878 (40.06)         1;0  0.0598 (0.16)          3           1
bench_inner_join[categorical-str-categorical]             16.6065 (6.29)     16.9145 (6.36)     16.7154 (6.31)     0.1727 (13.84)    16.6253 (6.29)     0.2310 (13.45)         1;0  0.0598 (0.16)          3           1
bench_inner_join[categorical-categorical-categorical]     17.7366 (6.72)     17.8603 (6.71)     17.7892 (6.72)     0.0639 (5.12)     17.7706 (6.73)     0.0928 (5.41)          1;0  0.0562 (0.15)          3           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

joshmarshall1:
Latest commit's performance metrics
```
------------------------------------------------------------------------------------- benchmark 'Inner Join': 30 tests -------------------------------------------------------------------------------------
Name (time in s)                                  Min                Max               Mean            StdDev             Median               IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_inner_join[int]                          2.0490 (1.0)       2.7882 (1.0)       2.2965 (1.0)      0.4258 (48.31)     2.0523 (1.0)      0.5544 (41.93)         1;0  0.4354 (1.0)           3           1
bench_inner_join_where[i-int-integer]          2.7195 (1.33)      2.8458 (1.02)      2.7683 (1.21)     0.0679 (7.70)      2.7397 (1.33)     0.0948 (7.17)          1;0  0.3612 (0.83)          3           1
bench_inner_join_where[s-int-integer]          2.9650 (1.45)      3.1074 (1.11)      3.0372 (1.32)     0.0712 (8.08)      3.0392 (1.48)     0.1068 (8.08)          1;0  0.3293 (0.76)          3           1
bench_inner_join_where[i-cat-integer]          3.1006 (1.51)      3.2181 (1.15)      3.1763 (1.38)     0.0657 (7.45)      3.2102 (1.56)     0.0881 (6.67)          1;0  0.3148 (0.72)          3           1
bench_inner_join_where[s-cat-integer]          3.3805 (1.65)      3.5756 (1.28)      3.4606 (1.51)     0.1021 (11.58)     3.4259 (1.67)     0.1463 (11.07)         1;0  0.2890 (0.66)          3           1
bench_inner_join_where[i-str-integer]          3.7535 (1.83)      4.1236 (1.48)      3.8972 (1.70)     0.1984 (22.51)     3.8146 (1.86)     0.2776 (20.99)         1;0  0.2566 (0.59)          3           1
bench_inner_join_where[s-str-integer]          3.8985 (1.90)      4.0001 (1.43)      3.9371 (1.71)     0.0550 (6.24)      3.9127 (1.91)     0.0762 (5.76)          1;0  0.2540 (0.58)          3           1
bench_inner_join_where[c-int-integer]          4.1418 (2.02)      4.3768 (1.57)      4.2358 (1.84)     0.1244 (14.11)     4.1888 (2.04)     0.1763 (13.33)         1;0  0.2361 (0.54)          3           1
bench_inner_join_where[c-cat-integer]          4.2407 (2.07)      4.4873 (1.61)      4.3398 (1.89)     0.1302 (14.77)     4.2913 (2.09)     0.1849 (13.98)         1;0  0.2304 (0.53)          3           1
bench_inner_join_where[c-str-integer]          4.8005 (2.34)      4.8181 (1.73)      4.8093 (2.09)     0.0088 (1.0)       4.8092 (2.34)     0.0132 (1.0)           1;0  0.2079 (0.48)          3           1
bench_inner_join[str]                          6.3471 (3.10)      6.7275 (2.41)      6.5563 (2.85)     0.1930 (21.89)     6.5941 (3.21)     0.2853 (21.58)         1;0  0.1525 (0.35)          3           1
bench_inner_join_where[i-int-string]           7.5725 (3.70)      8.1338 (2.92)      7.7826 (3.39)     0.3060 (34.72)     7.6417 (3.72)     0.4210 (31.84)         1;0  0.1285 (0.30)          3           1
bench_inner_join_where[s-int-string]           7.9245 (3.87)      8.1549 (2.92)      8.0287 (3.50)     0.1167 (13.24)     8.0068 (3.90)     0.1728 (13.07)         1;0  0.1246 (0.29)          3           1
bench_inner_join_where[i-cat-string]           7.9487 (3.88)      8.2080 (2.94)      8.0992 (3.53)     0.1346 (15.27)     8.1410 (3.97)     0.1945 (14.71)         1;0  0.1235 (0.28)          3           1
bench_inner_join_where[s-cat-string]           8.2842 (4.04)      8.8261 (3.17)      8.4777 (3.69)     0.3023 (34.30)     8.3229 (4.06)     0.4064 (30.74)         1;0  0.1180 (0.27)          3           1
bench_inner_join_where[i-str-string]           8.4600 (4.13)      8.5150 (3.05)      8.4919 (3.70)     0.0286 (3.24)      8.5008 (4.14)     0.0413 (3.12)          1;0  0.1178 (0.27)          3           1
bench_inner_join_where[s-str-string]           8.5977 (4.20)      9.0127 (3.23)      8.8743 (3.86)     0.2396 (27.18)     9.0125 (4.39)     0.3113 (23.54)         1;0  0.1127 (0.26)          3           1
bench_inner_join_where[c-cat-string]           9.0788 (4.43)      9.2807 (3.33)      9.1782 (4.00)     0.1010 (11.46)     9.1751 (4.47)     0.1515 (11.45)         1;0  0.1090 (0.25)          3           1
bench_inner_join_where[c-int-string]           9.0972 (4.44)      9.6202 (3.45)      9.2750 (4.04)     0.2990 (33.92)     9.1075 (4.44)     0.3923 (29.67)         1;0  0.1078 (0.25)          3           1
bench_inner_join_where[c-str-string]           9.6685 (4.72)     10.1043 (3.62)      9.9281 (4.32)     0.2296 (26.04)    10.0117 (4.88)     0.3268 (24.72)         1;0  0.1007 (0.23)          3           1
bench_inner_join[categorical]                 10.5764 (5.16)     11.1818 (4.01)     10.8430 (4.72)     0.3091 (35.06)    10.7707 (5.25)     0.4540 (34.33)         1;0  0.0922 (0.21)          3           1
bench_inner_join_where[i-int-categorical]     11.5400 (5.63)     11.6498 (4.18)     11.5810 (5.04)     0.0599 (6.80)     11.5533 (5.63)     0.0823 (6.23)          1;0  0.0863 (0.20)          3           1
bench_inner_join_where[s-int-categorical]     11.5993 (5.66)     12.2175 (4.38)     11.8599 (5.16)     0.3203 (36.34)    11.7628 (5.73)     0.4636 (35.06)         1;0  0.0843 (0.19)          3           1
bench_inner_join_where[i-cat-categorical]     12.0209 (5.87)     12.0495 (4.32)     12.0371 (5.24)     0.0146 (1.66)     12.0408 (5.87)     0.0214 (1.62)          1;0  0.0831 (0.19)          3           1
bench_inner_join_where[s-cat-categorical]     12.5090 (6.10)     12.9345 (4.64)     12.7254 (5.54)     0.2128 (24.14)    12.7326 (6.20)     0.3191 (24.13)         1;0  0.0786 (0.18)          3           1
bench_inner_join_where[s-str-categorical]     12.5234 (6.11)     12.7153 (4.56)     12.6336 (5.50)     0.0990 (11.23)    12.6620 (6.17)     0.1439 (10.88)         1;0  0.0792 (0.18)          3           1
bench_inner_join_where[i-str-categorical]     12.5865 (6.14)     13.0348 (4.67)     12.7737 (5.56)     0.2331 (26.45)    12.6997 (6.19)     0.3362 (25.43)         1;0  0.0783 (0.18)          3           1
bench_inner_join_where[c-int-categorical]     12.7796 (6.24)     13.6261 (4.89)     13.0739 (5.69)     0.4786 (54.29)    12.8159 (6.24)     0.6349 (48.01)         1;0  0.0765 (0.18)          3           1
bench_inner_join_where[c-cat-categorical]     13.0421 (6.37)     13.8255 (4.96)     13.4258 (5.85)     0.3920 (44.47)    13.4099 (6.53)     0.5876 (44.44)         1;0  0.0745 (0.17)          3           1
bench_inner_join_where[c-str-categorical]     13.8884 (6.78)     14.1880 (5.09)     13.9946 (6.09)     0.1677 (19.02)    13.9075 (6.78)     0.2247 (16.99)         1;0  0.0715 (0.16)          3           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

src/Broadcast.chpl
```diff
@@ -135,4 +145,42 @@ module Broadcast {
...
+    var expandedVals = makeDistArray(strSize, uint(8));
+    var agg = newDstAggregator(uint(8));
+    ref vals = segString.values.a;
+
+    forall (i, o, s) in zip(ind, offsets, 0..#size) with (ref agg, ref vals) {
+      var inds = if i == high then segOff[i]..segString.nBytes-1 else segOff[i]..segOff[i+1]-1;
+      var offs = if s == size - 1 then o..strSize-1 else o..offsets[s+1]-1;
+      forall (off, idx) in zip(offs, inds) with (ref agg, ref vals) {
+        agg.copy(expandedVals[off], vals[idx]);
+      }
```
pierce314159, line 181: im a bit worried it might be possible to have remote to remote aggregation here. I pulled down the PR and ran a string broadcast with gasnet and 2 locales. I'm not sure if this is the cause, but I did get a server crash with a `SIGSEGV(11)`
```python
>>> s = ak.array(["abc","abcde","ab"])
>>> segs = ak.array([0,2,5])
>>> ak.groupbyclass.broadcast(segs, s, 8)
```
from the logs
```
*** Caught a fatal signal (proc 1): SIGSEGV(11)
*** NOTICE (proc 1): Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace.
*** NOTICE (proc 1): We recommend linking the debug version of GASNet to assist you in resolving this application issue.
*** Caught a fatal signal (proc 1): SIGSEGV(11)
````

**Review by pierce314159 — CHANGES_REQUESTED** [review]

File: src/Broadcast.chpl (hunk 11 above)
Comment (joshmarshall1) [reply src/Broadcast.chpl]: The `agg` was the cause of this, saying the operation hit the memory limit. I reverted this to use the original bulk copy method, which shows equivalent times at `10**6` problem size. I had no problem running the bulk copy version with multiple locales

File: src/Broadcast.chpl [hunk 12]
```diff
@@ -135,4 +145,39 @@ module Broadcast {
...
+        strSize += (segs[i+1] - seg) * (segOff[i+1] - off);
+        d = segOff[i+1] - off;
+      }
+    }
+
+    var broadDist = broadcast(segs, diffs, size);
+    var offDiff = makeDistArray(size, int);
+    offDiff[1..] = broadDist[..size-2];
+
+    var offsets = (+ scan offDiff);
```
pierce314159, line 169: does this just end up being an exclusive scan? i.e. equivalent to
```chapel
var broadDist = broadcast(segs, diffs, size);
var offsets = (+ scan broadDist) - broadDist;
```

src/Broadcast.chpl
```diff
@@ -135,4 +145,39 @@ module Broadcast {
...
+    var r: [sD] int = 0..segString.size; 
+    var ind = broadcast(segs, r, size);
+    var expandedVals: [0..#strSize] uint(8);
```
pierce314159, line 172: should this be a `makeDistArray(strSize, uint(8));`

src/Broadcast.chpl
joshmarshall1: Yes, good catch!

src/Broadcast.chpl
joshmarshall1: Indeed. Fixed

arkouda/join.py
```diff
@@ -244,7 +250,11 @@ def inner_join(
             # Gather right whereargs
             rightWhere = whereargs[1][byRight.permutation][ranges]
             # Expand left whereargs
-            leftWhere = broadcast(fullSegs, whereargs[0][keep], ranges.size)
+            keep_where = whereargs[0][keep]
+            if isinstance(keep_where, Categorical):
+                leftWhere = broadcast(fullSegs, keep_where.codes, ranges.size)
+            else:
+                leftWhere = broadcast(fullSegs, keep_where, ranges.size)
```
pierce314159, line 257: could we make it so `broadcast` accepts cats too and just operates on the `codes`? That way we don't have to change this at all? I def feel like broadcasting on categoricals is useful behavior

src/BroadcastMsg.chpl
```diff
@@ -28,6 +29,12 @@ module BroadcastMsg {
                                        getModuleName(),
                                        "TypeError");
     }
+
+    const objType = msgArgs.getValueOf("objType");
+    if objType != "pdarray" {
+      return broadcastStrings(cmd, msgArgs, st);
+    }
```
pierce314159, line 36: u should the `ObjType` enum instead of doing string comparisons, something like:
```chapel
if objType.toUpper(): ObjType == ObjType.STRINGS {
    return broadcastStrings(cmd, msgArgs, st);
}
```

src/Broadcast.chpl
```diff
@@ -135,4 +145,36 @@ module Broadcast {
...
+    var r: [sD] int = 0..segString.size; 
+    var ind = broadcast(segs, r, size);
+    var expandedVals = makeDistArray(strSize, uint(8));
+    ref vals = segString.values.a;
+
+    forall (i, o, s) in zip(ind, offsets, 0..#size) with (ref vals) {
+      var inds = if i == high then segOff[i]..segString.nBytes-1 else segOff[i]..segOff[i+1]-1;
+      var offs = if s == size - 1 then o..strSize-1 else o..offsets[s+1]-1;
+      expandedVals[offs] = vals[inds];
+    }
```
pierce314159, line 176: im curious if we can make this a bit more similar to other strings workflows and take a localizing slice here and maybe do some aggregation. I think the advantage this buys us is to ensure the `vals` are local

Right now we calculate the ranges for `inds` and `offs`, but they both have to be the same length. So we really just need the starts of each and how long the str is. I think something like this might give us better perf. (warning this is psuedocode, i've not run it :)
```chapel
const lengths = segString.getLengths();
forall (so, o, len) in zip(segOff, offsets, lengths) with (var valAgg = newDstAggregator(uint(8))) {
  var localizedVals = new lowLevelLocalizingSlice(vals, so..#len);
  for k in 0..#len {
    valAgg.copy(expandedVals[o+k], localizedVals.ptr[k]);
  }
}
```

im not sure this is gonna be quicker or not, we should do some comm diags to compare. I'm also not sure if it's within the scope of this PR or not, but def wanted to make a note of it before i forgot

pierce314159 (approved):
A couple things to consider but nothing major is jumping out at me!

src/Broadcast.chpl
pierce314159: so i ran just some real quick perf numbers (super small scale like 100 long arrays) and i am seeing some lowered comm using the above
```python
In [15]: a = ak.arange(10)

In [16]: segs = ak.arange(0,100,10)

In [17]: a
Out[17]: array([0 1 2 3 4 5 6 7 8 9])

In [18]: segs
Out[18]: array([0 10 20 30 40 50 60 70 80 90])

In [19]: ak.broadcast(segs, a, 100)
Out[19]: array([0 0 0 ... 9 9 9])

In [20]: s = ak.cast(a, str)

In [21]: s
Out[21]: array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'])

In [22]: ak.broadcast(segs, s, 100)
Out[22]: array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '4', '4', '4', '4', '4', '4', '4', '4', '4', '4', '5', '5', '5', '5', '5', '5', '5', '5', '5', '5', '6', '6', '6', '6', '6', '6', '6', '6', '6', '6', '7', '7', '7', '7', '7', '7', '7', '7', '7', '7', '8', '8', '8', '8', '8', '8', '8', '8', '8', '8', '9', '9', '9', '9', '9', '9', '9', '9', '9', '9'])
```
and the comm diagnostics for this
```
USING whats there rn

|      0 | 818 | 300 |          0 |           601 |
|      1 | 811 | 100 |        200 |           600 |
USING agg and localizing slices

|      0 |   1 |   5 |          0 |             4 |
|      1 |  32 |   6 |         10 |             0 |
```

arkouda/groupbyclass.py
```diff
@@ -2092,14 +2101,21 @@ def broadcast(
...
+                "permute": permute,
+                "size": size,
+            },
+        ),
     )
-    return create_pdarray(repMsg)
+    if values.objType == Strings.objType:
+        return Strings.from_return_msg(repMsg)
+    else:
+        return create_pdarray(repMsg)
```
Ethan-DeBandi99, line 2121: Based on this it looks like the `values` parameter can be `pdarray` or `Strings`. However, the function definition of `values: pdarray` will not allow this.

src/Broadcast.chpl
pierce314159: posted a comment about comm, but mighta spoken too early the impl above is only writing the first couple vals, looking into it more

src/BroadcastMsg.chpl
Ethan-DeBandi99: Definitely agree here

Ethan-DeBandi99 (changes requested):

src/Broadcast.chpl
pierce314159: Digging a bit more, there's def perf improvements to be gained from an approach like this

I had a `10**4` long strings and ran both methods and recorded the comm and time:
```python
In [3]: def test(i):
   ...:     TOT = 10**i
   ...:     SIZE = 10**(i-1)
   ...:     a = ak.arange(SIZE)
   ...:     segs = ak.arange(0,TOT,10)
   ...:     s = ak.cast(a, str)
   ...:     rand = ak.random_strings_uniform(1, 16, SIZE)
   ...:     s = s.stick(rand, delimiter="_")
   ...:     return ak.broadcast(segs, s, TOT)
   ...: 

In [4]: test(5)
Out[4]: array(['0_WXVTUR', '0_WXVTUR', '0_WXVTUR', ... , '9999_OZH', '9999_OZH', '9999_OZH'])
```
this is the results:
```
new way time = 0

|      0 |  17 |   1 |          0 |            41 |
|      1 | 177 |  13 |          5 |             0 |
old way time = 115

| locale |    get |    put | execute_on | execute_on_nb |
| -----: | -----: | -----: | ---------: | ------------: |
|      0 | 800064 | 300000 |          0 |        600063 |
|      1 | 800203 | 100017 |     200004 |        600378 |
```

the code i suggest using is:
```chapel
proc broadcast(segs: [?sD] int, segString: borrowed SegString, size: int) throws {
  ref offs = segString.offsets.a;
  ref vals = segString.values.a;
  const high = sD.high;
  var strLens: [sD] int;
  var segLens: [sD] int;
  var expandedLen: int;

  forall (i, off, str_len, seg, seg_len) in zip (sD, offs, strLens, segs, segLens) with (+ reduce expandedLen) {
    if i == high {
      seg_len = size - seg;
      str_len = vals.size - off;
    } else {
      seg_len = segs[i+1] - seg;
      str_len = offs[i+1] - off;
    }
    expandedLen += seg_len * str_len;
  }
  var broadDist = broadcast(segs, strLens, size);
  const offsets = (+ scan broadDist) - broadDist;
  var expandedVals = makeDistArray(expandedLen, uint(8));

  forall (off, str_len, seg, seg_len) in zip(offs, strLens, segs, segLens) with (var valAgg = newDstAggregator(uint(8))) {
    var localizedVals = new lowLevelLocalizingSlice(vals, off..#str_len);
    for i in seg..#seg_len {
      var expValOff = offsets[i];
      for k in 0..#str_len {
        valAgg.copy(expandedVals[expValOff+k], localizedVals.ptr[k]);
      }
    }
  }
  return (expandedVals, offsets);
}
```

src/Broadcast.chpl
```diff
@@ -135,4 +145,36 @@ module Broadcast {
     expandedVals = (+ scan expandedVals);
     return (expandedVals == 1);
   }
+
+  proc broadcast(segs: [?sD] int, segString: borrowed SegString, size: int) throws {
+    ref segOff = segString.offsets.a;
+    const high = sD.high;
+    var strSize: int;
```
pierce314159, line 152: while i was looking into this i noticed we are missing the `broadcast` with permutation
 ```
proc broadcast(perm: [?D] int, segs: [?sD] int, segString: borrowed SegString) throws {
```
I feel like we should add that since it's possible for `permName` to be passed with a strings

pierce314159 (changes requested):

Issue #2373, opened 2023-04-12
`Categorical.update_hdf` Functionality


Ethan-DeBandi99:
This work requires #2372 to be completed.

Add the `update_hdf` functionality to categorical to enable overwriting an existing dataset with the Categorical object.

Ethan-DeBandi99:
Closed by PR #2394.

PR #2394, opened 2023-04-25
Closes #2372 & #2373 - Categorical HDF5 Format Update & `update_hdf` method

### LINKED ISSUE #2372: Update Categorical HDF5 Storage Format
Categorical currently saves to hdf5 as multiple datasets with a suffix indicating which property the dataset contains. This should be updated to use a group with the datasets contained within the group. The group should have the user provided name and the dataset names should indicate the property stored by that dataset.

We will need to maintain backwards compatability for reads.

This work is required to enable allowing `update_hdf` with categoricals.

Ethan-DeBandi99:
Closes #2372 & #2373 

- Updates Categorical to save to HDF5 using the following format, which allows us to view it as a single object
  - Group: Name provided by user
    - Group: categories (this stores a SegString as 2 datasets)
    - Dataset: codes
    - Dataset: NA_Codes
    - Dataset: permutation (if present in the object)
    - Dataset: segments (if present in object)
- Maintains read functionality for any categorical written with the previous format.
- Updates `Categorical.from_return_msg` to use a dictionary format rather than a delimited string.
  - This required additional updates to `attach/register` for Categoricals as they were using `from_return_msg`. These were updated to use the new dictionary formatted response as well. I will be adding an issues to update ALL attach/register functionality to this.
- Adds Testing for `Categorical.update_hdf`.
- Modified existing testing for new Categorical Storage format (though these were minor).
- Updates read functionality typing to allow for Categorical return.

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -825,6 +825,74 @@ class Categorical:
             args=args,
         )
 
+    def update_hdf(
+        self,
+        prefix_path,
+        dataset="categorical",
+        repack=True
+    ):
+        """
+        Overwrite the dataset with the name provided with this Categorical object. If
+        the dataset does not exist it is added.
+
+        Parameters
+        -----------
+        prefix_path : str
+            Directory and filename prefix that all output files share
+        dataset : str
+            Name of the dataset to create in files
+        repack: bool
+            Default: True
+            HDF5 does not release memory on delete. When True, the inaccessible
+            data (that was overwritten) is removed. When False, the data remains, but is
+            inaccessible. Setting to false will yield better performance, but will cause
+            file sizes to expand.
+
+        Returns
+        --------
+        None
+
+        Raises
+        -------
+        RuntimeError
+            Raised if a server-side error is thrown saving the SegArray
+
+        Notes
+        ------
+        - If file does not contain File_Format attribute to indicate how it was saved,
+          the file name is checked for _LOCALE#### to determine if it is distributed.
+        - If the dataset provided does not exist, it will be added
+        - Because HDF5 deletes do not release memory, this will create a copy of the
+          file with the new data
+        """
+        from arkouda.io import _get_hdf_filetype, _mode_str_to_int, _file_type_to_int, _repack_hdf
+
+        # determine the format (single/distribute) that the file was saved in
+        file_type = _get_hdf_filetype(prefix_path + "*")
+
+        args = {
+            "codes": self.codes,
+            "categories": self.categories,
+            "dset": dataset,
+            "write_mode": _mode_str_to_int("append"),
+            "filename": prefix_path,
+            "objType": "categorical",
+            "overwrite": True,
+            "file_format": _file_type_to_int(file_type),
+            "NA_codes": self._akNAcode
+        }
+        if self.permutation is not None and self.segments is not None:
+            args["permutation"] = self.permutation
+            args["segments"] = self.segments
+
+        generic_msg(
+            cmd="tohdf",
+            args=args,
+        )
+
+        if repack:
+            _repack_hdf(prefix_path)
+
     def to_parquet(
         self,
         prefix_path: str,
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -770,7 +770,7 @@ class Categorical:
             new_categories = new_arrays[0].categories
             new_codes = cast(pdarray, concatenate([arr.codes for arr in new_arrays], ordered=ordered))
             return Categorical.from_codes(new_codes, new_categories, NAvalue=self.NAvalue)
-    
+
     def to_hdf(
         self,
         prefix_path,
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -774,7 +774,7 @@ class Categorical:
     def to_hdf(
         self,
         prefix_path,
-        dataset="categorical",
+        dataset="categorical_array",
         mode="truncate",
         file_type="distribute",
     ):
@@ -828,7 +828,7 @@ class Categorical:
     def update_hdf(
         self,
         prefix_path,
-        dataset="categorical",
+        dataset="categorical_array",
         repack=True
     ):
         """
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -855,7 +855,7 @@ class Categorical:
         Raises
         -------
         RuntimeError
-            Raised if a server-side error is thrown saving the SegArray
+            Raised if a server-side error is thrown saving the Categorical
 
         Notes
         ------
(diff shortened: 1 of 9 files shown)


arkouda/io.py
```diff
@@ -444,24 +446,43 @@ def _parse_obj(obj: Dict) -> Union[Strings, pdarray, arkouda.array_view.ArrayVie
...
 
 
-def _dict_recombine_segarrays(df_dict):
+def _dict_recombine_segarrays_categoricals(df_dict):
     # this assumes segments will always have corresponding values.
     # This should happen due to save config
     seg_cols = ["_".join(col.split("_")[:-1]) for col in df_dict.keys() if col.endswith("_segments")]
-    df_dict_keys = [
-        "_".join(col.split("_")[:-1]) if col.endswith("_segments") or col.endswith("_values") else col
+    cat_cols = ["_".join(col.split(".")[:-1]) for col in df_dict.keys() if col.endswith(".categories")]
```
joshmarshall1, line 459: I could be wrong but I think this should be 
```python
cat_cols = [".".join(col.split(".")[:-1]) for col in df_dict.keys() if col.endswith(".categories")]
```

joshmarshall1 (commented):
One thing to look at, but otherwise looks good

arkouda/io.py
Ethan-DeBandi99: It should not matter, but I am seeing that if there is another `.` in the name it could cause and issue. I will update that. Thanks!

Ethan-DeBandi99: Updated

joshmarshall1 (approved):

src/RegistrationMsg.chpl
```diff
@@ -64,6 +64,10 @@ module RegistrationMsg
         return msgTuple;
     }
 
+    proc categorical_attachMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws{
+
+    }
+
```
pierce314159, line 70: i don't think is meant to be empty

src/HDF5Msg.chpl
```diff
@@ -951,19 +883,22 @@ module HDF5Msg {
...
         writeArkoudaMetaData(file_id, "%s/%s".format(group, SEGMENTED_VALUE_NAME), "pdarray", dtype_id);
 
-        var sd = segDim:C_HDF5.hsize_t;
-        C_HDF5.H5LTmake_dataset(file_id, "/%s/%s".format(group, SEGMENTED_OFFSET_NAME).c_str(), 1:c_int, sd, getDataType(int), segs.ptr);
-        writeArkoudaMetaData(file_id, "%s/%s".format(group, SEGMENTED_OFFSET_NAME), "pdarray", getDataType(int));
+        if write_offsets {
+            var localSegs: [0..#segs.size] int = segs.a;
+            var sd = segs.size:C_HDF5.hsize_t;
+            C_HDF5.H5LTmake_dataset(file_id, "%s/%s".format(group, SEGMENTED_OFFSET_NAME).c_str(), 1:c_int, sd, getDataType(int), c_ptrTo(localSegs));
+            writeArkoudaMetaData(file_id, "%s/%s".format(group, SEGMENTED_OFFSET_NAME), "pdarray", getDataType(int));
```
pierce314159, line 900: just want to verify that using `SEGMENTED_OFFSET_NAME` and `"pdarray"` won't be an issue in the segstr/categorical case

src/HDF5Msg.chpl
```diff
@@ -1155,6 +1082,129 @@ module HDF5Msg {
...
+        var codes_entry = st.lookup(msgArgs.getValueOf("codes"));
+        var codes = toSymEntry(toGenSymEntry(codes_entry), int);
+        var cat_entry:SegStringSymEntry = toSegStringSymEntry(st.lookup(msgArgs.getValueOf("categories")));
+        var cats = new SegString("", cat_entry);
+        var naCodes_entry = st.lookup(msgArgs.getValueOf("NA_codes"));
+        var naCodes = toSymEntry(toGenSymEntry(naCodes_entry), int);
+        var perm_seg_exist: bool = false;
+        if msgArgs.contains("permutation") && msgArgs.contains("segments") {
+            perm_seg_exist = true;
+        }
```
pierce314159, line 1108: same question of it's possible to only have one or the other

pierce314159 (commented):
A few things to look at

arkouda/io.py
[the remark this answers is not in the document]
pierce314159: Already addressed in josh's review

arkouda/categorical.py
```diff
@@ -189,6 +191,29 @@ def from_codes(
...
+        eles = json.loads(rep_msg)
+        codes = create_pdarray(eles["codes"])
+        cats = Strings.from_return_msg(eles["categories"])
+        na_code = create_pdarray(eles["_akNAcode"])
+
+        segments = None
+        perm = None
+        if "segments" in eles and "permutation" in eles:
+            segments = create_pdarray(eles["segments"])
+            perm = create_pdarray(eles["permutation"])
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Based on the `init` method these two would be present or not. They are both coming from the GroupBy, so I think this needs to remain as is because we should have both or neither.

arkouda/categorical.py
```diff
@@ -802,77 +773,125 @@ def concatenate(self, others: Sequence[Categorical], ordered: bool = True) -> Ca
...
-        determine the file format.
-        See Also
-        ---------
-        to_parquet
+        ------
+        - If file does not contain File_Format attribute to indicate how it was saved,
+          the file name is checked for _LOCALE#### to determine if it is distributed.
+        - If the dataset provided does not exist, it will be added
+        - Because HDF5 deletes do not release memory, this will create a copy of the
+          file with the new data
```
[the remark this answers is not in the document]
Ethan-DeBandi99: This was just to reiterate so that its less likely to be missed. I can reiterate that it does only if repack is set

arkouda/io.py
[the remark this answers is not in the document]
Ethan-DeBandi99: Already updated.

src/RegistrationMsg.chpl
Ethan-DeBandi99: Oh man... That was left over from my initial thought that I would need a separate message. I forget the exact reason. It should actually be removed.

src/HDF5Msg.chpl
Ethan-DeBandi99: It will not this allows the dataset to be read as a standalone pdarray. This was a feature requested by users.

src/HDF5Msg.chpl
Ethan-DeBandi99: Again, based on the init we should have both or neither so I maintained that idea through

pierce314159 (approved):
Looks good!

PR #2397, opened 2023-04-26
rough draft of diff approach to inner_join

pierce314159:
This is a rough draft for comparison purposes of an alternative approach to inner_join on str/cat. Creating this draft PR for ease of sharing


pierce314159:
I did verify that this passes my reproducers from the other PR. I also made sure it worked with `int, str, cat` as the `left right` arguments with any of `int, str, cat` as the where arguments

pierce314159:
This approach is now in #2376, closing in favor of that

PR #2401, opened 2023-04-26
Closes #2398 - Fixes Parquet List Columns with Nested Lists Error

### LINKED ISSUE #2398: Parquet List Columns with Nested Lists Error
Arkouda does not have the means to read multi-dimensional arrays. SegArrays currently store as lists of numeric/bool data types. However, through testing for issue #2396 it was discovered that when a Parquet list field contains another List field, an `Unsupported Data Type` error is produced. In these cases, we should not identify these columns as valid for Arkouda to read and simply skip over them.

Ethan-DeBandi99:
Closes #2398 

Updates the check check of the datatype contained within a list to return `ArrowTypes.notimplemented` if it is not a supported value. This allows any dataset that is not a valid Arkouda data type to be ignored. A log message is generated stating that the dataset was skipped. This still maintains read functionality for SegArrays. Code used to test this is below.

```python
f = pd.DataFrame({
    "idx": [0, 1],
    "l": [
        [[0, 2], [1]],
        [[2, 3], [4, 5]]
    ]
})

fname = "/path/to/file/list_test"

df.to_parquet(fname)

ak.connect()
print(ak.read_parquet(fname))
ak.disconnect()

# output from print - this is expected because column l will not be read
[0 1]
```

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -406,13 +406,6 @@ module ParquetMsg {
     // else if t == ARROWSTRING then return ArrowTypes.stringArr; // TODO - add handling for this case
     else if t == ARROWDOUBLE then return ArrowTypes.double;
     else if t == ARROWFLOAT then return ArrowTypes.float;
-    else if t == ARROWLIST then return ArrowTypes.list;
-    throw getErrorWithContext(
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good

Issue #2395, opened 2023-04-25
Register & Attach Retooling


Ethan-DeBandi99:
Currently, most register and attach workflows are using delimited strings in responses that are extremely cumbersome to parse (#2394 Updates Categorical). In order to make this more maintainable and less prone to mistakes if updates are needed in the future, we need to update the response messages for attach calls to use JSON return messages via `"%jt".format(map)`. This will make the code more readable and easier to maintain long term. This will also remove the error prone element of delimited strings.

It would also be good to give the full scope of this code a once over and determine if there are any other improvements that can be made. Currently, I believe there is a lot of duplicate code that we may be able to streamline.

Additionally, after PR #2392, we need to ensure that all register and attach workflows are functioning as desired.

Ethan-DeBandi99:
Adding a note that we need to review how `unregister` works because in a lot cases when the object is unregistered, the `self.name` is set to `None`. However, then you cannot interact with the object correctly. I am thinking we should do something like 

```python
# in the init 
self.name = "symbol table name"
self.reg_name = None

# when an object is registered, set the reg_name
self.reg_name = user_defined_name

# on unregister reset to non
self.reg_name = none
```
This way we never lose the symbol table object because if `ak.clear()` is never run the original name should still be in the symbol table.

Also, I think we want to make the registry separate from the symbol table. I believe there are a lot of edge cases currently that may cause issues that need to be thought through during this redesign to ensure proper functionality across the board.

Ethan-DeBandi99:
Adding a note here that PR #2406 is will add code (when it is merged) that is modeled after the dataframe register/attach for server<->server transfers. We will also need to update that when this is done. Either in this issue or as a separate issue.

Ethan-DeBandi99:
Noting a few things that I do not want to forget to do

- [ ] Do we want to maintain `unregister_objtype_by_name`. This would essentially just be a wrapper to call `ak.util.unregister`
- [ ] `is_registered()` methods
- [ ] `list_registry()` variants to return. Want to return the registered objects, the symtab entries that are part of the registered objects, and a combination return that associates the 2.
- [ ] Update docstrings since most will no longer be accurate.
- [ ] Clean up symbol table code interaction with the registry

Ethan-DeBandi99:
Remaining tasks to complete:
- [ ] Testing
- [ ] Code Clean-up
- [ ] Validate that everything currently functional (prior to updates) is still operational

PR #2406, opened 2023-05-01
Closes #2405: Add server <-> server transfers of pdarrays

### LINKED ISSUE #2405: Add server <-> server transfers of pdarrays
Add the ability to send a pdarray from one server to
another server through ZMQ sockets. This can be useful when
multiple Arkouda servers would like to operate on smaller
subsets of one giant dataset that can't fit on each Arkouda
server individually. For example, one large "mother" Arkouda
server could load in the entire dataset, down-select the data
as needed for each "child" server, then send the portion of
that array to the child server that would like to operate on
that portion of the dataset.

Prior to this PR, this would need to be done by writing the
down-selected portion from the mother server, reading those
files into the child server, and then operating. This aims
to replace the file system go-between of this use case,
simplifying the number of steps needed to get a pdarray
from one server to another and also potentially achieving 
better performance than would be possible using the file
system.

Usage:

Server-1 ("mother" server):
```python3
>>> a = ak.read('my_file') # file that's too big for other servers
>>> b = a[0:int((a.size/2))] # get a portion of the array to send to child
>>> b.send_array("child-name", 1234) # send smaller array to server running on "child-name"
```

Server-2 ("child" server):
```python3
a = ak.receive_array("mother-name", 1234) # receive array from "mother-name"
                                    # port must match the port from the `send` call
```

Types supported in initial implementation with `send_array`/`receive_array` interface:
- all native Arkouda datatypes (int, uint, bool, float)
- categorical
- strings
- segmented array

Additionally, DataFrames containing columns of the above supported types can be transferred with a slightly different interface of `DataFrame.transfer(hostname, port)`/`ak.receive_dataframe(hostname, port)`

bmcdonald3:
This PR adds the ability to send a pdarray from one server to
another server through ZMQ sockets. This can be useful when
multiple Arkouda servers would like to operate on smaller
subsets of one giant dataset that can't fit on each Arkouda
server individually. For example, one large "mother" Arkouda
server could load in the entire dataset, downselect the data
as needed for each "child" server, then send the portion of
that array to the child server that would like to operate on
that portion of the dataset.

Prior to this PR, this would need to be done by writing the
downselected portion from the mother server, reading those
files into the child server, and then operating. This aims
to replace the file system go-between of this use case,
simplifying the number of steps needed to get a pdarray
from one server to another and also potentially achieving 
better performance than would be possible using the file
system.

The performance could be improved, especially for the case
where the sending/receiving servers are of different element
sizes, but this initial work is focused on correctness.

Performance numbers collected on a Cray CS HDR transfers over
100G HDR-IB:

| Nodes     | Receiving 2  | Receiving 3  | Receiving 4  |
| --------: | -----------: | -----------: | -----------: |
| Sending 2 | 1.2385 GiB/s | 0.6264 GiB/s | 1.4077 GiB/s |
| Sending 3 | 0.6680 GiB/s | 1.6672 GiB/s | 0.6270 GiB/s |
| Sending 4 | 1.3586 GiB/s | 0.7278 GiB/s | 1.9504 GiB/s |

Types supported in initial implementation with `send_array`/`receive_array` interface:
- all native Arkouda datatypes (int, uint, bool, float)
- categorical
- strings
- segmented array

Additionally, DataFrames containing columns of the above supported types can be transferred with a slightly different interface of `DataFrame.transfer(hostname, port)`/`ak.receive_dataframe(hostname, port)`

Closes: #2405

### DIFF
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -195,10 +195,7 @@ class Categorical:
         eles = json.loads(rep_msg)
         codes = create_pdarray(eles["codes"])
         cats = Strings.from_return_msg(eles["categories"])
-        if "_akNAcode" in eles:
-            na_code = create_pdarray(eles["_akNAcode"])
-        else:
-            na_code = None
+        na_code = create_pdarray(eles["_akNAcode"])
 
         segments = None
         perm = None
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1446,5 +1446,3 @@ class Categorical:
             cmd="sendArray",
             args=args,
         )
-
-    
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1402,12 +1402,13 @@ class Categorical:
 
     def send_array(self, hostname: str, port: int_scalars):
         """
-        Sends a strings object to a different Arkouda server
+        Sends a Categorical object to a different Arkouda server
 
         Parameters
         ----------
         hostname : str
-            The hostname of the pdarray to receive the array
+            The hostname where the Arkouda server intended to
+            receive the Categorical is running.
         port : int_scalars
             The port to send the array over. This needs to be an
             open port (i.e., not one that the Arkouda server is
@@ -1423,7 +1424,7 @@ class Categorical:
 
         Returns
         -------
-        None
+        A message indicating a complete transfer
 
         Raises
         ------
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1396,3 +1396,52 @@ class Categorical:
                 result_categoricals[base_name] = Categorical.from_codes(**cat_parts)
 
         return removal_names, result_categoricals
+
+    def send_array(self, hostname: str, port: int_scalars):
+        """
+        Sends a strings object to a different Arkouda server
+
+        Parameters
+        ----------
+        hostname : str
+            The hostname of the pdarray to receive the array
+        port : int_scalars
+            The port to send the array over. This needs to be an
+            open port (i.e., not one that the Arkouda server is
+            running on). This will open up `numLocales` ports,
+            each of which in succession, so will use ports of the
+            range {port..(port+numLocales)} (e.g., running an
+            Arkouda server of 4 nodes, port 1234 is passed as
+            `port`, Arkouda will use ports 1234, 1235, 1236,
+            and 1237 to send the array data).
+            This port much match the port passed to the call to
+            `ak.receive_array()`.
+
+
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        ValueError
+            Raised if the op is not within the pdarray.BinOps set
+        TypeError
+            Raised if other is not a pdarray or the pdarray.dtype is not
+            a supported dtype
+        """
+        # hostname is the hostname to send to
+        args = {
+            "codes": self.codes,
+            "categories": self.categories,
+            "objType": "categorical",
+            "NA_codes": self._akNAcode,
+            "hostname": hostname,
+            "port": port
+        }
+        return generic_msg(
+            cmd="sendArray",
+            args=args,
+        )
+
+    
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -1400,7 +1400,7 @@ class Categorical:
 
         return removal_names, result_categoricals
 
-    def send_array(self, hostname: str, port: int_scalars):
+    def transfer(self, hostname: str, port: int_scalars):
         """
         Sends a Categorical object to a different Arkouda server
 
@@ -1438,7 +1438,7 @@ class Categorical:
         args = {
             "codes": self.codes,
             "categories": self.categories,
-            "objType": "categorical",
+            "objType": self.objType,
             "NA_codes": self._akNAcode,
             "hostname": hostname,
             "port": port
--- a/arkouda/categorical.py
+++ b/arkouda/categorical.py
@@ -195,7 +195,10 @@ class Categorical:
         eles = json.loads(rep_msg)
         codes = create_pdarray(eles["codes"])
         cats = Strings.from_return_msg(eles["categories"])
-        na_code = create_pdarray(eles["_akNAcode"])
+        if "_akNAcode" in eles:
+            na_code = create_pdarray(eles["_akNAcode"])
+        else:
+            na_code = None
 
         segments = None
         perm = None
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -2687,16 +2687,16 @@ def invert_permutation(perm):
     return coargsort([perm, arange(perm.size)])
 
 
-def receive_dataframe(hostname : str, port):
+def receive(hostname : str, port):
     """
-    Receive a pdarray sent by `pdarray.send_array()`.
+    Receive a pdarray sent by `dataframe.transfer()`.
 
     Parameters
     ----------
     hostname : str
-        The hostname of the pdarray that sent the array
+        The hostname of the dataframe that sent the array
     port : int_scalars
-        The port to send the array over. This needs to be an
+        The port to send the dataframe over. This needs to be an
         open port (i.e., not one that the Arkouda server is
         running on). This will open up `numLocales` ports,
         each of which in succession, so will use ports of the
@@ -2710,8 +2710,8 @@ def receive_dataframe(hostname : str, port):
     Returns
     -------
     pdarray
-        The pdarray sent from the sending server to the current
-        receiving server.
+        The dataframe sent from the sending server to the
+        current receiving server.
 
     Raises
     ------
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -634,7 +634,7 @@ class DataFrame(UserDict):
         msg_list = []
         for col in self._columns:
             if isinstance(self[col], Categorical):
-                msg_list.append(f"Categorical+{col}+{self[col].codes.name}+{self[col].categories.name}")
+                msg_list.append(f"Categorical+{col}+{self[col].codes.name}+{self[col].categories.name}+{self[col]._akNAcode.name}")
             elif isinstance(self[col], SegArray):
                 msg_list.append(f"SegArray+{col}+{self[col].segments.name}+{self[col].values.name}")
             elif isinstance(self[col], Strings):
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -21,6 +21,7 @@ from arkouda.groupbyclass import GROUPBY_REDUCTION_TYPES
 from arkouda.groupbyclass import GroupBy as akGroupBy
 from arkouda.groupbyclass import unique
 from arkouda.index import Index
+from arkouda.io import _dict_recombine_segarrays_categoricals, get_filetype, load_all, _build_objects
 from arkouda.numeric import cast as akcast
 from arkouda.numeric import cumsum, where
 from arkouda.pdarrayclass import RegistrationError, pdarray
@@ -43,6 +44,7 @@ __all__ = [
     "intersect",
     "invert_permutation",
     "intx",
+    "receive_dataframe",
 ]
 
 
@@ -592,6 +594,43 @@ class DataFrame(UserDict):
         new_df = DataFrame(df_dict)
         new_df._set_index(self.index.index[idx])
         return new_df.to_pandas(retain_index=True)[self._columns]
+
+    def transfer(self, hostname, port):
+        self.update_size()
+        idx = self._index
+        msg_list = []
+        for col in self._columns:
+            if isinstance(self[col], Categorical):
+                msg_list.append(f"Categorical+{col}+{self[col].codes.name}+{self[col].categories.name}")
+            elif isinstance(self[col], SegArray):
+                msg_list.append(f"SegArray+{col}+{self[col].segments.name}+{self[col].values.name}")
+            elif isinstance(self[col], Strings):
+                msg_list.append(f"Strings+{col}+{self[col].name}")
+            elif isinstance(self[col], Fields):
+                msg_list.append(f"Fields+{col}+{self[col].name}")
+            elif isinstance(self[col], IPv4):
+                msg_list.append(f"IPv4+{col}+{self[col].name}")
+            elif isinstance(self[col], Datetime):
+                msg_list.append(f"Datetime+{col}+{self[col].name}")
+            elif isinstance(self[col], BitVector):
+                msg_list.append(f"BitVector+{col}+{self[col].name}")
+            else:
+                msg_list.append(f"pdarray+{col}+{self[col].name}")
+
+        repMsg = cast(
+            str,
+            generic_msg(
+                cmd="sendDataframe",
+                args={
+                    "size"     : len(msg_list),
+                    "idx_name" : idx.name,
+                    "columns"  : msg_list,
+                    "hostname" : hostname,
+                    "port"     : port
+                },
+            ),
+        )
+        return repMsg
 
     def _shape_str(self):
         return f"{self.size} rows x {self._ncols()} columns"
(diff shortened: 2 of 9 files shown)


hokiegeek2:
@bmcdonald3 this looks cool. Question -> when an array is being transferred from one arkouda instance to another, the zmq  ports of the sending and receive servers block until the array transfer completes, correct?

bmcdonald3:
@hokiegeek2 Yes, that is correct

bmcdonald3:
@hokiegeek2 No, I don't have that information, but, based on the numbers above, we observed 0.1 GiB/s per-node as-is, and that was without targeting the high speed network (as mentioned, this code is unoptimized and there are clear next steps for optimization). If we assume this 50 TB array is over 50 nodes, that would be 5 GiB/s injection, so about 15 minutes.

But, aside from the raw numbers of this initial implementation, what I would say in response to that  is that, since this aims to replace writing/reading files (the only way to accomplish moving data from one server to another today), if we are targeting the high speed network (as indicated, the numbers in the table in the original PR are not, but saw large speedups when targeting infiniband), we have a much higher potential injection rate than doing so through the filesystem. If we assume 100 GiB/s infiniband (actual number) as the high bound, and 25 GiB/s to read/write files as a high bound (optimistic numbers), that would mean this would go 8x faster than reading/writing the files (since with files we have to read and write, the injection rate would have to take into both the writing and reading stages, so the injection rate here from reading/writing is actually 12.5 GiB/s, even though we can read/write at 25 GiB/s, so 8x lower than the 100 GiB/s potential from transferring via sockets).

So, while this would block the server for a while when transferring a large array, I don't see that as unique to this operation and think that we would see significant improvements from transferring the files as opposed to reading/writing the files in terms of performance (which is equivalent to the amount of unresponsive time you mention).

hokiegeek2:
@bmcdonald3 gotcha, okay. Just curious, what kind of network did you test on?

bmcdonald3:
Good question! This was tested using 1G Ethernet (I've updated the original description to say that).

In practice, we wouldn't want to use Ethernet for these transfers in most cases, so those performance numbers aren't particularly valuable I don't think and they were really collected as a correctness test more than for the performance and Elliot advised against putting out that performance table, but the table looked nice and I had already collected the numbers, so I had to include it haha.

I plan to run some actual studies using IPoIB in the near future to test more realistic numbers, but since this initial implementation was focused on establishing a baseline with correctness, rather than performance, I have not had a chance to do a full sweep of that yet.

pierce314159:
> @hokiegeek2 No, I don't have that information, but, based on the numbers above, we observed 0.1 GiB/s per-node as-is, and that was without targeting the high speed network (as mentioned, this code is unoptimized and there are clear next steps for optimization). If we assume this 50 TB array is over 50 nodes, that would be 5 GiB/s injection, so about 15 minutes.
> 
> But, aside from the raw numbers of this initial implementation, what I would say in response to that is that, since this aims to replace writing/reading files (the only way to accomplish moving data from one server to another today), if we are targeting the high speed network (as indicated, the numbers in the table in the original PR are not, but saw large speedups when targeting infiniband), we have a much higher potential injection rate than doing so through the filesystem. If we assume 100 GiB/s infiniband (actual number) as the high bound, and 25 GiB/s to read/write files as a high bound (optimistic numbers), that would mean this would go 8x faster than reading/writing the files (since with files we have to read and write, the injection rate would have to take into both the writing and reading stages, so the injection rate here from reading/writing is actually 12.5 GiB/s, even though we can read/write at 25 GiB/s, so 8x lower than the 100 GiB/s potential from transferring via sockets).
> 
> So, while this would block the server for a while when transferring a large array, I don't see that as unique to this operation and think that we would see significant improvements from transferring the files as opposed to reading/writing the files in terms of performance (which is equivalent to the amount of unresponsive time you mention).

@bmcdonald3 and @hokiegeek2 i gotta question. I think this is really interesting, im curious how this would compare in the case when you have multiple "child" servers? So if everyone on a team of `N` people wanted to play with their own copy of the filtered down dataset. With the transfer approach, we'd need to do `N` transfers (one for each "child" server). With the file system approach, we'd only have to do a single write and do `N` reads. Also (my ignorance of file IO is showing :) ) would it be possible for those reads to be done simultaneously? Because if so, it feels like for a large enough number of team-members the file system would be faster right? It's possible that number is too high to every realistically hit and it's possible that none of what i said made any sense lol

pierce314159:
> Elliot advised against putting out that performance table, but the table looked nice and I had already collected the numbers, so I had to include it haha.
> 
lmao the table _does_ look nice 😁

bmcdonald3:
> im curious how this would compare in the case when you have multiple "child" servers? So if everyone on a team of N people wanted to play with their own copy of the filtered down dataset. With the transfer approach, we'd need to do N transfers (one for each "child" server). With the file system approach, we'd only have to do a single write and do N reads. Also (my ignorance of file IO is showing :) ) would it be possible for those reads to be done simultaneously?

This is interesting question, but I think there are a lot of unknowns and this gets into a classic networking problem. If every single server did want the exact same dataset, then I think the most efficient way to do that would be to have the head server transfer to a child, then have that child become another worker node, so that child would then transfer it's data to another server, so you'd get something like: 1 node -> transfer -> 2 nodes -> transfer -> 4 nodes -> transfer -> 8 nodes ..., where each new child node is starting to do some transfers as well, so I think you could get `log(N)` time there.

This would definitely be a best case scenario though, it would be pretty hard to get a bunch of different people acting in unison to make something like that happen, but if the goal was just freeing up the one big server, then that one could just do one transfer and then have that child be the new server to farm things out, but all of this is speculative, since I don't have a concrete use case in mind that I know this is looking to help with, it is possible that IO is better, but it would depend on the use case.

What I was imagining based on past conversations was that every child server wanted a different portion of the dataset, which is a very different problem.

This is information that I stole from Elliot, so I can't say a ton about it, but networks and file systems typically have three different limits: limit for single thread transfer, limit for single node transfer, limit for total simultaneous transfer amongst all nodes. Inifiniband has a much higher total simultaneous transfer upper bound than the typical file system. So if we were really working with huge 50TB arrays like John mentioned, we would definitely be hitting that third bottleneck of total simultaneous transfer amongst all nodes, where if we were trying to read 50 TB in from the same filesystem on 5 different servers, that will basically be serialized and all the servers would be doing their reads one at a time, not in parallel. On much smaller data sets where you aren't capping out that total transfer rate on a single server, the parallelism from reading from multiple servers could come in handy, but that wouldn't be a case where the server was stalling for 15 minutes.

In summary, I am not an expert on this stuff and I think that the real value here is dependent on the actual use case in question.

pierce314159:
> This is interesting question, but I think there are a lot of unknowns and this gets into a classic networking problem. If every single server did want the exact same dataset, then I think the most efficient way to do that would be to have the head server transfer to a child, then have that child become another worker node, so that child would then transfer it's data to another server, so you'd get something like: 1 node -> transfer -> 2 nodes -> transfer -> 4 nodes -> transfer -> 8 nodes ..., where each new child node is starting to do some transfers as well, so I think you could get `log(N)` time there.
> 
Ooooo i didn't think about using the child servers to do transfers, that's clever
> if we were trying to read 50 TB in from the same filesystem on 5 different servers, that will basically be serialized and all the servers would be doing their reads one at a time, not in parallel. On much smaller data sets where you aren't capping out that total transfer rate on a single server, the parallelism from reading from multiple servers could come in handy, but that wouldn't be a case where the server was stalling for 15 minutes.
> 
Okay that makes a lot of sense! thanks ben/elliot

arkouda/io.py
```diff
@@ -1912,3 +1913,43 @@ def read_tagged_data(
...
+        The hostname of the pdarray that sent the array
+    port : int_scalars
+        The port to send the array over. This needs to be an
+        open port (i.e., not one that the Arkouda server is
+        running on). This will open up `numLocales` ports,
+        each of which in succession, so will use ports of the
+        range {port..(port+numLocales)} (e.g., running an
+        Arkouda server of 4 nodes, port 1234 is passed as
+        `port`, Arkouda will use ports 1234, 1235, 1236,
+        and 1237 to send the array data).
```
pierce314159, line 1934: would these ports need to be the same or distinct from those in `send_array`? I'm guessing distinct (or at least not necessarily the same) because we could have a different number of locales between the receiving and sending servers right?

arkouda/categorical.py
```diff
@@ -1356,3 +1359,51 @@ def parse_hdf_categoricals(
...
+            receive the Categorical is running.
+        port : int_scalars
+            The port to send the array over. This needs to be an
+            open port (i.e., not one that the Arkouda server is
+            running on). This will open up `numLocales` ports,
+            each of which in succession, so will use ports of the
+            range {port..(port+numLocales)} (e.g., running an
+            Arkouda server of 4 nodes, port 1234 is passed as
+            `port`, Arkouda will use ports 1234, 1235, 1236,
+            and 1237 to send the array data).
```
pierce314159, line 1380: i don't know much about networking but i wanted to throw this out there. Idk if it's a reasonable assumption that a block of num locales ports would be free. I know that num_locales can get pretty big in real use cases, so I just wanted to check that this was something ppl who know things about networks had thought about lol

pierce314159 (commented):
my initial review, I got a couple questions mostly me not understanding ports and stuff that well

I didn't really dig into the chpl side too much. I figured i would let @hokiegeek2 and @Ethan-DeBandi99 take a crack at that first since they have more expertise. i'll take a look on my next pass through

arkouda/categorical.py
```diff
@@ -200,7 +200,10 @@ def from_return_msg(cls, rep_msg) -> Categorical:
         eles = json.loads(rep_msg)
         codes = create_pdarray(eles["codes"])
         cats = Strings.from_return_msg(eles["categories"])
-        na_code = create_pdarray(eles["_akNAcode"])
+        if "_akNAcode" in eles:
+            na_code = create_pdarray(eles["_akNAcode"])
+        else:
+            na_code = None
```
Ethan-DeBandi99, line 206: `ak.Categorical._akNAcode` should always be defined. Looking at the `init`, there are cases where it will not be a `pdarray` (which means we there are alterations we need to make unrelated to this PR). However, in this case, `_akNAcode` should always be defined and should always be a `pdarray`.

arkouda/categorical.py
```diff
@@ -1356,3 +1359,51 @@ def parse_hdf_categoricals(
...
+            Raised if the op is not within the pdarray.BinOps set
+        TypeError
+            Raised if other is not a pdarray or the pdarray.dtype is not
+            a supported dtype
+        """
+        # hostname is the hostname to send to
+        args = {
+            "codes": self.codes,
+            "categories": self.categories,
+            "objType": "categorical",
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I agree with @pierce314159 that we should be using `self.objType`, but disagree that it can wait. We really should be doing this with all new code sending the object type to prevent adding additional tech debt.

arkouda/dataframe.py
```diff
@@ -641,6 +642,76 @@ def _get_head_tail_server(self):
         new_df._set_index(self.index.index[idx])
         return new_df.to_pandas(retain_index=True)[self._columns]
 
+    def transfer(self, hostname, port):
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Just noting that I agree with @pierce314159 here.

arkouda/categorical.py
```diff
@@ -1356,3 +1359,51 @@ def parse_hdf_categoricals(
...
+            "NA_codes": self._akNAcode,
```
Ethan-DeBandi99, line 1402: Based on the `__init__`, it looks like there are cases where this can be an `int`. Maybe I just haven't seen it yet, but are we checking that somewhere? If not, it would probably be best to create the `pdarray` here.

Again, this is definitely not checked in other locations in existing code. I will add an issue to look into this and make sure it gets updated.

arkouda/dataframe.py
```diff
@@ -641,6 +642,76 @@ def _get_head_tail_server(self):
...
+            elif isinstance(self[col], Fields):
+                msg_list.append(f"Fields+{col}+{self[col].name}")
+            elif isinstance(self[col], IPv4):
+                msg_list.append(f"IPv4+{col}+{self[col].name}")
+            elif isinstance(self[col], Datetime):
+                msg_list.append(f"Datetime+{col}+{self[col].name}")
+            elif isinstance(self[col], BitVector):
+                msg_list.append(f"BitVector+{col}+{self[col].name}")
+            else:
+                msg_list.append(f"pdarray+{col}+{self[col].name}")
```
Ethan-DeBandi99, line 698: We are trying to move away from delimited strings within the messaging layer of Arkouda. It looks like it would be a rather large up, but for what it is worth, we typically use multiple lists in these cases where the indexes line everything up. Additionally, that would allow for using the `objType` properties and the `ObjType` enum on the server for consistency. We could also do that here, but again I think we want to move away from delimited strings for messages. 

Also, noting that to that some classes/objects may need to have the `objType` property added in the class and to the enum on the server.

src/TransferMsg.chpl
```diff
@@ -0,0 +1,883 @@
...
+
+      var localeCount = receiveLocaleCount(hostname, port:string);
+      
+      sendDataFrameSetupInfo(port:string, numColumns, eleList: string);
+
+      var colNames = "";
+      for ele in eleList {
+        var ele_parts = ele.split("+");
+        sendColumnName(port, ele_parts[1]);
+        if ele_parts[0] == "Categorical" {
```
Ethan-DeBandi99, line 62: It would be good if we use the `ObjType` enum here. I think we will need to add some things to it for that.

src/TransferMsg.chpl
Ethan-DeBandi99, line 1: I have just a general implementation question. This may be due to a lack of experience with ZMQ, but just want to clarify. In general, all ZMQ traffic in Arkouda should be going through locale0. It looks like in this implementation, each Locale is getting its own port to communicate on, but doesn't that traffic still go through locale0? I know there have been previous things that we decided not to do because even opening a separate port would still block the existing an vice versa. I am just worried that if that is the case, even if multiple locales are sending or receiving at once they will be blocking each other. Maybe I am completely missing something, or just need to be reminded of something I've already had explained to me. If that is the case I apologize, but could we get clarification on this?

Ethan-DeBandi99 (commented):
Couple of things to address. And one question for clarification.

src/TransferMsg.chpl
```diff
@@ -0,0 +1,883 @@
...
+    use Map;
+    use ServerErrors;
+    use ServerConfig;
+    use ServerErrorStrings;
+
+    use SegmentedString;
+
+    proc sendDataFrameSetupInfo(port:string, numColumns: int, elements: string) throws {
+      var context: Context;
+      var socket = context.socket(ZMQ.PUSH);
```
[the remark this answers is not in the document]
bmcdonald3: This transfer will happen from locale0 on the sending server to locale0 on the receiving server and will only use 1 port. This transfer only happens once per dataframe transfer (see line 56).

arkouda/io.py
bmcdonald3: It will be the number of ports from the server with more locales, so, if we had a server sending with 4 locales and only receiving on 2, both of the servers will open up the 4 ports (since those are the ports that were sent on).

src/TransferMsg.chpl
bmcdonald3: > all ZMQ traffic in Arkouda should be going through locale0

Is there a reason this should be the case? I think this is the case for the usage of Arkouda's ZMQ today, but that is not the case with how it is used in this PR. 

> doesn't that traffic still go through locale0

No, each locale has it's own ZMQ instance that is handling the traffic.

I'm not sure if I'm answering the question here, but each separate locale is doing it's own communication, so there won't be any overlap with the existing usage of ZMQ for message sending or message sending from the different locales (which is why you see the performance get better as you increase the number of locales in the initial post). Does that help?

arkouda/categorical.py
bmcdonald3: OK I think I did something stupid there, working to get that fixed now, thanks for pointing out.

arkouda/io.py
pierce314159: cool thanks for clarifying!

arkouda/dataframe.py
Ethan-DeBandi99: @bmcdonald3 - I added a note to issue #2395 to make sure that when the code this is modeled after is updated to move away from delimited. I am not thrilled about additional tech debt here, but I think in this instance I am ok with it.

src/TransferMsg.chpl
Ethan-DeBandi99: @bmcdonald3 - I think so. I guess my only other question would be do the sockets on each locale block one another? I am assuming not based on the performance, but would rather do the sanity check now.

bmcdonald3: No, they don't. ZMQ messages are sent asynchronously in a non-blocking manner, but there are additional dials in ZMQ you can change to modify that behavior. For the Chapel ZMQ usage, we have it default to use `ZMQ_DONTWAIT`, which is their way of saying non-blocking.

Ethan-DeBandi99: Perfect thanks.

pierce314159 (approved):

bradcray:
Sweet!

PR #2409, opened 2023-05-02
Deprecation updates for `BitOps.popcount` and `bigint.mod`

jeremiah-corrado:
Updates for the following deprecations:

- `bigint.mod`: replaced by the standalone `BigInteger.mod` function
- `popcount`: renamed to `popCount`

Adds a compatibility module for bitops. A compatibility module for BigInteger was already set up.

### DIFF
--- a/src/compat/gt-130/ArkoudaBitOpsCompat.chpl
+++ b/src/compat/gt-130/ArkoudaBitOpsCompat.chpl
@@ -1,1 +1,7 @@
-module ArkoudaBitOpsCompat { }
+module ArkoudaBitOpsCompat {
+  use BitOps;
+
+  proc popCount(x: integral) {
+    return popcount(x);
+  }
+}
--- a/src/compat/gt-130/ArkoudaBitOpsCompat.chpl
+++ b/src/compat/gt-130/ArkoudaBitOpsCompat.chpl
@@ -1,0 +1,1 @@
+module ArkoudaBitOpsCompat { }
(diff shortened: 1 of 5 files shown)


src/compat/gt-130/ArkoudaBitOpsCompat.chpl
```diff
@@ -0,0 +1,7 @@
+module ArkoudaBitOpsCompat {
+  use BitOps;
+
+  proc popCount(x: integral) {
+    return popcount(x);
+  }
+}
```
bmcdonald3, line 7: For greater than 1.30, we have the `popCount` method in Chapel I think, so we don't need this overload (i.e., we need the `ArkoudaBitOpsCompat` module for all 3 versions of Chapel, but the one that supports the deprecated feature, we can just have it be an empty module, like https://github.com/Bears-R-Us/arkouda/blob/master/src/compat/gt-130/ArkoudaBigIntCompat.chpl).

Does that make sense? Also possible that I am missing something.

bmcdonald3 (changes requested):

src/compat/gt-130/ArkoudaBitOpsCompat.chpl
jeremiah-corrado: Yeah that makes sense, I'll remove that function.

bmcdonald3 (approved):
Looks good, feel free to promote to a regular PR, thanks Jeremiah!

jeremiah-corrado:
Great, thanks!

pierce314159 (approved):

PR #2415, opened 2023-05-08
Adjust these modules to avoid deprecation warnings from non-default Math symbols

lydia-duncan:
`exp`, `sin` and `cos` will stop being included in all Chapel programs by default in 1.31 and will instead require a `use` or `import` of the Math module. Change these modules to fully use the Math module, which will be compatible with previous versions of Chapel since the module already existed.

----


bmcdonald3 (approved):
Looks good, thanks Lydia! Ready to make it a regular PR 👍

pierce314159 (approved):

PR #2428, opened 2023-05-12
Closes #2425 - Import/Export lists from/to pandas

### LINKED ISSUE #2425: Import/Export lists from/to pandas
**Feature Description**
Export segarrays to pandas as a list and import columns containing lists as segarrays

Ethan-DeBandi99:
Closes #2425 

Noting that exporting an Arkouda DataFrame to list was already supported by calling `ak.DataFrame.to_pandas`. This is also supported for a standalone SegArray using `ak.SegArray.to_ndarray()`.

The previous implementation of `ak.DataFrame.__init__` did not handle cases where pandas DataFrames contained lists (ie SegArrays). This constructor case has been updated to include handling for list columns. These will be read in a SegArrays.

In order to support this, `ak.SegArray.from_multi_array` has been updated so that each array is a segment instead of each array containing a value for a column. This prevents the requirement that everything be the same size when using this method, allowing true SegArray functionality.  This resulted in some tests requiring updates in order to support this, but these were minor.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -2439,7 +2439,7 @@ class DataFrame(UserDict):
 
             elif parts[i] == "categorical":
                 colName = DataFrame._parse_col_name(parts[i + 1], dfName)[0]
-                cols[colName] = Categorical.from_return_msg(parts[i+2] + "+" + parts[i+3])
+                cols[colName] = Categorical.from_return_msg(parts[i + 2] + "+" + parts[i + 3])
                 i += 3
 
             elif parts[i] == "segarray":
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -562,7 +562,7 @@ class DataFrame(UserDict):
     def _get_head_tail_server(self):
         if self._empty:
             return pd.DataFrame()
-        # self.update_size()
+        self.update_size()
         maxrows = pd.get_option("display.max_rows")
         if self._size <= maxrows:
             newdf = DataFrame()
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -278,7 +278,13 @@ class DataFrame(UserDict):
             # convert the lists defining each column into a pdarray
             # pd.DataFrame.values is stored as rows, we need lists to be columns
             for key, val in initialdata.to_dict("list").items():
-                self.data[key] = array(val)
+                if isinstance(val[0], list):
+                    multi_array = []
+                    for r in val:
+                        multi_array.append(array(r))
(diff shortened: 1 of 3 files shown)


hokiegeek2 (approved):
looks good

jaketrookman (approved):
looks good

arkouda/dataframe.py
```diff
@@ -278,7 +278,13 @@ def __init__(self, initialdata=None, index=None):
             # convert the lists defining each column into a pdarray
             # pd.DataFrame.values is stored as rows, we need lists to be columns
             for key, val in initialdata.to_dict("list").items():
-                self.data[key] = array(val)
+                if isinstance(val[0], list):
+                    multi_array = []
+                    for r in val:
+                        multi_array.append(array(r))
```
pierce314159, line 284: Stylistic, but would something like this work?
```python
multi_array = [array(r) for r in val]
```
i think we could also make this entire block one line if ppl like that style (black would break it up so it wouldn't be too long)
```python
self.data[key] = SegArray.from_multi_array([array(r) for r in val]) if isinstance(val[0], list) else array(val)
```

tests/segarray_test.py
```diff
@@ -35,9 +35,10 @@ def test_creation(self):
         self.assertTrue(segarr.__eq__(segarr).all())
 
         multi_pd = ak.SegArray.from_multi_array(
-            [ak.array([10, 20, 30]), ak.array([11, 21, 31]), ak.array([12, 22, 32])]
+            [ak.array([10, 11, 12]), ak.array([20, 21, 22]), ak.array([30, 31, 32])]
         )
         self.assertIsInstance(multi_pd, ak.SegArray)
+        print(multi_pd.__str__())
```
pierce314159, line 41:
```suggestion

```

pierce314159 (commented):
logic looks good! small suggestions, mostly stylistic

arkouda/dataframe.py
```diff
@@ -278,13 +278,18 @@ def __init__(self, initialdata=None, index=None):
...
+                #     for r in val:
+                #         multi_array.append(array(r))
+                #     self.data[key] = SegArray.from_multi_array(multi_array)
+                # else:
+                #     self.data[key] = array(val)
+                self.data[key] = (
+                    SegArray.from_multi_array([array(r) for r in val])
+                    if isinstance(val[0], list)
+                    else array(val)
+                )
```
pierce314159, line 292:
```suggestion
                self.data[key] = (
                    SegArray.from_multi_array([array(r) for r in val])
                    if isinstance(val[0], list)
                    else array(val)
                )
```

pierce314159 (approved):

PR #2438, opened 2023-05-18
Closes #2434 - Strings Hashing bugfix

### LINKED ISSUE #2434: hashing columns containing strings
**Describe the bug**
```python
a = ak.array(['a','b'])
b = ak.array([1,2])

ak.hash([a,b])  # fails

ak.hash(a)  # this also fails, but isn't a problem because a.hash() still works
```

**Error Message**
from typeguard
```
TypeError: type of argument "pda" must be one of (pdarray, List); got list instead
```

**Is this a Blocking Issue**
workaround available for temporary work.

**Additional context**
seg_array also issue

joshmarshall1:
This PR closes #2434 

Patches a bug where Strings being passed into the hash api would fail. Also adds support for Categorical and SegArrays being hashed using their codes and values respectively. 

Due to mypy and circular imports in this file, I added the `hashable` type that is comprised of a `Union` of `pdarray, Strings, Categorical, and Segarray`. This was mainly due to the `# type: ignore` decorator causing the argument in the function definition to extend beyond the 105 character limit.

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -43,7 +43,7 @@ __all__ = [
     "ErrorMode",
 ]
 
-hashable = Union[pdarray, Categorical, SegArray, Strings]  # type: ignore
+hashable = Union[pdarray, Categorical, Strings]  # type: ignore
 
 
 class ErrorMode(Enum):
@@ -425,6 +425,15 @@ def cos(pda: pdarray) -> pdarray:
     return create_pdarray(repMsg)
 
 
+def _hash_helper(a):
+    from arkouda import Categorical as _Categorical
+
+    if isinstance(a, _Categorical):
+        return a.categories[a.codes]
+    else:
+        return a
+
+
 @typechecked
 def hash(
     pda: Union[hashable, List[hashable]],
@@ -479,35 +488,14 @@ def hash(
     the array.
     """
     from arkouda import Categorical as _Categorical
-    from arkouda import SegArray as _SegArray
 
-    if isinstance(pda, pdarray) or isinstance(pda, _Categorical) or isinstance(pda, _SegArray):
-        if isinstance(pda, _Categorical):
-            a = pda.codes
-        elif isinstance(pda, _SegArray):
-            a = pda.values
-        else:
-            a = pda
-
-        return _hash_single(a, full)
-    if isinstance(pda, Strings):
-        return pda.hash()
+    if isinstance(pda, (pdarray, _Categorical, Strings)):
+        pda = _hash_helper(pda)
+        return pda.hash() if isinstance(pda, Strings) else _hash_single(pda, full)
     elif isinstance(pda, List):
-        namesList = []
-        typesList = []
-        for n in pda:
-            objType = n.objType
-            if isinstance(n, Strings) or isinstance(n, pdarray):
-                namesList.append(n.name)
-                objType = n.objType
-            elif isinstance(n, _Categorical):
-                namesList.append(n.codes.name)
-                objType = n.codes.objType
-            elif isinstance(n, _SegArray):
-                namesList.append(n.values.name)
-                objType = n.values.objType
-
-            typesList.append(objType)
+        ready = [_hash_helper(a) for a in pda]
+        namesList = [n.name for n in ready]
+        typesList = [n.objType for n in ready]
 
         repMsg = type_cast(
             str,
@@ -517,7 +505,7 @@ def hash(
                     "nameslist": namesList,
                     "typeslist": typesList,
                     "length": len(pda),
-                    "size": len(pda[0].values) if isinstance(pda[0], _SegArray) else len(pda[0]),
+                    "size": len(pda[0]),
                 },
             ),
         )
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -43,6 +43,8 @@ __all__ = [
     "ErrorMode",
 ]
 
+hashable = Union[pdarray, Categorical, SegArray, Strings]  # type: ignore
+
 
 class ErrorMode(Enum):
     strict = "strict"
@@ -425,8 +427,8 @@ def cos(pda: pdarray) -> pdarray:
 
 @typechecked
 def hash(
-    pda: Union[pdarray, Categorical, SegArray, Strings,
-               List[Union[pdarray, Categorical, SegArray, Strings]]], full: bool = True
+    pda: Union[hashable, List[hashable]],
+        full: bool = True
 ) -> Union[Tuple[pdarray, pdarray], pdarray]:
     """
     Return an element-wise hash of the array.
@@ -476,13 +478,13 @@ def hash(
     cancel each other out, hence we do a rotation by the ordinal of
     the array.
     """
-    from arkouda import Categorical
-    from arkouda import SegArray
+    from arkouda import Categorical as _Categorical
+    from arkouda import SegArray as _SegArray
 
-    if isinstance(pda, pdarray) or isinstance(pda, Categorical) or isinstance(pda, SegArray):
-        if isinstance(pda, Categorical):
+    if isinstance(pda, pdarray) or isinstance(pda, _Categorical) or isinstance(pda, _SegArray):
+        if isinstance(pda, _Categorical):
             a = pda.codes
-        elif isinstance(pda, SegArray):
+        elif isinstance(pda, _SegArray):
             a = pda.values
         else:
             a = pda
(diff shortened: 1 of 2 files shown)


Ethan-DeBandi99 (changes requested):
It does not appear that there is any testing for hashing `Strings`. That needs to be added before this is merged (especially because it is a bugfix).

arkouda/numeric.py
```diff
@@ -474,21 +478,52 @@ def hash(
...
-        ),
-    )
+    from arkouda import Categorical as _Categorical
+    from arkouda import SegArray as _SegArray
+
+    if isinstance(pda, pdarray) or isinstance(pda, _Categorical) or isinstance(pda, _SegArray):
+        if isinstance(pda, _Categorical):
+            a = pda.codes
+        elif isinstance(pda, _SegArray):
+            a = pda.values
```
pierce314159, line 488: also i know i signed off hashing the values for segarray during initial design decisions, but after playing around a bit more, I don't think this is what we want to do. I think ultimately segarrays are gonna be difficult to handle (for reasons im about to get into). I think we're gonna wanna do some brainstorming for them, but they are def outside the scope of this bug fix and we should put in a separate issue to handle them

```python
# consider this segarray
>>> sa = ak.segarray(ak.array([0,3,6]), ak.array([0,1,2,3,4,5,0,1,2]))
>>> sa
SegArray([
[0 1 2]
[3 4 5]
[0 1 2]
])

# intuitively the first and third elem should has to the same thing
# since they are the same array and the second should be it's own thing

# we can't get this behavior by hashing the segments since 
# they don't capture the fact that sa[0] and sa[2] are identical
>>> sa.segments
array([0 3 6])

>>> ak.hash(sa.segments)
(array([14020738532124324502 10334800848021183012 12294710806803928155]),
 array([13660161615634858013 14420950124940837019 318151776295263894]))

# we can't hash the values cause they don't really match back up to the original segarray
>>> ak.hash(sa.values)
(array([14020738532124324502 17469661437632883388 2223077377302025035 10334800848021183012 12899154897215201871 1827545583036225533 14020738532124324502 17469661437632883388 2223077377302025035]),
 array([13660161615634858013 100355499037336788 7178333632624491744 14420950124940837019 4776206602043303594 14424044815130752497 13660161615634858013 100355499037336788 7178333632624491744]))

# we thought about hashing each sa[i] and then hashing the resulting pdarray
# but ak.hash(pdarray) returns a pdarray not a single value

# we can't just concatenate the values within a segment because how do u differentiate between 
# [1, 11, 2] and [11, 1, 2] for instance

# one absolutely awful idea that technically works... is to interpret each inner array as a string and then hash that
# (shown here using __repr__ but we would have to do cast server side because repr truncates)
>>> ak.hash(ak.array([ak.array(sa[i]).__repr__() for i in range(3)]))
(array([16660294113708371871 6932019251617944032 16660294113708371871]),
 array([10308859934083080982 18056049045803087039 10308859934083080982]))
# we see that the first and last do match, but it's such an ugly solution
# If we go this route maybe we do hex representation to shorten string lengths
```
These are just some quick ideas, but doing this well is gonna require some thought. I vote we hold off on segarray hashing for now

pierce314159 (changes requested):
I agree we should add some testing for this functionality. I suggested a helper func that will reduce the number of lines needed. And found that segarray hashing is gonna be complicated and we might wanna hold off on that for now

arkouda/numeric.py
pierce314159: This has been captured in #2459

arkouda/numeric.py
```diff
@@ -474,21 +478,52 @@ def hash(
...
+                namesList.append(n.name)
+                objType = n.objType
+            elif isinstance(n, _Categorical):
+                namesList.append(n.codes.name)
+                objType = n.codes.objType
+            elif isinstance(n, _SegArray):
+                namesList.append(n.values.name)
+                objType = n.values.objType
+
+            typesList.append(objType)
```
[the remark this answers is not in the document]
joshmarshall1: Added the helper function without the SegArray portion. That will be added in #2459

arkouda/numeric.py
joshmarshall1: Removed SegArray handling from this PR

joshmarshall1:
I modified the handling of `Categoricals` after realizing two `Categoricals` with completely different values could have identical `codes` arrays. It was decided the best way to handle this, at least for now, is to use `categorical.categories[categorical.codes]` to generate a `Strings` object that will be unique for every variation of Categoricals.

tests/numeric_test.py
```diff
@@ -227,6 +227,24 @@ def testHash(self):
...
+        h = ak.hash(s)
+        rh = ak.hash(s[rev])
+        self.assertListEqual(h[0].to_list(), rh[0][rev].to_list())
+        self.assertListEqual(h[1].to_list(), rh[1][rev].to_list())
+
+        # Two Categoricals with different categories, but identical codes
+        c1 = ak.Categorical(ak.array(["a", "a", "b", "b", "c"]))
+        c2 = ak.Categorical(ak.array(["d", "d", "e", "e", "f"]))
+        h1 = ak.hash(c1)
+        h2 = ak.hash(c2)
```
pierce314159, line 240:
```suggestion
        h1 = ak.hash(ak.Categorical(ak.array(["a", "a", "b", "b", "c"])))
        h2 = ak.hash(ak.Categorical(ak.array(["d", "d", "e", "e", "f"])))
```

pierce314159 (approved):
Looks good to me

arkouda/numeric.py
```diff
@@ -474,21 +487,31 @@ def hash(
...
+        return pda.hash() if isinstance(pda, Strings) else _hash_single(pda, full)
+    elif isinstance(pda, List):
+        ready = [_hash_helper(a) for a in pda]
+        namesList = [n.name for n in ready]
+        typesList = [n.objType for n in ready]
+
+        repMsg = type_cast(
+            str,
+            generic_msg(
+                cmd="efuncArr",
```
pierce314159, line 503: Why did we got with the command `efuncArr` here? I feel like `"hashlist"` or something like that would be more informative

pierce314159:
> I modified the handling of `Categoricals` after realizing two `Categoricals` with completely different values could have identical `codes` arrays. It was decided the best way to handle this, at least for now, is to use `categorical.categories[categorical.codes]` to generate a `Strings` object that will be unique for every variation of Categoricals.

I did some thinking on this, and it's way more efficient to hash the categories first then expand rather than to expand first and then hash

```python
>>> def test(j):
   ...:      import pandas as pd
   ...:      categories = ak.array([f'str {i}' for i in range(3)])
   ...:      codes = ak.randint(0, 3, 10**j)
   ...:
   ...:      timer1 = pd.Timestamp.now()
   ...:      expand_first_hi, expand_first_lo = ak.hash(categories[codes])
   ...:      timer2 = pd.Timestamp.now()
   ...:      print('expand then hash time')
   ...:      print(timer2-timer1)
   ...:
   ...:      timer1 = pd.Timestamp.now()
   ...:      h = ak.hash(categories)
   ...:      hash_first_hi, hash_first_lo = h[0][codes], h[1][codes]
   ...:      timer2 = pd.Timestamp.now()
   ...:      print('hash then expand time')
   ...:      print(timer2-timer1)
   ...:
   ...:      print()
   ...:      print(f'hash high matches: {ak.all(expand_first_hi == hash_first_hi)}')
   ...:      print(f'hash low matches: {ak.all(expand_first_lo == hash_first_lo)}')
   ...:

>>> test(8)
expand then hash time
0 days 00:00:10.223829
hash then expand time
0 days 00:00:01.357992

hash high matches: True
hash low matches: True
```
These are `10**8` long with comm=none

arkouda/numeric.py
```diff
@@ -422,9 +425,19 @@ def cos(pda: pdarray) -> pdarray:
     return create_pdarray(repMsg)
 
 
+def _hash_helper(a):
+    from arkouda import Categorical as _Categorical
+
+    if isinstance(a, _Categorical):
+        return a.categories[a.codes]
```
pierce314159, line 432: we should update how we handle cats to hash the categories and then do integer indexing instead

pierce314159 (changes requested):

arkouda/numeric.py
pierce314159: I think we should handle adding categoricals in a separate issue too, since we really want to do this idexing and hashing in the chapel code (so it can be easily called by `hashArrays`

pierce314159:
closing in favor of PR #2475 and issue #2462

pierce314159:
closing in favor of PR https://github.com/Bears-R-Us/arkouda/pull/2475 and issue https://github.com/Bears-R-Us/arkouda/issues/2462

PR #2440, opened 2023-05-18
Closes #2417 - `Filename_Codes` match `Categorical.codes`

### LINKED ISSUE #2417: read_tagged_data() sync of Filename_Code and categoricals codes
**Feature Description**
Very low priority. It would be convenient if the codes in the Filename_Code data columns of read_tagged_data() matched the codes of the associated categories in the filename list. (order not important). The benefit would be merely a conceptual simplification rather than any performance benefit.  It eliminates risk of errors human error caused by referencing c.categories with data['Filename_Code']. Again, very low priority.

**Example Behavior**
If applicable, please provide a code block showing expected behavior of the feature.

**Describe Workarounds**
Is there currently a way to achieve the desired functionality (even if it is not ideal)? If so, please provide it for context.

**Additional context**
Add any other context or screenshots about the feature request here. If the requested feature is based on another library (i.e. numpy, pandas, scipy), please provide a link to their supporting documentation for reference.

Ethan-DeBandi99:
Closes #2417 

Updates the filename `Categorical` generation to use `from_codes`. This uses an `arange(filelist.size)` for the code because we know the file list is ordered and the codes are assigned on the server based on the index. Verified with 11 locales that this ensures the `Categorical.codes` and associated categories align with the codes returned from the server.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -12,7 +12,7 @@ from arkouda.categorical import Categorical
 from arkouda.client import generic_msg
 from arkouda.groupbyclass import GroupBy
 from arkouda.pdarrayclass import create_pdarray, pdarray
-from arkouda.pdarraycreation import array
+from arkouda.pdarraycreation import array, arange
 from arkouda.segarray import SegArray
 from arkouda.strings import Strings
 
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
Looks good!!

jaketrookman (approved):
Looks good

PR #2447, opened 2023-05-19
Closes #2400: Add arc and hyperbolic trig functions

### LINKED ISSUE #2400: Add arc and hyperbolic trig functions
Duplicate the [numpy functionality](https://numpy.org/devdocs/reference/routines.math.html) for arc and hyperbolic trig funcs
- [x] arcsin
- [x] arccos
- [x] arctan
- [x] arctan2
- [x] sinh
- [x] cosh
- [x] tanh
- [x] arcsinh
- [x] arccosh
- [x] arctanh

jaketrookman:
This PR (Closes #2400) adds:

- the requested inverse and hyperbolic trig functions
- angle unit conversion functions
- testing for `int`, `uint`, and `float` datatypes.

Additional issues have been created to implement `where` functionality to these functions to more closely match numpy. (#2458 and #2461). This will allow for easy single scalar cases to be put in.

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -598,7 +598,7 @@ def arctan2(num: Union[numeric_scalars, pdarray], denom: Union[numeric_scalars, 
                 ),
             )
         )
-    if isinstance(num, pdarray) and isSupportedNumber(denom):
+    elif isinstance(num, pdarray) and isSupportedNumber(denom):
         return create_pdarray(
             type_cast(
                 str,
@@ -613,7 +613,7 @@ def arctan2(num: Union[numeric_scalars, pdarray], denom: Union[numeric_scalars, 
                 ),
             )
         )
-    if isSupportedNumber(num) and isinstance(denom, pdarray):
+    elif isSupportedNumber(num) and isinstance(denom, pdarray):
         return create_pdarray(
             type_cast(
                 str,
@@ -628,14 +628,10 @@ def arctan2(num: Union[numeric_scalars, pdarray], denom: Union[numeric_scalars, 
                 ),
             )
         )
-    if isSupportedNumber(num) and isSupportedNumber(denom):
-        raise TypeError(
-            "Scalar-scalar case is not supported. At least one argument must be a pdarray."
-        )
     else:
         raise TypeError(
             f"Unsupported types {type(num)} and/or {type(denom)}. Supported "
-            f"types are numeric scalars and pdarrays"
+            f"types are numeric scalars and pdarrays. At least one argument must be a pdarray."
         )
 
 
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -584,39 +584,54 @@ def arctan2(num: Union[numeric_scalars, pdarray], denom: Union[numeric_scalars, 
     TypeError
         Raised if the parameter is not a pdarray
     """
-    if (isinstance(num, pdarray) or isSupportedNumber(num)) and (
-        isinstance(denom, pdarray) or isSupportedNumber(denom)
-    ):
-        if isinstance(num, pdarray) and isinstance(denom, pdarray):
-            repMsg = generic_msg(
-                cmd="efunc2vv",
-                args={
-                    "func": "arctan2",
-                    "num": num,
-                    "denom": denom,
-                },
+    if isinstance(num, pdarray) and isinstance(denom, pdarray):
+        return create_pdarray(
+            type_cast(
+                str,
+                generic_msg(
+                    cmd="efunc2vv",
+                    args={
+                        "func": "arctan2",
+                        "num": num,
+                        "denom": denom,
+                    },
+                ),
             )
-        if isinstance(num, pdarray) and isSupportedNumber(denom):
-            repMsg = generic_msg(
-                cmd="efunc2vs",
-                args={
-                    "func": "arctan2",
-                    "num": num,
-                    "scalar": denom,
-                    "dtype": resolve_scalar_dtype(denom),
-                },
+        )
+    if isinstance(num, pdarray) and isSupportedNumber(denom):
+        return create_pdarray(
+            type_cast(
+                str,
+                generic_msg(
+                    cmd="efunc2vs",
+                    args={
+                        "func": "arctan2",
+                        "num": num,
+                        "scalar": denom,
+                        "dtype": resolve_scalar_dtype(denom),
+                    },
+                ),
             )
-        if isSupportedNumber(num) and isinstance(denom, pdarray):
-            repMsg = generic_msg(
-                cmd="efunc2sv",
-                args={
-                    "func": "arctan2",
-                    "scalar": num,
-                    "denom": denom,
-                    "dtype": resolve_scalar_dtype(num),
-                },
+        )
+    if isSupportedNumber(num) and isinstance(denom, pdarray):
+        return create_pdarray(
+            type_cast(
+                str,
+                generic_msg(
+                    cmd="efunc2sv",
+                    args={
+                        "func": "arctan2",
+                        "scalar": num,
+                        "denom": denom,
+                        "dtype": resolve_scalar_dtype(num),
+                    },
+                ),
             )
-        return create_pdarray(type_cast(str, repMsg))
+        )
+    if isSupportedNumber(num) and isSupportedNumber(denom):
+        raise TypeError(
+            "Scalar-scalar case is not supported. At least one argument must be a pdarray."
+        )
     else:
         raise TypeError(
             f"Unsupported types {type(num)} and/or {type(denom)}. Supported "
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -37,6 +37,19 @@ __all__ = [
     "cumprod",
     "sin",
     "cos",
+    "tan",
+    "arcsin",
+    "arccos",
+    "arctan",
+    "arctan2",
+    "sinh",
+    "cosh",
+    "tanh",
+    "arcsinh",
+    "arccosh",
+    "arctanh",
+    "rad2deg",
+    "deg2rad",
     "hash",
     "where",
     "histogram",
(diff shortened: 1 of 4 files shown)


tests/numeric_test.py
```diff
@@ -197,21 +197,335 @@ def testCumProd(self):
...
+        pda = ak.array(na)
         self.assertTrue(np.allclose(np.sin(na), ak.sin(pda).to_ndarray()))
+
         with self.assertRaises(TypeError):
-            ak.cos([range(0, 10)])
+            ak.sin([range(0, 10)])
 
     def testCos(self):
-        na = np.linspace(1, 10, 10)
+        na = np.linspace(1, 10, 10, "int64")
```
pierce314159, line 216: this isn't giving u what u want, you have to use the `dtype` keyword to set this argument like this
```python
>>> np.linspace(1, 10, 10, "int64")
array([ 1.,  2.,  3.,  4.,  5.,  6.,  7.,  8.,  9., 10.])

>>> np.linspace(1, 10, 10, "int64").dtype
dtype('float64')

>>> np.linspace(1, 10, 10, dtype="int64").dtype
dtype('int64')
```

also it might just make more sense to use `arange` since i think most ppl are more used to working with that
```python
>>>  np.arange(1, 10)
array([1, 2, 3, 4, 5, 6, 7, 8, 9])

>>> np.arange(1, 10).dtype
dtype('int64')

>>> np.arange(1, 10, dtype=np.uint).dtype
dtype('uint64')

>>> np.arange(1, 10, dtype=float).dtype
dtype('float64')
```

But as it stands right now, the tests are only testing the `float` case

src/EfuncMsg.chpl
```diff
@@ -300,6 +420,390 @@ module EfuncMsg
...
+             cmd,efunc,msgArgs.getValueOf("scalar"),dtype,name,rname));
+
+        select (dtype, denom.dtype) {
+            when (DType.Int64, DType.Int64) {
+                var val = msgArgs.get("scalar").getIntValue();
+                var ed = toSymEntry(denom, int);
+                select efunc {
+                    when "atan2" {
+                        var a = [(n) in zip(ed.a)] atan2(val,n);
+                        st.addEntry(rname, new shared SymEntry(a));
```
pierce314159, line 712: you don't have to do zippered iteration if there's only one item. Also we should a `ref`
```suggestion
                var ed = toSymEntry(denom, int);
                ref eda = ed.a;
                select efunc {
                    when "atan2" {
                        var a = [d in eda] atan2(val,d);
                        st.addEntry(rname, new shared SymEntry(a));
```
Also since we're applying this to every element i wonder how this compares to doing the forall implicitly using promotion
```chapel
var ed = toSymEntry(denom, int);
ref eda = ed.a;
select efunc {
    when "atan2" {
        st.addEntry(rname, new shared SymEntry(atan2(val, eda)));
    }
}
```

src/EfuncMsg.chpl
```diff
@@ -87,6 +87,46 @@ module EfuncMsg
                         var a = st.addEntry(rname, e.size, real);
                         a.a = cos(e.a);
                     }
+                    when "tan" {
+                        var a = st.addEntry(rname, e.size, real);
+                        a.a = tan(e.a);
```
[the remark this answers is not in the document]
pierce314159: we should do some perf comparisons, but if this performs better, we should update all the efunc methods to do this. so something like
```chapel
var e = toSymEntry(gEnt,int);
ref ea = e.a;
select efunc {
    when "abs" {
        st.addEntry(rname, new shared SymEntry(abs(ea)));
    }
    when "log" {
        st.addEntry(rname, new shared SymEntry(log(ea)));
    }
    when "exp" {
        st.addEntry(rname, new shared SymEntry(exp(ea)));
    }
    when "sin" {
        st.addEntry(rname, new shared SymEntry(sin(ea)));
    }
    when "cos" {
        st.addEntry(rname, new shared SymEntry(cos(ea)));
    }
```
because this also cuts down on lines of code. that would be outside the scope of this PR though

pierce314159 (changes requested):

src/EfuncMsg.chpl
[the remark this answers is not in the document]
Ethan-DeBandi99: I would agree with this. Generally, it is more performant to create the SymEntry from an array rather than creating an empty SymEntry and populating the array. I agree we should do performance testing, but I definitely think @pierce314159 is correct here.

arkouda/numeric.py
```diff
@@ -34,6 +34,19 @@
     "cumprod",
     "sin",
     "cos",
+    "tan",
+    "asin",
+    "acos",
+    "atan",
+    "atan2",
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Agree on this. We need to maintain consistency with the numpy API where it makes sense.

arkouda/numeric.py
```diff
@@ -34,6 +34,19 @@
...
+    "sinh",
+    "cosh",
+    "tanh",
+    "asinh",
+    "acosh",
+    "atanh",
+    "degrees",
+    "radians",
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Agree

src/EfuncMsg.chpl
jaketrookman: Updated. That makes sense, the zip is left over from the efunc3.

tests/numeric_test.py
jaketrookman: Good catch. I matched the format of another test, so assumed this was how the kwargs detected it. Went back in fixed that test too.

arkouda/numeric.py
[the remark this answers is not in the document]
jaketrookman: Updated. Makes sense, I was using the math module fxn names

src/EfuncMsg.chpl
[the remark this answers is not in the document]
jaketrookman: This makes a lot of sense. I changed all of the `efunc` fxns to this. we can create an issue and look at performance later on.

Ethan-DeBandi99 (approved):
Looks good. I was a little concerned about some of the typing based on the Chapel documentation, but I pulled it down and everything is working as far as I see. I asked @pierce314159 to double check to make sure I did not miss anything.

tests/numeric_test.py
```diff
@@ -197,21 +197,341 @@ def testCumProd(self):
...
+
+        na = np.arange(1, 10, dtype="float64")
+        pda = ak.array(na)
+        self.assertTrue(np.allclose(np.tan(na), ak.tan(pda).to_ndarray()))
+
+        with self.assertRaises(TypeError):
+            ak.tan([range(0, 10)])
+
+    def testArcsin(self):
+        na = np.arange(0, 1, 0.1, dtype="int64")
```
pierce314159, line 248: i don't feel like doing `arange` with a step of `0.1` is really what we want here
```python
In [2]: na = np.arange(0, 1, 0.1, dtype="int64")

In [3]: na
Out[3]: array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0])
```

pierce314159 (changes requested):
some small stuff mostly in the tests. The way they are set up rn, we end up with all zeros a fair amount of the time. We should also test outliers like 0, inf and -inf and stuff like that

The only other thing is i found some weird behavior with `arctan2` while testing. I'm not sure what's causing this
```python

In [16]: ak.arctan2(ak.arange(10), 2**63 + 1)
Out[16]: array([3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931 3.1415926535897931])

In [17]: np.arctan2(np.arange(10), 2**63 + 1)
Out[17]:
array([0.00000000e+00, 1.08420217e-19, 2.16840434e-19, 3.25260652e-19,
       4.33680869e-19, 5.42101086e-19, 6.50521303e-19, 7.58941521e-19,
       8.67361738e-19, 9.75781955e-19])

In [18]: np.arctan2(2**63 + 1, np.arange(10))
Out[18]:
array([1.57079633, 1.57079633, 1.57079633, 1.57079633, 1.57079633,
       1.57079633, 1.57079633, 1.57079633, 1.57079633, 1.57079633])

In [19]: ak.arctan2(2**63 + 1, ak.arange(10))
Out[19]: array([-1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966 -1.5707963267948966])
```

tests/numeric_test.py
```diff
@@ -180,7 +180,7 @@ def testCumSum(self):
         self.assertTrue(np.allclose(np.cumsum(na), ak.cumsum(pda).to_ndarray()))
 
         # Test uint case
-        na = np.linspace(1, 10, 10, "uint64")
+        na = np.linspace(1, 10, 10, dtype="uint64")
```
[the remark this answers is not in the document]
jaketrookman: Good catch

tests/numeric_test.py
jaketrookman: Updated

tests/numeric_test.py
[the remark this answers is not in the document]
jaketrookman: Updated

arkouda/numeric.py
```diff
@@ -447,7 +872,7 @@ def hash(
 
     Parameters
     ----------
-    pda : Union[pdarray, Strings, Segarray], List[Union[pdarray, Strings, Segarray]]]
+    pda : Union[pdarray, Strings, arctan2rray], List[Union[pdarray, Strings, Segarray]]]
```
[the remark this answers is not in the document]
jaketrookman: Good catch

pierce314159 (changes requested):
A few more suggested changes on the tests. I think we should go through them line by line in ipython and make sure the output at each step looks how we expect. 

Requesting changes because doing `assertListEqual` on float pdarrays can result in test failures. So we tend to use `assertTrue(np.allclose)` instead

tests/numeric_test.py
```diff
@@ -197,21 +197,341 @@ def testCumProd(self):
...
+        pda = ak.array(na)
         self.assertTrue(np.allclose(np.sin(na), ak.sin(pda).to_ndarray()))
+
         with self.assertRaises(TypeError):
-            ak.cos([range(0, 10)])
+            ak.sin([range(0, 10)])
 
     def testCos(self):
-        na = np.linspace(1, 10, 10)
+        na = np.arange(1, 10, dtype="int64")
```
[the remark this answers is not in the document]
pierce314159: i dont have a preference

tests/numeric_test.py
```diff
@@ -197,21 +197,397 @@ def testCumProd(self):
...
+            np.allclose(
+                np.arctan2(na1_float, na2_float), ak.arctan2(pda1_float, pda2_float).to_ndarray()
+            )
+        )
+
+        with self.assertRaises(TypeError):
+            ak.arctan2([range(0, 10)], [range(0, 10)])
+
+        # vector-scalar case
+        denom = np.array([5]).astype(np.uint)  # work around to get a scalar uint
```
pierce314159, line 354: just a reminder, we're prob gonna wanna update this to a more representative uint scalar (like `np.array([2**63 + 1]).astype(np.uint)`). This will need to be done as part of #2508

tests/numeric_test.py
```diff
@@ -197,21 +197,397 @@ def testCumProd(self):
...
+        pda = ak.array(na)
+        self.assertTrue(np.allclose(np.sinh(na), ak.sinh(pda).to_ndarray()))
+
+        with self.assertRaises(TypeError):
+            ak.sinh([range(0, 10)])
+
+        # Edge case: infinities
+        na = np.array([np.inf, -np.inf])
+        pda = ak.array(na)
+        self.assertTrue(np.allclose(np.sinh(na).tolist(), ak.sinh(pda).to_list(), equal_nan=True))
```
pierce314159, line 457: this is stylistic and doesn't need to be changed but i don't think u gotta call `to_list` on each of these. I think u could call `to_ndarray` on only `ak` one. i.e.
```python
self.assertTrue(np.allclose(np.sinh(na), ak.sinh(pda).to_ndarray(), equal_nan=True))
```

pierce314159 (approved):
looks good! nice work

arkouda/numeric.py
```diff
@@ -425,6 +438,418 @@ def cos(pda: pdarray) -> pdarray:
...
+                    "denom": denom,
+                    "dtype": resolve_scalar_dtype(num),
+                },
+            )
+        return create_pdarray(type_cast(str, repMsg))
+    else:
+        raise TypeError(
+            f"Unsupported types {type(num)} and/or {type(denom)}. Supported "
+            f"types are numeric scalars and pdarrays"
+        )
```
Ethan-DeBandi99, line 624: The logic here is not completely correct. The outer `if` will allow the for the cases covered inside it, but it will also allow for `isSupportedNumber(num) and isSupportedNumber(num)`. I do not believe that we want to include that case here, but we need to logic to properly error. The current workflow would try to run `create_pdarray`, but the `repMsg` variable does not exist. 

Here is an example for a single block which should be useable for each case which would be in an `if..elif..else` structure.

```python
if isinstance(num, pdarray) and isinstance(denom, pdarray):
    return create_pdarray(type_cast(str, generic_msg(
                cmd="efunc2vv",
                args={
                    "func": "arctan2",
                    "num": num,
                    "denom": denom,
                },
            )))
```

tests/numeric_test.py
```diff
@@ -197,21 +197,398 @@ def testCumProd(self):
...
+
+        with self.assertRaises(TypeError):
+            ak.arctan([range(0, 10)])
+
+        # Edge case: infinities
+        na = np.array([np.inf, -np.inf])
+        pda = ak.array(na)
+        self.assertTrue(np.allclose(np.arctan(na), ak.arctan(pda).to_ndarray(), equal_nan=True))
+
+    def testArctan2(self):
```
Ethan-DeBandi99, line 300: This test needs to include testing that the scalar-scalar case properly produces an error message. As it stands, the function logic needs to be modified to produce the expected error and the test needs to be updated to ensure that the case does in fact error.

Ethan-DeBandi99 (changes requested):
One thing with the logic that I found and related testing updates.

tests/numeric_test.py
jaketrookman: Updated tests

arkouda/numeric.py
```diff
@@ -425,6 +438,433 @@ def cos(pda: pdarray) -> pdarray:
...
+                    cmd="efunc2sv",
+                    args={
+                        "func": "arctan2",
+                        "scalar": num,
+                        "denom": denom,
+                        "dtype": resolve_scalar_dtype(num),
+                    },
+                ),
+            )
+        )
```
Ethan-DeBandi99, line 630: This should be updated to `if` then `elif`

arkouda/numeric.py
```diff
@@ -425,6 +438,433 @@ def cos(pda: pdarray) -> pdarray:
...
+    if isSupportedNumber(num) and isSupportedNumber(denom):
+        raise TypeError(
+            "Scalar-scalar case is not supported. At least one argument must be a pdarray."
+        )
```
Ethan-DeBandi99, line 634: This is essentially the same as the error in the `else` and can be removed. Combined with the `if/elif` update above, the else will then result in the desired error.

Ethan-DeBandi99 (changes requested):

arkouda/numeric.py
jaketrookman: Updated

arkouda/numeric.py
jaketrookman: I misinterpreted what your comment was asking for. Change made.

Ethan-DeBandi99 (approved):

pierce314159 (approved):

PR #2453, opened 2023-05-24
Closes #2304: `inner_join` on `Strings` and `Categorical`

### LINKED ISSUE #2304: Join on segstrings

pierce314159:
This PR (closes https://github.com/Bears-R-Us/arkouda/issues/2304) adds broadcasting (with and without permutation) for Strings and inner join on Strings and Categorical. Adds testing for all this functionality.

We can't add broadcasting to Categorical since we use `broadcast` in the init of Categorical so this causes circular dependencies. Otherwise we could just broadcast on the codes and build `from_codes`

This is built upon the work in https://github.com/Bears-R-Us/arkouda/pull/2376

I did some quick and dirty multi-locale comparisons of comm and time with the previous impl. This is a `10**4` long strings with ganset enabled and 2 locales:
```
new way time = 0

|      0 |  17 |   1 |          0 |            41 |
|      1 | 177 |  13 |          5 |             0 |
old way time = 115

| locale |    get |    put | execute_on | execute_on_nb |
| -----: | -----: | -----: | ---------: | ------------: |
|      0 | 800064 | 300000 |          0 |        600063 |
|      1 | 800203 | 100017 |     200004 |        600378 |
```

<details>
  <summary>quick python code used to generate the timings / comm</summary>

```python
In [3]: def test(i):
   ...:     TOT = 10**i
   ...:     SIZE = 10**(i-1)
   ...:     a = ak.arange(SIZE)
   ...:     segs = ak.arange(0,TOT,10)
   ...:     s = ak.cast(a, str)
   ...:     rand = ak.random_strings_uniform(1, 16, SIZE)
   ...:     s = s.stick(rand, delimiter="_")
   ...:     return ak.broadcast(segs, s, TOT)
   ...: 

In [4]: test(5)
Out[4]: array(['0_WXVTUR', '0_WXVTUR', '0_WXVTUR', ... , '9999_OZH', '9999_OZH', '9999_OZH'])
```
</details>

### DIFF
--- a/src/BroadcastMsg.chpl
+++ b/src/BroadcastMsg.chpl
@@ -9,7 +9,7 @@ module BroadcastMsg {
   use Message;
   use BigInteger;
   use SegmentedString;
-  
+
   private config const logLevel = ServerConfig.logLevel;
   private config const logChannel = ServerConfig.logChannel;
   const bmLogger = new Logger(logLevel, logChannel);
@@ -31,8 +31,7 @@ module BroadcastMsg {
                                        "TypeError");
     }
 
-    const objType = msgArgs.getValueOf("objType");
-    if objType != "pdarray" {
+    if msgArgs.getValueOf("objType").toUpper(): ObjType == ObjType.STRINGS {
       return broadcastStrings(cmd, msgArgs, st);
     }
 
@@ -129,10 +128,10 @@ module BroadcastMsg {
         }
         otherwise {
           throw new owned ErrorWithContext("Values array has unsupported dtype %s".format(gv.dtype:string),
-                                          getLineNumber(),
-                                          getRoutineName(),
-                                          getModuleName(),
-                                          "TypeError");
+                                           getLineNumber(),
+                                           getRoutineName(),
+                                           getModuleName(),
+                                           "TypeError");
         }
       }
     }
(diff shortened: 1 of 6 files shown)


src/BroadcastMsg.chpl
```diff
@@ -134,6 +140,46 @@ module BroadcastMsg {
     return new MsgTuple(repMsg, MsgType.NORMAL);    
   }
 
+  proc broadcastStrings(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
+    const gs = getGenericTypedArrayEntry(msgArgs.getValueOf("segName"), st);
+    const segs = toSymEntry(gs, int);
+    // Check that values exists (can be any dtype)
+    const sE = getSegString(msgArgs.getValueOf("valName"), st);
```
Ethan-DeBandi99, line 147: Not sure this comment fully makes sense here. This looks more like we are just accessing values, not checking existence. 

Also, calling `getSegString` with this comment is not straight forward because this would imply that values must be a `SegString` not any dtype.

@pierce314159  - Can you provide some clarification here?

Ethan-DeBandi99 (commented):
Overall, looks good. One question on a comment combined with the code that could use clarification.

src/BroadcastMsg.chpl
pierce314159: I won't lie, that comment is from the original PR and it looks like it was copied from here
https://github.com/Bears-R-Us/arkouda/blob/732c48fea6f2392b25f554dfda59b7eca10b8d6c/src/BroadcastMsg.chpl#L33-L34

I agree it doesn't need to be there! I'll go ahead and remove

Ethan-DeBandi99 (approved):

Issue #1267, opened 2022-04-12
unexpected results with GroupBy() when 2nd element is string array  


21771:
Accustomed to errors when forgetting to enclose GroupBy elements in square brackets. Surprised to find results returned when 2nd parameter is string array. 

example:
```python
a = ak.array([0, 1, 2, 1])
b = ak.array(['a', 'b', 'a', 'b'])

ak.GroupBy(a,b).unique_keys
# array([0,1,2,1])
```
Surprised at lack of error and don't understand results.

21771:
OK, now understand results. Issue appears to be that string array is being evaluated as True and interpreted as indicating that first array is ordered. Preferred behavior would be to throw an error rather than evaluate the string array as True.

PR #2456, opened 2023-05-24
Closes #2433: GroupBy back to client only

### LINKED ISSUE #2433: GroupBy Client Side only Implementation
Previously, we created a GroupBySymEntry and a GroupBy server side object to aide in future improvements. After further review of the code, this will end up presenting additional challenges and raises some concerns with maintainability of the codebase. As a result, we will be reverting to a client side only representation of the GroupBy object.

pierce314159:
This PR (closes #2433) goes back to a client only implementation of groupby


pierce314159:
the only code i know of that uses the server side groupby is some stuff @cdhollis94 has been working on. But it shouldn't be too hard to get the same functionality using `unique`. We can assist with the rewrite if need be

arkouda/groupbyclass.py
```diff
@@ -248,7 +255,7 @@ def __init__(
         # This prevents non-bool values that can be evaluated to true (ie non-empty arrays)
```
Ethan-DeBandi99, line 255: Just want to verify that this comment is still relevant since #1267 was closed.

Ethan-DeBandi99 (commented):
Looks good just want to verify that one comment is still necessary.

arkouda/groupbyclass.py
pierce314159: hmmm im not really sure, you added this comment in the PR that closed that issue
https://github.com/Bears-R-Us/arkouda/pull/1277/files

if a new dev looks at this and wonders why we are doing explicit type checking instead of using the `@typechecked`, this comment could lead them back to the issues that lead to this choice. I don't really think it's necessary. But ultimately i default to ur judgement since u added it

the full comment cause one line is missing from the diff
```
# Type Checks required because @typechecked was removed for causing other issues
# This prevents non-bool values that can be evaluated to true (ie non-empty arrays)
# from causing unexpected results. Experienced when forgetting to wrap multiple key arrays in [].
```

Ethan-DeBandi99: Yeah I think its fine. Just wanted to get your take on it.

Ethan-DeBandi99 (approved):

PR #2466, opened 2023-05-30
Closes #2454 - Update `SegArray.__getitem__` to Always Return `pdarray`

### LINKED ISSUE #2454: `SegArray.__getitem__` Returns `ndarray`. Should It?
@pierce314159 and I have had sever discussions about this, but I wanted to post a question about this to everyone. Feedback is appreciated.

Currently, `SegArray.__getitem__` returns an `ndarray`. This has always seemed a bit odd, but I want to make sure that any existing workflows are not relying on this. I would like to update this to return a `pdarray` if there are no objections.

Ethan-DeBandi99:
Closes #2454 

This PR updates `SegArray.__getitem__` to always return `pdarray` instead of `ndarray`. This prevents issues with large objects being returned.

Some testing required updates to support this change. This involved updating `ndarray.tolist()` to `pdarray.to_list()`.

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -259,7 +259,7 @@ class SegArray:
         if isSupportedInt(i):
             start = self.segments[i]
             end = self.segments[i] + self.lengths[i]
-            return self.values[start:end].to_ndarray()
+            return self.values[start:end]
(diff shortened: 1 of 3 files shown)


pierce314159 (approved):
looks good to me!

PR #2469, opened 2023-05-31
Closes #2424 - Adds `SegArray` support for `Strings` Values

### LINKED ISSUE #2424: Allow dataframe rows to contain lists of strings (i.e. segarray or similar)
**Feature Description**
Allow dataframe rows to contain lists of strings (i.e. segarray or similar). Either arkouda strings or categorical representation underneath not as important as being able to represent lists of text.

**Example Behavior**
```python
language = ak.array(['english','spanish'])
words = ak.array(['one,two,three','uno,dos,tres'])
strs, segs = words.split(',', return_segments=True)
```

### would like something like this to work and be supported in dataframes
```python
x = ak.segarray(segs, strs)
df = ak.DataFrame({'language':language, 'examples': x })
```

**Describe Workarounds**
none to my knowledge

**Additional context**
Part of request to have general import/export of lists from/to pandas.

Ethan-DeBandi99:
Closes #2424

This PR makes the necessary updates to `SegArray` to allow its values to be a `Strings` object.

Additional supporting changes also allow for set operations on 2 Strings/Categorical objects. This was previously allowed by the parameter typing, but the actual workflow would not allow for it.

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -539,7 +539,7 @@ class SegArray:
         """
         longenough, newj = self._normalize_index(j)
         ind = (self.segments + newj)[longenough]
-        if compressed or self.dtype == str_: # Strings not supported by uncompressed version
+        if compressed or self.dtype == str_:  # Strings not supported by uncompressed version
             res = self.values[ind]
         else:
             res = zeros(self.size, dtype=self.dtype) + default
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -23,6 +23,7 @@ from arkouda.pdarrayclass import (
 )
 from arkouda.pdarraycreation import arange, array, ones, zeros
 from arkouda.pdarraysetops import concatenate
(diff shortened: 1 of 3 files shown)


arkouda/segarray.py
```diff
@@ -640,6 +647,8 @@ def append_single(self, x, prepend=False):
         SegArray
             Copy of original SegArray with values from x appended to each sub-array
         """
+        if self.dtype == str_:
+            raise TypeError("String elements are immutable and cannot accept a single value")
```
pierce314159, line 651: I think we should be able to do this, but I think it's outside the scope of this PR. I'll open an issue to capture this

pierce314159 (approved):

arkouda/segarray.py
Ethan-DeBandi99: Agreed. Thanks

pierce314159: Captured in #2471

PR #2470, opened 2023-05-31
Closes #1211 - Remove TaskErrors Workaround

### LINKED ISSUE #1211: Remove TaskErrors workaround once 1.26 is required by Arkouda
In https://github.com/Bears-R-Us/arkouda/issues/1153, it was identified that `TaskErrors` from `forall` loops were not being caught by the server, so https://github.com/chapel-lang/chapel/issues/19378 was opened and https://github.com/Bears-R-Us/arkouda/pull/1188 was implemented with a short-term workaround for the issue, which was wrapping all throwing `forall` loops in a `try/catch` and re-throwing the error, which would enable the server to catch the error without crashing.

The `TaskErrors` bug was fixed for the 1.26 release, so the extra `try/catch` should no longer be needed if using a version of Chapel > 1.26, so these `try/catch` blocks should be removed once 1.26 is the oldest supported version of Arkouda https://github.com/Bears-R-Us/arkouda/blob/667abfb84ffb9772705a92a2ff86c7d5caaae1c0/src/ParquetMsg.chpl#L109-L127

Ethan-DeBandi99:
Closes #1211 

Since we are beyond 1.26 being are minimum supported version, we are able to make this change. The try/catches that are no longer needed have been removed. I also cleaned up some unneeded code that was commented out.

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -126,25 +126,22 @@ module ParquetMsg {
       var locFiledoms = subdoms;
       var locOffsets = fileOffsets;
       
-      try {
-        forall (off, filedom, filename) in zip(locOffsets, locFiledoms, locFiles) {
-          for locdom in A.localSubdomains() {
-            const intersection = domain_intersection(locdom, filedom);
+      forall (off, filedom, filename) in zip(locOffsets, locFiledoms, locFiles) {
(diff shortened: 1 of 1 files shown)


pierce314159 (approved):
good job

jaketrookman (approved):
Looks good

PR #2475, opened 2023-05-31
Closes #2459 and #2434: `ak.hash` for `Segarray` and `Strings`

### LINKED ISSUE #2434: hashing columns containing strings

### LINKED ISSUE #2459: Hashing segarray
also i know i signed off hashing the values for segarray during initial design decisions, but after playing around a bit more, I don't think this is what we want to do. I think ultimately segarrays are gonna be difficult to handle (for reasons im about to get into). I think we're gonna wanna do some brainstorming for them, but they are def outside the scope of this bug fix and we should put in a separate issue to handle them

```python
# consider this segarray
>>> sa = ak.segarray(ak.array([0,3,6]), ak.array([0,1,2,3,4,5,0,1,2]))
>>> sa
SegArray([
[0 1 2]
[3 4 5]
[0 1 2]
])

# intuitively the first and third elem should hash to the same thing since they are identical arrays

# we can't get this behavior by hashing the segments since 
# they don't capture the fact that sa[0] and sa[2] are identical
>>> sa.segments
array([0 3 6])

>>> ak.hash(sa.segments)
(array([14020738532124324502 10334800848021183012 12294710806803928155]),
 array([13660161615634858013 14420950124940837019 318151776295263894]))

# we can't hash the values cause they don't really match back up to the original segarray
>>> ak.hash(sa.values)
(array([14020738532124324502 17469661437632883388 2223077377302025035 10334800848021183012 12899154897215201871 1827545583036225533 14020738532124324502 17469661437632883388 2223077377302025035]),
 array([13660161615634858013 100355499037336788 7178333632624491744 14420950124940837019 4776206602043303594 14424044815130752497 13660161615634858013 100355499037336788 7178333632624491744]))

# we thought about hashing each sa[i] and then hashing the resulting pdarray
# but ak.hash(pdarray) returns a pdarray not a single value

# we can't just concatenate the values within a segment because how do u differentiate between 
# [1, 11, 2] and [11, 1, 2] for instance

# one absolutely awful idea that technically works... is to interpret each inner array as a string and
# then hash that (shown here using __repr__ but we'd have to cast server side because repr truncates)
>>> ak.hash(ak.array([ak.array(sa[i]).__repr__() for i in range(3)]))
(array([16660294113708371871 6932019251617944032 16660294113708371871]),
 array([10308859934083080982 18056049045803087039 10308859934083080982]))
# we see that the first and last do match, but it's such an ugly solution
# if we go this route maybe we do hex representation to shorten string lengths
```
These are just some quick thoughts, but doing this well is gonna require some thought. I vote we hold off on segarray hashing for now

_Originally posted by @pierce314159 in https://github.com/Bears-R-Us/arkouda/pull/2438#discussion_r1204584631_

pierce314159:
This PR (closes #2459 and closes #2434) adds:
- segarray hashing `sa.hash()` and `ak.hash(sa)` for pdarray and Strings values
- support for Strings in `ak.hash`
- testing for these cases

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -1,3 +1,4 @@
+import json
 from enum import Enum
 from typing import ForwardRef, List, Optional, Tuple, Union
 from typing import cast as type_cast
@@ -426,11 +427,12 @@ def cos(pda: pdarray) -> pdarray:
 
 
 def _hash_helper(a):
-    from arkouda import SegArray as _Segarray
+    from arkouda import SegArray as Segarray_
 
-    # TODO move to JSON (see issue #2464), right now we '+' as a delimiter
-    if isinstance(a, _Segarray):
-        return f"{a.segments.name}+{a.values.name}+{a.values.objType}"
+    if isinstance(a, Segarray_):
+        return json.dumps(
+            {"segments": a.segments.name, "values": a.values.name, "valObjType": a.values.objType}
+        )
     else:
         return a.name
 
@@ -487,9 +489,9 @@ def hash(
     because equivalent values will cancel each other out, hence we
     do a rotation by the ordinal of the array.
     """
-    from arkouda import SegArray as _Segarray
+    from arkouda import SegArray as Segarray_
 
-    if isinstance(pda, (pdarray, Strings, _Segarray)):
+    if isinstance(pda, (pdarray, Strings, Segarray_)):
         return _hash_single(pda, full) if isinstance(pda, pdarray) else pda.hash()
     elif isinstance(pda, List):
         types_list = [a.objType for a in pda]
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -24,6 +24,7 @@ from arkouda.pdarraycreation import array
 from arkouda.strings import Strings
 
 Categorical = ForwardRef("Categorical")
+SegArray = ForwardRef("SegArray")
 
 __all__ = [
     "cast",
@@ -41,6 +42,8 @@ __all__ = [
     "isnan",
     "ErrorMode",
 ]
+
+hashable = Union[pdarray, Strings, SegArray]  # type: ignore
 
 
 class ErrorMode(Enum):
@@ -422,16 +425,26 @@ def cos(pda: pdarray) -> pdarray:
     return create_pdarray(repMsg)
 
 
+def _hash_helper(a):
+    from arkouda import SegArray as _Segarray
+
+    # TODO move to JSON (see issue #2464), right now we '+' as a delimiter
+    if isinstance(a, _Segarray):
+        return f"{a.segments.name}+{a.values.name}+{a.values.objType}"
+    else:
+        return a.name
+
+
 @typechecked
 def hash(
-    pda: Union[pdarray, List[pdarray]], full: bool = True
+    pda: Union[hashable, List[hashable]], full: bool = True
 ) -> Union[Tuple[pdarray, pdarray], pdarray]:
     """
-    Return an element-wise hash of the array.
+    Return an element-wise hash of the array or list of arrays.
 
     Parameters
     ----------
-    pda : Union[pdarray, list[pdarray]]
+    pda : Union[pdarray, Strings, Segarray], List[Union[pdarray, Strings, Segarray]]]
 
     full : bool
         This is only used when a single pdarray is passed into hash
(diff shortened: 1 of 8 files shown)


arkouda/numeric.py
```diff
@@ -422,16 +425,26 @@ def cos(pda: pdarray) -> pdarray:
...
 
 
+def _hash_helper(a):
+    from arkouda import SegArray as _Segarray
+
+    # TODO move to JSON (see issue #2464), right now we '+' as a delimiter
+    if isinstance(a, _Segarray):
+        return f"{a.segments.name}+{a.values.name}+{a.values.objType}"
+    else:
+        return a.name
```
Ethan-DeBandi99, line 435: We have the components to handle this now. Not for complex JSON, but for a simple case like this we can use `jsonToMap` in `GenSymIO.chpl` to convert to a map. I do think limiting our tech debt moving forward would be would be advisable.

Let me know if you need an example of how to do this with the list of JSON elements.

arkouda/numeric.py
```diff
@@ -468,30 +481,38 @@ def hash(
...
-                "typeslist": [n.objType for n in pda],
-                "length": len(pda),
-                "size": len(pda[0]),
-            },
-        ),
-    )
-
-    a, b = repMsg.split("+")
-    return create_pdarray(a), create_pdarray(b)
+    from arkouda import SegArray as _Segarray
```
Ethan-DeBandi99, line 490: Just noting again that we may want switch this to `SegArray_` to align with our existing cases. Though I would be open to adding an issue to change everything to this format moving forward as well.

arkouda/segarray.py
```diff
@@ -361,6 +361,9 @@ def __eq__(self, other):
             eq[leneq] = intersection
         return eq
 
+    def __len__(self) -> int:
+        return self.segments.size
```
Ethan-DeBandi99, line 365: Should we not just return `self.size` here? It is set from `segments.size`, so the 2 should be equivalent.

Ethan-DeBandi99 (changes requested):
Logic looks good overall, just a few house keeping things to address.

arkouda/numeric.py
```diff
@@ -42,6 +43,8 @@
     "ErrorMode",
 ]
 
+hashable = Union[pdarray, Strings, SegArray]  # type: ignore
```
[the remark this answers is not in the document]
pierce314159: Uhm I actually don't know, it seems like the `@typechecked` hash function is perfectly happy with `hashable` as is, so i guess it's fine?

arkouda/numeric.py
```diff
@@ -422,16 +425,26 @@ def cos(pda: pdarray) -> pdarray:
     return create_pdarray(repMsg)
 
 
+def _hash_helper(a):
+    from arkouda import SegArray as _Segarray
```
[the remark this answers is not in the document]
pierce314159: absolutely agree with being consistent! ill get this updated

arkouda/numeric.py
pierce314159: awesome! ill add that in!

arkouda/segarray.py
pierce314159: yah that's a better idea!

arkouda/numeric.py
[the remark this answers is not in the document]
Ethan-DeBandi99: Ah ok... I see why. You have a `# type: ignore` on this line. I believe that may defeat the purpose.

Ethan-DeBandi99 (changes requested):
Just the forward ref. Otherwise looks good!

arkouda/numeric.py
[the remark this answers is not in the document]
pierce314159: ur totally right! I updated this, I had to remove the `@typechecked` so i just expanded the manual typechecking to also handle the types within a list

arkouda/numeric.py
```diff
@@ -422,16 +428,26 @@ def cos(pda: pdarray) -> pdarray:
     return create_pdarray(repMsg)
 
 
-@typechecked
+def _hash_helper(a):
```
Ethan-DeBandi99, line 431: should this  be `a: hashable`?

Ethan-DeBandi99 (approved):
I am not following why the `@typechecked` need to be removed. We use it in other places with `groupable`. Although, maybe `groupable is working differently because of the forward ref being in the `groupable_element_type`. Either way I think this is good to go.

src/SegmentedMsg.chpl
```diff
@@ -627,16 +626,30 @@ module SegmentedMsg {
             forall (h, h1, h2) in zip(hashes, hash1.a, hash2.a) {
                 (h1,h2) = h:(uint,uint);
             }
-            var repMsg = "created " + st.attrib(name1) + "+created " + st.attrib(name2);
-            smLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
-            return new MsgTuple(repMsg, MsgType.NORMAL);
+            repMsg = "created " + st.attrib(name1) + "+created " + st.attrib(name2);
+        }
+        when ObjType.SEGARRAY {
+            // check to make sure symbols defined
```
jaketrookman, line 632: In what case would it not be defined at this point?

pierce314159: if somehow the segments or values arrays were deleted accidentally (we see this happen sometimes when things go out of scope)

arkouda/numeric.py
pierce314159: sure! It's not user facing or typechecked so idk if it's completely necessary, but i don't mind making the change it could be helpful for devs who want to know what types this functions expects

Ethan-DeBandi99 (approved):

jaketrookman (approved):
Looks good, thanks for the explanation

PR #2479, opened 2023-06-01
Changes for Chapel `c_memcpy` replacement with `OS.POSIX.memcpy`

jeremiah-corrado:
Recently `c_memcpy` was deprecated in favor of `OS.POSIX.memcpy`. This PR adds a compatibility module and makes changes to a few modules to use the new symbol.

Resolves: https://github.com/Bears-R-Us/arkouda/issues/2480


bmcdonald3 (approved):
Looks good to me, ready to promote 👍 thanks Jeremiah!

src/AryUtil.chpl
```diff
@@ -9,6 +9,9 @@ module AryUtil
     use MultiTypeSymEntry;
     use ServerErrors;
     use BitOps;
+    use OS.POSIX;
+
+    use ArkoudaPOSIXCompat;
```
Ethan-DeBandi99, line 14: Do we need to import both of these? If there is a compat module, I feel like normally we just import the compat module.

Ethan-DeBandi99 (commented):
Just one question.

pierce314159 (approved):
looks good

src/AryUtil.chpl
jeremiah-corrado: I don't think both are needed. I added `OS.POSIX` to prepare for whenever the compatibility module is removed, but I can remove it for now to avoid ambiguity.

pierce314159 (approved):

PR #2484, opened 2023-06-02
Fixes #2481 - Multi-Column Parquet does not handle Empty Files Properly

### LINKED ISSUE #2481: Multi-Column Parquet does not handle Empty Files Properly
The multi-column Parquet code does handle locales that have no data on them. These locales should get an empty dataset generated, but currently they do not. The code only generates the dataset if `locDom.size>0`, but does not handle the `locDom.size=0` case. @bmcdonald3 reported that this is causing a missing dataset issue. I am seeing it manifest differently on my machine, but I believe that the issues are ultimately caused by the same thing.

Ethan-DeBandi99:
Closes #2481 

Corrects the issue with SegArray where the Object Type and Data Type was not being set properly when the local domain size on a given locale was 0. This was causing issues during the read. The workflow has been restructured to always set these so that the metadata gets set accordingly. 

I verified that the test that was failing is now passing with larger numbers of locales.

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -765,10 +765,12 @@ module ParquetMsg {
     var rnames: list((string, string, string)); // tuple (dsetName, item type, id)
     
     for (dsetidx, dsetname) in zip(dsetdom, dsetnames) do {
+        writeln("\n\n Getting Type Info For %s\n\n".format(dsetname));
+        types[dsetidx] = getArrType(filenames[0], dsetname);
         for (i, fname) in zip(filedom, filenames) {
             var hadError = false;
             try {
-                types[dsetidx] = getArrType(fname, dsetname);
+                // types[dsetidx] = getArrType(fname, dsetname);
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good to me

src/ParquetMsg.chpl
```diff
@@ -765,10 +765,11 @@ module ParquetMsg {
     var rnames: list((string, string, string)); // tuple (dsetName, item type, id)
     
     for (dsetidx, dsetname) in zip(dsetdom, dsetnames) do {
+        types[dsetidx] = getArrType(filenames[0], dsetname);
         for (i, fname) in zip(filedom, filenames) {
             var hadError = false;
             try {
-                types[dsetidx] = getArrType(fname, dsetname);
+                // types[dsetidx] = getArrType(fname, dsetname);
```
pierce314159, line 772:
```suggestion

```

pierce314159 (approved):

PR #2494, opened 2023-06-09
Closes #2476: Updates Chapel Tutorial

### LINKED ISSUE #2476: Update the chapel tutorial
It's been a while since we first wrote the chapel tutorial, it's probably time for a revisit. Some things that come up a good deal in arkouda that aren't covered right now:

- const, ref (when to use them, how they help)
- diagnostics (comm and timing)
- coforall loops (when to use vs forall)
- mem management (borrowed, owned, shared)
- records/classes
- maybe a section on common pitfalls that will kill performance

pierce314159:
This PR (closes #2476) adds updates to chapel tutorial to cover topics often used in arkouda that were missed in the first iteration. Namely `coforall`s, locality, and performance/diagnostics

### DIFF
--- a/training/CHAPEL_TUTORIAL.md
+++ b/training/CHAPEL_TUTORIAL.md
@@ -567,7 +567,8 @@ Expected Output:
 [Locales](https://chapel-lang.org/docs/users-guide/locality/localesInChapel.html)
 can be thought of as chunk of memory that can do computation. Things that are co-located within a single locale
 are close to each other in the system and can interact with one another relatively cheaply. Things that are in distinct
-locales can still interact with each other in the same ways, but this is more expensive.
+locales can still interact with each other in the same ways, but this is more expensive since transferring data between
+the locales will result in more communication.
 
 Say `x` and `y` are both on `locale_i`. When on `locale_i`, we say both `x` and `y` are local.
 
--- a/training/CHAPEL_TUTORIAL.md
+++ b/training/CHAPEL_TUTORIAL.md
@@ -20,28 +20,39 @@ This guide will familiarize new developers with Chapel concepts regularly used i
    3. [Example 8: `forall` Expressions](#ex8)
    4. [Example 9: Factorial with may-parallel `forall` Expression and Reduction](#ex9)
    5. [Try It Yourself 1: Perfect Squares <=25](#TIY1)
-6. [Ternary](#ternary)
-   1. [Example 10: Absolute Value Ternary](#ex10)
-   2. [Example 11: Ternary and `forall` Expression](#ex11)
+6. [`coforall` Loops](#coforall)
+   1. [Example 10: `coforall` loop](#ex10)
+7. [Ternary](#ternary)
+   1. [Example 11: Absolute Value Ternary](#ex11)
+   2. [Example 12: Ternary and `forall` Expression](#ex12)
    3. [Try It Yourself 2: Array Absolute Value](#TIY2)
-7. [Introspection](#introspection)
-   1. [Example 12: Introspection](#ex12)
-8. [Promotion](#promotion)
-   1. [Example 13: Promotion](#ex13)
-9. [Zippered Iteration and Aggregation](#zip_and_agg)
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Great changes and updates.

training/CHAPEL_TUTORIAL.md
```diff
@@ -354,6 +365,44 @@ Expected Output:
...
 ```
+<a id="coforall"></a>
+## `Coforall` loops
+The second most common parallel loop in arkouda is the [`coforall`](https://chapel-lang.org/docs/users-guide/taskpar/coforall.html) loop.
+The biggest difference between a `coforall` and [`forall`](https://chapel-lang.org/docs/users-guide/datapar/forall.html) loop
+is the way tasks are scheduled.
+
+A `coforall` loop creates one distinct task per loop iteration, each of which executes a copy of the loop body.
+
+But a `forall` loop is a bit more complicated. A forall-loop creates a variable number of tasks to execute the loop
```
Ethan-DeBandi99, line 376: I think it may be better to explain this in the `forall` section and then refer back to it here. We introduce the `forall` and give examples, but don't really explain it in that section.

training/CHAPEL_TUTORIAL.md
```diff
@@ -482,6 +531,109 @@ Promotion is equivalent to a `forall` loop. For this example the equivalent loop
...
+```
+
+<a id="locality"></a>
+## Locality
+<a id="locale"></a>
+### Locales
+[Locales](https://chapel-lang.org/docs/users-guide/locality/localesInChapel.html)
+can be thought of as chunk of memory that can do computation. Things that are co-located within a single locale
+are close to each other in the system and can interact with one another relatively cheaply. Things that are in distinct
+locales can still interact with each other in the same ways, but this is more expensive.
```
Ethan-DeBandi99, line 570: I would add mention that the expense is due to the locales having to communicate/transfer data. Otherwise, if someone is not familiar with distributed computing this will seem extremely vague.

training/CHAPEL_TUTORIAL.md
```diff
@@ -586,7 +738,7 @@ Nice! We'll use `getLengths`, `values`, and `offsets` in the next section as wel
 
 [Aggregation](https://chapel-lang.org/docs/modules/packages/CopyAggregation.html#copyaggregation) is used to copy
 local values into a remote array (`DstAggregator`) or copy values from a remote array into local variables (`SrcAggregator`).
-This provides a significant speed up when doing batch assignments.
+This provides a significant speed-up and lower communication when doing batch assignments.
```
Ethan-DeBandi99, line 741:
```suggestion
This provides a significant speed-up and reduces communication when doing batch assignments.
```

Ethan-DeBandi99 (commented):
Overall, looks good. Few small things I wanted to bring attention to, but ultimately nothing to hold up on.

training/CHAPEL_TUTORIAL.md
pierce314159: i feel like getting into how tasks are created in that section will be confusing if we don't have simpler context of `coforall` first. The chapel docs linked to in the forall section go into detail about this (so more experienced readers can explore more, but less experienced readers can wait for the context provided by `coforall`)

training/CHAPEL_TUTORIAL.md
pierce314159: yah this is a bit handwavy, ill update this

training/CHAPEL_TUTORIAL.md
pierce314159: spoke a bit with ethan about this. We debated switching the order to introduce coforall first, but think it might be putting the wagon before the horse to jump right into talking about task creation. It might be more approachable to talk about an abstract parallel loop first before we can get into the weeds about how things are actually implemented.

We think we're gonna have @TheOddCzar789 take a look and see which he thinks is more intuitive

Ethan-DeBandi99 (approved):

PR #2498, opened 2023-06-13
Closes #2053: Strings Property Accessors

### LINKED ISSUE #2053: Strings Property Accessors
Ability to access the `bytes` and `offsets` properties directly for Strings Objects.

If we do not want to directly expose, create `_bytes` and `_offsets` property definitions that are not directly exposed.

jaketrookman:
This PR (Closes #2053) allows access the bytes and offsets properties directly for Strings Objects and adds testing.

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -18,7 +18,7 @@ from arkouda.dtypes import (
     str_scalars,
     translate_np_dtype,
 )
-from arkouda.infoclass import information
+from arkouda.infoclass import information, list_registry, list_symbol_table
 from arkouda.logger import getArkoudaLogger
 from arkouda.match import Match, MatchType
 from arkouda.pdarrayclass import (
@@ -26,6 +26,7 @@ from arkouda.pdarrayclass import (
     parse_single_value,
     pdarray,
     unregister_pdarray_by_name,
+    attach_pdarray,
 )
 
 __all__ = ["Strings"]
@@ -183,6 +184,8 @@ class Strings:
         except Exception as e:
             raise ValueError(e)
 
+        self._bytes: Optional[pdarray] = None
+        self._offsets: Optional[pdarray] = None
         self.dtype = npstr
         self._regex_dict: Dict = dict()
         self.logger = getArkoudaLogger(name=__class__.__name__)  # type: ignore
@@ -353,6 +356,70 @@ class Strings:
         return create_pdarray(
             generic_msg(cmd="segmentLengths", args={"objType": self.objType, "obj": self.entry})
         )
+
+    def get_bytes(self):
+        """
+        Getter for the bytes component (uint8 pdarray) of this Strings.
+
+        Returns
+        -------
+        pdarray, uint8
+            Pdarray of bytes of the string accessed
+
+        Example
+        -------
+        >>> x = ak.array(['one', 'two', 'three'])
+        >>> x.get_bytes()
+        [111 110 101 0 116 119 111 0 116 104 114 101 101 0]
+        """
+        bytes_name = self.name + "_bytes"
+        registry = list_registry()
+
+        if self._bytes is None and bytes_name in registry:
+            self._bytes = attach_pdarray(bytes_name)
+
+        if self._bytes is None or self._bytes.name not in list_symbol_table():
+            self._bytes = create_pdarray(
+                generic_msg(
+                    cmd="getSegStringProperty", args={"property": "get_bytes", "obj": self.entry}
+                )
+            )
+
+        if self.is_registered() and self._bytes.name not in registry:
+            self._bytes.register(self.name + "_bytes")
+        return self._bytes
+
+    def get_offsets(self):
+        """
+        Getter for the offsets component (int64 pdarray) of this Strings.
+
+        Returns
+        -------
+        pdarray, int64
+            Pdarray of offsets of the string accessed
+
+        Example
+        -------
+        >>> x = ak.array(['one', 'two', 'three'])
+        >>> x.get_offsets()
+        [0 4 8]
+        """
+        offsets_name = self.name + "_offsets"
+        registry = list_registry()
+
+        if self._offsets is None and offsets_name in registry:
+            self._offsets = attach_pdarray(offsets_name)
+
+        if self._offsets is None or self._offsets.name not in list_symbol_table():
+            self._offsets = create_pdarray(
+                generic_msg(
+                    cmd="getSegStringProperty", args={"property": "get_offsets", "obj": self.entry}
+                )
+            )
+
+        if self.is_registered() and self._offsets.name not in registry:
+            self._offsets.register(self.name + "_offsets")
+        return self._offsets
 
     def encode(self, toEncoding: str, fromEncoding: str = "UTF-8"):
         """
@@ -2074,7 +2141,12 @@ class Strings:
           the file name is checked for _LOCALE#### to determine if it is distributed.
         - If the dataset provided does not exist, it will be added
         """
-        from arkouda.io import _mode_str_to_int, _file_type_to_int, _get_hdf_filetype, _repack_hdf
+        from arkouda.io import (
+            _file_type_to_int,
+            _get_hdf_filetype,
+            _mode_str_to_int,
+            _repack_hdf,
+        )
 
         # determine the format (single/distribute) that the file was saved in
         file_type = _get_hdf_filetype(prefix_path + "*")
@@ -2343,6 +2415,10 @@ class Strings:
         """
         self.entry.register(user_defined_name)
         self.name = user_defined_name
+        if self._bytes is not None:
+            self._bytes.register(self.name + "_bytes")
+        if self._offsets is not None:
+            self._offsets.register(self.name + "_offsets")
         return self
 
     def unregister(self) -> None:
@@ -2373,6 +2449,10 @@ class Strings:
         """
         self.entry.unregister()
         self.name = None
+        if self._bytes is not None:
+            self._bytes.unregister()
+        if self._offsets is not None:
+            self._offsets.unregister()
 
     @staticmethod
     @typechecked
@@ -2405,9 +2485,17 @@ class Strings:
         Registered names/Strings objects in the server are immune to deletion
         until they are unregistered.
         """
+        from arkouda.pdarrayclass import attach_pdarray
+
         rep_msg: str = cast(str, generic_msg(cmd="attach", args={"name": user_defined_name}))
         s = Strings.from_return_msg(rep_msg)
         s.name = user_defined_name
+        registry = list_registry()
+        bytes_name, offsets_name = f"{user_defined_name}_bytes", f"{user_defined_name}_offsets"
+        if bytes_name in registry:
+            s._bytes = attach_pdarray(bytes_name)
+        if offsets_name in registry:
+            s._offsets = attach_pdarray(offsets_name)
         return s
 
     @staticmethod
@@ -2426,3 +2514,9 @@ class Strings:
         register, unregister, attach, is_registered
         """
         unregister_pdarray_by_name(user_defined_name)
+        registry = list_registry()
+        bytes_name, offsets_name = f"{user_defined_name}_bytes", f"{user_defined_name}_offsets"
+        if bytes_name in registry:
+            unregister_pdarray_by_name(bytes_name)
+        if offsets_name in registry:
+            unregister_pdarray_by_name(offsets_name)
--- a/arkouda/util.py
+++ b/arkouda/util.py
@@ -6,7 +6,7 @@ from arkouda.categorical import Categorical
 from arkouda.client import generic_msg, get_config, get_mem_used
 from arkouda.client_dtypes import BitVector, BitVectorizer, IPv4
 from arkouda.groupbyclass import GroupBy, broadcast
-from arkouda.infoclass import list_symbol_table
+from arkouda.infoclass import list_registry, list_symbol_table
 from arkouda.pdarrayclass import RegistrationError, create_pdarray
 from arkouda.pdarraycreation import arange
 from arkouda.pdarraysetops import unique
@@ -248,7 +248,18 @@ def attach(name: str, dtype: str = "infer"):
         dtype = repMsg.split()[2]
         repMsg = repMsg[len(repType) + 1 :]
         if dtype == "str":
-            return Strings.from_return_msg(repMsg)
+            s = Strings.from_return_msg(repMsg)
+            registry = list_registry()
+            bytes_name, offsets_name = f"{name}_bytes", f"{name}_offsets"
+            if bytes_name in registry:
+                s._bytes = create_pdarray(
+                    cast(str, generic_msg(cmd="attach", args={"name": bytes_name}))
+                )
+            if offsets_name in registry:
+                s._offsets = create_pdarray(
+                    cast(str, generic_msg(cmd="attach", args={"name": offsets_name}))
+                )
+            return s
         else:
             return create_pdarray(repMsg)
     else:
--- a/src/SegmentedMsg.chpl
+++ b/src/SegmentedMsg.chpl
@@ -170,6 +170,40 @@ module SegmentedMsg {
     return new MsgTuple(repMsg, MsgType.NORMAL);
   }
 
+  proc getSegStringPropertyMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
+    var pn = Reflection.getRoutineName();
+    const property = msgArgs.getValueOf("property");
+    const name = msgArgs.getValueOf("obj");
+
+    var genSym = toGenSymEntry(st.lookup(name));
+
+    if genSym.dtype != DType.Strings{
+      var errorMsg = notImplementedError(pn, "%s".format(dtype2str(genSym.dtype)));
+      smLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);                      
+      return new MsgTuple(errorMsg, MsgType.ERROR);
+    }
+    
+    var ssentry = toSegStringSymEntry(genSym);
+    var rname = st.nextName();
+    select property{
+      when "get_bytes" {
+        st.addEntry(rname, ssentry.bytesEntry);
+      }
+      when "get_offsets" {
+        st.addEntry(rname, ssentry.offsetsEntry);
+      }
+      otherwise {
+        var errorMsg = notImplementedError(pn,property);
+        smLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);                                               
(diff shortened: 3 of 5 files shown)


tests/registration_test.py
```diff
@@ -533,6 +533,12 @@ def test_strings_groupby_attach(self):
         self.assertIsInstance(sAttach.segments, ak.pdarray)
         self.assertIsInstance(sAttach.unique_keys, ak.Strings)
 
+    def test_string_property_accessor(self):
+        s = ak.array(["abc", "123", "abc"])
+        print()
+        print(s.bytes.dtype)
+        print(s.offsets.dtype)
+
```
jaketrookman, line 541: Not sure why my updated tests did not get added here. Update coming.

hokiegeek2 (approved):
looks good, just need to run `mypy arkouda` and `flake8 arkouda` from the project root dir and fix the corresponding errors

src/SegmentedMsg.chpl
```diff
@@ -170,6 +170,40 @@ module SegmentedMsg {
...
+    
+    var ssentry = toSegStringSymEntry(genSym);
+    var rname = st.nextName();
+    select property{
+      when ("mybytes"){
+        st.addEntry(rname, ssentry.bytesEntry);
+      }
+      when ("myoffsets"){
+        st.addEntry(rname, ssentry.offsetsEntry);
+      }
```
pierce314159, line 194: this is completely stylistic and there's no need to change but u don't need the parens here

tests/registration_test.py
```diff
@@ -533,6 +533,52 @@ def test_strings_groupby_attach(self):
...
+        soffsets = s.myoffsets
+        self.assertListEqual(soffsets.to_list(), offsets_answers.to_list())
+        self.assertTrue(len(ak.list_registry()) == 3)
+
+        s.unregister()
+        self.assertTrue(len(ak.list_registry()) == 0)
+
+
+        # Tests for unregistering each property
```
pierce314159, line 561: you don't need two new lines here, if u run `black` on this file and it should reformat it correctly

tests/registration_test.py
```diff
@@ -533,6 +533,52 @@ def test_strings_groupby_attach(self):
         self.assertIsInstance(sAttach.segments, ak.pdarray)
         self.assertIsInstance(sAttach.unique_keys, ak.Strings)
 
+    def test_string_property_accessor(self):
```
pierce314159, line 536: it might be worth adding a test where u attach to make sure that functionality is working correctly

pierce314159 (changes requested):
Most of the logic looks good! I offered a slight reworking to avoid duplicating the accessors

arkouda/strings.py
```diff
@@ -183,6 +183,8 @@ def __init__(self, strings_pdarray: pdarray, bytes_size: int_scalars) -> None:
         except Exception as e:
             raise ValueError(e)
 
+        self._mybytes: Optional[pdarray] = None
+        self._myoffsets: Optional[pdarray] = None
```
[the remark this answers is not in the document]
Ethan-DeBandi99: @pierce314159 - if we leave the property accessors named as they are yes. However, I am not sure that is the naming scheme we want to use here.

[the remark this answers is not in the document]
pierce314159: yeah i agree with u, another thing to consider is using `values` instead? I know there are sections of the chpl code where we refer to that attribute of a segstring that way. But also know that naming collisions with segarray are thing to keep in mind

tests/registration_test.py
jaketrookman: I would appreciate you taking a look at the testing I added to see if this covers all the cases we need.

tests/registration_test.py
jaketrookman: Updated

src/SegmentedMsg.chpl
jaketrookman: Updated

arkouda/strings.py
```diff
@@ -354,6 +356,60 @@ def get_lengths(self) -> pdarray:
...
+        """
+        from arkouda.infoclass import list_registry
+
+        if (self._mybytes is None) or (self._mybytes.name not in list_registry()):
+            self._mybytes = create_pdarray(
+                generic_msg(cmd="getSegStringProperty", args={"property": "mybytes", "obj": self.entry})
+            )
+            if self.is_registered():
+                self._mybytes.register(self.name + "_mybytes")
+        return self._mybytes
```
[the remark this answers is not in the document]
jaketrookman: Updated

arkouda/strings.py
```diff
@@ -354,6 +356,60 @@ def get_lengths(self) -> pdarray:
             generic_msg(cmd="segmentLengths", args={"objType": self.objType, "obj": self.entry})
         )
 
+    @property
+    def mybytes(self):
+        """
+        Property accessor that returns a pdarray of bytes (uint8) of the corresponding accessed string.
```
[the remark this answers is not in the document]
jaketrookman: Updated

arkouda/strings.py
```diff
@@ -354,6 +356,60 @@ def get_lengths(self) -> pdarray:
...
+        pdarray, uint8
+            Pdarray of bytes of the string accessed
+
+        Example
+        -------
+        >>> x = ak.array(['one', 'two', 'three'])
+        >>> x.mybytes
+        [111 110 101 0 116 119 111 0 116 104 114 101 101 0]
+        """
+        from arkouda.infoclass import list_registry
```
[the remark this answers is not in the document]
jaketrookman: Updated

arkouda/strings.py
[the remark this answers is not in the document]
jaketrookman: Updated

TheOddCzar789 (commented):
**DISREGARD all of the below, this was my fault.**

make test failed for me  
```
______________________ RegistrationTest.test_string_property_accessor ______________________
self = <registration_test.RegistrationTest testMethod=test_string_property_accessor>

    def test_string_property_accessor(self):
        cleanup()
        s = ak.array(["one", "two", "three", "123"])
        s.register("sname")
        ak.clear()
        sAttach = ak.Strings.attach("sname")
        self.assertListEqual(s.to_list(), sAttach.to_list())
>       s.get_bytes()

tests/registration_test.py:543:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
arkouda/strings.py:384: in get_bytes
    generic_msg(
arkouda/client.py:871: in generic_msg
    return cast(Channel, channel).send_string_message(cmd=cmd, args=msg_args,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
```

arkouda/strings.py
```diff
@@ -354,6 +357,72 @@ def get_lengths(self) -> pdarray:
...
+        pdarray, int64
+            Pdarray of offsets of the string accessed
+
+        Example
+        -------
+        >>> x = ak.array(['one', 'two', 'three'])
+        >>> x.get_offsets()
+        [0 4 8]
+        """
+        offsets_name = self.name + "_get_offsets"
```
Ethan-DeBandi99, line 409: Again, I don't think we want the `get` in the name here.

arkouda/strings.py
```diff
@@ -2343,6 +2417,10 @@ def register(self, user_defined_name: str) -> Strings:
         """
         self.entry.register(user_defined_name)
         self.name = user_defined_name
+        if self._get_bytes is not None:
+            self._get_bytes.register(self.name + "_get_bytes")
+        if self._get_offsets is not None:
+            self._get_offsets.register(self.name + "_get_offsets")
```
Ethan-DeBandi99, line 2423: just noting so we don't miss it here as well that we don't need `get` in the name.

tests/registration_test.py
```diff
@@ -533,6 +533,114 @@ def test_strings_groupby_attach(self):
...
```
Ethan-DeBandi99, line 536: As a general note, I think it would be good to add some comments to indicate what each section is intended to test. It makes coming back to things like this were there are a lot of cases easier to identify what is going on.

tests/registration_test.py
```diff
@@ -533,6 +533,114 @@ def test_strings_groupby_attach(self):
...
+        cleanup()
+        s = ak.array(["one", "two", "three", "123"])
+        s.register("sname")
+        ak.clear()
+        s.get_bytes()
+        s.get_offsets()
+        sAttach = ak.Strings.attach("sname")
+        self.assertListEqual(s.to_list(), sAttach.to_list())
+        self.assertListEqual(s.get_bytes().to_list(), sAttach.get_bytes().to_list())
+        self.assertListEqual(s.get_offsets().to_list(), sAttach.get_offsets().to_list())
```
Ethan-DeBandi99, line 561: In my opinion, these 2 blocks are essentially testing the same thing, I think we can simplify it a bit to just one block. I made a few modifications to my suggestion. Adds validation that the property is no longer `None` after calling the get methods.

```suggestion
b = s.get_bytes()
self.assertTrue(s._bytes is not None)
o = s.get_offsets()
self.assertTrue(s._offsets is not None)
self.assertListEqual(s.to_list(), sAttach.to_list())
self.assertListEqual(s.get_bytes().to_list(), sAttach.get_bytes().to_list())
self.assertListEqual(s.get_offsets().to_list(), sAttach.get_offsets().to_list())
```

tests/registration_test.py
```diff
@@ -533,6 +533,114 @@ def test_strings_groupby_attach(self):
...
+        ak.clear()
+        s.get_offsets()
+        sAttach = ak.Strings.attach("sname")
+        self.assertListEqual(s.to_list(), sAttach.to_list())
+        self.assertListEqual(s.get_bytes().to_list(), sAttach.get_bytes().to_list())
+        self.assertListEqual(s.get_offsets().to_list(), sAttach.get_offsets().to_list())
+        s.get_bytes()
+        self.assertListEqual(s.to_list(), sAttach.to_list())
+        self.assertListEqual(s.get_bytes().to_list(), sAttach.get_bytes().to_list())
+        self.assertListEqual(s.get_offsets().to_list(), sAttach.get_offsets().to_list())
```
Ethan-DeBandi99, line 575: Again, I think this block effectively tests the same thing as the previous 2. If we want to maintain it to ensure that the calling order does not present any issues, I would suggest updating to the format from my previous comment.

tests/registration_test.py
```diff
@@ -533,6 +533,114 @@ def test_strings_groupby_attach(self):
...
+        self.assertListEqual(sbytes.to_list(), bytes_answers.to_list())
+        self.assertTrue(len(ak.list_registry()) == 2)
+
+        soffsets = s.get_offsets()
+        offsets_answers = ak.array([0, 4, 8, 14])
+        self.assertListEqual(soffsets.to_list(), offsets_answers.to_list())
+        self.assertTrue(len(ak.list_registry()) == 3)
+
+        s.unregister()
+        self.assertTrue(len(ak.list_registry()) == 0)
```
Ethan-DeBandi99, line 612: We should add a `cleanup()` call here.

tests/registration_test.py
```diff
@@ -533,6 +533,114 @@ def test_strings_groupby_attach(self):
...
+        # Tests for unregistering each property
+        s = ak.array(["one", "two", "three", "123"])
+        s.register("sname")
+        sbytes = s.get_bytes()
+        soffsets = s.get_offsets()
+        self.assertTrue(len(ak.list_registry()) == 3)
+        sbytes.unregister()
+        self.assertTrue(len(ak.list_registry()) == 2)
+        soffsets.unregister()
+        self.assertTrue(len(ak.list_registry()) == 1)
```
Ethan-DeBandi99, line 623: I would also check that in addition to the length of the registry, we check that the registered name is not in `list_registry`

arkouda/strings.py
Ethan-DeBandi99, line 1: One issue that I think is a problem with all the registration code is that when an object is registered the `user_defined_name` is set as `self.name` If the object is unregistered, this can cause issues because the symbol table name will no longer be available.

I think to correct this we need to add a `registered_name=None` property in the init that is set when an object is registered and reverted to `None` on calls to `unregister`. 

This issue will cause problems within the logic of `get_bytes()` and `get_offsets()`, but it is probably outside the scope of this PR. @pierce314159 what are your thoughts? I am thinking we should add an issue to track this and update for all types since it is not an issue specific to Strings.

Ethan-DeBandi99 (changes requested):
Logic looks good. Just a few things to adjust before this is ready to merge. One thing that will cause problems is probably outside the scope of this PR and should be handled as a separate issue, but want to get @pierce314159 thoughts on that before creating the issue.

arkouda/strings.py
pierce314159: yah i think adding and issue and working in separate is the way to go

Ethan-DeBandi99: @pierce314159 / @jaketrookman - can one of you submit the issue for this and post the issue # in this thread?

Ethan-DeBandi99 (approved):

arkouda/strings.py
pierce314159: Captured in issue #2571

arkouda/strings.py
```diff
@@ -354,6 +357,72 @@ def get_lengths(self) -> pdarray:
             generic_msg(cmd="segmentLengths", args={"objType": self.objType, "obj": self.entry})
         )
 
+    def get_bytes(self):
+        """
+        Property accessor that returns a pdarray of bytes (uint8) of the
+        corresponding accessed Strings or SegString.
```
pierce314159, line 363: Not super important but this doc string isn't accurate anymore. Since this a getter not a property accessor now. Also we don't have to use such vague language of `corresponding accessed Strings` because this called on a specfic Strings (i.e. `my_str.get_bytes`) so we know it's this Strings that we want the bytes component of. Does that make sense?
```suggestion
        Getter for the bytes component (uint8 pdarray) of this Strings.
```

arkouda/strings.py
```diff
@@ -354,6 +357,72 @@ def get_lengths(self) -> pdarray:
...
+            )
+
+        if self.is_registered() and self._bytes.name not in registry:
+            self._bytes.register(self.name + "_bytes")
+        return self._bytes
+
+    def get_offsets(self):
+        """
+        Property accessor that returns a pdarray of offsets (int64) of the
+        corresponding accessed Strings or SegString.
```
pierce314159, line 396:
```suggestion
        Getter for the offsets component (int64 pdarray) of this Strings.
```

arkouda/util.py
```diff
@@ -225,6 +225,8 @@ def attach(name: str, dtype: str = "infer"):
     Attaches to a known element name. If a type is passed, the server will use that type
     to pull the corresponding parts, otherwise the server will try to infer the type
     """
+    from arkouda.infoclass import list_registry
+
```
pierce314159, line 229: Is this here for circular dependency reasons? if not we should move it to the top with the rest of the imports

arkouda/util.py
```diff
@@ -248,7 +250,17 @@ def attach(name: str, dtype: str = "infer"):
...
+            s = Strings.from_return_msg(repMsg)
+            registry = list_registry()
+            if f"{name}_bytes" in registry:
+                s._bytes = create_pdarray(
+                    cast(str, generic_msg(cmd="attach", args={"name": name + "_bytes"}))
+                )
+            if f"{name}_offsets" in registry:
+                s._offsets = create_pdarray(
+                    cast(str, generic_msg(cmd="attach", args={"name": name + "_offsets"}))
+                )
```
pierce314159, line 262: I don't love us double defining the name for bytes/offsets in 2 places and 2 different ways (one with f strings and one with adding strings. I don't think see any issues but it seems error prone (one gets updated but not the other). Maybe do something like
```python
registry = list_registry()
bytes_name, offsets_name = f"{name}_bytes", f"{name}_offsets"
if bytes_name in registry:
    s._bytes = create_pdarray(
        cast(str, generic_msg(cmd="attach", args={"name": bytes_name}))
    )
if offsets_name in registry:
    s._offsets = create_pdarray(
        cast(str, generic_msg(cmd="attach", args={"name": offsets_name}))
    )
```

pierce314159 (commented):
very close, i didn't see anything major

TheOddCzar789 (approved):

arkouda/strings.py
```diff
@@ -359,8 +359,7 @@ def get_lengths(self) -> pdarray:
 
     def get_bytes(self):
         """
-        Property accessor that returns a pdarray of bytes (uint8) of the
-        corresponding accessed Strings or SegString.
+        Getter for the offsets component (int64 pdarray) of this Strings.
```
pierce314159, line 362: you changed both to be the docstring of `offsets`
```suggestion
        Getter for the bytes component (uint8 pdarray) of this Strings.
```

arkouda/strings.py
```diff
@@ -2405,9 +2485,16 @@ class method to return a Strings object attached to the registered name in the a
...
+        from arkouda.pdarrayclass import attach_pdarray
+
         rep_msg: str = cast(str, generic_msg(cmd="attach", args={"name": user_defined_name}))
         s = Strings.from_return_msg(rep_msg)
         s.name = user_defined_name
+        registry = list_registry()
+        if f"{user_defined_name}_bytes" in registry:
+            s._bytes = attach_pdarray(user_defined_name + "_bytes")
+        if f"{user_defined_name}_offsets" in registry:
+            s._offsets = attach_pdarray(user_defined_name + "_offsets")
```
pierce314159, line 2497: Same argument to be made here as `util/attach` from my last review. We're double defining the name for bytes/offsets in 2 places and 2 different ways

arkouda/strings.py
```diff
@@ -2426,3 +2513,8 @@ def unregister_strings_by_name(user_defined_name: str) -> None:
         register, unregister, attach, is_registered
         """
         unregister_pdarray_by_name(user_defined_name)
+        registry = list_registry()
+        if f"{user_defined_name}_bytes" in registry:
+            unregister_pdarray_by_name(user_defined_name + "_bytes")
+        if f"{user_defined_name}_offsets" in registry:
+            unregister_pdarray_by_name(user_defined_name + "_offsets")
```
pierce314159, line 2520: Same argument to be made here as `util/attach` from my last review. We're double defining the name for bytes/offsets in 2 places and 2 different ways

pierce314159 (changes requested):

arkouda/strings.py
jaketrookman: Updated

arkouda/strings.py
jaketrookman: Updated

pierce314159 (approved):
Looks good!

PR #2507, opened 2023-06-15
Fixes #2506 - Categorical Optional Components Required Bug

### LINKED ISSUE #2506: 05.05 unable to load saved dataframes from previous versions containing Categorical types
Dataframes saved to disk with ak.DataFrame.to_hdf() with Arkouda versions earlier than 05.05 AND containing ak.Categorical types are unable to be loaded by version 05.05. This issue appears to result from arkouda/io.py in the _dict_recombine_segarrays_categoricals() function that expects all categoricals to have the .permutation key set. 

KeyError: 'appIP_ipowner.permutation'

Attempted to workaround by reloading each dataset individually and reconstructing the dataframe, which almost works, but causes a journey into the layers of inception to figure out what codes actually go with what categories and reconstructing the NA value appropriately. 

Dataframes saved from and loaded into version 05.05 seem to work fine and this appears to only be a backward compatibility issue. I'm assuming this relates to segarray implementations in 05.05, but the on disk storage of the arrays also appears different as 05.05 does not appear to create the ._akNAcode, .categories, and .codes files as before. 

If possible, would like backwards compatibility to allow users to load existing stored dataframes.

Ethan-DeBandi99:
Closes #2506 

Updates `_dict_recombine_segarrays_categoricals()` to properly reconstruct `Categorical` objects from  the components available. Since `.permutation` and `.segments` are not required, there is no guarantee they can be used to recreate the object. This is now handled by checking for these components existence and if they are not available, calling `ak.Categorical.from_codes` with `permutation=None` and `segments=None`. If these components are read from the file, they will be used.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -498,8 +498,10 @@ def _dict_recombine_segarrays_categoricals(df_dict):
             else Categorical.from_codes(
                 df_dict[f"{col}.codes"],
                 df_dict[f"{col}.categories"],
-                permutation=df_dict[f"{col}.permutation"],
-                segments=df_dict[f"{col}.segments"],
+                permutation=df_dict[f"{col}.permutation"]
+                if f"{col}.permutation" in df_dict_keys
+                else None,
+                segments=df_dict[f"{col}.segments"] if f"{col}.segments" in df_dict_keys else None,
                 _akNAcode=df_dict[f"{col}._akNAcode"],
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good to me

pierce314159 (approved):
Looks good! Nice work!

PR #2511, opened 2023-06-16
Closes #2504 - Adds New Testing Architecture

### LINKED ISSUE #2504: New Testing Architecture
Configure the framework for the updates to the testing architecture.

- pytest.ini
- conftest.py

These will most likely need to be kept separate, but during testing should be named so that they are the default to ensure proper functionality when we switch over.

Ethan-DeBandi99:
Closes #2504 

Adds new testing architecture for updates to the testing framework. `PROTO_tests` directory and `pytest_PROTO.ini` contain the components required for the new framework. As configured, there will be no interference with running the existing testing in CI. The new framework allows for more customization at runtime of the testing to ensure that we have more coverage. Once the tests are updated, the updated benchmarks that were previously created will be integrated so that we will only need to maintain a single set of configuration files. This format will also support output of JSON logs for testing with the pytest-json-report plugin. This dependency has been added to the `arkouda-env-dev.yml` file.

`numeric_tests.py` has been updated to work with the new framework as an example for updating the other test files.


PROTO_tests/tests/numeric_test.py
```diff
@@ -0,0 +1,362 @@
+import arkouda as ak
+import pytest
+import numpy as np
+
+from arkouda.dtypes import npstr
+
+NUMERIC_TYPES = [ak.int64, ak.float64, ak.bool, ak.uint64]
+CAST_TYPES = [ak.int64, ak.float64, ak.bool, ak.uint64, npstr]
```
pierce314159, line 8: `CAST_TYPES` seems to be parametrization of this inner loop in `testCast`
```python
for t2 in ak.DTypes:
    t2 = ak.dtype(t2)
```

But the inner loop seems to cover more types than those listed, most notably `bigint` is missing
```python
In [4]: for t2 in ak.DTypes:
   ...:     t2 = ak.dtype(t2)
   ...:     print(t2)
   ...:
bigint
int64
float64
uint64
<U0
bool
float64
uint8
int64

In [5]: ak.DTypes
Out[5]:
frozenset({'bigint',
           'bool',
           'float',
           'float64',
           'int',
           'int64',
           'str',
           'uint64',
           'uint8'})
```

PROTO_tests/tests/numeric_test.py
```diff
@@ -0,0 +1,362 @@
...
+    def test_cast(self, prob_size, cast_to):
+        arrays = {
+            ak.int64: ak.randint(-(2 ** 48), 2 ** 48, prob_size),
+            ak.float64: ak.randint(0, 1, prob_size, dtype=ak.float64),
+            ak.bool: ak.randint(0, 2, prob_size, dtype=ak.bool),
+        }
+
+        for t1, orig in arrays.items():
+            other = ak.cast(orig, cast_to)
+            assert orig.size == other.size
```
pierce314159, line 88: once we update `CAST_TYPES` to include `bigint`, we'll need to add back in something like this check
```suggestion
        for t1, orig in arrays.items():
            if t1 == ak.float64 and cast_to == ak.bigint:
                # we don't support casting a float to a bigint
                continue
            other = ak.cast(orig, cast_to)
            assert orig.size == other.size
```

PROTO_tests/tests/numeric_test.py
```diff
@@ -0,0 +1,362 @@
+import arkouda as ak
+import pytest
+import numpy as np
+
+from arkouda.dtypes import npstr
+
+NUMERIC_TYPES = [ak.int64, ak.float64, ak.bool, ak.uint64]
+CAST_TYPES = [ak.int64, ak.float64, ak.bool, ak.uint64, npstr]
+
+VALID_CAST = [
```
pierce314159, line 10: i don't love the name `VALID_CAST` since these aren't the only valid casts, I think `roundtripable` was more descriptive because it's basically saying that these type combos can be casts both way (i.e. `int -> float` or `float -> int` both work)

PROTO_tests/tests/numeric_test.py
```diff
@@ -0,0 +1,362 @@
...
+        assert np.allclose(np.cumsum(na), ak.cumsum(pda).to_ndarray())
+
+        # Test uint case
+        na = np.linspace(1, 10, 10, "uint64")
+        pda = ak.cast(pda, ak.uint64)
+
+        assert np.allclose(np.cumsum(na), ak.cumsum(pda).to_ndarray())
+
+        with pytest.raises(TypeError):
+            ak.cumsum([range(0, 10)])
```
pierce314159, line 185: definitely this one (and likely some of the others) could be parameterized on `NUMERIC_TYPES`. I'm not sure if we wanna do that in this pass or not

maybe something like:
```python
@pytest.mark.parametrize("num_type", NUMERIC_TYPES)
def test_cumsum(self):
    na = np.arange(1, 10, dtype=num_type)
    pda = ak.array(na)

    assert np.allclose(np.cumsum(na), ak.cumsum(pda).to_ndarray())
```

pierce314159 (changes requested):

PROTO_tests/tests/numeric_test.py
```diff
@@ -0,0 +1,362 @@
...
+
+        # Make sure unseeded runs differ
+        a = ak.randint(0, 2**32, prob_size, dtype=numeric_type)
+        b = ak.randint(0, 2**32, prob_size, dtype=numeric_type)
+        assert not ((a == b).all())
+
+        # Make sure seeded results are same
+        a = ak.randint(0, 2**32, prob_size, dtype=numeric_type, seed=seed)
+        b = ak.randint(0, 2**32, prob_size, dtype=numeric_type, seed=seed)
+        assert a.to_list() == b.to_list()
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I didn't even think about this. I definitely think we should update to `ak.all` where we are using dynamic sizing. We can easily alter this with a safe problem size if needed for debugging.

PROTO_tests/conftest.py
```diff
@@ -0,0 +1,100 @@
...
+            print(
+                "Started arkouda_server in TEST_CLASS mode with "
+                "host: {} port: {} locales: {}".format(server, port, pytest.nl)
+            )
+        except Exception as e:
+            raise RuntimeError(
+                "in configuring or starting the arkouda_server: {}, check "
+                + "environment and/or arkouda_server installation",
+                e,
+            )
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Updated to us `f''`

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
Ethan-DeBandi99: Updated applicable location to use `.all()`.

PROTO_tests/tests/numeric_test.py
Ethan-DeBandi99: Updated to use the list comprehension @pierce314159 noted above.

PROTO_tests/tests/numeric_test.py
Ethan-DeBandi99: Added the check

PROTO_tests/tests/numeric_test.py
Ethan-DeBandi99: Renamed to `ROUNDTRIP_CAST`

PROTO_tests/tests/numeric_test.py
Ethan-DeBandi99: I parameterized the other functions I thought made sense. I added a few other lists to use based on the typing that was valid for the functions being tested.

Ethan-DeBandi99:
Adding a note the PR #2447 adds new tests to the `numeric_test` file. When that is merged, the new testing will need to be updated. That can be done as part of this PR or as a separate PR.

PROTO_tests/tests/numeric_test.py
```diff
@@ -121,8 +128,9 @@ def test_str_cast_errors(self, num_type):
         assert valid.to_list() == validans.to_list()
         assert np.allclose(ans, res.to_ndarray(), equal_nan=True)
 
-    def test_histogram(self):
-        pda = ak.randint(10, 30, 40)
+    @pytest.mark.parametrize("num_type", INT_FLOAT)
+    def test_histogram(self, num_type):
```
pierce314159, line 132: kinda crazy to me that histogram isn't implemented for `uint`. I feel like we should have that. I just added issue #2521 for this, we can update this test to be `NO_BOOL` as part of that

pierce314159 (approved):
looks good to me!

PR #2523, opened 2023-06-23
Closes #2493 - Multi-Column SegArray with String Values

### LINKED ISSUE #2493: Multi-Column SegArray with String Values Parquet Support
PR #2492 adds SegArray with String value support in the single column case. Support for this case needs to be added to the multi-column case as well.

Ethan-DeBandi99:
Closes #2493 

Adds support for SegArray with String values to the Multi-Column processing for Parquet files. Adds testing to ensure proper functionality. Also, updates some variable names to make things more readable.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -256,17 +256,18 @@ int64_t cpp_getStringColumnNumBytes(const char* filename, const char* colname, v
         while (ba_reader->HasNext() && numRead < numElems) {
           parquet::ByteArray value;
           (void)ba_reader->ReadBatch(1, &definition_level, nullptr, &value, &values_read);
-          if(values_read > 0) {
-            offsets[i] = value.len + 1;
-            byteSize += value.len + 1;
-            numRead += values_read;
-          } else {
-            offsets[i] = 1;
-            byteSize+=1;
-            numRead+=1;
+          if ((ty == ARROWLIST && definition_level == 3) || ty == ARROWSTRING) {
+            if(values_read > 0) {
+              offsets[i] = value.len + 1;
+              byteSize += value.len + 1;
+              numRead += values_read;
+            } else {
+              offsets[i] = 1;
+              byteSize+=1;
+              numRead+=1;
+            }
+            i++;
           }
-          i++;
-
         }
       }
       return byteSize;
@@ -552,7 +553,6 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
         int16_t definition_level; // needed for any type that is nullable
 
         std::shared_ptr<parquet::ColumnReader> column_reader = row_group_reader->Column(idx);
-
         if(lty == ARROWINT64 || lty == ARROWUINT64) {
           auto chpl_ptr = (int64_t*)chpl_arr;
           parquet::Int64Reader* reader =
@@ -592,13 +592,13 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
             parquet::ByteArray value;
             (void)reader->ReadBatch(1, &definition_level, nullptr, &value, &values_read);
             // if values_read is 0, that means that it was a null value
-            if(values_read > 0) {
+            if(values_read > 0 && definition_level == 3) {
               for(int j = 0; j < value.len; j++) {
                 chpl_ptr[i] = value.ptr[j];
                 i++;
               }
+              i++; // skip one space so the strings are null terminated with a 0
             }
-            i++; // skip one space so the strings are null terminated with a 0
           }
         } else if(lty == ARROWBOOLEAN) {
           auto chpl_ptr = (bool*)chpl_arr;
@@ -1110,7 +1110,6 @@ int cpp_writeMultiColToParquet(const char* filename, void* column_names,
                 segSize = offset_ptr[offIdx+1] - offset_ptr[offIdx];
               }
               if (segSize > 0) {
-
                 for (int64_t s=0; s<segSize; s++) {
                   int16_t def_lvl = 3;
                   int16_t rep_lvl = (s == 0) ? 0 : 1;
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -862,7 +862,13 @@ std::shared_ptr<parquet::schema::GroupNode> SetupSchema(void* column_names, void
         fields.push_back(parquet::schema::PrimitiveNode::Make(cname_ptr[i], parquet::Repetition::REQUIRED, parquet::Type::DOUBLE, parquet::ConvertedType::NONE));
       }
     } else if(dtypes_ptr[i] == ARROWSTRING) {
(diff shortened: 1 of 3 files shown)


tests/parquet_test.py
```diff
@@ -577,6 +580,7 @@ def test_segarray_string(self):
         words = ak.array(['one,two,three', 'uno,dos,tres'])
         strs, segs = words.split(',', return_segments=True)
         x = ak.SegArray(segs, strs)
+        # x = ak.SegArray(ak.array([0, 0, 2]), ak.array(["a", "b", "c", "d"]))
```
jaketrookman, line 583: Does this still need to be here?

src/ParquetMsg.chpl
```diff
@@ -1304,22 +1304,41 @@ module ParquetMsg {
...
-          
+
+          segment_ct[i] += locDom.size;
+          if values.dtype == DType.Strings && locDom.size > 0 {
+            var e: SegStringSymEntry = toSegStringSymEntry(values);
+            var segStr = new SegString("", e);
+            ref ss = segStr;
+            var lens = ss.getLengths();
+            const lastOffset = if sa.size == 0 then 0 else sa[high];
+            const lastOffsetIdx = segStr.offsets.a.domain.high;
```
jaketrookman, line 1315: I think I am able to unpack everything except the `.domain` part of this. Can i get a quick explanation of why you need it after calling the array part.

tests/parquet_test.py
Ethan-DeBandi99: Nope it was left over from some extra confirmation I was doing. It has been removed.

src/ParquetMsg.chpl
Ethan-DeBandi99: The `.domain` is needed so that we can access the largest index of the offsets array. This is used in the case when the last offset value is the start of the last segment on the locale. Accessing this allows for localization of all offsets contained in any segments on the current locale.

jaketrookman: Thank you for the explanation. Looks good

src/ArrowFunctions.cpp
```diff
@@ -1078,31 +1084,88 @@ int cpp_writeMultiColToParquet(const char* filename, void* column_names,
...
+              idxQueue_str.pop();
+            }
+            
+            auto offset_ptr = (int64_t*)offset_arr[i];
+            int64_t offIdx = 0; // index into offsets
+
+            if (x > 0){
+              offIdx = idxQueue_segarray.front();
+              idxQueue_segarray.pop();
+            }
```
pierce314159, line 1102: could we combine the two `if (x > 0)` cases to set `byteIdx` and `offIdx` at the same time?
```chapel
auto offset_ptr = (int64_t*)offset_arr[i];
int64_t byteIdx = 0;
int64_t offIdx = 0; // index into offsets
// identify the starting byte and offset index
if (x > 0){
  byteIdx = idxQueue_str.front();
  idxQueue_str.pop();

  offIdx = idxQueue_segarray.front();
  idxQueue_segarray.pop();
}
```

Also i point this out in the push case as well, but should we be doing `byteIdx = idxQueue_str.front();` or `byteIdx = idxQueue_segarray.front();`

src/ArrowFunctions.cpp
```diff
@@ -1078,31 +1084,88 @@ int cpp_writeMultiColToParquet(const char* filename, void* column_names,
...
+                int16_t* def_lvl = new int16_t[segSize] { 1 };
+                int16_t* rep_lvl = new int16_t[segSize] { 0 };
+                ba_writer->WriteBatch(segSize, def_lvl, rep_lvl, nullptr);
+              }
+              offIdx++;
+              count++;
+            }
+            if (numLeft - count > 0) {
+              idxQueue_segarray.push(byteIdx);
+              idxQueue_segarray.push(offIdx);
```
pierce314159, line 1139: i noticed we are pushing `byteIdx` onto `idxQueue_segarray` here, but above we are popping it from `idxQueue_str`. I just wanted to double check that this is intentional

src/ParquetMsg.chpl
```diff
@@ -1304,22 +1304,41 @@ module ParquetMsg {
...
+            var startOffsetIdx = sa[locDom.low];
+            var endOffsetIdx = if (lastOffset == sa[locDom.high]) then lastOffsetIdx else sa[locDom.high + 1] - 1;
+            var offIdxRange = startOffsetIdx..endOffsetIdx;
+            var str_bytes: int;
+            for i in offIdxRange do str_bytes += lens[i];
```
pierce314159, line 1320: it probably doesn't matter (and we do it in other places) but for clarity maybe we should change the looping variable of this `for` loop to be different from the outer `forall` loop. They both use `i` and it threw me off for a sec. It's up to you though, I don't think this should affect the functionality at all

tests/parquet_test.py
```diff
@@ -515,13 +515,16 @@ def test_multicol_write(self):
             "c_6": ak.SegArray(ak.array([0, 5, 10]), ak.randint(0, 1, 15, dtype=ak.bool)),
             "c_7": ak.array(np.random.uniform(0, 100, 3)),
             "c_8": ak.SegArray(ak.array([0, 9, 14]), ak.array(np.random.uniform(0, 100, 20))),
-            "c_9": ak.array(["abc", "123", "xyz"])
+            "c_9": ak.array(["abc", "123", "xyz"]),
+            "c_10": ak.SegArray(ak.array([0, 2, 5]), ak.array(["a", "b", "c", "d", "e", "f", "g", "h", "i"])),
+            "c_11": ak.SegArray(ak.array([0, 2, 2]), ak.array(["a", "b", "c", "d", "e", "f", "g", "h", "i"])),
+            "c_12": ak.SegArray(ak.array([0, 0, 2]), ak.array(["a", "b", "c", "d", "e", "f", "g", "h", "i"])),
+            "c_13": ak.SegArray(ak.array([0, 5, 8]), ak.array(["a", "b", "c", "d", "e", "f", "g", "h", "i"])),
```
pierce314159, line 522: should we maybe update one of these to have strings that are longer than 1 character to be sure that functions as expected?

src/ArrowFunctions.cpp
Ethan-DeBandi99: Oh yeah, good idea. I think I was trying to keep them separate in my head while adding this to make sure I could track everything, but definitely no need for 2 `if` blocks.

Also, good catch on the `idxQueue` stuff. The `byteIdx` value should be pushed to `idxQueue_str` and the `offIdx` to `idxQueue_segarray`. This has been updated as well.

src/ArrowFunctions.cpp
Ethan-DeBandi99: Yes, `byteIdx` should be being pushed to`idxQueue_str`. Updated.

tests/parquet_test.py
Ethan-DeBandi99: Probably would not hurt. I added an extra column to test this.

pierce314159 (commented):
looks good other than one commented out line that can be removed

pierce314159 (approved):

jaketrookman (approved):

PR #2525, opened 2023-06-25
2524 extend arkouda client to enable alternative channel implementations

### LINKED ISSUE #2524: Extend Arkouda client to enable alternative Channel implementations
**Feature Description**
The arkouda client [Channel](https://github.com/Bears-R-Us/arkouda/blob/c5eb42f48c0f91e389b09d808f9d33e315975421/arkouda/client.py#L150) is currently set by a factory method, which does not allow for alternative Channel implementations that are not within the Arkouda codebase. The proposed change is as follows:

1. Add a Channel parameter named access_channel to the client [connect](https://github.com/Bears-R-Us/arkouda/blob/c5eb42f48c0f91e389b09d808f9d33e315975421/arkouda/client.py#L330) method that defaults to None
2. If the access_channel parameter is None, create the default ZmqChannel object via the client _get_channel method and set the global channel object. Otherwise, set the global channel reference to the access_channel parameter.
4. Add default server and port values to Channel init method to better foster passing in alternative Channel implementations

**Example Behavior**
```
import arkouda as ak
from arkouda_proxy_client.client import GrpcChannel
channel = GrpcChannel(connect_url='localhost:50053', user='kjyost')
ak.connect(access_channel=channel)
```

image

**Describe Workarounds**
None, as there is no way to specify a Channel object that is not created by the client _get_channel factory method

**Additional context**
The conn

hokiegeek2:
Closes #2524

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -166,21 +166,21 @@ class Channel():
     """
     __slots__ = ('url', 'user', 'token', 'logger')
 
-    def __init__(self, user: str, server: str='localhost', port: int=5555, token: str = None,
+    def __init__(self, user: str, server: str = 'localhost', port: int = 5555, token: str = None,
                  connect_url: str = None) -> None:
         '''
         user : str
             Arkouda user who will use the Channel to connect to the arkouda_server
         server : str, optional
-            The hostname of the server (must be visible to the current machine). 
+            The hostname of the server (must be visible to the current machine).
             Defaults to `localhost`.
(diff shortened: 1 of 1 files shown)


hokiegeek2:
this demonstrates both the ZmqChannel and GrpcChannel-configured clients connecting to arkouda_server and arkouda_proxy_server, respectively:

image

Ethan-DeBandi99 (approved):
Looks good

PR #2527, opened 2023-06-27
Closes #2529: Adjust getUserMetrics to return borrowed

### LINKED ISSUE #2529: Adjust getUserMetrics to return borrowed
Fix Arkouda compilation failures after the Chapel PR https://github.com/chapel-lang/chapel/pull/21894 .

In particular, there is a change in behavior for statements like this:

`var userMetrics : borrowed CounterTable = this.getUserMetrics(...);`
because after that PR, the 'shared' result of `getUserMetrics` is destroyed at the end of the statement, before 'userMetrics' can be used.

`getUserMetrics` is changed to return a 'borrowed CounterTable' which seems to better match how it is used in other places in the module and addresses the issue.

mppf:
This PR is intended to fix Arkouda compilation failures after the Chapel PR https://github.com/chapel-lang/chapel/pull/21894 .

In particular, there is a change in behavior for statements like this:

     var userMetrics : borrowed CounterTable = this.getUserMetrics(...);

because after that PR, the 'shared' result of getUserMetrics is destroyed at the end of the statement, before 'userMetrics' can be used.

This PR changes getUserMetrics to return a 'borrowed CounterTable' which seems to better match how it is used in other places in the module and addresses the issue.

Closes #2529

- [x] make test-python passes, except for io_test.py, which I think reflects a configuration issue on my end

### DIFF
--- a/src/MetricsMsg.chpl
+++ b/src/MetricsMsg.chpl
@@ -114,9 +114,9 @@ module MetricsMsg {
         var metrics = new map(keyType=User,valType=shared CounterTable);
         var users = new Users();
 
-        proc getUserMetrics(user: User) {
+        proc getUserMetrics(user: User) : borrowed CounterTable {
             if this.metrics.contains(user: User) {
-              return try! this.metrics[user];
+                return try! this.metrics[user];
             } else {
                 var userMetrics = new shared CounterTable();
                 this.metrics.add(user, userMetrics);


bmcdonald3 (approved):
Looks good, thanks Michael. I agree that switching to `borrowed` makes sense, but I do worry that a subtle change like this could be difficult to understand for someone without context for the change, given the error message of `error: Illegal use of dead value`, but I don't feel strongly enough to push that back to a design discussion or anything like that.

pierce314159 (approved):
Seems reasonable to me, but i would like @hokiegeek2's two cents since the userMetrics stuff is in his domain

Ethan-DeBandi99 (approved):

hokiegeek2 (approved):
@pierce314159 thanks for the heads-up! I think the error message as pointed out by @bmcdonald3 is confusing, so I recommend that it be updated in a future chapel release.

PR #2531, opened 2023-06-28
Closes #2526: Add arc/hyperbolic Tests to new test framework

### LINKED ISSUE #2526: Add arc/hyperbolic Tests to new test framework
The tests that were added in PR #2447 need to be added to `PROTO_tests/tests/numeric_tests.py`. This file was converted as an example for converting other tests, but we need to make sure we don't lose coverage from the tests added after the conversion.

jaketrookman:
This PR (Closes #2526)
Testing implemented to match #2400 in the new testing framework in `numeric_test.py`.

### DIFF
--- a/PROTO_tests/tests/numeric_test.py
+++ b/PROTO_tests/tests/numeric_test.py
@@ -283,7 +283,9 @@ class TestNumeric:
             ak.arctan2(
                 np.array([range(0, 10)]).astype(num_type), np.array([range(10, 20)]).astype(num_type)
             )
+        with pytest.raises(TypeError):
             ak.arctan2(num[0], np.array([range(10, 20)]).astype(num_type))
+        with pytest.raises(TypeError):
             ak.arctan2(np.array([range(0, 10)]).astype(num_type), denom[0])
 
     @pytest.mark.parametrize("num_type", NO_BOOL)
--- a/PROTO_tests/tests/numeric_test.py
+++ b/PROTO_tests/tests/numeric_test.py
@@ -18,6 +18,7 @@ NP_TRIG_ARRAYS = {
             np.array([np.nan, -np.inf, -0.0, 0.0, np.inf]),
         ]
     ),
+    ak.bool: np.arange(10) % 2 == 0,
     ak.uint64: np.arange(2**64 - 10, 2**64, dtype=np.uint64),
 }
 
@@ -231,43 +232,21 @@ class TestNumeric:
 
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_arcsin(self, num_type):
-        if num_type == ak.uint64:
-            na = np.arange(0, 2).astype(num_type)
-        elif num_type == ak.float64:
-            na = np.linspace(-1, 1)
-        else:
-            na = np.arange(-1, 2).astype(num_type)
+        na = NP_TRIG_ARRAYS[num_type]
         pda = ak.array(na, dtype=num_type)
 
         assert np.allclose(np.arcsin(na), ak.arcsin(pda).to_ndarray(), equal_nan=True)
         with pytest.raises(TypeError):
             ak.arcsin(np.array([range(0, 10)]).astype(num_type))
 
-        # Edge case: Infinities and NaNs
-        if num_type == ak.float64:
-            na = np.array([np.inf, np.nan, -np.inf])
-            pda = ak.array(na)
-            assert np.allclose(np.arcsin(na), ak.arcsin(pda).to_ndarray(), equal_nan=True)
-
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_arccos(self, num_type):
-        if num_type == ak.uint64:
-            na = np.arange(0, 2).astype(num_type)
-        elif num_type == ak.float64:
-            na = np.linspace(-1, 1)
-        else:
-            na = np.arange(-1, 2).astype(num_type)
+        na = NP_TRIG_ARRAYS[num_type]
         pda = ak.array(na, dtype=num_type)
 
         assert np.allclose(np.arccos(na), ak.arccos(pda).to_ndarray(), equal_nan=True)
         with pytest.raises(TypeError):
             ak.arccos(np.array([range(0, 10)]).astype(num_type))
-
-        # Edge case: Infinities and NaNs
-        if num_type == ak.float64:
-            na = np.array([np.inf, np.nan, -np.inf])
-            pda = ak.array(na)
-            assert np.allclose(np.arccos(na), ak.arccos(pda).to_ndarray(), equal_nan=True)
 
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_arctan(self, num_type):
@@ -278,17 +257,11 @@ class TestNumeric:
         with pytest.raises(TypeError):
             ak.arctan(np.array([range(0, 10)]).astype(num_type))
 
-        # Edge case: infinities and NaNs
-        if num_type == ak.float64:
-            na = np.array([np.inf, np.nan, -np.inf])
-            pda = ak.array(na)
-            assert np.allclose(np.arctan(na), ak.arctan(pda).to_ndarray(), equal_nan=True)
-
     @pytest.mark.parametrize("num_type", NO_BOOL)
     @pytest.mark.parametrize("denom_type", NO_BOOL)
     def test_arctan2(self, num_type, denom_type):
-        na_num = NP_TRIG_ARRAYS[num_type]
-        na_denom = np.arange(0, 10).astype(denom_type)
+        na_num = np.random.permutation(NP_TRIG_ARRAYS[num_type])
+        na_denom = np.random.permutation(NP_TRIG_ARRAYS[denom_type])
 
         pda_num = ak.array(na_num, dtype=num_type)
         pda_denom = ak.array(na_denom, dtype=denom_type)
@@ -313,24 +286,6 @@ class TestNumeric:
             ak.arctan2(num[0], np.array([range(10, 20)]).astype(num_type))
             ak.arctan2(np.array([range(0, 10)]).astype(num_type), denom[0])
 
-        # Edge case: Infinities
-        na_inf = np.array([np.inf, np.nan, -np.inf])
-        pda_inf = ak.array(na_inf)
-        na_fin = np.array([0, 5, 10]).astype(denom_type)
-        pda_fin = ak.array(na_fin)
-        assert np.allclose(
-            np.arctan2(na_inf, na_fin), ak.arctan2(pda_inf, pda_fin).to_ndarray(), equal_nan=True
-        )
-        assert np.allclose(
-            np.arctan2(na_fin, na_inf), ak.arctan2(pda_fin, pda_inf).to_ndarray(), equal_nan=True
-        )
-        assert np.allclose(
-            np.arctan2(na_inf, denom[0]), ak.arctan2(pda_inf, denom[0]).to_ndarray(), equal_nan=True
-        )
-        assert np.allclose(
-            np.arctan2(num[0], na_inf), ak.arctan2(num[0], pda_inf).to_ndarray(), equal_nan=True
-        )
-
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_sinh(self, num_type):
         na = NP_TRIG_ARRAYS[num_type]
@@ -339,12 +294,6 @@ class TestNumeric:
         assert np.allclose(np.sinh(na), ak.sinh(pda).to_ndarray(), equal_nan=True)
         with pytest.raises(TypeError):
             ak.sinh(np.array([range(0, 10)]).astype(num_type))
-
-        # Edge case: infinities and NaNs
-        if num_type == ak.float64:
-            na = np.array([np.inf, np.nan, -np.inf])
-            pda = ak.array(na)
-            assert np.allclose(np.sinh(na), ak.sinh(pda).to_ndarray(), equal_nan=True)
 
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_cosh(self, num_type):
@@ -355,12 +304,6 @@ class TestNumeric:
         with pytest.raises(TypeError):
             ak.cosh(np.array([range(0, 10)]).astype(num_type))
 
-        # Edge case: infinities and NaNs
-        if num_type == ak.float64:
-            na = np.array([np.inf, np.nan, -np.inf])
-            pda = ak.array(na)
-            assert np.allclose(np.cosh(na), ak.cosh(pda).to_ndarray(), equal_nan=True)
-
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_tanh(self, num_type):
         na = NP_TRIG_ARRAYS[num_type]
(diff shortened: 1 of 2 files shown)


PROTO_tests/tests/numeric_test.py
```diff
@@ -197,7 +193,10 @@ def test_cumprod(self, num_type):
...
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_sin(self, num_type):
-        na = np.linspace(1, 10, 10).astype(num_type)
+        if num_type == ak.float64:
+            na = np.linspace(0, 10)
+        else:
+            na = np.arange(0, 10).astype(num_type)
         pda = ak.array(na, dtype=num_type)
 
         assert np.allclose(np.sin(na), ak.sin(pda).to_ndarray())
```
pierce314159, line 202: I feel like we should update all the `np.allclose` to have `equal_nan=True`. This doesn't seem to factor in for these examples, but would if we add the edge cases at some point

PROTO_tests/tests/numeric_test.py
```diff
@@ -206,13 +205,242 @@ def test_sin(self, num_type):
...
 
+    @pytest.mark.parametrize("num_type", NO_BOOL)
+    def test_tan(self, num_type):
+        if num_type == ak.float64:
+            na = np.linspace(0, 10)
+        else:
+            na = np.arange(0, 10).astype(num_type)
+        pda = ak.array(na, dtype=num_type)
+
+        assert np.allclose(np.tan(na), ak.tan(pda).to_ndarray())
```
pierce314159, line 226: I feel like we should update all the `np.allclose` to have `equal_nan=True`. This doesn't seem to factor in for these examples, but would if we add the edge cases at some point

PROTO_tests/tests/numeric_test.py
```diff
@@ -206,13 +205,242 @@ def test_sin(self, num_type):
...
+            pda_num = ak.array(na_num, dtype=num_type)
+        else:
+            na_num = np.arange(10, 0, -1).astype(num_type)
+            pda_num = ak.array(na_num, dtype=num_type)
+        if denom_type == ak.float64:
+            na_denom = np.linspace(0, 10, 10)
+            pda_denom = ak.array(na_denom, dtype=denom_type)
+        else:
+            na_denom = np.arange(0, 10).astype(denom_type)
+            pda_denom = ak.array(na_denom, dtype=denom_type)
```
pierce314159, line 291: `pda_num/denom` is the same in all these cases, so we should be able to pull it out.
```python
if num_type == ak.float64:
    na_num = np.linspace(0, 1, 10)
else:
    na_num = np.arange(10, 0, -1).astype(num_type)
if denom_type == ak.float64:
    na_denom = np.linspace(0, 10, 10)
else:
    na_denom = np.arange(0, 10).astype(denom_type)
pda_num = ak.array(na_num, dtype=num_type)
pda_denom = ak.array(na_denom, dtype=denom_type)
```

We could also turn the others into ternaries, but im not sure if that's easier to read or not
```python
na_num = np.linspace(0, 1, 10) if num_type == ak.float64 else np.arange(10, 0, -1).astype(num_type)
na_denom = np.linspace(0, 10, 10) if denom_type == ak.float64 else np.arange(0, 10).astype(denom_type)
pda_num = ak.array(na_num, dtype=num_type)
pda_denom = ak.array(na_denom, dtype=denom_type)
```

PROTO_tests/tests/numeric_test.py
```diff
@@ -206,13 +205,242 @@ def test_sin(self, num_type):
...
+        assert np.allclose(np.arctan2(num[0], na_denom), ak.arctan2(num[0], pda_denom).to_ndarray())
+        assert np.allclose(np.arctan2(na_num, denom[0]), ak.arctan2(pda_num, denom[0]).to_ndarray())
+
+        with pytest.raises(TypeError):
+            ak.arctan2(
+                np.array([range(0, 10)]).astype(num_type), np.array([range(10, 20)]).astype(num_type))
+            ak.arctan2(num[0], np.array([range(10, 20)]).astype(num_type))
+            ak.arctan2(np.array([range(0, 10)]).astype(num_type), denom[0])
+
+        # Edge case: Infinities and Zeros
```
pierce314159, line 306: we say `Infinities and Zeros` but i don't think we're testing any zeros

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,7 +193,10 @@ def test_cumprod(self, num_type):
 
     @pytest.mark.parametrize("num_type", NO_BOOL)
     def test_sin(self, num_type):
-        na = np.linspace(1, 10, 10).astype(num_type)
+        if num_type == ak.float64:
+            na = np.linspace(0, 10)
+        else:
+            na = np.arange(0, 10).astype(num_type)
```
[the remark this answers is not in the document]
Ethan-DeBandi99: @pierce314159 - I think this is a great idea. 

@jaketrookman - can we make sure to apply this to all tests in the file where applicable?

PROTO_tests/tests/numeric_test.py
jaketrookman: Updated

PROTO_tests/tests/numeric_test.py
jaketrookman: Updated

PROTO_tests/tests/numeric_test.py
jaketrookman: Updated

PROTO_tests/tests/numeric_test.py
jaketrookman: Updated, the zeros are tested in the main test block

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
jaketrookman: Updated, I was able to get a majority of the trig functions to this format. I did not see very many other places where this structure could be used on the other functions, but definitely looks like a good thing to implement the new testing structure.

TheOddCzar789 (approved):

PROTO_tests/tests/numeric_test.py
```diff
@@ -10,6 +10,12 @@
 INT_FLOAT = [ak.int64, ak.float64]
 CAST_TYPES = [ak.dtype(t) for t in ak.DTypes]
 
+NP_TRIG_ARRAYS = {
+    ak.int64: np.arange(0, 10, dtype=np.int64),
+    ak.float64: np.linspace(0, 10),
+    ak.uint64: np.arange(0, 10, dtype=np.uint64),
+}
```
pierce314159, line 17: this doesn't pick as good of representative examples as I was looking for. We don't get any negative ints or floats, uint's never go into the `> 2**63` range, and we don't test any float edge cases

PROTO_tests/tests/numeric_test.py
pierce314159: we still need to update this one to use `equal_nan=True`

PROTO_tests/tests/numeric_test.py
pierce314159: we still need to update this one to use `equal_nan=True`

pierce314159 (commented):
Some the issues from my last review are still lingering

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +204,249 @@ def test_cumprod(self, num_type):
...
+        if num_type == ak.float64:
+            na = np.array([np.inf, np.nan, -np.inf])
+            pda = ak.array(na)
+            assert np.allclose(np.arctan(na), ak.arctan(pda).to_ndarray(), equal_nan=True)
+
+    @pytest.mark.parametrize("num_type", NO_BOOL)
+    @pytest.mark.parametrize("denom_type", NO_BOOL)
+    def test_arctan2(self, num_type, denom_type):
+        na_num = NP_TRIG_ARRAYS[num_type]
+        na_denom = np.arange(0, 10).astype(denom_type)
```
pierce314159, line 291: I don't love hardcoding the `na_denom` to be fixed to be fixed 0..10 of whatever type. Especially because it means `denom[0]` below ill only ever test the scalar case when it is equal to 0. I think it would be better to take random permutations of the representative arrays, so something the below instead
```python
na_num = np.random.permutation(NP_TRIG_ARRAYS[num_type])
na_denom = np.random.permutation(NP_TRIG_ARRAYS[denom_type])
```

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +204,249 @@ def test_cumprod(self, num_type):
...
+            ak.arctan2(np.array([range(0, 10)]).astype(num_type), denom[0])
+
+        # Edge case: Infinities
+        na_inf = np.array([np.inf, np.nan, -np.inf])
+        pda_inf = ak.array(na_inf)
+        na_fin = np.array([0, 5, 10]).astype(denom_type)
+        pda_fin = ak.array(na_fin)
+        assert np.allclose(
+            np.arctan2(na_inf, na_fin), ak.arctan2(pda_inf, pda_fin).to_ndarray(), equal_nan=True
+        )
```
pierce314159, line 323: I think this should be handled by the above once we make the `np.random.permutation` change

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +204,249 @@ def test_cumprod(self, num_type):
...
+            na = np.array([np.inf, np.nan, -np.inf])
+            pda = ak.array(na)
+            assert np.allclose(np.arcsinh(na), ak.arcsinh(pda).to_ndarray(), equal_nan=True)
+
+    @pytest.mark.parametrize("num_type", NO_BOOL)
+    def test_arccosh(self, num_type):
+        if num_type == ak.float64:
+            na = np.linspace(1, 10)
+        else:
+            na = np.arange(1, 10).astype(num_type)
```
pierce314159, line 399: is there a reason this test doesn't use `NP_TRIG_ARRAYS`?

PROTO_tests/tests/numeric_test.py
```diff
@@ -10,6 +10,17 @@
...
 
+NP_TRIG_ARRAYS = {
+    ak.int64: np.arange(-5, 5),
+    ak.float64: np.concatenate(
+        [
+            np.linspace(-3.5, 3.5, 5),
+            np.array([np.nan, -np.inf, -0.0, 0.0, np.inf]),
+        ]
+    ),
+    ak.uint64: np.arange(2**64 - 10, 2**64, dtype=np.uint64),
```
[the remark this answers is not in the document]
jaketrookman: No that is a good point. I will add that in there, just in case we want to use it elsewhere in a future test.

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +204,249 @@ def test_cumprod(self, num_type):
...
+
+        assert np.allclose(np.arctan(na), ak.arctan(pda).to_ndarray(), equal_nan=True)
+        with pytest.raises(TypeError):
+            ak.arctan(np.array([range(0, 10)]).astype(num_type))
+
+        # Edge case: infinities and NaNs
+        if num_type == ak.float64:
+            na = np.array([np.inf, np.nan, -np.inf])
+            pda = ak.array(na)
+            assert np.allclose(np.arctan(na), ak.arctan(pda).to_ndarray(), equal_nan=True)
```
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/numeric_test.py
jaketrookman: You are correct that these are now handled, but the next two test are for the vector-scalar and scalar-vector cases. So I will remove the repeated testing.

jaketrookman: Updated

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +205,159 @@ def test_cumprod(self, num_type):
...
+
+        assert np.allclose(
+            np.arctan2(na_num, na_denom), ak.arctan2(pda_num, pda_denom).to_ndarray(), equal_nan=True
+        )
+        assert np.allclose(
+            np.arctan2(num[0], na_denom), ak.arctan2(num[0], pda_denom).to_ndarray(), equal_nan=True
+        )
+        assert np.allclose(
+            np.arctan2(na_num, denom[0]), ak.arctan2(pda_num, denom[0]).to_ndarray(), equal_nan=True
+        )
```
pierce314159, line 280: how about we update to something like this so we aren't testing the same scalars everytime
```suggestion
        num_scal = pda_num[0]
        denom_scal = pda_denom[0]

        assert np.allclose(
            np.arctan2(na_num, na_denom), ak.arctan2(pda_num, pda_denom).to_ndarray(), equal_nan=True
        )
        assert np.allclose(
            np.arctan2(num_scal, na_denom), ak.arctan2(num_scal, pda_denom).to_ndarray(), equal_nan=True
        )
        assert np.allclose(
            np.arctan2(na_num, denom_scal), ak.arctan2(pda_num, denom_scal).to_ndarray(), equal_nan=True
        )
```

TheOddCzar789 (approved):

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +205,159 @@ def test_cumprod(self, num_type):
...
+        assert np.allclose(
+            np.arctan2(na_num, denom_scal), ak.arctan2(pda_num, denom_scal).to_ndarray(), equal_nan=True
+        )
+
+        with pytest.raises(TypeError):
+            ak.arctan2(
+                np.array([range(0, 10)]).astype(num_type), np.array([range(10, 20)]).astype(num_type)
+            )
+            ak.arctan2(num[0], np.array([range(10, 20)]).astype(num_type))
+            ak.arctan2(np.array([range(0, 10)]).astype(num_type), denom[0])
```
Ethan-DeBandi99, line 287: If all of these are expected to raise an error, they need to be checked individually. What is here will raise the expected `TypeError` on the first and the next 2 statements are never executed.

Also, the next 2 statements will not produce a `TypeError` as it stands because `num` and `denom` do not exist. You will raise a `NameError` here.

Ethan-DeBandi99 (changes requested):
The error handling portion of the `arctan2` test is not correct. This needs to be updated before merging. Otherwise, looks good.

pierce314159 (changes requested):

PROTO_tests/tests/numeric_test.py
```diff
@@ -197,22 +205,161 @@ def test_cumprod(self, num_type):
...
+    @pytest.mark.parametrize("denom_type", NO_BOOL)
+    def test_arctan2(self, num_type, denom_type):
+        na_num = np.random.permutation(NP_TRIG_ARRAYS[num_type])
+        na_denom = np.random.permutation(NP_TRIG_ARRAYS[denom_type])
+
+        pda_num = ak.array(na_num, dtype=num_type)
+        pda_denom = ak.array(na_denom, dtype=denom_type)
+
+        num = np.array([1]).astype(num_type)
+        denom = np.array([5]).astype(denom_type)
```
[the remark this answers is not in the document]
jaketrookman: Updated

Ethan-DeBandi99 (approved):
Looks good

pierce314159 (approved):
Looks good!

PR #2532, opened 2023-06-28
Closes #2505: Update `resolve_scalar_dtype` to handle `uint` and `bigint`

### LINKED ISSUE #2505: weirdness with `resolve_scalar_dtype`
the way we handle determining the type of a scalar is a bit funky and we handle it differently in different places. This was discovered in #2447.

taking a look at this example, we see that `2**63 + 1` is being interpreted as an `int` i.e. `-9223372036854775807` when it should be a `uint`. 
```python
In [3]: ak.dtypes.resolve_scalar_dtype(2**63 + 1)
Out[3]: 'int64'

In [4]: ak.arange(10) + (2**63 + 1)
Out[4]: array([-9223372036854775807 -9223372036854775806 -9223372036854775805 -9223372036854775804 -9223372036854775803 -9223372036854775802 -9223372036854775801 -9223372036854775800 -9223372036854775799 -9223372036854775798])

In [5]: ak.dtypes.resolve_scalar_dtype(2**65)
Out[5]: 'int64'

In [6]: ak.arange(10) + 2**65
Out[6]: array(["36893488147419103232" "36893488147419103233" "36893488147419103234" "36893488147419103235" "36893488147419103236" "36893488147419103237" "36893488147419103238" "36893488147419103239" "36893488147419103240" "36893488147419103241"])
```

`resolve_scalar_dtype` also sees `2**65` as a `int` instead of a `bigint`

we handle this differently in different spots:
https://github.com/Bears-R-Us/arkouda/blob/3bf0ea9e17b2ffe20aa69a84574d0daf3965f664/arkouda/dtypes.py#L261-L266

https://github.com/Bears-R-Us/arkouda/blob/3bf0ea9e17b2ffe20aa69a84574d0daf3965f664/arkouda/message.py#L214-L215

https://github.com/Bears-R-Us/arkouda/blob/3bf0ea9e17b2ffe20aa69a84574d0daf3965f664/arkouda/pdarrayclass.py#L360-L369

I think we should add these checks for `if >=2**63 then uint` and `if >=2**64 then bigint` to `resolve_scalar_dtype` and then update all these sites to use that instead

pierce314159:
This PR (closes #2505) updates `resolve_scalar_dtype` to correctly handle `bigint` and `uint` scalars. It also updates places where this is handled differently to just call that function.

### DIFF
--- a/arkouda/dtypes.py
+++ b/arkouda/dtypes.py
@@ -32,7 +32,7 @@ __all__ = [
     "all_scalars",
     "get_byteorder",
     "get_server_byteorder",
-    "isSupportedNumber"
+    "isSupportedNumber",
 ]
 
 NUMBER_FORMAT_STRINGS = {
@@ -121,7 +121,6 @@ The DType enum defines the supported Arkouda data types in string form.
 
 
 class DType(Enum):
-
     BOOL = "bool"
     FLOAT = "float"
     FLOAT64 = "float64"
@@ -260,7 +259,11 @@ def resolve_scalar_dtype(val: object) -> str:  # type: ignore
         return "bool"
     # Python int or np.int* or np.uint*
     elif isinstance(val, int) or (hasattr(val, "dtype") and cast(np.uint, val).dtype.kind in "ui"):
-        if isinstance(val, np.uint64):
+        # we've established these are int, uint, or bigint,
+        # so we can do comparisons
+        if isSupportedInt(val) and val >= 2**64:  # type: ignore
+            return "bigint"
+        elif isinstance(val, np.uint64) or val >= 2**63:  # type: ignore
             return "uint64"
         else:
             return "int64"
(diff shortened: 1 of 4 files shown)


Ethan-DeBandi99 (approved):
Looks good to me

jaketrookman (approved):
Everything looks good to me. Could you just explain why we need to switch from `int` to `int64` here.

pierce314159:
> Everything looks good to me. Could you just explain why we need to switch from `int` to `int64` here.

@jaketrookman yeah! so previously our `message.py` was using `type(val).__name__` to determine the dtype of scalars and the dtype of elements within a list. Now that we've updated it to use `resolve_scalar_dtype`, it returns `int64` instead of `int` this is nice because it matches the pdarray dtypes, but also it allows for `uint64` and `bigint`.

So previously if you looked at the logs `2**63 - 1`, `2**63 + 1`, and `2**65 + 1` would all have `dtype: int`. But now they'll have `int64`, `uint64`, and `bigint` respectively

PR #2533, opened 2023-06-29
dynamic available memory calculation for all locales

### LINKED ISSUE #2477: Dynamic available memory calculation for all locales
**Feature Description**
Arkouda currently utilizes the physical memory on the locale0 host or memMax to estimate whether a command request will cause 1..n locales to exceed physical memory or memMax and cause 1..n locales to crash. We need to do better than this. 

I wrote a simple chapel function that retrieves the available memory from /proc/meminfo:

```
module MemoryReporting {
  use IO;

  proc getMemory() throws {
        var aFile = open('/proc/meminfo', ioMode.r);
        var lines = aFile.reader().lines();
        var line : string;

        var memTotal:int;
        var memAvail:int;

        for line in lines do {
            if line.find('MemTotal:') >= 0 {
              writeln(line);
            }

            if line.find('MemAvailable:') >= 0 {
              var splits = line.split('MemAvailable:');
              writeln('available memory %s'.format(splits[1]));
            }
        }
  }

  proc main() {
    try! getMemory();
  }
}

```

image

Since our target users are on systems that are running either Linux or Unix (which both have /proc/meminfo, the proposal is as follows:

1. iterate over all locales
2. calculate avail memory for each
3. take total, projected memory required for command, divide by numLocales, and confirm if there's avail memory to execute on each locale
4. alert user if 1..n locales do not have sufficient RAM

hokiegeek2:
Closes #2477 

This PR delivers the following dynamic Arkouda memory tracking, reporting, and mgmt features:

1. on each locale MemoryMgmt checks to see if the projected memory required for an incoming cmd can be accommodated by the memory allocated to the Arkouda chapel process _on every node_. If so, the command is allowed to proceed.
2. if the answer to (1) is false, on each locale MemoryMgmt checks to see if additional memory can be allocated to accommodate the new, projected total Arkouda memory _on every node_. If so, the command is allowed to proceed. If not, a server-side error is thrown and the locales that cannot support the new malloc request are logged.
3. message is sent back to the client that OOM occurred on 1..n locales
4. new CLI arguments to enable dynamic memory mgmt. The default is the existing, static memory tracking, reporting, and  mgmt Arkouda feature

### DIFF
--- a/src/MemoryMgmt.chpl
+++ b/src/MemoryMgmt.chpl
@@ -1,7 +1,32 @@
 module MemoryMgmt {
 
     use Subprocess;
-    use Memory;
+    use Memory.Diagnostics;
+    use Logging;
+    
+    private config const logLevel = LogLevel.DEBUG;
+    private config const logChannel = LogChannel.CONSOLE;
+    const mmLogger = new Logger(logLevel,logChannel);
+
+    /*
+     * Indicates whether static locale host memory or dynamically-captured
+     * memory allocated to the Arkouda process is used to estimate whether
+     * sufficient memory is available to execute the requested command.
+     */
+    enum MemMgmtType {STATIC,DYNAMIC}
+
+    /*
+     * The percentage of currently available memory on each locale host 
+     * that is the limit for memory allocated to each Arkouda locale.
+     */
+    config const availableMemoryPct: real = 90;
+
+    /*
+     * Config param that indicates whether the static memory mgmt logic in
+     * ServerConfig or dynamic memory mgmt in this module will be used. The
+     * default is static memory mgmt.
+     */
+    config const memMgmtType = MemMgmtType.STATIC;
 
     proc getArkoudaPid() : string throws {
         var pid = spawn(["pgrep","arkouda"], stdout=pipeStyle.pipe);
@@ -17,7 +42,7 @@ module MemoryMgmt {
         return pid_string;
     }
 
-    proc getArkoudaMemAlloc() : string throws {
+    proc getArkoudaMemAlloc() : uint(64) throws {
         var pid = getArkoudaPid();
 
         var sub = spawn(["pmap", pid], stdout=pipeStyle.pipe);
@@ -33,35 +58,42 @@ module MemoryMgmt {
         }
 
         sub.close();
-        return malloc_string;
+        return malloc_string:uint(64) * 1000;
     }
     
-    proc getAvailMemory() : int throws {
+    proc getAvailMemory() : uint(64) throws {
         var aFile = open('/proc/meminfo', ioMode.r);
         var lines = aFile.reader().lines();
         var line : string;
 
-        var memAvail:int;
+        var memAvail:uint(64);
 
         for line in lines do {
             if line.find('MemAvailable:') >= 0 {
                 var splits = line.split('MemAvailable:');
-                memAvail = splits[1].strip().strip(' kB'):int;
+                memAvail = splits[1].strip().strip(' kB'):uint(64);
             }
         }
-        
-        return memAvail;
+
+        return (AutoMath.round(availableMemoryPct/100 * memAvail)*1000):uint(64);
     }
 
-    proc localeMemAvailable(reqMemory: int, memLimitPct: int) : bool throws {
+    proc localeMemAvailable(reqMemory) : bool throws {
         var arkMemAlloc = getArkoudaMemAlloc();
-        var arkMemUsed = getMemory()/1000;
+        var arkMemUsed = memoryUsed();
         var availMemory = getAvailMemory();
 
-        if reqMemory + arkMemUsed <= arkMemAlloc {
+        mmLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
+                 "reqMemory: %i arkMemAlloc: %i arkMemUsed: %i availMemory: %i".format(reqMemory,
+                                                                                       arkMemAlloc,
+                                                                                       arkMemUsed,
+                                                                                       availMemory));
+        var newArkoudaMemory = reqMemory:int + arkMemUsed:int;
+        
+        if newArkoudaMemory:int <= arkMemAlloc:int {
             return true;
         } else {
-            if reqMemory + arkMemUsed <= availMemory {
+            if newArkoudaMemory:int <= availMemory {
                 return true;
             } else {
                 return false;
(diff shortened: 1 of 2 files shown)


src/MemoryMgmt.chpl
```diff
@@ -0,0 +1,140 @@
...
+
+    /*
+     * Config param that indicates whether the static memory mgmt logic in
+     * ServerConfig or dynamic memory mgmt in this module will be used. The
+     * default is static memory mgmt.
+     */
+    config const memMgmtType = MemMgmtType.STATIC;
+
+    proc getArkoudaPid() : string throws {
+        var pid = spawn(["pgrep","arkouda"], stdout=pipeStyle.pipe);
```
Ethan-DeBandi99, line 32: I don't think it matters too much (I test both and either works), but should we use `arkouda_server` instead of just `arkouda` for the process name?

Ethan-DeBandi99 (approved):
Looks good. 1 thing to look at, but I don't think it will impact functionality.

src/MemoryMgmt.chpl
hokiegeek2: @Ethan-DeBandi99 ah, I will check that out

hokiegeek2: @Ethan-DeBandi99 that works, will update accordingly, good idea

image

src/MemoryMgmt.chpl
```diff
@@ -0,0 +1,140 @@
...
+
+      coforall loc in Locales with (ref overMemLimit) {
+        on loc {
+            if !localeMemAvailable(reqMemory) {
+                overMemLimit = true;
+            }
+        }
+      }
+      
+      return if overMemLimit then false else true;
```
pierce314159, line 130: not important at all but we could just `return !overMemLimit;`

src/MemoryMgmt.chpl
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    proc main() {
+      try {
+          writeln("is mem available: %t".format(isMemAvailable(10000000)));
+      } catch e: Error{
+          try! writeln("error: %t".format(e));
+      }
+   }
```
pierce314159, line 139: this looks like it was used for testing while running this as a stand alone. Do we still want this?

src/ServerConfig.chpl
```diff
@@ -291,17 +307,43 @@ module ServerConfig
...
-                          routineName=getRoutineName(),
-                          moduleName=getModuleName(),
-                          errorClass="ErrorWithContext");                                        
+                    scLogger.error(getModuleName(),getRoutineName(),getLineNumber(), msg);  
+                    throw getErrorWithContext(
+                              msg=msg,
+                              lineNumber=getLineNumber(),
+                              routineName=getRoutineName(),
+                              moduleName=getModuleName(),
+                              errorClass="ErrorWithContext");
```
pierce314159, line 333: wouldn't it make sense to call `checkStaticMemoryLimit` here or am i misreading

pierce314159 (commented):
@hokiegeek2 got some quick questions but i think this looks good

src/MemoryMgmt.chpl
```diff
@@ -0,0 +1,140 @@
...
+
+        while sub.stdout.readLine(line) {
+            if line.find("total") > 0 {
+                var splits = line.split('total');
+                malloc_string = splits(1).strip().strip('K');
+            }
+        }
+
+        sub.close();
+        return malloc_string:uint(64) * 1000;
```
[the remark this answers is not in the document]
hokiegeek2: @pierce314159 at least on the system's I've tested AWS and my homebrew servers running Ubuntu and CentOS 7. I will, however, research and submit a follow-in issue if there's an issue (awesome pun, you're welcome).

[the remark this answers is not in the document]
hokiegeek2: @pierce314159 yeah, in regard to changing the executable name, excellent point, will add issue

src/MemoryMgmt.chpl
hokiegeek2: @pierce314159 good point, gonna rip it out now

hokiegeek2: done

src/MemoryMgmt.chpl
hokiegeek2: @pierce314159 gotcha, will add issue

hokiegeek2: @Ethan-DeBandi99 just pushed update w/ pgrep looking for arkouda_server

src/MemoryMgmt.chpl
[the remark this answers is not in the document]
hokiegeek2: @pierce314159 added [process name issue](https://github.com/Bears-R-Us/arkouda/issues/2535)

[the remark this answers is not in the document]
hokiegeek2: @pierce314159 yes, [always in KB](https://unix.stackexchange.com/questions/199482/does-proc-pid-status-always-use-kb#:~:text=Yes%2C%20it's%20always%20in%20kB,not%201000)%20to%20be%20exact.)

src/ServerConfig.chpl
hokiegeek2: ROTFL, why yes, yes it would

jaketrookman (approved):

src/ServerConfig.chpl
hokiegeek2: @pierce314159 I actually ripped the checkStaticMemoryLimit outta there...really did not help readability.

src/MemoryMgmt.chpl
hokiegeek2: @pierce314159 I will revamp that with a couple of other things later if that's okay

src/MemoryMgmt.chpl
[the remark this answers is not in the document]
pierce314159: okay cooool!!

src/MemoryMgmt.chpl
pierce314159: yah absolutely just wanted to throw it out there while it was fresh on my mind

src/ServerConfig.chpl
pierce314159: it looks like it's still there to me, idk if the changes didn't make their way to github somehow. But this is also something that can be addressed in a future cleanup deal.

pierce314159 (approved):
Looks good to me!

PR #2534, opened 2023-06-29
Closes #2268 - Fixes Sporadic failures of Parquet `segarray_write` test with gasnet

### LINKED ISSUE #2268: Sporadic failures of Parquet `segarray_write` test with gasnet
**Describe the bug**
This failure:
```
        # bool with empty segments
        a = [0, 1, 1]
        c = [1, 0]
        s = ak.SegArray.from_parts(ak.array([0, 0, len(a), len(a), len(a), len(a) + len(c)]), ak.array(a + c,
                                   dtype=ak.bool))
        with tempfile.TemporaryDirectory(dir=ParquetTest.par_test_base_tmp) as tmp_dirname:
            s.to_parquet(f"{tmp_dirname}/int_test")

            rd_data = ak.read_parquet(f"{tmp_dirname}/int_test*")
            for i in range(6):
>               self.assertListEqual(s[i].tolist(), rd_data[i].tolist())
E               AssertionError: Lists differ: [False, True, True] != []
E
E               First list contains 3 additional elements.
E               First extra element 0:
E               False
E
E               - [False, True, True]
E               + []

tests/parquet_test.py:483: AssertionError
```
popped up on nightly testing. I am able to reproduce this issue on 16 locales, but it appears to only show up about once in every 100 trials.

**To Reproduce**

- Edit `pytest.ini` to run with 4 or more locales (not sure if this is necessary, but I couldn't reproduce with 2 locales, but could with 16; fairly confident anything 4 or over would work based on the type of failure and what we've seen here in the past)
- Modify `test_segarray_write` to run a large number of trials (I seem to catch this about 1 in every 100 runs)
- Run `python3 -m pytest tests/parquet_test.py -k "test_segarray_write"` to run just this test

Ethan-DeBandi99:
Closes #2268 

This PR updates the processing for SegArray reads and writes to prevent memory corruption errors from occurring. From my testing, these were the errors I was seeing. They were different than those reported. Either way these errors needed to be resolved as they occur mainly when a locale only contains empty segments.

- Adds handling for writing to locales with segment values where all segments are empty
- Adds handling for reading from files with only empty segments
- Adds better memory management and cleanup of memory used during processing
- Updates testing file names
- Adds column with strings longer than 1 character for testing

If these updates do not also address the specific reported issue, we will reopen the issue and continue troubleshooting.

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -195,15 +195,18 @@ module ParquetMsg {
     var fileOffsets = (+ scan sizes) - sizes;
     var segmentOffsets = (+ scan rows_per_file) - rows_per_file;
     
-    coforall loc in A.targetLocales() do on loc {
+    coforall loc in A.targetLocales() do on loc { // TODO return to coforall
       var locFiles = filenames;
       var locFiledoms = subdoms;
       var locOffsets = fileOffsets; // value count offset
       var locSegOffsets = segmentOffsets; // indicates which segment index is first for the file
+      var pqErr = new parquetErrorMsg();
 
-      forall (s, off, filedom, filename) in zip(locSegOffsets, locOffsets, locFiledoms, locFiles) {
+      // forall (s, off, filedom, filename) in zip(locSegOffsets, locOffsets, locFiledoms, locFiles) {
+      for (s, off, filedom, filename) in zip(locSegOffsets, locOffsets, locFiledoms, locFiles) {
         for locdom in A.localSubdomains() {
           const intersection = domain_intersection(locdom, filedom);
+          
           if intersection.size > 0 {
             var pqErr = new parquetErrorMsg();
             if ty == ArrowTypes.stringArr {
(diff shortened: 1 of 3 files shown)


tests/parquet_test.py
```diff
@@ -423,27 +423,29 @@ def test_segarray_write(self):
             rd_data = ak.read_parquet(f"{tmp_dirname}/int_test*")
             for i in range(3):
                 self.assertListEqual(s[i].to_list(), rd_data[i].to_list())
+        print("Finished Int")
```
pierce314159, line 426:
```suggestion

```

tests/parquet_test.py
```diff
@@ -423,27 +423,29 @@ def test_segarray_write(self):
...
         s = ak.SegArray(ak.array([0, 0, len(a), len(a), len(a), len(a)+len(c)]), ak.array(a + c))
         with tempfile.TemporaryDirectory(dir=ParquetTest.par_test_base_tmp) as tmp_dirname:
-            s.to_parquet(f"{tmp_dirname}/int_test")
+            s.to_parquet(f"{tmp_dirname}/int_test_empty")
 
-            rd_data = ak.read_parquet(f"{tmp_dirname}/int_test*")
+            rd_data = ak.read_parquet(f"{tmp_dirname}/int_test_empty*")
             for i in range(6):
                 self.assertListEqual(s[i].to_list(), rd_data[i].to_list())
+        print("Finished Int Empty")
```
pierce314159, line 438:
```suggestion

```

src/ParquetMsg.chpl
```diff
@@ -1000,7 +994,12 @@ module ParquetMsg {
 
     var pqErr = new parquetErrorMsg();
 
-    if c_writeListColumnToParquet(filename.localize().c_str(), c_ptrTo(locOffsets), c_ptrTo(localVals),
+    var valPtr: c_void_ptr = nil;
+    if localVals.size != 0 {
+      valPtr = c_ptrTo(localVals);
+    }
+
+    if c_writeListColumnToParquet(filename.localize().c_str(), c_ptrTo(locOffsets), valPtr,
```
pierce314159, line 1002: is there a scenario where we might wanna do something similar with `c_ptrTo(locOffsets)`?

pierce314159 (approved):
All the logic looks good to me! It looks like github is complaining about a merge conflict. Once that gets resolved and the leftover prints get removed, this should be good to go

src/ParquetMsg.chpl
Ethan-DeBandi99: @pierce314159 - No there should not be. In those cases the intersection size will be 0 and the write will not run.

tests/parquet_test.py
Ethan-DeBandi99: Manually removed this when resolving merge conflicts.

tests/parquet_test.py
Ethan-DeBandi99: Manually removed this when resolving merge conflicts.

jaketrookman (approved):
Looks good

jaketrookman (approved):

pierce314159 (approved):

Issue #2522, opened 2023-06-23
Merge `ObjType` and `ObjectType` enums


stress-tess:
We got 2 different enums for object types. One used primarily for message / parameter objects and the other used for file IO / most select statements. We should dig into this to see if these can be merged into one thing

Ethan-DeBandi99:
Its worth noting here that these do vary slightly. The `ObjType` enum is strictly for Arkouda objects. The `ObjectType` enum used by the message parameter parser also includes values for Scalar and JSON. I am not positive that merging these is the right course of action based on their intended purpose. It may be more beneficial to rename `ObjectType` to `ParameterType` or something of that nature.

pierce314159:
Okay the rename is seems fine to me!

PR #2544, opened 2023-06-30
Closes #2542 - `message_test.py` Conversion for new test framework

### LINKED ISSUE #2542: `message_test.py` Conversion for new test framework
Part of #2499 

Configure `message_test.py` to work in new testing framework

Ethan-DeBandi99:
Closes #2542 

Configures `message_test.py` for the new test framework.

Updates the argument testing to add a bit of coverage. The tests are restructured to ensure that the arg conversion to a string that is expected by the message processing is checking the string and not doing any other modifications/alteration to test accuracy.

### DIFF
--- a/PROTO_tests/tests/message_test.py
+++ b/PROTO_tests/tests/message_test.py
@@ -1,0 +1,234 @@
+from arkouda.message import MessageFormat, MessageType, ReplyMessage, RequestMessage
+import pytest
+import json
+from arkouda.client import _json_args_to_str
+import arkouda as ak
+
+
+class TestMessage:
+    def test_message_format(self):
+        assert MessageFormat.BINARY == MessageFormat("BINARY")
+        assert MessageFormat.STRING == MessageFormat("STRING")
+        assert "BINARY" == str(MessageFormat.BINARY)
+        assert "STRING" == str(MessageFormat.STRING)
+        assert "BINARY" == repr(MessageFormat.BINARY)
+        assert "STRING" == repr(MessageFormat.STRING)
+
+        with pytest.raises(ValueError):
+            MessageFormat("STR")
+
+    def test_message_type(self):
+        assert MessageType.NORMAL == MessageType("NORMAL")
+        assert MessageType.WARNING == MessageType("WARNING")
+        assert MessageType.ERROR == MessageType("ERROR")
+
+        assert "NORMAL" == str(MessageType.NORMAL)
+        assert "WARNING" == str(MessageType.WARNING)
+        assert "ERROR" == str(MessageType.ERROR)
+        assert "NORMAL" == repr(MessageType.NORMAL)
+        assert "WARNING" == repr(MessageType.WARNING)
+        assert "ERROR" == repr(MessageType.ERROR)
+
+        with pytest.raises(ValueError):
+            MessageType("STANDARD")
+
+    def test_request_msg(self):
+        msg = RequestMessage(user="user1", token="token", cmd="connect", format=MessageFormat.STRING)
+        msgDupe = RequestMessage(user="user1", token="token", cmd="connect", format=MessageFormat.STRING)
+        msgNonDupe = RequestMessage(
+            user="user1", token="token", cmd="connect", format=MessageFormat.BINARY
+        )
+
+        assert "user1" == msg.user
+        assert "token" == msg.token
+        assert "connect" == msg.cmd
+        assert MessageFormat.STRING == msg.format
+
+        assert msg == msgDupe
+        assert msg != msgNonDupe
+
+        assert (
+            "RequestMessage(user='user1', token='token', cmd='connect', format=STRING, args=None, size=-1)"
+            == str(msg)
+        )
+
+        assert (
+            "RequestMessage(user='user1', token='token', cmd='connect', format=STRING, args=None, size=-1)"
+            == repr(msg)
+        )
+
+        assert (
+            '{"user": "user1", "token": "token", "cmd": "connect", "format": "STRING", "args": "", "size": -1}'
+            == json.dumps(msg.asdict())
+        )
+
+        # self.assertFalse(self.assertRaises(Exception, json.loads(json.dumps(msg.asdict()))))
+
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/message_test.py
```diff
@@ -0,0 +1,234 @@
...
+            "RequestMessage(user='user1', token='token', cmd='connect', format=STRING, args=None, size=-1)"
+            == repr(msg)
+        )
+
+        assert (
+            '{"user": "user1", "token": "token", "cmd": "connect", "format": "STRING", "args": "", "size": -1}'
+            == json.dumps(msg.asdict())
+        )
+
+        # self.assertFalse(self.assertRaises(Exception, json.loads(json.dumps(msg.asdict()))))
```
pierce314159, line 65:
```suggestion

```

PROTO_tests/tests/message_test.py
```diff
@@ -0,0 +1,234 @@
...
+    def test_addl_str(self):
+        val = "abc"
+        size, args = _json_args_to_str({"arg": val})
+        expected = json.dumps([
+            json.dumps({"key": "arg", "objType": "VALUE", "dtype": ak.resolve_scalar_dtype(val), "val": val}),
+        ])
+        assert args == expected
+
+    @pytest.mark.parametrize("dtype", [int])
+    def test_list_arg(self, dtype):
```
pierce314159, line 131: should we do this test for dtypes other than just `int`? if not, i feel like maybe don't parameterize it and just use `int`?

PROTO_tests/tests/message_test.py
```diff
@@ -0,0 +1,234 @@
...
+            }
+        }
+        size, args = _json_args_to_str(data)
+        expected = json.dumps([
+            json.dumps({"key": "json_1", "objType": "DICT", "dtype": "dict", "val": json.dumps([
+                json.dumps({"key": "param1", "objType": "VALUE", "dtype": ak.resolve_scalar_dtype(data["json_1"]["param1"]), "val": "1"}),
+                json.dumps({"key": "param2", "objType": "VALUE", "dtype": ak.resolve_scalar_dtype(data["json_1"]["param2"]), "val": "abc"}),
+                json.dumps({"key": "param3", "objType": "LIST", "dtype": ak.resolve_scalar_dtype(data["json_1"]["param3"][0]), "val": json.dumps([str(x) for x in data["json_1"]["param3"]])}),
+                json.dumps({"key": "param4", "objType": "PDARRAY", "dtype": str(data["json_1"]["param4"].dtype), "val": data["json_1"]["param4"].name}),
+                json.dumps({"key": "param5", "objType": "SEGSTRING", "dtype": "str", "val": data["json_1"]["param5"].name}),
```
pierce314159, line 231: we use `data["json_1"]` a lot in this chunk. I'm wondering if it might be cleaner to save it off? I don't have super strong feelings about it tho

pierce314159 (approved):
Looks great! couple small comments but only thing that needs to be done is removing commented out line

jaketrookman (approved):
Besides Pierce's comments, everything else looks good

PROTO_tests/tests/message_test.py
Ethan-DeBandi99: I thought about that as well, but wasn't sure. I went ahead and updated it.

PROTO_tests/tests/message_test.py
Ethan-DeBandi99: This should us the `SCALAR_TYPES`. I initially built it with `int` and forgot to update once everything was done. It should be updated now.

PROTO_tests/tests/message_test.py
Ethan-DeBandi99: Removed with the other changes.

jaketrookman (approved):
Looks good

TheOddCzar789 (approved):

PROTO_tests/tests/message_test.py
```diff
@@ -0,0 +1,348 @@
...
+            ]
+        )
+        assert args == expected
+
+    @pytest.mark.parametrize("dtype", SCALAR_TYPES)
+    def test_list_arg(self, dtype):
+        l1 = [0, 1, 2, 3]
+        l1 = [dtype(x) for x in l1]
+        l2 = [9, 8, 7]
+        l2 = [dtype(x) for x in l2]
```
pierce314159, line 154: This is tiny and doesn't need to be updated, but here's the suggestion in case u want to
```suggestion
        l1 = [dtype(x) for x in [0, 1, 2, 3]]
        l2 = [dtype(x) for x in [9, 8, 7]]
```

PROTO_tests/tests/message_test.py
```diff
@@ -0,0 +1,348 @@
...
+                        "objType": "LIST",
+                        "dtype": ak.resolve_scalar_dtype(l2[0]),
+                        "val": json.dumps([str(x) for x in l2]),
+                    }
+                ),
+            ]
+        )
+        assert args == expected
+
+    def test_list_addl_str(self):
```
pierce314159, line 179: what does `addl` mean?

PROTO_tests/tests/message_test.py
pierce314159, line 1: it's kinda weird that `ObjectType(Enum)` has `SEGARRAY`, `DATETIME`, and `TIMEDELTA` which never seem used. I think this outside the scope of this PR and will be handled when we combine this enum with `ObjType` in #2522

pierce314159 (approved):
This looks good to me!

PROTO_tests/tests/message_test.py
Ethan-DeBandi99: additional

PROTO_tests/tests/message_test.py
Ethan-DeBandi99: I think they initially were added, but we change how anything inheriting from pdarray is handled. Also, we removed SegArray from the server so it is no longer a single name. I think I agree, these should be updated, but can be handled in the other issue.

pierce314159 (approved):

PR #2546, opened 2023-07-03
Closes #2545: Update `MemoryMgmt` to use compat modules

### LINKED ISSUE #2545: Update `MemoryMgmt`  to use compat modules
Add a compatibility shim for `ioMode` to avoid deprecation warnings when making with chpl `1.30`

EDIT:
This already exists in `ArkoudaFileCompat`. I also noticed there's a compat for`ArkoudaMemDiagnosticsCompat` that we should use to avoid breaking gt 1.30 builds

pierce314159:
This PR (closes #2545) updates `MemoryMgmt` to use `ArkoudaFileCompat` and `ArkoudaMemDiagnosticsCompat` to avoid deprecation warnings

### DIFF
--- a/src/MemoryMgmt.chpl
+++ b/src/MemoryMgmt.chpl
@@ -1,8 +1,10 @@
 module MemoryMgmt {
 
     use Subprocess;
-    use Memory.Diagnostics;
     use Logging;
+
+    use ArkoudaMemDiagnosticsCompat;
+    use ArkoudaFileCompat;
     
     private config const logLevel = LogLevel.DEBUG;
     private config const logChannel = LogChannel.CONSOLE;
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):

hokiegeek2 (approved):
just as an FYI, this is a breaking change for 1.29.0.

pierce314159:
> just as an FYI, this is a breaking change for 1.29.0.

@hokiegeek2 and i just talked about his offline, but we should be good since there is already a [compat module which adds `ioMode`](https://github.com/Bears-R-Us/arkouda/blob/master/src/compat/e-129/ArkoudaFileCompat.chpl). And also we're about to drop 1.29 support now that 1.31 is out (see #2551)

PR #2551, opened 2023-07-05
Closes #2550: Drop support for Chapel 1.29

### LINKED ISSUE #2550: Drop CI testing for Chapel version 1.29
Now that Chapel 1.31 is released, we are dropping CI testing for 1.29 since it is no longer supported.

bmcdonald3:
Now that Chapel 1.31 has been released, Chapel version 1.29 is no longer supported by Arkouda.

Closes #2550


Makefile
pierce314159, line 1: do we wanna go ahead and add stuff for `gt-131` and `e-131`? or wait until theres a need to add a compat module for that?

pierce314159 (changes requested):
The only thing I have is we should revisit the compat modules and remove any that we only had because differences of `1.29` (i.e. where the `e-130` and `gt-130` version are the same). I didn't look terribly close but I think these might be a few 
- `ArkoudaArrayCompat`
- `ArkoudaBlockCompat`
- `ArkoudaMapCompat`
- `ArkoudaRegexCompat`

Ethan-DeBandi99:
Just noting that per our SOP on updating supported Chapel versions, this should not be merged until a week after the 1.31 support is added.

Makefile
bmcdonald3: Sorry, I'm not sure that I'm following this, what would we add there?

bmcdonald3:
I think we have been supporting Chapel 1.31 since the Chapel release that was on June 22nd (2 weeks ago), it just hasn't been in the CI, but we have been testing it in nightly testing.

Makefile
pierce314159: oh i was just saying rn we have `gt-130`, which would capture `1.31` and changes that are being made to chapel main (which will become `1.32`). So i was asking if we wanted to go ahead and change that to `e-131` and `gt-131` to differentiate between 1.31 and 1.32. We don't have any use cases for this yet i was just asking if we wanted to go ahead and update the makefile in prep for that stuff

pierce314159: this is very much a question. I don't have strong feelings either way and im totally down for us to just add this when becomes necessary (i.e. there's need for a compat module because of change of API in chapel main)

bmcdonald3: I think I'd be in favor of waiting on making the change to the eq-131, eq-132 dirs until that's necessary, but I did remove those extra compat files, so there shouldn't be any empty compat modules anymore.

pierce314159 (approved):

Makefile
pierce314159: @bmcdonald3 I noticed you didn't remove `Regex` and `File` compat modules. Does it not work to remove those even though they are the same for the `e-130` and `gt-130` versions.

Ethan-DeBandi99 (approved):

PR #2552, opened 2023-07-05
Closes #2548: Recommend Chapel 1.31.0 and use it for CI testing

### LINKED ISSUE #2548: Recommend Chapel 1.31.0 and use it for CI testing
Chapel 1.31 is live! This issue will track updating our CI, docs, and Makefile to reflect the new version

Either @bmcdonald3 or I can work this

bmcdonald3:
Chapel 1.31.0 includes improvements to the bigint library and further improves Arkouda compilation times with the LLVM backend.

Closes: #2548

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -159,7 +159,7 @@ jobs:
     env:
       CHPL_RT_NUM_THREADS_PER_LOCALE: ${{matrix.threads}}
     container:
-      image: chapel/${{matrix.image}}:1.30.0
+      image: chapel/${{matrix.image}}:1.31.0
     steps:
     - uses: actions/checkout@v3
     - name: Install dependencies
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -95,35 +95,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        chpl-version: ['1.29.0']
-    container:
-      image: chapel/chapel:${{matrix.chpl-version}}
-    steps:
-    - uses: actions/checkout@v3
-    - name: Install dependencies
-      run: |
-        apt-get update && apt-get install -y -V ca-certificates lsb-release wget
-        wget https://apache.jfrog.io/artifactory/arrow/$(lsb_release --id --short | tr 'A-Z' 'a-z')/apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb
-        apt-get install -y -V ./apache-arrow-apt-source-latest-$(lsb_release --codename --short).deb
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
pierce314159, line 1: on line 162 of this file it says: `image: chapel/${{matrix.image}}:1.30.0`

I'm not sure if that should be updated

pierce314159 (commented):
I think there's two spots that may have gotten missed. One in the `CI.yml` and one in [`REQUIREMENTS.md`](https://github.com/Bears-R-Us/arkouda/blob/04a26087599a0022be1ede167f7f968b0852dd47/pydoc/setup/REQUIREMENTS.md?plain=1#L7)

.github/workflows/CI.yml
bmcdonald3: Good catch, updated that one. 

For the requirements, I think that Chapel 1.30 is still the lowest requirement since it is still supported, so the line that says `Chapel 1.30 or later` still seems accurate to me

pierce314159 (approved):
Looks good! thanks ben!

.github/workflows/CI.yml
```diff
@@ -119,6 +119,34 @@ jobs:
...
+      run: |
+        chpl --version
+    - name: Build/Install Arkouda
+      run: |
+        make
+        python3 -m pip install -e .[dev]
+    - name: Arkouda unit tests
+      run: |
+        make test-python
+
```
Ethan-DeBandi99, line 149: I don't think we need a duplicate of the `arkouda_chpl_portability`. You should be able to just update the `chapel-version` matrix to `["1.29.0", "1.30.0"]`. This was we aren't duplicating things unnecessarily.

Ethan-DeBandi99 (changes requested):

.github/workflows/CI.yml
pierce314159: i think the reason ben did this was because we decided to do "recommend 1.31" and "drop 1.29" as separate PRs. They remove the the 1.29 version in https://github.com/Bears-R-Us/arkouda/pull/2551

pierce314159: sorry i completely misunderstood what u were saying, this is why i shouldn't comment on PRs right after waking up 😅

bmcdonald3: Ya, like Pierce mentioned, I was doing this because I was dropping 1.29 in another PR, so this way there are no merge conflicts. Seeing https://github.com/Bears-R-Us/arkouda/pull/2551#issuecomment-1625580404, I'll update this one though.

pierce314159 (approved):

Ethan-DeBandi99 (approved):

PR #2557, opened 2023-07-07
Closes #2471:  ak.full() for Strings

### LINKED ISSUE #2471: `ak.full` for Strings and `append_single` string segarray support
This came up in #2469 for `append_single`. I think getting that working shouldn't be too bad if we implement `ak.full` for segstrings (and i think this is functionality that would be useful in and of itself)

for the proof of concept i hacked together a `full` using `broadcast`, but we should def just write some chapel code to do this and it should be pretty efficient
```python
>>> def full(size, x):
   ...:     return ak.broadcast(ak.array([0]), ak.array([f'{x}']), size)
   ...:

>>> s = ak.array([f'str {i}' for i in range(3)])
>>> s.stick(full(s.size, 'abc'))
array(['str 0abc', 'str 1abc', 'str 2abc'])
```

_Originally posted by @pierce314159 in https://github.com/Bears-R-Us/arkouda/pull/2469#discussion_r1212016330_

TheOddCzar789:
This PR (Closes #2471):
  -  expands the existing ak.full() function to enable the creation of a Strings object of provided size that is filled with provided str
  - adds testing for creation of Strings object via ak.full()

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -34,6 +34,7 @@ __all__ = [
     "zeros",
     "ones",
     "full",
+    "_full_string",
     "zeros_like",
     "ones_like",
     "full_like",
@@ -499,10 +500,10 @@ def ones(
 @typechecked
 def full(
     size: Union[int_scalars, str],
-    fill_value: int_scalars,
+    fill_value: Union[int_scalars, str],
     dtype: Union[np.dtype, type, str, BigInt] = float64,
     max_bits: Optional[int] = None,
-) -> pdarray:
+) -> Union[pdarray, Strings]:
     """
     Create a pdarray filled with fill_value.
 
@@ -519,7 +520,7 @@ def full(
 
     Returns
     -------
-    pdarray
+    pdarray or Strings
         array of the requested size and dtype filled with fill_value
 
     Raises
@@ -543,8 +544,12 @@ def full(
     >>> ak.full(5, 5, dtype=ak.bool)
     array([True, True, True, True, True])
     """
+
     if not np.isscalar(size):
         raise TypeError(f"size must be a scalar, not {size.__class__.__name__}")
+    if type(fill_value) == str:
+       return _full_string(size, fill_value)
+
     dtype = akdtype(dtype)  # normalize dtype
     dtype_name = dtype.name if isinstance(dtype, BigInt) else cast(np.dtype, dtype).name
     # check dtype for error
@@ -557,6 +562,28 @@ def full(
         a.max_bits = max_bits
     return a
 
+@typechecked
+def _full_string(
+    size: Union[int_scalars, str],
+    fill_value: str,
+) -> Strings:
+    """
+    Create a Strings object filled with fill_value.
+
+    Parameters
+    ----------
+    size: int_scalars
+        Size of the array (only rank-1 arrays supported)
+    fill_value: str
+        Value with which the array will be filled
+
+    Returns
+    -------
+    Strings
+        array of the requested size and dtype filled with fill_value
+"""
+    repMsg = generic_msg(cmd="segmentedFull", args={"size": size, "fill_value": fill_value})
+    return Strings.from_return_msg(cast(str, repMsg))
 
 @typechecked
 def zeros_like(pda: pdarray) -> pdarray:
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -548,7 +548,7 @@ def full(
     if not np.isscalar(size):
         raise TypeError(f"size must be a scalar, not {size.__class__.__name__}")
     if type(fill_value) == str:
-       return _full_string(size, fill_value)
+        return _full_string(size, fill_value)
 
     dtype = akdtype(dtype)  # normalize dtype
     dtype_name = dtype.name if isinstance(dtype, BigInt) else cast(np.dtype, dtype).name
@@ -561,6 +561,7 @@ def full(
     if max_bits:
         a.max_bits = max_bits
     return a
+
 
 @typechecked
 def _full_string(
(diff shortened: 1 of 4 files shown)


src/SegmentedString.chpl
```diff
@@ -772,7 +774,7 @@ module SegmentedString {
       return (whereOffs, whereVals);
     }
 
-   proc segStrWhere(other: ?t, condition: [] bool, newLens: [] int) throws where t == owned SegString {
+    proc segStrWhere(other: ?t, condition: [] bool, newLens: [] int) throws where t == owned SegString {
```
jaketrookman, line 777: It looks like you might have an extra space here that may mess up alignment

tests/pdarray_creation_test.py
```diff
@@ -419,6 +419,10 @@ def test_full(self):
         string_len_full = ak.full("5", 5)
         self.assertEqual(5, len(string_len_full))
 
+        strings_full = ak.full(5, "test")
+        self.assertIsInstance(strings_full, ak.Strings)
+        self.assertEqual(5, len(strings_full))
+
```
jaketrookman, line 425: What would happen here with a size of 0. May want to add a test case that shows a nonscalar raises and error.

arkouda/pdarraycreation.py
```diff
@@ -34,6 +34,7 @@
     "zeros",
     "ones",
     "full",
+    "_full_string",
```
Ethan-DeBandi99, line 37: If we are exposing full_string here, I don't know that we need the `_` at the beginning. That is used to denote private methods that are not user facing, but adding it here makes it user facing.

Alternatively, looking at the usage of the method, I think having it as private would be the right call. Based on that we would not need to expose it within the `__all__`

src/SegmentedString.chpl
TheOddCzar789: The original version is actually missing a space.

tests/pdarray_creation_test.py
```diff
@@ -419,6 +419,10 @@ def test_full(self):
         string_len_full = ak.full("5", 5)
         self.assertEqual(5, len(string_len_full))
 
+        strings_full = ak.full(5, "test")
+        self.assertIsInstance(strings_full, ak.Strings)
+        self.assertEqual(5, len(strings_full))
```
Ethan-DeBandi99, line 424: It may also be beneficial here to verify that all of the values in the resulting Strings object are the value provided.

Ethan-DeBandi99 (changes requested):
Overall, this looks good. I think we need to make one adjustment with the client function. I provided 2 suggestions, but I think I am leaning towards keeping the function name as is and removing the function from the `__all__` because the usage is consistent with a private function.

arkouda/pdarraycreation.py
```diff
@@ -543,8 +544,12 @@ def full(
     >>> ak.full(5, 5, dtype=ak.bool)
     array([True, True, True, True, True])
     """
+
     if not np.isscalar(size):
         raise TypeError(f"size must be a scalar, not {size.__class__.__name__}")
+    if type(fill_value) == str:
```
pierce314159, line 550: it doesn't make a difference in this case but in general we try to use `isinstance` for checks like this. For this case that'd look like:
```python
if isinstance(fill_value, str):
```

There's a couple reasons for this, it makes it easier to check multiple types at once. So for instance if we wanted to act if `fill_value` is any of `str`, `int`, or `bool`, then 
```python
# checking the type directly is much longer
if type(fill_value) == str or type(fill_value) == int or type(fill_value) == bool:
# than an isinstance check
if isinstance(fill_value, (str, int, bool)):
```
It also correctly handles subclasses (classes that inherit from another) better, for example
```python
>>> class rectangle:
     def __init__(self, length, width):
         self.length = length
         self.width = width

>>> class square(rectangle):
     # rectangle where both sides are same size
     def __init__(self, side_len):
         rectangle.__init__(self, side_len, side_len)
         self.side_len = side_len

# isinstance knows a square is a rectangle and a rectangle is not necessarily a square
>>> isinstance(square(10), rectangle)
True
>>> isinstance(rectangle(5, 10), square)
False

# but checking the type directly does not
>>> type(square(10)) == rectangle
False
```


A good example of this in arkouda is `DateTime` which inherits from `pdarray`

arkouda/pdarraycreation.py
```diff
@@ -558,6 +563,30 @@ def full(
...
+    size: int_scalars
+        Size of the array (only rank-1 arrays supported)
+    fill_value: str
+        Value with which the array will be filled
+
+    Returns
+    -------
+    Strings
+        array of the requested size and dtype filled with fill_value
+"""
```
pierce314159, line 585: we need to indent this `"""`

tests/pdarray_creation_test.py
pierce314159: it seems to handle this correctly and returns an empty Strings
```python
>>> ak.full(0, 'str')
array([])
```

pierce314159 (approved):
Ethan stole all my comments, but i still found some stuff to talk about :)

Great job! I think this looks really good

arkouda/pdarraycreation.py
TheOddCzar789: Removed

tests/pdarray_creation_test.py
pierce314159: Looks like this test didn't get added

pierce314159 (commented):
Looks good other than forgetting to add the test that ethan suggested

tests/pdarray_creation_test.py
```diff
@@ -419,6 +419,12 @@ def test_full(self):
...
+        for i in range(len(strings_full)):
+            self.assertEqual(strings_full[i], "test")
```
Ethan-DeBandi99, line 426: Just noting that testing is one of the few cases where using `to_ndarray` and `to_list` is valid. This will test exactly what was request, so no changes are needed (purely for informational purposes), but we can also do something like below:

```python
self.assertListEqual(strings_full.to_list(), ["test"] * 5)
```

Ethan-DeBandi99 (approved):
One comment added for informational purposes, so no changes needed unless you prefer the alternative. Looks good.

pierce314159 (approved):

PR #2563, opened 2023-07-11
Closes #2561: Add casting directly from bool to bigint

### LINKED ISSUE #2561: Add casting from bool to bigint to compat modules
Add the code ben put in https://github.com/chapel-lang/chapel/issues/22715 to our compat modules

pierce314159:
This PR (closes #2561) adds the code to cast directly from bool to bigint to the compat modules. The only place I was able to make this change is in the class private variables in [`OperatorMsg`](https://github.com/Bears-R-Us/arkouda/blob/0eb300869728f1cc272cbdc4838b69f330a77152/src/OperatorMsg.chpl#L2045-L2061) (trying to change these cause me to have an internal compilation error)

### DIFF
--- a/src/compat/e-129/ArkoudaBigIntCompat.chpl
+++ b/src/compat/e-129/ArkoudaBigIntCompat.chpl
@@ -1,5 +1,9 @@
 module ArkoudaBigIntCompat {
-  use BigInteger;
+  public use BigInteger;
+
+  inline operator bigint.:(src: bool, type toType: bigint): bigint throws {
+    return new bigint(src:int);
+  }
 
   proc mod(ref result: bigint, const ref a: bigint, const ref b: bigint) {
     result.mod(a, b);
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -8,7 +8,6 @@ module BinOp
   use Logging;
   use Message;
   use BitOps;
-  use BigInteger;
 
   use ArkoudaBigIntCompat;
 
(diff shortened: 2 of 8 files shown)


bmcdonald3 (approved):

jaketrookman (approved):

src/compat/e-129/ArkoudaBigIntCompat.chpl
```diff
@@ -1,5 +1,9 @@
 module ArkoudaBigIntCompat {
-  use BigInteger;
+  public use BigInteger;
+
+  inline operator bigint.:(src: bool, type toType: bigint): bigint throws {
```
jaketrookman, line 4: Looks good, but can i just get an explantation of what the `operator` part of this `inline` function does.

pierce314159: yah, it's worth noting i grabbed this snippet from @bmcdonald3 in the chapel issue, but i can try and explain it.

this is just defining an operator in this case `:`. It allows us to apply `:` on a `bool` and the type of `bigint`. So now when the code has something like `myBool: bigint`, it's able to find this operator and call this function with `src=myBool` and `toType=bigint`. Does that make sense?

I believe all operators in chapel are defined this way and so you would see similar something similar for other more conventional binary operations like `+`

jaketrookman: Thank you :)

TheOddCzar789 (approved):

Ethan-DeBandi99 (approved):

PR #2565, opened 2023-07-12
Fixes #2564 - SetOps Test Sporadic Failures

### LINKED ISSUE #2564: SetOps Test Sporadic Failures
@pierce314159 noticed that the SetOps testing in the new test framework was experiencing sporadic failures. The issue is due to arrays being switch around in the validation data.

Ethan-DeBandi99:
Closes #2564

Corrects the configuration of the validation data. This was sporadic due to not every random set created having an intersection.

I validated all other testing and the configurations look correct.

### DIFF
--- a/PROTO_tests/tests/setops_test.py
+++ b/PROTO_tests/tests/setops_test.py
@@ -177,12 +177,14 @@ class TestSetOps:
 
     @pytest.mark.parametrize("size", pytest.prob_size)
     def test_in1d_multiarray_str(self, size):
+        if size < 1000:
+            size = 1000
         # large general test
-        a = ak.random_strings_uniform(1, 5, size)
-        b = ak.random_strings_uniform(1, 5, size)
+        a = ak.random_strings_uniform(1, 2, size)
+        b = ak.random_strings_uniform(1, 2, size)
 
-        c = ak.random_strings_uniform(1, 5, size)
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):

PROTO_tests/tests/setops_test.py
pierce314159, line 1: we chatted about this a bit offline. We think it would be smart to drop the length of the random strings generated to increase the odds of having a collision / non-empty intersection. We should also prob do this for other tests that require a non-empty intersection (in1d comes to mind)

I'd prefer `1,2` so we have some variation in length and we're not only testing things of uniform length. This should give up a sample space of something like `26**2 + 26 = 702`, so if we pick a prob_size larger than that we'd be guaranteed a collision. So if `prob_size < 1000` then set to `1000` otherwise do `prob_size`.

pierce314159 (changes requested):

PROTO_tests/tests/setops_test.py
Ethan-DeBandi99: @pierce314159 - I updated the testing for `intersect1d` and `in1d` to use `random_strings_uniform(1, 2, size)` and added checks for if the size is `<1000` to use `1000` as our minimum size.

pierce314159 (approved):

PR #2578, opened 2023-07-14
Closes #2570: `bigint_agg_test.py` Conversion for new test framework

### LINKED ISSUE #2570: `bigint_agg_test.py` Conversion for new test framework
Convert `bigint_agg_test.py` to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

TheOddCzar789:
This PR (Closes #2570):
  - Converts file to new test framework by adding pytest

### DIFF
--- a/PROTO_tests/tests/bigint_agg_test.py
+++ b/PROTO_tests/tests/bigint_agg_test.py
@@ -1,7 +1,16 @@
-import arkouda as ak
 import numpy as np
-
 import pytest
+
+import arkouda as ak
+
+
+def gather_scatter(a):
+    rev = ak.array(np.arange(len(a) - 1, -1, -1))
+    a2 = a[rev]
+    res = ak.zeros(len(a), dtype=a.dtype)
+    res[:] = a2
+    res[rev] = a2
+    return res
 
 
 class TestBigInt:
@@ -50,12 +59,3 @@ class TestBigInt:
         res = ak.ones_like(bi_arr)
         bi_arr[:] = res
         assert bi_arr.to_list() == res.to_list()
-
-
-def gather_scatter(a):
-    rev = ak.array(np.arange(len(a) - 1, -1, -1))
-    a2 = a[rev]
-    res = ak.zeros(len(a), dtype=a.dtype)
-    res[:] = a2
-    res[rev] = a2
-    return res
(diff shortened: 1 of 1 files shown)


jaketrookman (changes requested):
It might be worth making `SIZE` an array of values and parameterizing it to allow for easier updates for future implementation.

PROTO_tests/tests/bigint_agg_test.py
pierce314159, line 1: I agree with @jaketrookman that we should use `pytest.prob_size` as a parameterization for `SIZE`

PROTO_tests/tests/bigint_agg_test.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_change_size(self, size):
+        # Go from 256 bigint values down to just 1
+        bits = ak.randint(0, 2**32, size, dtype=ak.uint64)
+        bi_arr = ak.bigint_from_uint_arrays([bits, bits, bits, bits])
+        res = ak.ones_like(bi_arr)
+        bi_arr[:] = res
+        assert bi_arr.to_list() == res.to_list()
+
+
+def gather_scatter(a):
```
jaketrookman, line 55: Does this need a `@staticmethod` or is it ok because it is outside of the Class indentation

Ethan-DeBandi99 (commented):
Overall, looks good. Added one recommendation for moving a function, but otherwise agree with previous comments and the changes made on those fronts.

PROTO_tests/tests/bigint_agg_test.py
```diff
@@ -0,0 +1,61 @@
+import arkouda as ak
+import numpy as np
+
+import pytest
```
pierce314159, line 4: same deal with running `isort`

PROTO_tests/tests/bigint_agg_test.py
pierce314159: i agree with ethan here

pierce314159 (approved):
Couple small things, but this looks mostly good to go to me. Nice work!

PROTO_tests/tests/bigint_agg_test.py
TheOddCzar789: Moved.

PROTO_tests/tests/bigint_agg_test.py
TheOddCzar789: Done

jaketrookman (approved):
Looks good

pierce314159 (approved):
Looks good!

PR #2587, opened 2023-07-18
Fixes #2567 - Error when reading SegArray containing `nan` with parquet

### LINKED ISSUE #2567: Error when reading SegArray containing `nan` with parquet
This might not be the most straightforward way to reproduce, it's modeled after `test_segarray_read`

```python
>>> df = pd.DataFrame({"FloatList": [[3.14, np.nan, 2.23], [3.08], [np.inf, 6.8], [-0.0]]})
>>> table = pa.Table.from_pandas(df)
>>> pq.write_table(table, "segarray_parquet")
>>> ak.read_parquet("segarray_parquet*")
```
This results in a segmentation fault
```
[1]    8794 segmentation fault  ./arkouda_server
```

If we update `df` to contain 2 segarrays, it doesn't error but it does give incorrect results (treating the `nan` as a segment boundary)
```python
>>> df = pd.DataFrame({"FloatList": [[3.14, np.nan, 2.23], [3.08], [np.inf, 6.8], [-0.0]], "IntList": [[0, 1, -2], [0, 1], [3, -4, 5, 6], [-1, 2, 3]]})
>>> table = pa.Table.from_pandas(df)
>>> pq.write_table(table, "segarray_parquet")
>>> ak.read_parquet("segarray_parquet*")
{'FloatList': SegArray([
 [3.140000e+00]
 [2.230000e+00]
 [3.080000e+00]
 [inf 6.800000e+00 -0.000000e+00]
 ]),
 'IntList': SegArray([
 [0 1 -2]
 [0 1]
 [3 -4 5 6]
 [-1 2 3]
 ])}
```

Ethan-DeBandi99:
Closes #2567 

This PR corrects detection of `NaN` values and properly returns them. It is worth noting the Pandas will return these as Null when reading from Parquet so the 2 do not align 100% for testing. Previously there was a bad assumption that the SegArray case was handling `NaN` because it properly handles `inf`. This has been updated and detection/processing has been configured to properly return NaN values.

Adds Testing to validate functionality.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -419,20 +419,19 @@ int64_t cpp_getListColumnSize(const char* filename, const char* colname, void* c
           while (float_reader->HasNext()) {
             float value;
             (void)float_reader->ReadBatch(1, &definition_level, &rep_lvl, &value, &values_read);
-            // not worried about NaN here so don't need to check values read.
-            if (values_read == 0 || (!first && rep_lvl == 0)) {
+            if ((values_read == 0 && definition_level != 2) || (!first && rep_lvl == 0)) {
               seg_sizes[i] = seg_size;
               i++;
               seg_size = 0;
             }
-            if (values_read != 0) {
+            if (values_read != 0 || (values_read == 0 && definition_level == 2)) {
               seg_size++;
               vct++;
               if (first) {
                 first = false;
               }
             }
-            if (values_read != 0 && !float_reader->HasNext()){
+            if ((values_read != 0 || (values_read == 0 && definition_level == 2)) && !float_reader->HasNext()){
               seg_sizes[i] = seg_size;
             }
           }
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good

tests/parquet_test.py
```diff
@@ -534,6 +534,18 @@ def test_segarray_string(self):
             rd_data = ak.read_parquet(f"{tmp_dirname}/segarray_test_empty_*")
             self.assertListEqual(s.to_list(), rd_data.to_list())
 
+    def test_float_edge(self):
+        df = pd.DataFrame({"FloatList": [[3.14, np.nan, 2.23], [], [3.08], [np.inf, 6.8], [-0.0]]})
```
pierce314159, line 538: I pulled this down to mess around a bit with the nan in different spots and i'm seeing `0.0` instead of `nan` for all nan after the last non-`nan` is read
```python
df = pd.DataFrame({"FloatList": [[3.14, 2.23], [], [3.08], [np.inf, 6.8], [-0.0, np.nan, np.nan, np.nan]]})
```

printing out `pd_l` and `ak_l` shows the difference
```
pd_l: [[3.14, 2.23], [], [3.08], [inf, 6.8], [-0.0, nan, nan, nan]]
ak_l: [[3.14, 2.23], [], [3.08], [inf, 6.8], [-0.0, 0.0, 0.0, 0.0]]
```

src/ArrowFunctions.cpp
```diff
@@ -617,18 +615,29 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
...
-              chpl_ptr[i+j] = (double)tmpArr[j];
+            float* tmpArr = new float[batchSize]; // this will not include NaN values
+            int16_t* def_lvl = new int16_t[batchSize];
+            int16_t* rep_lvl = new int16_t[batchSize];
+            (void)reader->ReadBatch(batchSize, def_lvl, rep_lvl, tmpArr, &values_read);
+
+            int64_t tmp_offset = 0; // used to properly access tmp after NaN encountered
+            int64_t val_ct = 0;
+            int64_t j = 0;
+            while (val_ct < values_read){
```
pierce314159, line 626: I think this is the cause of the edge case i noted in the tests (all NAN's after `values_read` are treated as zeroes) because we exit the while loop after the last non-NAN. Maybe we intialize the tmpArr to be `NAN`s? Or keep looping until we've hit everything in `def_lvl` and keep writing `NAN`s when it equals 2? idk if that would fix the problem, i'm just spitballing at this point

pierce314159 (changes requested):
Found an edge case of the edge case :)

TheOddCzar789 (approved):
Nothing to add

jaketrookman (approved):
Besides the test cases, looks good

src/ArrowFunctions.cpp
Ethan-DeBandi99: This was definitely on the right track. I made a few adjustments to keep things as simple as possible.

pierce314159 (approved):
Looks great! Thanks!!

PR #2589, opened 2023-07-18
Closes #2588: Remove BigInt casting from 132 compat and add new compat dir

### LINKED ISSUE #2588: Remove BigInt casting from 132 compat and add new compat dir
Since the bool to bigint casting was added to both the Chapel
side and the Arkouda side, there was a disambiguation error
being hit when the compiler was trying to distinguish between
those two identical signatures. To solve this, compat modules
were added for 1.32, removing the double reference to the
casting method.

bmcdonald3:
Since the bool to bigint casting was added to both the Chapel side and the Arkouda side, there was a disambiguation error being hit when the compiler was trying to distinguish between those two identical signatures. To solve this, compat modules were added for 1.32, removing the double reference to the casting method.

Closes #2588


pierce314159 (approved):
Looks good to me!

Ethan-DeBandi99 (approved):

PR #2590, opened 2023-07-18
Closes #2537:  `array_view_test` conversion for new framework

### LINKED ISSUE #2537: `array_view_test.py` conversion for new test framework
Convert `array_view_test.py` to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

jaketrookman:
This PR (Closes #2537) updates  `array_view_test` to the new testing framework.

Updated to be parameterized for different dimensioned arrays in a variety of scale.

### DIFF
--- a/PROTO_tests/tests/array_view_test.py
+++ b/PROTO_tests/tests/array_view_test.py
@@ -7,13 +7,13 @@ import pytest
 import arkouda as ak
 
 SHAPE = {
-    4: (2, 2),
+    56: (7, 8),
     30: (5, 3, 2),
-    252: (4, 9, 7),
-    693: (3, 7, 3, 11),
+    27: (3, 3, 3),
+    48: (2, 3, 4, 2),
 }
 
-SIZE = [4, 30, 252, 693]
+SIZE = [56, 30, 27, 48]
 NO_BOOL = [ak.int64, ak.float64, ak.uint64]
 
 
--- a/PROTO_tests/tests/array_view_test.py
+++ b/PROTO_tests/tests/array_view_test.py
@@ -1,9 +1,9 @@
 from itertools import product
 
 import numpy as np
-import arkouda as ak
 import pytest
 
+import arkouda as ak
 
 SHAPE = {
     4: (2, 2),
@@ -13,7 +13,6 @@ SHAPE = {
 }
 
 SIZE = [4, 30, 252, 693]
-TRUTH = [(True, True, True), (False, True, True), (True, False, True), (True, True, False)]
 NO_BOOL = [ak.int64, ak.float64, ak.uint64]
 
 class TestArrayView:
--- a/PROTO_tests/tests/array_view_test.py
+++ b/PROTO_tests/tests/array_view_test.py
@@ -1,0 +1,204 @@
+from itertools import product
+
+import numpy as np
+import arkouda as ak
+import pytest
+
+
+SHAPE = {
+    4: (2, 2),
+    30: (5, 3, 2),
+    252: (4, 9, 7),
+    693: (3, 7, 3, 11),
+}
+
+SIZE = [4, 30, 252, 693]
+TRUTH = [(True, True, True), (False, True, True), (True, False, True), (True, True, False)]
+NO_BOOL = [ak.int64, ak.float64, ak.uint64]
+
+class TestArrayView:
+    @pytest.mark.parametrize("size", SIZE)
+    @pytest.mark.parametrize("num_type", NO_BOOL)
+    def test_mulitdimensional_array_creation(self, size, num_type):
+        n = np.array([[0, 0], [0, 1], [1, 1]])
+        a = ak.array([[0, 0], [0, 1], [1, 1]])
+        assert np.array_equal(n.tolist(), a.to_list())
+        n = np.arange(size).reshape(SHAPE[size])
+        a = ak.arange(size).reshape(SHAPE[size])
+        assert np.array_equal(n.tolist(), a.to_list())
+        n = np.arange(size, dtype=num_type).reshape(SHAPE[size])
+        a = ak.arange(size, dtype=num_type).reshape(SHAPE[size])
+        assert np.array_equal(n.tolist(), a.to_list())
+
+    def test_arrayview_int_indexing(self):
+        nd = np.arange(9).reshape(3, 3)
+        pd_reshape = ak.arange(9).reshape(3, 3)
+        pd_array = ak.array([[0, 1, 2], [3, 4, 5], [6, 7, 8]])
+
+        nd_ind = [nd[i, j] for (i, j) in product(range(3), repeat=2)]
+        reshape_ind = [pd_reshape[i, j] for (i, j) in product(range(3), repeat=2)]
+        array_ind = [pd_array[i, j] for (i, j) in product(range(3), repeat=2)]
+        assert nd_ind == reshape_ind
+        assert nd_ind == array_ind
+
+        with pytest.raises(IndexError):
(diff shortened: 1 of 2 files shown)


tests/array_view_test.py
```diff
@@ -93,17 +93,21 @@ def test_get_bool_pdarray(self):
...
         a[False, True, True] = 5
         self.assertListEqual(n.tolist(), a.to_list())
+        # print(n)
         n[True, False, True] = 6
+        # print(n)
         a[True, False, True] = 6
         self.assertListEqual(n.tolist(), a.to_list())
+        # print(n)
         n[True, True, False] = 13
+        # print(n)
```
jaketrookman, line 110: Forgot to remove these print statements

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+SHAPE = {
+    4: (2, 2),
+    30: (5, 3, 2),
+    252: (4, 9, 7),
+    693: (3, 7, 3, 11),
+}
+
+SIZE = [4, 30, 252, 693]
+TRUTH = [(True, True, True), (False, True, True), (True, False, True), (True, True, False)]
```
TheOddCzar789, line 16: This isn't used anywhere

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,204 @@
+from itertools import product
+
+import numpy as np
+import arkouda as ak
+import pytest
```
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/array_view_test.py
jaketrookman: Updated

Ethan-DeBandi99 (approved):

PROTO_tests/tests/array_view_test.py
pierce314159, line 1: seems like we need to run `black` again on this file

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,203 @@
...
+        assert np.array_equal(uav[iind], uav[uind])
+
+    @pytest.mark.parametrize("size", SIZE)
+    def test_set_index(self, size):
+        inav = np.arange(size).reshape(SHAPE[size])
+        unav = np.arange(size, dtype=np.uint64).reshape(SHAPE[size])
+        iav = ak.arange(size).reshape(SHAPE[size])
+        uav = ak.arange(size, dtype=ak.uint64).reshape(SHAPE[size])
+
+        nind = (1,) * inav.ndim
```
jaketrookman, line 76: Extra close review here about what values should be populating this

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,203 @@
...
+            # cannot reshape array of size 9 into shape (4,3)
+            ak.arange(9).reshape(4, 3)
+
+    @pytest.mark.parametrize("size", SIZE)
+    def test_int_list_indexing(self, size):
+        N = len(SHAPE[size])
+        iav = ak.arange(size).reshape(SHAPE[size])
+        uav = ak.arange(size, dtype=np.uint64).reshape(SHAPE[size])
+
+        iind = ak.ones(N, dtype=ak.uint64)
```
pierce314159, line 64: we define `N` but only use it once

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,203 @@
...
+        uind = ak.cast(iind, np.uint64)
```
pierce314159, line 65: I think one of these (`iind`) is meant to have integer dtype, but they're both `uint` here

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,203 @@
...
+            n = np.arange(size).reshape(SHAPE[size])
+            a = ak.arange(size).reshape(SHAPE[size])
+            truth = [True] * N
+            if i > 0:
+                truth[i - 1] = False
+            truth = tuple(truth)
+            n_bool_list = n[truth].tolist()
+            a_bool_list = a[truth].to_list()
+
+            assert n_bool_list == a_bool_list
```
pierce314159, line 103: We don't need to redefine `n` and `a` every time and also there might be a more direct to compute `truth`
```suggestion
        n = np.arange(size).reshape(SHAPE[size])
        a = ak.arange(size).reshape(SHAPE[size])
        for i in range(0, N+1):
            truth = tuple(j != i-1 if i > 0 else True for j in range(N))
            n_bool_list = n[truth].tolist()
            a_bool_list = a[truth].to_list()

            assert n_bool_list == a_bool_list
```

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,203 @@
...
+        for i in range(0, N+1):
+            n = np.arange(size).reshape(SHAPE[size])
+            a = ak.arange(size).reshape(SHAPE[size])
+            truth = [True] * N
+            if i > 0:
+                truth[i-1] = False
+            truth = tuple(truth)
+            n[truth] = 7
+            a[truth] = 7
+            assert n.tolist() == a.to_list()
```
pierce314159, line 118: same suggestions for moving `n` and `a` out of loop and constructing truth using a tuple comprehension

pierce314159 (commented):
Not a ton to comment on, this seems pretty close. Good work!

PROTO_tests/tests/array_view_test.py
pierce314159: Yah i think all 1's should be fine, but if we wanted to do random indexing we could do something like this
```python
import random
tuple(random.randint(0,y-1) for y in shape)
```

PROTO_tests/tests/array_view_test.py
jaketrookman: Updated `truth` construction, but keeping `n` and `a` inside the loop to regenerate between each test. Also added a random value to test each time

PROTO_tests/tests/array_view_test.py
jaketrookman: Updated

PROTO_tests/tests/array_view_test.py
jaketrookman: Updated

PROTO_tests/tests/array_view_test.py
jaketrookman: Updated

PROTO_tests/tests/array_view_test.py
jaketrookman: Updated

PROTO_tests/tests/array_view_test.py
```diff
@@ -0,0 +1,201 @@
...
+import numpy as np
+import pytest
+
+import arkouda as ak
+
+SHAPE = {
+    4: (2, 2),
+    30: (5, 3, 2),
+    252: (4, 9, 7),
+    693: (3, 7, 3, 11),
```
pierce314159, line 13: I think we should lower these numbers especially for the ones that have more dimensions. For `test_reshape_order` we do a distinct index for each combination and that can take a while with multi-locale. 
so I would shift the larger dimension sizes to the ones with few dimensions

jaketrookman: Updated

Ethan-DeBandi99 (approved):

PR #2595, opened 2023-07-19
provide client access to arkouda locale memory information

### LINKED ISSUE #2568: Provide client access to arkouda locale memory information
**Feature Description**
The [MemoryMgmt](https://github.com/Bears-R-Us/arkouda/blob/master/src/MemoryMgmt.chpl) module contains logic that determines whether there is sufficient (1) memory allocated to Arkouda to execute the requested command and (2) if not, if there's available memory on the locale host that can be allocated to arkouda to execute the requested command. A user requested the capability for reporting the memory status on each locale, which is a great idea.

**Example Behavior**
ak.get_status() will return (1) memory allocated to arkouda (2) memory used by arkouda within the allocated arkouda memory and (3) available memory on the locale host; this functionality will require a MemoryMgmtMsg module to provide a service endpoint.

**Additional context**
added on behalf of @21771

hokiegeek2:
Closes #2568 

This PR does the following:

1. adds per-locale memory reporting functions in MemoryMgmt and record to encapsulate memory stats for transport to client
2. adds StatusMsg module to enable client access to per-locale memory stats
3. adds client.py method for retrieving per-locale memory stats


src/MemoryMgmt.chpl
```diff
@@ -79,6 +89,44 @@ module MemoryMgmt {
...
+        var line : string;
+
+        var totalMem:uint(64);
+
+        for line in lines do {
+            if line.find('MemTotal:') >= 0 {
+                var splits = line.split('MemTotal:');
+                totalMem = splits[1].strip().strip(' kB'):uint(64);
+            }
+        }
```
Ethan-DeBandi99, line 105: should we add a break after `totalMem` is set to avoid extra iterations that are not necessary?

src/MemoryMgmt.chpl
Ethan-DeBandi99, line 1: @hokiegeek2 - This is not an issue, but want to correct something that I believe I may have informed you incorrectly on. In this case, a map should work fine. Nothing wrong with using the record here, but wanted to let you know. The map generation is only occurring when it will be populated and returned, so it should not cause any of the issues we saw with the metrics use case.

Ethan-DeBandi99 (commented):
Overall, this looks good. One spot I think we can add a break and another comment to correct information I provided based on me misunderstanding the configuration that was being described.

src/StatusMsg.chpl
```diff
@@ -0,0 +1,24 @@
...
+    use MultiTypeSymbolTable;
+    use MultiTypeSymEntry;
+    
+    private config const logLevel = ServerConfig.logLevel;
+    private config const logChannel = ServerConfig.logChannel;
+    const sLogger = new Logger(logLevel, logChannel);
+    
+    /**
+     * 
+     */
```
jaketrookman, line 16: I assume this is left over and can be deleted

src/MemoryMgmt.chpl
hokiegeek2: @Ethan-DeBandi99 gotcha, cool, I'd like to stick w/ the record here since it's strongly typed and we do have multiple object types.

src/MemoryMgmt.chpl
hokiegeek2: @Ethan-DeBandi99 good call, added that, testing once build is complete

src/StatusMsg.chpl
hokiegeek2: @jaketrookman ah, good catch, wanted to add some docs there, will do

src/MemoryMgmt.chpl
hokiegeek2: @Ethan-DeBandi99 Added, please check it out.

Ethan-DeBandi99: Looks good

Ethan-DeBandi99 (approved):
Looks good. Once the doc string is added or the comment is removed, this should be good to go

jaketrookman (approved):

src/MemoryMgmt.chpl
```diff
@@ -74,11 +84,51 @@ module MemoryMgmt {
...
+                memStatuses[here.id] = new LocaleMemoryStatus(total_mem=totalMem,
+                                                              avail_mem=availMem,
+                                                              pct_avail_mem=pctAvailMem:int,
+                                                              arkouda_mem_alloc=getArkoudaMemAlloc(),
+                                                              mem_used=memoryUsed(),
+                                                              locale_id=here.id,
+                                                              locale_hostname=here.hostname);                                      
+            }
+        }
+        return memStatuses;
```
pierce314159, line 130: I was originally trying to add `memStatuses` to the looping var to make things more clear and avoid having to do the `ref` with intent (since that can be racy and cause problems). Something like the following
```chapel
var memStatuses: [makeDistDom(numLocales-1)] LocaleMemoryStatus;

coforall (loc, m) in zip(Locales, memStatuses) {
    on loc {
        var availMem = getAvailMemory();
        var totalMem = getTotalMemory();
        var pctAvailMem = (availMem:real/totalMem)*100:int;

        m = new LocaleMemoryStatus(total_mem=totalMem,
                                  avail_mem=availMem,
                                  pct_avail_mem=pctAvailMem:int,
                                  arkouda_mem_alloc=getArkoudaMemAlloc(),
                                  mem_used=memoryUsed(),
                                  locale_id=here.id,
                                  locale_hostname=here.hostname);                                      
    }
}
// localize memStatuses somehow so it lives on locale0 before returning
return memStatuses;
```
but that requires we distribute and localize `memStatuses`. I think since we only ever access `memStatuses[here.id]`, we won't have any race conditions because two loop iterations won't touch the same section at the same time.

So this is just a roundabout way of saying I think what's here should be fine lol

src/StatusMsg.chpl
pierce314159, line 16: Since this isn't user facing, I don't think it's terribly important to have doc strings. I'm fine with adding them or just deleting the block comment. Adding a suggestion for convenience in case u decide to go the "just delete" path
```suggestion
```

pierce314159 (approved):
Looks good to me!

pierce314159 (approved):

Ethan-DeBandi99 (approved):

PR #2599, opened 2023-07-20
Closes #2541 - DataFrame HDF5 Snapshot Format

### LINKED ISSUE #2541: DataFrame HDF5 Storage
DataFrames are currently written to HDF5 with all of the columns at the root level of the file. This means that in order to maintain the integrity of the DataFrame, it is the only thing that can be saved in the file. This presents some challenges with snapshotting especially because DataFrames must be saved in their own files to allow them to be restored.

The goal here is to begin saving DataFrames with a `dsetname="dataframe"` (similar to all other objects). Then, that `dsetname` will be used to create an HDF5 group. The columns will then be saved under that group.

Columns will still be able to be accessed individually using `/dsetname/column`.

Additionally, we will need to maintain `ak.DataFrame.load` to allow loading an entire file as a DataFrame. `ak.load` and the other read functionality will now return any saved dataframes as DataFrames. Dataframes saved prior to this update will function the same as before with these functions, loading the columns individually.

Ethan-DeBandi99:
Closes #2541 

This PR adds support for saving DataFrame objects to HDF5 as a group so that they can be stored in the same file as other Arkouda objects. This format will only be used for Snapshots. 

The standard storage workflow for DataFrames is maintained as to not impact current workflows for users. If desired, we can expose both methods in the future.

Adds testing for the `ak.DataFrame_to_hdf_snapshot` method for now. This will be replaced once snapshotting is implemented since it will be tested there.

**Supporting Modifications**
- Adds `ak.DataFrame.objType`
- Adds `DATAFRAME` to the `ObjType` enum in the server code
- Adds processing support for `ak.Strings` indexes.
- Updates the existing `ak.DataFrame.from_return_msg` to `ak.DataFrame.from_attach_msg` since it is only used there and in the future the attach will be updated to use the same format as is used in `ak.DataFrame.from_return_msg`.


TheOddCzar789 (approved):
I didn't see any issues...

src/GenSymIO.chpl
```diff
@@ -215,7 +215,7 @@ module GenSymIO {
             else if (akType == ObjType.SEGARRAY || akType == ObjType.CATEGORICAL) {
                 create_str = id;
             }
-            else if akType == ObjType.GROUPBY {
+            else if (akType == ObjType.GROUPBY || akType == ObjType.DATAFRAME) {
                 create_str = id;
             }
```
pierce314159, line 220: could we combine these two if's?

src/HDF5Msg.chpl
```diff
@@ -1905,6 +1998,474 @@ module HDF5Msg {
...
+                                    dtype = getDataType(int);
+                                }
+                                when DType.UInt64 {
+                                    var key_entry = st.lookup(name);
+                                    var key = toSymEntry(toGenSymEntry(key_entry), uint);
+
+                                    // localize permutation and write dataset
+                                    var localkey: [0..#key.size] uint = key.a;
+                                    writeLocalDset(file_id, "/%s/%s".format(group, dset), c_ptrTo(localkey), key.size, int);
+                                    dtype = getDataType(uint);
```
pierce314159, line 2100:
```suggestion
                                    var localkey: [0..#key.size] uint = key.a;
                                    writeLocalDset(file_id, "/%s/%s".format(group, dset), c_ptrTo(localkey), key.size, uint);
                                    dtype = getDataType(uint);
```

src/HDF5Msg.chpl
```diff
@@ -1905,6 +1998,474 @@ module HDF5Msg {
...
+                // write dataframe meta data
+                writeDataFrameMetaData(file_id, group, objType: string, num_cols);
+                C_HDF5.H5Fclose(file_id);
+            } 
+            when MULTI_FILE {
+                var filenames = prepFiles(filename, mode, idx_gen);
+
+                // need to add the group to all files
+                coforall (loc, idx) in zip(Locales, filenames.domain) do on loc {
+                    const localeFilename = filenames[idx];
```
pierce314159, line 2252: why don't we just loop over `filenames` instead looping the domain and then indexing?

pierce314159 (commented):
Few small things

src/GenSymIO.chpl
Ethan-DeBandi99: done

src/HDF5Msg.chpl
Ethan-DeBandi99: done

src/HDF5Msg.chpl
Ethan-DeBandi99: This needs to be done using the domain for proper functionality of the `c_str()` calls.

pierce314159: makes sense!

pierce314159 (approved):
Looks good!

TheOddCzar789 (approved):

jaketrookman (approved):

PR #2601, opened 2023-07-21
Fixes #2579 - Compression when writing bools with Parquet

### LINKED ISSUE #2579: Compression when writing bools with parquet
While working #2539, it seems like we don't have compression support for writing boolean arrays with parquet. We should look into if that is supported by `pyarrow` and if it is we should figure out how to add it

Ethan-DeBandi99:
Closes #2579

This PR removes the specification of `RLE` encoding when writing Parquet files. This was only being done with compression and is not necessary as Parquet will automatically determine the best encoding strategy if one is not specified. This allows `bool` values to be written to Parquet with no issue and does not impact writing of any other types.

Updates existing test framework to verify to test compression specifically with `bool` values.

Updates the `PROTO_tests` of Parquet to test compression on all types.

Updates `pytest_PROTO.ini` test path so that it runs properly without modification.


pierce314159 (approved):
Looks great!

TheOddCzar789 (approved):

jaketrookman (approved):

PR #2604, opened 2023-07-21
Closes #2603: Rework `arange` handling of `uint` and `bigint` arguments

### LINKED ISSUE #2603: rework `arange` handling of `uint` and `bigint` arguments
We need to rework `arange`, right now the chapel code assumes if the arguments aren't `bigint` that they are `int`. This leads to server crash when arguments are `uint`s in the `2**63..2**64` range

for example:
```python
# fits in an int, so it works
>>> ak.arange(15)
array([0 1 2 3 4 5 6 7 8 9 10 11 12 13 14])

# in bigint range, only works if start and stop are specified
>>> ak.arange(2**64, 2**64 + 15)
array(["18446744073709551616" "18446744073709551617" "18446744073709551618" "18446744073709551619" "18446744073709551620" "18446744073709551621" "18446744073709551622" "18446744073709551623" "18446744073709551624" "18446744073709551625" "18446744073709551626" "18446744073709551627" "18446744073709551628" "18446744073709551629" "18446744073709551630"])

# in bigint range but only stop specified, doesn't error but gives incorrect results
>>> ak.arange(2**64 + 15)
array(["0" "1" "2" "3" "4" "5" "6" "7" "8" "9" "10" "11" "12" "13" "14"])

# upper end of uint, start and stop specified is treated as int and gives incorrect results
>>> ak.arange(2**63, 2**63 + 15)
array([-9223372036854775808 -9223372036854775807 -9223372036854775806 -9223372036854775805 -9223372036854775804 -9223372036854775803 -9223372036854775802 -9223372036854775801 -9223372036854775800 -9223372036854775799 -9223372036854775798 -9223372036854775797 -9223372036854775796 -9223372036854775795 -9223372036854775794])

# in upper end of uint, only stop specified causes server crash
>>> ak.arange(2**63 + 15)
```
the server crashes with this error message
```
/Users/pierce/proj/arkouda//src/SymArrayDmap.chpl:58: error: Attempting to allocate > max(size_t) bytes of memory
```

I think we can get around this by using `resolve_scalar_dtype` to determine the type of the arguments. Then we can either pass this to the chpl side and do a select or just get this info from message args

It's worth noting this isn't terrible to workaround (we can just add the `2**63` after the fact), but this should def be handled correctly
```python
>>> ak.arange(15, dtype=ak.uint64) + 2**63
array([9223372036854775808 9223372036854775809 9223372036854775810 9223372036854775811 9223372036854775812 9223372036854775813 9223372036854775814 9223372036854775815 9223372036854775816 9223372036854775817 9223372036854775818 9223372036854775819 9223372036854775820 9223372036854775821 9223372036854775822])

>>> ak.arange(15) + 2**64
array(["18446744073709551616" "18446744073709551617" "18446744073709551618" "18446744073709551619" "18446744073709551620" "18446744073709551621" "18446744073709551622" "18446744073709551623" "18446744073709551624" "18446744073709551625" "18446744073709551626" "18446744073709551627" "18446744073709551628" "18446744073709551629" "18446744073709551630"])
```

pierce314159:
This PR (closes #2603) reworks how arange handles `uint` and `bigint` arguments. This update provides the correct return for something in the `> 2**63` range such as `ak.arange(2**63, 2**63+15)`

Note, we might need to update #2584 to account for this changed test

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -15,15 +15,14 @@ from arkouda.dtypes import (
     bigint,
 )
 from arkouda.dtypes import dtype as akdtype
+from arkouda.dtypes import float64, get_byteorder, get_server_byteorder
+from arkouda.dtypes import int64 as akint64
 from arkouda.dtypes import (
-    float64,
-    get_byteorder,
-    get_server_byteorder,
-    int64,
     int_scalars,
     isSupportedInt,
     isSupportedNumber,
     numeric_scalars,
+    resolve_scalar_dtype,
 )
 from arkouda.dtypes import uint64 as akuint64
 from arkouda.pdarrayclass import create_pdarray, pdarray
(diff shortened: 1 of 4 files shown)


arkouda/pdarraycreation.py
```diff
@@ -15,15 +15,14 @@
     bigint,
 )
 from arkouda.dtypes import dtype as akdtype
+from arkouda.dtypes import float64, get_byteorder, get_server_byteorder
```
Ethan-DeBandi99, line 18: If these are not being aliased, is there a reason they were moved from the import 2 lines below?

Ethan-DeBandi99 (commented):
One minor thing. Otherwise, I think this looks all good. Definitely will need to update testing if this is merged before the other PR.

arkouda/pdarraycreation.py
Ethan-DeBandi99: If that is what `isort` is doing let's leave it. Just wanted to check there wasn't some other reason.

Ethan-DeBandi99 (approved):

TheOddCzar789 (approved):

PR #2606, opened 2023-07-24
Closes #2512 - Snapshot to HDF5 

### LINKED ISSUE #2512: Snapshot to HDF5
In order to allow users to quickly recover in the event of server crash or to resume work at a later date, we will be adding a snapshot feature. Initially, this will take all registered objects and write them to an HDF5 file (options to capture more objects will be added in a later update). A restore feature will be added along with this that will allow the user to restore these objects.

The initial restore implementation will return a dictionary of objects, but this will be further developed to recreate the Python variables on the client.

Ethan-DeBandi99:
Closes #2512

This PR adds the ability to snapshot to HDF5 within Arkouda. The workflow takes and saves off all supported Arkouda objects in the namespace using `ak.snapshot(filename)`. These snapshots can then be loaded using `ak.restore(filename)`. The initial hope was to run the variables to the namespace upon restore, but there were a few scoping issues that prevented this in all situations. Instead, of using one method in certain situations and another elsewhere, I felt it better for this feature to remain consistent. As a result, `ak.restore` will return a dictionary identical to those from `ak.read` and `ak.load_all`. This uses the DataFrame storage format added in PR #2599, so on loading the file, DataFrames are returned as DataFrames within the return dictionary.

The return dictionary will be such that the key is the variable name that was in use at the time of the snapshot and the value will be the object.

Supporting updates to testing. The snapshot test is removed from `dataframe_test.py` and then full snapshot testing is added `io_test.py`. Expands testing to include segarray in dataframes.

Fixes some oversights from #2599 with SegArray

### DIFF
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -1250,9 +1250,9 @@ class IOTest(ArkoudaTest):
 
         with tempfile.TemporaryDirectory(dir=IOTest.io_test_dir) as tmp_dirname:
             ak.snapshot(f"{tmp_dirname}/arkouda_snapshot_test")
-            print(dir())
             # delete variables
             del df
+            del df_str_idx
             del a
             del s
             del c
@@ -1261,6 +1261,8 @@ class IOTest(ArkoudaTest):
             # verify no longer in the namespace
             with self.assertRaises(NameError):
                 self.assertTrue(not df)
+            with self.assertRaises(NameError):
+                self.assertTrue(not df)df_str_idx
             with self.assertRaises(NameError):
                 self.assertTrue(not a)
             with self.assertRaises(NameError):
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -1262,7 +1262,7 @@ class IOTest(ArkoudaTest):
             with self.assertRaises(NameError):
                 self.assertTrue(not df)
             with self.assertRaises(NameError):
-                self.assertTrue(not df)df_str_idx
+                self.assertTrue(not df_str_idx)
             with self.assertRaises(NameError):
                 self.assertTrue(not a)
             with self.assertRaises(NameError):
--- a/tests/io_test.py
+++ b/tests/io_test.py
@@ -1217,6 +1217,79 @@ class IOTest(ArkoudaTest):
             self.assertListEqual(x.segments.to_list(), rd.segments.to_list())
             self.assertListEqual(x.values.to_list(), rd.values.to_list())
 
+    def test_snapshot(self):
+        from pandas.testing import assert_frame_equal
+        df = ak.DataFrame(
+            {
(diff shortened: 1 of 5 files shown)


tests/io_test.py
```diff
@@ -1217,6 +1217,79 @@ def test_segarray_str_hdf5(self):
...
+        g = ak.GroupBy(a)
+        g_ref = {
+            "perm": g.permutation.to_list(),
+            "keys": g.keys.to_list(),
+            "segments": g.segments.to_list(),
+        }
+
+        with tempfile.TemporaryDirectory(dir=IOTest.io_test_dir) as tmp_dirname:
+            ak.snapshot(f"{tmp_dirname}/arkouda_snapshot_test")
+            print(dir())
```
jaketrookman, line 1253: Looks like an old print statement

tests/io_test.py
```diff
@@ -1217,6 +1217,79 @@ def test_segarray_str_hdf5(self):
...
+            # delete variables
+            del df
+            del a
+            del s
```
pierce314159, line 1257: don't we also need to `del df_str_idx`?

tests/io_test.py
```diff
@@ -1217,6 +1217,79 @@ def test_segarray_str_hdf5(self):
...
+        df_str_idx._set_index(["A" + str(i) for i in range(len(df))])
+        col_order = df.columns
+        df_ref = df.to_pandas()
+        df_str_idx_ref = df_str_idx.to_pandas()
+        a = ak.randint(0, 10, 100)
+        a_ref = a.to_list()
+        s = ak.random_strings_uniform(0, 5, 50)
+        s_ref = s.to_list()
+        c = ak.Categorical(s)
+        c_ref = c.to_list()
```
pierce314159, line 1243: I think we should pick Categorical to have some repeats so it's more similar to actual uses of cat

pierce314159 (commented):
Looks good beyond one small comment in the test

tests/io_test.py
pierce314159: quick offline discussion with ethan, this shouldn't really effect things at all. I was thinking more the parquet case where things being different sizes could mess things up (if we have repeats then the size of unique_keys wouldn't match like permutation for instance). But we're dealing with hdf5 so that doesn't matter. I don't think this needs to be changed

TheOddCzar789 (approved):
Didn't see anything else to add.

tests/io_test.py
Ethan-DeBandi99: Oh yeah. I missed that. It should be there now

jaketrookman (approved):
Looks good

pierce314159 (approved):

PR #2608, opened 2023-07-24
Closes #2607 - `groupby_test.py` conversion to new framework

### LINKED ISSUE #2607: `groupby_test.py` Conversion to new test framework
Update `groupby_test.py` to work with within the new test framework

### LINKED ISSUE #2610: `nan_test.py` conversion to new test framework
Update `nan_test.py` to work in the new testing framework.

Ethan-DeBandi99:
Closes #2607 
Closes #2610

Adds `groupby_test.py` to the new test framework. 

A few tests were running the same assertions as `test_count` so I removed those assertions in favor of running `test_count` separately. 

A lot of these tests are run over multiple `dtypes`, but the test workflows were different enough that I am not sure parameterization is a good choice. I believe some tests could be expanded, but felt that this may be better to do later as the coverage is fairly good.

If I missed anything or anyone feels that we should add additional testing for any reason, please let me know.

### DIFF
--- a/PROTO_tests/tests/groupby_test.py
+++ b/PROTO_tests/tests/groupby_test.py
@@ -9,13 +9,7 @@ from arkouda.groupbyclass import GroupByReductionType
 def to_tuple_dict(labels, values):
     # transforms labels from list of arrays into a list of tuples by index and builds a dictionary
     # labels: [array(['b', 'a', 'c']), array(['b', 'a', 'c'])] -> [('b', 'b'), ('a', 'a'), ('c', 'c')]
-    return {
-        label: value
-        for (label, value) in zip(
-            [index_tuple for index_tuple in zip(*[pda.to_ndarray() for pda in labels])],
-            values.to_ndarray(),
-        )
-    }
+    return dict(zip(list(zip(*[pda.to_ndarray() for pda in labels])), values.to_ndarray()))
 
 
 class TestGroupBy:
--- a/PROTO_tests/tests/groupby_test.py
+++ b/PROTO_tests/tests/groupby_test.py
@@ -1,0 +1,579 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+import arkouda as ak
+from arkouda.groupbyclass import GroupByReductionType
+
+
+def to_tuple_dict(labels, values):
+    # transforms labels from list of arrays into a list of tuples by index and builds a dictionary
+    # labels: [array(['b', 'a', 'c']), array(['b', 'a', 'c'])] -> [('b', 'b'), ('a', 'a'), ('c', 'c')]
+    return {
+        label: value
+        for (label, value) in zip(
+            [index_tuple for index_tuple in zip(*[pda.to_ndarray() for pda in labels])],
+            values.to_ndarray(),
+        )
+    }
+
+class TestGroupBy:
+    GROUPS = 8
+    LEVELS = [1, 2]
+
+    @classmethod
+    def setup_class(cls):
+        cls.bvalues = ak.randint(0, 1, 10, dtype=bool)
+        cls.fvalues = ak.randint(0, 1, 10, dtype=float)
+        cls.ivalues = ak.array([4, 1, 3, 2, 2, 2, 5, 5, 2, 3])
+        cls.uvalues = ak.cast(cls.ivalues, ak.uint64)
+        cls.svalues = ak.cast(cls.ivalues, str)
+        cls.bivalues = ak.cast(cls.ivalues, ak.bigint)
+        cls.igb = ak.GroupBy(cls.ivalues)
+        cls.ugb = ak.GroupBy(cls.uvalues)
+        cls.sgb = ak.GroupBy(cls.svalues)
+        cls.bigb = ak.GroupBy(cls.bivalues)
+
+    def make_arrays(self, size):
+        keys = np.random.randint(0, self.GROUPS, size, dtype=np.uint64)
+        keys2 = np.random.randint(0, self.GROUPS, size)
+        i = np.random.randint(0, size // self.GROUPS, size)
+        u = np.random.randint(0, size // self.GROUPS, size, dtype=np.uint64)
+        f = np.random.randn(size)  # normally dist random numbers
+        b = (i % 2) == 0
+        d = {"keys": keys, "keys2": keys2, "int64": i, "uint64": u, "float64": f, "bool": b}
+
+        return d
+
+    def groupby_to_arrays(self, df: pd.DataFrame, kname, vname, op, levels):
+        g = df.groupby(kname)[vname]
+        agg = g.aggregate(op.replace("arg", "idx"))
+        if op == "prod":
+            # There appears to be a bug in pandas where it sometimes
+            # reports the product of a segment as NaN when it should be 0
+            agg[agg.isna()] = 0
+        if levels == 1:
+            keys = agg.index.values
+        else:
+            keys = tuple(zip(*(agg.index.values)))
+        return keys, agg.values
+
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/groupby_test.py
```diff
@@ -0,0 +1,577 @@
...
+def to_tuple_dict(labels, values):
+    # transforms labels from list of arrays into a list of tuples by index and builds a dictionary
+    # labels: [array(['b', 'a', 'c']), array(['b', 'a', 'c'])] -> [('b', 'b'), ('a', 'a'), ('c', 'c')]
+    return {
+        label: value
+        for (label, value) in zip(
+            [index_tuple for index_tuple in zip(*[pda.to_ndarray() for pda in labels])],
+            values.to_ndarray(),
+        )
+    }
```
pierce314159, line 18: Never look at old code. Either of these are a cleaner way to do this than what I originally had
```python
return {
    label: value
    for (label, value) in zip(
        list(zip(*[pda.to_ndarray() for pda in labels])),
        values.to_ndarray(),
    )
}
```
or
```python
return dict(zip(list(zip(*[pda.to_ndarray() for pda in labels])), values.to_ndarray()))
```

pierce314159 (commented):
Looks good. The only thing keeping me from approving is wanting to talk about `test_pandas_equivalency` and how not having prints might change things

PROTO_tests/tests/groupby_test.py
```diff
@@ -0,0 +1,577 @@
...
+            akg = ak.GroupBy(akdf["keys"])
+            keyname = "keys"
+        elif levels == 2:
+            akg = ak.GroupBy([akdf["keys"], akdf["keys2"]])
+            keyname = ["keys", "keys2"]
+
+        pdkeys, pdvals = self.groupby_to_arrays(df, keyname, "int64", "count", levels)
+        akkeys, akvals = akg.count()
+
+        self.compare_keys(pdkeys, akkeys, levels, pdvals, akvals)
```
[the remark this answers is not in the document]
Ethan-DeBandi99: Initially, I did not do the `op` as parameterized because it would rerun the first component of the test. However, it may be best to split this into 2 tests and parameterize on op. Or just parameterize the existing one knowing some components will be run more than necessary.

PROTO_tests/tests/groupby_test.py
Ethan-DeBandi99: For what it is worth, I am not sure this is 100% a necessary change as it will ultimately provide the same output. However, I agree its a bit cleaner, so its updated.

PROTO_tests/tests/groupby_test.py
```diff
@@ -0,0 +1,577 @@
...
+        keys, counts = self.igb.count()
+
+        assert [1, 2, 3, 4, 5] == keys.to_list()
+        assert [1, 4, 2, 1, 2] == counts.to_list()
+
+    def test_broadcast_ints(self):
+        keys, counts = self.igb.count()
+
+        results = self.igb.broadcast(1 * (counts > 2), permute=False)
+        assert [0, 1, 1, 1, 1, 0, 0, 0, 0, 0] == results.to_list()
```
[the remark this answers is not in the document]
Ethan-DeBandi99: I agree, but with the variance between each, I don't know that putting the time into these adaptations is beneficial at the moment.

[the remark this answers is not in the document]
Ethan-DeBandi99: Added issue #2613 to track.

Ethan-DeBandi99:
One other thing that I just started when working `nan_test`, it is only testing `nan` in relation to GroupBy. I think that should be moved here. I am going to convert this PR to draft for now since there are a few things to discuss.

PROTO_tests/tests/groupby_test.py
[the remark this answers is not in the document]
pierce314159: sweet thanks! yah I glanced at it and didn't see a real need for the variance, but that can be explored further in the issue

PROTO_tests/tests/groupby_test.py
```diff
@@ -0,0 +1,577 @@
...
+        for vname in ("int64", "uint64", "float64", "bool"):
+            for op in ak.GroupBy.Reductions:
+                do_check = True
+                try:
+                    pdkeys, pdvals = self.groupby_to_arrays(df, keyname, vname, op, levels)
+                except Exception:
+                    do_check = False
+                try:
+                    akkeys, akvals = akg.aggregate(akdf[vname], op)
+                except Exception as E:
```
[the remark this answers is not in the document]
Ethan-DeBandi99: After discussing with Pierce. We decided it would be best to leave the prints always on for now and they can be used by devs if needed. Otherwise, they are not harming anything.

PROTO_tests/tests/groupby_test.py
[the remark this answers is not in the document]
Ethan-DeBandi99: Parameterized. I added `count` to the parameterization as well to cover everything in one go.

Ethan-DeBandi99:
I added the `nan_test.py` test case to this file as well.

PROTO_tests/tests/groupby_test.py
```diff
@@ -89,33 +98,75 @@ def test_pandas_equivalency(self, size, levels):
...
+        if do_check:
+            for i in range(pdvals.size):
+                if np.isnan(pdvals[i]):
+                    pdvals[i] = 0.0  # clear out any nans to match ak implementation
+            print()
+            print(akg.keys.to_list())
+            print(pdkeys)
+            print(pdvals)
+            print(akkeys)
+            print(akvals)
```
pierce314159, line 167: these prints seem to be leftover debugging for the test but I might be wrong about that

PROTO_tests/tests/groupby_test.py
```diff
@@ -89,33 +98,75 @@ def test_pandas_equivalency(self, size, levels):
...
+        d = self.make_arrays_nan(size)
+        df = pd.DataFrame(d)
+        akdf = {k: ak.array(v) for k, v in d.items()}
+
+        akg = ak.GroupBy(akdf["keys"])
+        keyname = "keys"
+
+        if op == "count":
+            print(f"Doing .count()")
+        else:
```
pierce314159, line 144: we don't need this cause `"count"` isn't in `NAN_OPS` right?

PROTO_tests/tests/groupby_test.py
```diff
@@ -89,33 +98,75 @@ def test_pandas_equivalency(self, size, levels):
...
+            print(f"\nDoing aggregate(float64, {op})")
+
+        do_check = True
+        try:
+            pdkeys, pdvals = self.groupby_to_arrays(df, keyname, "float64", op, 1)
+        except Exception:
+            print("Pandas does not implement")
+            do_check = False
+        try:
+            akkeys, akvals = akg.count() if op == "count" else akg.aggregate(akdf["float64"], op, True)
```
pierce314159, line 154: we don't the `if op == "count"` check for the nan test right?

PROTO_tests/tests/groupby_test.py
Ethan-DeBandi99: Yeah I think my brain decided to leave it since I put in the issue to get it working, but it should be removed for now. Removed.

PROTO_tests/tests/groupby_test.py
Ethan-DeBandi99: Same as above. Removed.

pierce314159 (approved):

jaketrookman (approved):
Looks good

Issue #2602, opened 2023-07-21
`io_test.py` conversion for new framework


pierce314159:
Convert `io_test.py` to new framework

PR #2618, opened 2023-07-25
Closed #2616: `index_test.py` conversion for new framework

### LINKED ISSUE #2616: `index_test.py` Conversion for new test framework
Convert `index_test.py` to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

jaketrookman:
This PR (Closes #2616) updates `index_test.py` to the new testing framework.

Removes the `test_save` test and will be handled by @pierce314159 in his HDF5 testing in #2602.

### DIFF
--- a/PROTO_tests/tests/index_test.py
+++ b/PROTO_tests/tests/index_test.py
@@ -1,0 +1,105 @@
+import pytest
+
+import arkouda as ak
+
+
+class TestIndex:
+    def test_index_creation(self):
+        idx = ak.Index(ak.arange(5))
+
+        assert isinstance(idx, ak.Index)
+        assert idx.size == 5
+        assert idx.to_list() == [i for i in range(5)]
+
+    def test_multiindex_creation(self):
+        # test list generation
+        idx = ak.MultiIndex([ak.arange(5), ak.arange(5)])
+        assert isinstance(idx, ak.MultiIndex)
+        assert idx.levels == 2
+        assert idx.size == 5
+
+        # test tuple generation
+        idx = ak.MultiIndex((ak.arange(5), ak.arange(5)))
+        assert isinstance(idx, ak.MultiIndex)
+        assert idx.levels == 2
+        assert idx.size == 5
+
+        with pytest.raises(TypeError):
+            idx = ak.MultiIndex(ak.arange(5))
+
+        with pytest.raises(ValueError):
+            idx = ak.MultiIndex([ak.arange(5), ak.arange(3)])
+
+    def test_is_unique(self):
+        i = ak.Index(ak.array([0, 1, 2]))
+        assert i.is_unique
+
+        i = ak.Index(ak.array([0, 1, 1]))
+        assert not i.is_unique
+
+    def test_factory(self):
+        idx = ak.Index.factory(ak.arange(5))
+        assert isinstance(idx, ak.Index)
+
+        idx = ak.Index.factory([ak.arange(5), ak.arange(5)])
+        assert isinstance(idx, ak.MultiIndex)
+
+    def test_argsort(self):
+        idx = ak.Index.factory(ak.arange(5))
+        i = idx.argsort(False)
+        assert i.to_list() == [4, 3, 2, 1, 0]
+
+        idx = ak.Index(ak.array([1, 0, 4, 2, 5, 3]))
+        i = idx.argsort()
+        # values should be the indexes in the array of idx
+        assert i.to_list() == [1, 0, 3, 5, 2, 4]
+
+    def test_concat(self):
+        idx_1 = ak.Index.factory(ak.arange(5))
+
+        idx_2 = ak.Index(ak.array([2, 4, 1, 3, 0]))
+
+        idx_full = idx_1.concat(idx_2)
+        assert idx_full.to_list() == [0, 1, 2, 3, 4, 2, 4, 1, 3, 0]
+
+    def test_lookup(self):
+        idx = ak.Index.factory(ak.arange(5))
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/index_test.py
```diff
@@ -0,0 +1,105 @@
...
+import arkouda as ak
+
+
+class TestIndex:
+    def test_index_creation(self):
+        idx = ak.Index(ak.arange(5))
+
+        assert isinstance(idx, ak.Index)
+        assert idx.size == 5
+        assert idx.to_list() == [i for i in range(5)]
```
pierce314159, line 12: This is completely stylistic and optional

i wrote this originally so im really dissing myself, I think it's cleaner to do `list(range(5))` here instead
```suggestion
        assert idx.to_list() == list(range(5))
```

PROTO_tests/tests/index_test.py
```diff
@@ -0,0 +1,105 @@
...
+        lk = idx.lookup(ak.array([0, 4]))
+        assert lk.to_list() == [True, False, False, False, True]
+
+    def test_multi_argsort(self):
+        idx = ak.Index.factory([ak.arange(5), ak.arange(5)])
+        s = idx.argsort(False)
+        s.to_list() == [4, 3, 2, 1, 0]
+
+        s = idx.argsort()
+        s.to_list() == [i for i in range(5)]
```
pierce314159, line 76: we are missing `assert` here, so this isn't actually testing anything

We could also add the same stylistic flair with answer lists
```suggestion
        assert s.to_list() == list(reversed(range(5)))

        s = idx.argsort()
        assert s.to_list() == list(range(5))
```

PROTO_tests/tests/index_test.py
```diff
@@ -0,0 +1,105 @@
...
+            (1, 1),
+            (2, 2),
+            (3, 3),
+            (4, 4),
+            (0, 0),
+            (1, 1),
+            (2, 2),
+            (3, 3),
+            (4, 4),
+        ]
```
pierce314159, line 97: I don't like the way this looks. We can prob get a more readable reformatting if we declare the list of tuples on a separate line
```python
        ans = [(0, 0), (1, 1), (2, 2), (3, 3), (4, 4), (0, 0), (1, 1), (2, 2), (3, 3), (4, 4)]
        assert idx_full.to_pandas().tolist() == ans
```

pierce314159 (changes requested):

PROTO_tests/tests/index_test.py
jaketrookman: Updated

PROTO_tests/tests/index_test.py
jaketrookman: Updated

PROTO_tests/tests/index_test.py
```diff
@@ -0,0 +1,95 @@
+import pytest
+
+import arkouda as ak
+
+
+class TestIndex:
+    def test_index_creation(self):
+        idx = ak.Index(ak.arange(5))
```
TheOddCzar789, line 8: It looks like we could use pytest.prob_size for many of these.

jaketrookman: Updated, also added some ternaries for @pierce314159 :)

PROTO_tests/tests/index_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+        idx_full = idx_1.concat(idx_2)
+        assert idx_full.to_list() == [0, 1, 2, 3, 4, 2, 4, 1, 3, 0]
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_lookup(self, size):
+        idx = ak.Index.factory(ak.arange(size))
+        lk = idx.lookup(ak.array([0, size - 1]))
+
+        assert lk.to_list() == [True if (i == 0 or i == size - 1) else False for i in range(size)]
```
pierce314159, line 73: since we are doing `True if cond else False`, we can just do `cond` since this will be the correct value. So in this case
```python
assert lk.to_list() == [i == 0 or i == size - 1 for i in range(size)]
```
or more compactly
```python
assert lk.to_list() == [i in [0, size - 1] for i in range(size)]
```

PROTO_tests/tests/index_test.py
```diff
@@ -0,0 +1,102 @@
...
+        assert idx_full.to_pandas().tolist() == [(i, i) for i in range(size)] * 2
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_multi_lookup(self, size):
+        idx = ak.Index.factory([ak.arange(size), ak.arange(size)])
+        truth = [0, 3, 2]
+        lk = ak.array(truth)
+        result = idx.lookup([lk, lk])
+
+        assert result.to_list() == [True if i in truth else False for i in range(size)]
```
pierce314159, line 102: same deal here
```python
assert result.to_list() == [i in truth for i in range(size)]
```

pierce314159 (commented):
super small stuff

PROTO_tests/tests/index_test.py
jaketrookman: Updated

PROTO_tests/tests/index_test.py
jaketrookman: Updated

TheOddCzar789 (approved):

pierce314159 (approved):

PR #2622, opened 2023-07-26
Closes #2605: `dtypes_test.py` Conversion for new test framework

### LINKED ISSUE #2605: `dtypes_test.py` Conversion for new test framework

TheOddCzar789:
This PR (Closes https://github.com/Bears-R-Us/arkouda/issues/2605):

I've added parameterization for a few tests; however, the way it is currently written has some of these tests run more than is necessary, and I'm not sure if that is worth the trade-off.


PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,185 @@
...
+    def test_isSupportedInt(self, supported, unsupported):
+        """
+        Tests for both True and False scenarios of the isSupportedInt method.
+        """
+        # assert dtypes.isSupportedInt(1)
+        # assert dtypes.isSupportedInt(np.int64(1))
+        # assert dtypes.isSupportedInt(np.int64(1.0))
+        # assert not dtypes.isSupportedInt(1.0)
+        # assert not dtypes.isSupportedInt("1")
+        # assert not dtypes.isSupportedInt("1.0")
```
jaketrookman, line 95: Either leftover or unimplemented tests

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,185 @@
...
+        assert not dtypes.isSupportedFloat(unsupported)
+
+        # assert dtypes.isSupportedFloat(1.0)
+        # assert dtypes.isSupportedFloat(float(1))
+        # assert dtypes.isSupportedFloat(np.float64(1.0))
+        # assert dtypes.isSupportedFloat(np.float64(1))
+        # assert not dtypes.isSupportedFloat(np.int64(1.0))
+        # assert not dtypes.isSupportedFloat(int(1.0))
+        # assert not dtypes.isSupportedFloat("1")
+        # assert not dtypes.isSupportedFloat("1.0")
```
jaketrookman, line 116: Either leftover or unimplemented tests

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,185 @@
...
+
+        :return: None
+        :raise: AssertionError if 1.. test cases fail
+        """
+        assert "bool" == dtypes.resolve_scalar_dtype(True)
+        assert "int64" == dtypes.resolve_scalar_dtype(1)
+        assert "float64" == dtypes.resolve_scalar_dtype(float(0.0))
+        assert "str" == dtypes.resolve_scalar_dtype("test")
+        assert "int64" == dtypes.resolve_scalar_dtype(np.int64(1))
+        assert "<class 'list'>" == dtypes.resolve_scalar_dtype([1])
```
jaketrookman, line 77: I think these and a lot of other tests can be parameterized using some dictionaries.

jaketrookman (changes requested):

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Removed

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Removed

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+class TestDTypes:
+    @pytest.mark.parametrize("dtype", SUPPORTED_NP_DTYPES)
+    def test_check_np_dtype(self, dtype):
+        """
+        Tests dtypes.check_np_dtype method
+
+        :return: None
+        :raise: AssertionError if 1.. test cases fail
+        """
```
pierce314159, line 22: I think we can get rid of these doc strings. I know they're present in the current test, but I don't think they're really helping anyone

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,188 @@
...
+import pytest
+
+import arkouda as ak
+from arkouda import dtypes
+
+"""
+DtypesTest encapsulates arkouda dtypes module methods
+"""
+
+SUPPORTED_NP_DTYPES = [bool, int, float, str, np.int64, np.float64, np.uint8, np.uint64, np.str_]
```
pierce314159, line 11: we lost `np.bool_` which was present in the original `test_check_np_dtype`

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,188 @@
...
+            (np.bool_, 1, "bool"),
+            (bool, 1, "bool"),
+            (np.int64, 8, "int"),
+            (np.float64, 8, "float"),
+            (np.uint8, 1, "uint"),
+            (np.str_, 0, "str"),
+            (str, 0, "str"),
+        ],
+    )
+    def test_translate_np_dtype(self, dtype, byte_size, kind):
```
pierce314159, line 44: I think we're getting carried away with parameterization in this file. I don't really see what it's buying us in this case.

It doesn't end up being any less lines of code or easier to read to me. I could be convinced otherwise, i wonder what @Ethan-DeBandi99 thinks. I think if we used the same dictionary in multiple functions it would make sense, but since we only use it once, i'm leaning towards not doing it

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,188 @@
...
+        [
+            ("bool", True),
+            ("int64", 1),
+            ("float64", float(0.0)),
+            ("str", "test"),
+            ("int64", np.int64(1)),
+            ("<class 'list'>", [1]),
+        ],
+    )
+    def test_resolve_scalar_dtype(self, dtype, val):
```
pierce314159, line 71: I think for `resolve_scalar_dtype`, i'd be more interested in the edge cases. Does it properly handle `2**63+1` as a uint? Can it figure out that `+/- np.inf` and `np.nan` are floats? negative integers?

We should also add testing for bigint here

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+        :return: None
+        :raise: AssertionError if 1.. test cases fail
+        """
+        assert dtype == dtypes.resolve_scalar_dtype(val)
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_pdarrays_datatypes(self, size):
+        assert dtypes.dtype("float64") == ak.ones(size).dtype
+        assert dtypes.dtype("str") == ak.array(["string {}".format(i) for i in range(0, size)]).dtype
```
pierce314159, line 83: should we test more dtypes? like `uint`, `int`, and `bigint`? Also this is just be particular but using f strings is a cleaner way to declare the strings array
```python
ak.array([f"string {i}" for i in range(size)]).dtype
```

pierce314159 (changes requested):

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Is the alternative having 
`d_tuple = dtypes.translate_np_dtype(np.dtype(dtype))
  
assert kind == d_tuple[0]
        assert byte_size == d_tuple[1]`

repeated for each test case (without the parameters, of course)?

pierce314159: if i were converting it, i'd do something like the below because I feel like it's more readable
```python
def test_translate_np_dtype(self):
    for b in [np.bool_, bool]:
        assert ("bool", 1) == dtypes.translate_np_dtype(np.dtype(b))

    for s in [np.str_, str]:
        assert ("str", 0) == dtypes.translate_np_dtype(np.dtype(s))

    assert ("int", 8) == dtypes.translate_np_dtype(np.dtype(np.int64))
    assert ("uint", 8) == dtypes.translate_np_dtype(np.dtype(np.uint64))
    assert ("float", 8) == dtypes.translate_np_dtype(np.dtype(np.float64))
    assert ("uint", 1) == dtypes.translate_np_dtype(np.dtype(np.uint8))
```

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Removed

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Added

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: I didn't love the format in the old test file but yes, your suggestion is best of all.  Updated

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: (Pending; I have some questions about bigint I'd like to discuss offline)

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated (with the exception of bigint)

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,167 @@
...
+            ("float64", float(0.0)),
+            ("str", "test"),
+            ("int64", np.int64(1)),
+            ("<class 'list'>", [1]),
+            ("float64", np.nan),
+            ("float64", -np.inf),
+        ],
+    )
+    def test_resolve_scalar_dtype(self, dtype, val):
+        assert dtype == dtypes.resolve_scalar_dtype(val)
```
pierce314159, line 63: I think most of these parameterizations should be refactored. In particular the cases when it's used only once and declared in the `pytest.mark.parametrize` block

pierce314159 (changes requested):
We seem to have the same issue here as in #2584 where github is showing commits that have already been merged into master as new changes. Let me know if you need help fixing this

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Done

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+        assert "uint64" == dtypes.resolve_scalar_dtype(2**63 + 1)
+        assert "bigint" == dtypes.resolve_scalar_dtype(2**64)
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_pdarrays_datatypes(self, size):
+        assert dtypes.dtype("int64") == ak.array(np.arange(size)).dtype
+        assert dtypes.dtype("uint64") == ak.array(np.arange(size), ak.uint64).dtype
+        assert dtypes.dtype("bool") == ak.ones(size, ak.bool).dtype
+        assert dtypes.dtype("float64") == ak.ones(size).dtype
+        assert dtypes.dtype("str") == ak.array([f"string {i}" for i in range(size)]).dtype
```
pierce314159, line 71: So generating the strings in this way can get really expensive as the `prob_size` grows. A more scalable way is something like
```python
 ak.random_strings_uniform(1, 16, size=size)
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+                [ak.ones(size, dtype=ak.uint64), ak.arange(size, dtype=ak.uint64)]
+            ).dtype
+        )
+        assert dtypes.dtype("bigint") == ak.array([i for i in range(2**200, 2**200 + size)]).dtype
+
+    @pytest.mark.parametrize("supported", [1, np.int64(1), np.int64(1.0), 1, np.uint32(1)])
+    @pytest.mark.parametrize("unsupported", [1.0, "1"])
+    def test_isSupportedInt(self, supported, unsupported):
+        assert dtypes.isSupportedInt(supported)
+        assert not dtypes.isSupportedInt(unsupported)
```
pierce314159, line 84: So two things, the smaller thing is we repeat `1` in the supported list. But more importantly due to the nature of parameterization we are duplicating these checks a lot. The way the parametrization works, it will run with every possible combination so something like this (with `S` and `U` being used for shorthad
```python
S = 1, U =1.0
S = np.int64(1), U =1.0
S = np.int64(1.0), U =1.0
S = 1, U =1.0  # notice we have 1 twice in supported
S = np.uint32(1), U =1.0
```
and then the same thing for `unsupported = "1"`, as you can see we are checking that the same values are unsupported over and over. It would be more efficient to do a loop so we only check each value once
```python
def test_isSupportedInt(self, supported, unsupported):
    for supported in 1, np.int64(1), np.int64(1.0), np.uint32(1):
        assert dtypes.isSupportedInt(supported)
    for unsupported in 1.0, "1":
        assert not dtypes.isSupportedInt(unsupported)
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @pytest.mark.parametrize("supported", [1.0, float(1), np.float64(1), np.float64(1.0)])
+    @pytest.mark.parametrize("unsupported", [np.int64(1.0), int(1.0), "1.0"])
+    def test_isSupportedFloat(self, supported, unsupported):
+        assert dtypes.isSupportedFloat(supported)
+        assert not dtypes.isSupportedFloat(unsupported)
```
pierce314159, line 90: Same deal here we are checking the same values are supported or not over and over

pierce314159 (changes requested):
This is getting pretty close

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+    def test_check_np_dtype(self, dtype):
+        dtypes.check_np_dtype(np.dtype(dtype))
+
+    def test_check_np_dtype_errors(self):
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype(np.dtype(np.int16))
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype("np.str")
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype("bigint")
```
pierce314159, line 36: so doing `"bigint"` is no different than doing `"np.str"` since they are both strings. I think it would be better to updated this to 
```python
dtypes.check_np_dtype(ak.bigint)
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+        assert "int64" == dtypes.resolve_scalar_dtype(np.int64(1))
+        assert "float64" == dtypes.resolve_scalar_dtype(float(0.0))
+        assert "float64" == dtypes.resolve_scalar_dtype(float(-1.0))
+        assert "float64" == dtypes.resolve_scalar_dtype(np.nan)
+        assert "float64" == dtypes.resolve_scalar_dtype(np.inf)
+        assert "float64" == dtypes.resolve_scalar_dtype(-np.inf)
+        assert "str" == dtypes.resolve_scalar_dtype("test")
+        assert "<class 'list'>" == dtypes.resolve_scalar_dtype([1])
+        assert "uint64" == dtypes.resolve_scalar_dtype(2**63 + 1)
+        assert "bigint" == dtypes.resolve_scalar_dtype(2**64)
```
pierce314159, line 63: we can just do this in a cleaner way with `for` loops. It also makes it easier to more edge case testing without having to add more lines
```python
def test_resolve_scalar_dtype(self):
    for b in True, False:
        assert "bool" == dtypes.resolve_scalar_dtype(b)

    for i in np.iinfo(np.int64).min, -1, 0, 3, np.iinfo(np.int64).max:
        assert "int64" == dtypes.resolve_scalar_dtype(i)

    floats = [
        -np.inf,
        np.finfo(np.float64).min,
        -3.14,
        -0.0,
        0.0,
        7.0,
        np.finfo(np.float64).max,
        np.inf,
        np.nan,
    ]
    for f in floats:
        assert "float64" == dtypes.resolve_scalar_dtype(f)

    for s in "test", '"', " ", "":
        assert "str" == dtypes.resolve_scalar_dtype(s)
    assert "<class 'list'>" == dtypes.resolve_scalar_dtype([1])

    assert "uint64" == dtypes.resolve_scalar_dtype(2**63 + 1)
    assert "bigint" == dtypes.resolve_scalar_dtype(2**64)
```

It ends up being more lines of code than what's there just because I added more test cases for floats that black formats it in an expanded way

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+        assert dtypes.dtype("uint64") == ak.array(np.arange(size), ak.uint64).dtype
+        assert dtypes.dtype("bool") == ak.ones(size, ak.bool).dtype
+        assert dtypes.dtype("float64") == ak.ones(size).dtype
+        assert dtypes.dtype("str") == ak.random_strings_uniform(1, 16, size=size).dtype
+        assert (
+            dtypes.dtype("bigint")
+            == ak.bigint_from_uint_arrays(
+                [ak.ones(size, dtype=ak.uint64), ak.arange(size, dtype=ak.uint64)]
+            ).dtype
+        )
```
pierce314159, line 77: maybe slightly more visually appealing
```python
bi = ak.bigint_from_uint_arrays(
    [ak.ones(size, dtype=ak.uint64), ak.arange(size, dtype=ak.uint64)]
).dtype
assert dtypes.dtype("bigint") == bi
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+        assert dtypes.dtype("bigint") == ak.array([i for i in range(2**200, 2**200 + size)]).dtype
```
pierce314159, line 78: let's just use the arkouda `arange` function here
```python
assert dtypes.dtype("bigint") == ak.arange(2**200, 2**200 + size).dtype
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def test_isSupportedInt(self):
+        for supported in 1, np.int64(1), np.int64(1.0), np.uint32(1):
```
pierce314159, line 81: I would like to see this test expanded to test things other than `1`
```python
for supported in -10, 1, np.int64(1), np.int64(1.0), np.uint32(1), 2**63 + 1, 2**200:
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,156 @@
...
+            assert dtypes.isSupportedInt(supported)
+        for unsupported in 1.0, "1":
+            assert not dtypes.isSupportedInt(unsupported)
+
+    def test_isSupportedFloat(self):
+        for supported in 1.0, float(1), np.float64(1), np.float64(1.0):
```
pierce314159, line 87: Same here
```python
for supported in np.nan, -np.inf, 3.1, -0.0, float(1), np.float64(1):
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,166 @@
...
+class TestDTypes:
+    @pytest.mark.parametrize("dtype", SUPPORTED_NP_DTYPES)
+    def test_check_np_dtype(self, dtype):
+        dtypes.check_np_dtype(np.dtype(dtype))
+
+    def test_check_np_dtype_errors(self):
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype(np.dtype(np.int16))
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype(ak.bigint)
```
pierce314159, line 34: I didn't mean for you to get rid of the `check_np_dtype("np.str")` case, I was just pointing out that it and `check_np_dtype("bigint")` were both checking the same thing

PROTO_tests/tests/dtypes_test.py
pierce314159: I'm sorry if I wasn't clear before I think we should keep the `"np.str"` check and just edit the `"bigint"` check to do `ak.bigint` instead

TheOddCzar789: Nah that was my mistake.  Updated.

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,168 @@
...
+        assert np.int64 == dtypes.SeriesDTypes["int64"]
+        assert np.int64 == dtypes.SeriesDTypes["<class 'numpy.int64'>"]
+        assert np.float64 == dtypes.SeriesDTypes["float64"]
+        assert np.float64 == dtypes.SeriesDTypes["<class 'numpy.float64'>"]
+        assert np.bool_ == dtypes.SeriesDTypes["bool"]
+        assert np.dtype(bool) == dtypes.SeriesDTypes["bool"]
+        assert np.bool_ == dtypes.SeriesDTypes["<class 'bool'>"]
+        assert np.dtype(bool) == dtypes.SeriesDTypes["<class 'bool'>"]
+        assert np.int64 == dtypes.SeriesDTypes["datetime64[ns]"]
+        assert np.int64 == dtypes.SeriesDTypes["timedelta64[ns]"]
```
pierce314159, line 136: This can be simplified / cleaned up
```python
def test_SeriesDTypes(self):
  for dt in "int64", "<class 'numpy.int64'>", "datetime64[ns]", "timedelta64[ns]":
    assert dtypes.SeriesDTypes[dt] == np.int64
  
  for dt in "string", "<class 'str'>":
    assert dtypes.SeriesDTypes[dt] == np.str_
  
  for dt in "float64", "<class 'numpy.float64'>":
    assert dtypes.SeriesDTypes[dt] == np.float64
  
  for dt in "bool", "<class 'bool'>":
    assert dtypes.SeriesDTypes[dt] == np.bool_
```

I removed the checks for `np.dtype(bool)` because this is equivalent to `np.bool_`
```python
In [3]: np.dtype(bool) == np.bool_
Out[3]: True
```

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,168 @@
...
+
+        for s in "test", '"', " ", "":
+            assert "str" == dtypes.resolve_scalar_dtype(s)
+        assert "<class 'list'>" == dtypes.resolve_scalar_dtype([1])
+
+        assert "uint64" == dtypes.resolve_scalar_dtype(2 ** 63 + 1)
+        assert "bigint" == dtypes.resolve_scalar_dtype(2 ** 64)
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_pdarrays_datatypes(self, size):
```
pierce314159, line 79: i don't really think we gain anything by parameterizing this on `size`. I don't think having this be large will have anything to do with the dtype

TheOddCzar789: Removed

PROTO_tests/tests/dtypes_test.py
TheOddCzar789: Updated

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,168 @@
...
+    def test_check_np_dtype(self, dtype):
+        dtypes.check_np_dtype(np.dtype(dtype))
+
+    def test_check_np_dtype_errors(self):
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype(np.dtype(np.int16))
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype("np.str")
+        with pytest.raises(TypeError):
+            dtypes.check_np_dtype(ak.bigint)
```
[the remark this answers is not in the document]
TheOddCzar789: Updated

pierce314159 (changes requested):
The file looks good, but your git history is wonky again. If you look at the files changed tab, it says there are 78. I imagine this due to a botched rebase

TheOddCzar789:
> The file looks good, but your git history is wonky again. If you look at the files changed tab, it says there are 78. I imagine this due to a botched rebase

I think I managed to fix this while preserving my desired commits...

jaketrookman (approved):

PROTO_tests/tests/dtypes_test.py
```diff
@@ -0,0 +1,163 @@
...
+
+class TestDTypes:
+    @pytest.mark.parametrize("dtype", SUPPORTED_NP_DTYPES)
+    def test_check_np_dtype(self, dtype):
+        dtypes.check_np_dtype(np.dtype(dtype))
+
+    def test_check_np_dtype_errors(self):
+        for dt in np.dtype(np.int16), "np.str", ak.bigint:
+            with pytest.raises(TypeError):
+                dtypes.check_np_dtype(dt)
```
pierce314159, line 33: we're getting really nit picky, but since this is so small I think I actually like parameterizing this better
```suggestion
    @pytest.mark.parametrize("dtype", [np.dtype(np.int16), "np.str", ak.bigint])
    def test_check_np_dtype_errors(self, dtype):
        with pytest.raises(TypeError):
            dtypes.check_np_dtype(dtype)
```

pierce314159 (approved):
Looks good! Only comment I have is super nit picky

pierce314159 (approved):

jaketrookman (approved):

PR #2623, opened 2023-07-27
Closes #2602: `io_test.py` conversion for new framework

### LINKED ISSUE #2602: `io_test.py` conversion for new framework

pierce314159:
This PR (closes #2602) converts the tests from `io_test.py` to new framework increasing coverage and edge case testings in several spots. This add classes in the new `io_test` for `TestHDF5` and `TestCSV`

It also makes a small update to `arrayview` to have it's named arguments match other arkouda objects for `to_hdf`


TheOddCzar789 (approved):

jaketrookman (approved):
Looks good

PROTO_tests/tests/io_test.py
```diff
@@ -491,3 +525,735 @@ def test_ipv4_columns(self, comp):
...
+        with tempfile.TemporaryDirectory(dir=TestHDF5.hdf_test_base_tmp) as tmp_dirname:
+            file_name = f"{tmp_dirname}/hdf_test_correct"
+            ak_arr.to_hdf(file_name, "my_dset")
+
+            # test read_hdf with glob
+            gen_arr = ak.read_hdf(f"{file_name}*", "my_dset")
+            assert (ak_arr == gen_arr).all()
+
+            # test read_hdf with filenames
+            gen_arr = ak.read_hdf(filenames=[f"{file_name}_LOCALE{i:04d}" for i in range(pytest.nl)])
```
Ethan-DeBandi99, line 547: This is the only case where we do not provide the dataset name. I would recommend that we test both cases for everything. This way we ensure all the workflows are operating as expected. 

I am thinking it may be easiest to have a `test_read_and_write` and `test_read_and_write_dset_provided`. That way we are fully testing each case. 

We may also want to test when a list of datasets is specified with multiple dsets. Again, testing that `ls` works properly when not supplied, a subset can be selected properly and all can be specified by the user. This may be overkill, but I think it would be beneficial.

Ethan-DeBandi99 (commented):
Overall, this looks good. I am not seeing any major missing components. I do have 1 suggestion around the read/write testing just to make sure we fully cover all angles.

PROTO_tests/tests/io_test.py
pierce314159: I believe this should be updated. We now have both `test_read_and_write` and `test_read_and_write_dset_provided`. 

The request to test reading from a file with multiple dataset and specify no datasets, 1 dataset, half datasets, and all datasets was added to `test_read_and_write_with_dict`. 

Let me know if I misunderstood what you were asking for

Ethan-DeBandi99 (approved):

PR #2629, opened 2023-08-01
Closes #2572: `bit_ops.py` Conversion for new test framework

### LINKED ISSUE #2572: `bitops_test.py` Conversion for new test framework
Convert bitops_test.py to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

TheOddCzar789:
This PR (Closes https://github.com/Bears-R-Us/arkouda/issues/2572):

Didn't see much opportunity for parameterization, including pytest.prob_size (due to some of the tests comparing each element of a hardcoded answer array against the generated test array).


PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,219 @@
...
+        assert self.uint_arr.popcount().to_list() == ans_uint
+        assert self.bigint_arr.popcount().to_list() == ans_uint
+
+        # Edge case input
+        ans = [1, 64, 63]
+        assert ak.popcount(self.edge_cases).to_list() == ans
+
+        ans_uint = np.array(ans, ak.uint64).tolist()
+        assert ak.popcount(self.edge_cases_uint).to_list() == ans_uint
+        assert ak.popcount(self.edge_cases_bigint).to_list() == ans_uint
```
pierce314159, line 33: so we're not gaining anything by converting the answer to `uint` here because it won't change the values. I should've known better when I originally added this. Once we realize that, we can clean this up a good deal

```python
for arr in self.int_arr, self.uint_arr, self.bigint_arr:
    assert arr.popcount().to_list() == [0, 1, 1, 2, 1, 2, 2, 3, 1, 2]

for arr in self.edge_cases, self.edge_cases_uint, self.edge_cases_bigint:
    assert ak.popcount(arr).to_list() == [1, 64, 63]
```
I think most of the functions can be updated in a similar way, so I won't comment on each one, but they should all be updated

PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,219 @@
...
+
+        ans = [63, 0, 0]
+        assert ak.ctz(self.edge_cases).to_list() == ans
+
+        ans_uint = np.array(ans, ak.uint64).tolist()
+        assert ak.ctz(self.edge_cases_uint).to_list() == ans_uint
+        assert ak.ctz(self.edge_cases_bigint).to_list() == ans_uint
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_bigint_bitops(self, size):
```
pierce314159, line 81: since this is just a correctness test that shouldn't change based on the size of the array, I think I'm leaning towards not parameterizing on `size` for this one

PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,219 @@
...
+            assert par_ans.to_list() == bi.parity().to_list()
+            assert clz_ans.to_list() == bi.clz().to_list()
+            assert ctz_ans.to_list() == bi.ctz().to_list()
+
+    @pytest.mark.parametrize("dtype", [bool, str, float])
+    def test_dtypes(self, dtype):
+        f = ak.cast(ak.array(range(10)), dtype)
+
+        with pytest.raises(TypeError):
+            ak.popcount(f)
```
pierce314159, line 171: let's name this `arr` not `f` since that's usually only used for floats. Also we should continue to check both `arr.popcount()` and `ak.popcount(arr)`

I think renaming this would also be worthwhile since we're only testing error that the dtypes throw errors

PROTO_tests/tests/bitops_test.py
TheOddCzar789: Updated

PROTO_tests/tests/bitops_test.py
TheOddCzar789: Removed

PROTO_tests/tests/bitops_test.py
TheOddCzar789: Updated

PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def test_rotl(self):
+        # vector <<< scalar
+        rotated = self.int_arr.rotl(5)
+        shifted = self.int_arr << 5
+        # No wraparound, so these should be equal
+        assert rotated.to_list() == shifted.to_list()
+
+        r = ak.rotl(self.edge_cases, 1)
+        assert r.to_list() == [1, -1, -2]
```
pierce314159, line 150:
```suggestion
        assert ak.rotl(self.edge_cases, 1).to_list() == [1, -1, -2]
```

PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # vector <<< vector
+        rotated = self.int_arr.rotl(self.int_arr)
+        shifted = self.int_arr << self.int_arr
+        # No wraparound, so these should be equal
+        assert rotated.to_list() == shifted.to_list()
+
+        r = ak.rotl(self.edge_cases, ak.array([1, 1, 1]))
+        assert r.to_list() == [1, -1, -2]
```
pierce314159, line 159:
```suggestion
        assert ak.rotl(self.edge_cases, ak.array([1, 1, 1])).to_list() == [1, -1, -2]
```

PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def test_rotr(self):
+        # vector <<< scalar
+        rotated = (1024 * self.int_arr).rotr(5)
+        shifted = (1024 * self.int_arr) >> 5
+        # No wraparound, so these should be equal
+        assert rotated.to_list() == shifted.to_list()
+
+        r = ak.rotr(self.edge_cases, 1)
+        assert r.to_list() == [2**62, -1, -(2**62) - 1]
```
pierce314159, line 174:
```suggestion
        assert ak.rotr(self.edge_cases, 1).to_list() == [2**62, -1, -(2**62) - 1]
```

PROTO_tests/tests/bitops_test.py
```diff
@@ -0,0 +1,188 @@
...
+        # No wraparound, so these should be equal
+        assert rotated.to_list() == shifted.to_list()
+
+        r = ak.rotl(self.edge_cases, ak.array([1, 1, 1]))
+        assert r.to_list() == [1, -1, -2]
+
+        # scalar <<< vector
+        rotated = ak.rotl(-(2**63), self.int_arr)
+        ans = [-(2**63), 1, 2, 4, 8, 16, 32, 64, 128, 256]
+        assert rotated.to_list() == ans
```
pierce314159, line 164:
```suggestion
        assert rotated.to_list() == [-(2**63), 1, 2, 4, 8, 16, 32, 64, 128, 256]
```

pierce314159 (approved):
This looks good! I have a few suggestion but they're all stylistic. So those don't have to be implemented for this to get merged

Just a few places where we define an array that only gets used once. These used to be formatted in a weird way by black but they don't with the new format. It's possible there's other spots, I just commented on the couple of places I saw

pierce314159 (approved):
Looks good

pierce314159 (approved):

PR #2632, opened 2023-08-02
Closes #2633: Resolve Arkouda deprecation warnings for 1.32 to date

### LINKED ISSUE #2633: Resolve Arkouda deprecation warnings for 1.32 to date
During the development of Chapel 1.32, there has been a lot of progress made with the IO module and some pretty big changes that impact Arkouda regarding deprecations. This PR brings Arkouda up to date with those changes, maintaining backwards compatibility.

Included here are:
- `map.addOrSet` deprecated in favor of `map.addOrReplace`
- `%jt` deprecated in favor of usage of a `JsonSerializer`
- `%t` deprecated in favor of a wildcard `%?` format token
- change in default precision for printing real values (only affects display)

Also noting that the goal is that this should be the last release that we are making any breaking changes, so shouldn't have to maintain these compatibility modules too much longer.

bmcdonald3:
During the development of Chapel 1.32, there has been a lot of progress made with the IO module and some pretty big changes that impact Arkouda regarding deprecations. This PR brings Arkouda up to date with those changes, maintaining backwards compatibility.

Included here are:
- `map.addOrSet` deprecated in favor of `map.addOrReplace`
- `%jt` deprecated in favor of usage of a `JsonSerializer`
- `%t` deprecated in favor of a wildcard `%?` format token
- change in default precision for printing real values (only affects display)

Also noting that the goal is that this should be the last release that we are making any breaking changes, so shouldn't have to maintain these compatibility modules too much longer.

Closes #2633


Ethan-DeBandi99 (commented):
The only think I see is that we have `use *Compat` and `use ChapelModule` for the corresponding modules. Will this have any unexpected consequences? Otherwise, it looks like everything has been addressed. Thanks for taking care of all these changes @bmcdonald3

bmcdonald3:
> we have use *Compat and use ChapelModule for the corresponding modules. Will this have any unexpected consequences?

I don't think any of these usages here will cause problems. Each module in Chapel has it's own namespace (which is why we have to do something like `use ServerConfig` in all the modules separately), so when we pull in a module that has something like a `use Map;` that is already `use`d in a module that is pulling it in, that won't cause any problem. 

> if we have use ArkoudaMapCompat, we should not need use Map right?

In the case of the map compat, that is correct, we don't need both of those `use`s (since we have `public use Map;` in the compat module, if it were just a regular `use`, we'd need both), but it doesn't cause any issues having both. If you prefer, I can take out the `use Map` there, but it isn't impacting anything and will need to be re-added when we get rid of the map compat, so I could go either way.

Ethan-DeBandi99:
@bmcdonald3 - sounds good!

Ethan-DeBandi99 (approved):

src/ParquetMsg.chpl
```diff
@@ -1925,7 +1926,7 @@ module ParquetMsg {
           st.addEntry(valName, entryVal);
           rnames.pushBack((dsetname, ObjType.PDARRAY, valName));
         } else {
-          var errorMsg = "Null indices only supported on Parquet string columns, not {} columns".format(ty);
+          var errorMsg = "Null indices only supported on Parquet string columns, not {} columns".doFormat(ty);
```
pierce314159, line 1929: so this was in the existing code, but `{}` isn't how format strings work in chapel right? I think someone accidentally did it the python way. I think this needs to be updated to use `%?` instead
```suggestion
          var errorMsg = "Null indices only supported on Parquet string columns, not %? columns".doFormat(ty);
```

tests/operator_tests.py
```diff
@@ -732,10 +732,15 @@ def test_str_repr(self):
...
-            ak.linspace(0, 10, 20).__str__(),
+        # Printing floating point values changed precision in Chapel 1.32
+        self.assertTrue(
+            ("[1.100000e+00 2.300000e+00 5.000000e+00]" == ak.array([1.1, 2.3, 5]).__str__()) or
+            ("[1.1 2.3 5.0]" == ak.array([1.1, 2.3, 5]).__str__())
+        )
+        self.assertTrue(
+            ("[0.000000e+00 5.263158e-01 1.052632e+00 ... 8.947368e+00 9.473684e+00 1.000000e+01]" == \
+            ak.linspace(0, 10, 20).__str__()) or
+            ("[0.0 0.526316 1.05263 ... 8.94737 9.47368 10.0]" == ak.linspace(0, 10, 20).__str__())
```
pierce314159, line 743: this is totally fine, but I think it's a little hard to read. I think at some point i think it would be nice to update it to be something like
```python
# Printing floating point values changed precision in Chapel 1.32
answers = ["[1.100000e+00 2.300000e+00 5.000000e+00]", "[1.1 2.3 5.0]"]
self.assertTrue(ak.array([1.1, 2.3, 5]).__str__() in answers)

answers = [
    "[0.000000e+00 5.263158e-01 1.052632e+00 ... 8.947368e+00 9.473684e+00 1.000000e+01]",
    "[0.0 0.526316 1.05263 ... 8.94737 9.47368 10.0]",
]
self.assertTrue(ak.linspace(0, 10, 20).__str__() in answers)
```

tests/operator_tests.py
```diff
@@ -745,13 +750,14 @@ def test_str_repr(self):
         # Test __repr__()
         self.assertEqual("array([1 2 3])", ak.array([1, 2, 3]).__repr__())
         self.assertEqual("array([1 2 3 ... 17 18 19])", ak.arange(1, 20).__repr__())
-        self.assertEqual(
-            "array([1.1000000000000001 2.2999999999999998 5])", ak.array([1.1, 2.3, 5]).__repr__()
+        self.assertTrue(
+            "array([1.1000000000000001 2.2999999999999998 5])" == ak.array([1.1, 2.3, 5]).__repr__() or
+            "array([1.1 2.3 5])" == ak.array([1.1, 2.3, 5]).__repr__()
```
pierce314159, line 755: same thing here as in the `__str__` case. Would be nice to update for readability but not pressing

pierce314159 (commented):
A few small things to look into!

tests/operator_tests.py
bmcdonald3: You're a genius Pierce...

pierce314159: 😁

PR #2639, opened 2023-08-04
Fixes #2617: Bug in comparison of segarrays containing empty segments

### LINKED ISSUE #2617: Bug in comparison of segarrays containing empty segments
While working #2602, I found this bug in comparison between segarrays containing empty segments
```python
# comparison of segarray without empty segments, behaves as expected
>>> sa = ak.SegArray(ak.array([0, 9, 14]), ak.arange(-10, 10))
>>> sa 
SegArray([
[-10 -9 -8 -7 -6 -5 -4 -3 -2]
[-1 0 1 2 3]
[4 5 6 7 8 9]
])
>>> sa == sa
array([True True True])

# comparison of segarray with empty segments, errors
>>> sa = ak.SegArray(ak.array([0, 0, 9, 14]), ak.arange(-10, 10))
>>> sa 
SegArray([
[]
[-10 -9 -8 -7 -6 -5 -4 -3 -2]
[-1 0 1 2 3]
[4 5 6 7 8 9]
])
>>> sa == sa

File ~/proj/arkouda/arkouda/segarray.py:358, in SegArray.__eq__(self, other)
    356 leneq = self.lengths == other.lengths
    357 if leneq.sum() > 0:
--> 358     selfcmp = self[leneq]
    359     othercmp = other[leneq]
    360     intersection = self.all(selfcmp.values == othercmp.values)

File ~/proj/arkouda/arkouda/segarray.py:272, in SegArray.__getitem__(self, i)
    270     ends = starts + self.lengths[i]
    271     newsegs, inds = gen_ranges(starts, ends)
--> 272     return SegArray(newsegs, self.values[inds])
    273 else:
    274     raise TypeError(f"Invalid index type: {type(i)}")

File ~/proj/arkouda/arkouda/pdarrayclass.py:609, in pdarray.__getitem__(self, key)
    607     if kind == "bool" and self.size != key.size:
    608         raise ValueError(f"size mismatch {self.size} {key.size}")
--> 609     repMsg = generic_msg(
    610         cmd="[pdarray]",
    611         args={
    612             "array": self,
    613             "idx": key,
    614         },
    615     )
    616     return create_pdarray(repMsg)
    617 else:

RuntimeError: Error: pdarrayIndexMsg: OOBindex 1 > 19
```

This was running with 3 locales because that was what I had built for my testing. But I did have @jaketrookman run this with `comm=none` to what seems to be the same results

pierce314159:
This PR (fixes #2617):
- Fixes bug in equality check of segarrays containing empty segments. Due to bug in gen_ranges, not using the `selfcmp` groupby, and size mismatch since `segarr.all()` excludes empty segs
- Fixed issue with empty segs in `gen_ranges`
- Added support for indexing segarray with a uint pdarray
- Fixed error message in indexing and dataframe indexing
- Added tests for segarray comparisons

### DIFF
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -12,8 +12,9 @@ from arkouda.dtypes import resolve_scalar_dtype
 from arkouda.groupbyclass import GroupBy, broadcast
 from arkouda.numeric import cumsum
 from arkouda.pdarrayclass import create_pdarray, pdarray
-from arkouda.pdarraycreation import arange, array, ones, zeros
+from arkouda.pdarraycreation import arange, array, zeros
 from arkouda.pdarraysetops import concatenate, in1d
+from arkouda.segarray import gen_ranges as seg_gen_ranges
 from arkouda.strings import Strings
 
 __all__ = ["join_on_eq_with_dt", "gen_ranges", "compute_join_size"]
@@ -137,19 +138,8 @@ def gen_ranges(starts: pdarray, ends: pdarray) -> Tuple[pdarray, pdarray]:
     ranges : pdarray, int64
         The actual ranges, flattened into a single array
     """
-    if starts.size != ends.size:
-        raise ValueError("starts and ends must be same size")
-    if starts.size == 0:
-        return zeros(0, dtype=akint64), zeros(0, dtype=akint64)
-    lengths = ends - starts
-    if not (lengths > 0).all():
-        raise ValueError("all ends must be greater than starts")
-    segs = cumsum(lengths) - lengths
-    totlen = lengths.sum()
-    slices = ones(totlen, dtype=akint64)
-    diffs = concatenate((array([starts[0]]), starts[1:] - starts[:-1] - lengths[:-1] + 1))
-    slices[segs] = diffs
-    return segs, cumsum(slices)
+    # only maintain one version of gen_ranges
+    return seg_gen_ranges(starts, ends)
 
 
 @typechecked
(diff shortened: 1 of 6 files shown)


arkouda/join.py
```diff
@@ -137,19 +138,8 @@ def gen_ranges(starts: pdarray, ends: pdarray) -> Tuple[pdarray, pdarray]:
...
-    if not (lengths > 0).all():
-        raise ValueError("all ends must be greater than starts")
-    segs = cumsum(lengths) - lengths
-    totlen = lengths.sum()
-    slices = ones(totlen, dtype=akint64)
-    diffs = concatenate((array([starts[0]]), starts[1:] - starts[:-1] - lengths[:-1] + 1))
-    slices[segs] = diffs
-    return segs, cumsum(slices)
+    # only maintain one version of gen_ranges
+    return seg_gen_ranges(starts, ends)
```
Ethan-DeBandi99, line 142: Looking at the configuration of the code, I am wonder if we shouldn't just delete the function from here and only maintain it in one place. Since they are named identically, there does not seem to be a need to maintain both

Ethan-DeBandi99 (commented):
Just one thing to discuss. Otherwise, looks good.

arkouda/join.py
pierce314159: I agree, I think it would be better to delete. The only reason I didn't was if users specifically import from the module i.e. `from join import gen_ranges`. I imagine this is pretty rare and they'd just do `ak.gen_ranges`

jaketrookman (approved):
Looks good

arkouda/join.py
pierce314159: I actually decided to just remove the version in `segarray.py` since the one in `join` is the version that `ak.gen_ranges` points to. Now in segarry we just do `from join import gen_ranges`

TheOddCzar789 (approved):

Ethan-DeBandi99 (approved):

PR #2643, opened 2023-08-07
Closed #2620: `indexing_test.py` conversion for new framework

### LINKED ISSUE #2620: `indexing_test.py` Conversion for new test framework
Convert `indexing_test.py` to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

jaketrookman:
This PR (Closes #2620) updates `indexing_test.py` to the new testing framework.

### DIFF
--- a/PROTO_tests/tests/indexing_test.py
+++ b/PROTO_tests/tests/indexing_test.py
@@ -1,0 +1,104 @@
+import numpy as np
+import pytest
+
+import arkouda as ak
+
+NUM_TYPES = [ak.int64, ak.uint64, ak.float64, ak.bool, ak.bigint]
+BIGINT_AND_UINT_TYPES = [ak.bigint, ak.uint64]
+
+
+def make_key_arrays(size):
+    ikeys = ak.arange(size)
+    ukeys = ak.arange(size, dtype=ak.uint64)
+    return ikeys, ukeys
+
+
+def trial(num_types, size):
+    array_dict = {
+        ak.int64: (i := ak.randint(0, size, size)),
+        ak.uint64: (u := ak.cast(i, ak.uint64)),
+        ak.float64: ak.array(np.random.randn(size)),
+        ak.bool: (i % 2) == 0,
+        ak.bigint: ak.cast(u, ak.bigint),
+        ak.str_: ak.cast(i, str),
+    }
+    return array_dict[num_types]
+
+
+class TestIndexing:
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("num_types", NUM_TYPES)
+    def test_pdarray_uint_indexing(self, prob_size, num_types):
+        # for every pda in array_dict test indexing with uint array and uint scalar
+        ikeys, ukeys = make_key_arrays(prob_size)
+        pda = trial(num_types, prob_size)
+        assert pda[np.uint(2)] == pda[2]
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_strings_uint_indexing(self, prob_size):
+        # test Strings array indexing with uint array and uint scalar
+        ikeys, ukeys = make_key_arrays(prob_size)
+        pda = trial(ak.str_, prob_size)
+        assert pda[np.uint(2)] == pda[2]
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("bigint_and_uint_types", BIGINT_AND_UINT_TYPES)
+    def test_bool_indexing(self, prob_size, bigint_and_uint_types):
+        pda_test = trial(bigint_and_uint_types, prob_size)
+        pda_uint = ak.cast(pda_test, ak.uint64)
+        pda_bool = trial(ak.bool, prob_size)
+        assert pda_uint[pda_bool].to_list() == pda_test[pda_bool].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("num_types", NUM_TYPES)
+    def test_set_uint(self, prob_size, num_types):
+        test_size = prob_size // 2
+        ikeys, ukeys = make_key_arrays(prob_size)
+        pda = trial(num_types, prob_size)
+
+        # set [int] = val with uint key and value
+        pda[np.uint(2)] = np.uint(5)
+        assert pda[np.uint(2)] == pda[2]
+
+        # set [slice] = scalar/pdarray
+        pda[:test_size] = -2
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+        pda[:test_size] = ak.cast(ak.arange(test_size), num_types)
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+        # set [pdarray] = scalar/pdarray with uint key pdarray
+        pda[ak.arange(test_size, dtype=ak.uint64)] = np.uint(3)
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+        pda[ak.arange(test_size, dtype=ak.uint64)] = ak.cast(ak.arange(test_size), num_types)
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+        if num_types == ak.bigint:
+            # bigint specific set [int] = val with uint key and value
+            pda[np.uint(2)] = 2**200
+            assert pda[2] == 2**200
+
+            # bigint specific set [slice] = scalar/pdarray
+            pda[:prob_size] = 2**200
+            assert pda[:prob_size].to_list() == ak.full(prob_size, 2**200, ak.bigint).to_list()
+            pda[:prob_size] = ak.arange(prob_size, dtype=ak.bigint)
+            assert pda[:prob_size].to_list() == ak.arange(prob_size, dtype=ak.uint64).to_list()
+
+            # bigint specific set [pdarray] = scalar/pdarray with uint key pdarray
+            pda[ak.arange(prob_size, dtype=ak.uint64)] = 2**200
+            assert pda[:prob_size].to_list() == ak.full(prob_size, 2**200, ak.bigint).to_list()
+            pda[ak.arange(prob_size)] = ak.arange(prob_size, dtype=ak.bigint)
+            assert pda[:prob_size].to_list() == ak.arange(prob_size, dtype=ak.uint64).to_list()
+
+    def test_indexing_with_uint(self):
+        # verify reproducer from #1210 no longer fails
+        a = ak.arange(10) * 2
+        b = ak.cast(ak.array([3, 0, 8]), ak.uint64)
+        a[b]
+
+    def test_bigint_indexing_preserves_max_bits(self):
+        max_bits = 64
+        a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=max_bits)
+        assert max_bits == a[ak.arange(10)].max_bits
+        assert max_bits == a[:].max_bits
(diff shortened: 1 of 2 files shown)


TheOddCzar789 (approved):

pierce314159 (commented):
Im going to finish this review later, but wanted to go ahead an post the one thing I have in case you want to ask about it

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,104 @@
...
+        # test Strings array indexing with uint array and uint scalar
+        ikeys, ukeys = make_key_arrays(prob_size)
+        pda = trial(ak.str_, prob_size)
+        assert pda[np.uint(2)] == pda[2]
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("bigint_and_uint_types", BIGINT_AND_UINT_TYPES)
+    def test_bool_indexing(self, prob_size, bigint_and_uint_types):
+        pda_test = trial(bigint_and_uint_types, prob_size)
```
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,104 @@
...
+        assert pda[np.uint(2)] == pda[2]
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_strings_uint_indexing(self, prob_size):
+        # test Strings array indexing with uint array and uint scalar
+        ikeys, ukeys = key_arrays(prob_size)
+        pda = value_array(ak.str_, prob_size)
+        assert pda[np.uint(2)] == pda[2]
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
```
pierce314159, line 44: there's no reason these need to be separate tests, we could either create a `NUM_AND_STR_TYPES` or just do 
```python
@pytest.mark.parametrize("prob_size", pytest.prob_size)
@pytest.mark.parametrize("num_types", NUM_TYPES + [ak.str_])
def test_pdarray_uint_indexing(self, prob_size, num_types):
    # for every pda in array_dict test indexing with uint array and uint scalar
    ikeys, ukeys = key_arrays(prob_size)
    pda = value_array(num_types, prob_size)
    assert pda[np.uint(2)] == pda[2]
    assert pda[ukeys].to_list() == pda[ikeys].to_list()
```

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,104 @@
...
+def value_array(num_types, size):
+    array_dict = {
+        ak.int64: (i := ak.randint(0, size, size)),
+        ak.uint64: (u := ak.cast(i, ak.uint64)),
+        ak.float64: ak.array(np.random.randn(size)),
+        ak.bool: (i % 2) == 0,
+        ak.bigint: ak.cast(u, ak.bigint),
+        ak.str_: ak.cast(i, str),
+    }
+    return array_dict[num_types]
```
pierce314159, line 25: this is actually a pretty inefficient way of doing this if we are only ever returning one array at a time. Because for every dtype we are calculating all these arrays even if we only use one

Also we should switch the var name to be `dtype` (prob the function name too) and pick more representative examples for each dtype
```python
def make_array(dtype, size):
    if dtype is ak.int64:
        return ak.randint(-size, size, size)
    elif dtype is ak.uint64:
        return ak.randint(0, size, size, dtype=dtype)
    elif dtype is ak.float64:
        return ak.randint(-size, size, size, dtype=dtype)
    elif dtype is ak.bool:
        return (ak.randint(0, size, size) % 2) == 0
    elif dtype is ak.bigint:
        return ak.randint(0, size, size, dtype=ak.uint64) + 2**200
    elif dtype is ak.str_:
        return ak.random_strings_uniform(1, 16, size=size)
    return None
```

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,104 @@
...
+        assert pda[np.uint(2)] == pda[2]
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", BIGINT_AND_UINT_TYPES)
+    def test_bool_indexing(self, prob_size, dtype):
+        pda_test = value_array(dtype, prob_size)
+        pda_uint = ak.cast(pda_test, ak.uint64)
+        pda_bool = value_array(ak.bool, prob_size)
+        assert pda_uint[pda_bool].to_list() == pda_test[pda_bool].to_list()
```
pierce314159, line 52: since the left hand side is always `uint` and the right hand side is `uint` or `bool` theres a case where we are essentially comparing

```python
pda_uint[pda_bool].to_list() == pda_uint[pda_bool].to_list()
```

I think a better way of doing this (especially if we update `make_array` to pick more representative examples). Is to just get a random uint array and construct int and bigint from that
```python
def test_bool_indexing(self, prob_size):
    u = make_array(ak.uint64, prob_size)
    b = make_array(ak.bool, prob_size)
    assert u[b].to_list() == ak.cast(u, ak.int64)[b].to_list()
    assert u[b].to_list() == ak.cast(u, ak.bigint)[b].to_list()
```

Notice we no longer have a need for `BIGINT_AND_UINT_TYPES`

pierce314159 (changes requested):

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,98 @@
...
+            assert pda[:prob_size].to_list() == ak.full(prob_size, 2**200, ak.bigint).to_list()
+            pda[:prob_size] = ak.arange(prob_size, dtype=ak.bigint)
+            assert pda[:prob_size].to_list() == ak.arange(prob_size, dtype=ak.uint64).to_list()
+
+            # bigint specific set [pdarray] = scalar/pdarray with uint key pdarray
+            pda[ak.arange(prob_size, dtype=ak.uint64)] = 2**200
+            assert pda[:prob_size].to_list() == ak.full(prob_size, 2**200, ak.bigint).to_list()
+            pda[ak.arange(prob_size)] = ak.arange(prob_size, dtype=ak.bigint)
+            assert pda[:prob_size].to_list() == ak.arange(prob_size, dtype=ak.uint64).to_list()
+
```
pierce314159, line 87: spoke with @jaketrookman about this offline. We are currently special casing for when `dtype == ak.bigint` to make sure we can set values that are representative of `bigint`. But we should really be testing to make sure we can set value that are representative of every dtype

I think let's create a `value_scalar` func that takes in dtype and returns a representative scalar and use that for the scalar, then use `value_array` for setting using `[slice]` and `[pdarray]`

This will result in less code (since we won't have to special case bigint) and better coverage

pierce314159 (changes requested):
Looks good other than my request for `test_set_uint`

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,98 @@
...
+    return ikeys, ukeys
+
+
+def value_array(dtype, size):
+    if dtype is ak.int64:
+        return ak.randint(-size, size, size)
+    elif dtype is ak.uint64:
+        return ak.randint(0, size, size, dtype=dtype)
+    elif dtype is ak.float64:
+        return ak.randint(-size, size, size, dtype=dtype)
```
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,109 @@
...
+        # set [slice] = scalar/pdarray
+        pda[:prob_size] = val
+        assert pda[:prob_size].to_list() == ak.full(prob_size, val, dtype=dtype).to_list()
+        pda[:prob_size] = ak.arange(prob_size, dtype=dtype)
+        assert pda[:prob_size].to_list() == ak.arange(prob_size, dtype=dtype).to_list()
+
+        # set [pdarray] = scalar/pdarray with uint key pdarray
+        pda[ak.arange(prob_size, dtype=ak.uint64)] = val
+        assert pda[:prob_size].to_list() == ak.full(prob_size, val, dtype=dtype).to_list()
+        pda[ak.arange(prob_size)] = ak.arange(prob_size, dtype=dtype)
```
pierce314159, line 96: for setting it to an array, I think instead of `arange` lets use `value_array` that way we're getting more representative examples

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,109 @@
...
+        pda[:test_size] = -2
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+        pda[:test_size] = ak.cast(ak.arange(test_size), dtype)
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+
+        # set [pdarray] = scalar/pdarray with uint key pdarray
+        pda[ak.arange(test_size, dtype=ak.uint64)] = np.uint(3)
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
+        pda[ak.arange(test_size, dtype=ak.uint64)] = ak.cast(ak.arange(test_size), dtype)
+        assert pda[ukeys].to_list() == pda[ikeys].to_list()
```
pierce314159, line 80: I don't think we really need these now that we are covering all dtypes with representative examples

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # set [int] = val with uint key and value
+        val = value_scalar(dtype, prob_size)[0]
+        pda[np.uint(2)] = val
+        assert pda[2] == val
+
+        # set [slice] = scalar/pdarray
+        pda[:prob_size] = val
+        assert pda[:prob_size].to_list() == ak.full(prob_size, val, dtype=dtype).to_list()
+        pda[:prob_size] = ak.arange(prob_size, dtype=dtype)
```
pierce314159, line 84: I wanted you to use `value_array` here too
```suggestion
        pda[:prob_size] = value_array(dtype, prob_size)
```

PROTO_tests/tests/indexing_test.py
```diff
@@ -0,0 +1,103 @@
...
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_bool_indexing(self, prob_size):
+        u = value_array(ak.uint64, prob_size)
+        b = value_array(ak.bool, prob_size)
+        assert u[b].to_list() == ak.cast(u, ak.int64)[b].to_list()
+        assert u[b].to_list() == ak.cast(u, ak.bigint)[b].to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", NUM_TYPES)
+    def test_set_uint(self, prob_size, dtype):
```
pierce314159, line 61: this test fails, prob because we set `pda[ak.arange(prob_size)] = value_array(dtype, prob_size)`, but test that `pda[:prob_size].to_list() == ak.arange(prob_size, dtype=dtype).to_list()`

pierce314159 (changes requested):
`test_set_uint` is failing

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated

Ethan-DeBandi99 (approved):

PR #2655, opened 2023-08-11
Closes #2626 `join_test.py` Conversion for new test framework

### LINKED ISSUE #2626: `join_tests.py` Conversion for new test framework

TheOddCzar789:
This PR (Closes https://github.com/Bears-R-Us/arkouda/issues/2626):

Combined three tests into one via parameterization and renamed some vars to correct case and improve clarity

### DIFF
--- a/PROTO_tests/tests/join_test.py
+++ b/PROTO_tests/tests/join_test.py
@@ -1,0 +1,226 @@
+import numpy as np
+import pytest
+
+import arkouda as ak
+
+"""
+Encapsulates a variety of arkouda join_on_eq_with_dt test cases.
+"""
+
+
+class TestJoin:
+    @classmethod
+    def setup_class(cls):
+        cls.size = 1000
+        cls.a1 = ak.ones(cls.size, dtype=np.int64)
+        cls.a2 = ak.arange(0, cls.size, 1)
+        cls.t1 = cls.a1
+        cls.t2 = cls.a1 * 10
+        cls.dt = 10
+        ak.verbose = False
+
+    @pytest.mark.parametrize("dt_type", ["true_dt", "abs_dt", "pos_dt"])
+    def test_join_on_eq_by_dt(self, dt_type):
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, self.dt, dt_type)
+        nl = ak.get_config()["numLocales"]
+        assert self.size // nl, x.size
+        assert self.size // nl, y.size
+
+    def test_join_on_eq_with_true_dt_with_result_limit(self):
+        nl = ak.get_config()["numLocales"]
+        lim = (self.size + nl) * self.size
+        res_size = self.size * self.size
+        x, y = ak.join_on_eq_with_dt(
+            self.a1, self.a1, self.a1, self.a1, self.dt, "true_dt", result_limit=lim
+        )
+        assert res_size, x.size
+        assert res_size, y.size
+
+    def test_join_on_eq_with_abs_dt_outside_window(self):
+        """
+        Should get 0 answers because N^2 matches but 0 within dt window
+        """
+        dt = 8
+        x, y = ak.join_on_eq_with_dt(self.a1, self.a1, self.t1, self.t1 * 10, dt, "abs_dt")
+        assert 0 == x.size
+        assert 0 == y.size
+
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "abs_dt")
+        assert 0 == x.size
+        assert 0 == y.size
+
+    def test_join_on_eq_with_pos_dt_outside_window(self):
+        """
+        Should get 0 answers because N matches but 0 within dt window
+        """
+        dt = 8
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "pos_dt")
+        assert 0 == x.size
+        assert 0 == y.size
+
+        dt = np.int64(8)
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "pos_dt")
+        assert 0 == x.size
+        assert 0 == y.size
+
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "pos_dt", int(0))
+        assert 0 == x.size
+        assert 0 == y.size
+
+    def test_gen_ranges(self):
+        start = ak.array([0, 10, 20])
+        end = ak.array([10, 20, 30])
+
+        segs, ranges = ak.join.gen_ranges(start, end)
+        assert segs.to_list() == [0, 10, 20]
+        assert ranges.to_list() == list(range(30))
+
+        with pytest.raises(ValueError):
+            segs, ranges = ak.join.gen_ranges(ak.array([11, 12, 41]), end)
+
+    def test_inner_join(self):
+        left = ak.arange(10)
+        right = ak.array([0, 5, 3, 3, 4, 6, 7, 9, 8, 1])
+
+        l, r = ak.join.inner_join(left, right)
+        assert left[l].to_list() == right[r].to_list()
+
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,226 @@
...
+        cls.a2 = ak.arange(0, cls.size, 1)
+        cls.t1 = cls.a1
+        cls.t2 = cls.a1 * 10
+        cls.dt = 10
+        ak.verbose = False
+
+    @pytest.mark.parametrize("dt_type", ["true_dt", "abs_dt", "pos_dt"])
+    def test_join_on_eq_by_dt(self, dt_type):
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, self.dt, dt_type)
+        nl = ak.get_config()["numLocales"]
```
pierce314159, line 25: we should use `pytest.nl` instead. This applies to all the places we declare `nl`

PROTO_tests/tests/join_test.py
pierce314159, line 1: I might be missing them but it seems we dropped several tests from the original `test_join_on_eq_with_true_dt`, `test_join_on_eq_with_abs_dt`, and `test_join_on_eq_with_pos_dt`

PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,226 @@
...
+        Should get 0 answers because N^2 matches but 0 within dt window
+        """
+        dt = 8
+        x, y = ak.join_on_eq_with_dt(self.a1, self.a1, self.t1, self.t1 * 10, dt, "abs_dt")
+        assert 0 == x.size
+        assert 0 == y.size
+
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "abs_dt")
+        assert 0 == x.size
+        assert 0 == y.size
```
pierce314159, line 50: feels like we could reduce some code by looping here and substituting the third from last argument

PROTO_tests/tests/join_test.py
TheOddCzar789: These have all been combined into `test_join_on_eq_by_dt`

PROTO_tests/tests/join_test.py
TheOddCzar789: Updated

PROTO_tests/tests/join_test.py
TheOddCzar789: I've done this in a couple other spots.  Here, both the first and fourth args are different, so I haven't (yet) made this change.  Please review the loops added in the latest commit and let me know your thoughts on them and this one.

PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,214 @@
...
+        cls.t1 = cls.a1
+        cls.t2 = cls.a1 * 10
+        cls.dt = 10
+        ak.verbose = False
+
+    @pytest.mark.parametrize("dt_type", ["true_dt", "abs_dt", "pos_dt"])
+    def test_join_on_eq_by_dt(self, dt_type):
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, self.dt, dt_type)
+        assert self.size // pytest.nl, x.size
+        assert self.size // pytest.nl, y.size
```
pierce314159, line 26: so this actually isn't testing anything. It needs to be `assert self.size // pytest.nl == x.size`. Otherwise it's just checking the truthyness of `self.size // pytest.nl` which will be true as long as it's non-zero

To show this:
```python
# this outputs nothing because 17 is truthy
>>> assert 17, 9

# raises an assertion error because 17 == 9 is false
>>> assert 17 == 9
AssertionError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 assert 17 == 9

AssertionError:

# raises an assertion error because 0 is falsey
>>> assert 0, 17
AssertionError                            Traceback (most recent call last)
Cell In[7], line 1
----> 1 assert 0, 17

AssertionError: 17
```

PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def test_join_on_eq_with_true_dt_with_result_limit(self):
+        lim = (self.size + pytest.nl) * self.size
+        res_size = self.size * self.size
+        x, y = ak.join_on_eq_with_dt(
+            self.a1, self.a1, self.a1, self.a1, self.dt, "true_dt", result_limit=lim
+        )
+        assert res_size, x.size
+        assert res_size, y.size
```
pierce314159, line 35: this has the same issue of needing `==`

PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,214 @@
...
+        assert 0 == y.size
+
+    def test_join_on_eq_with_pos_dt_outside_window(self):
+        """
+        Should get 0 answers because N matches but 0 within dt window
+        """
+        for dt in 8, np.int64(8):
+            x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "pos_dt")
+            assert 0 == x.size
+            assert 0 == y.size
```
pierce314159, line 57: since this is just a python comparison we can just do
```python
assert 0 == x.size == y.size
```

This is applicable in a couple other spots in this file where `x.size` and `y.size` are being compared to the same value

PROTO_tests/tests/join_test.py
pierce314159: Yah I'm not sure it's worth it, let's hold off on it

pierce314159 (changes requested):
pretty close, the only thing that for sure needs to be changed is updating the couple of spots to use `==` for the asserts

PROTO_tests/tests/join_test.py
TheOddCzar789: Updated

PROTO_tests/tests/join_test.py
TheOddCzar789: Updated

PROTO_tests/tests/join_test.py
TheOddCzar789: Updated

jaketrookman (approved):

PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,209 @@
...
+    def test_join_on_eq_with_abs_dt_outside_window(self):
+        """
+        Should get 0 answers because N^2 matches but 0 within dt window
+        """
+        dt = 8
+        x, y = ak.join_on_eq_with_dt(self.a1, self.a1, self.t1, self.t1 * 10, dt, "abs_dt")
+        assert 0 == x.size == y.size
+
+        x, y = ak.join_on_eq_with_dt(self.a2, self.a1, self.t1, self.t2, dt, "abs_dt")
+        assert 0 == x.size == y.size
```
pierce314159, line 45: So I know we said we'd hold off on turning this into a loop since 2 things were changing (1st and 4th arguments), but looking at this a bit more closely I found that for the fourth argument, we have `t2` and `t1 * 10`. Since `t2 = a1 * 10` and `t1 = a1`, these are equivalent. So we should be able to fix that as `t2` and loop only changing the first argument

PROTO_tests/tests/join_test.py
```diff
@@ -0,0 +1,209 @@
...
+Encapsulates a variety of arkouda join_on_eq_with_dt test cases.
+"""
+
+
+class TestJoin:
+    @classmethod
+    def setup_class(cls):
+        cls.size = 1000
+        cls.a1 = ak.ones(cls.size, dtype=np.int64)
+        cls.a2 = ak.arange(0, cls.size, 1)
```
pierce314159, line 16: we don't have to do the first and last arguments here (stop = 0 and stride =1 by default)
```suggestion
        cls.a2 = ak.arange(cls.size)
```

PROTO_tests/tests/join_test.py
TheOddCzar789: Updated

Ethan-DeBandi99 (approved):
Looks good to me

pierce314159 (approved):

PR #2669, opened 2023-08-15
Closes #2654 `sort_test.py` conversion for new test framework

### LINKED ISSUE #2654: `sort_test.py` Conversion for new test framework

TheOddCzar789:
This PR (Closes https://github.com/Bears-R-Us/arkouda/issues/2654):

Added some looping for brevity and added sorting of floats.

### DIFF
--- a/PROTO_tests/tests/sort_test.py
+++ b/PROTO_tests/tests/sort_test.py
@@ -11,13 +11,24 @@ def make_ak_arrays(dtype):
     if dtype in ["int64", "uint64"]:
         return ak.randint(0, 100, 1000, dtype)
     elif dtype == "float64":
-        return ak.array(np.random.rand(100) * 10000)
+        return ak.array(np.random.rand(1000) * 10000)
     elif dtype == "bigint":
         return ak.randint(0, 100, 1000, dtype=ak.uint64)
     return None
 
 
+def np_is_sorted(arr):
+    return np.all(arr[:-1] <= arr[1:])
+
+
 class TestSort:
+    @pytest.mark.parametrize("dtype", NUMERIC_AND_BIGINT_TYPES)
+    def test_is_sorted(self, dtype):
+        pda = make_ak_arrays(dtype)
+        sorted_pda = ak.sort(pda)
+        assert ak.is_sorted(sorted_pda)
+        assert np_is_sorted(sorted_pda.to_ndarray())
+
     @pytest.mark.parametrize("algo", SortingAlgorithm)
     @pytest.mark.parametrize("dtype", NUMERIC_AND_BIGINT_TYPES)
     def test_sort_dtype(self, algo, dtype):
--- a/PROTO_tests/tests/sort_test.py
+++ b/PROTO_tests/tests/sort_test.py
@@ -4,68 +4,71 @@ import pytest
 import arkouda as ak
 from arkouda.sorting import SortingAlgorithm
 
+NUMERIC_AND_BIGINT_TYPES = ["int64", "float64", "uint64", "bigint"]
+
+
+def make_ak_arrays(dtype):
+    if dtype in ["int64", "uint64"]:
+        return ak.randint(0, 100, 1000, dtype)
+    elif dtype == "float64":
+        return ak.array(np.random.rand(100) * 10000)
+    elif dtype == "bigint":
+        return ak.randint(0, 100, 1000, dtype=ak.uint64)
+    return None
+
 
 class TestSort:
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/sort_test.py
```diff
@@ -0,0 +1,71 @@
...
+        for algo in SortingAlgorithm:
+            spda = ak.sort(pda_float, algo)
+            assert ak.is_sorted(spda)
+
+        pda = ak.randint(0, 100, size, dtype=ak.uint64)
+        pda_bigint = pda + 2**200
+        for algo in SortingAlgorithm:
+            sorted_pda = ak.sort(pda, algo)
+            sorted_bi = ak.sort(pda_bigint, algo)
+            assert (sorted_bi - 2**200).to_list() == sorted_pda.to_list()
```
pierce314159, line 27: I don't think it's super important to parameterize on `size` (especially if we're only drawing from `0-100` no matter what), but I think parameterizing on `algo` and `dtype` would make this way more compressed

Every function in this test should be able to parameterized on `algo`. If you want an example of how to parameterize on `algo` and `dtype`, I already converted `coargsort` and I think it does something similar

pierce314159 (changes requested):
I think we should parameterize on `algo` and `dtype` like we do in the `coargsort` test that was already converted

PROTO_tests/tests/sort_test.py
TheOddCzar789: Updated

PROTO_tests/tests/sort_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class TestSort:
+    @pytest.mark.parametrize("algo", SortingAlgorithm)
+    @pytest.mark.parametrize("dtype", NUMERIC_AND_BIGINT_TYPES)
+    def test_sort_dtype(self, algo, dtype):
+        pda = make_ak_arrays(dtype)
+
+        if dtype != "bigint":
+            assert ak.is_sorted(ak.sort(pda, algo))
```
Ethan-DeBandi99, line 27: If we are going to use `ak.is_sorted` to validate `ak.sort`, we need to validate `ak.is_sorted` is functioning properly first. A test should be added for `ak.is_sorted` or after this assertion we should validate that the result is seen as sorted by NumPy. Personally, I am leaning towards a combination of both adding a test for `ak.is_sorted` and always validating against NumPy. @pierce314159 what are your thoughts?

It looks like if we add the `NumPy` validation it will need to be done in all tests.

Ethan-DeBandi99 (changes requested):
We are using something that is not validated to validate other functionality. I provided 2 options forward (I am leaning towards a combo though).

PROTO_tests/tests/sort_test.py
pierce314159: I agree I think definitely validate `ak.is_sorted` is correct. I'm also down with always comparing against numpy

pierce314159 (approved):
Looks good to me other than ethan's comment

PROTO_tests/tests/sort_test.py
TheOddCzar789: I added `np_is_sorted` and `test_is_sorted` for review, but I'm not sure it does quite what is being asked.  I could not find a cleaner way to verify sortedness in numpy.

Ethan-DeBandi99: I think the added test is achieving what we want on the testing of `ak.is_sorted()`. From a quick search the helper method does appear to be doing what is considered the accepted check within NumPy to determine if an array is sorted. I am going back and forth on whether or not we should  be using `np_is_sorted(ak_array.to_ndarray())` in place of `ak.is_sorted` for the other validations. I may be splitting hairs. @pierce314159 - what are your thoughts? I am torn between covering everything and saying that if we test `ak.is_sorted` it is acceptable to use for other validations.

pierce314159: I get where you're coming from. I think if we did some edge_case testing with `is_sorted` I'd be comfortable only using that. But given https://github.com/Bears-R-Us/arkouda/issues/2703, I'm somewhat doubtful it would behave correctly. I think either is good for now and I'll revisit it during #2703

Ethan-DeBandi99: I am fine with pushing this for now since we have an issue to revisit. I'll approve it. Thanks for weighing in @pierce314159.

Ethan-DeBandi99 (approved):

PR #2671, opened 2023-08-15
Closed #2651: `series_test.py` conversion for new framework

### LINKED ISSUE #2651: `series_test.py` Conversion for new test framework
Convert `series_test.py` to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

jaketrookman:
This PR (Closes #2651) updates `series_test.py` to the new testing framework.

*This is just a direct conversion. I would appreciate some help identifying any places where some parameterization could be added.

### DIFF
--- a/PROTO_tests/tests/series_test.py
+++ b/PROTO_tests/tests/series_test.py
@@ -82,21 +82,19 @@ class TestSeries:
                 index=ak.arange(prob_size // 2, prob_size),
             )
         )
-        assert added.index.to_list() == ak.arange(prob_size).to_list()
-        assert added.values.to_list() == ak.arange(prob_size, dtype=dtype).to_list()
+        assert (added.index == ak.arange(prob_size)).all()
+        assert (added.values == ak.arange(prob_size, dtype=dtype)).all()
 
     @pytest.mark.parametrize("prob_size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64, ak.float64])
     def test_topn(self, prob_size, dtype):
-        prob_size_half = prob_size // 2
-        top = ak.Series(ak.arange(prob_size, dtype=dtype)).topn(prob_size_half)
-        assert top.values.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
-        assert top.index.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
+        top = ak.Series(ak.arange(100, dtype=dtype)).topn(50)
+        assert top.values.to_list() == list(range(99, 49, -1))
+        assert top.index.to_list() == list(range(99, 49, -1))
 
-    @pytest.mark.parametrize("prob_size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", NUMERICAL_TYPES)
     @pytest.mark.parametrize("dtype_index", NUMERICAL_TYPES)
-    def test_sort(self, prob_size, dtype, dtype_index):
+    def test_sort(self, dtype, dtype_index):
         def gen_perm(n):
             idx_left = set(range(n))
             perm = np.arange(n)
@@ -109,8 +107,8 @@ class TestSeries:
                 idx_left.remove(inds[1])
             return perm
 
-        ordered = ak.arange(prob_size, dtype=dtype)
-        perm = ak.array(gen_perm(prob_size), dtype=dtype_index)
+        ordered = ak.arange(100, dtype=dtype)
+        perm = ak.array(gen_perm(100), dtype=dtype_index)
 
         idx_sort = ak.Series(data=ordered, index=perm).sort_index()
         assert idx_sort.index.to_list() == ordered.to_list()
@@ -125,7 +123,7 @@ class TestSeries:
     def test_head_tail(self, prob_size, dtype):
         s = ak.Series(ak.arange(prob_size, dtype=dtype))
 
-        for i in range(prob_size):
+        for i in range(1, prob_size, 5):
             head = s.head(i)
             assert head.index.to_list() == list(range(i))
             assert head.values.to_list() == ak.arange(i, dtype=dtype).to_list()
@@ -183,11 +181,8 @@ class TestSeries:
                 assert isinstance(df, pd.DataFrame)
                 assert_frame_equal(ref_df, df)
 
-    @pytest.mark.parametrize("prob_size", pytest.prob_size)
-    def test_index_as_index_compat(self, prob_size):
+    def test_index_as_index_compat(self):
         # added to validate functionality for issue #1506
-        df = ak.DataFrame(
-            {"a": ak.arange(prob_size), "b": ak.arange(prob_size), "c": ak.arange(prob_size)}
-        )
+        df = ak.DataFrame({"a": ak.arange(10), "b": ak.arange(10), "c": ak.arange(10)})
         g = df.groupby(["a", "b"])
         g.broadcast(g.sum("c"))
--- a/PROTO_tests/tests/series_test.py
+++ b/PROTO_tests/tests/series_test.py
@@ -85,9 +85,8 @@ class TestSeries:
         assert (added.index == ak.arange(prob_size)).all()
         assert (added.values == ak.arange(prob_size, dtype=dtype)).all()
 
-    @pytest.mark.parametrize("prob_size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64, ak.float64])
-    def test_topn(self, prob_size, dtype):
+    def test_topn(self, dtype):
         top = ak.Series(ak.arange(100, dtype=dtype)).topn(50)
         assert top.values.to_list() == list(range(99, 49, -1))
         assert top.index.to_list() == list(range(99, 49, -1))
--- a/PROTO_tests/tests/series_test.py
+++ b/PROTO_tests/tests/series_test.py
@@ -1,0 +1,193 @@
+import numpy as np
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
+
+import arkouda as ak
+
+DTYPES = [ak.int64, ak.uint64, ak.bool, ak.float64, ak.bigint, ak.str_]
+NO_STRING = [ak.int64, ak.uint64, ak.bool, ak.float64, ak.bigint]
+NUMERICAL_TYPES = [ak.int64, ak.uint64, ak.float64, ak.bigint]
+INTEGRAL_TYPES = [ak.int64, ak.uint64, ak.bool, ak.bigint]
+
+
+class TestSeries:
+    @pytest.mark.parametrize("dtype", DTYPES)
+    def test_series_creation(self, dtype):
+        idx = ak.arange(3, dtype=dtype)
+        for val in idx, ak.array(["A", "B", "C"]):
+            ans = ak.Series(data=val, index=idx).to_list()
+            for series in (
+                ak.Series(data=val, index=idx),
+                ak.Series(data=val),
+                ak.Series(val, idx),
+                ak.Series(val),
+                ak.Series((idx, val)),
+            ):
+                assert isinstance(series, ak.Series)
+                assert isinstance(series.index, ak.Index)
+                assert series.to_list() == ans
+
+        with pytest.raises(TypeError):
+            ak.Series(index=idx)
+
+        with pytest.raises(TypeError):
+            ak.Series((ak.arange(3),))
+
+        with pytest.raises(TypeError):
+            ak.Series()
+
+        with pytest.raises(ValueError):
+            ak.Series(data=ak.arange(3), index=ak.arange(6))
+
+    @pytest.mark.parametrize("dtype", INTEGRAL_TYPES)
+    @pytest.mark.parametrize("dtype_index", [ak.int64, ak.uint64])
+    def test_lookup(self, dtype, dtype_index):
+        pda = ak.arange(3, dtype=dtype)
+        for val in pda, ak.array(["A", "B", "C"]):
+            s = ak.Series(data=val, index=ak.arange(3, dtype=dtype_index))
+
+            for key in (
+                1,
+                ak.Index(ak.array([1], dtype=dtype_index)),
+                ak.Index(ak.array([0, 2], dtype=dtype_index)),
+            ):
+                lk = s.locate(key)
+                assert isinstance(lk, ak.Series)
+                key = ak.array(key) if not isinstance(key, int) else key
+                assert (lk.index == s.index[key]).all()
+                assert (lk.values == s.values[key]).all()
+
+            # testing multi-index lookup
+            mi = ak.MultiIndex([pda, pda[::-1]])
+            s = ak.Series(data=val, index=mi)
+            lk = s.locate(mi[0])
+            assert isinstance(lk, ak.Series)
+            assert lk.index.index == mi[0].index
+            assert lk.values[0] == val[0]
+
+            # ensure error with scalar and multi-index
+            with pytest.raises(TypeError):
+                s.locate(0)
+
+            with pytest.raises(TypeError):
+                s.locate([0, 2])
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", NO_STRING)
+    def test_add(self, prob_size, dtype):
+        added = ak.Series(ak.arange(prob_size // 2, dtype=dtype)).add(
+            ak.Series(
+                data=ak.arange(prob_size // 2, prob_size, dtype=dtype),
+                index=ak.arange(prob_size // 2, prob_size),
+            )
+        )
+        assert added.index.to_list() == ak.arange(prob_size).to_list()
+        assert added.values.to_list() == ak.arange(prob_size, dtype=dtype).to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64, ak.float64])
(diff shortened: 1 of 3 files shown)


PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,224 @@
...
+                "val_0": [0, 1, 2, 3, 4, 0, 0, 0, 0, 0, 0],
+                "val_1": [0, 0, 0, 0, 0, 5, 6, 7, 8, 9, 10],
+            }
+        )
+        assert ((ref_df == df.to_pandas()).all()).all()
+
+    def test_pdconcat(self):
+        v = ak.arange(5)
+        i = ak.arange(5)
+        s = ak.Series(data=v, index=i)
```
TheOddCzar789, line 199: I notice each pair of (v, i) vars in this function are the same.  Should we combine them instead of assigning the same values twice?

jaketrookman: Updated

TheOddCzar789 (approved):

pierce314159:
I added some updates to this, marking as a draft until those make it in

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,150 @@
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
+
+import arkouda as ak
+
+
+class TestSeries:
+    def test_series_creation(self):
```
Ethan-DeBandi99, line 9: `ak.Series` can have pdarray, Strings and Categorical values. I would add a file level global

```python
DTYPES = [ak.int64, ak.uint64, ak.bool, ak.float64, ak.bigint, ak.str_]
```
and then update this to function definition to parameterize on `DTYPES`. Also, add an `if` block for `dtype == ak.str_` to test `Categorical`. 

Then, we should split the error handling testing into its own function.

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,150 @@
...
+        with pytest.raises(TypeError):
+            ak.Series((ak.arange(3),))
+
+        with pytest.raises(TypeError):
+            ak.Series()
+
+        with pytest.raises(ValueError):
+            ak.Series(data=ak.arange(3), index=ak.arange(6))
+
+    def test_lookup(self):
```
Ethan-DeBandi99, line 36: Again, we should parameterize on the allowed `DTYPE` values and handle `Categorical` as a special case of `ak.str_`.

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,150 @@
...
+            assert lk.values[0] == val[0]
+
+            # ensure error with scalar and multi-index
+            with pytest.raises(TypeError):
+                s.locate(0)
+
+            with pytest.raises(TypeError):
+                s.locate([0, 2])
+
+    def test_add(self):
```
Ethan-DeBandi99, line 62: This also should be tested for all `DTYPES`.

PROTO_tests/tests/series_test.py
Ethan-DeBandi99, line 1: I marked a few tests specifically, but most of these need to be updated to parameterize on `DTYPES` and `Categorical` for values. I am also thinking we need to do a 2nd `DTYPE` parameter to handle different index value types as well.

It may also be worth while to parameterize on `pytest.prob_size` as well.

Ethan-DeBandi99 (changes requested):
Mostly looked at parameterization for better coverage.

PROTO_tests/tests/series_test.py
jaketrookman: Added `dtype` testing for several functions where applicable. After going through some issues with @pierce314159 , there are some larger issues that can be readdressed later with the handling of `Categorical` for strings on specific tests. It is properly handled in some cases, in those cases it is tested.

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,188 @@
...
+        assert added.values.to_list() == ak.arange(prob_size, dtype=dtype).to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.float64])
+    # @pytest.mark.parametrize("dtype", NUMERICAL_TYPES)
+    def test_topn(self, prob_size, dtype):
+        prob_size_half = prob_size // 2
+        top = ak.Series(ak.arange(prob_size, dtype=dtype)).topn(prob_size_half)
+        assert top.values.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
+        assert top.index.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
```
jaketrookman, line 90: I do need some help with this test. For `uint` and `bigint` there is an error that shows up on the messaging from the client. It shows up as an unsupported type, but I can not find where they are actually being called in the messaging.

Ethan-DeBandi99: @jaketrookman - This is on the server side processing of `argmaxk` used by `topn`. If you look in `src/KExtremeMsg.chpl` at `maxkMsg`, you will see that it is currently only defined for `int64` and `float64`. Thus, that function will need to be updated in order to support `uint` (which PR #2719 appears to handle) and `bigint` or the test updated to not run on the unsupported types.

PROTO_tests/tests/series_test.py
Ethan-DeBandi99: Is #2717 part of this? If there are issues related to it, then Issues need to be added to GitHub and linked here so that they are tracked and can be addressed if they will not be resolved as part of this PR.

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,188 @@
...
+            ak.Series((ak.arange(3),))
+
+        with pytest.raises(TypeError):
+            ak.Series()
+
+        with pytest.raises(ValueError):
+            ak.Series(data=ak.arange(3), index=ak.arange(6))
+
+    @pytest.mark.parametrize("dtype", INTEGRAL_TYPES)
+    def test_lookup(self, dtype):
```
Ethan-DeBandi99, line 44: This still is not testing with different `dtypes` for the index.

PROTO_tests/tests/series_test.py
Ethan-DeBandi99: Added a new comment, but this only tests different `dtypes` on values now. It still needs to test different `dtypes` for the index.

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,188 @@
...
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.float64])
+    # @pytest.mark.parametrize("dtype", NUMERICAL_TYPES)
+    def test_topn(self, prob_size, dtype):
+        prob_size_half = prob_size // 2
+        top = ak.Series(ak.arange(prob_size, dtype=dtype)).topn(prob_size_half)
+        assert top.values.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
+        assert top.index.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", NUMERICAL_TYPES)
```
Ethan-DeBandi99, line 93: This should have a separate parameter for the index dtype because we do not only want to test when the index type matches the value type

Ethan-DeBandi99 (changes requested):
Few things still are not all the way there.

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated, but as there is no current implementation of `min/max reduce` for type `bigint`. An issue has been created to look into future implementation.

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,193 @@
...
+        assert val_sort.index.to_pandas().tolist() == perm.to_list()
+        assert val_sort.values.to_list() == ordered.to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", DTYPES)
+    def test_head_tail(self, prob_size, dtype):
+        s = ak.Series(ak.arange(prob_size, dtype=dtype))
+
+        for i in range(prob_size):
+            head = s.head(i)
```
pierce314159, line 129: this is should not be parmeterized on problem size. This will take forever for even a moderate problem prob size (say 1000) because we are doing thousands of arkouda operations and array creations, not even mentioning doing `list(range(i))` for every i less than prob_size. I wouldn't do this for any more than `n=10`

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,193 @@
...
+        assert added.index.to_list() == ak.arange(prob_size).to_list()
+        assert added.values.to_list() == ak.arange(prob_size, dtype=dtype).to_list()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64, ak.float64])
+    def test_topn(self, prob_size, dtype):
+        prob_size_half = prob_size // 2
+        top = ak.Series(ak.arange(prob_size, dtype=dtype)).topn(prob_size_half)
+        assert top.values.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
+        assert top.index.to_list() == list(range(prob_size - 1, prob_size_half - 1, -1))
```
pierce314159, line 94: This also shouldn't really be parameterized on `prob_size` because `list(range(prob_size))` can quickly get too large for the client to handle

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,193 @@
...
+                assert isinstance(df, pd.DataFrame)
+                assert_frame_equal(ref_df, df)
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_index_as_index_compat(self, prob_size):
+        # added to validate functionality for issue #1506
+        df = ak.DataFrame(
+            {"a": ak.arange(prob_size), "b": ak.arange(prob_size), "c": ak.arange(prob_size)}
+        )
+        g = df.groupby(["a", "b"])
```
pierce314159, line 192: we aren't gaining anything by testing prob_size here

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,193 @@
...
+    @pytest.mark.parametrize("dtype", NO_STRING)
+    def test_add(self, prob_size, dtype):
+        added = ak.Series(ak.arange(prob_size // 2, dtype=dtype)).add(
+            ak.Series(
+                data=ak.arange(prob_size // 2, prob_size, dtype=dtype),
+                index=ak.arange(prob_size // 2, prob_size),
+            )
+        )
+        assert added.index.to_list() == ak.arange(prob_size).to_list()
+        assert added.values.to_list() == ak.arange(prob_size, dtype=dtype).to_list()
```
pierce314159, line 86: since we are parameterizing on `prob_size` this might get too big to call `.to_list`. I would suggest we do this instead
```python
assert (added.index == ak.arange(prob_size)).all()
assert (added.values == ak.arange(prob_size, dtype=dtype)).all()
```

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,193 @@
...
+        ordered = ak.arange(prob_size, dtype=dtype)
+        perm = ak.array(gen_perm(prob_size), dtype=dtype_index)
+
+        idx_sort = ak.Series(data=ordered, index=perm).sort_index()
+        assert idx_sort.index.to_list() == ordered.to_list()
+        assert idx_sort.values.to_list() == perm.to_list()
+
+        val_sort = ak.Series(data=perm, index=ordered).sort_values()
+        assert val_sort.index.to_pandas().tolist() == perm.to_list()
+        assert val_sort.values.to_list() == ordered.to_list()
```
pierce314159, line 121: same deal, we might want to avoid `to_list` if we are parameterizing on `prob_size`

pierce314159 (changes requested):

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: As `binopvvMsg` does not handle `bigint` and `float64` due to ambiguity, after discussing with @pierce314159, we are going to keep this structure but remove the parameterizing on `prob_size`.

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", DTYPES)
+    def test_head_tail(self, prob_size, dtype):
+        s = ak.Series(ak.arange(prob_size, dtype=dtype))
+
+        for i in range(1, prob_size, 5):
+            head = s.head(i)
```
pierce314159, line 127: this is the same concern as my last review just on a different test. If we have a large problem size this is just not feasible

pierce314159 (changes requested):

PROTO_tests/tests/series_test.py
jaketrookman: Updated

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,187 @@
...
+        added = ak.Series(ak.arange(prob_size // 2, dtype=dtype)).add(
+            ak.Series(
+                data=ak.arange(prob_size // 2, prob_size, dtype=dtype),
+                index=ak.arange(prob_size // 2, prob_size),
+            )
+        )
+        assert (added.index == ak.arange(prob_size)).all()
+        assert (added.values == ak.arange(prob_size, dtype=dtype)).all()
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
```
TheOddCzar789, line 88: This can be removed as it's not longer used (and removed from the def parameter list)

TheOddCzar789 (approved):
All I see is the leftover prob_size parameter

PROTO_tests/tests/series_test.py
jaketrookman: Good catch, updated

TheOddCzar789 (approved):

PROTO_tests/tests/series_test.py
```diff
@@ -0,0 +1,186 @@
...
+        size = 100
+        s = ak.Series(ak.arange(size, dtype=dtype))
+
+        head = s.head(size)
+        assert head.index.to_list() == list(range(size))
+        assert head.values.to_list() == ak.arange(size, dtype=dtype).to_list()
+
+        tail = s.tail(size)
+        assert tail.index.to_list() == ak.arange(size)[-size:size].to_list()
+        assert tail.values.to_list() == ak.arange(size, dtype=dtype)[-size:size].to_list()
```
pierce314159, line 131: the `-size:size` thing doesn't really make sense to me. It's just the same as `ak.arange`
```python
In [4]: size = 10

In [5]: ak.arange(size)[-size:size]
Out[5]: array([0 1 2 3 4 5 6 7 8 9])
```

I say we just go back to what I put in the PR in ur fork

```suggestion
    def test_head_tail(self, dtype):
        n = 10
        s = ak.Series(ak.arange(n, dtype=dtype))

        for i in range(n):
            head = s.head(i)
            assert head.index.to_list() == list(range(i))
            assert head.values.to_list() == ak.arange(i, dtype=dtype).to_list()

            tail = s.tail(i)
            assert tail.index.to_list() == ak.arange(n)[-i:n].to_list()
            assert tail.values.to_list() == ak.arange(n, dtype=dtype)[-i:n].to_list()
```

arkouda/index.py
```diff
@@ -806,5 +806,9 @@ def lookup(self, key):
             raise TypeError("MultiIndex lookup failure")
         # if individual vals convert to pdarrays
         if not isinstance(key[0], pdarray):
-            key = [array([x]) for x in key]
+            from arkouda.numeric import cast as akcast
```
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/series_test.py
jaketrookman: Updated

pierce314159 (approved):

PR #2673, opened 2023-08-15
Closes #2672, #2662 - Update `CommPrimitives` functions and precision in `test_str_repr`

### LINKED ISSUE #2662: Update precision in `test_str_repr` test
In https://github.com/chapel-lang/chapel/pull/22924, the default precision for printing `real` values in Chapel was changed, and this impacts one of the Arkouda tests that will need to be updated.

### LINKED ISSUE #2672: Replace primitive calls to `chpl_comm_get` and `chpl_comm_put`
In https://github.com/chapel-lang/chapel/pull/22884 Chapel deprecated and removed sending `c_ptr` and `ddata` as actuals to the primitive calls `chpl_comm_get` and `chpl_comm_put`. These calls were exposed in Arkouda through the `CommPrimitives` module, and guidance from @ronawho suggested that removing and replacing these calls by using the Chapel standard module `Communication` would be appropriate. The `Communication` module was introduced in Chapel 1.28, so it should be OK to introduce this dependency without any `compat` changes.

arezaii:
This PR resolves #2672 by replacing the use of Chapel `__primitive` calls for performing bulk `get`/`put` operations with calls to the Chapel `Communication` module, introduced in Chapel 1.28. 
This change is only for backend calls to `PUT` and `GET`, and should have no effect on any Arkouda interface.

While here, also updated the expected results in `operator_tests::test_str_repr` to resolve #2662.

TESTING:

- [x] make check
- [x] make test
- [x] make mypy
- [x] flake8 arkouda

Reviewer should note:
(also removed a bunch of trailing spaces)

### DIFF
--- a/src/AryUtil.chpl
+++ b/src/AryUtil.chpl
@@ -16,20 +16,20 @@ module AryUtil
     param bitsPerDigit = RSLSD_bitsPerDigit;
     private param numBuckets = 1 << bitsPerDigit; // these need to be const for comms/performance reasons
     private param maskDigit = numBuckets-1; // these need to be const for comms/performance reasons
-    
+
     private config const logLevel = ServerConfig.logLevel;
     private config const logChannel = ServerConfig.logChannel;
     const auLogger = new Logger(logLevel, logChannel);
-    
+
     /*
-      Threshold for the amount of data that will be printed. 
+      Threshold for the amount of data that will be printed.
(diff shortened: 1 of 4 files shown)


pierce314159 (approved):
Looks good to me!

bmcdonald3 (approved):

jaketrookman (approved):

PR #2692, opened 2023-08-18
Closes #2690 - `ak.load_all` glob expression update

### LINKED ISSUE #2690: `ak.load_all` glob expression update
The current glob expression used by `ak.load_all` is greedy. It has been requested that this be updated to only load the files with the name as specified. The existing glob expression is `filename*`. This needs to be updated to `filename_LOCALE*` to get the desired functionality. 

For reference, the example case given was stated as below:

If a group of files, `my_data`, and another group of files, `my_data_tests`, are in the same directory and `ak.load_all("my_data")` is called, the result is a dataset loading all data from `my_data_LOCALE*` and `my_data_tests_LOCALE*`. The desired functionality is to only load `my_data_LOCALE*`.

Ethan-DeBandi99:
Closes #2690

Updates the glob expression from `filename*` to `filename_LOCALE*`. The previous functionality is still available through passing `filename*` to `ak.read()`. This also adds an error message to ensure that the user does not pass a glob expression to `ak.load()`.

### DIFF
--- a/arkouda/io.py
+++ b/arkouda/io.py
@@ -1572,7 +1572,8 @@ def load(
     with a .parquet appended to the prefix_path.
     Parquet files were previously ALWAYS stored with a ``.parquet`` extension.
 
-    This function will be deprecated when glob flags are added to read_* functions
+    ak.load does not support loading a single file. This function automatically creates a glob expression.
+    For loading single HDF5 files without the _LOCALE#### suffix please use ak.read().
 
     CSV files without the Arkouda Header are not supported.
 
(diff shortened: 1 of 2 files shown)


arkouda/io.py
```diff
@@ -1572,7 +1572,8 @@ def load(
     with a .parquet appended to the prefix_path.
     Parquet files were previously ALWAYS stored with a ``.parquet`` extension.
 
-    This function will be deprecated when glob flags are added to read_* functions
+    ak.load does not support loading a single file.
+    For loading single HDF5 files without the _LOCALE#### suffix please use ak.read().
```
jaketrookman, line 1576: Should this be "use ak.read_hdf()"?

Ethan-DeBandi99: No, `ak.load` can read HDF5, parquet, and CSV. Thus, we want this to be pointed to the generic `read` as well. HDF5 is just the one currently supporting writing to a single file, but all should support reading single files.

jaketrookman: Makes sense. Thank you for clarifying.

jaketrookman (approved):
Looks good to me

pierce314159 (approved):
Looks good!

PR #2693, opened 2023-08-18
Closes #2688: `operator_test.py` conversion for new test framework

### LINKED ISSUE #2688: `operator_test.py` conversion to new framework
Convert `operator_test.py` to use new test framework

pierce314159:
This PR (closes #2688) converts `operator_test.py` to new test framework

### DIFF
--- a/PROTO_tests/tests/operator_test.py
+++ b/PROTO_tests/tests/operator_test.py
@@ -1,0 +1,774 @@
+import warnings
+from itertools import product
+
+import numpy as np
+import pytest
+
+import arkouda as ak
+
+NUMERIC_TYPES = ["int64", "float64", "bool", "uint64"]
+
+
+def make_np_arrays(size, dtype):
+    if dtype == "int64":
+        return np.random.randint(-(2**32), 2**32, size=size, dtype=dtype)
+    elif dtype == "uint64":
+        return np.random.randint(-(2**32), 2**32, size=size).astype(dtype)
+    elif dtype == "float64":
+        return np.random.uniform(-(2**32), 2**32, size=size)
+    elif dtype == "bool":
+        return np.random.randint(0, 1, size=size, dtype=dtype)
+    return None
+
+
+class TestOperator:
+    def test_numpy_equivalency(self, size=100, verbose=pytest.verbose):
+        # ignore numpy warnings like divide by 0
+        np.seterr(all="ignore")
+        global pdarrays
+        pdarrays = {
+            "int64": ak.arange(0, size, 1),
+            "uint64": ak.array(np.arange(2**64 - size, 2**64, 1, dtype=np.uint64)),
+            "float64": ak.linspace(-2, 2, size),
+            "bool": (ak.arange(0, size, 1) % 2) == 0,
+        }
+        global ndarrays
+        ndarrays = {
+            "int64": np.arange(0, size, 1),
+            "uint64": np.arange(2**64 - size, 2**64, 1, dtype=np.uint64),
+            "float64": np.linspace(-2, 2, size),
+            "bool": (np.arange(0, size, 1) % 2) == 0,
+        }
+        global scalars
+        scalars = {"int64": 5, "uint64": np.uint64(2**63 + 1), "float64": -3.14159, "bool": True}
+        dtypes = pdarrays.keys()
+        if verbose:
+            print("Operators: ", ak.pdarray.BinOps)
+            print("Dtypes: ", dtypes)
+            print("pdarrays: ")
+            for k, v in pdarrays.items():
+                print(k, ": ", v)
+            print("ndarrays: ")
+            for k, v in ndarrays.items():
+                print(k, ": ", v)
+            print("scalars: ")
+            for k, v in scalars.items():
+                print(k, ": ", v)
+
+        def do_op(lt, rt, ls, rs, isarkouda, oper):
+            evalstr = ""
+            if ls:
+                evalstr += f'scalars["{lt}"]'
+            else:
+                evalstr += f'{("ndarrays", "pdarrays")[isarkouda]}["{lt}"]'
+            evalstr += f" {oper} "
+            if rs:
+                evalstr += f'scalars["{rt}"]'
+            else:
+                evalstr += f'{("ndarrays", "pdarrays")[isarkouda]}["{rt}"]'
+            res = eval(evalstr)
+            return res
+
+        results = {
+            "neither_implement": [],  # (expression, ak_error)
+            "arkouda_minus_numpy": [],  # (expression, ak_result, error_on_exec?)
+            "numpy_minus_arkouda": [],  # (expression, ak_result, error_on_exec?)
+            "both_implement": [],
+        }  # (expression, ak_result, error_on_exec?, dtype_mismatch?, value_mismatch?)
+        tests = 0
+        for ltype, rtype, op in product(dtypes, dtypes, ak.pdarray.BinOps):
+            if op in ("<<<", ">>>"):
+                continue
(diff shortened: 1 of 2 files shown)


PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
+    elif dtype == "bool":
+        return np.random.randint(0, 1, size=size, dtype=dtype)
+    return None
+
+
+class TestOperator:
+    def test_numpy_equivalency(self, size=100, verbose=pytest.verbose):
+        # ignore numpy warnings like divide by 0
+        np.seterr(all="ignore")
+        global pdarrays
```
Ethan-DeBandi99, line 28: Maybe I am missing something, but I don't think we need the global definitions here. Was there a reason we made these global definitions?

PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
+            for expression, err in results["neither_implement"]:
+                print(expression)
+        print(f'# ops implemented by numpy but not arkouda: {len(results["numpy_minus_arkouda"])}')
+        if verbose:
+            for expression, err, flag in results["numpy_minus_arkouda"]:
+                print(expression)
+        print(f'# ops implemented by arkouda but not numpy: {len(results["arkouda_minus_numpy"])}')
+        if verbose:
+            for expression, res, flag in results["arkouda_minus_numpy"]:
+                print(expression, " -> ", res)
```
Ethan-DeBandi99, line 149: These could all be wrapped in a single `if verbose` it looks like

PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
+        print(f"  Matching results:         {matches} / {nboth}")
+        print(f"  Arkouda execution errors: {len(execerrors)} / {nboth}")
+        if verbose:
+            print("\n".join(map(": ".join, execerrors)))
+        print(f"  Dtype mismatches:         {len(dtypeerrors)} / {nboth}")
+        if verbose:
+            print("\n".join(map(": ".join, dtypeerrors)))
+        print(f"  Value mismatches:         {len(valueerrors)} / {nboth}")
+        if verbose:
+            print("\n".join(map(": ".join, valueerrors)))
```
Ethan-DeBandi99, line 173: Again, we could use a single `if verbose` block here.

PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
+        assert ak_invert.to_list() == np_invert.tolist()
+
+    def test_uint_bool_binops(self):
+        # Test fix for issue #1932
+        # Adding support to binopvv to correctly handle uint and bool types
+        ak_uint = ak.arange(10, dtype=ak.uint64)
+        ak_bool = ak_uint % 2 == 0
+        assert (ak_uint + ak_bool).to_list() == (ak.arange(10) + ak_bool).to_list()
+
+    def test_float_uint_binops(self):
```
Ethan-DeBandi99, line 235: I wish we could parameterize the operator here, but I am not sure we can without a lot of headache.

PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
+        )
+        bi_arr = ak.array(
+            [0, 1, 2, 3, 4, 2**64 - 5, 2**64 - 4, 2**64 - 3, 2**64 - 2, 2**64 - 1],
+            dtype=ak.bigint,
+            max_bits=64,
+        )
+        np_arr = np.array(
+            [0, 1, 2, 3, 4, 2**64 - 5, 2**64 - 4, 2**64 - 3, 2**64 - 2, 2**64 - 1],
+            dtype=np.uint,
+        )
```
Ethan-DeBandi99, line 454: Would it make sense to add a helper function?

PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
```
Ethan-DeBandi99, line 498: Again, maybe a helper function just to reduce the number of places required for potential updates in the future?

Ethan-DeBandi99 (commented):
Few things to look at, but overall looks good.

PROTO_tests/tests/operator_test.py
pierce314159: to be honest, this was just leftover from how it used to be in `run_test` where that lived outside the class. I think you're right that it doesn't need to be that way anymore. I'll update this

PROTO_tests/tests/operator_test.py
pierce314159: I believe this the same deal, we only wanna print the mismatches with the appropriate header

PROTO_tests/tests/operator_test.py
```diff
@@ -0,0 +1,774 @@
...
+NUMERIC_TYPES = ["int64", "float64", "bool", "uint64"]
+
+
+def make_np_arrays(size, dtype):
+    if dtype == "int64":
+        return np.random.randint(-(2**32), 2**32, size=size, dtype=dtype)
+    elif dtype == "uint64":
+        return np.random.randint(-(2**32), 2**32, size=size).astype(dtype)
+    elif dtype == "float64":
+        return np.random.uniform(-(2**32), 2**32, size=size)
```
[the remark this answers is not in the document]
pierce314159: yah I think we should add an issue for this. Right now we only do float edge case testing on `//` and `%`

PROTO_tests/tests/operator_test.py
pierce314159: One idea I had was something like the below, but i feel like that ends up being less readable and harder to tell if we're testing the right thing

```python
for lfn, rfn in zip([`__add__`, `__sub__`, `__mul__`, ...], [`__radd__`, `__rsub__`, `__rmul__`, ...]):
    ak_ans = getattr(ak.pdarrayclass, fn)(ak_uint, akf)
    np_ans = getattr(np.whereever, fn)(np_uint, npf)
    assert np.allclose(ak_ans.to_ndarray(), np_ans, equal_nan=True)
```

PROTO_tests/tests/operator_test.py
pierce314159: sure! I can make that change

PROTO_tests/tests/operator_test.py
Ethan-DeBandi99: There are 3 separate loops, all inside a separate `if verbose`. Why can we not have a single `if` and then the loops under?

PROTO_tests/tests/operator_test.py
Ethan-DeBandi99: Same response as above.

PROTO_tests/tests/operator_test.py
pierce314159: Upon trying to delete the globals, I found that `do_op` requires these to be globals in order for the `eval` to work properly
```python
def do_op(lt, rt, ls, rs, isarkouda, oper):
    evalstr = ""
    if ls:
        evalstr += f'scalars["{lt}"]'
    else:
        evalstr += f'{("ndarrays", "pdarrays")[isarkouda]}["{lt}"]'
    evalstr += f" {oper} "
    if rs:
        evalstr += f'scalars["{rt}"]'
    else:
        evalstr += f'{("ndarrays", "pdarrays")[isarkouda]}["{rt}"]'
    res = eval(evalstr)
    return res
```

Ethan-DeBandi99: Ah ok. Sounds good. As long as there is a reason for them.

Ethan-DeBandi99 (approved):

PR #2707, opened 2023-08-22
Closes #2702: Capture only user-specified arkouda commands on client

### LINKED ISSUE #2702: capture only user-specified arkouda commands on client
**Feature Description**
The recently added [command logging](https://github.com/Bears-R-Us/arkouda#arkouda-command-logging) logs incoming commands along with any corresponds command arguments and requesting user. Client-side debug logging has demonstrated there are some implementation details that negatively impact the efficacy for command logging.

One example is the client invocation of ak.ones() results in three separate client server calls, none of which are called ones:

image

Looking at the ak.ones code, the invocations are as follows:

1. create the pdarray via 
2. fill with ones via pdarray.fill(1)
3. pdarray.\_\_repr\_\_ to display back to the client

In another example, ak.unique starts with the unique command which is followed by multiple delete commands deleting intermediate arrays along with a call to \_\_repr\_\_:

image

A possible enhancement would be to log all commands as submitted by the user via the client. One approach satisfies one of the command logging use cases--namely, to capture all commands within a workflow for later use in a non-interactive setting such as as cron job--is to generate and capture the command history within the Python client. Longer-term, being able to precisely capture solely user-specified commands server-side will require additional message metadata and processing logic. However, this issue is designed to satisfy the above-mentioned use case.

hokiegeek2:
Closes #2702 

This PR adds the following:

1. history module containing a base abstract HistoryRetriever and two derived classes that retrieve Python command history
2. ShellHistoryRetriever that retrieves Python command history from Python REPL shell
3. NotebookHistoryRetriever that retrieves Python command history from Jupyter notebook and IPython shells
4. Integration of HistoryRetriever classes into client module via generate_history method


arkouda/history.py
```diff
@@ -0,0 +1,113 @@
...
+        List[str]
+            A list of commands from the Python shell, Jupyter notebook, or IPython notebook
+        """
+        length_of_history = readline.get_current_history_length()
+        num_to_return = num_commands if num_commands else length_of_history
+
+        if command_filter:
+            return [readline.get_history_item(i + 1) for i in range(length_of_history)
+                    if self._filter_arkouda_command(readline.get_history_item(i + 1),
+                                                    command_filter)][-num_to_return:]
```
pierce314159, line 68: This is likely not super important at all since I imagine the history is relatively small. 

It looks like we are running the filter on every element in the history before whittling it down to `num_to_return`. I'm wondering if we'll ever hit a situation where the history is quite long and the user wants only a few things if this could take longer than necessary. 

Using a while loop instead would prevent this, but it probably doesn't make a big difference in actual performance

pierce314159 (commented):
Looks good other than one probably unintentional edit to the `get_mem_avail` docstring

arkouda/client.py
```diff
@@ -959,11 +1003,7 @@ def get_mem_avail(unit: str = "b", as_percent: bool = False) -> int:
     """
     Compute the amount of memory available to be used.
 
-    Parameters
-     ----------
-     unit : str {'b', 'kb', 'mb', 'gb', 'tb', 'pb'}
-         unit of return ('b' by default)
-     as_percent : bool
+c     as_percent : bool
```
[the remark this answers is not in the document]
hokiegeek2: @pierce314159 nope, fixed|

arkouda/client.py
```diff
@@ -959,10 +1003,6 @@ def get_mem_avail(unit: str = "b", as_percent: bool = False) -> int:
     """
     Compute the amount of memory available to be used.
 
-    Parameters
-     ----------
-     unit : str {'b', 'kb', 'mb', 'gb', 'tb', 'pb'}
-         unit of return ('b' by default)
```
pierce314159, line 965: we prob wanna add this part back too

arkouda/history.py
hokiegeek2: @pierce314159 yeah, good point, but I figured the length of history is gonna be small enough that this will not be noticeable performance hit. Also, can't use while loop in list comprehension.

arkouda/client.py
hokiegeek2: @pierce314159 yeesh, that's no bueno, fixed!

pierce314159 (approved):

jaketrookman (approved):
Looks good

PR #2709, opened 2023-08-23
Closes #2708 - `special_objType` support for HDF5

### LINKED ISSUE #2708: Simple types inheriting pdarrays (DateTime, TimeDelta, IPv4)

Ethan-DeBandi99:
Closes #2708

This PR adds support for persisting special object types that inherit `pdarray`. These include `IPv4`, `Datetime`, and `Timedelta`. The workflow still uses the `pdarray` write logic on the server, but will assign the special object type on load. This prevents the users from having to identify and recreate these objects from the `pdarray` returned. Instead, they are returned as the object type they were when saved.

Testing has been added to ensure that the object type is preserved. This is using the same logic on the server as pdarray, so we do not need to retest in depth.

### DIFF
--- a/PROTO_tests/tests/io_test.py
+++ b/PROTO_tests/tests/io_test.py
@@ -1247,6 +1247,47 @@ class TestHDF5:
             ret_idx = ak.read_hdf(f"{file_name}*")
             assert (idx == ret_idx).all()
 
+    def test_special_objtype(self):
+        """
+                This test is simply to ensure that the dtype is persisted through the io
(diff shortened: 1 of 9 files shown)


TheOddCzar789 (approved):

pierce314159 (commented):
Everything that's here looks good to me, do we want to add `BitVector`? If we don't or we wanna hold off for now, then I'll re-review to approve

Ethan-DeBandi99:
@pierce314159 - BitVector requires some additional information be stored, so I was planning to handle that as a separate issue.

PR #2710, opened 2023-08-24
Closes #2697: `datetime_test.py` conversion for new test framework

### LINKED ISSUE #2697: `datetime_test.py` conversion to new framework
Convert `datetime_test.py` to use new test framework

pierce314159:
This PR (closes #2697) converts `datetime_test.py` to new test framework

### DIFF
--- a/PROTO_tests/tests/datetime_test.py
+++ b/PROTO_tests/tests/datetime_test.py
@@ -184,7 +184,7 @@ class TestDatetime:
                 metrics["ak_supported"] += 1
                 compare_flag = True
                 try:
-                    pdret = eval("scsca {} pdfcvec".format(op))
+                    pdret = eval(f"scsca {op} pdfcvec")
                 except TypeError:
                     if verbose:
                         print(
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good

PROTO_tests/tests/datetime_test.py
```diff
@@ -0,0 +1,336 @@
...
+            else:
+                try:
+                    ret = eval(f"scsca {op} fcvec")
+                except Exception as e:
+                    raise TypeError(f"{secondclass} scalar {op} {firstclass}") from e
+                assert isinstance(ret, return_type)
+                metrics["ak_supported"] += 1
+                compare_flag = True
+                try:
+                    pdret = eval("scsca {} pdfcvec".format(op))
```
pierce314159, line 187:
```suggestion
                    pdret = eval(f"scsca {op} pdfcvec")
```

Ethan-DeBandi99 (approved):
I am just spitballing here, but... Should `datetime64` and `timedelta` not be `dtypes`? It seems like based on NumPy's implementation that maybe they should be dtypes housed by a pdarray, but that will complicate some things. I am just thinking that would be more consistent with NumPy. This is totally unrelated to the PR, just something I thought of while reviewing. I think this should be fine to merge.

PR #2722, opened 2023-08-28
Closes #2723: Fix bug with naming after new registry refactor

### LINKED ISSUE #2723: Attempting to unregister an array causes issues since `name` field isn't set
https://github.com/Bears-R-Us/arkouda/pull/2638 introduced a bug when attempting to unregister an array. When creating a registry entry, the name field isn't set, but is then attempted to be used to unregister an array. To get around that, in the initializers, the name field needs to be set. This PR handles that for the ArrayRegEntry, which was causing testing failures, but there are other types that have multiple names where it isn't as clear what "name" should be. I suspect those cases need to be fixed as well.

bmcdonald3:
#2638 introduced a bug when attempting to unregister an array. When creating a registry entry, the `name` field isn't set, but is then attempted to be used to unregister an array. To get around that, in the initializers, the name field needs to be set. This PR handles that for the `ArrayRegEntry`, which was causing testing failures, but there are other types that have multiple names where it isn't as clear what "name" should be. I suspect those cases need to be fixed as well.

Closes #2723

### DIFF
--- a/src/MultiTypeRegEntry.chpl
+++ b/src/MultiTypeRegEntry.chpl
@@ -70,6 +70,7 @@ module MultiTypeRegEntry {
         proc init(array_name: string, objType: ObjType) {
             super.init(objType);
             this.array = array_name;
+            this.name = array_name;
         }
 
         proc asMap(st: borrowed SymTab): map(string, string) throws {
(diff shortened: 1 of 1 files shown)


Ethan-DeBandi99 (approved):
Thanks for catching this and knocking it out!

pierce314159 (approved):

bradcray:
Thanks for the merge @pierce314159!  And thanks for identifying and fixing this, @bmcdonald3!

PR #2733, opened 2023-08-29
Closes #2732: Fix missed symbol table entry creation for distributed builds

### LINKED ISSUE #2732: Fix missed symbol table entry creation for distributed builds
In #2653, the compatability shim was added to support the new tryCreateArray throwing interface and that interface removed initializers for block-distributed symbol table entries, replacing with the new createSymEntry helper function. There were a few instances of this missed in the KExtreme code, which was causing distributed builds to fail.

bmcdonald3:
In #2653, the compatability shim was added to support the new tryCreateArray throwing interface and that interface removed initializers for block-distributed symbol table entries, replacing with the new createSymEntry helper function. There were a few instances of this missed in the KExtreme code, which was causing distributed builds to fail.

Closes #2732

### DIFF
--- a/src/KExtremeMsg.chpl
+++ b/src/KExtremeMsg.chpl
@@ -49,7 +49,7 @@ module KExtremeMsg
             when (DType.Int64) {
                 var e = toSymEntry(gEnt,int);
                 var aV = if !returnIndices then computeExtremaValues(e.a, k) else computeExtremaIndices(e.a, k);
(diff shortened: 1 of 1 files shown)


hokiegeek2 (approved):
looks good to me

pierce314159 (approved):
Thanks for the quick fix @bmcdonald3!

PR #2735, opened 2023-08-29
Closes #2734: Fix remaining missed symbol table entry creation for distributed builds

### LINKED ISSUE #2734: Fix remaining missed symbol table entry creation for distributed builds
In https://github.com/Bears-R-Us/arkouda/pull/2653, the compatibility shim was added to support the new `tryCreateArray` throwing interface and that interface removed initializers for block-distributed symbol table entries, replacing with the new createSymEntry helper function. There were a few instances of this missed in the KExtreme code, which was causing distributed builds to fail.

This is related to #2732, which only covered occurrences that were in `KExtremeMsg`

pierce314159:
This PR (closes #2734) fixes the remaining missed symbol table entry creation for distributed builds

This is related to #2732, which only covered occurrences that were in `KExtremeMsg`. I'm not sure if all these spots need updating but I was still seeing similar failures when making with gasnet enabled, but in other files. I would especially like @bmcdonald3 to review since this is based on his PR

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -257,7 +257,7 @@ module ArgSortMsg
           var merged = mergeNumericArrays(numDigits, arrSize, totalDigits, bitWidths, negs, names, st);
 
           var iv = argsortDefault(merged, algorithm=algorithm);
-          st.addEntry(ivname, new shared SymEntry(iv));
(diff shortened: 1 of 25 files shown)


bmcdonald3 (approved):
Thanks Pierce! Sorry for the mistakes 😨

hokiegeek2 (approved):
looks good

PR #2741, opened 2023-08-30
Fixes #2737: Address sanitization errors on `unregister`

### LINKED ISSUE #2737: Address sanitizer errors on unregister
When building chapel with the address sanitizer (instructions [here](https://chapel-lang.org/docs/developer/bestPractices/Sanitizers.html)). We see an access after free error when something gets unregistered

```python
>>> a = ak.arange(10)
>>> a.register('pda')
array([0 1 2 3 4 5 6 7 8 9])
>>> a.unregister()
```

This results in a `AddressSanitizer: heap-use-after-free` error. This seems to be the result of removing the name from the registry table
```chapel
tab.remove(are.name);
```

This is the error that led to the nightly failures that caused us to skip `testNA` in https://github.com/Bears-R-Us/arkouda/issues/2726. But this doesn't just apply to categorical, it happens when unregistering any object

pierce314159:
This PR (fixes #2737) changes `registry.lookup` to return a `shared` object instead of `borrowed`. This makes it so later references to `gre` in `unregisterMsg` don't cause us to access freed memory

### DIFF
--- a/src/Registry.chpl
+++ b/src/Registry.chpl
@@ -160,7 +160,7 @@ module Registry {
             bre.setName(name);
         }
 
-        proc unregister_array(are: borrowed ArrayRegEntry) throws {
+        proc unregister_array(are: shared ArrayRegEntry) throws {
             registered_entries.remove(are.array);
             tab.remove(are.name);
         }
@@ -173,12 +173,12 @@ module Registry {
             }
         }
 
-        proc unregister_segarray(sre: borrowed SegArrayRegEntry) throws {
+        proc unregister_segarray(sre: shared SegArrayRegEntry) throws {
             unregister_segarray_components(sre);
             tab.remove(sre.name);
         }
 
-        proc unregister_dataframe(dfre: borrowed DataFrameRegEntry) throws {
+        proc unregister_dataframe(dfre: shared DataFrameRegEntry) throws {
             registered_entries.remove(dfre.idx);
             for c in dfre.columns {
                 var gre = c: borrowed GenRegEntry;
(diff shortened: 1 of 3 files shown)


hokiegeek2 (approved):
looks good to me

bmcdonald3 (approved):

jaketrookman (approved):
Looks good. After discussing with @pierce314159, either he or @bmcdonald3 will shortly be putting up a compat module to fix the CI test issue

TheOddCzar789 (approved):

src/Registry.chpl
```diff
@@ -279,12 +279,12 @@ module Registry {
...
-        proc unregister_bitvector(bre: borrowed BitVectorRegEntry) throws {
+        proc unregister_bitvector(bre: shared BitVectorRegEntry) throws {
             registered_entries.remove(bre.array);
             tab.remove(bre.name);
         }
 
-        proc lookup(name: string): borrowed AbstractRegEntry throws {
+        proc lookup(name: string): shared AbstractRegEntry throws {
             checkTable(name, "lookup");
             return tab[name];
```
bmcdonald3, line 289: To resolve the CI build issues, I would replace the `return tab[name]` with `table.table[slot].val`. This touches some internals that aren't really user-facing, but, in my opinion, is the most simple workaround here.

bmcdonald3: For context, we have to do this because there was a bug in Chapel 1.30 with the Map module (that I introduced, it was all my fault 😅)

PR #2745, opened 2023-09-01
Fixes #2644, #2645: Fix uint cast to str/float and add str cast to str

### LINKED ISSUE #2644: Bug casting from uint pdarray to strings/float
We currently don't have the ability to cast from a uint pdarray to a strings. I found this while reviewing the [pdarray_creation_tests PR](https://github.com/Bears-R-Us/arkouda/pull/2584)

```python
# int pdarray converts just fine
>>> ak.cast(ak.arange(10), str)
array(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'])

# uint pdarray causes errors
>>> ak.cast(ak.arange(10, dtype=ak.uint64), str)
Cell In[12], line 1
----> 1 ak.cast(ak.arange(10, dtype=ak.uint64), str)

RuntimeError: Error: castMsg: UInt64 : str not implemented
```

The issue seems to be this block. For all other types we do `"str"` not `"string"`
https://github.com/Bears-R-Us/arkouda/blob/5eb7b5f5d328cb02bbc60ce6f2d7a4daf940c0a3/src/CastMsg.chpl#L85-L87

Similarly we can't cast from `uint` to `float` because we write `"float"` instead `"float64"`

### LINKED ISSUE #2645: Add casting from Strings to Strings
There's no reason we shouldn't be able to handle this. Most pdarray types are able to cast to same type

pierce314159:
This PR (fixes #2644 and fixes #2645) fixes uint casting to str and float and adds casting from str to str

### DIFF
--- a/PROTO_tests/tests/numeric_test.py
+++ b/PROTO_tests/tests/numeric_test.py
@@ -88,13 +88,16 @@ class TestNumeric:
     def test_cast(self, prob_size, cast_to):
         arrays = {
             ak.int64: ak.randint(-(2**48), 2**48, prob_size),
+            ak.uint64: ak.randint(0, 2**48, prob_size, dtype=ak.uint64),
             ak.float64: ak.randint(0, 1, prob_size, dtype=ak.float64),
             ak.bool: ak.randint(0, 2, prob_size, dtype=ak.bool),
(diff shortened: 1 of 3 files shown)


TheOddCzar789 (approved):
Looks good

jaketrookman (approved):
Looks good

Issue #2749, opened 2023-09-04
Extend dataframe groubpy aggregations to accept lists of column names


pierce314159:
Extend dataframe groubpy aggregations to accept lists of column names. Leaving this empty will apply the aggregation to all columns
```python
>>> username = ["Alice", "Bob", "Alice", "Carol", "Bob", "Alice"]
>>> userid = [111, 222, 111, 333, 222, 111]
>>> item = [0, 0, 1, 1, 2, 0]
>>> day = [5, 5, 6, 5, 6, 6]
>>> amount = [0.5, 0.6, 1.1, 1.2, 4.3, 0.6]
>>> p = pd.DataFrame({"userName": username, "userID": userid, "item": item, "day": day, "amount": amount})
>>> p.groupby('userID').first()
       userName  item  day  amount
userID
111       Alice     0    5     0.5
222         Bob     0    5     0.6
333       Carol     1    5     1.2
```

Issue #2521, opened 2023-06-22
Add `uint` support to histogram


pierce314159:
We want to be able to do `ak.histogram` on `uint`, like:
```python
>>> pda = ak.randint(10, 30, 40, dtype=ak.uint64)
>>> bins, result = ak.histogram(pda, bins=20)
```

PR #2750, opened 2023-09-04
If keys are not provided, use all columns

Jemeljanov:
Closes issue #2749

Not parsing value to groupby aggregates all columns.

Kindly review the changes.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -1307,7 +1307,7 @@ class DataFrame(UserDict):
             return self
         return self[array(random.sample(range(self._size), n))]
 
-    def GroupBy(self, keys, use_series=False):
+    def GroupBy(self, keys=None, use_series=False):
         """
         Group the dataframe by a column or a list of columns.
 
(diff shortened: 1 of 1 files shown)


pierce314159:
hi @Jemeljanov, thanks for the PR! We appreciate the contribution!

This is slightly different than what was requested in the issue (I'm sure I could've made the issue more clear, but I was mostly writing it with me as the intended audience). This PR makes it so the grouping keys are all column names if none are specified. The issue was meant to have the aggregation be performed on all column names if none are specified. It's subtle but I think the below explains the difference a bit:
```python
# This PR is acting on grouping_keys, for the issue we want to act on aggregate_on
df.grouping(grouping_keys).agg(aggregate_on)
```

I put up a PR for this issue as well (see #2751) which adds testing and resolves another issue, so I'm going to close this in favor of that one. But don't let that discourage you from trying your hand at future arkouda issues! In the future, it'd prob be a good idea to drop a comment in any issue you'd like to work to avoid duplicating effort. That would also make it easier for the dev team to answer any question you may have while working an issue.

If you are looking for another issue to work, I'd recommend https://github.com/Bears-R-Us/arkouda/issues/2521 as a nice starting point!

pierce314159:
It's also worth noting that if there's a different reason that we'd like to have this behavior, feel free to add an issue explaining why and open up a new PR against it!

PR #2755, opened 2023-09-06
Fixes #2703: Sort bug with `nan`s

### LINKED ISSUE #2703: Bug in `sort` and `extrema` functions' handling of float edge cases
```python
>>> np_arr = np.array([-3.14, np.inf, np.nan, -np.inf, 3.14, 0.0, 3.14, -8])
>>> np_sorted = np.sort(np_arr)
>>> np_sorted
array([ -inf, -8.  , -3.14,  0.  ,  3.14,  3.14,   inf,   nan])

ak.sort(ak.array(np_arr))
array([0 -inf -8 -3.1400000000000001 3.1400000000000001 3.1400000000000001 inf nan])
```
After playing around with this a bit more, it seems to only mis-sort `0`. and this only seems to happen if a `nan` is present

Once this is done we should add float edge case testing for `extrema_test` and `sort_test`

pierce314159:
This PR (fixes #2703)

When a `nan` is present in `a` the value of `min reduce a` will equal `nan`. So `signbit(min reduce a)` will be false even if there are negatives present. This was causing the sort to mishandle `0.0` (it left them at the front instead of moving them to the correct spot)

I updated the code to do the same thing it used to if `min reduce a` is not a `nan`, and when it is to find the signbits of all values see if any are true (i.e. `| reduce signbit(a)`). So in the common case, it should be equivalent and in the more rare `nan`s present case, it does `signbits` on every value and 2 reductions.

It would be simpler to always do `| reduce signbit(a)`, and I feel like calling `signbit` on every value of `a` then reducing shouldn't be too much more expensive than reducing first and doing only one `signbit` call. But I know the sort code is super optimized, so I tried to keep it as minimal impact on the common case as i could.

If @ronawho doesn't mind looking this over and making sure I'm not doing something dumb that will kill the performance / there's not a better way, i'd appreciated it!

```python
>>> ak.sort(ak.array([np.inf,-9, -3.0, 0.0, 3.0, np.nan]))
# before this PR 
array([0 -9 -3 3 inf nan])
# after this PR
array([-9 -3 0 3 inf nan])
```

### DIFF
--- a/src/AryUtil.chpl
+++ b/src/AryUtil.chpl
@@ -248,7 +248,9 @@ module AryUtil
 
     inline proc getBitWidth(a: [?aD] real): (int, bool) {
       const bitWidth = numBits(real);
-      const negs = signbit(min reduce a);
+      const minVal = min reduce a;
+      // nan is returned for a `min reduce` even if negatives are present in that case do an `| reduce` over all signbits
+      const negs = if isnan(minVal) then | reduce signbit(a) else signbit(minVal);
       return (bitWidth, negs);
     }
 
--- a/src/AryUtil.chpl
+++ b/src/AryUtil.chpl
@@ -248,9 +248,7 @@ module AryUtil
 
     inline proc getBitWidth(a: [?aD] real): (int, bool) {
       const bitWidth = numBits(real);
-      const minVal = min reduce a;
-      // nan is returned for a `min reduce` even if negatives are present in that case do an `| reduce` over all signbits
-      const negs = if isnan(minVal) then | reduce signbit(a) else signbit(minVal);
+      const negs = | reduce signbit(a);
       return (bitWidth, negs);
     }
 
--- a/tests/sort_test.py
+++ b/tests/sort_test.py
@@ -77,7 +77,9 @@ class SortTest(ArkoudaTest):
 
     def test_nan_sort(self):
         # Reproducer from #2703
-        np_arr = np.array([-3.14, np.inf, np.nan, -np.inf, 3.14, 0.0, 3.14, -8])
-        self.assertTrue(
-            np.allclose(np.sort(np_arr), ak.sort(ak.array(np_arr)).to_ndarray(), equal_nan=True)
-        )
+        neg_arr = np.array([-3.14, np.inf, np.nan, -np.inf, 3.14, 0.0, 3.14, -8])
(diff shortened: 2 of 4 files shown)


src/AryUtil.chpl
```diff
@@ -248,7 +248,9 @@ module AryUtil
 
     inline proc getBitWidth(a: [?aD] real): (int, bool) {
       const bitWidth = numBits(real);
-      const negs = signbit(min reduce a);
+      const minVal = min reduce a;
+      // nan is returned for a `min reduce` even if negatives are present in that case do an `| reduce` over all signbits
+      const negs = if isnan(minVal) then | reduce signbit(a) else signbit(minVal);
       return (bitWidth, negs);
```
ronawho, line 254: This doesn't use the minVal except to check the sign bit, so probably don't need to bother computing it and can instead just do the `const negs = | reduce signbit(a);`

Having the extra reduce today shouldn't have any noticeable performance impact. It's an extra read of the array, but that's really fast compared to the sorting itself.

tests/sort_test.py
```diff
@@ -75,3 +74,10 @@ def testErrorHandling(self):
             # Test attempt to sort Strings object, which is unsupported
             with self.assertRaises(TypeError):
                 ak.sort(ak.array(["String {}".format(i) for i in range(0, 10)]), algo)
+
+    def test_nan_sort(self):
+        # Reproducer from #2703
+        np_arr = np.array([-3.14, np.inf, np.nan, -np.inf, 3.14, 0.0, 3.14, -8])
```
ronawho, line 80: I think it'd be worth having a positive version with nan/inf as well.

src/AryUtil.chpl
pierce314159: Awesome! I was thinking the same thing but wanted to make sure!

ronawho: And just seeing you mentioned that in the OP, but clearly I didn't read anything before my name :)

`signbit` is  trivially fast to call (shouldn't be any more expensive than the comparison used to check for `min`) so I'd just use that instead of trying to limit when you call signbit.

tests/sort_test.py
pierce314159: a version with only positive values? That's a good idea, just to make sure it still handles things correctly when `negs` is False

ronawho (approved):

jaketrookman (approved):
Looks good to me

TheOddCzar789 (approved):

PR #2756, opened 2023-09-07
Closes #2744 - Implement `skew` and `hist_all`

### LINKED ISSUE #2744: Implement `skew` and `hist_all`
Add `skew` to arkouda similar to what is done in scipy https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.skew.html

egelberg:
### `skew`
Computes the sample skewness of an array (added to `pdarrayclass.py`). Based on [scipy.stats.skew](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.skew.html)
  - Skewness > 0 means there's greater weight in the right tail of the distribution.
  - Skewness < 0 means there's greater weight in the left tail of the distribution.
  - Skewness == 0 means the data is normally distributed.

### `hist_all`
Creates a grid plot histogramming all numeric columns in ak dataframe.
- Binning is done automatically utilizing [Doane's formula](https://en.wikipedia.org/wiki/Histogram#Doane's_formula)
- Example:
```
import arkouda as ak
from arkouda.plotting import hist_all 
ak_df = ak.DataFrame({"a": ak.array(np.random.randn(100)),
                      "b": ak.array(np.random.randn(100)),
                      "c": ak.array(np.random.randn(100)),
                      "d": ak.array(np.random.randn(100))
                     })
hist_all(ak_df)
```
PNG image


arkouda/pdarrayclass.py
```diff
@@ -3073,6 +3073,45 @@ def sqrt(pda: pdarray, where: Union[bool, pdarray] = True) -> pdarray:
...
+    cubed_deviations = deviations ** 3
+
+    std_dev = pda.std()
+
+    if std_dev != 0:
+        skewness = cubed_deviations.mean() / (std_dev ** 3)
+    else:
+        skewness = 0
+
+    return skewness
```
pierce314159, line 3112: completely unnecessary stylistic suggestion to cut down on number of lines, feel free to ignore :)
```python
cubed_deviations = deviations ** 3
std_dev = pda.std()
return 0 if std_dev == 0 else cubed_deviations.mean() / (std_dev ** 3)
```

arkouda/pdarrayclass.py
```diff
@@ -3073,6 +3073,45 @@ def sqrt(pda: pdarray, where: Union[bool, pdarray] = True) -> pdarray:
     return power(pda, 0.5, where)
 
 
+@typechecked
+def skew(pda: pdarray) -> np.float64:
+
+    """
+    Computes the sample skewness of an array.
+    Skewness > 0 means there's greater weight in the right tail of the distribution.
```
pierce314159, line 3081: The scipy version takes a `bias` argument, is that something we should have in the arkouda version?

> bias bool, optional
If False, then the calculations are corrected for statistical bias.

arkouda/plotting.py
```diff
@@ -50,3 +57,99 @@ def plot_dist(b, h, log=True, xlabel=None, newfig=True):
...
+        elif isinstance(x, Timedelta):
+            bins = (
+                timedelta_range(x.min(), x.max(), periods=num_bins)
+                .to_ndarray()
+                .astype("int")
+            )
+        else:
+            bins = np.linspace(x.min(), x.max(), num_bins + 1)[:-1]
+
+        print(col, num_bins)
```
pierce314159, line 150: Just checking that this is intended to be printing and not leftover from debugging

arkouda/plotting.py
```diff
@@ -50,3 +57,99 @@ def plot_dist(b, h, log=True, xlabel=None, newfig=True):
...
+    else:
+        axes = [axes]
+
+    for col in cols:
+        try:
+            ax = axes[cols.index(col)]
+            x = ak_df[col]
+
+            if x.dtype == "float64":
+                x = x[isnan(x)]
```
pierce314159, line 110: maybe im not understanding but shouldn't this be `~isnan`? because otherwise `x` is only the `nan` values
```suggestion
            if x.dtype == "float64":
                x = x[~isnan(x)]
```

arkouda/plotting.py
```diff
@@ -50,3 +57,99 @@ def plot_dist(b, h, log=True, xlabel=None, newfig=True):
...
+            g1 = skew(x)
+
+        except ValueError:
+            GB_df = GroupBy(ak_df[col])
+            new_labels = arange(GB_df.unique_keys.size)
+            newcol = GB_df.broadcast(new_labels)
+            x = newcol[0 : ak_df.size]
+
+            if x.dtype == "float64":
+                x = x[isnan(x)]
```
pierce314159, line 122: same question about adding `~`

arkouda/plotting.py
```diff
@@ -50,3 +57,99 @@ def plot_dist(b, h, log=True, xlabel=None, newfig=True):
...
+
+            n = len(x)
+            g1 = skew(x)
+
+        except ValueError:
+            GB_df = GroupBy(ak_df[col])
+            new_labels = arange(GB_df.unique_keys.size)
+            newcol = GB_df.broadcast(new_labels)
+            x = newcol[0 : ak_df.size]
```
pierce314159, line 119: stylistic, the `0` is implied for slices

arkouda/plotting.py
```diff
@@ -50,3 +57,99 @@ def plot_dist(b, h, log=True, xlabel=None, newfig=True):
...
+                timedelta_range(x.min(), x.max(), periods=num_bins)
+                .to_ndarray()
+                .astype("int")
+            )
+        else:
+            bins = np.linspace(x.min(), x.max(), num_bins + 1)[:-1]
+
+        print(col, num_bins)
+
+        ax.bar(bins, h[1].to_ndarray(), width=bins[1] - bins[0])
```
pierce314159, line 152: when attempting run this locally, im hitting an error on this line
```python
>>> ak_df = ak.DataFrame({"a": ak.randint(0,100,10),"b": ak.randint(0,100,10)})

>>> ak.hist_all(ak_df)
a 5
Cell In[5], line 1
----> 1 ak.hist_all(ak_df)

File ~/proj/arkouda/arkouda/plotting.py:152, in hist_all(ak_df, cols, log)
    148     bins = np.linspace(x.min(), x.max(), num_bins + 1)[:-1]
    150 print(col, num_bins)
--> 152 ax.bar(bins, h[1].to_ndarray(), width=bins[1] - bins[0])
    153 ax.set_title(col, size=8)
    154 if x.max() > 100 * x.min():

AttributeError: 'numpy.ndarray' object has no attribute 'bar'
```

pierce314159 (changes requested):
Few comments, mostly style stuff but I'm seeing one error in `hist_all`

jaketrookman (approved):
Besides the aforementioned changes, looks good

arkouda/plotting.py
egelberg: Doh.

arkouda/plotting.py
egelberg: Definitely supposed to be that, I think the black formatting may have screwed it up. Changing now.

arkouda/plotting.py
egelberg: Hmm... I was able to recreate this but only when the df had 2 or 3 columns (1 and 4+ worked fine 🤔 ).
I'll have to dig in

egelberg: Figured it out
`int(len(cols) ** 0.5)` when `len(cols) is 2 or 3` = 1
This portion of the code is used to create an automatic grid for the histogram plots, so it's a pretty easy fix using `math.ceil`

arkouda/pdarrayclass.py
egelberg: @pierce314159 Good catch, the answer is yes. I just took care of this in the latest commit to try and get it to match [what scipy is doing](https://github.com/scipy/scipy/blob/4ec80491eee7bbb85bd8a248c9d974b1792790fd/scipy/stats/_mstats_basic.py#L2732-L2741). Tested it and it looks good!

arkouda/pdarrayclass.py
```diff
@@ -3074,7 +3074,7 @@ def sqrt(pda: pdarray, where: Union[bool, pdarray] = True) -> pdarray:
 
 
 @typechecked
-def skew(pda: pdarray) -> np.float64:
+def skew(pda: pdarray, bias: bool=True) -> np.float64:
```
TheOddCzar789, line 3077: Looks like flake8 wants spaces in here:
`bool = True`

TheOddCzar789 (approved):

arkouda/plotting.py
```diff
@@ -50,3 +57,97 @@ def plot_dist(b, h, log=True, xlabel=None, newfig=True):
...
+                .to_ndarray()
+                .astype("int")
+            )
+        else:
+            bins = np.linspace(x.min(), x.max(), num_bins + 1)[:-1]
+
+        ax.bar(bins, h[1].to_ndarray(), width=bins[1] - bins[0])
+        ax.set_title(col, size=8)
+        if x.max() > 100 * x.min():
+            ax.set_yscale("log")
```
pierce314159, line 153: I might be misreading but it seems like the `log` bool argument is never actually used. It looks the scale being set to log is trigged by this chunk of logic instead

pierce314159 (commented):
Just one question about the `log` argument

arkouda/plotting.py
egelberg: @pierce314159 good call, it's a holdover from testing. I'm removing it so the yscale is automatically determined for every column.
If a user wants the ability to modify the scale moving forward we can think about adding it back in.

jaketrookman (approved):

pierce314159 (approved):

PR #2775, opened 2023-09-14
Closes #2774: Resolve deprecation warnings in advance of 1.32 release

### LINKED ISSUE #2774: Resolve 1.32 Deprecation warnings
Deprecations addressed from Chapel 1.32:

- values modified in loop body must be specified as `ref` with intents
- `c_string` deprecated, not replaced, but can use `c_ptrConst(c_char))` now
- `domain.dist` deprecated in favor of `domain.distribution`
- `NAN`, `INFINITY` deprecated in favor of `nan` and `inf`
- casting to bytes/strings deprecated in favor of type method creation functions
- `chpl_task_yield` deprecated in favor of `currentTask.yieldExecution`
- file reader `kind` deprecated for use of serializers
- `writeThis` deprecated in favor of `serialize`
- generic class values must now have `(?)` to indicate they are generic
- `Block` deprecated in favor of `blockDist`
- `date.isoCalendar` deprecated in favor of `date.isoWeekDate`
- `JsonSerializer` changed name to `jsonSerializer`
- `dateTime.createFromTimestamp` deprecated in favor of `dateTime.createUtcFromTimestamp().getDate()`

bmcdonald3:
Deprecations addressed from Chapel 1.32:

- values modified in loop body must be specified as `ref` with intents
- `c_string` deprecated, not replaced, but can use `c_ptrConst(c_char))` now
- `domain.dist` deprecated in favor of `domain.distribution`
- `NAN`, `INFINITY` deprecated in favor of `nan` and `inf`
- casting to bytes/strings deprecated in favor of type method creation functions
- `chpl_task_yield` deprecated in favor of `currentTask.yieldExecution`
- file reader `kind` deprecated for use of serializers
- `writeThis` deprecated in favor of `serialize`
- generic class values must now have `(?)` to indicate they are generic
- `Block` deprecated in favor of `blockDist`
- `date.isoCalendar` deprecated in favor of `date.isoWeekDate`
- `JsonSerializer` changed name to `jsonSerializer`
- `dateTime.createFromTimestamp` deprecated in favor of `dateTime.createUtcFromTimestamp().getDate()`

Closes #2774

### DIFF
--- a/src/asd.chpl
+++ b/src/asd.chpl
@@ -1,0 +1,8 @@
+use AryUtil;
+
+var values: [0..10] int;
+var res: [0..10] int;
+
+coforall loc in Locales with (ref values, ref res) {
+  var slice = new lowLevelLocalizingSlice(values, start..#len);
+}
--- a/src/asd.chpl
+++ b/src/asd.chpl
@@ -1,8 +1,0 @@
-use AryUtil;
-
-var values: [0..10] int;
-var res: [0..10] int;
-
-coforall loc in Locales with (ref values, ref res) {
-  var slice = new lowLevelLocalizingSlice(values, start..#len);
-}
--- a/src/compat/e-130/ArkoudaSymEntryCompat.chpl
+++ b/src/compat/e-130/ArkoudaSymEntryCompat.chpl
@@ -1,0 +1,28 @@
+module ArkoudaSymEntryCompat {
+  use MultiTypeSymEntry;
+  use Map;
+
+  type SymEntryAny = SymEntry;
+  type mapAny = map;
+
+  override proc SymEntry.writeThis(f) throws {
+    use Reflection;
+    proc writeField(f, param i) throws {
+      if !isArray(getField(this, i)) {
+        f.write(getFieldName(this.type, i), " = ", getField(this, i):string);
+      } else {
+        f.write(getFieldName(this.type, i), " = ", formatAry(getField(this, i)));
+      }
+    }
+
+    super.writeThis(f);
+    f.write(" {");
+    param nFields = numFields(this.type);
+    for param i in 0..nFields-2 {
+      writeField(f, i);
+      f.write(", ");
+    }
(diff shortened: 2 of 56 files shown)


TheOddCzar789 (approved):

bmcdonald3:
Converting this PR to a draft as we are discussing on the Chapel team about making the `ref` changes unstable, rather than generating a deprecation warning (meaning that we wouldn't need to make the changes in Arkouda). If we end up making it unstable, I will undo all the `ref` changes, which will lessen the impact/size of this round of deprecations on Arkouda.

src/asd.chpl
pierce314159, line 1: is this leftover from testing?

src/compat/e-130/ArkoudaSymEntryCompat.chpl
pierce314159, line 1: is this compat module imported anywhere? I feel like it should be in `src/MultiTypeSymEntry.chpl`

pierce314159 (commented):
2 small things to look into

src/asd.chpl
pierce314159: no worries!!

pierce314159 (approved):
Looks good to me!

src/compat/e-130/ArkoudaSymEntryCompat.chpl
bmcdonald3: Ya, you are right about that... In 1.32, we deprecated `writeThis` in favor of serialize, so these needed to be pulled out into their own modules. With the missing imports I had, these would have just been calling the default printing methods for the symbol table entries. Seeing that this still passed testing and looking into it a bit, I think that we are only actually using the `__str__` method, so we might not need these `writeThis` and `serialize` methods anymore... I don't think we should wrap that into this PR, but something I'll be thinking about.

pierce314159: that makes sense!

PR #2781, opened 2023-09-22
Closes #2716: Add dataframe merge functionality

### LINKED ISSUE #2716: Add support for `merge` and `join` on `DataFrame`s
Add the ability to `merge` and `join` on arkouda dataframes this should mimic the pandas dataframe [`merge`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html) and [`join`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.join.html#pandas.DataFrame.join)

egelberg:
This PR closes #2716

The purpose of this branch is essentially to start to mimic pandas [merge](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html). It is not currently an exact copy of pandas.

An example of the functionality can be seen when running the following in a jupyter notebook attached to an arkouda server:
```
people_left = ak.array(['josh', 'brennan', 'zac', 'brennan', 'ross'])
people_right = ak.array(['josh', 'brennan', 'zac', 'sam', 'ross'])
desserts = ak.array(['cake', 'cookies', 'cookies', 'pie', 'ice cream'])
fruits = ak.array(['apple', 'pear', 'orange', 'apple', 'banana'])

ak_desserts = ak.DataFrame({'person': people_left, 'dessert': desserts})
ak_fruits= ak.DataFrame({'person': people_right, 'fruit': fruits})

merge(ak_desserts, ak_fruits, on='person', how='inner')

merge(ak_desserts, ak_fruits, on='person', how='left')

merge(ak_desserts, ak_fruits, on='person', how='right')
```

It has proven to be performative at scale
Test 1:
- left_df: 4B rows
- right_df: 4B rows
- inner join: 4B -> 2.78B rows in 23 seconds
- right join: 4B -> 4B rows in 67 seconds

Test 2:
- left_df: 4B rows
- right_df: 2B rows
- inner join: Reduced to 1.4B rows in 19 seconds
- right join: Reduced to 2B rows in 52 seconds


egelberg:
Seeing that the following import is causing some errors:
`from arkouda.dataframe import DataFrame`

@pierce314159 any thoughts? I'm only importing it to enforce a parameter type in the function inputs

pierce314159:
> Seeing that the following import is causing some errors: `from arkouda.dataframe import DataFrame`
> 
> @pierce314159 any thoughts? I'm only importing it to enforce a parameter type in the function inputs

I wonder if it would make sense for these functions to be in `dataframe.py` instead

egelberg:
Ya I think you're right, that appears to be how [pandas](https://github.com/pandas-dev/pandas/blob/v2.1.1/pandas/core/frame.py#L10471-L10504) does it

egelberg:
@pierce314159 I think this should be good now! I ended up solving the [integer cast issue](https://github.com/Bears-R-Us/arkouda/issues/2788) in the right_join_merge code itself. I think I can still take a crack at the `numeric.cast` method as well, but this at least sets this merge up for the next release!

arkouda/dataframe.py
```diff
@@ -2706,3 +2708,152 @@ def invert_permutation(perm):
...
+    left_cols.remove(on)
+    right_cols = right.columns.copy()
+    right_cols.remove(on)
+
+    new_dict = {on: left[on][ij[0]]}
+
+    for col in left_cols:
+        new_dict[col] = left[col][ij[0]]
+    for col in right_cols:
+        new_dict[col] = right[col][ij[1]]
```
pierce314159, line 2752: let's say we have `col_i` which is in both `left.columns` and `right.columns`, but `col_i` is not in `on`. What the expected behavior for `col_i` in the inner joined return df?

Right now we are setting it equal to whatever values the `right` df has when indexed/permuted to meet the inner join condition (since we set it equal to left first and the overwrite it with right). I just want to make sure that's what we are intending to happen

If we are down to just always take `col_i` to be the right df, then we could have some unnecessary work since we first set `new_dict['col_i']` equal to `left['col_i][ij[0]]` then overwrite it with `right['col_i'][ij[1]]`. This could be quite a lot of unused indexing especially if `(left.columns ∩ right.columns) - on` is quite large

We can avoid this by doing something like `left_cols = set(left_cols) - set(right_cols)`. Hopefully this makes sense, if not we can jump on call to talk through it

arkouda/dataframe.py
```diff
@@ -2706,3 +2708,152 @@ def invert_permutation(perm):
...
+        The name of the DataFrame column the join is being
+        performed on
+
+    Returns
+    -------
+    ij_ak_df : DataFrame
+        Inner-Joined Arkouda DataFrame
+    """
+
+    ij = inner_join(left[on], right[on])
```
pierce314159, line 2740: completely stylistic, feel free to ignore. but maybe we make this return be 
```python
left_inds, right_inds = inner_join(left[on], right[on])
```
i just feel like it's slightly more readable to have like `left[left_inds]` than `left[ij[0]]`

arkouda/dataframe.py
```diff
@@ -2706,3 +2708,152 @@ def invert_permutation(perm):
...
+    new_dict = {on: left[on][ij[0]]}
+
+    for col in left_cols:
+        new_dict[col] = left[col][ij[0]]
+    for col in right_cols:
+        new_dict[col] = right[col][ij[1]]
+
+    ij_ak_df = DataFrame(new_dict)
+
+    return ij_ak_df
```
pierce314159, line 2756: also completely stylistic but we never use this variable so we can just return right away
```suggestion
    return DataFrame(new_dict)
```

pierce314159 (changes requested):
I only really got a chance to look at `inner_join_merge` in depth. I'm about to run to a meeting, and I figured I'd go ahead and post what I have so you can start going through it. I'll look at the rest after

arkouda/dataframe.py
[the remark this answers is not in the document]
egelberg: I think yes to both. First one seems easy, do you have an idea of how to make #2 happen?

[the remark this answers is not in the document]
pierce314159: yah i good example is some of the functions in `pdarrayclass`

Adding to `__all__`:
https://github.com/Bears-R-Us/arkouda/blob/401154c104c2dfb928df3a9ebc8cf83b85e850fd/arkouda/pdarrayclass.py#L39

class level i.e. `pda.sum()`:
https://github.com/Bears-R-Us/arkouda/blob/401154c104c2dfb928df3a9ebc8cf83b85e850fd/arkouda/pdarrayclass.py#L850-L854

top level i.e `ak.sum(pda)`
https://github.com/Bears-R-Us/arkouda/blob/401154c104c2dfb928df3a9ebc8cf83b85e850fd/arkouda/pdarrayclass.py#L2139-L2161

arkouda/dataframe.py
```diff
@@ -2706,3 +2708,152 @@ def invert_permutation(perm):
...
+    right_ak_df = DataFrame.append(in_left, not_in_left)
+
+    return right_ak_df
+
+
+@typechecked
+def merge(
+    left: DataFrame,
+    right: DataFrame,
+    on: str,
```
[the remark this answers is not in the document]
egelberg: Correct! This would absolutely be next up on the list of enhancements, but I figured with this PR just utilizing the current functionality to create single-column inner/right/left joins would be optimal

arkouda/dataframe.py
```diff
@@ -2706,3 +2708,152 @@ def invert_permutation(perm):
...
+    how: str
```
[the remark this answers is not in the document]
egelberg: This is related to the above comment, I inherently agree but I think we'd need the ability to join on multiple columns in this instance?

arkouda/dataframe.py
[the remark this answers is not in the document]
egelberg: Agreed, I have some notebooks I can probably turn into a test file

arkouda/dataframe.py
egelberg: This is a great question, and one I hadn't accounted for.

The pandas behavior is to retain both columns and add a suffix to the end of each to indicate which is which (default is `_x` for left and `_y` for right), see pic below. This should be fairly straightforward to add

image

egelberg: Addressed with [this commit](https://github.com/Bears-R-Us/arkouda/pull/2781/commits/243ea8966be66aa47355ef2df5828163f7a67f48)

pierce314159: Nice!! We should prob add these arguments to the other functions as well

egelberg:
@pierce314159 Any idea what this failure means?

> ERROR: Cannot install blosc2==2.2.9 and oldest-supported-numpy==0.1 because these package versions have conflicting dependencies.

pierce314159:
> @pierce314159 Any idea what this failure means?
> 
> > ERROR: Cannot install blosc2==2.2.9 and oldest-supported-numpy==0.1 because these package versions have conflicting dependencies.

yeah it's not a problem specific to this PR. It's a dependency issue that PR #2808 will fix. Once we get that merged, we'll merge/rebase to get those changes in this PR

arkouda/dataframe.py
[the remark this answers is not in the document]
pierce314159: sweet! the requirement to expand `inner_join` to accept a list of pdarrays has been captured in #2810. We can update `merge` once that's been worked

jaketrookman (approved):
Looks good, the one thing that will need to be done is the addition of this testing to the new framework.

arkouda/dataframe.py
```diff
@@ -2706,3 +2823,182 @@ def invert_permutation(perm):
...
+    in_left = inner_join_merge(left, right, on, left_suffix, right_suffix)
+    in_left_cols = in_left.columns.copy()
+    in_left_cols.remove(on)
+
+    not_in_left = right[find(setdiff1d(right[on], left[on]), right[on])]
+    for col in not_in_left.columns:
+        if col in left_cols:
+            new_col = col + right_suffix
+            not_in_left[new_col] = not_in_left[col]
+            not_in_left = not_in_left.drop(col, axis=1)
```
pierce314159, line 2933: it's worth noting there is a boolean `inplace` flag for `drop`

arkouda/dataframe.py
```diff
@@ -2706,3 +2823,182 @@ def invert_permutation(perm):
...
+    DataFrame
+        Joined Arkouda DataFrame
+    """
+
+    if how == 'inner':
+        return inner_join_merge(left, right, on)
+    elif how == 'right':
+        return right_join_merge(left, right, on)
+    elif how == 'left':
+        return right_join_merge(right, left, on, left_suffix='_y', right_suffix='_x')
```
pierce314159, line 3002: i have to imagine we also we also want to pass in `left_suffix` and `right_suffix` in the `'inner'` and `'right'` join cases and also use what is passed in, instead of hardcoding in the `'left'` join case. What is here works fine if everything stays the default but will ignore any attempt by the user to change these suffixes. I'll go ahead and add these as a suggestion and commit them in the interest of getting this in for this release. If this isn't what u intended, it should be pretty easy to change
```suggestion
    if how == 'inner':
        return inner_join_merge(left, right, on, left_suffix, right_suffix)
    elif how == 'right':
        return right_join_merge(left, right, on, left_suffix, right_suffix)
    elif how == 'left':
        return right_join_merge(right, left, on, right_suffix, left_suffix)
```

pierce314159 (approved):
Looks good, there might be improvements to be made but that can be done in future iterations

jaketrookman (approved):

PR #2787, opened 2023-09-26
Closes #2461: Potential optimization for functions with `where` argument

### LINKED ISSUE #2461: Potential optimization for functions with `where` argument
Had discussion with @pierce314159 about our current implementation of `where` functions across several different implemented functions, like power, sqrt, etc.

We want to ensure that we are not wasting computational time while parsing through the `where` conditions.
Example:
If we are operating on a large array to apply an exponent, with a condition that only has a single value of true, every value in the array will be operated on.
```python
where(condition, a**power, a)
```
This does the exponentiation for every single element in this large array, but we only take the a single exponentiated value from our calculations, wasting computation time.

There is also concern that we could be applying an illegal computation in places where we know the condition is `False` and do not want to apply it, like taking the square root of a negative number.

Possible new implementation would only apply the desired operation to the desired positions; we would be creating an additional copy of the array and several variables that may take up significant space.
Example code:

```python
new_a = a[condition]
new_power = power[condition]
new_a = new_a**new_power
return_a = a[:]
return_a[condition] = new_a
```

jaketrookman:
This PR (Closes #2461) implements the `where` condition for trig functions, first draft

@pierce314159 , if you could take a look at this and see if I am on the right track for implementation and any style suggestions would be great.
If you could take a look at `sin`, `arctan2`, and `rad2deg`, that would help me apply the changes to the rest of the functions. Many of them are implemented the same way, so the other way I could condense this current version is a helper function that take in the input array(s), where condition, and the name of the trig function and handles the repMsg call logic.

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -770,19 +770,50 @@ def arctan2(
         )
         return create_pdarray(repMsg)
     elif where is False:
-        return num / denom
+        if isinstance(num, pdarray) and isinstance(denom, pdarray):
+            ret = num / denom
+        if isinstance(num, pdarray) and not isinstance(denom, pdarray):
+            ret = num / denom
+        if isinstance(denom, pdarray) and not isinstance(num, pdarray):
+            ret = num / denom
+        return ret
     else:
-        repMsg = type_cast(
-            str,
-            generic_msg(
-                cmd="efunc2",
-                args={
-                    "func": "arctan2",
-                    "A": num[where],
-                    "B": denom[where],
-                },
-            ),
-        )
+        if isinstance(num, pdarray) and isinstance(denom, pdarray):
+            repMsg = type_cast(
+                str,
+                generic_msg(
+                    cmd="efunc2",
+                    args={
+                        "func": "arctan2",
+                        "A": num[where],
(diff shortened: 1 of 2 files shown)


arkouda/numeric.py
```diff
@@ -606,17 +768,71 @@ def arctan2(num: Union[pdarray, numeric_scalars], denom: Union[pdarray, numeric_
...
-    )
+        return create_pdarray(repMsg)
+    elif where is False:
+        if isinstance(num, pdarray) and isinstance(denom, pdarray):
+            ret = num / denom
+        if isinstance(num, pdarray) and not isinstance(denom, pdarray):
+            ret = num / denom
+        if isinstance(denom, pdarray) and not isinstance(num, pdarray):
+            ret = num / denom
+        return ret
```
jaketrookman, line 779: `mypy` has problems with just a return statement of `num / denom`. The previous logic ensures that we are in a case where either one or both are pdarrays, and it kept making the assumption that this division could lead to a non-array value.

pierce314159 (commented):
I agree that factoring this out to a helper function to reduce duplicated code seems like the right idea

PR #2792, opened 2023-09-28
Closes #2789: Updates to HDF5 for Index

### LINKED ISSUE #2789: Update `index.update_hdf5`
Update `index.update_hdf5` code to have the `dataset = index`, add testing for `update_hdf5`, and change the code that will cause deprecation warnings for `1.32`

https://github.com/Bears-R-Us/arkouda/issues/2724

pierce314159:
This PR (closes #2789) makes small updates to the index hdf5 code

### DIFF
--- a/PROTO_tests/tests/io_test.py
+++ b/PROTO_tests/tests/io_test.py
@@ -1303,18 +1303,44 @@ class TestHDF5:
 
             assert isinstance(rd_idx, ak.MultiIndex)
             assert idx.to_list() == rd_idx.to_list()
-        # handle categorical cases as well
-        if dtype1 == ak.str_:
-            t1 = ak.Categorical(t1)
-        if dtype2 == ak.str_:
-            t2 = ak.Categorical(t2)
-        idx = ak.Index.factory([t1, t2])
-        with tempfile.TemporaryDirectory(dir=TestHDF5.hdf_test_base_tmp) as tmp_dirname:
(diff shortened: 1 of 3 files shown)


PROTO_tests/tests/io_test.py
```diff
@@ -1303,18 +1303,44 @@ def test_multi_index(self, dtype1, dtype2, size):
...
+                rd_idx = ak.read_hdf(f"{tmp_dirname}/idx_test*")
+
+                assert isinstance(rd_idx, ak.MultiIndex)
+                assert idx.to_list() == rd_idx.to_list()
+
+    def test_hdf_overwrite_index(self):
+        # test repack with a single object
+        a = ak.Index(ak.arange(1000))
+        b = ak.Index(ak.randint(0, 100, 1000))
+        c = ak.Index(ak.arange(15))
```
jaketrookman, line 1325: Is this worth parameterizing?

TheOddCzar789 (approved):

PROTO_tests/tests/io_test.py
pierce314159: It's not a for loop, so there's no way to parameterize. We use all three of the indices

jaketrookman: Makes sense.

jaketrookman (approved):

PR #2793, opened 2023-09-28
Closes #2472, #2800: Update IndexingMsg to use `&` instead of `mod` for setting bigint pdarrays

### LINKED ISSUE #2472: Update IndexingMsg to use `&` instead of `mod` for setting bigint pdarrays
in IndexingMsg there are several places where we use `.mod` to make sure the indices we are setting the bigint pdarray are less than `max_size`. We should do this with an `&` instead to be consistent with other code because it performs better. Here's one of the places that needs updating
https://github.com/Bears-R-Us/arkouda/blob/9bf1266da3b7abee3c2c1df672241c019b484c75/src/IndexingMsg.chpl#L1316-L1322

### LINKED ISSUE #2800: bug in bigint indexing with `max_bits` set
Reproducer:
```python
In [3]: bi = ak.arange(2**200 - 1, 2**200+4)

In [4]: bi
Out[4]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [5]: bi[:] = ak.arange(2**200 - 1, 2**200+4)

In [6]: bi
Out[6]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [7]: bi.max_bits = 201

In [8]: bi
Out[8]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [9]: bi[:] = ak.arange(2**200 - 1, 2**200+4)

In [10]: bi
Out[10]: array([3 4 5 6 7])
```

jaketrookman:
This PR (Closes #2472 and closes #2800) updates the function calls of `mod` in IndexingMsg to use `%`, as they are more efficient.
The `bool` case still needs to be updated as there is not a way to directly cast a `bigint` to a `bool`.

Leaving as a draft until I know if this last case can be completed without this casting feature.

### DIFF
--- a/PROTO_tests/tests/indexing_test.py
+++ b/PROTO_tests/tests/indexing_test.py
@@ -106,5 +106,5 @@ class TestIndexing:
 
     def test_handling_bigint_max_bits(self):
         a = ak.arange(2**200 - 1, 2**200 + 11)
-        a.max_bits = 3
+        a[:] = ak.array(a, dtype=ak.bigint, max_bits=3)
         assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
--- a/PROTO_tests/tests/indexing_test.py
+++ b/PROTO_tests/tests/indexing_test.py
@@ -106,5 +106,5 @@ class TestIndexing:
 
     def test_handling_bigint_max_bits(self):
         a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=3)
-        a[:] = ak.array(a, dtype=ak.bigint)
+        a[:] = ak.arange(2**200 - 1, 2**200 + 11)
         assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
--- a/PROTO_tests/tests/indexing_test.py
+++ b/PROTO_tests/tests/indexing_test.py
@@ -105,6 +105,6 @@ class TestIndexing:
         assert max_bits == a[:].max_bits
 
     def test_handling_bigint_max_bits(self):
-        a = ak.arange(2**200 - 1, 2**200 + 11)
-        a[:] = ak.array(a, dtype=ak.bigint, max_bits=3)
+        a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=3)
+        a[:] = ak.array(a, dtype=ak.bigint)
         assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
--- a/PROTO_tests/tests/indexing_test.py
+++ b/PROTO_tests/tests/indexing_test.py
@@ -103,3 +103,8 @@ class TestIndexing:
         a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=max_bits)
         assert max_bits == a[ak.arange(10)].max_bits
         assert max_bits == a[:].max_bits
+
+    def test_handling_bigint_max_bits(self):
+        a = ak.arange(2**200 - 1, 2**200 + 11)
+        a.max_bits = 3
+        assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -607,8 +607,8 @@ module IndexingMsg
             when (DType.BigInt, DType.BigInt) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = valueArg.getBigIntValue();
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -618,8 +618,8 @@ module IndexingMsg
             when (DType.BigInt, DType.Int64) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = valueArg.getIntValue():bigint;
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -629,8 +629,8 @@ module IndexingMsg
             when (DType.BigInt, DType.UInt64) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = valueArg.getUIntValue():bigint;
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -640,8 +640,8 @@ module IndexingMsg
             when (DType.BigInt, DType.Bool) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = valueArg.getBoolValue():bigint;
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -1161,8 +1161,8 @@ module IndexingMsg
             when (DType.BigInt, DType.BigInt) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = value.getBigIntValue();
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -1172,8 +1172,8 @@ module IndexingMsg
             when (DType.BigInt, DType.Int64) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = value.getIntValue():bigint;
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -1183,8 +1183,8 @@ module IndexingMsg
             when (DType.BigInt, DType.UInt64) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = value.getUIntValue():bigint;
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
@@ -1194,8 +1194,8 @@ module IndexingMsg
             when (DType.BigInt, DType.Bool) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = value.getBoolValue():bigint;
-                var max_size = 1:bigint;
                 if e.max_bits != -1 {
+                    var max_size = 1:bigint;
                     max_size <<= e.max_bits;
                     max_size -= 1;
                     val &= max_size;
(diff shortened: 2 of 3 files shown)


pierce314159:
@jaketrookman this isn't really what the issue was getting at

For this instead of using `%` or `.mod`, i was think we use `&`. This is what we do in other sections of code and it's more efficient. Basically instead of `val % 2**max_bits` we do `val & ((2**max_bits)-1)` so it functions as a mask instead

You can see something similar to this done in `operatorMsg`
```chpl
var max_size = 1:bigint;
var has_max_bits = max_bits != -1;
if has_max_bits {
  max_size <<= max_bits;
  max_size -= 1;
}
...
if has_max_bits {
  li &= local_max_size;
}
```

Also just looking at this code statically, unless im misreading something what's already there seems to be incorrect. Because it's doing `val % max_bits` not `val % 2**max_bits`. Let me look into this a bit

pierce314159:
yup this code is incorrect when `max_bits` is set... see reproducer below
```python
In [3]: bi = ak.arange(2**200 - 1, 2**200+4)

In [4]: bi
Out[4]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [5]: bi[:] = ak.arange(2**200 - 1, 2**200+4)

In [6]: bi
Out[6]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [7]: bi.max_bits = 201

In [8]: bi
Out[8]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [9]: bi[:] = ak.arange(2**200 - 1, 2**200+4)

In [10]: bi
Out[10]: array([3 4 5 6 7])
```

src/IndexingMsg.chpl
```diff
@@ -607,32 +607,56 @@ module IndexingMsg
...
-                if e.max_bits != -1 {
-                  mod(val, val, e.max_bits);
+                var max_size = 1:bigint;
+                var has_max_bits = e.max_bits != -1;
+                if has_max_bits {
+                    max_size <<= e.max_bits;
+                    max_size -= 1;
+                }
+                if has_max_bits {
+                    val &= max_size;
```
pierce314159, line 617: we check the same condition twice in a row, we really only need one `if` block here

In the same vein, we don't really need to declare `has_max_bits` as a variable since it should really only ever be used once

src/IndexingMsg.chpl
```diff
@@ -1308,26 +1356,44 @@ module IndexingMsg
...
-                if x.max_bits != -1 {
-                    mod(y.a, y.a, x.max_bits);
+                var max_size = 1:bigint;
+                var has_max_bits = x.max_bits != -1;
+                if has_max_bits {
+                    max_size <<= x.max_bits;
+                    max_size -= 1;
+                }
+                if has_max_bits {
+                    y.a &= max_size;
```
pierce314159, line 1366: this should be done in a forall loop over the array and use a `with ref` intent to create a task local version of `max_size`, you can look at [this PR](https://github.com/Bears-R-Us/arkouda/pull/2803/files) for reference 

This is because this operation gets turned into a forall loop kinda like the one below. See [promotion](https://github.com/Bears-R-Us/arkouda/blob/master/training/CHAPEL_TUTORIAL.md#promotion) in the chpl tutorial

```chapel
forall y in y.a {
    y &= max_size;
}
```
So `max_size` will need to be remote value forwarded to be accessible on all locales. The issue is remote value forwarding a bigint used to be pretty inefficient. To workaround this in the past we created a task local copy of `max_size`, which would look something like
```chapel
forall y in y.a with (var local_max_size = max_size) {
    y &= local_max_size;
}
```

At some point this was fixed (i don't remember exactly when), so this might not be a problem anymore. Maybe @bmcdonald3 or @ronawho knows offhand

pierce314159 (changes requested):

src/IndexingMsg.chpl
bmcdonald3: Looking at https://github.com/chapel-lang/chapel/blob/main/CHANGES.md#performance-optimizations--improvements-1, it looks like this was fixed in the 1.31 release, so I think as long as we support 1.30, we still need these.

pierce314159: awesome! Thanks @bmcdonald3!!

jaketrookman: Updated

src/IndexingMsg.chpl
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -103,3 +103,8 @@ def test_bigint_indexing_preserves_max_bits(self):
         a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=max_bits)
         assert max_bits == a[ak.arange(10)].max_bits
         assert max_bits == a[:].max_bits
+
+    def test_handling_bigint_max_bits(self):
+        a = ak.arange(2**200 - 1, 2**200 + 11)
+        a.max_bits = 3
+        assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
```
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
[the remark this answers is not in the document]
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -103,3 +103,8 @@ def test_bigint_indexing_preserves_max_bits(self):
         a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=max_bits)
         assert max_bits == a[ak.arange(10)].max_bits
         assert max_bits == a[:].max_bits
+
+    def test_handling_bigint_max_bits(self):
+        a = ak.arange(2**200 - 1, 2**200 + 11)
+        a[:] = ak.array(a, dtype=ak.bigint, max_bits=3)
+        assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
```
pierce314159, line 110: this still isn't testing what u think it is, this test will still work on master. I ran the below python block on master, hopefully it'll make it a bit easier to see what the problem is. If not, we can jump on a call and talk through it
```python
# this is what you test is doing, as you can see it works on master without your fix
>>> a = ak.arange(2**200 - 1, 2**200 + 11)
>>> a[:] = ak.array(a, dtype=ak.bigint, max_bits=3)
>>> a
array([7 0 1 2 3 4 5 6 7 0 1 2])

# this is what's broken on master
>>> a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=3)
>>> a[:] = ak.arange(2**200 - 1, 2**200 + 11)
>>> a
array([0 1 2 0 1 2 0 1 2 0 1 2])
```
A good way to avoid this in the future is to write your tests first for bug fixes. By verifying that a test was failing and is no longer, you can have more confidence you actually fixed the problem

src/IndexingMsg.chpl
```diff
@@ -607,32 +607,44 @@ module IndexingMsg
             when (DType.BigInt, DType.BigInt) {
                 var e = toSymEntry(gEnt,bigint);
                 var val = valueArg.getBigIntValue();
+                var max_size = 1:bigint;
```
pierce314159, line 610: i think let's declare this `var` within the `if` since it's only ever used `if e.max_bits != -1`, so there's no point declaring it if not

src/IndexingMsg.chpl
```diff
@@ -1308,26 +1332,37 @@ module IndexingMsg
...
             when (DType.BigInt, DType.Int64) {
                 var x = toSymEntry(gX,bigint);
                 var y = toSymEntry(gY,int);
                 var ya = y.a:bigint;
+                var max_size = 1:bigint;
                 if x.max_bits != -1 {
-                    mod(ya, ya, x.max_bits);
+                    max_size <<= x.max_bits;
+                    max_size -= 1;
+                    ya &= max_size;
```
pierce314159, line 1353: you only updated one case of the select statement to do:
```chapel
forall y in y.a with (var local_max_size = max_size) {
    y &= local_max_size;
}
```
You need to update all the others too

tests/indexing_test.py
```diff
@@ -89,3 +89,8 @@ def test_bigint_indexing_preserves_max_bits(self):
         a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=max_bits)
         self.assertEqual(max_bits, a[ak.arange(10)].max_bits)
         self.assertEqual(max_bits, a[:].max_bits)
+
+    def test_handling_bigint_max_bits(self):
+        a = ak.arange(2**200 - 1, 2**200 + 11)
+        a[:] = ak.array(a, dtype=ak.bigint, max_bits=3)
+        self.assertListEqual([7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2], a.to_list())
```
pierce314159, line 96: same deal this test also needs to be updated

pierce314159 (changes requested):

src/IndexingMsg.chpl
jaketrookman: Updated

src/IndexingMsg.chpl
jaketrookman: Updated

tests/indexing_test.py
jaketrookman: Updated

PROTO_tests/tests/indexing_test.py
```diff
@@ -103,3 +103,8 @@ def test_bigint_indexing_preserves_max_bits(self):
         a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=max_bits)
         assert max_bits == a[ak.arange(10)].max_bits
         assert max_bits == a[:].max_bits
+
+    def test_handling_bigint_max_bits(self):
+        a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=3)
+        a[:] = ak.array(a, dtype=ak.bigint)
+        assert [7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2] == a.to_list()
```
pierce314159, line 110: So this does test the fix, but it's not as representative for bigint as I'd like.

This is because we are using the same array `a`, but this array has already been modded into the right range so instead of being `ak.arange(2**200 - 1, 2**200 + 11)` at time of indexing, it is already `[7, 0, 1, 2, 3, 4, 5, 6, 7, 0, 1, 2]`

On master it does in fact give a different result (which is good cause it means it tests the fix in this PR)
```python
In [3]: a = ak.arange(2**200 - 1, 2**200 + 11, max_bits=3)

In [4]: a
Out[4]: array([7 0 1 2 3 4 5 6 7 0 1 2])

In [5]: a[:] = ak.array(a, dtype=ak.bigint)

In [6]: a
Out[6]: array([1 0 1 2 0 1 2 0 1 0 1 2])
```

But I would rather it be `a[:] = ak.arange(2**200 - 1, 2**200 + 11)`

pierce314159 (commented):
one nit pick on the tests but the logic looks great!

PROTO_tests/tests/indexing_test.py
jaketrookman: Updated, good catch

pierce314159 (approved):
Looks good!

Issue #2788, opened 2023-09-27
NaNs & int dtypes in the `cast` method


egelberg:
### **Feature Description**
Arkouda's cast function revealed a particular behavior that is contrary to pandas: if a NaN value is being added to a column, and that column was previously of dtype uint, the NaN value is replaced with 0 ([documentation](https://bears-r-us.github.io/arkouda/autoapi/arkouda/numeric/index.html#arkouda.numeric.cast)) and the dtype is maintained.

The difference in pandas is that the NaN values are kept, and the entire column is recast to a float (see example below)

I'm of the opinion that casting NaNs as 0s could be potentially harmful in data conversion (especially as we're exploring importing/exporting files to/from parquet), and could potentially start to skew distributions at scale. However, this appears to have been [a purposeful decision](https://github.com/Bears-R-Us/arkouda/blame/master/arkouda/numeric.py#L82-L92).

After discussing with the team today, there seems to be agreement that NaNs should be handled in the same method as in Pandas (casting the integer column as a float column to accomodate NaNs).

### **Example Behavior**
**`left_df`**
```
uid | int_col
------------
aaa | 0
bbb | 1
```
**`right_df`**
```
uid | name
------------
aaa | Bob
bbb | Tom
ccc | Kay
```
**Current Right Join utilizing the `cast` method would produce the following**
```
uid | name | int_col
-----------------
aaa | Bob  | 0
bbb | Tom  | 1
ccc | Kay  | 0
```
**Utilizing the suggested change in the `cast` method would produce the following**
```
uid | name | int_col
-----------------
aaa | Bob  | 0.0
bbb | Tom  | 1.0
ccc | Kay  | NaN
```
**This is in line with current pandas behavior**

PR #2813, opened 2023-10-13
Fixes #2804: Missing values in CSV bug

### LINKED ISSUE #2804: Server crashes reading CSV file with missing values
**Describe the bug**
Reading in a CSV file with the Arkouda `read_csv` method causes the server to shutdown or return a non-descriptive error message if some column data is missing for some of the rows in the CSV file. 

There are two possible outcomes.
1. The server shuts down with a `Caught a fatal signal (proc 0): SIGSEGV(11)` error.
2. The server does not shut down but the user gets a `RuntimeError: 0 errors:` message in Python with accompanying server message: `run Line 720 ERROR [Chapel] <<< readcsv resulted in error: 0 errors:`.

**To Reproduce**
Run the Python code below and attempt to read in a CSV file with at least one missing "cell".

Code:
```Python
nodes_df = ak.read_csv("path/to/nodes.csv")
```

File:
```
**HEADER**
int64,str,int64,str,str
*/HEADER/*
vertex,name,born,brand,model
34,Ann,1990,,
69,Dan,1975,,
89,,,Tesla,Model X
89,,,Tesla,Model X
```

**Expected behavior**
Have the server not shut down and return a descriptive message to the user saying empty cells in CSV files are not currently accepted, or load in the data with `NaN`s or empty cells.

**Error Message**
Provided in error description above.

**Is this a Blocking Issue**
The only work-around I came up with was loading in placeholder data for the cells. This would probably not be feasible for larger CSV files. 

Example:
```
**HEADER**
int64,str,int64,str,str
*/HEADER/*
vertex,name,born,brand,model
34,Ann,1990,X,X
69,Dan,1975,X,X
89,X,0,Tesla,Model X
89,X,0,Tesla,Model X
```

**`ak.get_config()` Output**
```
{'arkoudaVersion': 'v2023.09.06', 'chplVersion': '1.32.0', 'ZMQVersion': '4.3.4', 'HDF5Version': '1.12.2', 'serverHostname': 'n61.cluster.local', 'ServerPort': 5555, 'numLocales': 2, 'numPUs': 128, 'maxTaskPar': 128, 'physicalMemory': 549739035728, 'distributionType': 'BlockDom(1,int(64),one,unmanaged DefaultDist)', 'LocaleConfigs': [{'id': 0, 'name': 'n61.cluster.local', 'numPUs': 128, 'maxTaskPar': 128, 'physicalMemory': 549739035728}, {'id': 1, 'name': 'n62.cluster.local', 'numPUs': 128, 'maxTaskPar': 128, 'physicalMemory': 549739035728}], 'authenticate': False, 'logLevel': 'INFO', 'logChannel': 'CONSOLE', 'regexMaxCaptures': 20, 'byteorder': 'little', 'autoShutdown': False, 'serverInfoNoSplash': False}
```

**Additional context**
None.

pierce314159:
Missing values in a CSV are interpreted as an empty string, which causes a bad cast when the column dtype `!= string`. This PR (fixes #2804) throws an `IOError` for missing values in a CSV file unless the `allow_errors` flag is set

If `allow_errors` is True, then any missing values are replaced with `min(col_dtype)`. In the future, we can look into converting the column to a float and using `nan`s. This matches pandas, but we don't know if there's missing values in a column until we've already started reading. So I'm not convinced there's gonna be an efficient way to do this. I'll sync with @egelberg and @alvaradoo and see what makes since

Working this bug uncovered an issue when an error is thrown in a parallel loop, this is being investigated further in https://github.com/chapel-lang/chapel/issues/23648. This would lose information on the error (giving the error message `RuntimeError: 0 errors:` and occasionally resulting in server crashes with `SIGSEGV(11)/SIGBUS(10)`

This PR works around this by using a boolean, `hadError`, and doing an or reduction across all the tasks, so we can throw the error after the parallel loop has completed

This is what the error message and result looks like with multi-locale on the reproducer from the issue. @alvaradoo if u don't mind taking a look and making sure it's seems reasonable, I'd appreciate it!
```python
>>> ak.read_csv("empty_test.csv")
Cell In[3], line 1
----> 1 ak.read_csv("empty_test.csv")

File ~/proj/arkouda/arkouda/io.py:995, in read_csv(filenames, datasets, column_delim, allow_errors)
    992 elif datasets is None:
    993     datasets = get_columns(filenames, col_delim=column_delim, allow_errors=allow_errors)
--> 995 rep_msg = generic_msg(
    996     cmd="readcsv",
    997     args={
    998         "filenames": filenames,
    999         "nfiles": len(filenames),
   1000         "datasets": datasets,
   1001         "num_dsets": len(datasets),
   1002         "col_delim": column_delim,
   1003         "allow_errors": allow_errors,
   1004     },
   1005 )
   1006 rep = json.loads(rep_msg)  # See GenSymIO._buildReadAllMsgJson for json structure
   1007 _parse_errors(rep, allow_errors)

File ~/proj/arkouda/arkouda/client.py:970, in generic_msg(cmd, args, payload, send_binary, recv_binary)
    968     else:
    969         assert payload is None
--> 970         return cast(Channel, channel).send_string_message(cmd=cmd, args=msg_args,
    971                                                           size=size, recv_binary=recv_binary)
    972 except KeyboardInterrupt as e:
    973     # if the user interrupts during command execution, the socket gets out
    974     # of sync reset the socket before raising the interrupt exception
    975     cast(Channel, channel).connect(timeout=0)

File ~/proj/arkouda/arkouda/client.py:503, in ZmqChannel.send_string_message(self, cmd, recv_binary, args, size, request_id)
    501 # raise errors or warnings sent back from the server
    502 if return_message.msgType == MessageType.ERROR:
--> 503     raise RuntimeError(return_message.msg)
    504 elif return_message.msgType == MessageType.WARNING:
    505     warnings.warn(return_message.msg)

RuntimeError: IOError Line 385 In CSVMsg.read_files_into_dist_array: This CSV is missing values. To read anyway and replace these with min(col_dtype), set allow_errors to True

>>> ak.read_csv("empty_test.csv", allow_errors=True)
{'vertex': array([34 69 89 89]),
 'name': array(['Ann', 'Dan', '', '']),
 'born': array([1990 1975 -9223372036854775808 -9223372036854775808]),
 'brand': array(['', '', 'Tesla', 'Tesla']),
 'model': array(['', '', 'Model X', 'Model X'])}
```

### DIFF
--- a/src/CSVMsg.chpl
+++ b/src/CSVMsg.chpl
@@ -248,7 +248,7 @@ module CSVMsg {
                                     lineNumber=getLineNumber(), 
                                     routineName=getRoutineName(), 
                                     moduleName=getModuleName(), 
-                                    errorClass='IOError'
+                                    errorClass="IOError"
                             );
                         }
                     }
@@ -323,9 +323,9 @@ module CSVMsg {
         return (row_ct, hasHeader, new list(dtypes));
     }
 
-    proc read_files_into_dist_array(ref A: [?D] ?t, dset: string, filenames: [] string, filedomains: [] domain(1), skips: set(string), hasHeaders: bool, col_delim: string, offsets: [] int) throws {
-
-        coforall loc in A.targetLocales() with (ref A) do on loc {
+    proc read_files_into_dist_array(ref A: [?D] ?t, dset: string, filenames: [] string, filedomains: [] domain(1), skips: set(string), hasHeaders: bool, col_delim: string, offsets: [] int, allowErrors: bool) throws {
+        var hadError = false;
+        coforall loc in A.targetLocales() with (ref A, | reduce hadError) do on loc {
             // Create local copies of args
             var locFiles = filenames;
             var locFiledoms = filedomains;
@@ -359,9 +359,18 @@ module CSVMsg {
                     for locdom in A.localSubdomains() {
                         const intersection = domain_intersection(locdom, filedom);
                         if intersection.size > 0 {
-                            forall x in intersection {
+                            forall x in intersection with (| reduce hadError) {
                                 var row = lines[x-offsets[file_idx]+data_offset].split(col_delim);
-                                A[x] = row[colidx]: t;
+                                if t == string || row[colidx] != "" {
+                                    // only write into A[x] if the value is non-empty and the col dtype is not a string
+                                    A[x] = row[colidx]: t;
+                                }
+                                else if allowErrors {
+                                    A[x] = min(t);
+                                }
+                                else {
+                                    hadError |= true;
+                                }
                             }
                         }
                     }
@@ -369,6 +378,14 @@ module CSVMsg {
                     csvFile.close();
                 }
             }
+        }
+        if hadError {
+            throw getErrorWithContext(
+                msg="This CSV is missing values. To read anyway and replace these with min(col_dtype), set allow_errors to True",
+                lineNumber=getLineNumber(),
+                routineName=getRoutineName(),
+                moduleName=getModuleName(),
+                errorClass="IOError");
         }
     }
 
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good

pierce314159:
Just want to link this PR to #2788 since these both touch on auto converting something with missing values to float and using `NaN`s

PR #2821, opened 2023-10-23
Add support for reading Decimal128 Parquet columns

bmcdonald3:
Previously, when reading a file with a Decimal128 column, Arkouda would silently ignore the column and the data would be dropped. This PR adds the ability to read Decimal128 Parquet columns, but they are cast to float C++ values and stored into a 64-bit `real` Chapel array, so some precision can be lost in the conversion.

Longer term, we would like to add support for arbitrary-precision values into Chapel, which would result in a new type, but that would be able to more accurately display Decimal128 values.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -112,6 +112,8 @@ int cpp_getType(const char* filename, const char* colname, char** errMsg) {
       return ARROWDOUBLE;
     else if(myType->id() == arrow::Type::LIST)
       return ARROWLIST;
+    else if(myType->id() == arrow::Type::DECIMAL)
+      return ARROWDECIMAL;
     else {
       std::string fname(filename);
       std::string dname(colname);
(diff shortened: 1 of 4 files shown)


bmcdonald3:
Marking as draft as I am realizing some additional checks need to be done in the C++ code that I am unsure of how to do at the moment.

bradcray:
> we would like to add support for arbitrary-precision values into Chapel

Capturing some discussion Ben and I were having off-repo:  Another option would be to provide support for fixed-size fixed-point values in Chapel, which would be more space-efficient than arbitrary-precision values since no heap storage would be required.

src/ArrowFunctions.cpp
```diff
@@ -849,6 +876,24 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
...
+        auto chpl_ptr = (double*)chpl_arr;
+        parquet::FixedLenByteArray value;
+        parquet::FixedLenByteArrayReader* reader =
+          static_cast<parquet::FixedLenByteArrayReader*>(column_reader.get());
+        startIdx -= reader->Skip(startIdx);
+
+        while (reader->HasNext() && i < numElems) {
+          if((numElems - i) < batchSize)
+            batchSize = numElems - i;
+          (void)reader->ReadBatch(1, nullptr, nullptr, &value, &values_read);
```
pierce314159, line 889: I don't really see how `batchSize` factors into this since we are always calling `ReadBatch` with a size of 1

src/ArrowFunctions.cpp
```diff
@@ -849,6 +876,24 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
...
+          arrow::Decimal128 v;
+          PARQUET_ASSIGN_OR_THROW(v,
+                                  ::arrow::Decimal128::FromBigEndian(value.ptr, byteLength));
```
pierce314159, line 892: does it matter if the server's byte order is little endian? I just want to make sure `FromBigEndian` will still work

pierce314159 (commented):
Looks good, just want to verify that big vs little endian doesn't matter before approving

src/ArrowFunctions.cpp
bmcdonald3: You are right, that isn't doing anything in this case, it was just leftover from copy/paste of the integer section. I'll remove that.

src/ArrowFunctions.cpp
bmcdonald3: No, the `FromBigEndian` is referring to the Arrow `FixedLenByteArray` type, which is always stored in big endian itself. There isn't a way to convert from a `FixedLenByteArray` type itself to a decimal, so here we have to use the `FromBigEndian` helper function on the byte array pointer, which is always big endian, regardless of machine architecture.

jaketrookman (approved):

pierce314159 (approved):

PR #2411, opened 2023-05-05
Add scaffolding for multi dimensional arrays in Arkouda

bmcdonald3:
This PR updates the `SymEntry` class to have a `param dimensions` field that allows stamping out of multiple different pdarray dimensions in preparation for adding Array API functionality. No behavior will change after this PR is merged, but development of multi-dimensional pdarrays of the form below will be able to begin:

1. Add a `param dimensions=1` field to a function to add functionality to, like:
```chpl
proc randintMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, param dimensions=1): MsgTuple throws {
```

2. Update the code in the function to behave as expected with the different dimensions

3. Add a 1D, 2D... function to call this new function with different dimensions using signatures that can be registered in the command map:
```chpl
    proc randintMsg1D(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
      return randintMsg(cmd, msgArgs, st, 1);
    }

    proc randintMsg2D(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
      return randintMsg(cmd, msgArgs, st, 2);
    }
```

4. Register the new functions:
```chpl
    registerFunction("randint1D", randintMsg1D, getModuleName());
    registerFunction("randint2D", randintMsg2D, getModuleName());
```

5. update the client to somehow send the new messages as desired

PR #2829, opened 2023-10-25
Scaffolding for multi-dimensional arrays

jeremiah-corrado:
Continuing off of @bmcdonald3's work in https://github.com/Bears-R-Us/arkouda/pull/2411, this PR adds some scaffolding for supporting ND arrays:

* `SymEntry` is now generic w.r.t. its array rank
* `createMsg` and `setMsg` are set up to support ND arrays as a proof of concept
* `__str__` is supported for ND arrays

Additionally, the build system is modified such that users can controll a server's maximum array rank (dimensionality) by setting the  `max_array_dims` field in `ServerModules.cfg`. At build time, `serverModuleGen.py` will read that value and create multidimensional message handlers up to the given dimension for any message handler that opts in to ND support.

### Opting into ND support for a message handler

1. add a `param nd: int` formal to the procedure, e.g.,
```chapel
  proc fooMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, param nd: int = 1): MsgTuple throws { ... }
```

2. modify the procedure to handle muti-dimensional arrays. Because `nd` is a param value, it can be used to call out to different implementations for various ranks at compile time, e.g., 
```chapel
  if nd == 1
    then return fast1DImpl(cmd, msgArgs, st);
    else return multiDimImpl(cmd, msgArgs, st);
```

3. add an `@arkouda.registerND` annotation to the procedure. This will handle instantiating and registering the message handler for each supported dimensionality.
```chapel
@arkouda.registerND
proc fooMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, param nd: int = 1): MsgTuple throws { ... }
```

This will make the following message commands available: `foo1D`, `foo2D`, ... (up to the servers maximum supported rank).

<details>

<summary>how `@arkouda.registerND` works...</summary>

At build time, the `serverModuleGen.py` script will generate a file: `src/multi_dim_support/nd_array_stamps.chpl` with instatiations of each message handler annotated with `@arkouda.registerND`.

For the above example, with `max_array_dims: 3`, the `nd_array_stamps.chpl` file will contain the following:

```chapel
use CommandMap, Message, MultiTypeSymbolTable;

proc _nd_gen_fooMsg1D(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws do
  return fooMsg(cmd, msgArgs, st, 1);
registerFunction("foo1D", _nd_gen_fooMsg1D);
                
proc _nd_gen_fooMsg2D(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws do
  return fooMsg(cmd, msgArgs, st, 2);
registerFunction("foo2D", _nd_gen_fooMsg2D);
                
proc _nd_gen_fooMsg3D(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws do
  return fooMsg(cmd, msgArgs, st, 3);
registerFunction("foo3D", _nd_gen_fooMsg3D);

use Foo.chpl
```

This file is compiled with the other chapel source code, making the command instatiations available.

</details>


arkouda/pdarraycreation.py
```diff
@@ -490,7 +490,7 @@ def ones(
     # check dtype for error
     if dtype_name not in NumericDTypes:
         raise TypeError(f"unsupported dtype {dtype}")
-    repMsg = generic_msg(cmd="create", args={"dtype": dtype_name, "size": size})
+    repMsg = generic_msg(cmd="create1D", args={"dtype": dtype_name, "shape": size})
```
bmcdonald3, line 493: Should we make these client-side functions handle the dimensionality as well? I don't think there are instructions in the PR for how to get the client to work with ND arrays, but if we did the `"create" + str(ndim) + "D"` or something like that to build up the command string, that should just work with the increased number of dimensions I think.

bmcdonald3 (commented):
Nice work, just one comment on the client-side handling. Also, it would be cool if we could get some numbers on the build time increases that we see with varying numbers of dimensions, even with just the couple of functions stamped out.

That makes me think, it would probably be a good idea to collect some performance runs with this to make sure that the symbol table entry modifications aren't hurting things too bad... I don't think they would, especially since it's a `param` field that was added, but you never know...

arkouda/pdarraycreation.py
jeremiah-corrado: I was thinking about that, but I didn't want replace the `size` argument on `ak.zeros` and `ak.ones` with a `shape` argument for now because I wasn't sure if it's okay to modify those interfaces. Would that change be allowed?

I can definitely parameterize the calls to `set` by the arrays dimension without any interface changes though.

jeremiah-corrado:
Looks like there is a non-trivial hit to compilation time when using multiple dimensions in `createMsg` and `setMsg`

Here are some compilation time data collected on my laptop with everything commented out in ServerModules.cfg:

| `max_array_dims` | compilation time (sec) |
| :--- | :--- |
| 1 | 13.33 |
| 2 | 14.62 |
| 3 | 15.14 |
| 4 | 16.18 |
| 5 | 17.36 |
| 6 | 18.55 |

src/multi_dim_support/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
bmcdonald3, line 2: So the stamped out Chapel code will be placed here?

bmcdonald3 (approved):
This looks good to me; to check my understanding, what is happening here is that we are generating the multidimensional Chapel code, but there is nothing handled automatically on the client-side to support calling that just yet?

That makes sense to me, just making sure I've got it straight after passing over the code.

bmcdonald3:
@pierce314159 I think this PR is ready to go if you have a chance to look at it sometime. We are modifying the Chapel server to allow support of ND arrays, but they aren't actually usually from the client yet.

Since the modifications are mainly from a `param` field, this doesn't really affect anything else in the code base and Jeremiah has run some performance tests and saw little to no impact from this change.

Increasing the dimensions will increase compilation time, but we wouldn't expect people to be increasing the dimensions just yet, since they would have to write their own Python code to make that work.

Let us know if you have thoughts on this. Thanks.

src/multi_dim_support/.gitignore
jeremiah-corrado: Yep, it goes in a single source file in this directory that gets compiled with everything else.

jeremiah-corrado:
> what is happening here is that we are generating the multidimensional Chapel code, but there is nothing handled automatically on the client-side to support calling that just yet?

That's right. I was thinking we'll start adding client-side changes in a later PR.

pierce314159 (approved):
Looks good to me! Thanks @jeremiah-corrado and @bmcdonald3!!

There seems to be a merge conflict but once that's resolved, I think it should be good to go

src/AryUtil.chpl
```diff
@@ -34,13 +34,52 @@ module AryUtil
...
+                for dd in 0..<d.rank {
+                    const dMax = d.dim(dd).high;
+                    indices[3][dd] = dMax;
+                    indices[4][dd] = dMax;
+                    indices[5][dd] = dMax;
+                }
+
+                indices[3][dimIdx] = dimSize-3;
+                indices[4][dimIdx] = dimSize-2;
+                indices[5][dimIdx] = dimSize-1;
```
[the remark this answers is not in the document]
jeremiah-corrado: Yeah because of the zero based arrays those two queries end up being the same. Reading your comment and looking back through code makes me realize that it isn't super clear what this procedure is doing, so I'll comment it up a bit as well.

pierce314159 (approved):

jeremiah-corrado:
Thanks @pierce314159!

PR #2832, opened 2023-10-30
Add directory with files for Parquet C++ comparison

bmcdonald3:
This is adding some files so that a Parquet C++ comparison can be run.


pierce314159 (approved):
Looks great!! Thanks Ben

Also, I know you mention this in the README, but I wanted to note that this comparison is for `int` columns only. I think this is the right thing to do. Since there's known shortcomings in the arkouda parquet code's handling of strings, I don't think it's worth writing code for strings read comparison at this point

PR #2848, opened 2023-11-15
Closes #2838: Expand dataframe merge functions to accept multiple columns

### LINKED ISSUE #2838: Expand dataframe merge functions to accept multiple columns
Once https://github.com/Bears-R-Us/arkouda/pull/2837 is merged, `inner_join` will be able to handle sequences of pdarrays. Using that we can enable the dataframe merge functions that @egelberg added to accept multiple columns. When no columns are specified we should merge on all columns to match the pandas default

pierce314159:
This PR (closes #2838) expands the dataframe merge functions to act on multiple columns. When no value is provided for `on`, it defaults to the intersection of the columns of the left and right dataframe. `inner_join_merge` and `right_join_merge` were turned into helper functions that aren't exposed to the user to more closely match the pandas merge functionality where these are only avialble through `merge`

### DIFF
--- a/PROTO_tests/tests/dataframe_test.py
+++ b/PROTO_tests/tests/dataframe_test.py
@@ -666,3 +666,34 @@ class TestDataFrame:
         assert df.index.to_list() == df2.index.to_list()
         assert df["a"].to_list() == df2["a"].to_list()
         assert df["b"].to_list() == df2["b"].to_list()
+
+    def test_multi_col_merge(self):
+        size = 1000
+        seed = 1
+        a = ak.randint(-size // 10, size // 10, size, seed=seed)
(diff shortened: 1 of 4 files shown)


egelberg:
@pierce314159 Overall LGTM for what it's supposed to do, I think a next step would be adding support for string column merges, but this might have to wait until String optimizations are taken care of?

pierce314159:
> @pierce314159 Overall LGTM for what it's supposed to do, I think a next step would be adding support for multi-column string merges, but this might have to wait until String optimizations are taken care of?

Yea absolutely. I think the thing holding this up is the use of `in1d` which currently only acts on sequences of pdarrays and not sequences of Strings. But I agree that would be good to add

PR #2856, opened 2023-11-16
Closes #2855: Read batches for Parquet strings

### LINKED ISSUE #2855: Read batches for Parquet strings
In the first implementation of Parquet, strings were the only data type being read one at a time, rather than in batches since we can't read directly into a `uint(8)` buffer, so we were reading one value at a time and then copying it into the Chapel buffer. Rather than doing that one a time, we can switch to reading batches and then bulk copying to improve performance. Here are some performance numbers collected on 2 nodes of a Cray XC with varying batch sizes:

| batchSize | exec time |
| ---------:| ---------:|
| 1         | 9.63s     |
| 10        | 3.66s     |
| 256       | 3.05s     |
| 512       | 3.03s     |
| 8192      | 3.05s     |

So, we see about a 3x improvement.

bmcdonald3:
In the first implementation of Parquet, strings were the only data type being read one at a time, rather than in batches since we can't read directly into a `uint(8)` buffer, so we were reading one value at a time and then copying it into the Chapel buffer. Rather than doing that one a time, we can switch to reading batches and then bulk copying to improve performance. Here are some performance numbers collected on 2 nodes of a Cray XC with varying batch sizes:

| batchSize | exec time |
| ---------:| ---------:|
| 1         | 9.63s     |
| 10        | 3.66s     |
| 256       | 3.05s     |
| 512       | 3.03s     |
| 8192      | 3.05s     |

So, we see about a 3x improvement.

Closes #2855


bmcdonald3:
I believe the test that is failing is because of the `writeMultiCol` parquet function was not accounting for null values and thus didn't care about the definition level, which we now have to use to detect null values, since we can no longer just check `values_read` to give us indices of where there may be a null string. 

I think it will require some modifications to the `writeMultiCol` function in order for this to work.

src/ArrowFunctions.cpp
```diff
@@ -275,25 +275,47 @@ int64_t cpp_getStringColumnNumBytes(const char* filename, const char* colname, v
...
+          if((numElems - numRead) < batchSize)
+            batchSize = numElems-numRead;
+          std::vector<parquet::ByteArray> string_values(batchSize);
+          std::vector<int16_t> definition_level(batchSize);
+          (void)ba_reader->ReadBatch(batchSize, definition_level.data(), nullptr, string_values.data(), &values_read);
+          numRead += values_read;
+          if (ty == ARROWSTRING) {
+            auto numCols = file_metadata -> num_columns();
+            int string_index = 0;
+            for(int idx = 0; idx < definition_level.size(); idx++) {
```
pierce314159, line 292: This doens't need to be updated because they should always be equivalent but I feel like `idx < batchSize` is 
a bit clearer. But since we're using `idx` to indexing to `definition_level` maybe this is better? idk i don't have terribly strong feeligns

jaketrookman (approved):

src/ArrowFunctions.cpp
bmcdonald3: Well, the thing is that `idx < batchSize` is not the same as `idx < definition_level.size()`. When you read a batch in Parquet, sometimes some of the reads will "fail", so you don't actually read a full `batchSize` in.

The `definition_level.size()` will return the number of elements that were actually read in and initialized by the `ReadBatch` call, so you can think of something like `ReadBatch(10,...)` actually only reading in 6 values or something like that, which would result in `definition_level.size()` being 6, so that is why we have that distinction here.

src/ArrowFunctions.cpp
```diff
@@ -275,25 +275,47 @@ int64_t cpp_getStringColumnNumBytes(const char* filename, const char* colname, v
...
+              auto lvl = definition_level[idx];
+              auto value = string_values[string_index];
+              if(lvl != 0 || numCols > 1) {
+                auto value = string_values[string_index];
```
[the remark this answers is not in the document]
bmcdonald3: You've caught me redhanded... this was a fragment of something that I was trying earlier, where I was basing one of the conditionals on `value`, but couldn't do that since `value` could potentially be uninitialized.

Will remove that first value set.

src/ArrowFunctions.cpp
pierce314159: oh gotcha! Yah def leave it as that then 😁

pierce314159 (approved):

jaketrookman (approved):

PR #2860, opened 2023-11-24
Fixes #2666: Segarray.filter bug with empty segments

### LINKED ISSUE #2666: `SegArray.filter()` Incorrect Behavior with Empty Segments
`SegArray.filter()` appears to be dropping the segments that are empty before filtering. Empty segments resulting from the filter are handled properly.

```python
# NOTE - uses function to build segarray components used in testing
a, b = self.build_repeat_filter_data(dtype)
segments = ak.array([0, len(a), len(a), len(a), len(a)+len(b)])
flat = ak.array(a + b)
sa = ak.SegArray(segments, flat)
result = sa.filter(3)
```

pierce314159:
This PR (fixes #2666) fixes a bug where segarray.filter always dropped empty segments

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -11,7 +11,7 @@ import numpy as np  # type: ignore
 from arkouda.client import generic_msg
 from arkouda.dtypes import bool as akbool
 from arkouda.dtypes import int64 as akint64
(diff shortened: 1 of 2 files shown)


QuinsulonPhD (approved):
With my current knowledge of Arkouda, the logic appears sound.

jaketrookman (approved):

PR #2861, opened 2023-11-24
Fixes #2661: `segarray.remove_repeats` bug with empty segments

### LINKED ISSUE #2661: `SegArray.remove_repeats()` Empty First Segment Incorrect Behavior
While updating the SegArray testing for the new framework, I discovered that `remove_repeats` returns a negative value for the 2nd segment value which is not correct. I provided example code below to reproduce. We will need to update the new test framework for this when it is resolved. I have a note in the code to do so.

```python
segments = ak.array([0, 0, len(a), len(a), len(a), len(a)+len(b)])
flat = ak.array(a + b)
sa = ak.SegArray(segments, flat)
result = sa.remove_repeats()
```

In `ak.SegArray.remove_repeats`, if you add a `print(truesegs)` on ln 713, you will see the negative referenced above.

pierce314159:
This PR (fixes #2661) fixes a bug in `segarray.remove_repeats` handling of empty segments. This was a very similar problem as in #2666

### DIFF
--- a/arkouda/segarray.py
+++ b/arkouda/segarray.py
@@ -661,18 +661,10 @@ class SegArray:
         isrepeat[self.segments[self.non_empty]] = False
         truepaths = self.values[~isrepeat]
         nhops = self.grouping.sum(~isrepeat)[1]
-        truesegs = cumsum(nhops) - nhops
         # Correct segments to properly assign empty lists - prevents dropping empty segments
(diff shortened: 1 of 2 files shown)


QuinsulonPhD (approved):
Same remarks towards lack of knowledge of operation-- the logic appears sound.

jaketrookman (approved):

PR #2865, opened 2023-11-28
Array API infrastructure and partial implementation

jeremiah-corrado:
Building off of https://github.com/Bears-R-Us/arkouda/pull/2829, adds all the function signatures and a general skeleton for the [Python Array API](https://data-apis.org/array-api/latest/), as well as some of the implementation. 

The frontend implementation lives in its own directory `/arkouda/array_api/` mirroring what [numpy does](https://github.com/numpy/numpy/tree/main/numpy/array_api). This allows a user to import the arkouda implementation with:
```python
import arkouda.array_api as Array
```
Which can then be plugged into other Python libraries that are generic wrt their `Array` implementation.

---

This PR also has some of the backend implementation, including:
* [array creation](https://data-apis.org/array-api/latest/API_specification/creation_functions.html)
* [element wise operations](https://data-apis.org/array-api/latest/API_specification/elementwise_functions.html)
* [linear algebra](https://data-apis.org/array-api/latest/API_specification/linear_algebra_functions.html) (except `tensordot`)
* [argsort](https://data-apis.org/array-api/latest/API_specification/generated/array_api.argsort.html#array_api.argsort)
* [broadcasting](https://data-apis.org/array-api/latest/API_specification/broadcasting.html)

Note that the linear algebra implementations have not been optimized for performance — this can come in a later PR.

The current broadcasting implementation is also somewhat antithetical to the intention of broadcasting [^1] because it copies values from degenerate (single element) dimensions to fill a larger array. This is the simplest solution and will work okay, but could use excess memory. Instead, we may want to consider adding fields to the `SymEntry` class that allow individual dimensions to be marked as "broadcasted". When used in a computation (e.g., a dot product), the value in the "broadcasted" dimension could be used in a promoted sense instead of being copied into an array of the proper size ahead of time.

Additionally, we may want to adjust how scalar arrays ("zero-dimensional arrays") and uninitialized arrays (created from [`empty`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.empty.html#array_api.empty)) are handled. Specifically, it may make sense to create separate classes for them rather than using a single-element 1D array, and an `empty` marker on the server side respectively.

[^1]: From the docs: "Broadcasting refers to the automatic (implicit) expansion of array dimensions to be of equal sizes without copying array data for the purpose of making arrays with different shapes have compatible shapes for element-wise operations."


bmcdonald3:
Hey @pierce314159, this PR looks huge and intimidating to review, but I'll note that it is actually much less so when you consider what is actually in it.

For the most part, this PR is not touching existing code, where things that are added in the `/arkouda/array_api` folder, rather than just the `/arkouda/` folder, where our existing stuff lives. 

What that means is that this is not impacting existing code, but is rather just adding a second new array interface that is largely a direct translation of something like https://github.com/numpy/numpy/tree/main/numpy/array_api. Because of this, I don't think a lot of this needs to be looked at deeply, where there is a lot of placeholder code for functionality to come, so, given the lack of impact on the existing Arkouda framework (i.e., none of Arkouda standard unit tests or any code that users are using would be affected by these changes), my stance is that this should be merged to allow follow up work and doesn't need a ton of effort to review at this point.

All of that is really to say, this is really a starting point to build off of and doesn't affect existing code, so I don't think there should be concern about needing to spend a lot of time looking at every line here.

bmcdonald3 (approved):
Looks good, exciting stuff 🥳

PR #2868, opened 2023-11-29
Closes #2863: Implement `histogramdd`

### LINKED ISSUE #2862: Implement `histogram2d`
@ItsQuinnMoore's work in https://github.com/Bears-R-Us/arkouda-contrib/pull/124 seems very similar to the functionality of `ak.histogram`. This issue is to investigate if there's a way to slightly modify existing arkouda code to get a 2d (or even n-dimensional) histogram

arbitrary dimension histogram will be added in #2863

### LINKED ISSUE #2863: Implement `histogramdd`
Related to #2862, implement [`histogramdd`](https://numpy.org/doc/stable/reference/generated/numpy.histogramdd.html) respectively

pierce314159:
This PR (closes #2863) adds the logic for a histogram with an arbitrary nubmer of dimensions. This mimics `np.histogramdd`. Added tests against the numpy version to ensure we have the same resulting histogram when acting on the same arrays.

This is based on the existing single dimensional histogram, so it has 3 different implementations based on number of bins to take advantage of the time / space tradeoff when total number of bins is relatively small. I tested all three versions against numpy with and without gasnet enabled

This is related to the work being done by @ItsQuinnMoore

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -1325,7 +1325,7 @@ def histogram2d(
     Examples
     --------
     >>> x = ak.arange(0, 10, 1)
-    >>> y = ak.arange(9, 0, -1)
+    >>> y = ak.arange(9, -1, -1)
     >>> nbins = 3
     >>> h, x_edges, y_edges = ak.histogram2d(x, y, bins=nbins)
     >>> h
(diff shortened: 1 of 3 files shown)


quinsulon (approved):
I see no issues, given my current understanding of the Chapel syntax.

pierce314159:
converting to draft until #2872 is merged and then I'll need to make the same update to this test

jaketrookman (approved):
Looks good

PR #2869, opened 2023-11-30
Array creation perf patch

jeremiah-corrado:
https://github.com/Bears-R-Us/arkouda/pull/2829 introduced a perf regression in array creation performance.

This is likely due to the [IO-based tuple parsing](https://github.com/Bears-R-Us/arkouda/blob/4d6141411f7d811f22daa3804696512f4b36ed0c/src/Message.chpl#L458-L474) that was added to `createMsg` to parse the array shape. This PR bypasses that code at compile time when `nd` is 1 to avoid the multidimensional array overhead for the common case.

### DIFF
--- a/src/Message.chpl
+++ b/src/Message.chpl
@@ -258,11 +258,10 @@ module Message {
         Parse value as a tuple of integers with the given size
         */
         proc getTuple(param size: int): size*int throws {
-            try {
+            if size == 1 {
+                return (this.getIntValue(),);
+            } else {
                 return parseJsonTuple(this.val, size);
-            } catch {
-                var x: size*int; x[0] = this.getIntValue();
-                return x;
             }
         }
 


pierce314159 (approved):
looks good to me

bmcdonald3:
Do we have any performance numbers collected for this change? I'd be curious to see the impact on a real machine before merging.

jeremiah-corrado:
It looks like #2829 also made a change to the array declaration in `SymEntry` which was causing a slowdown to array creation. With that fixed, I'm seeing the following performance for the array-creation benchmark on a 16 node machine:

```
zeros Average rate = 429.13 GiB/sec
ones Average rate = 292.98 GiB/sec
randint Average rate = 286.91 GiB/sec
```

which is on par with the performance we were seeing before the ND-array PR went in.

jaketrookman (approved):

PR #2870, opened 2023-12-01
DataFrame initializer improvements

brandon-neth:
Two changes (with tests) and a docstring.
1. Column labels can be provided as an optional argument for use with unlabelled data
2. DataFrames can be initialized with dicts/lists that contain native lists and tuples in addition to pdarrays. Lists and tuples are converted to pdarrays as part of the initialization.

Tested locally.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -205,7 +205,7 @@ class DataFrame(UserDict):
 
     objType = "DataFrame"
 
-    def __init__(self, initialdata=None, index=None):
+    def __init__(self, initialdata=None, index=None, columns=None):
         super().__init__()
         self.registered_name = None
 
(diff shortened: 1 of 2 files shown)


arkouda/dataframe.py
```diff
@@ -278,8 +294,19 @@ def __init__(self, initialdata=None, index=None):
             # Initial data is a list of arkouda arrays
             elif isinstance(initialdata, list):
                 # Create string IDs for the columns
-                keys = [str(x) for x in range(len(initialdata))]
+                keys = []
+                if columns is not None:
+                    if any([not isinstance(label, str) for label in columns]):
```
pierce314159, line 299: Small suggestion, we could pass the generator into `any` directly and avoid the list creation
```suggestion
                    if any(not isinstance(label, str) for label in columns):
```

pierce314159 (approved):
Looks good to me!

bmcdonald3 (approved):

pierce314159 (approved):

PR #2881, opened 2023-12-13
Closes #2880: Add parallel writing when writing pdarrays to Parquet

### LINKED ISSUE #2880: Add parallel writing when writing pdarrays to Parquet
This PR adds support for a `parallelWriteThreshold` flag that
allows a user to determine the size of files of the Parquet files
to be written and then write those files in parallel, appending a
`_CORE####` to the end of the file name.

By running the Arkouda server with:
`./arkouda_server --ParquetMsg.parallelWriteThreshold=<num>`, a
user is able to control the size of the files that are going to be
written.

This is currently only supported on pdarrays of natively-supported
datatypes (meaning not strings or dataframes), but follow work is
on the way.

bmcdonald3:
This PR adds support for a `parallelWriteThreshold` flag that
allows a user to determine the size of files of the Parquet files
to be written and then write those files in parallel, appending a
`_CORE####` to the end of the file name.

By running the Arkouda server with:
`./arkouda_server --ParquetMsg.parallelWriteThreshold=<num>`, a
user is able to control the size of the files that are going to be
written.

This is currently only supported on pdarrays of natively-supported
datatypes (meaning not strings or dataframes), but follow work is
on the way.

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -46,6 +46,8 @@ module ParquetMsg {
   // Undocumented for now, just for internal experiments
   private config const batchSize = getEnvInt("ARKOUDA_SERVER_PARQUET_BATCH_SIZE", 8192);
 
+  private config const parallelWriteThreshold = 512*1024*1024 / numBytes(int);
+
   extern var ARROWINT64: c_int;
   extern var ARROWINT32: c_int;
   extern var ARROWUINT64: c_int;
@@ -417,6 +419,7 @@ module ParquetMsg {
                                         dtype, compression,
                                         errMsg): int;
     var dtypeRep = toCDtype(dtype);
+    var doParallel = if A.size > parallelWriteThreshold then true else false;
     var prefix: string;
     var extension: string;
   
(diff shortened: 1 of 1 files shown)


src/ParquetMsg.chpl
```diff
@@ -46,6 +46,8 @@ module ParquetMsg {
   // Undocumented for now, just for internal experiments
   private config const batchSize = getEnvInt("ARKOUDA_SERVER_PARQUET_BATCH_SIZE", 8192);
 
+  private config const parallelWriteThreshold = 512*1024*1024 / numBytes(int);
```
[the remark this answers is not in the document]
bmcdonald3: No, that will set the entire expression. The default expression here represents 512 MB of integer elements.

jaketrookman (approved):
Looks good

src/ParquetMsg.chpl
```diff
@@ -417,6 +419,7 @@ module ParquetMsg {
                                         dtype, compression,
                                         errMsg): int;
     var dtypeRep = toCDtype(dtype);
+    var doParallel = if A.size > parallelWriteThreshold then true else false;
```
pierce314159, line 422: okay this is pedantic, but this could just be `var doParallel = A.size > parallelWriteThreshold;` right?

src/ParquetMsg.chpl
```diff
@@ -454,10 +457,32 @@ module ParquetMsg {
...
+          if !doParallel {
+            if c_writeColumnToParquet(myFilename.localize().c_str(), valPtr, 0,
+                                      dsetname.localize().c_str(), locDom.size, rowGroupSize,
+                                      dtypeRep, compression, c_ptrTo(pqErr.errMsg)) == ARROWERROR {
+              pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+            }
+          } else {
+            var fileSizes: [0..#loc.maxTaskPar] int = locDom.size/loc.maxTaskPar;
+            // First file has the extra elements if it isn't evenly divisible by maxTaskPar
+            fileSizes[0] += locDom.size - ((locDom.size/loc.maxTaskPar)*loc.maxTaskPar);
```
pierce314159, line 469: i had to convince myself that the integer cast of the float divide would always round down. This seems to check out and the adjustment gave what i expected on a small example!
```chapel
var fileSizes: [0..#6] int = 10/6;  // 1.6666 verify this rounds down
writeln(fileSizes);
var leftOver = 10 - ((10/6)*6); 
writeln(leftOver);
```
```
1 1 1 1 1 1
4
```

I will say in my small example this resulted in a pretty unbalanced distribution, but I *think* that in a real case that `locDom.size` would be large enough relative to `loc.maxTaskPar` that it would be pretty uniform... I'm just now realizing that's probably part of the motivation for having a `parallelWriteThreshold` lol

src/ParquetMsg.chpl
```diff
@@ -454,10 +457,32 @@ module ParquetMsg {
...
+
+            var offsets = + scan fileSizes;
+
+            forall i in fileSizes.domain {
+              var suffix = '%04i'.format(idx): string;
+              var parSuffix = '%04i'.format(i): string;
+              const parFilename = filename + "_LOCALE" + suffix + "_CORE" + parSuffix + ".parquet";
+              var oi = if i == 0 then i else offsets[i-1];
```
pierce314159, line 477: instead of doing this back by one, couldn't you do
```chapel
var offsets = (+ scan fileSizes) - fileSizes;
forall (i, off, len) in zip(fileSizes.domain, offsets, fileSizes) {
...
```
I don't think this would have any performance difference, but this is more similar to how we calculate offsets in other places. I normally prefer looping variables over indexing when possible because it makes easier for me to tell at a glance what's local, but that doesn't apply here. So there's def no need to change, but verification that I'm thinking about this correctly would be helpful!

pierce314159 (approved):
A whole lot of comments from me all just to say... LGTM lol 

Nice work ben! This looks great!! 🚀

src/ParquetMsg.chpl
pierce314159: Since `leftOver` the remaineder of `locDom.size/loc.maxTaskPar`, it should be guaranteed to be less than `fileSizes.size`. So if we wanted to distribute the remained more evenly we could do something like
```chapel
var fileSizes: [0..#6] int = 10/6;
writeln(fileSizes);
var leftOver = 10 - ((10/6)*6); 
writeln(leftOver);
fileSizes[0..#leftOver] += 1;
writeln(fileSizes);
```
```
1 1 1 1 1 1
4
2 2 2 2 1 1
```
But this is probably overengineering something that isn't actually a problem

pierce314159:
Since @bmcdonald3 is out and none of my comments are blocking, I'll go ahead and merge this. Thanks again ben!!!

bradcray:
Reviewing PRs from our team today, I came across this one and wondered about its status.  My understanding is:
* It only handles ints in its current state, while ultimately one would want to do the same for strings (though maybe that doesn't need to be done in this PR?)
* Ben handed it off to @vasslitvinov when he took his new position, though it may have fallen between the cracks with other things Vass was also handling
* Vass will be unable to work on this for the next few months, so we should consider having someone else pick it up

Tagging @e-kayrakli for awareness and @jhh67 due to his recent Parquet work (albeit in the runtime rather than module-level code).

src/ParquetMsg.chpl
drculhane: There are pathological cases where this could make a big difference.  I lean towards implementing this change in Tess's comment.

PR #2892, opened 2024-01-04
Closes ticket #2831 to align dataframe.groupby().size(), dataframe.groupby().sum() to pandas

ajpotts:
This ticket changes DataFrame.GroupBy.size() to be more similar to the pandas API.   It will now return a dataframe as default, with an as_series optional parameter to allow a series to be returned.  

The groupby count operation was temporarily set to be an alias of size both in dataframe.GroupBy.count() and groupbyclass.GroupBy.count().

In dataframe.py,  def _make_aggop returns an aggregate function, such as dataframe.GroupBy.sum().  The output has been changed to return a dataframe.DataFrame object as default, with an optional parameter as_series to return a Series object instead.  The groupby column names have been added as well.

dataframe.GroupBy also now has additional parameters gb_key_names=None, as_index=True.  The gb_key_names allow storage of the key names of the columns used to group by, and the as_index parameter allows the user to select whether the group by column will be used as the index in the returned dataframe.DataFrame or series.Series.

### DIFF
--- a/PROTO_tests/tests/dataframe_test.py
+++ b/PROTO_tests/tests/dataframe_test.py
@@ -3,13 +3,12 @@ import itertools
 import numpy as np
 import pandas as pd
 import pytest
-from pandas.testing import assert_frame_equal, assert_series_equal, assert_series_equal
+from pandas.testing import assert_frame_equal, assert_series_equal
 
 import arkouda as ak
 
 
 class TestDataFrame:
-
     @staticmethod
     def build_pd_df():
         username = ["Alice", "Bob", "Alice", "Carol", "Bob", "Alice"]
@@ -17,8 +16,8 @@ class TestDataFrame:
         item = [0, 0, 1, 1, 2, 0]
         day = [5, 5, 6, 5, 6, 6]
         amount = [0.5, 0.6, 1.1, 1.2, 4.3, 0.6]
-        bi = [2 ** 200, 2 ** 200 + 1, 2 ** 200 + 2, 2 ** 200 + 3, 2 ** 200 + 4, 2 ** 200 + 5]
-        ui = (np.arange(6).astype(ak.uint64)) + 2 ** 63
+        bi = [2**200, 2**200 + 1, 2**200 + 2, 2**200 + 3, 2**200 + 4, 2**200 + 5]
+        ui = (np.arange(6).astype(ak.uint64)) + 2**63
         return pd.DataFrame(
             {
                 "userName": username,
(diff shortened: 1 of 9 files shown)


arkouda/dataframe.py
```diff
@@ -56,37 +58,186 @@ def groupby_operators(cls):
...
+
+    """
 
-    def __init__(self, gb, df):
+    def __init__(self, gb, df, gb_key_names=None, as_index=True):
         self.gb = gb
         self.df = df
-        for attr in ["nkeys", "size", "permutation", "unique_keys", "segments"]:
+        self.gb_key_names = gb_key_names
+        self.as_index = as_index
```
pierce314159, line 93: we talked about adding `skipna` to the init. I think we might be able to just add it as a class attribute here and then pass it into `self.gb.aggregate` in `aggop`. I'm up for this to be in this PR if it's easy or looked into in a separate issue, whatever u prefer!

arkouda/dataframe.py
```diff
@@ -56,37 +58,186 @@ def groupby_operators(cls):
...
                     )
+                elif isinstance(self.gb_key_names, list) and len(self.gb_key_names) == 1:
+                    return DataFrame(
+                        {c: self.gb.aggregate(self.df.data[c], opname)[1] for c in colnames},
+                        index=Index(self.gb.unique_keys, name=self.gb_key_names[0]),
+                    )
+                elif isinstance(self.gb_key_names, list):
+                    column_dict = {
+                        self.gb_key_names[i]: self.unique_keys[i] for i in range(len(self.gb_key_names))
+                    }
```
pierce314159, line 148: purely stylistic suggestion, so no need to change if u dont want to. I think you can accomplish the same thing with
```python
column_dict = dict(zip(self.gb_key_names, self.unique_keys))
```
which is a bit more pythonic

arkouda/dataframe.py
```diff
@@ -56,37 +58,186 @@ def groupby_operators(cls):
...
+    def __return_agg_series(self, values):
+        if self.as_index:
+            if isinstance(self.gb_key_names, str):
+                return Series(values, index=Index(self.gb.unique_keys, name=self.gb_key_names))
+            elif len(self.gb_key_names) == 1:
+                return Series(values, index=Index(self.gb.unique_keys, name=self.gb_key_names[0]))
+            else:
+                return Series(values)
+        else:
+            return Series(values)
```
pierce314159, line 213: i think this could be simplified a bit
```python
        if self.as_index:
            if isinstance(self.gb_key_names, str):
                return Series(values, index=Index(self.gb.unique_keys, name=self.gb_key_names))
            elif len(self.gb_key_names) == 1:
                return Series(values, index=Index(self.gb.unique_keys, name=self.gb_key_names[0]))
        return Series(values)
```

pierce314159 (commented):
A few stylistic suggestions, but looks good overall! Well done!!

The only thing holding up an approval is a typo in the proto test

PROTO_tests/tests/dataframe_test.py
```diff
@@ -3,21 +3,22 @@
 import numpy as np
 import pandas as pd
 import pytest
-from pandas.testing import assert_frame_equal, assert_series_equal
+from pandas.testing import assert_frame_equal, assert_series_equal, assert_series_equal
```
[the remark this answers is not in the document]
ajpotts: This has already been resolved.  Thanks!

arkouda/dataframe.py
```diff
@@ -56,37 +58,186 @@ def groupby_operators(cls):
...
+    """
+    A DataFrame that has been grouped by a subset of columns
+
+    Parameters
+    ----------
+
+    gb_key_names : str or list(str), default=None
+        The column name(s) associated with the aggregated columns.
+
+    as_index : as_index, default=True
```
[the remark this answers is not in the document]
ajpotts: Good catch.  Thanks!

PROTO_tests/tests/series_test.py
```diff
@@ -187,4 +187,6 @@ def test_index_as_index_compat(self):
         # added to validate functionality for issue #1506
         df = ak.DataFrame({"a": ak.arange(10), "b": ak.arange(10), "c": ak.arange(10)})
         g = df.groupby(["a", "b"])
-        g.broadcast(g.sum("c"))
+        series =
```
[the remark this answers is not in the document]
ajpotts: That was an extra line that was not needed and must have been pasted in by mistake.  Sorry about that :)

arkouda/dataframe.py
ajpotts: This is better.  Thanks for the tip.  I made the change.

arkouda/dataframe.py
ajpotts: That is cleaner.  I made the change.  Thanks!

jaketrookman (approved):
Looks good to me.
For future reference, let the person who requested the change resolve the request to ensure that their issue was properly addressed.

pierce314159 (approved):
Looks good!

PR #2904, opened 2024-01-17
Closes #2894 add dropna option to dataframe groupby

### LINKED ISSUE #2894: add dropna option to dataframe.GroupBy
dataframe.GroupBy should have an option dropna, with default dropna=True to match pandas api:

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html

ajpotts:
This change adds a dropna option to dataframe.DataFrame.groupby().
It also adds a dataframe.DataFrame.sort_index function.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -2155,7 +2155,7 @@ class DataFrame(UserDict):
     def __reindex(self, idx):
         if isinstance(self.index, MultiIndex):
             new_index = MultiIndex(self.index[idx].values, name=self.index.name, names=self.index.names)
-        elif isinstance(self.index, Index) and hasattr(self.index, "name"):
+        elif isinstance(self.index, Index):
             new_index = Index(self.index[idx], name=self.index.name)
         else:
             new_index = Index(self.index[idx])
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -22,7 +22,7 @@ from arkouda.dtypes import uint64 as akuint64
 from arkouda.groupbyclass import GROUPBY_REDUCTION_TYPES
 from arkouda.groupbyclass import GroupBy as akGroupBy
 from arkouda.groupbyclass import unique
-from arkouda.index import Index
+from arkouda.index import Index, MultiIndex
 from arkouda.join import inner_join
 from arkouda.numeric import cast as akcast
 from arkouda.numeric import cumsum, where
@@ -72,6 +72,7 @@ class GroupBy:
         (only implemented for single dimensional aggregates).
         Otherwise, treat aggregated column as a dataframe column.
 
+
     Attributes
     ----------
     gb :    arkouda.groupbyclass.GroupBy
@@ -83,7 +84,10 @@ class GroupBy:
         The column name(s) associated with the aggregated columns.
     as_index : bool (default=True)
         If True the grouped values of the aggregation keys will be treated as an index.
-
+    dropna : bool (default=True)
+        If True, and the groupby keys contain NaN values,
+        the NaN values together with the corresponding row will be dropped.
+        Otherwise, the rows corresponding to NaN values will be kept.
     """
 
     def __init__(self, gb, df, gb_key_names=None, as_index=True):
@@ -91,7 +95,6 @@ class GroupBy:
         self.df = df
         self.gb_key_names = gb_key_names
         self.as_index = as_index
-
         for attr in ["nkeys", "permutation", "unique_keys", "segments"]:
             setattr(self, attr, getattr(gb, attr))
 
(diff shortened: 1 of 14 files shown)


arkouda/groupbyclass.py
```diff
@@ -252,8 +256,30 @@ def __init__(
...
         assume_sorted: bool = False,
+        dropna: bool = True,
         **kwargs,
     ):
+        from arkouda.numeric import isnan
+
+        def drop_na_keys():
+            if self.dropna is True:
+                # if self.keys is None:
+                #     pass
```
pierce314159, line 267: leftover commented out code

arkouda/series.py
```diff
@@ -337,7 +368,7 @@ def sort_values(self, ascending: bool = True) -> Series:
                 idx = argsort(self.values)[arange(self.values.size - 1, -1, -1)]
         else:
             idx = argsort(self.values)
-        return Series(index=self.index[idx], data=self.values[idx])
+        return self.__reindex(idx)  # Series(index=self.index[idx], data=self.values[idx])
```
pierce314159, line 371: leftover commented out code

arkouda/dataframe.py
```diff
@@ -2111,6 +2152,33 @@ def coargsort(self, keys, ascending=True):
             i = i[arange(self.size - 1, -1, -1)]
         return i
 
+    def __reindex(self, idx):
+        if isinstance(self.index, MultiIndex):
+            new_index = MultiIndex(self.index[idx].values, name=self.index.name, names=self.index.names)
+        elif isinstance(self.index, Index) and hasattr(self.index, "name"):
+            new_index = Index(self.index[idx], name=self.index.name)
+        else:
+            new_index = Index(self.index[idx])
```
pierce314159, line 2161: I don't think the `hasattr(self.index, "name")` is doing anything here since an `Index` will always have a name attribute even if it is set to `None`

arkouda/dataframe.py
```diff
@@ -83,15 +84,17 @@ class GroupBy:
         The column name(s) associated with the aggregated columns.
     as_index : bool (default=True)
         If True the grouped values of the aggregation keys will be treated as an index.
-
+    dropna : bool (default=True)
+        If True, and the groupby keys contain NaN values,
+        the NaN values together with the corresponding row will be dropped.
+        Otherwise, the rows corresponding to NaN values will be kept.
```
pierce314159, line 90: I'm not sure if this should be listed as an attribute since we never call anything like `self.dropna`

arkouda/series.py
```diff
@@ -306,29 +314,52 @@ def topn(self, n: int = 10) -> Series:
 
         return Series(index=k.index[idx], data=v[idx])
 
+    def __reindex(self, idx):
+        if isinstance(self.index, MultiIndex):
+            new_index = MultiIndex(self.index[idx].values, name=self.index.name, names=self.index.names)
+        elif isinstance(self.index, Index) and hasattr(self.index, "name"):
+            new_index = Index(self.index[idx], name=self.index.name)
+        else:
+            new_index = Index(self.index[idx])
```
pierce314159, line 323: same comment about `hasattr(self.index, "name")`

pierce314159 (commented):
This looks great! Nice work! Just a couple of small comments

arkouda/groupbyclass.py
ajpotts: Deleted.  Thanks for catching that.

arkouda/series.py
ajpotts: Deleted.

arkouda/dataframe.py
```diff
@@ -2111,6 +2152,33 @@ def coargsort(self, keys, ascending=True):
             i = i[arange(self.size - 1, -1, -1)]
         return i
 
+    def __reindex(self, idx):
```
jaketrookman, line 2155: Does this need to have a double underscore or just one? I know there are a few reserved functions that are reserved for it.

pierce314159: I think we tend to use only one underscore for our own hidden functions to more easily differentiate between them and the [dunder functions](https://www.geeksforgeeks.org/dunder-magic-methods-python/) (double underscore). This is ultimately a stylistic choice but for the sake of a unified style, I'd be in favor of changing this to follow our convention of a single leader underscore

arkouda/dataframe.py
ajpotts: I fixed this.  Thanks!

arkouda/dataframe.py
ajpotts: OK, I changed it.  Thanks!

pierce314159 (approved):

PR #2910, opened 2024-01-19
Closes #2905 add strings isalnum

### LINKED ISSUE #2905: Add `strings.isalnum`
We should add `s.isalnum()` function which returns a boolean array where index `i` is true iff `s[i]` contains only alphanumeric characters. This is based on [`isalnum` in numpy](https://numpy.org/doc/stable/reference/generated/numpy.char.isalnum.html) 

To do this we should use the [chapel `isAlnum` proc](https://chapel-lang.org/docs/language/spec/strings.html#String.string.isAlnum)

This is quite similar to to `is_lower/upper`, but for the first pass let's ignore the `computeOnSegments` optimization

Somewhat unrelated, we should consider renaming `is/to_upper/lower/title` in order to match [numpy version of these functions](https://numpy.org/doc/stable/reference/routines.char.html)

ajpotts:
This ticket adds the Strings.isalnum() function to align with the numpy functionality:
https://numpy.org/doc/stable/reference/generated/numpy.char.isalnum.html

### DIFF
--- a/PROTO_tests/tests/string_test.py
+++ b/PROTO_tests/tests/string_test.py
@@ -494,6 +494,12 @@ class TestString:
         expected = ak.arange(40) >= 30
         assert istitle.to_list() == expected.to_list()
 
+    def test_string_isalnum(self):
+        not_alnum = ak.array([f"%Strings {i}" for i in range(3)])
+        alnum = ak.array([f"Strings{i}" for i in range(3)])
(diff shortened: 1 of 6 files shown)


pierce314159 (approved):
Looks good to me! Great job!!

jaketrookman (approved):
Looks good

PR #2921, opened 2024-01-19
Closes #2912 add isalpha for pdarray

### LINKED ISSUE #2912: add isalpha for pdarray
Add isalpha function for pdarray to mirror the numpy function:
https://numpy.org/doc/stable/reference/generated/numpy.char.isalpha.html

Returns true for each element if all characters in the string are alphabetic and there is at least one character, false otherwise.

Make use of the chapel function is Alpha:
https://chapel-lang.org/docs/language/spec/strings.html?highlight=istitle#String.string.isAlpha

ajpotts:
This adds the isalpha function to arkouda.strings.String.  

It closes Closes #2912 add isalpha for pdarray.

### DIFF
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -501,6 +501,14 @@ module SegmentedString {
       return computeOnSegments(offsets.a, values.a, SegFunction.StringIsAlphaNumeric, bool);
     }
 
+    /*
+      Returns list of bools where index i indicates whether the string i of the SegString is alphabetic
+      :returns: [domain] bool where index i indicates whether the string i of the SegString is alphabetic
+    */
+    proc isalpha() throws {
+      return computeOnSegments(offsets.a, values.a, SegFunction.StringIsAlphabetic, bool);
+    }
+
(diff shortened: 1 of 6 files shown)


src/SegmentedString.chpl
```diff
@@ -1449,6 +1457,13 @@ module SegmentedString {
     return interpretAsString(values, rng, borrow=true).isAlnum();
   }
 
+  /*
+    The SegFunction called by computeOnSegments for isalnum
```
jaketrookman, line 1461: Looks good, I think you only forgot to switch this one to `isalpha`

jaketrookman (changes requested):

src/SegmentedString.chpl
ajpotts: Thanks for pointing that out.  It was correct in my local copy.  Wondering if I pushed the wrong commit somehow?  Anyway, it should be fixed.  I pushed again and added an extra example in the unit test to differentiate better between isalpha and isalphanum.

pierce314159 (approved):
Looks good to me!

jaketrookman (approved):

PR #2930, opened 2024-01-24
Modify fix to performance regression in #2902

jeremiah-corrado:
https://github.com/Bears-R-Us/arkouda/pull/2902 commented out an assignment to the `ndim` field in some `SymEntry` initializers to correct a performance regression. This had the side effect of disabling some multidimensional array support that relied on queries to `ndim` on the client side.

This PR maintains the current performance and re-enables ND array support by setting the `ndim` field in a different way. Namely, the value is not assigned from a param formal, but rather from the value of another field set before `init this`. 

Rerunning the same 2 Locale scan benchmark from #2902 on master and this branch, performance is unaffected:

| before 2902 | master | this PR |
| :--- | :--- | :--- |
| 42.43 | 68.62 | 67.94 |

### DIFF
--- a/src/compat/e-132/ArkoudaSymEntryCompat.chpl
+++ b/src/compat/e-132/ArkoudaSymEntryCompat.chpl
@@ -67,6 +67,7 @@ module ArkoudaSymEntryCompat {
     this.a = try! makeDistArray((...args), etype);
     init this;
     this.shape = tupShapeString(this.tupShape);
+    this.ndim = this.tupShape.size;
   }
 
   /*
@@ -87,5 +88,6 @@ module ArkoudaSymEntryCompat {
     this.max_bits=max_bits;
     init this;
     this.shape = tupShapeString(this.tupShape);
+    this.ndim = this.tupShape.size;
   }
 }
--- a/src/compat/eq-131/ArkoudaSymEntryCompat.chpl
+++ b/src/compat/eq-131/ArkoudaSymEntryCompat.chpl
@@ -61,7 +61,7 @@ module ArkoudaSymEntryCompat {
     this.a = try! makeDistArray((...args), etype);
     this.complete();
     this.shape = tupShapeString(this.tupShape);
-    this.ndim = N;
+    this.ndim = this.tupShape.size;
   }
 
   /*
@@ -82,6 +82,6 @@ module ArkoudaSymEntryCompat {
     this.max_bits=max_bits;
     this.complete();
     this.shape = tupShapeString(this.tupShape);
-    this.ndim = D.rank;
+    this.ndim = this.tupShape.size;
   }
 }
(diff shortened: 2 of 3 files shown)


bmcdonald3 (approved):
Thanks for confirming you could reproduce the old numbers; looks good to me 👍

bradcray:
Do either of you have any insight into why the old assignment was such a performance killer?  We had some guesses in the perf meeting, but none of them seemed plausible (but clearly something is going off the rails...).

jeremiah-corrado:
I haven't been able to come up with anything so far. I'm planning to bring this up in the perf meeting tomorrow to hopefully get some pointers on how to potentially diagnose the issue though.

bmcdonald3:
Jeremiah and I were talking about this offline, but the difference here is that `D.rank` was a param value, whereas the fix uses a non-param value of `this.tupShape.size;`, so switching from `this.ndim = D.rank;` to `this.ndim = this.tupShape.size;` is switching from assigning from a param, to a non-param, so we think that the paramness of `D.rank` was causing that issue.

This seems odd, since I would think a param would only be better, but Jeremiah was speculating that it could be that, because of the param value there, we were stamping out different initializers that would then get dispatched, which could potentially be causing the problem.

So, we aren't certain why the performance hit is happening, but we are pretty certain it's from switching from a non-param to a param assignment.

bradcray:
Thanks for the additional info!

I could imagine the stamping out being problematic, but would think we'd only stamp out one per unique number of dimensions (so probably 1-3 in practice?).  But that's definitely a good guess and maybe we're being overly conservative in terms of stamping things out in other ways?

If the perf hit happens using the C back-end (not just the LLVM back-end), it'd be interesting to capture the generated code for the two versions and see whether there's a blow up in size, which would support that theory.  If there wasn't, I think the thing to do would be to compare the generated code for that initializer between the two versions to see how it differs.

(Also, I agree that, my intuition would also be that the param version would only behave better, which is part of why I remain curious about this.  If a user were to hit this, I wouldn't be able to explain it to them, so wouldn't expect them to figure it out on their own, and that's disconcerting).

bradcray:
@jeremiah-corrado : Copying and rephrasing a comment from https://github.com/Cray/chapel-private/issues/5892, realizing that it's more relevant here:

Rather than using a post-initialization assignment to `.ndim`, could ndim be set in the initialization phase of the `init()` (i.e., before the `init this;` or `this.complete();`) by invoking the superclass initializer (e.g., `super.init(ndim=...);`)—assuming I'm not missing anything that requires `ndim` to be set late like this? 

The modest benefits to doing so would be:
* better style (initialize fields in order from superclass to child class)
* touches fields in cache-friendlier order (not that there are so many that this is a big deal)
* avoid re-assignment of a field (once to initialize, a second time to assign)

I don't expect this would have a significant performance advantage, but the style improvement feels like good motivation on their own.

PR #2931, opened 2024-01-24
Indexing methods for `Series`

brandon-neth:
Implementations for indexing methods on `Series`: setting and getting with `[]`, `iloc`, `loc`, `iat`, and `at`. Testing compares the outputs of these methods with the outputs of the same methods for pandas Series objects. Also compares error behavior.


src/In1d.chpl
```diff
@@ -87,4 +89,25 @@ module In1d
...
+                if keyVal == arr[i] then indexList.pushBack(i);
+            indexLists[keyIdx] = indexList;
+        }
+
+        var finalList : list(int);
+        for l in indexLists do
+            finalList.pushBack(l);
+        
+        return finalList.toArray();
+    }
```
bmcdonald3, line 112: Noting that we talked about the performance here; we don't expect it to perform well, but are focusing on correctness this first pass.

src/In1dMsg.chpl
```diff
@@ -78,6 +79,77 @@ module In1dMsg
...
+        param pn = Reflection.getRoutineName();
+        var repMsg: string; // response message
+
+        var vname = st.nextName();
+
+        var gEnt: borrowed GenSymEntry = getGenericTypedArrayEntry(msgArgs.getValueOf("keys"), st);
+        var gEnt2: borrowed GenSymEntry = getGenericTypedArrayEntry(msgArgs.getValueOf("arr"), st);
+
+        select(gEnt.dtype, gEnt2.dtype) {
+          when (DType.Int64, DType.Int64) {
```
bmcdonald3, line 100: Is it a feature that both of them have to be the same type? In other words, say we had a `uint` array and an `int` array that both contained logically equivalent values, that is not supported?

brandon-neth: I think that would be okay to support; I was basing it off of the set operations, which require equivalent types iirc. Here we're thinking something like `when (Dtype.Int64, Dtype.Int64), (Dtype.Int64, Dtype.uInt64), (Dtype.uInt64, Dtype.Int64), (Dtype.uInt64, Dtype.uInt64)`?

brandon-neth: modification to code above: we have to cast the arrays to a particular datatype; how should we choose which? Or should we do four separate cases

bmcdonald3: > Here we're thinking something like when (Dtype.Int64, Dtype.Int64), (Dtype.Int64, Dtype.uInt64), (Dtype.uInt64, Dtype.Int64), (Dtype.uInt64, Dtype.uInt64)?

Ya, that is what I was thinking.

I don't think this necessarily has to be added in this PR, just something I was asking about, I think it's fine to limit for now.

I guess I'm not sure we need casting though, in Chapel, you are able to do something like:
```chpl
var a: int = 10;
var b: uint = 10;

writeln(a == b);
```
and that works just fine, so I would assume we can do the same kind of thing here, but is that not the case?

If we have to cast for some reason, that would be unfortunate and I don't think we would want to add that unless someone has a desire for it explicitly.

brandon-neth: If we have all four options handle in a single `when`, then we have to pick types for the entries when we run `var e = toSymEntry(gEnt,/* int or uint */);`. 

If we do four separate cases where the arrays are cast to the correct types, that would work fine, but would require some changes to the `indexof1d` computation implementation, as it expects the same type. Not a problem, but I think we should have compatibility between int64 and uint64 done more systematically than have it just supported for this one method. So maybe an improvement for another sprint.

bmcdonald3: OK, sounds good to me.

bmcdonald3 (approved):

PR #2933, opened 2024-01-26
Closes #2918 add isempty for pdarray

### LINKED ISSUE #2918: add isempty for pdarray
add isempty function for pdarray:

For each entry:
    true – when the string is empty
    false – otherwise

Make use of chapel function : 
https://chapel-lang.org/docs/language/spec/strings.html?highlight=istitle#String.string.isEmpty

ajpotts:
Adds function strings.Strings.isempty().

Closes #2918


jaketrookman (approved):
Looks good

pierce314159 (approved):

PR #2936, opened 2024-01-30
Array API manipulation functions

jeremiah-corrado:
Implementation of the [Array API manipulation functions](https://data-apis.org/array-api/latest/API_specification/manipulation_functions.html)

Adds implementations of:
* `concat`
* `expand_dims`
* `flip`
* `permute_dims`
* `reshape`
* `roll`
* `squeeze`
* `stack`

Like broadcasting, `reshape` and `squeeze` both create arrays whose rank can/will be different than the array being operated on. As such, these procedures also get compiled for various pairs of supported ranks, which required some modifications to `parseServerConfig.py`

Tests are added for each of the new functions.

Adds a separate Array API module config file with modules that are only needed for the ArrayAPI.


arkouda/array_api/_manipulation_functions.py
```diff
@@ -14,26 +14,67 @@ def broadcast_arrays(*arrays: Array) -> List[Array]:
 
     See its docstring for more information.
     """
-    raise ValueError("broadcast_arrays not implemented")
+
+    shapes = [a.shape for a in arrays]
+    bcShape = shapes[0]
+    for shape in shapes[1:]:
```
bmcdonald3, line 20: Do we need a check here to see if `shapes` is of length 1 or is `[1:]` just `None` if it is length 1? Or does this function require more than 1 array?

src/ManipulationMsg.chpl
```diff
@@ -0,0 +1,915 @@
...
+        mLogger.error(getModuleName(),pn,getLineNumber(),errorMsg);
+        return new MsgTuple(errorMsg,MsgType.ERROR);
+      }
+    }
+  }
+
+  private proc flatten(const ref a: [?d] ?t): [] t throws {
+    var flat = makeDistArray({0..<d.size}, t);
+    forall idx in flat.domain with (var agg = newSrcAggregator(t)) do
+      agg.copy(flat[idx], a[d.orderToIndex(idx)]);
```
bmcdonald3, line 288: These copy patterns are hard for me to wrap my head around... 

It seems here we are just iterating through the second array and copying the values over through aggregators into the flat array, but I am just thikning if doing more of a bulk assign where we figure out where the elements go in the flat array based on the dimensions and then do a bulk assign would be better?

e.g.:

```
flat: [ 0 0 0 0 ]
a   : [ 1 2 ; 3 4 ]
```
Since we know `a`s dimensions, we can do something like `flat[0..#flat.length/(a[0]).length)] = a[0]` 

Probably doesn't matter and no resaon to do that premature performance work before we have a reason to, but just curious what led to this pattern.

bmcdonald3 (approved):
Not sure I have a ton of feedback, I think the API answering the qeustions of design have those bases covered and the tests look good, so since we are focused on correctness and not performance off the bat, I think this looks good.

I am curious about a few things I mentioned and one other thing I'll say is the amount of `param for` loops we have is unfortunate, but I understand we need those because of how we are doing the dimensionality, that just really jumped out at me taking a pass over the code.

Looks good 👍

arkouda/array_api/_manipulation_functions.py
jeremiah-corrado: `shapes[1:]` returns an empty list if `shapes` only has a length of 1. So I believe this should essentially just create a copy of the array if it's only called on one array.

> Or does this function require more than 1 array?

And it looks like the [spec](https://data-apis.org/array-api/latest/API_specification/generated/array_api.broadcast_arrays.html#array_api.broadcast_arrays) says "1 or more arrays", so I don't think an extra check would be needed.

PR #2944, opened 2024-02-02
Closes #2906 shift operator for boolean vectors

### LINKED ISSUE #2906: shift operator for boolean vectors
In numpy:
`np_bool = np.array([1, 0, 1, 0, 0, 1, 0], dtype='bool') `
`np_bool << True`
returns:
`array([2, 0, 2, 0, 0, 2, 0], dtype=int8) `

Arkouda should have the same functionality, including left and right shift for both scalars and vectors.

ajpotts:
Adds left and right shift x << y and x >> y in the case where x is a boolean array and y is a boolean array or scalar.

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -1230,15 +1230,11 @@ module BinOp
           when "<<" {
             if(val){
               e.a = val:int << r.a:int;
-            }else{
-              e.a = val:int;
             }
           }
           when ">>" {
             if(val){
               e.a = val:int >> r.a:int;
-            }else{
-              e.a = val:int;
             }
           }
           otherwise {
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -697,10 +697,18 @@ module BinOp
     else if (l.etype == bool && val.type == bool) {
       select op {
         when ">>" {
-          e.a = l.a:int >> val:int;
+          if(val){
+            e.a = l.a:int >> val:int;
+          }else{
+            e.a = l.a:int;
+          }
         }
         when "<<" {
-          e.a = l.a:int << val:int;
+          if(val){
+            e.a = l.a:int << val:int;
+          }else{
+            e.a = l.a:int;
+          }
         }
         otherwise {
           var errorMsg = notImplementedError(pn,l.dtype,op,dtype);
@@ -1220,10 +1228,18 @@ module BinOp
     } else if (r.etype == bool && val.type == bool)  {
       select op {
           when "<<" {
-            e.a = val:int << r.a:int;
+            if(val){
+              e.a = val:int << r.a:int;
+            }else{
+              e.a = val:int;
+            }
           }
           when ">>" {
-            e.a = val:int >> r.a:int;
+            if(val){
+              e.a = val:int >> r.a:int;
+            }else{
+              e.a = val:int;
+            }
           }
           otherwise {
             var errorMsg = notImplementedError(pn,dtype,op,r.dtype);
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -458,6 +458,22 @@ module BinOp
         }
       var repMsg = "created %s".doFormat(st.attrib(rname));
       return new MsgTuple(repMsg, MsgType.NORMAL);
+    } else if (l.etype == bool && r.etype == bool)  {
+      select op {
+          when "<<" {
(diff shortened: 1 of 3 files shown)


src/OperatorMsg.chpl
```diff
@@ -148,6 +148,10 @@ module OperatorMsg
           when (DType.Bool, DType.Bool) {
             var l = toSymEntry(left,bool, nd);
             var r = toSymEntry(right,bool, nd);
+            if (op == "<<") || (op == ">>" )  {
+              var e = st.addEntry(rname, l.tupShape, uint);
+              return doBinOpvv(l, r, e, op, rname, pn, st);
```
pierce314159, line 153: i don't think it matters since we're only shifting bools by bools, but I wanted to point out that we create `e` with `uint` type but we cast them to `int` in `doBinOp`

src/BinOp.chpl
```diff
@@ -1185,6 +1217,22 @@ module BinOp
...
       var repMsg = "created %s".doFormat(st.attrib(rname));
       return new MsgTuple(repMsg, MsgType.NORMAL);
+    } else if (r.etype == bool && val.type == bool)  {
+      select op {
+          when "<<" {
+            e.a = val:int << r.a:int;
+          }
+          when ">>" {
+            e.a = val:int >> r.a:int;
+          }
```
pierce314159, line 1227: It's not necessary but we could wrap these in an `if val` block to skip doing anything when `val` is false. Chapel zeroes out arrays when they are created and false shifted in either direction would remain 0, so we could just leave `e.a` as it is

But that's optimizing for a case that's prob not terribly common and isn't terribly expensive so maybe it would be better to just avoid the clutter

src/BinOp.chpl
```diff
@@ -679,6 +694,23 @@ module BinOp
...
+          e.a = l.a:int >> val:int;
+        }
+        when "<<" {
+          e.a = l.a:int << val:int;
+        }
+        otherwise {
+          var errorMsg = notImplementedError(pn,l.dtype,op,dtype);
+          omLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
+          return new MsgTuple(errorMsg, MsgType.ERROR);
+        }
```
pierce314159, line 709: similar to the other case, we could do something like
```chapel
if val {
  // do shift
}
else {
  e.a = l.a;
}
```
since an array shifted in either direction by `False/0` would just remain the same

pierce314159 (approved):
Looks good to me! I threw out some ideas to skip the shift when the scalar is `False`, but it's not clear to me if it's worth the added clutter. I'm good with whatever you want to do!

src/BinOp.chpl
ajpotts: I made this change.  Thanks for reminding me.

src/BinOp.chpl
ajpotts: Made the change.  Thanks.

src/OperatorMsg.chpl
ajpotts: Thanks for catching this.  I changed the type to int to better match numpy.  Numpy returns int(8), but when I tried to change to int(8) I received this error: 
RuntimeError: IOError Line 97 In MultiTypeSymbolTable.addEntry: The server was not configured to support pdarray's of type int(8). Please update the configuration and recompile.

jaketrookman (approved):
Looks good

src/BinOp.chpl
```diff
@@ -1185,6 +1225,30 @@ module BinOp
...
       var repMsg = "created %s".doFormat(st.attrib(rname));
       return new MsgTuple(repMsg, MsgType.NORMAL);
+    } else if (r.etype == bool && val.type == bool)  {
+      select op {
+          when "<<" {
+            if(val){
+              e.a = val:int << r.a:int;
+            }else{
+              e.a = val:int;
+            }
```
pierce314159, line 1235: In the else block, we know `val` is `False` and since casting `False` to int gives `0`,  the else block is setting `e.a = 0;`. `e.a` was initialized to be all zeroes, so this operation isn't necessary. We can just do 
```chapel 
if val {
  e.a = val:int << r.a:int;
}
```
with no else block. The same goes for the `">>"` case

ajpotts: Ok, I made the change.

pierce314159 (approved):

PR #2946, opened 2024-02-05
Closes #2916 add isspace for pdarrays

### LINKED ISSUE #2916: add isspace for pdarrays
add isspace function to match numpy function for ndarrays:
https://numpy.org/doc/stable/reference/generated/numpy.char.isspace.html

Returns true for each element if there are only whitespace characters in the string and there is at least one character, false otherwise.

May rely on chapel function isSpace:
https://chapel-lang.org/docs/language/spec/strings.html?highlight=istitle#String.string.isSpace

ajpotts:
This ticket adds the isspace function for Strings.

closes #2916


jaketrookman (approved):
Looks good

pierce314159 (approved):

PR #2948, opened 2024-02-05
Closes #2919 add capitalize to pdarrays

### LINKED ISSUE #2919: add capitalize to pdarrays
Add function capitalize to pdarrays to match what is in numpy for ndarrays:
https://numpy.org/doc/stable/reference/generated/numpy.char.capitalize.html#numpy.char.capitalize

Return a copy of a with only the first character of each element capitalized.

ajpotts:
Add capitalize function to Strings.

Closes #2919

### DIFF
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -450,9 +450,9 @@ module SegmentedString {
 
     /*
       Given a SegString, return a new SegString with first character of each original element replaced with its uppercase equivalent
-      and the remaining characters replaced with their lowercase equivalent
+      and the remaining characters replaced with their lowercase equivalent.  The first character following a space character will be uppercase.
       :returns: Strings – Substrings with first characters replaced with uppercase equivalent and remaining characters replaced with
-      their lowercase equivalent
+      their lowercase equivalent.  The first character following a space character will be uppercase.
     */
     proc title() throws {
       ref origVals = this.values.a;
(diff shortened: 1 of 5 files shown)


src/SegmentedString.chpl
```diff
@@ -469,6 +469,32 @@ module SegmentedString {
...
+        for b in interpretAsBytes(origVals, off..#len, borrow=true).toLower() {
+          valAgg.copy(capitalizedVals[off+i], b:uint(8));
+          i += 1;
+        }
+        i = 0;
+        for b in interpretAsString(origVals, off..#len, borrow=true).item(0).toUpper().bytes(){
+          valAgg.copy(capitalizedVals[off+i], b:uint(8));
+          i += 1;
+        }
+      }
```
pierce314159, line 494: I wanted to get rid of having to call `toUpper` on the entire string when we only need the first character to be upper. I did something that ended up being pretty convoluted but seemed to work before I realized that we have the `uint(8)` value of the lowercase character, so we can just shift it to get the upper case value

```chapel
forall (off, len) in zip(offs, lengths) with (var valAgg = newDstAggregator(uint(8))) {
  var i = 0;
  for b in interpretAsBytes(origVals, off..#len, borrow=true).toLower() {
    if i == 0 {
      valAgg.copy(capitalizedVals[off+i], b - 32);
    }
    else {
      valAgg.copy(capitalizedVals[off+i], b:uint(8));
    }
    i += 1;
  }
}
```
The offset of 32 can be seen here: https://en.wikipedia.org/wiki/ASCII#Printable_characters

pierce314159 (commented):
Looks good! One suggestion to avoid the `toUpper` call

src/SegmentedString.chpl
ajpotts: .item(0) does pull off the first character, from what I'm understanding from the chapel docs: https://chapel-lang.org/docs/language/spec/strings.html?highlight=encode#String.string.item  



Returns

    A new [string](https://chapel-lang.org/docs/language/spec/strings.html?highlight=encode#String.string) with the complete multibyte character starting at the specified codepoint index from 1..string.numCodepoints


The wording seems to imply that characters can be more than one byte, which is why I was avoiding just pulling off the first byte.

pierce314159: Sorry you're def right that this doesn't call `toUpper` on the whole string, but it does duplicate the `interpretAsString` call which is not ideal since that is the expensive bit (it could require creating a new buffer or fetching data if it's not all local or split across locales). But it seems unavoidable because of the multi-byte case like u said

I didn't consider the multi-byte case, but you're absolutely right! I ran a quick test of your code and numpys capitalize with some strings where the first character is a multi-byte, and your code works correctly 
```python
>>> np.char.capitalize(np.array(["äaaa", "åaaa", "âaaa"]))
array(['Äaaa', 'Åaaa', 'Âaaa'], dtype='<U4')

>>> ak.array(["äaaa", "åaaa", "âaaa"]).capitalize()
array(['Äaaa', 'Åaaa', 'Âaaa'])
```

A modification that only selected the first byte (something like this)
```chapel
# we would need to do a check to pick the smaller of 2 and len, but this is just a proof of concept
for b in interpretAsBytes(origVals, off..#2, borrow=true).toUpper(){
  valAgg.copy(capitalizedVals[off], b:uint(8));
}
```
Capitalizes characters with single byte, but fails on multi-bytes
```python
>>> ak.array(["äaaa", "åaaa", "âaaa"]).capitalize()
array(['äaaa', 'åaaa', 'âaaa'])
```

So i think we should leave this as is, great job amanda!

pierce314159 (approved):

jaketrookman (approved):

pierce314159 (approved):

PR #2950, opened 2024-02-05
Closes #2949: Add compat modules for 1.34

### LINKED ISSUE #2949: Fix 1.33 build issues after sort changes
Since the sort changes were added, Arkouda was failing to build with 1.33, as 1.33 and 1.34 were using the same compat modules. Splitting 1.34 off into it's own directory so that we can keep those changes separate.

bmcdonald3:
Since the sort changes were added, Arkouda was failing to build with 1.33, as 1.33 and 1.34 were using the same compat modules. Splitting 1.34 off into it's own directory so that we can keep those changes separate.

Closes #2949


pierce314159 (approved):
looks good!

PR #2962, opened 2024-02-09
Closes #2896 DataFrame columns should return an Index

### LINKED ISSUE #2896: DataFrame.columns should return an Index
DataFrame.columns returns a list, but should return an Index to better align with Pandas:
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.columns.html

The following example demonstrates the discrepancy:

`data = {"key1":["valuew","valuex","valuew","valuex"],"key2":["valueA","valueB","valueA","valueB"],"key3":["value1","value2","value3","value4"],"count":[34,25,11,4],"nums":[1,2,5,21]}`

`pd_df = pd.DataFrame(data)`
`ak_df = ak.DataFrame({k:ak.array(v) for k,v in data.items()})`

`print(pd_df.columns)`
Index(['key1', 'key2', 'key3', 'count', 'nums'], dtype='object')

`print(ak_df.columns)`
['key1', 'key2', 'key3', 'count', 'nums']

ajpotts:
Closes #2896 

This changes the return type of DataFrame.columns to ak.index.Index.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -624,7 +624,7 @@ class DataFrame(UserDict):
             self._nrows = initialdata._nrows
             self._bytes = initialdata._bytes
             self._empty = initialdata._empty
-            self._columns = initialdata._columns
+            self._column_names = initialdata._column_names
             if index is None:
                 self._set_index(initialdata.index)
             else:
@@ -637,7 +637,7 @@ class DataFrame(UserDict):
             self._nrows = initialdata.shape[0]
             self._bytes = 0
             self._empty = initialdata.empty
-            self._columns = initialdata.columns.tolist()
+            self._column_names = initialdata.columns.values
 
             if index is None:
                 self._set_index(initialdata.index.values.tolist())
@@ -660,7 +660,7 @@ class DataFrame(UserDict):
         self._empty = True
 
         # Initial attempts to keep an order on the columns
-        self._columns = []
+        self._column_names = []
         self._set_index(index)
 
         # Add data to the DataFrame if there is any
@@ -685,7 +685,7 @@ class DataFrame(UserDict):
                     self._empty = False
                     UserDict.__setitem__(self, key, val)
                     # Update the column index
-                    self._columns.append(key)
+                    self._column_names.append(key)
 
             # Initial data is a list of arkouda arrays
             elif isinstance(initialdata, list):
@@ -711,7 +711,7 @@ class DataFrame(UserDict):
                     self._empty = False
                     UserDict.__setitem__(self, key, col)
                     # Update the column index
-                    self._columns.append(key)
+                    self._column_names.append(key)
 
             # Initial data is invalid.
             else:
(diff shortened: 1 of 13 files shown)


pierce314159 (commented):
I'm not quite sure what's going on, but I pulled down this PR to try it out. When I attempt to tab complete to see the attributes of the dataframe, it doesn't provide the drop down and server seems to indicate that new pdarrays are being created. So this is something we should dig into next week. Here is the code I was running for reference
```python
In [3]: key = ak.array([0, 0, 1, 1, 2, 2, 3, 3])
   ...: val = ak.array([0, 9, 0, 0, 0, 0, 0, 0])
   ...: df = ak.DataFrame({"key": key, "val": val})

In [4]: df.
```

ajpotts:
I made the changes to fix the ipython side effect and it probably needs to be re-reviewed because I changed how I did it.  

I basically changed the name of the current "columns" attribute to "column_names" and then created a new property "columns" which converts "column_names" to an index.

arkouda/dataframe.py
```diff
@@ -730,23 +730,23 @@ def __init__(self, initialdata=None, index=None, columns=None):
             self.update_nrows()
 
     def __getattr__(self, key):
-        if key not in self.columns:
+        if key not in self.column_names:
```
pierce314159, line 733: should this be `self._column_names`

arkouda/dataframe.py
```diff
@@ -1525,12 +1525,43 @@ def shape(self):
...
+        |  1 |      2 |      4 |
+        +----+--------+--------+
+
+        >>> df.columns
+        Index(array(['col1', 'col2']), dtype='<U0')
+        """
+        return Index(self._column_names)
+
+    @property
+    def column_names(self):
```
pierce314159, line 1564: is `column_names` a property of a pandas dataframe?

pierce314159 (commented):
maybe I'm wrong but I feel like we can just have a `_columns` as a hidden attribute which is a list of strings and then have the property function for `columns` convert it to an index before sending it to the user

Because right now we have a `df.column_names` property visible to the user which pandas doesn't have, which isn't a huge deal but I don't think we need it

arkouda/dataframe.py
ajpotts: We could do it that way, and I think that's how I had it before.  If we do it that way, then every time you loop over the column names you would have to either call df._columns (hidden attribute) or df.columns.values, which creates an index, sends the data to arkouda server, only immediately to pull all the values back, which seemed kind of inefficient.  The column names are small now, but keep in mind, could be more voluminous once we start implementing features like on-hot-encoding.  But I'm open to whatever everyone can agree with.

arkouda/dataframe.py
ajpotts: No, it would be a new feature.

arkouda/dataframe.py
ajpotts: Actually, you would have to call df.columns.values.tolist() to loop over the columns b/c in arkouda indexes store the value as pdarrays rather than lists.

ajpotts: I made the changes we discussed.  Let me know what you think.

arkouda/util.py
```diff
@@ -288,45 +292,6 @@ def attach_all(names: list):
...
-    Helper for summing two sparse matrices together
-
-    Parameters
-    -----------
-    idx1: pdarray
-        indices for the first sparse matrix
-    idx2: pdarray
-        indices for the second sparse matrix
-    val1: pdarray
-        values for the first sparse matrix
```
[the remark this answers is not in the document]
pierce314159: If i had to guess i'd say this was due to a rebase gone awry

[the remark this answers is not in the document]
ajpotts: It should be fixed now.  Thanks for catching that.

pierce314159 (approved):

jaketrookman (approved):
Looks good

PR #2967, opened 2024-02-12
Array API stats functions

jeremiah-corrado:
Implementation of the [Array API Statistical Functions](https://data-apis.org/array-api/latest/API_specification/statistical_functions.html):

* max
* min
* mean
* std
* var
* prod (reduce)
* sum (reduce)

And the two [searching functions](https://data-apis.org/array-api/latest/API_specification/searching_functions.html) that were already implemented for 1D arrays:
* argmin
* argmax

Also adds multidimensional indexing assignments:
```python
a[2, :, 8] = 3.14
```


arkouda/pdarrayclass.py
```diff
@@ -725,64 +724,122 @@ def __getitem__(self, key):
...
+                            "array": self,
+                            "start": start,
+                            "stop": stop,
+                            "stride": stride,
+                            "value": value,
+                        },
+                    )
+                else:
+                    generic_msg(
+                        cmd="[slice]=val-1D",
```
bmcdonald3, line 776: I am a little confused by the use of the `1D`s here. Why is it only on some of them and why is it hardcoded to 1D rather than doing the f string based on number of dims?

src/ReductionMsg.chpl
```diff
@@ -32,253 +33,347 @@ module ReductionMsg
...
+          select op {
+            when "argmin" {
+              const (minVal, minLoc) = minloc reduce zip(eIn.a, eIn.a.domain);
+              s = minLoc;
             }
+            when "argmax" {
+              const (maxVal, maxLoc) = maxloc reduce zip(eIn.a, eIn.a.domain);
+              s = maxLoc;
+            }
+            otherwise halt("unreachable");
```
bmcdonald3, line 277: We usuaully return a msgtuple with the error type to avoid shutting down the whole server on this type of thing (we never want the server to crash or halt because that means all the data is lost).

bmcdonald3 (approved):
Pretty much all my feedback is maybe replace those `otherwise halt` paths with `return new MsgTuple(errorMsg, MsgType.ERROR);`, rest looks good, nice work.

arkouda/pdarrayclass.py
jeremiah-corrado: The `1D` is only on the `[int]=val` and `[slice]=val` commands because I didn't update the other two to support ND arrays yet — that will probably come in a future PR because I don't think it's needed for the array API. It's hard-coded to one here because this code branch is only for 1D arrays. There is more code further down that uses the array's rank in the command.

src/ReductionMsg.chpl
jeremiah-corrado: There is a check up above (line 257) that will return an error if the operation isn't recognized so the halt shouldn't be reachable (I had to put it there to make the compiler happy).

I think it would be clearer to return the error message from the select statement instead of checking beforehand, but returning from within a `forall` isn't allowed, so that's ruled out for the second select statement in this function. Do you think it would be better to refactor the code for the first select statement at least and leave the second as-is, or is it okay?

bmcdonald3: OK, that makes sense to me, sousnds good as-is.

PR #2987, opened 2024-02-21
Remove deprecation messages for reader/writer locking default change

jeremiah-corrado:
In Chapel's next release, warnings will be emitted about the default value of `locking` changing in the following IO procedures:
* `file.reader`
* `file.writer`
* `openReader`
* `openWriter`

This PR makes some modifications to silence those warnings. In some cases, `locking` is specified explicitly. In other cases, uses of `reader`/`writer` are switched to `openReader`/`openWriter` where a flag is used to silence the warning.


src/CSVMsg.chpl
```diff
@@ -66,17 +66,14 @@ module CSVMsg {
...
+        var reader = openReader(filename);
         var lines = reader.lines().strip();
         if lines[0] == CSV_HEADER_OPEN {
             idx = 3; // set to first line after header
         }
 
         var col_delim: string = msgArgs.getValueOf("col_delim");
         var column_names = lines[idx].split(col_delim);
-        reader.close();
-        csvFile.close();
```
pierce314159, line 79: we removed the reader close and file close here but in another section we only remove the file close

src/CSVMsg.chpl
```diff
@@ -300,7 +296,6 @@ module CSVMsg {
         var row_ct: int = lines.size - data_start_offset;
 
         reader.close();
-        csvFile.close();
```
pierce314159, line 303: here we keep the `reader.close()`, I was just wanted to point out the fact that they're different in case it matters

pierce314159 (commented):
I just have one question about inconsistencies in closing the reader in case that makes a difference. But outside of that looks good to me!

jeremiah-corrado:
In both cases, calling `close` isn't strictly necessary, as the fileReader's `deinit` will call it when the variable falls out of scope. I think I left the `close` in the first case though because there is quite a bit more code in that scope after the fileReader is done being used.

pierce314159 (approved):

Issue #2991, opened 2024-02-22
`chpldoc` failure


pierce314159:
I'm seeing the a[ doc build failure](https://github.com/Bears-R-Us/arkouda/actions/runs/7980387267/job/21789909998) for our CI on master. It seems to have been happening ever since [we updated the CI to run 1.33](https://github.com/Bears-R-Us/arkouda/commits/master/)

Looking at the error message:
```
Warning, treated as error:
/tmp/chpldoc-root.deleteme-6811vu/source/modules/__w/arkouda/arkouda/src/NumPyDType.rst:100:Inline substitution_reference start-string without end-string.
error: building html output from chpldoc sphinx project
make: *** [Makefile:430: /__w/arkouda/arkouda//docs/server/index.html] Error 1
Error: Process completed with exit code 2.
```

It looks like it's failing on the chapel docs in NumPyDType. I'm not sure it's related but @jeremiah-corrado last updated the docs in that file, so maybe he could take a look at it? If I had to venture a guess, I would thing maybe this block is causing trouble?
```chapel
    /*
      Return the dtype that can store the result of
      an operation between two dtypes for the following
      operations: +, -, *, **, //, %, &, |, ^, <<, >>,

      follows Numpy's rules for type promotion
      (of which the array-api promotion rules are a subset)
    */
```
since it's using `*`, but I wouldn't expect that to cause issues

PR #2992, opened 2024-02-22
Resolve formatting issue in NumpyDType server docs

jeremiah-corrado:
Add backticks around symbols in NumpyDType docs to prevent them from being parsed as rst special characters.  With this change, chpldoc was running successfully on my machine.

resolves: #2991


pierce314159 (approved):
Thanks!!

PR #3014, opened 2024-03-01
Support for creating Array API objects from numpy arrays

jeremiah-corrado:
Add support for creating array-api Arrays from numpy arrays.  This also puts the server infrastructure in place needed to create multidimensional pdarrays from numpy arrays (however, no updates are made to the frontend to support this yet). 

Because the `arrayMsg` handler has a special signature, a separate command map is added to store the `array1D` -> `arrayND` commands. The server config script is modified to populate that map for procs with the `@arkouda.registerArrayMsg` annotation.


bmcdonald3 (approved):
Looks good to me

PR #3015, opened 2024-03-01
Closes #2913: add isdecimal

### LINKED ISSUE #2913: add isdecimal for pdarray
Add isdecimal function to pdarray to match what numpy has for ndarray:
https://numpy.org/doc/stable/reference/generated/numpy.char.isdecimal.html

For each element, return True if there are only decimal characters in the element.

There does not appear to be a corresponding chapel function yet.

jaketrookman:
This PR (Closes #2913) adds the Strings.isdecimal() function to align with the numpy functionality

### DIFF
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -618,6 +618,14 @@ class Strings:
         array(['Strings 0', 'Strings 1', 'Strings 2', '120', '121', '122'])
         >>> strings.isdecimal()
         array([False False False True True True])
+
+        Special Character Examples
+        --------------------------
+        >>> special_strings = ak.array(["3.14", "\u0030", "\u00B2", "2³₇", "2³x₇"])
+        >>> special_strings
+        array(['3.14', '0', '²', '2³₇', '2³x₇'])
+        >>> special_strings.isdecimal()
+        array([False True False False False])
         """
         return create_pdarray(
             generic_msg(
@@ -881,6 +889,14 @@ class Strings:
         array(['Strings 0', 'Strings 1', 'Strings 2', '120', '121', '122'])
         >>> strings.isdigit()
         array([False False False True True True])
+        
+        Special Character Examples
+        --------------------------
+        >>> special_strings = ak.array(["3.14", "\u0030", "\u00B2", "2³₇", "2³x₇"])
+        >>> special_strings
+        array(['3.14', '0', '²', '2³₇', '2³x₇'])
+        >>> special_strings.isdigit()
+        array([False True True True False])
         """
         return create_pdarray(
             generic_msg(
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -618,9 +618,7 @@ class Strings:
         array(['Strings 0', 'Strings 1', 'Strings 2', '120', '121', '122'])
         >>> strings.isdecimal()
         array([False False False True True True])
-
         Special Character Examples
-        --------------------------
         >>> special_strings = ak.array(["3.14", "\u0030", "\u00B2", "2³₇", "2³x₇"])
         >>> special_strings
         array(['3.14', '0', '²', '2³₇', '2³x₇'])
@@ -889,9 +887,7 @@ class Strings:
         array(['Strings 0', 'Strings 1', 'Strings 2', '120', '121', '122'])
         >>> strings.isdigit()
         array([False False False True True True])
-        
         Special Character Examples
-        --------------------------
         >>> special_strings = ak.array(["3.14", "\u0030", "\u00B2", "2³₇", "2³x₇"])
         >>> special_strings
         array(['3.14', '0', '²', '2³₇', '2³x₇'])
--- a/arkouda/strings.py
+++ b/arkouda/strings.py
@@ -590,6 +590,43 @@ class Strings:
         return Strings.from_return_msg(cast(str, rep_msg))
 
     @typechecked
+    def isdecimal(self) -> pdarray:
+        """
+        Returns a boolean pdarray where index i indicates whether string i of the
+        Strings has all decimal characters.
+
+        Returns
+        -------
+        pdarray, bool
+            True for elements that are decimals, False otherwise
+
+        Raises
+        ------
+        RuntimeError
+            Raised if there is a server-side error thrown
+
+        See Also
+        --------
+        Strings.isdigit
+
+        Examples
+        --------
+        >>> not_decimal = ak.array([f'Strings {i}' for i in range(3)])
+        >>> decimal = ak.array([f'12{i}' for i in range(3)])
+        >>> strings = ak.concatenate([not_decimal, decimal])
+        >>> strings
+        array(['Strings 0', 'Strings 1', 'Strings 2', '120', '121', '122'])
+        >>> strings.isdecimal()
+        array([False False False True True True])
+        """
+        return create_pdarray(
+            generic_msg(
+                cmd="checkChars",
+                args={"subcmd": "isDecimal", "objType": self.objType, "obj": self.entry},
+            )
+        )
+
+    @typechecked
     def capitalize(self) -> Strings:
         """
         Returns a new Strings from the original replaced with the first letter capitilzed
--- a/src/SegmentedString.chpl
+++ b/src/SegmentedString.chpl
@@ -474,7 +474,7 @@ module SegmentedString {
       :returns: [domain] bool where index i indicates whether the string i of the SegString is a decimal
     */
     proc isDecimal() throws {
-      return computeOnSegments(offsets.a, values.a, SegFunction.StringIsDigit, bool);
+      return computeOnSegments(offsets.a, values.a, SegFunction.StringIsDecimal, bool);
     }
     
     /*
@@ -1493,10 +1493,6 @@ module SegmentedString {
     return myRegex.search(interpretAsString(values, rng, borrow=true)).matched;
   }
 
-  inline proc stringIsDecimal(ref values, rng) throws {
-    return interpretAsString(values, rng, borrow=true).isDecmial();
-  }
-
   /*
     The SegFunction called by computeOnSegments for isLower
   */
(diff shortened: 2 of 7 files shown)


pierce314159 (changes requested):
It looks like the output from one of the examples in the tests doesn't match `np.char.isdigit`

```python
>>>          example2 = ak.array(
   ...:             [
   ...:                 "",
   ...:                 "string1",
   ...:                 "stringA",
   ...:                 "String",
   ...:                 "12345",
   ...:                 "Hello\tWorld",
   ...:                 " ",
   ...:                 "\n",
   ...:                 "3.14",
   ...:                 "\u0030",  # Unicode for zero
   ...:                 "\u00B2",
   ...:             ]
   ...:         )

>>> example2.isdigit()
array([False False False False True False False False False True False])

>>> np.char.isdigit(example2.to_ndarray())
array([False, False, False, False,  True, False, False, False, False, True,  True])

>>> example2.isdigit().to_ndarray() == np.char.isdigit(example2.to_ndarray())
array([ True,  True,  True,  True,  True,  True,  True,  True,  True, True, False])
```

src/SegmentedString.chpl
```diff
@@ -1524,7 +1536,35 @@ module SegmentedString {
...
+      proc toUint64(c) {
+        var ret:uint = 0;
+        var i = 0;
+        for b in c.bytes() {
+          var tmp = b:uint;
+          ret |= (tmp << (i*8));
+          i += 1;
+        }
+        return ret;
+      }
```
stress-tess, line 1561: prob worth adding a comment to explain why this function is here/ what it's doing. I don't think it's obvious that it's here to handle multi-byte cases (or how it does that) and to make it easier to use our integer In1d proc

src/SegmentedString.chpl
```diff
@@ -1524,7 +1536,35 @@ module SegmentedString {
     The SegFunction called by computeOnSegments for isdigit
   */
   inline proc stringIsDigit(ref values, rng) throws {
-    return interpretAsString(values, rng, borrow=true).isDigit();
+    use In1d;
+    const special_digits = "⁰¹²³⁴⁵⁶⁷⁸⁹₀₁₂₃₄₅₆₇₈₉";
+    const myString = interpretAsString(values, rng, borrow=true);
```
stress-tess, line 1541: we're mixing `_` and camelCase. We normally use camelCase for our chpl code

src/SegmentedString.chpl
```diff
@@ -1524,7 +1536,35 @@ module SegmentedString {
...
+        for b in c.bytes() {
+          var tmp = b:uint;
+          ret |= (tmp << (i*8));
+          i += 1;
+        }
+        return ret;
+      }
+      var special_array = [s in special_digits] toUint64(s);
+      var myString_array = [b in myString] toUint64(b);
+      var digitTruth = [b in myString] b.isDigit();
```
stress-tess, line 1564: we're mixing `_` and camelCase. We normally use camelCase for our chpl code

arkouda/strings.py
```diff
@@ -589,6 +589,43 @@ def title(self) -> Strings:
...
+
+        Examples
+        --------
+        >>> not_decimal = ak.array([f'Strings {i}' for i in range(3)])
+        >>> decimal = ak.array([f'12{i}' for i in range(3)])
+        >>> strings = ak.concatenate([not_decimal, decimal])
+        >>> strings
+        array(['Strings 0', 'Strings 1', 'Strings 2', '120', '121', '122'])
+        >>> strings.isdecimal()
+        array([False False False True True True])
```
stress-tess, line 620: would it be worth adding an example with some of the weirder cases?

stress-tess (approved):
The logic looks great! just some small suggested about style and adding comments

src/SegmentedString.chpl
jaketrookman: Added

src/SegmentedString.chpl
jaketrookman: Fixed

src/SegmentedString.chpl
jaketrookman: Fixed

arkouda/strings.py
jaketrookman: Added

ajpotts (changes requested):
Looks good!  Only problem is that I can't get it to compile with gasnet:
```
                       normalize :   1.088 seconds
                 checkNormalized :   0.022 seconds
           buildDefaultFunctions :   0.194 seconds
             createTaskFunctions :   0.050 seconds
/home/amandapotts/git/arkouda//src/Unique.chpl:62: error: unable to resolve return type
make: *** [Makefile:359: arkouda_server] Error 1

```

src/SegmentedString.chpl
```diff
@@ -1520,11 +1528,48 @@ module SegmentedString {
...
+        for b in c.bytes() {
+          var tmp = b:uint;
+          ret |= (tmp << (i*8));
+          i += 1;
+        }
+        return ret;
+      }
+      var specialArray = [s in specialDigits] toUint64(s);
+      var myStringArray = [b in myString] toUint64(b);
+      var digitTruth = [b in myString] b.isDigit();
```
stress-tess, line 1569: in re: to gasnet failing to compile. I was able to get it to compile (i haven't checked if it's working correctly) by making these distributed arrays.
```chapel
var specialArray = makeDistArray([s in specialDigits] toUint64(s));
var myStringArray = makeDistArray([b in myString] toUint64(b));
var digitTruth = makeDistArray([b in myString] b.isDigit());
```

So that seems to work. But it's prob not what we want to do because why distribute these small arrays when we don't need to. The issue is calling `uniqueSortWithInverse` with non-distributed arrays has 2 code paths, the normal one will return non-distributed arrays if that's what's passed in. But there's an early out is `size == 0` which is always distributed. So now there are cases where it sometimes returns distributed arrays and sometimes not and the compiler is rightfully upset about that lol

I'm looking into this more to see if i can find a better solution

stress-tess: yah this is better, just change all the `makeDistArray` calls in `uniqueSortWithInverse ` to use `aD.size` and it fixes the issue. Diff included below
```diff
diff --git a/src/Unique.chpl b/src/Unique.chpl
index 56295452a..8dd14078e 100644
--- a/src/Unique.chpl
+++ b/src/Unique.chpl
@@ -62,9 +62,9 @@ module Unique
     proc uniqueSortWithInverse(a: [?aD] ?eltType) throws {
         if (aD.size == 0) {
             try! uLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),"zero size");
-            var u = makeDistArray(0, eltType);
-            var c = makeDistArray(0, int);
-            var inv = makeDistArray(0, int);
+            var u = makeDistArray(aD.size, eltType);
+            var c = makeDistArray(aD.size, int);
+            var inv = makeDistArray(aD.size, int);
             return (u, c, inv);
         }
         var sorted = makeDistArray(aD, eltType);
@@ -84,7 +84,7 @@ module Unique
         // check there's enough room to create a copy for scan and throw if creating a copy would go over memory limit
         overMemLimit(numBytes(int) * bcast.size);
         bcast = (+ scan bcast);
-        var inv = makeDistArray(aD, int);
+        var inv = makeDistArray(aD.size, int);
         forall (p, b) in zip(perm, bcast) with (var agg = newDstAggregator(int)) {
             agg.copy(inv[p], b);
         }
```

stress-tess: i just got off a video chat with jake, but thought it might be helpful for others to know how to go about debugging an issue like this. What I did was look at the locations of the error
```
/home/amandapotts/git/arkouda//src/Unique.chpl:62: error: unable to resolve return type
```
and saw line 62 of `Unique.chpl` was `uniqueSortWithInverse`. The error message says chpl is unable to figure out the return type, so I looked for all the spots where we return in that `proc` and tried to get their return types to be as similar as possible. `uniqueSortWithInverse` is included below but for simplicity I remove everything that's not a variable declaration or a return, because all that we care about
```chpl
proc uniqueSortWithInverse(a: [?aD] ?eltType) throws {
    if (aD.size == 0) {
        var u = makeDistArray(0, eltType);
        var c = makeDistArray(0, int);
        var inv = makeDistArray(0, int);
        return (u, c, inv);
    }
    var (u, c) = uniqueFromSorted(sorted);
    ...
    var inv = makeDistArray(aD, int);
    ...
    return (u, c, inv);
}
```

Looking at `uniqueFromSorted`, I see one code path is:
```chpl
proc uniqueFromSorted(sorted: [?aD] ?eltType, param needCounts = true) throws {
    ...
    if (allUnique == aD.size) {
        try!  uLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                "early out already unique");
        var u = makeDistArray(aD.size, eltType);
        u = sorted; // array is already unique
        if (needCounts) {
            var c = makeDistArray(aD.size, int);
            c = 1;
            return (u,c);
        } else {
            return u;
        }
    }
    ...
}
```
so `u` and `c` are declared with `makeDistArray(aD.size, int);`. so our two code paths are
```chpl
proc uniqueSortWithInverse(a: [?aD] ?eltType) throws {
    if (aD.size == 0) {
        var u = makeDistArray(0, eltType);
        var c = makeDistArray(0, int);
        var inv = makeDistArray(0, int);
        return (u, c, inv);
    }
    var u = makeDistArray(aD.size, eltType);
    var c = makeDistArray(aD.size, int);
    var inv = makeDistArray(aD, int);
    return (u, c, inv);
}
```
So if I change them all to `makeDistArray(aD.size, int)`, we'd be sure the return types line up. This doesn't mess up the early out because we know `aD.size == 0`

jaketrookman: Updated and confirmed that it builds and runs on gasnet

stress-tess (approved):

ajpotts (approved):

PR #3016, opened 2024-03-02
Closes #3008: Add generator sym entry and stateful uniform distribution

### LINKED ISSUE #3008: Add generator sym entry and stateful uniform distribution
Add symbol entry for generators so they can persist and add a uniform distributions with state (so consecutive calls to `rng.intergers(...)` don't return same array)

Most of the generator functions we have right now rely on a uniform distribution (`integers`, `random`, `uniform`) so we can convert all these to use the stateful uniform distribution

stress-tess:
This PR (closes #3008) adds a generator sym entry to the symbol table. It also add a stateful uniform distribution. So repeated method calls won't return the same array with a set seed, but a new generator with the same seed will produce the same sequence of random arrays. It updates the methods which draw a from a uniform distribution to use the stateful variant

Example:
```python
>>> rng = ak.random.default_rng(74)

>>> rng.integers(0, 2**32, size=3, dtype="uint")
array([103369013 434448015 1372864948])

>>> rng.uniform(-1.0, 1.0, size=3)
array([-0.31903782753174481 0.35896419541356117 0.45980632831507906])

>>> rng.integers(0, 1, size=5, dtype="bool")
array([False True False True True])

>>> rng.integers(-(2**32), 2**32, size=3, dtype="int")
array([-83456679 -540246299 -3756329859])

# repeat
>>> rng = ak.random.default_rng(74)

>>> rng.integers(0, 2**32, size=3, dtype="uint")
array([103369013 434448015 1372864948])

>>> rng.uniform(-1.0, 1.0, size=3)
array([-0.31903782753174481 0.35896419541356117 0.45980632831507906])

>>> rng.integers(0, 1, size=5, dtype="bool")
array([False True False True True])

>>> rng.integers(-(2**32), 2**32, size=3, dtype="int")
array([-83456679 -540246299 -3756329859])
```

Its adds tests for this functionality. Note the tests for the legacy functions are copied over from the existing ones in `pdarraycreation`

### DIFF
--- a/PROTO_tests/tests/random_test.py
+++ b/PROTO_tests/tests/random_test.py
@@ -53,49 +53,6 @@ class TestRandom:
     def test_uniform(self):
         # verify same seed gives different but reproducible arrays
         rng = ak.random.default_rng(18)
-        first = rng.integers(-(2**32), 2**32, 10)
-        second = rng.integers(-(2**32), 2**32, 10)
-        assert first.to_list() != second.to_list()
-
-        rng = ak.random.default_rng(18)
-        same_seed_first = rng.integers(-(2**32), 2**32, 10)
-        same_seed_second = rng.integers(-(2**32), 2**32, 10)
-        assert first.to_list() == same_seed_first.to_list()
-        assert second.to_list() == same_seed_second.to_list()
-
-        # test endpoint
-        rng = ak.random.default_rng()
-        all_zero = rng.integers(0, 1, 20)
-        assert all(all_zero.to_ndarray() == 0)
-
-        not_all_zero = rng.integers(0, 1, 20, endpoint=True)
-        assert any(not_all_zero.to_ndarray() != 0)
-
-        # verify that switching dtype and function from seed is still reproducible
-        rng = ak.random.default_rng(74)
-        uint_arr = rng.integers(2**32, size=10, dtype="uint")
-        float_arr = rng.uniform(-1.0, 1.0, size=5)
-        bool_arr = rng.integers(0, 1, size=20, dtype="int")
-        int_arr = rng.integers(-(2**32), 2**32, size=10, dtype="int")
-
-        rng = ak.random.default_rng(74)
-        same_seed_uint_arr = rng.integers(2**32, size=10, dtype="uint")
-        same_seed_float_arr = rng.uniform(-1.0, 1.0, size=5)
-        same_seed_bool_arr = rng.integers(0, 1, size=20, dtype="int")
-        same_seed_int_arr = rng.integers(-(2**32), 2**32, size=10, dtype="int")
-
-        assert uint_arr.to_list() == same_seed_uint_arr.to_list()
-        assert float_arr.to_list() == same_seed_float_arr.to_list()
-        assert bool_arr.to_list() == same_seed_bool_arr.to_list()
-        assert int_arr.to_list() == same_seed_int_arr.to_list()
-
-        # verify within bounds (lower inclusive and upper exclusive)
-        rng = ak.random.default_rng()
-        bounded_arr = rng.integers(-5, 5, 1000)
-        assert all(bounded_arr.to_ndarray() >= -5)
-        assert all(bounded_arr.to_ndarray() < 5)
-        # verify same seed gives different but reproducible arrays
-        rng = ak.random.default_rng(18)
         first = rng.uniform(-(2**32), 2**32, 10)
         second = rng.uniform(-(2**32), 2**32, 10)
         assert first.to_list() != second.to_list()
--- a/PROTO_tests/tests/random_test.py
+++ b/PROTO_tests/tests/random_test.py
@@ -1,0 +1,270 @@
+import numpy as np
+import pytest
+
+import arkouda as ak
+
+
+class TestRandom:
+    def test_integers(self):
+        # verify same seed gives different but reproducible arrays
+        rng = ak.random.default_rng(18)
+        first = rng.integers(-(2**32), 2**32, 10)
+        second = rng.integers(-(2**32), 2**32, 10)
+        assert first.to_list() != second.to_list()
+
+        rng = ak.random.default_rng(18)
+        same_seed_first = rng.integers(-(2**32), 2**32, 10)
+        same_seed_second = rng.integers(-(2**32), 2**32, 10)
+        assert first.to_list() == same_seed_first.to_list()
+        second.to_list() == same_seed_second.to_list()
+
+        # test endpoint
+        rng = ak.random.default_rng()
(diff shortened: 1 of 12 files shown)


pierce314159:
I forgot there's been changes to the chpl random module in recent versions, I'll need to make this compatible with chpl 1.31 and 1.32

pierce314159:
I did my best updating the random compat modules, but that was mostly just me loosening the memory management requirements to make the errors go away lol. @jeremiah-corrado seems to have touched those wouldn't mind double checking my work I'd appreciate it

src/compat/e-132/ArkoudaRandomCompat.chpl
```diff
@@ -2,14 +2,24 @@ module ArkoudaRandomCompat {
...
     }
     proc init(type t, seed) {
       eltType = t;
-      r = new owned PCGRandomStream(eltType, seed);
+      r = new shared PCGRandomStream(eltType, seed);
+    }
+    proc ref fill(ref arr: []) where arr.isRectangular() {
+      var cpyArr = arr;
+      r.fillRandom(cpyArr);
+      arr = cpyArr;
```
jeremiah-corrado, line 17: I'm not seeing why the copy is needed here?

jeremiah-corrado (approved):
Looks good to me! Just the one question about an extra array copy in the compat modules:

src/compat/e-132/ArkoudaRandomCompat.chpl
pierce314159: Thank you for catching this! I think I must've been misinterpreting an error I was hitting (and solved it some other way)

In my head I thought since `fillRandom` took and array literal back in 1.31 and `fill` takes a ref that this was necessary

arkouda/random/__init__.py
```diff
@@ -1,9 +1,8 @@
-from ._generator import Generator, default_rng
+from ._generator import Generator, default_rng  # noqa
```
ajpotts, line 1: I don't understand the comment # noqa

tests/random_test.py
```diff
@@ -0,0 +1,286 @@
...
+        self.assertEqual(int_arr.to_list(), same_seed_int_arr.to_list())
+
+        # verify within bounds (lower inclusive and upper exclusive)
+        rng = ak.random.default_rng()
+        bounded_arr = rng.integers(-5, 5, 1000)
+        self.assertTrue(all(bounded_arr.to_ndarray() >= -5))
+        self.assertTrue(all(bounded_arr.to_ndarray() < 5))
+
+    def test_uniform(self):
+        # verify same seed gives different but reproducible arrays
```
ajpotts, line 53: The first few examples of this test seem redundant with test_integers.  Could they be removed?

PROTO_tests/tests/random_test.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # verify within bounds (lower inclusive and upper exclusive)
+        rng = ak.random.default_rng()
+        bounded_arr = rng.integers(-5, 5, 1000)
+        assert all(bounded_arr.to_ndarray() >= -5)
+        assert all(bounded_arr.to_ndarray() < 5)
+
+    def test_uniform(self):
+        # verify same seed gives different but reproducible arrays
+        rng = ak.random.default_rng(18)
```
ajpotts, line 55: Same comment as below, the first few examples seem redundant to test_integers.  Are they necessary?

ajpotts (commented):
A few questions, as marked, but othewise looks good!

arkouda/random/__init__.py
stress-tess: it tells flake8 to ignore this line because it complains about `default_rng` being imported but not included in the `__all__`

I'm using it here because we want `default_rng` to be visible to the user only under the `Random` module. So you can call `ak.Random.default_rng()`, but not `ak.default_rng()`. This was the easiest way I saw to get this behavior, but I'm open to suggestions!

arkouda/random/_generator.py
```diff
@@ -23,9 +28,11 @@ class Generator:
     default_rng : Recommended constructor for `Generator`.
     """
 
-    def __init__(self, seed=None):
+    def __init__(self, name_dict=None, seed=None, state=1):
```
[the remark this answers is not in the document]
stress-tess: I can definitely do that! I didn't initially because `default_rng()` is the preferred constructor on generators, but it can't hurt to include that info

tests/random_test.py
stress-tess: good catch! can't believe i missed that! i'll get that fixed

arkouda/random/__init__.py
ajpotts: I googled it and it looks like there is another option:
https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used

Change setup.cfg
to add `__init__.py: F401`
like this:

```
[flake8]
max-line-length = 105
extend-ignore = E203
per-file-ignores =
    PROTO_tests/tests/operator_test.py: E501
    PROTO_tests/tests/symbol_table_test.py: F841
    __init__.py: F401

```

But either way is fine with me.

stress-tess: i think leaving it as the `# noqa` is the way to go, since we could want that error on other lines in the file if we accidentally left them out of the `__all__`. It's just this one specifically that is the exception to the rule

arkouda/random/_generator.py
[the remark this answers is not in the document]
stress-tess: absolutely! I'll get right on it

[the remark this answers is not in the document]
stress-tess: done!

tests/random_test.py
stress-tess: done!

PROTO_tests/tests/random_test.py
stress-tess: done!

arkouda/random/__init__.py
ajpotts: Sounds good.

tests/random_test.py
ajpotts: Thanks!

PROTO_tests/tests/random_test.py
ajpotts: Thanks!

arkouda/random/_generator.py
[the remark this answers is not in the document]
ajpotts: I appreciate it.  Thanks!

ajpotts (approved):
Looks good!

jaketrookman (approved):
Looks good

PR #3021, opened 2024-03-07
SegArray optimization & bug fix

brandon-neth:
This PR should improve the dataframe benchmark performance by removing two sources of redundant computation for SegArray columns. It also fixes a bug in the benchmark that only saved the result of the final run. 

Both redundant computations appear in the SegArray initializer, which in my profiling contributed the most to the dataframe benchmark execution time. The first redundant computation occurred when calculating the number of non-empty segments, which was calculated both when setting the `_non_empty` field and when counting them. The second redundant computation was the calculation of segment lengths, which occurs in the `gen_ranges` function, then again in the `SegArray` initializer. I added an optional argument to `gen_ranges` to return the lengths to pass straight to the initializer instead of calculating it again during initialization.

### DIFF
--- a/arkouda/join.py
+++ b/arkouda/join.py
@@ -117,7 +117,7 @@ def join_on_eq_with_dt(
     return resI, resJ
 
 
-def gen_ranges(starts, ends, stride=1):
+def gen_ranges(starts, ends, stride=1, return_lengths=False):
     """
     Generate a segmented array of variable-length, contiguous ranges between pairs of
     start- and end-points.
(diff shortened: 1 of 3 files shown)


bmcdonald3 (approved):
Cool, good catch! Just curious, do you have any performance runs you can put in the description so we have a sense of what kind of improvement to expect? An example of that would be something like https://github.com/Bears-R-Us/arkouda/pull/2290

stress-tess (approved):
nice work!! I would also like to see the improved times, but I'll go ahead and merge in the meantime

brandon-neth:
Sure, these results are from my Mac. Average of 5 runs for 10k-element-long columns.

Original version: .9077 seconds
With non-empty fix: .7495 seconds
Reusing lengths: 0.6762 seconds

PR #3025, opened 2024-03-08
Closes #3024: Add merge based workflow and update sort workflow for sparse sum helper

### LINKED ISSUE #3024: Add merge based workflow and update sort workflow for sparse sum helper
Currently the sparse sum helper combines the two sorted indices array by concatenating and calling radix sort. Which seems like overkill since what we want is to merge two sorted lists

We should implement a merge based workflow to see if we can take advantage of this extra structure to get better performance. We should also add an unordered concat helper to improve the sort based workflow

stress-tess:
This PR (closes #3024) implements a merge based workflow for combining the two sorted index arrays to hopefully take advantage of this extra structure to get better performance than the sort workflow. It also adds an unordered concat helper to improve the sort based workflow


stress-tess:
@bmcdonald3 was nice enough to run a quick test on a 16 nodes cluster for me, and it seems like the merge workflow is a bit faster and produced correct results. So I'm cautiously optimistic that this will perform better on the users machines

The results:
```
merge time = 1.4320 sec
sort time = 1.9938 sec
gb time = 2.2158 sec
```

The quick test:
```python
import time
import arkouda as ak
ak.connect() 

cfg = ak.get_config()
N = 10**8 * cfg["numLocales"]
select_from = ak.arange(N)
inds1 = select_from[ak.randint(0, 10, N) % 3 == 0]
inds2 = select_from[ak.randint(0, 10, N) % 3 == 0]
vals1 = ak.randint(-(2**32), 2**32, N)[inds1]
vals2 = ak.randint(-(2**32), 2**32, N)[inds2]

start = time.time()
merge_idx, merge_vals = ak.util.sparse_sum_help(inds1, inds2, vals1, vals2, merge=True)
end = time.time()
print("merge time = {:.4f} sec".format(end - start))

start = time.time()
sort_idx, sort_vals = ak.util.sparse_sum_help(inds1, inds2, vals1, vals2, merge=False)
end = time.time()
print("sort time = {:.4f} sec".format(end - start))

start = time.time()
gb_idx, gb_vals = ak.GroupBy(ak.concatenate([inds1, inds2], ordered=False)).sum(ak.concatenate((vals1, vals2), ordered=False))
end = time.time()
print("gb time = {:.4f} sec".format(end - start))

assert (merge_idx == sort_idx).all()
assert (merge_idx == gb_idx).all()
assert (merge_vals == sort_vals).all()
assert (merge_vals == gb_vals).all()
```

bmcdonald3 (approved):

src/ArraySetops.chpl
```diff
@@ -114,30 +122,149 @@ module ArraySetops
       return uniqueSort(aux, false);
     }
 
-    proc sortHelper(idx1: [?D] ?t, idx2: [] t, aSize, bSize) throws {
-      // all we are doing is merging two sorted lists, there has to
-      // be a better way than concatenating and doing a full sort
-
-      // eventually we want an if statement to determine if we want to merge or sort
+    proc combineHelper(const ref idx1: [?D] ?t, const ref idx2: [] t, const ref val1: [] ?t2, const ref val2: [] t2, doMerge = false) throws {
```
ajpotts, line 125: Remember to add the fetch condition.

arkouda/util.py
```diff
@@ -292,10 +292,13 @@ def attach_all(names: list):
     return {n: attach(n) for n in names}
 
 
-def sparse_sum_help(idx1, idx2, val1, val2):
+def sparse_sum_help(idx1, idx2, val1, val2, merge=True):
```
ajpotts, line 295: Is there a way to unit test this?

ajpotts (commented):
Looks good.  I just had a few questions, as marked.

src/AryUtil.chpl
```diff
@@ -207,12 +208,33 @@ module AryUtil
...
+      var ret = makeDistArray(retSize, t);
+      if ordered {
+        ret[0..#a.size] = a;
+        ret[a.size..#b.size] = b;
+      }
+      else {
+        // if unordered, we interleave the arrays by concatenating each locale's
+        // local part of the arrays to lower communication between locales
+        // this is a simplified version of the logic in concatenateMsg
+        var blocksizes: [PrivateSpace] int;
```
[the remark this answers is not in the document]
stress-tess: yah it's not super easy to find, but it's in there! https://chapel-lang.org/docs/modules/dists/PrivateDist.html

src/ArraySetops.chpl
```diff
@@ -114,30 +122,149 @@ module ArraySetops
...
+            permutedVals[writeToResultSlice] = [(key, val) in tmp] val;
+          }
+        }
+      }
+      else {
+        var perm = makeDistArray(allocSize, int);
+        forall (s, p, sp) in zip(sortedIdx, perm, radixSortLSD(concatArrays(idx1, idx2, ordered=false))) {
+          (s, p) = sp;
+        }
+        const vals = concatArrays(val1, val2, ordered=false);
```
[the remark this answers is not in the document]
stress-tess: yah so it's prob worth a comment, so we are concatenating the indices array and then the values arrays. This will make sure that indices will line up with their associated values

src/ArraySetops.chpl
```diff
@@ -114,30 +122,149 @@ module ArraySetops
...
+        forall (p, i) in zip(permutedVals, perm) with (var agg = newSrcAggregator(t2)) {
+          agg.copy(p, vals[i]);
```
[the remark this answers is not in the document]
stress-tess: excellent question!

So for:
```chapel 
forall (s, p, sp) in zip(sortedIdx, perm, radixSortLSD(concatArrays(idx1, idx2, ordered=false))) {
  (s, p) = sp;
}
```
Firstly we know arrays that are distributed the same way because they are the same size and block distibuted (If they weren't the same size, chpl would yell at us about trying to zip them together). So this means that `s`, `p`, and `sp` will always be on the same locale as each other (i.e. they are all local to each other). Since they're on the same locale, there's no comm and no need for aggregation. Creating aggregators carries a bit overhead, so it's best to avoid if it's not necessary.

Note: this is a big reason why I'm a fan of using looping variables instead of indexing into the arrays since it makes it easier to tell at a glance that these values are local to the locale where the computation is happening. (also i just think it looks cleaner). We could rewrite this as the below (and it would be equivalent), but it would be tougher to know what was local:
```chapel
var sortedArr = radixSortLSD(concatArrays(idx1, idx2, ordered=false));
forall i in sortedIdx.domain {
  (sortedIdx[i], perm[i]) = sortedArr[i];
}
```

for the example in the comment:
```chapel
forall (p, i) in zip(permutedVals, perm) with (var agg = newSrcAggregator(t2)) {
  agg.copy(p, vals[i]);
}
```
Using the reasoning in above case, we are confident that `p` and `i` are local to one another. But we don't know that `p` and `vals[i]` live on the same locale. if `vals[i]` lives on a different locale (and there's a very good chance it does) then we want to avoid the comm of writing `p` int `vals[i]`. So using an aggregator allows us to wait until we have a couple that need to go to a locale and send them all over at once

Onces again this highlights how nice following the convention of looping variables / zipping is when possible because the fact that we are indexing into vals is an indicator that it could be remote (i.e. living on a different local)

src/ArraySetops.chpl
stress-tess: thanks for the reminder!

arkouda/util.py
stress-tess: Good question! there is... with a catch lol. A slightly modified version of the quick and dirty benchmark I had ben run would work to verify we are matching the sum aggregation. The catch is that our CI doesn't run multi-locale and that's the only case where this could run into trouble, so it wouldn't be super useful for us

But when this gets pulled onto user machines if they run `make test` with a multi-locale build, it would give us some info if that test failed. So it is worth adding

src/ArraySetops.chpl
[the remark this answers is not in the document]
ajpotts: I should have looked more closely.  Thanks for clarifying.

arkouda/util.py
stress-tess: Added!

src/ArraySetops.chpl
stress-tess: Added!

jaketrookman (approved):
Looks great

src/AryUtil.chpl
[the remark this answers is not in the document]
ajpotts: Interesting.  Thanks!

src/ArraySetops.chpl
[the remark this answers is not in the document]
ajpotts: Great explanation.  Thanks!

arkouda/util.py
ajpotts: Thanks!

src/ArraySetops.chpl
ajpotts: Thanks!

ajpotts (approved):

PR #3027, opened 2024-03-11
Array API improvements

jeremiah-corrado:
Several small improvements to the array-api implementation:

- implementation of missing special methods on Array type
- implementation of `where` and `nonzero` methods
- bug fix for creating 0D arrays
- bug fix for applying stats functions to 1D arrays
- fix a bug where ManipulationMsg.chpl wasn't building in a multi-locale config
- added framework for intercepting numpy calls with array-api functions (see numpy's [docs](https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_function__))
- added skipNan option to stats and (some) reduction operations
- improved printing of ND arrays — now a recursive generalization of the 1D format. for example, a 2D array summary would look like: 
```[[0 1 2 ... 98 99 100] [0 1 2 ... 98 99 100] [0 1 2 ... 98 99 100] ... [0 1 2 ... 98 99 100] [0 1 2 ... 98 99 100] [0 1 2 ... 98 99 100]]```

### CODE REVIEW

File: ServerModulesArrayApi.cfg [hunk 1]
```diff
@@ -18,7 +18,7 @@ In1dMsg
 IndexingMsg
 JoinEqWithDTMsg
 KExtremeMsg
-LinalgMsg
+# LinalgMsg
```
bmcdonald3, line 21: Intentionally commented out?

arkouda/array_api/_array_object.py
```diff
@@ -533,3 +774,11 @@ def size(self) -> int:
...
     def T(self) -> Array:
         raise ValueError("Not implemented")
+
+
+def implements_numpy(numpy_function):
+    """Register an __array_function__ implementation for MyArray objects."""
+    def decorator(func):
+        HANDLED_FUNCTIONS[numpy_function] = func
+        return func
+    return decorator
```
bmcdonald3, line 784: In order for this to work, the Arkouda function header has to exactly match the NumPy function header, is that correct?

src/compat/e-132/SymArrayDmapCompat.chpl
```diff
@@ -40,7 +40,7 @@ module SymArrayDmapCompat
                 }
                 // fix the annoyance about boundingBox being empty
                 else {
-                  return {0..#0} dmapped blockDist(boundingBox={0..0});
+                  return dom dmapped blockDist(boundingBox=dom.expand(1));
```
bmcdonald3, line 43: What is the point of this change? I thought there was a reason why it had to be the old way, but I guess I could have just made that up since I don't know what the reason would be.

bmcdonald3 (approved):
Looks good, just a couple of questions.

ServerModulesArrayApi.cfg
jeremiah-corrado: Yeah, that module is an optional part of the array api and it's not finished yet, so I figured I'd remove it for now to speed up compilation times.

arkouda/array_api/_array_object.py
jeremiah-corrado: That's right

jeremiah-corrado: (one cavait: I think the header can be missing some optional arguments and it will still work as long as the caller doesn't use them)

src/compat/e-132/SymArrayDmapCompat.chpl
jeremiah-corrado: It fixes a compilation error I was seeing in multi-locale builds where the compiler couldn't figure out the return type of this function (`makeDistDom`). The first branch of this conditional calls `return blockDist.createDomain(dom)`, which can return a mutlidimensional domain, but this branch was always returning a single dimensional domain.

The new code still creates an empty domain with a non-empty bounding box but its rank matches the domain from the first branch of the conditional.

PR #3039, opened 2024-03-15
Fixes #3035 - Inconsistent results when broadcasting with empty segments

### LINKED ISSUE #3035: inconsistent results when broadcasting with empty segments on multilocale
A user reported this and I was able to reproduce. Below is what I'm seeing with gasnet and 3 locales
```python
>>> segments = ak.array([3,3,5,6,6,7,7])
>>> values = ak.arange(7)
>>> size = 10
>>> ak.broadcast(segments, values, size)
array([0 0 0 1 1 2 3 4 4 4])

>>> ak.broadcast(segments, values, size)
array([0 0 0 1 1 2 3 4 4 4])

>>> ak.broadcast(segments, values, size)
array([0 0 0 0 0 1 2 3 3 3])
```

stress-tess:
This PR (fixes #3035) a race condition bug that caused inconsistent results when broadcasting with empty segments (though both results given were incorrect).

We get around this by masking out the segments and values associated with the empty segments. We only do this when empty segments exist, so it shouldn't affect the performance of the common case too much but will give correct results in the empty seg case.

More details on bug in comment below

### DIFF
--- a/PROTO_tests/tests/groupby_test.py
+++ b/PROTO_tests/tests/groupby_test.py
@@ -209,6 +209,30 @@ class TestGroupBy:
         assert bcast.sum() == check
         assert (bcast[:-1] >= bcast[1:]).all()
 
+    def test_empty_segs_broadcast(self):
+        # verify the reproducer from issue #3035 gives correct answer
+        # note: this was due to a race condition, so it will only appear with multiple locales
+
+        # test with int and bool vals
+        for vals in ak.arange(7), (ak.arange(7) % 2 == 0):
+            segs = ak.array([3, 3, 5, 6, 6, 7, 7])
+            size = 10
+            perm = ak.array([9, 1, 0, 5, 7, 2, 8, 4, 3, 6])
+
+            # filter out empty segs
+            non_empty_segs = ak.array([False, True, True, False, True, False, True])
+            compressed_segs = segs[non_empty_segs]
+            compressed_vals = vals[non_empty_segs]
+
+            assert (
+                ak.broadcast(segs, vals, size).to_list()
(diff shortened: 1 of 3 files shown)


stress-tess:
The result from the reproducer with multi-locale enabled:
```python
>>> segments = ak.array([3,3,5,6,6,7,7])
>>> values = ak.arange(7)
>>> size = 10
>>> ak.broadcast(segments, values, size)
array([0 0 0 1 1 2 3 4 4 4])

>>> ak.broadcast(segments, values, size)
array([0 0 0 1 1 2 3 4 4 4])

>>> ak.broadcast(segments, values, size)
array([0 0 0 0 0 1 2 3 3 3])
```

This unexpected result is due to a race condition in this `forall` loop:
https://github.com/Bears-R-Us/arkouda/blob/9ab6f1774055690507844c1439932529b037deff/src/Broadcast.chpl#L151-L153

if we print out the values for segs and diffs we get:
```
segs: 3 3 5 6 6 7 7
diffs: 0 1 1 1 1 1 1
```
we see that the seg `3` has two associated diffs, so `expandedVals[3]` will vary depending on the order that these two segments write

But the actual result should be should be `array([0 0 0 1 1 2 4 6 6 6])` since
- we don't specify what goes in `(0,3)` so it remains uninitialized (i.e. `0`)
- `0` goes in segment `(3,3)` which is empty
- `1` goes in segment `(3,5)`
- `2` goes in segment `(5,6)`
- `3` goes in segment `(6,6)` which is empty
- `4` goes in segment `(6,7)`
- `5` goes in segment `(7,7)` which is empty
- `6` goes in segment `(7,10)`

jaketrookman (approved):

PR #3061, opened 2024-03-22
Closes #3054: Dynamically switch to batching for larger csv writes

### LINKED ISSUE #3054: Dynamically switch to batching for larger csv writes
The optimization in https://github.com/Bears-R-Us/arkouda/pull/3053 gets much better performance when writing small csv files, but runs into memory issues above a certain size since we put all the data that needs to be written on a locale into a chapel native strings array

I think we can trade some runtime for less memory usage by writing in blocks. The idea is to take slices of `localSubdomain.size/N`, so the strings array will only have `(all data the locale needs to write) / N` in a strings array at once. I would like to use memory usage to try and dynamically switch over to writing block

stress-tess:
This PR (closes #3054) adds the ability to write chunks of data when writing csv files that would cause us to run out of memory when using the new optimizatioin. We use the arkouda memory management functions to determine approximately how much memory is available on each locale, and then divide the data we want to write on that locale into slices of that size. This makes sure the chapel native strings array is as big as possible without running out of. Hopefully we'll keep the performance bump for the small cases

### DIFF
--- a/src/CSVMsg.chpl
+++ b/src/CSVMsg.chpl
@@ -205,8 +205,8 @@ module CSVMsg {
         // calculate the amount of memory required to write the csv file all in one go
         var memRequired: uint = 0;
         var numStrings = 0;
-        for (cname, dt) in zip(dtypes, datasets) {
-            if dt == "Strings" {
+        for (cname, dt) in zip(datasets, dtypes) {
+            if dt == "str" {
                 var ss = new SegString("", toSegStringSymEntry(st.lookup(cname)));
                 // strings vals are uint(8), so they're only one byte but overcount
                 // because they're not evenly distributed across locales
@@ -220,7 +220,10 @@ module CSVMsg {
 
         // if we can put the entire df in mem (i.e. memLim > (memUsed + memRequired)), then we can do it all in one go.
         // Otherwise we can only put (memLim - memUsed) in memory at a time, so that's our batchSize
-        const numBatches = if memLim > (memUsed + memRequired) then 1 else ceil(memRequired:real / (memLim - memUsed)):int;
+        // if memLim > memUsed then we use chunks no bigger than 5% of total memory (memLim is 90% so dividing by 18 gives us 5%)
+        const numBatches = if memLim > memUsed then ceil(memRequired:real / (memLim - memUsed)):int else ceil(memRequired:real / (memLim / 18)):int;
+
+        csvLogger.info(getModuleName(),getRoutineName(),getLineNumber(), "Start csv write with %i batches".doFormat(numBatches));
 
         coforall loc in Locales do on loc {
             const localeFilename = filenames[loc.id];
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Great improvements, looks good to me

src/CSVMsg.chpl
```diff
@@ -197,7 +197,32 @@ module CSVMsg {
...
+                numStrings += 1;
+            }
+        }
+        // assume pdarrays are 64 bits / 8 bytes (this is an overestimate for bools, but that's fine)
+        memRequired += ((ndsets-numStrings) * row_count * 8);
+        memRequired = (memRequired:real / numLocales):uint;
+
+        // if we can put the entire df in mem (i.e. memLim > (memUsed + memRequired)), then we can do it all in one go.
+        // Otherwise we can only put (memLim - memUsed) in memory at a time, so that's our batchSize
+        const numBatches = if memLim > (memUsed + memRequired) then 1 else ceil(memRequired:real / (memLim - memUsed)):int;
```
ajpotts, line 223: I don't think the ternary is necessary:
`memLim > (memUsed + memRequired)`  =>
`memLim - memUsed >  memRequired` = >
`1 > memRequired/(memLim - memUsed)` =>
`ceil(memRequired:real / (memLim - memUsed)) == 1`

ajpotts, line 223: Also, what happens if `memUsed > memLim`?

ajpotts (approved):
Looks good other than the questions I marked.

src/CSVMsg.chpl
```diff
@@ -197,7 +197,32 @@ module CSVMsg {
...
-        coforall loc in Locales do on loc{
+        // only use 90% of getMemLimit to give ourselves some buffer
+        // these are each per locale measurements
+        const memLim = (getMemLimit() * (.9)):uint;
+        const memUsed = getMemUsed();
+
+        // calculate the amount of memory required to write the csv file all in one go
+        var memRequired: uint = 0;
+        var numStrings = 0;
+        for (cname, dt) in zip(dtypes, datasets) {
```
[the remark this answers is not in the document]
stress-tess: Good catch! I just got these backwards! This section was never running, I'll go ahead fix that!

src/CSVMsg.chpl
stress-tess: Ah yah! I originally was gonna use a bool to switch between the orig code and the batch. But eventually I changed it to always batch and sometimes only have one batch. It seems I didn't think too hard once I had that change

stress-tess: yaaaah that would give us a negative number of batches, which is def not what we want. It likely means we'll run out of memory since we'd already be at 90% of our memory limit. I think we should prob check for that.

If we've exceeded 90% of mem on the system then we will break the chunks into at most size 5% to give us some buffer

stress-tess: If we were over 95% and trying to write a large csv, we will likely run out of memory during the operation, but I feel like that's reasonable

stress-tess: so we can use this new and improved (hopefully actually useful) ternary! lol
```chapel
const numBatches = if memLim > memUsed then ceil(memRequired:real / (memLim - memUsed)):int else ceil(memRequired:real / (memLim / 18)):int;
```

PR #3064, opened 2024-03-25
Closes #3063: Fix deprecation warnings

### LINKED ISSUE #3063: compat modules deprecation warnings
update our compat modules to fix the deprecations warnings

stress-tess:
Closes #3063

Read _*just barely*_ enough about this-intent on the chapel docs to get the warning to go away, so hopefully it is in fact what we want (hopefully jeremiah can validate that 😁). [This bit of the chpl docs](https://chapel-lang.org/docs/language/spec/methods.html#the-method-receiver-and-the-this-argument) seems to state a `ref` this-intent is what we are looking for since we're modifying `this`


jeremiah-corrado:
Yep, that's right. My bad for not catching this initially.

jaketrookman (approved):

PR #3070, opened 2024-03-29
Array API Set functions

jeremiah-corrado:
Implement the Array API set functions (documentation [here](https://data-apis.org/array-api/latest/API_specification/set_functions.html)).


bmcdonald3 (approved):
Looks good.

PR #3072, opened 2024-04-02
Fix --print-used-modules for functions registered with `@arkouda.registerND`

jeremiah-corrado:
Some quality of life fixes related to the Array API work:

* add `modName` arguments to registration in `parseServerConfig.py` s.t. the --print-used-modules flag works for functions registered via `@arkouda.registerND`
* fix formatting in `parseServerConfig.py`
* rename: ServerConfigArrayAPI.cfg  => ArrayAPIServerConfig.cfg to make autocomplete faster in the common case
* remove some unused modules from ArrayAPIServerConfig.cfg to speed up Array-API compilation

### DIFF
--- a/ArrayAPIServerModules.cfg
+++ b/ArrayAPIServerModules.cfg
@@ -2,37 +2,20 @@
 # a module from a build
 
 ArgSortMsg
-ArraySetopsMsg
-BroadcastMsg
 CastMsg
-ConcatenateMsg
-CSVMsg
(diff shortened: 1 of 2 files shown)


bmcdonald3 (approved):
Thanks for doing this! @stress-tess let us know if this fixes what you were running into!

bmcdonald3 (approved):

PR #3076, opened 2024-04-03
Add support for large string Parquet type

bmcdonald3:
The new Taxi cab Parquet data stores some strings as "large strings" so we need to account for those in Arkouda".


stress-tess (approved):
Looks good!

PR #3082, opened 2024-04-05
Closes #3083: Optimize Parquet string read code 

### LINKED ISSUE #3083: Optimize Parquet string reading
This PR reworks the existing Parquet string read implementation by implementing a distinct string path, due to differences in the layout of the strings.

Parquet strings cannot be read directly into Chapel buffers, like they can for other Arkouda types, like int and bool, due to Arrow having a unique representation of strings. 

The way that Parquet reading was implemented prior to this PR was reading each Parquet file once, the first time to get the number of bytes from the file, which requires reading the entire file, since number of bytes is not available in the metadata, and then the second time to actually read and store the data into the Arkouda buffer. Each Parquet string was read 1 value at a time and then deep copied on the C++ side into the Chapel buffer.

This PR takes a different approach, reading all of the values in once, storing them in an intermediary data structure to avoid the second file read.

To get the data from Arrow into Chapel, we are creating a row group reader for each row group in the file, which needs to be kept alive to avoid cleaning up the data from the actual Arrow strings. Once we have all of the data stored in the interim data structure, we can then do the copies into the Chapel buffer in parallel on the Chapel side.

Performance results reading one large file collected on an XC with a Lustre file system:

| nl   | master | my branch |
| ----:| ------:| ---------:|
| 1    | 8.17s  | 1.56s     |
| 2    | 8.06s  | 1.95s     |
| 4    | 8.32s  | 1.90s     |
| 8    | 8.31s  | 1.96s     |

So, this results in a 4-5x speedup.

The reason that 1 nodes reads faster than 2 nodes is that each file is assigned to a locale, rather than a chunk (like the existing implementation), which means that the first node will read the contents and then broadcast that out to the other nodes, which results in a modest amount of overhead.

bmcdonald3:
This PR reworks the existing Parquet string read implementation by implementing a distinct string path, due to differences in the layout of the strings.

Parquet strings cannot be read directly into Chapel buffers, like they can for other Arkouda types, like int and bool, due to Arrow having a unique representation of strings. 

The way that Parquet reading was implemented prior to this PR was reading each Parquet file once, the first time to get the number of bytes from the file, which requires reading the entire file, since number of bytes is not available in the metadata, and then the second time to actually read and store the data into the Arkouda buffer. Each Parquet string was read 1 value at a time and then deep copied on the C++ side into the Chapel buffer.

This PR takes a different approach, reading all of the values in once, storing them in an intermediary data structure to avoid the second file read.

To get the data from Arrow into Chapel, we are creating a row group reader for each row group in the file, which needs to be kept alive to avoid cleaning up the data from the actual Arrow strings. Once we have all of the data stored in the interim data structure, we can then do the copies into the Chapel buffer in parallel on the Chapel side.

Performance results reading one large file collected on an XC with a Lustre file system:

| nl   | master | my branch |
| ----:| ------:| ---------:|
| 1    | 8.17s  | 1.56s     |
| 2    | 8.06s  | 1.95s     |
| 4    | 8.32s  | 1.90s     |
| 8    | 8.31s  | 1.96s     |

So, this results in a 4-5x speedup.

The reason that 1 nodes reads faster than 2 nodes is that each file is assigned to a locale, rather than a chunk (like the existing implementation), which means that the first node will read the contents and then broadcast that out to the other nodes, which results in a modest amount of overhead.

Closes #3083


src/ParquetMsg.chpl
```diff
@@ -705,6 +725,102 @@ module ParquetMsg {
...
+
+  proc getRowGroupNums(ref distFiles, ref numRowGroups) {
+    coforall loc in distFiles.targetLocales() with (ref numRowGroups) do on loc {
+      var locFiles: [distFiles.localSubdomain()] string = distFiles[distFiles.localSubdomain()];
+      for i in locFiles.domain {
+        c_openFile(locFiles[i].localize().c_str(), getReaderIdx(i,0));
+        numRowGroups[i] = c_getNumRowGroups(getReaderIdx(i,0));
+        for j in 2..numRowGroups[i] {
+          c_openFile(locFiles[i].localize().c_str(), getReaderIdx(i,j-1));
+        }
```
stress-tess, line 741: I don't totally get why `j` starts at `2` and not `1` because we have `getReaderIdx(i,0)` calls ahead of this

ohhh I think I just answered my own question lol. the first bit handles the first row group in the file and the for loop handles any remaining row groups...

which leads me to the question, why not remove the first 2 lines and just have 
```chapel
for 1 in 2..numRowGroups[i] {
  c_openFile(locFiles[i].localize().c_str(), getReaderIdx(i,j-1));
}
```

src/ParquetMsg.chpl
```diff
@@ -705,6 +725,102 @@ module ParquetMsg {
...
+        var fname = locFiles[i];
+        var locDsetname = dsetname;
+        var startIdx = locSubdoms[i].low;
+        for rg in 0..#numRowGroups[i] {
+          var totalBytes = 0;
+          startIdxs[i][rg] = startIdx;
+
+          var numRead = 0;
+
+          if c_readParquetColumnChunks(fname.localize().c_str(), 8192, len, getReaderIdx(i,rg), c_ptrTo(numRead), c_ptrTo(externalData[i][rg]), c_ptrTo(containsNulls[i][rg]), c_ptrTo(errs[i][1].errMsg)) == ARROWERROR {
```
stress-tess, line 775: why a batch size of 8192? I feel like I vaguely remember that empirically this had the best performance. if that is the case then I think adding a comment would be helpful

src/ParquetMsg.chpl
```diff
@@ -705,6 +725,102 @@ module ParquetMsg {
...
+            errs[i] = (true, errs[i][1]);
+          }
+          var tmp: [startIdx..#numRead] int;
+          forall (id, j) in zip(0..#numRead, startIdx..#numRead) with (+ reduce totalBytes) {
+            ref curr = (externalData[i][rg]: c_ptr(MyByteArray))[id];
+            tmp[j] = curr.len + 1; // this was only change
+            totalBytes += curr.len+1;
+          }
+          entrySeg.a[startIdx..#numRead] = tmp;
```
stress-tess, line 784: I'm struggling to understand exactly how this bit works

stress-tess (approved):
Looks good to me!! Thanks for doing this! I don't envy u having to figure out how to keep the maps from going out of scope switching from c++ to chpl lol

src/ArrowFunctions.h
```diff
@@ -41,6 +41,33 @@ extern "C" {
...
+                                void** outData, bool* containsNulls, char** errMsg);
+  int cpp_readParquetColumnChunks(const char* filename, int64_t batchSize, int64_t numElems,
+                                  int64_t readerIdx, int64_t* numRead,
+                                  void** outData, bool* containsNulls, char** errMsg);
+
+  int c_getNumRowGroups(int64_t readerIdx);
+  int cpp_getNumRowGroups(int64_t readerIdx);
+  
+  void c_free(void* arr);
+  void cpp_free(void* arr);
```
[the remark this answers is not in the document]
bmcdonald3: Leftovers from something I tried earlier, removing.

src/ParquetMsg.chpl
```diff
@@ -119,6 +124,21 @@ module ParquetMsg {
...
 
+  proc getRGSubdomains(bytesPerRG: [?D] ?t, maxRowGroups: int) {
+    var rgSubdomains: [D] [0..#maxRowGroups] domain(1);
+
+    var offset = 0;
+    for i in D {
+      for rg in 0..#maxRowGroups {
+        if bytesPerRG[i][rg] != 0 {
+          rgSubdomains[i][rg] = {offset..#bytesPerRG[i][rg]};
+          offset += bytesPerRG[i][rg];
```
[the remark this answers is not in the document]
bmcdonald3: Well, since we are basing the next output on the result of the previous with `offset += bytesPerRG[i][rg];`, I don't think we can parallelize as-is. We could pull that into a precompute to parallelize, but this is such a small amount of execution time, I don't think it matters.

src/ParquetMsg.chpl
bmcdonald3: Ya, good point.

src/ParquetMsg.chpl
bmcdonald3: Whoops, thanks for pointing this out, we actually have a private config const that we should be using here so that users can modify that value, I had hardcoded it in some testing, but we don't want that.

src/ParquetMsg.chpl
bmcdonald3: The external data holds the Parquet `ByteArray`s that actually store the data and we need to get all the bytes from that data to create our Chapel array. So, we do that by getting the chunk of this columns data (`externalData[i][rg]`) and then cast that to a byte array (it's a C++ class, so we have to have a C wrapper to it, which is why `MyByteArray` isn't the real Arrow type, but it's same amount of memory) and then we get the actual individual string with the `[id]` index. From there, we can get the total number of bytes by adding each strings length and then we get the segmented lengths of the actual strings into our seg array here too with the `tmp[j] = curr.len+1;`. The `+1` is because the Arrow strings are not null-terminated, so we need to have 1 more character in our Arkouda array per string.

bmcdonald3:
Also, in regards to your higher level comment, the hard part here wasn't getting the maps in place, it was figuring out that we needed them in the first place! There is no indication in the API that you need to keep column readers in scope to keep string allocated data around, so that was a long rabbit chase to figure that one out.

src/ParquetMsg.chpl
bmcdonald3: Actually, we need this, the loop requires that we know the number of row groups, which we don't know until the first one has been computed.

src/ParquetMsg.chpl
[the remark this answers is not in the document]
stress-tess: oh! my bad I wasn't paying close enough attention! but yah ur right, it shouldn't matter either

src/ParquetMsg.chpl
stress-tess: ahhhh okay! that makes sense

src/ParquetMsg.chpl
stress-tess: ah! I think the biggest misunderstanding here was me being real dumb and not reading the name of `entrySeg` lol. I thought that was the vals 🤦‍♀️  this makes sense now, thanks ben!

stress-tess (approved):

jaketrookman (approved):
Looks good

stress-tess:
awesome @bmcdonald3 it looks like this will be good to go once the merge conflicts are resolved (sorry I messed u up by merging mine first 😅

bmcdonald3:
@stress-tess should be good to go

stress-tess (approved):

Issue #3088, opened 2024-04-08
failure in `test_sparse_sum_helper`


stress-tess:
@bmcdonald3 reported a failure in `test_sparse_sum_helper` during their nightly run of our tests with the error:
```
FAILED tests/util_test.py::UtilTest::test_sparse_sum_helper - ValueError: shape mismatch [1271] [1269]
```
in my experience this is due to some off by ones in the index math

unfortunately this relies on `ak.randint` without a set seed, so reproducing this might be difficult to track down. I should definitely add a randomly generated set seed to this test so any future failures will be easier to reproduce

We should also consider having the sort based workflow be the default and have the merge based be opt in

stress-tess:
I'm really struggling to reproduce this. I've modified the test to have a randomly selected set seed:
```python
def test_sparse_sum_helper(self):
    seed = np.random.default_rng().choice(2**63)
    ...
    def are_pdarrays_equal(pda1, pda2):
        # we first check the sizes so that we won't hit shape mismatch
        # before we can print the seed (due to short-circuiting)
        return (pda1.size == pda2.size) and ((pda1 == pda2).all())

    cond = (
        are_pdarrays_equal(merge_idx, sort_idx)
        and are_pdarrays_equal(merge_idx, gb_idx)
        and are_pdarrays_equal(merge_vals, sort_vals)
    )

    if not cond:
        print(f"Failure with seed = {seed}")
    self.assertTrue(cond)
```

I've verified that when I intentionally modify one of the arrays I'm testing to have different values that it fails and prints out the seed. It will look like the following:
```
----------------------------------------------- Captured stdout call -----------------------------------------------
Failure with seed = 3875030313406367833
```

So now I just have to find a seed that fails. I changed the `ARKOUDA_RUNNING_MODE` in `pytest.ini` to be `CLIENT`, so I can run the test with a multi-locale server. I then used the following bash loop to run the test over and over:
```
for run in {1..25}; do python3 -m pytest tests/util_test.py::UtilTest::test_sparse_sum_helper; don
```
I've tried this with different numbers of locales, but I still can't get  a failure

bmcdonald3:
Interesting... It only popped up for one night on our nightly test and then resolved, so I'd say we can push this to the side for now and I'll let you know if it pops up with any consistency. Thanks for looking into it so quick!

stress-tess:
[...]

PR #3090, opened 2024-04-08
Part of #3088: Generate seed for sparse sum test

stress-tess:
This PR is more of a bandaid since I can't reproduce the error in #3088. I added randomly generated seeds that get printed to stdout when the test fails. This will make it easier to reproduce when it pops back up in the nightly

In the meantime I switched the sort workflow to be the default, so the merge workflow is opt in

More info about how I went about trying to reproduce it can be found in https://github.com/Bears-R-Us/arkouda/issues/3088#issuecomment-2043527577

### DIFF
--- a/arkouda/util.py
+++ b/arkouda/util.py
@@ -303,7 +303,7 @@ def attach_all(names: list):
     return {n: attach(n) for n in names}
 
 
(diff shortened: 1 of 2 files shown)


bmcdonald3 (approved):
Thanks tess!

jaketrookman (approved):

Issue #3087, opened 2024-04-08
Failure reading `int64` column with null values with parquet


stress-tess:
a user encountered this issue when trying to read a dataset from the [nyc tax cab data](https://data.cityofnewyork.us/browse?limitTo=datasets&q=yellow+taxi+trip+data&sortBy=relevance):
```python
ak.read_parquet('yellow_tripdata_2023-02.parquet')
```
resulting in: 
```
RuntimeError: 1 errors: Error: ParquetError 142 coforall_fn:ParquetMsg Unexpected end of stream
```

The user thinks this is due to `nan`s being present in some of the columns and that it's related to https://github.com/Bears-R-Us/arkouda/issues/2005

stress-tess:
At first I thought it was our handling of `int32s`. Because reading the file with pandas and attempting to convert to an arkouda dataframe:
```python
ak.DataFrame(pd.read_parquet('yellow_tripdata_2023-02.parquet'))
```
gives
```
RuntimeError: Unhandled data type int32
```

Which was handled in https://github.com/Bears-R-Us/arkouda/pull/2333
```python
In [6]: df_pd = pd.DataFrame({
   ...: "int16": pd.Series([2**15-1, -2**15], dtype=np.int16),
   ...: "int32": pd.Series([2**31-1, -2**31], dtype=np.int32),
   ...: "uint16": pd.Series([2**15-1, 2**15], dtype=np.uint16),
   ...: "uint32": pd.Series([2**31-1, 2**31], dtype=np.uint32),
   ...: })
   ...: fname = 'pd_test'
   ...: df_pd.to_parquet(fname)
   ...: df_ak = ak.DataFrame(ak.read(fname+"*"))

In [7]: df_pd
Out[7]:
   int16       int32  uint16      uint32
0  32767  2147483647   32767  2147483647
1 -32768 -2147483648   32768  2147483648

In [8]: df_ak
Out[8]:
   int16       int32  uint16      uint32
0  32767  2147483647   32767  2147483647
1 -32768 -2147483648   32768  2147483648 (2 rows x 4 columns)
```
But it turns out that our initializer just can't handle smaller int types right now

```python
In [9]: ak.DataFrame(df_pd)
RuntimeError: Unhandled data type int16
```

stress-tess:
A little more digging makes me think it's failing on `'passenger_count'`, which does have `nan`s

```python
In [6]: pddf['passenger_count']
Out[6]:
0          2.0
1          1.0
[...]

PR #3091, opened 2024-04-09
Fix Parquet double reads to properly account for null values

bmcdonald3:
In the Parquet taxi cab dataset, we were hitting end of stream errors when reading in null values and this was identified by Tess in https://github.com/Bears-R-Us/arkouda/issues/3087.

To fix this, we switch to reading in single values in the case where nulls are allowed so that we can check the definition level before inserting a value into the Parquet array.

Potential optimizations are an `allow_nan` flag to avoid the check and then, additionally, we could modify the previous implementation by reading the definition level and values into temporary vectors and then only assign the value if the value is non-null.

The reason it didnt' work before was that it was bsaed off of `batchSize` to loop over the values, but batch size doesn't actually guarantee how many values were read, due to row group intricacies.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -847,36 +847,34 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
           delete[] rep_lvl;
         }
       } else if(ty == ARROWDOUBLE) {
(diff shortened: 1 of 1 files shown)


stress-tess (approved):
Way simpler and fixes the bug I was seeing! Thanks for take care of this so quickly!

stress-tess (approved):

PR #3093, opened 2024-04-10
Pdarray indexing

jeremiah-corrado:
Implement multidimensional pdarray indexing. 

For example, the following indexing patterns now match numpy's behavior, where `A` is a 3D pdarray and `a` is a 1D pdarray:

```python
x = A[a, a, a]
x = A[a, a, 1]
x = A[a, a, :]
x = A[a, 1, 1]
x = A[a, :, 1]
x = A[a, ...]
```

Also implements rank expansion via 'None' indices. For example, the following would return a 4D array, where the 3rd dimension has a size of 1:

```python
x = A[1, :, None, a]
```

*also fixes a bug in unstackMsg


src/IndexingMsg.chpl
```diff
@@ -380,9 +380,147 @@ module IndexingMsg
...
+                return new MsgTuple(errorMsg,MsgType.ERROR);
+            }
+        }
+    }
+
+    private proc multiIndexShape(inShape: ?N*int, idxDims: [?d] int, outSize: int): (bool, int, N*int) {
+        var minShape: N*int = inShape,
+            firstRank = -1;
+
+        if d.size > N then return (false, -1, minShape);
```
bmcdonald3, line 497: It seems like this `-1` is conveying the same information as `false`. I don't know if we need the additional `bool` value, but I also don't think it's hurting anything, so don't need to change. Am I missing something here though?

src/ManipulationMsg.chpl
```diff
@@ -1063,36 +1063,36 @@ module ManipulationMsg {
...
-        const repMsg = try! '+'.join([rn in rnames] "created " + st.attrib(rn));
-        mLogger.info(getModuleName(),pn,getLineNumber(),repMsg);
-        return new MsgTuple(repMsg, MsgType.NORMAL);
+      // copy the data from the input array to the output arrays while unstacking
+      for arrIdx in 0..<numReturnArrays {
+        forall idx in eOuts[arrIdx].a.domain with (
+          var agg = newSrcAggregator(t),
+          const imap = new unstackIdxMapper(nd, arrIdx, axis)
+        ) {
+          const inIdx = imap(if nd == 2 then (idx,) else idx);
```
bmcdonald3, line 1088: is this hardcoded 2 what we want or do we want like greater than 1? (Don't understand enough to know, just an actual question)

bmcdonald3 (approved):
Looks good, couple of questions.

src/IndexingMsg.chpl
jeremiah-corrado: There are cases where the second value in the tuple can be greater than `-1`, but the first value is still `false`. For example, if one of the values in `idxDims` is out of bounds, but `firstRank` has already been set, the function could return `(false, 1, ...)`.

This did make me realize that I should be putting the variable `firstRank` in that tuple here instead of the hard-coded `-1`, so I fixed that.

src/ManipulationMsg.chpl
jeremiah-corrado: Yeah it's supposed to be a `2` because when `nd==1`, the output array's domain will have a rank of `nd-1`. So this is the same thing as saying `if eOuts[arrIdx].a.domain.rank == 1 then (idx, ) else idx`. I couldn't decide which was less confusing, so lmk if you think it should be changed.

bmcdonald3 (approved):

PR #3094, opened 2024-04-10
Fixes #3087: Failure reading non-float parquet columns with null values

### LINKED ISSUE #3087: Failure reading `int64` column with null values with parquet

stress-tess:
This PR fixes #3087, a bug encountered by user while reading the nyc taxi cab data. The parquet file contained an `int64` column with null values. Pandas handles this by converting into a `float64` column and replaces the null values with `NaN`.

Properly handling null values is slower than the regular workflow because we can't read in batches. Since non-float columns having nulls is a fairly rare case, I decided to default to the faster way so the common case wouldn't take a performance hit. In the rare case, we fail with an informative error message that specifies which flag needs to be set to handle the null values. Rerunning with that flag behaves correctly and returns an array that matches pandas

```python
>>> pddf = pd.read_parquet('/Users/pierce/Downloads/yellow_tripdata_2023-02.parquet')
>>> akdf = ak.DataFrame(ak.read_parquet('/Users/pierce/Downloads/yellow_tripdata_2023-02.parquet'))

RuntimeError: ParquetError 145 readFilesByName:ParquetMsg Unexpected end of stream. This may be due to null values in a non-float column. Try again with the flag has_non_float_nulls=True

>>> akdf = ak.DataFrame(ak.read_parquet('/Users/pierce/Downloads/yellow_tripdata_2023-02.parquet', has_non_float_nulls=True))

>>> akdf['passenger_count']
array([2.00000000000000000 1.00000000000000000 1.00000000000000000 ... nan nan nan])

>>> np.allclose(pddf['passenger_count'].to_list(), akdf['passenger_count'].to_list(), equal_nan=True)
True
```

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -721,7 +721,7 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
   }
 }
 
-int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colname, int64_t numElems, int64_t startIdx, int64_t batchSize, int64_t byteLength, char** errMsg) {
+int cpp_readColumnByName(const char* filename, void* chpl_arr, bool* where_null_chpl, const char* colname, int64_t numElems, int64_t startIdx, int64_t batchSize, int64_t byteLength, bool hasNonFloatNulls, char** errMsg) {
   try {
     int64_t ty = cpp_getType(filename, colname, errMsg);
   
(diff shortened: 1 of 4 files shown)


src/ArrowFunctions.cpp
```diff
@@ -756,45 +756,92 @@ int cpp_readColumnByName(const char* filename, void* chpl_arr, const char* colna
...
         parquet::Int64Reader* reader =
           static_cast<parquet::Int64Reader*>(column_reader.get());
         startIdx -= reader->Skip(startIdx);
 
-        while (reader->HasNext() && i < numElems) {
-          if((numElems - i) < batchSize)
-            batchSize = numElems - i;
-          (void)reader->ReadBatch(batchSize, nullptr, nullptr, &chpl_ptr[i], &values_read);
-          i+=values_read;
+        if (not hasNonFloatNulls) {
```
bmcdonald3, line 765: Wow, I did not know that C++ supported `not`!

src/ParquetMsg.chpl
```diff
@@ -820,45 +824,76 @@ module ParquetMsg {
...
-          readFilesByName(entryVal.a, filenames, sizes, dsetname, ty);
+          readFilesByName(entryVal.a, whereNull, filenames, sizes, dsetname, ty, hasNonFloatNulls=hasNonFloatNulls);
           var valName = st.nextName();
-          st.addEntry(valName, entryVal);
+          if hasNonFloatNulls && (|| reduce whereNull) {
+            // if we have non-float nulls and there's at least one null
+            var floatEntry = createSymEntry(len, real);
+            floatEntry.a = (entryVal.a):real;
+            ref fa = floatEntry.a;
+            [(t, f) in zip(whereNull, fa)] if t then f = nan;
```
bmcdonald3, line 839: I am confused by this, aren't we setting the null values in the C++ code?

stress-tess: ah yah we do but since they're not float arrays, it ends up being `0`s. I should just remove that line setting it to nan in c++ code cause it isn't doing anything and it's just confusing

bmcdonald3 (approved):
Null flag was a good idea! One alternative would be, rather than modifying the error message, we could see if that particular error is thrown and then, if it is, run the nan code automatically. It would read twice in that case though, so I'm not sure what the best decision is.

Also, I think that we can do some optimizations here, but probably best left for another PR. With strings, we have to worry about data being cleaned up because it is memory being allocated by Arrow, but since all our memory here is allocated in Chapel, we don't have that problem.

The optimization would to be to read in batches into a third temporary array and then do the assignment in parallel on the Chapel side of things.

bmcdonald3 (approved):
Rejoice for Parquet now reads taxi data!

jaketrookman (approved):

stress-tess:
> Null flag was a good idea! One alternative would be, rather than modifying the error message, we could see if that particular error is thrown and then, if it is, run the nan code automatically. It would read twice in that case though, so I'm not sure what the best decision is.
> 
> Also, I think that we can do some optimizations here, but probably best left for another PR. With strings, we have to worry about data being cleaned up because it is memory being allocated by Arrow, but since all our memory here is allocated in Chapel, we don't have that problem.
> 
> The optimization would to be to read in batches into a third temporary array and then do the assignment in parallel on the Chapel side of things.

I think these are both great ideas!! I think we should def look into them. I created issue https://github.com/Bears-R-Us/arkouda/issues/3095 to capture these. I think for the time being I'll go ahead and merge this as a stop gap

PR #3101, opened 2024-04-17
Closes #3020 dataframe.dropna

### LINKED ISSUE #3020: DataFrame.dropna
Create a new function to replicate the capabilities of pandas DataFrame.dropna:
https://pandas.pydata.org/docs/dev/reference/api/pandas.DataFrame.dropna.html

ajpotts:
Adds the following DataFrame functions:
dropna
all
any
isna
notna

closes #3020

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -4155,6 +4155,7 @@ class DataFrame(UserDict):
             return Series(array(count_values_list), index=Index(array(index_values_list)))
         elif (isinstance(axis, int) and axis == 1) or (isinstance(axis, str) and axis == "columns"):
             first = True
+            count_values = arange(0)
             for col in self.columns:
                 if is_numeric(self[col]):
                     if first:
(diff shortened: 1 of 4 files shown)


arkouda/dataframe.py
```diff
@@ -4402,6 +4406,480 @@ def merge(
...
+        if (isinstance(mask, bool) and mask is True) or (
+            isinstance(mask, Series) and akall(mask.values) is True
+        ):
+            result = self.copy(deep=None)
+        else:
+            if (isinstance(axis, int) and axis == 0) or (isinstance(axis, str) and axis == "index"):
+                if self.empty is True:
+                    result = DataFrame()
+                else:
+                    result = self[mask].copy(deep=True)
```
stress-tess, line 4870: I noticed that we have both deep and shallow copies in this one. Is there a reason to not be consistent in how we're copying?

also at some point do we want to try and add an `inplace` flag like in the pandas version? I haven't thought through it, so I'm mostly asking if you ran into problems adding it or if it was left out to get this functionality added sooner

stress-tess (approved):
Looks great! Super thorough test! I really like the commutativity of the methods with `to_pandas`

my only question is about deep vs shallow copies for `dropna`, but I don't think it needs to hold this up

arkouda/dataframe.py
ajpotts: Admittedly that could be more consistent.  I was going to look into it further when I added the `inplace` flag, which I thought would be easier once Brandon's indexing updates were merged in.  I've already made tickets for the ways we could better align with pandas:
https://github.com/Bears-R-Us/arkouda/issues/3104
https://github.com/Bears-R-Us/arkouda/issues/3105
https://github.com/Bears-R-Us/arkouda/issues/3106
https://github.com/Bears-R-Us/arkouda/issues/3107

stress-tess: okay awesome!

PR #3109, opened 2024-04-18
Dataframe Indexing

brandon-neth:
This pull request (re)implements indexing methods for DataFrames to match the behavior of Pandas indexing. Specifically, the `[]` indexers now index columns by label, `loc` and `at` index rows by label, and `iloc` and `iat` index rows by position. It also updates uses of DataFrames throughout the repository to match these new behaviors.


ajpotts (commented):
Here are my comments so far.  Thanks for all this great work. I know it was a lot!

tests/dataframe_test.py
```diff
@@ -1442,5 +1437,667 @@ def test_to_markdown(self):
...
+            with self.assertRaises(TypeError):
+                if isinstance(keys[0], int):
+                    a = df['int']
+                else:
+                    a = df[3]
+            with self.assertRaises(TypeError):
+                b = df[1.0]
+
+        # slices
+        def slice_check(_slice_access, slice_access):
```
ajpotts, line 1554: I'm not going to comment on every instance, but in general, these checks can be more succinct using a combination of `.to_pandas`, `assert_frame_equal`, and `assert_series_equal`.

arkouda/dataframe.py
```diff
@@ -1026,6 +1217,69 @@ def _get_head_tail_server(self):
...
+        # To stay consistent with numpy, provide the old index values
+        return DataFrame(initialdata=result, index=self.index.index[key])
+
+    def column_labels(self):
+        """
+        Return the column labels.
+        """
+        return self._columns
+
+    def add_new_rows(self, key):
```
ajpotts, line 1246: Every function should have a docstring, if it is public.

ajpotts (changes requested):
This is such a huge amount of code and I am really looking forward to using these new features!  Thank you so much for this!  Please look at my comments and let me know what you think.

arkouda/series.py
```diff
@@ -252,15 +262,25 @@ def validate_key(
...
+                "Series [] only supports indexing by scalars, lists of scalars, "
+                "and arrays of scalars. Received {}".format(type(key))
             )
         return key
 
     @typechecked
-    def __getitem__(self, _key: Union[supported_scalars, pdarray, Strings, List]):
+    def __getitem__(self, _key: Union[supported_scalars, pdarray, Strings, List, SegArray, Series]):
         """
         Gets values from Series.
```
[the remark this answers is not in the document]
brandon-neth: No you're right! Good catch. I think I added it when updating the initializer.

arkouda/dataframe.py
```diff
@@ -753,75 +758,280 @@ def __delitem__(self, key):
...
+                raise TypeError(f"Expected key of type {type(self._columns[0])}, received {type(key)}")
+            return key
+
+        if isinstance(key, pdarray) and key.dtype == akbool:
+            if len(key) != len(self):
+                raise ValueError("Boolean mask arguments must have the same length as the DataFrame.")
+            return key
+
+        if isinstance(key, (pdarray, Strings, Categorical, SegArray)):
+            for k in key.to_ndarray():
```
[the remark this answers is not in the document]
brandon-neth: Good point. I'll change this to get the first element of the key list and use that instead.

arkouda/dataframe.py
```diff
@@ -753,75 +758,280 @@ def __delitem__(self, key):
...
+            return self.validate_key(key.values)
+        if isinstance(key, list):
+            return self.validate_key(array(key))
+
+        if isinstance(key, slice):
+            if key.start is not None and key.start < 0:
+                raise ValueError("Negative start index not supported")
+            if key.stop is not None and key.stop > len(self):
+                raise ValueError("Slice stop index out of range")
+            return key
```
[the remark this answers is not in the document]
brandon-neth: I think just distance between the implementations. I'll do the same here for uniformity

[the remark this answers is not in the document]
brandon-neth: On second thought, it's because slice indexing behavior is different from array indexing behavior for dataframes. slices do rows while arrays do columns. This maintains that difference so it can be used to distinguish in the body of `__getitem__`

arkouda/dataframe.py
```diff
@@ -1026,6 +1217,69 @@ def _get_head_tail_server(self):
...
+            key = arange(start, stop, step)
+        if key.dtype == akbool:
+            key = arange(key.size)[key]
+        result = {}
+        for k in self._columns:
+            result[k] = UserDict.__getitem__(self, k)[key]
+        # To stay consistent with numpy, provide the old index values
+        return DataFrame(initialdata=result, index=self.index.index[key])
+
+    def column_labels(self):
```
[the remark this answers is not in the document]
brandon-neth: Gotcha! I'll replace the uses.

ajpotts (approved):
Thanks for all your work on this!  Super excited to start using it :)

arkouda/dataframe.py
```diff
@@ -3694,7 +3981,7 @@ def sort_values(self, by=None, ascending=True):
             i = self.coargsort(by, ascending=ascending)
         else:
             raise TypeError("Column name(s) must be str or list/tuple of str")
-        return self[i]
+        return self._get_rows(i)
```
stress-tess, line 3984: I was just wondering why we switched to using `_get_rows` here when most of the other cases like this use `iloc`

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+            return self._get_row_col(key[0], key[1])
+        if isinstance(key, list):
+            key = array(key)
+        if isinstance(key, Series):
+            key = key.values
+        if is_supported_scalar(key) and self.df.index.dtype == dtype(type(key)):
+            return self.df._get_rows(indexof1d(array([key]), self.df.index.values))
+
+        if isinstance(key, pdarray) and key.dtype == self.df.index.dtype:
+            return self.df._get_rows(indexof1d(key, self.df.index.values))
```
stress-tess, line 5627: do we expect there to be multiple occurrences of `key`/the values of `key` in `self.df.index.values`? if not then it would be more efficient to use `ak.find` rather than `indexof1d`

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+
+        if isinstance(key, slice):
+            if key.start is not None and not akany(in1d(array([key.start]), self.df.index.values)):
+                raise KeyError(f"Index {key.start} not found in DataFrame index")
+            if key.stop is not None and not akany(in1d(array([key.stop]), self.df.index.values)):
+                raise KeyError(f"Index {key.stop} not found in DataFrame index")
```
stress-tess, line 5633:
```suggestion
            if key.start is not None and find(array([key.start]), self.df.index.values) != -1:
                raise KeyError(f"Index {key.start} not found in DataFrame index")
            if key.stop is not None and find(array([key.stop]), self.df.index.values) != -1:
                raise KeyError(f"Index {key.stop} not found in DataFrame index")
```
similarly `find` would be faster here. `find` returns `-1` if the index is not found

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+
+            start_idx = (indexof1d(array([key.start]), self.df.index.values)[0]
+                         if key.start is not None else 0)
+            stop_idx = (indexof1d(array([key.stop]), self.df.index.values)[0] + 1
+                        if key.stop is not None else self.df.index.size)
```
stress-tess, line 5638: okay im sorry if im starting to sound like a broken record lol, but if we only want the first occurrence (which it seems we do since we're indexing at `0`, `find` is the way to go.

```python
start_idx = (find(array([key.start]), self.df.index.values) if key.start is not None else 0)
stop_idx = (find(array([key.stop]), self.df.index.values) + 1 if key.stop is not None else self.df.index.size)
```

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+            return self._set_row_col_scalar_val(row_key, col_key, val)
+        else:
+            assert (isinstance(val, (pdarray, Series, Strings, SegArray))), "Invalid value type"
+            return self._set_row_col_vector_val(row_key, col_key, val)
+
+    def _set_row_col_scalar_val(self, row_key, col_key, val):
+        if is_supported_scalar(row_key):
+            if not self.df.index.dtype == dtype(type(row_key)):
+                raise TypeError("Row key must be of the same type as the DataFrame index")
+            if akany(in1d(array([row_key]), self.df.index.values, invert=True)):
```
stress-tess, line 5678:
```suggestion
            if ak.find(ak.array([5row_key), self.df.index.values) != -1:
```

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+                self.df._add_new_rows(row_key)
+            # updating a single row
+            row_idx = indexof1d(array([row_key]), self.df.index.values)
```
stress-tess, line 5681: if `self.df.index.values` is unique, then `find` is faster. I'm prob not gonna comment on it anymore just cause I don't want to spam, but I think it's worth going through and thinking about which is better for the all places we call `indexof1d`

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+            # updating multiple rows
+            row_idx = indexof1d(row_key, self.df.index.values)
+            self.df.data[col_key][row_idx] = val
+
+        if isinstance(row_key, pdarray) and row_key.dtype == akbool:
+            self.df.data[col_key][row_key] = val
+        if isinstance(row_key, slice):
+            if row_key.start is not None and not (in1d(array([row_key.start]), self.df.index.values)):
+                raise ValueError(f"Index {row_key.start} not found in DataFrame index")
+            if row_key.stop is not None and not akany(in1d(array([row_key.stop]), self.df.index.values)):
```
stress-tess, line 5699: similar logic to an above comment. and it seems to pop up in handful of other places as well

arkouda/series.py
```diff
@@ -252,15 +262,25 @@ def validate_key(
...
+        elif isinstance(key, slice):
+            start = key.start if key.start is not None else 0
+            stop = key.stop if key.stop is not None else self.size
+            if start < 0:
+                raise IndexError("Slice start must be non-negative")
+            if stop > self.size:
+                raise IndexError("Slice stop must be less than or equal to the size of the Series")
+            if start > stop:
+                raise IndexError("Slice start must be less than or equal to the stop")
+            key = arange(start, stop)
```
stress-tess, line 274: just checking if we want `stride` to be supported

arkouda/series.py
```diff
@@ -1596,8 +1631,19 @@ def validate_key(self, key):
...
+        elif isinstance(key, slice):
+            start = key.start if key.start is not None else 0
+            stop = key.stop if key.stop is not None else self.series.size
+            if start < 0:
+                raise IndexError("Slice start must be non-negative")
+            if stop > self.series.size:
+                raise IndexError("Slice stop must be less than or equal to the size of the Series")
+            if start > stop:
+                raise IndexError("Slice start must be less than or equal to the stop")
+            key = arange(start, stop)
```
stress-tess, line 1643: same stride question

stress-tess (changes requested):
Great work overall! This must've been a huge lift! I'm glad we're finally going to have indexing consistent with pandas!

Just a few small gotchas. honestly half of comments are just suggesting `find` instead of `indexof1d` when possbile lol, I stopped commenting that at some point but prob should've stopped sooner...

I didn't look at the tests yet, but I think there's enough here to go ahead and post a review. and I didn't look at the indexers terribly closely after `_LocIndexer` because the logic seemed pretty similar

also is there an associated issue with this PR? if so we should link it

EDIT:
I realized once my PR gets merged some of comments about `find` being faster than `indexof1d` aren't relevant anymore because `indexof1d` would be equivalent to running `find` if the search space is unique... bc it only takes the slower code path if `all_occurrences` is True and there are duplicates in the search space. I def should've realized this sooner considering I was the one to write that code but ya know how it goes

 I went ahead and resolved the affected comments. The comments about using `find` instead of `akany(in1d(...`  and using `find` instead of `indexof1d(...)[0]` still hold

arkouda/dataframe.py
```diff
@@ -5326,6 +5609,322 @@ def from_return_msg(cls, rep_msg):
...
+    def _set_row_col_scalar_val(self, row_key, col_key, val):
+        if is_supported_scalar(row_key):
+            if not self.df.index.dtype == dtype(type(row_key)):
+                raise TypeError("Row key must be of the same type as the DataFrame index")
+            if akany(in1d(array([row_key]), self.df.index.values, invert=True)):
+                self.df._add_new_rows(row_key)
+            # updating a single row
+            row_idx = indexof1d(array([row_key]), self.df.index.values)
+            if row_idx.size == 0:
+                raise ValueError(f"Index {row_key} not found in DataFrame index")
```
stress-tess, line 5683: also this logic depends on the size of the return being `0` when the key is not found. If my PR gets merged which updates `indexof1d` to use `find`, this will no longer be the case and it will return `-1` instead. so if excluding indices that are not found is the desired behavior, then I need to update https://github.com/Bears-R-Us/arkouda/pull/3169

```python
>>> ak.find(ak.array([7]), ak.arange(5))
array([-1])

>>> ak.indexof1d(ak.array([7]), ak.arange(5))
array([])
```

arkouda/dataframe.py
```diff
@@ -829,74 +834,280 @@ def __delitem__(self, key):
             self._empty = True
         self.update_nrows()
 
-    def __getitem__(self, key):
-        # convert series to underlying values
-        # Should check for index alignment
+    def validate_key(self, key):
```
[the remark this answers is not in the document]
brandon-neth: I think we should; I lean towards making it a separate issue to get this into the code base first though

arkouda/dataframe.py
```diff
@@ -829,74 +834,280 @@ def __delitem__(self, key):
...
+                result._columns.append(k)
+            result._empty = False
+            result._set_index(self.index)
+            return result
+
+        raise TypeError("key not supported: {}".format(key))
+
+    def validate_value(self, value):
+        if isinstance(value, Series):
+            # TODO: check index alignment
```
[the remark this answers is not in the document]
brandon-neth: Index alignment was something that was a TODO when I started and I left it as a TODO here for another issue

arkouda/dataframe.py
```diff
@@ -829,74 +834,280 @@ def __delitem__(self, key):
...
+
+        # boolean mask
+        if isinstance(key, pdarray) and key.dtype == akbool:
+            return self._get_rows(key)
+
+        # slice
+        if isinstance(key, slice):
+            start_index = key.start if key.start is not None else 0
+            stop_index = key.stop if key.stop is not None else len(self.index)
+            return self._get_rows(arange(start_index, stop_index, 1))
```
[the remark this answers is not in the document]
brandon-neth: Good catch. I was approaching it with strided indexing being a future work issue, but its a light lift to get it supported here.

arkouda/dataframe.py
```diff
@@ -829,74 +834,280 @@ def __delitem__(self, key):
...
+
+            UserDict.__setitem__(self, key, value)
+            return
+
+        if isinstance(key, pdarray) and key.dtype == akbool:
+            # TODO: validate length and width of value matches number of entries needed
+            if not isinstance(value, DataFrame):
+                raise ValueError("Expected DataFrame type for boolean mask assignment")
+            if len(intersect1d(self.columns, value.columns)) != 0:
+                raise ValueError("Right-hand side columns do not match left-hand side columns")
```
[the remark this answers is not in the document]
brandon-neth: Nope!

arkouda/dataframe.py
brandon-neth: Only change is that we're looking for `== -1` because this error triggers when the start or stop aren't there.

arkouda/dataframe.py
brandon-neth: That is the desired behavior! The return of indexof1d should just be a list of indices, potentially empty.

stress-tess: sweet! I'll get that updated! I also want to note that will change how we want to handle these type of checks:
https://github.com/Bears-R-Us/arkouda/pull/3109#discussion_r1599199235

which should be small change so I can handle that as part of the indexof1d update!

arkouda/dataframe.py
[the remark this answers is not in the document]
brandon-neth: I've removed the uses either way so it should be okay!

arkouda/dataframe.py
brandon-neth: you are correct, we want the first occurrence.

stress-tess (approved):
looks good!! Thanks brandon!

PR #3117, opened 2024-04-24
Closes #3115: Fix non-deterministic sparse_sum failure

### LINKED ISSUE #3115: non-deterministic sparse sum failure
The sparse sum helper failed the nightly with the following:
```
num locales: 2
Failure with seeds:
8086790153783974714,
2380734683647922470,
2906987507681887800,
4967208342496478642
```

Even with runtime checks enabled I wasn't able to reproduce until I made an address sanitizer (`asan`) build. And even then it seems to be non-deterministic and sometimes passes. This leads me to believe there is a race condition

stress-tess:
This PR (closes #3115) fixes the non-determinstic `sparse_sum_helper` failure. 

I was seeing sporadically runs where both the `if` and the `if else` would execute. Since the fetch conditions are mutually exclusive, we can just use an `if` for both. This seems to avoid the issue

I also didn't realize that `on` statements were non-blocking, so there's potential for them to execute concurrently on different locales. Since my logic requires the previous locale to finish before the next begins, I've added an atomic counter to enforce this

Thanks @jeremiah-corrado for the atomic counter logic! And everyone else for putting up with me while I tried to figure out what was going on 😅


src/ArraySetops.chpl
```diff
@@ -267,8 +275,8 @@ module ArraySetops
             twoArrayRadixSort(tmp, new KeysRanksComparator());
 
             const writeToResultSlice = (segs[here.id]+numSendBack)..#tmp.size;
-            sortedIdx[writeToResultSlice] = [(key, val) in tmp] key;
-            permutedVals[writeToResultSlice] = [(key, val) in tmp] val;
+            sortedIdx[writeToResultSlice] = [(key, _) in tmp] key;
+            permutedVals[writeToResultSlice] = [(_, val) in tmp] val;
```
ajpotts, line 279: Is there any advantage in using `_` here?

src/ArraySetops.chpl
```diff
@@ -148,8 +148,12 @@ module ArraySetops
 
       var aMin = -1, aMax = -1, bMin = -1, bMax = -1;
       // we iterate over locales serially because the logic relies on the mins and maxs referring to the previous locale
-      for loc in Locales {
-        on loc {  // perform loop body computation on locale #loc
+      // use an atomic counter to ensure the previous locale finishes before the next begins (on statements are non-blocking)
+      var count: atomic int;
```
ajpotts, line 152: You could also use a sync variable :)

ajpotts (approved):
Good to know about this issue!  Thanks for figuring it out!  It will definitely help to be aware of this going forward.

src/ArraySetops.chpl
stress-tess: It's just stylistic! It makes it easier to tell at a glance which looping variables are actually being used

this was one of the style suggestions I got from the `chpl_check` linter

src/ArraySetops.chpl
stress-tess: I did it with a sync var at first! And I think they should be equivalent. But I do like the explicitness of each locale waiting for the counter to hit their number before beginning as a guarantee that they happen both serially and in order

jaketrookman (approved):
Looks good

PR #3138, opened 2024-04-25
Closes #3118: Move `choice` implementation into arkouda

### LINKED ISSUE #3118: Add choice implementation to arkouda
in PR #3114 we encountered a subtle bug edge case with `choice` in chpl versions before `2.0`. We decided to skip the test until a correct implementation could could be added to arkouda

stress-tess:
This PR closes #3118 by moving the implemention of `choice` into arkouda for chpl versions prior to 2.0. This allows us to no longer skip `test_choice_flags`.

We could have done this as part of #3114, since it was much less involved than I thought it would be. I was expecting to need to copy several internal procs and records from chpl's random modules, but it more or less just worked to copy it over and change `sample` to `stream.sample`

### DIFF
--- a/src/compat/e-132/ArkoudaRandomCompat.chpl
+++ b/src/compat/e-132/ArkoudaRandomCompat.chpl
@@ -32,7 +32,70 @@ module ArkoudaRandomCompat {
     }
 
     proc ref sample(d: domain(?), n: int, withReplacement = false): [] d.idxType throws  where is1DRectangularDomain(d) {
-      return r.choice(d, n, withReplacement);
+      return choiceUniform(r, d, n, withReplacement);
+
+      /* _choice branch for uniform distribution */
+      proc choiceUniform(stream, X: domain(?), size: ?sizeType, replace: bool) throws
+      {
+        use Set;
+        use Math;
+
+        const low = X.low,
+              stride = abs(X.stride);
+
+        if isNothingType(sizeType) {
+          // Return 1 sample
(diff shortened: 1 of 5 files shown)


src/compat/e-132/ArkoudaRandomCompat.chpl
```diff
@@ -32,7 +32,70 @@ module ArkoudaRandomCompat {
     }
 
     proc ref sample(d: domain(?), n: int, withReplacement = false): [] d.idxType throws  where is1DRectangularDomain(d) {
-      return r.choice(d, n, withReplacement);
+      return choiceUniform(r, d, n, withReplacement);
+
+      /* _choice branch for uniform distribution */
+      proc choiceUniform(stream, X: domain(?), size: ?sizeType, replace: bool) throws
```
ajpotts, line 38: Why doesn't `stream` need an intent?

ajpotts (approved):
Looks OK to me!  I can't verify the chapel data types, but the logic looks good.

src/compat/e-132/ArkoudaRandomCompat.chpl
stress-tess: by leaving off a type intent, it's generic. So technically anything could be passed in as the `stream` variable

Since it's only ever called by line 35, we could give it the intent `: shared PCGRandomStream(eltType)`

The reason I didn't was because it didn't have a type intent in the chpl sorce code for the `1.32` release, so I wanted to make minimal modifications to fix the bug

https://github.com/chapel-lang/chapel/blob/1.32.0/modules/standard/Random.chpl#L328

PR #3147, opened 2024-04-29
Closes #3116 remove DataFrame._columns

### LINKED ISSUE #3116: remove DataFrame._columns
Remove `DataFrame._columns` and use `DataFrame.keys()` and `DataFrame.columns.values` as a replacement.  This will eliminate redundant storage that needs to be managed and could become out of date.

ajpotts:
Closes #3116 remove DataFrame._columns

Having researched this, the only reason to keep the `._columns` attribute is to preserve the order.  The downside is that it could cause inconsistencies if accessed outside the `__setitem__ `method.  To remove this attribute and also preserve the column order we could make `DataFrame` an `OrderedDict` rather than a `UserDict`.  However, since the two classes have different APIs it would be a larger refactor.  

This change implements short term fix to remove instances where `._columns` is modified outside of the `__setitem__` method.

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -685,9 +685,7 @@ class DataFrame(UserDict):
                     if len(sizes) > 1:
                         raise ValueError("Input arrays must have equal size.")
                     self._empty = False
-                    UserDict.__setitem__(self, key, val)
-                    # Update the column index
-                    self._columns.append(key)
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):

stress-tess (approved):
Looks good to me!

PR #3151, opened 2024-05-02
3130 numeric test slight revamp

### LINKED ISSUE #3130: Update `numeric_test`

drculhane:
This does several things.

Closes #3130

Fixes all the numeric tests that were failing.

Revamps trigonometric and hyperbolic tests, in the same fashion as the _trig_test_helper did (which is now _trig_and_hyp_test_helper).  Adds an _infinity_edge_case_helper.

The only trig or hyp function not included is arctan2, since it needs two inputs (numerator and denominator).  Adds a denominator array with no zeroes for these cases.  There are already zero denominator tests as special cases at the end of test_arctan2, so zero denominators aren't needed in the more general tests.

Includes seed in every test that uses random numbers, so that results will be reproducible.

Combines log and exp tests into one function, because they were identical.

Combines cumsum and cumprod for the same reason.

Sets "ignore" on overflow, invalid data, and divide-by-zero in tests where that info is accurate but unhelpful.  This eliminates dozens of runtime warnings that clutter up the test output.  This ignoring can be undone by commenting out every line that invokes np.seterr (3 each in _trig_and_test_helper, _infinity_edge_case_helper).

### DIFF
--- a/PROTO_tests/tests/numeric_test.py
+++ b/PROTO_tests/tests/numeric_test.py
@@ -3,22 +3,54 @@ import pytest
 import arkouda as ak
 from arkouda.dtypes import npstr
 
+prob_size = 1000
 NUMERIC_TYPES = [ak.int64, ak.float64, ak.bool, ak.uint64]
 NO_BOOL = [ak.int64, ak.float64, ak.uint64]
 NO_FLOAT = [ak.int64, ak.bool, ak.uint64]
 INT_FLOAT = [ak.int64, ak.float64]
 
+# There are many ways to create a vector of alternating values.
+# This is a fairly fast and fairly straightforward approach.
+
+def alternate(L,R,n):
+    v = np.full(n,R)
+    v[::2] = L
+    return v
+
+#np.random.seed (8675309)
+
+#  The following tuples support a simplification of the trigonometric
+#  and hyperbolic testing.
+
+TRIGONOMETRICS = (
+        (np.sin,ak.sin),
+        (np.cos,ak.cos),
+        (np.tan,ak.tan),
+        (np.arcsin,ak.arcsin),
+        (np.arccos,ak.arccos),
+        (np.arctan,ak.arctan)
+    )
+
+HYPERBOLICS = (
+        (np.sinh,ak.sinh),
+        (np.cosh,ak.cosh),
+        (np.tanh,ak.tanh),
+        (np.arcsinh,ak.arcsinh),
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/numeric_test.py
```diff
@@ -455,6 +507,8 @@ def test_value_counts(self, num_type):
 
     def test_value_counts_error(self):
         pda = ak.linspace(1, 10, 10)
+        # with pytest.raises(TypeError):
+        #   ak.value_counts(pda)
```
ajpotts, line 511: Why is this commented out?

PROTO_tests/tests/numeric_test.py
```diff
@@ -341,97 +402,88 @@ def test_arctan2(self, num_type, denom_type):
 
         assert np.allclose(
             [
-                np.arctan2(na_num[i], na_denom[i]) if truth_np[i] else na_num[i] / na_denom[i]
+                (
+                    np.arctan2(na_num[i], na_denom[i])
+                    if truth_np[i]
+                    else na_num[i] / na_denom[i]
+                )
```
ajpotts, line 409: Can you check your black settings?  It still seems like these are being reformatted more than necessary.

ajpotts (approved):
Looks good modulo the comments.  Thanks for doing this!

PROTO_tests/tests/numeric_test.py
drculhane: Because once again, I forgot to simply delete it after testing.  So that's fixed.  This used to raise a TypeError, but doesn't anymore.

PROTO_tests/tests/numeric_test.py
drculhane: I've just written a script to use the correct line length in black.  That should fix this going forward.

PROTO_tests/tests/numeric_test.py
```diff
@@ -3,22 +3,41 @@
 import arkouda as ak
 from arkouda.dtypes import npstr
 
+prob_size = 1000
```
jaketrookman, line 6: I would take this out, although it is good for running quick tests, it will overwrite any set value of the `prob_size`.

drculhane: Yep.  Another oversight.  Have removed it, committed, and pushed.

stress-tess (approved):

PR #3154, opened 2024-05-03
Closes #3079 and #3080: Sum and Plus Equal of Boolean Arrays 

### LINKED ISSUE #3079: plus equals (bool,bool)
Add += for (bool, bool).  The following illustrates the differences between numpy and pandas:

```python

ak_x = ak.array([True,False])
np_x = np.array([True, False])
np_x += True
ak_x += True
```
RuntimeError: Error: opeqvsMsg: unrecognized type: (bool,bool)

```python
np_x += np_x
ak_x += ak_x
```
RuntimeError: Error: opeqvvMsg: bool += bool not implemented

### LINKED ISSUE #3080: sum boolean pdarrays
Add sum of two boolean vectors to match numpy.

```python
np_x = np.array([True, False])
ak_x = ak.array([True, False])

 np_x + np_x
array([ True, False])
ak_x + ak_x
```
RuntimeError: Error: binopvvMsg: bool + bool not implemented

jaketrookman:
This PR (Closes #3080) adds the addition binop to Bool-Bool with vector-vector, scalar-vector, and vector-scalar cases. And this PR (Closes #3079) adds the plus equal opeq Bool-Bool with vector-vector, scalar-vector.

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -524,7 +524,7 @@ module BinOp
             e.a = l.a:int >= val:int;
           }
           when "+" {
-            e.a = (l.a:int + val:int):bool;
+            e.a = l.a | val;
           }
           otherwise {
             var errorMsg = notImplementedError(pn,l.dtype,op,dtype);
@@ -918,7 +918,7 @@ module BinOp
             e.a = val:int >= r.a:int;
           }
           when "+" {
-            e.a = (val:int + r.a:int):bool;
+            e.a = val | r.a;
           }
           otherwise {
             var errorMsg = notImplementedError(pn,dtype,op,r.dtype);
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -109,6 +109,9 @@ module BinOp
           when ">=" {
             e.a = l.a:int >= r.a:int;
           }
+          when "+" {
+            e.a = (l.a:int + r.a:int):bool;
+          }
           otherwise {
             var errorMsg = notImplementedError(pn,l.dtype,op,r.dtype);
             omLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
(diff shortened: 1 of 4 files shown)


src/OperatorMsg.chpl
```diff
@@ -1346,6 +1346,11 @@ module OperatorMsg
                     when "|=" {l.a |= r.a;}
                     when "&=" {l.a &= r.a;}
                     when "^=" {l.a ^= r.a;}
+                    when "+=" {
+                      ref la = l.a;
+                      ref ra = r.a;
+                      [(li,ri) in zip(la,ra)] li = if (li == true || ri == true) then true else false;
+                    }
```
stress-tess, line 1353: I might be missing something, but this seems equivalent to the `l.a |= r.a;` case

src/BinOp.chpl
```diff
@@ -109,6 +109,9 @@ module BinOp
           when ">=" {
             e.a = l.a:int >= r.a:int;
           }
+          when "+" {
+            e.a = (l.a:int + r.a:int):bool;
+          }
```
stress-tess, line 114: similarly I don't think we need casts. It seems `e.a = l.a | r.a` should do what we want

src/OperatorMsg.chpl
jaketrookman: You are totally right, I totally overlooked the operation right above where I added this.

src/BinOp.chpl
jaketrookman: Yes, this is much better than the casting. Changed

ajpotts (approved):
Looks good to me.  Thanks for taking care of this!

src/OperatorMsg.chpl
```diff
@@ -1838,6 +1839,21 @@ module OperatorMsg
...
                 }
             }
+            when (DType.Bool, DType.Bool) {
+                var l = toSymEntry(left, bool, nd);
+                var val = value.getBoolValue();
+                select op {
+                    when "+=" {
+                      ref la = l.a;
+                      [li in la] la = if (li == true || val == true) then true else false;
+                    }
```
stress-tess, line 1849: I think this could also be `l.a |= val`?

stress-tess (commented):
one last area where we might be able to simplify

src/OperatorMsg.chpl
jaketrookman: My bad, left this one out, great catch

stress-tess (approved):
looks good!

PR #3159, opened 2024-05-06
Support for Zarr IO 

brandon-neth:
This PR adds support for reading and writing potentially multi-dimensional arrays from and to zarr stores. Current support only allows for whole-array IO but future development may add support for partial chunk reading/writing.


src/ZarrMsg.chpl
```diff
@@ -0,0 +1,125 @@
...
+
+  @arkouda.registerND(cmd_prefix="readAllZarr")
+  proc readAllZarrMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, param nd: int): MsgTuple throws {
+    const storePath = msgArgs.getValueOf("store_path"), dtype = str2dtype(msgArgs.getValueOf("dtype"));
+    const rname = st.nextName();
+    select dtype {
+      when DType.Float64 {
+        var ar1 = readZarrArray(storePath, real, nd);
+        const a = makeDistArray(ar1);
+        st.addEntry(rname, createSymEntry(a));
```
bmcdonald3, line 42: Am I understanding this correctly that we are reading into a DR (meaning all memory on locale 0) and then turning that into a distributed array?

brandon-neth: `readZarrArray` reads in the store into a distributed array using all locales; the `makeDistArray` call was added while debugging a comms issue, but I've fixed that elsewhere. These calls will be removed.

brandon-neth:
I've got the compatibility modules working on my machine for 1.33 and 1.32. For 1.31, changes to IO modules will mean reimplementing the JSON module using 1.31 parlance. This is a significant lift, so I'm going to prioritize resolving the c-blosc dependency first.

Makefile
```diff
@@ -264,6 +277,7 @@ check-idn2: $(IDN2_CHECK)
 	$(DEP_INSTALL_DIR)/$@ -nl 1
 	@rm -f $(DEP_INSTALL_DIR)/$@ $(DEP_INSTALL_DIR)/$@_real
 
+
```
bmcdonald3, line 280: Added this whitespace by accident?

src/SymArrayDmap.chpl
```diff
@@ -79,15 +79,19 @@ module SymArrayDmap {
...
     }
 
     proc makeDistArray(in a: [?D] ?etype) throws
       where D.rank > 1 && (MyDmap == Dmap.defaultRectangular || !a.isDefaultRectangular())
     {
-      return a;
+      const dom = makeDistDom((...D.shape));
+      var res: [dom] etype;
+      res = a;
+      return res;
```
bmcdonald3, line 93: What is the point of these changes? These have had performance impacts in the past, so I think we need to make sure they are OK before changing thee. Noting that the first function has a `where` clause, so we know it's only 1D and shouldn't need to do tuple unrolling there.

src/compat/eq-131/ArkoudaZarrCompat.chpl
```diff
@@ -0,0 +1,411 @@
+module ArkoudaZarrCompat {
```
bmcdonald3, line 1: Why do we have distinct modules for each one here? It seems they are all at least largely the same.

Also, we don't support 1.31 anymore.

bmcdonald3 (changes requested):

Makefile
brandon-neth: Yep. Deleted.

src/compat/eq-131/ArkoudaZarrCompat.chpl
brandon-neth: Zarr is only supported in Chapel as of 2.0, so the compatability module just includes the Zarr module body for the <2.0 versions. I'll remove the 1.31 version though.

src/SymArrayDmap.chpl
brandon-neth: When reading a Zarr store, the `readZarrArray` method returns a distributed array. If I, without this code change, try to add that array to the symbol table, I get the following error: 
```
/Users/brandonneth/arkouda//src/MultiTypeSymEntry.chpl:261: error: cannot assign expression of type [BlockDom(1,int(64),one,unmanaged DefaultDist)] real(64) to field 'a' of type [domain(1,int(64),one)] real(64)
```

I think this is caused by the fact that the generic `SymEntry` class only supports the `a` field having the default type of `makeDistArray`. I ended up reverting the code in this file and instead creating an array with the correct type and copying into it within the ZarrMsg.chpl file.

bmcdonald3 (approved):
Now that Arkouda <2.0 isn't supported, if we just drop the 1.32, 1.33, 1.34 compat modules, then I think this is good to merge.

bmcdonald3 (approved):

PR #3160, opened 2024-05-06
Implement missing stats, search and sort functions for Array API

jeremiah-corrado:
Add implementations of the following Array API functions:
* [sort](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sort.html#array_api.sort)
* [search_sorted](https://data-apis.org/array-api/latest/API_specification/generated/array_api.searchsorted.html#array_api.searchsorted)
* [cumulative_sum](https://data-apis.org/array-api/latest/API_specification/generated/array_api.cumulative_sum.html#array_api.cumulative_sum)

### DIFF
--- a/src/ArgSortMsg.chpl
+++ b/src/ArgSortMsg.chpl
@@ -372,6 +372,8 @@ module ArgSortMsg
           }
         }
         when SortingAlgorithm.RadixSortLSD {
+          // TODO: make a version of radixSortLSD_ranks that does the sort on
+          // slices of `A` directly instead of creating a copy for each slice
           for idx in DD {
             const sliceDom = domOnAxis(D, idx, axis),
                   aSliced1D = removeDegenRanks(A[sliceDom], 1),
--- a/src/SortMsg.chpl
+++ b/src/SortMsg.chpl
@@ -5,13 +5,14 @@ module SortMsg
     use ArkoudaTimeCompat as Time;
     use Math only;
     use Sort only;
+    use Search only;
     use Reflection;
     use ServerErrors;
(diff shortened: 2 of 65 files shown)


src/ArgSortMsg.chpl
```diff
@@ -372,6 +372,8 @@ module ArgSortMsg
           }
         }
         when SortingAlgorithm.RadixSortLSD {
+          // TODO: make a version of radixSortLSD_ranks that does the sort on
+          // slices of `A` directly instead of creating a copy for each slice
```
bmcdonald3, line 376: 👍

src/SortMsg.chpl
```diff
@@ -29,89 +30,190 @@ module SortMsg
...
+        throw getErrorWithContext(
+          msg="Unrecognized side: %s".doFormat(side),
+          lineNumber=getLineNumber(),
+          pn,
+          moduleName=getModuleName(),
+          errorClass="NotImplementedError"
+        );
+      }
+
+      // TODO: add support for Float32
```
bmcdonald3, line 163: Why doesn't this work with `Float32` today?

bmcdonald3 (approved):
Looks good, just a couple questions

src/SortMsg.chpl
jeremiah-corrado: It could, but would require a separate code path that I didn't add yet. I think that work could be saved for when we start adding more support for all the Array API types, but probably isn't needed yet.

bmcdonald3: Sounds good to me, thanks.

PR #3163, opened 2024-05-07
Closes #3089: Avoid OOM Crashes caused due to `in` intents on `makeDistArray`

### LINKED ISSUE #3089: Arkouda crash on argsort
**Describe the bug**
ak.argsort _appears_ to crash Arkouda with an uncaught OOM, first producing the message below followed by an arkouda_server crash:

```
projected memory use of 77%
```

Also, the radixSortLSDCore invocation within the ak.DataFrame.groupby  _appears_ to crash Arkouda with an uncaught OOM, first producing the message below followed by an arkouda_server crash:

```
projected pct memory used of 84%
```

**To Reproduce**
Steps to reproduce the ak.argsort behavior:
1. export CHPL_RT_NUM_THREADS_PER_LOCALE=3
2. ./arkouda_server -nl 6 --logLevel=LogLevel.DEBUG memMax=12000000000,  --ServerDaemon.daemonTypes=ServerDaemonType.DEFAULT,ServerDaemonType.METRICS
3. Load a 1-D array with the following attributes: 
`'[{"name":"<array name>", "dtype":"int64", "size":1393000000, "ndim":1, "shape":[1393000000], "itemsize":8, "registered":false}]'`
4. ak.argsort(pd)
6. Logging stops, for AOS squeue shows Arkouda job has stopped

Steps to reproduce the ak.groupby behavior:
1. export CHPL_RT_NUM_THREADS_PER_LOCALE=3
2. ./arkouda_server -nl 6 --logLevel=LogLevel.DEBUG memMax=40000000000,  --ServerDaemon.daemonTypes=ServerDaemonType.DEFAULT,ServerDaemonType.METRICS
3. Load a 1-D array with the following attributes: 
`'[{"name":"<array name>", "dtype":"int64", "size":2786000000, "ndim":1, "shape":[2786000000], "itemsize":8, "registered":false}]'`
4. dft = {'<column name>', pdarray}
5. df = ak.DataFrame(dft)
6. df.groupby(keys=['<column name>').count()
7. Logging stops, for AOS squeue shows Arkouda job has stopped

**Expected behavior**
Should get overMemLimit error

**Error Message**
logging stops

**Is this a Blocking Issue**
Not sure how widespread this error is potentially, but it's a very popular command

ShreyasKhandekar:
`in` intents were causing array copies to be made in the `makeDistArray` functions outside of the `tryCreateArray` interface which caused the server to crash. 
By creating the array using the shape and immediately copying the original array after allocation we can avoid the OOM crash.

This resolves https://github.com/Bears-R-Us/arkouda/issues/3089
See more info about the nature of this issue here: https://github.com/chapel-lang/chapel/issues/25005

### DIFF
--- a/src/RadixSortLSD.chpl
+++ b/src/RadixSortLSD.chpl
@@ -67,11 +67,12 @@ module RadixSortLSD
     private proc radixSortLSDCore(ref a:[?aD] ?t, nBits, negs, comparator) throws {
         try! rsLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
                                        "type = %s nBits = %?".doFormat(t:string,nBits));
-        var temp = makeDistArray(a);
-        
+        var temp = makeDistArray((...aD.shape), a.eltType);
+        temp = a;
+
         // create a global count array to scan
         var globalCounts = makeDistArray((numLocales*numTasks*numBuckets), int);
-        
+
         // loop over digits
         for rshift in {0..#nBits by bitsPerDigit} {
             const last = (rshift + bitsPerDigit) >= nBits;


jeremiah-corrado:
> So I imagine this would apply to any case where we are passing an array into makeDistArray?

The copy only happens in cases where an array is passed into `makeDistArray`, and then is used again after being passed in. For example:
```chapel
var a: [d] real = ...
...
var b = makeDistArray(a); // a copied here
...
writeln(a); // because it's used again here
```

However, if it isn't used again after being passed in, then the compiler moves the array into `makeDistArray`, rather than copying it in. For example:
```chapel
proc foo(): [] real {
  var a: [d] real = ...
  ...
  return makeDistArray(a); // a moved here
}
```

So there are probably some other cases where a change like this one would be useful, but I don't think it applies to all cases.

stress-tess:
> > So I imagine this would apply to any case where we are passing an array into makeDistArray?
> 
> The copy only happens in cases where an array is passed into `makeDistArray`, and then is used again after being passed in. For example:
> 
> ```chapel
> var a: [d] real = ...
> ...
> var b = makeDistArray(a); // a copied here
> ...
> writeln(a); // because it's used again here
> ```
> 
> However, if it isn't used again after being passed in, then the compiler moves the array into `makeDistArray`, rather than copying it in. For example:
> 
> ```chapel
> proc foo(): [] real {
>   var a: [d] real = ...
>   ...
>   return makeDistArray(a); // a moved here
> }
> ```
> 
> So there are probably some other cases where a change like this one would be useful, but I don't think it applies to all cases.

oh nice! yah that should be way less common. Thanks for clarifying!

PR #3164, opened 2024-05-07
Closes #3148: Update IO functions to always return a dictionary

### LINKED ISSUE #3148: Update IO functions to always return a dictionary
Users have requested we have `read_parquet` always return a dictionary. Right now passing it a single column will return that column type (for example pdarray) but it would be nice to be consistent, so it's always valid to do `ak.DataFrame(ak.read_parquet(...))`

stress-tess:
Our IO functions used to return a dictionary if multiple columns or datasets were provided and if only was provided to just return the object. It has been requested by users that our IO functions always return a dictionary for consistency sake. This PR (closes #3148) makes the changes necessary for this and also updates all tests that relied on the old functionality

The majority of the changes are updating the tests. We index using the dataset name where explicit. Otherwise we use `.popitem()` which returns the `(key,item)` tuple (since there is only one). We can then index at `[1]` to get the object that would've been returned previously


ajpotts (approved):
Looks good!  Thanks for taking this one.

PR #3180, opened 2024-05-13
Closes #3167: Add `normal` to random number generators

### LINKED ISSUE #3167: Add `normal` to random number generators
Add `normal` to our generators, we already have a `fillRandom` helper in `Random.chpl` so this shouldn't be a big lift

Relevant numpy docs:
- https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.normal.html

stress-tess:
This PR (closes #3167) adds `normal` to our generators by transforming the results of standard normal. It also updates `standard_normal` to use generators to make sure we have the same multiple calls of the same function give different output as the other rand funcs. Adds goodness of fit testing against scipy to give confidence that we are actually drawing from the correct distribution

Added this to the docs with correct formatting of the math equations. I also added some deps to the conda env yaml files that were introduced in #3077. Also I changed the eol for that file in my pycharm from carriage return `\r\n` to just newline `\n`. This makes it look like every line in the yaml changed. In the `Files changed` tab you can click the setting wheel and hide whitespace to see what was actually added

### DIFF
--- a/arkouda/random/_generator.py
+++ b/arkouda/random/_generator.py
@@ -210,8 +210,8 @@ class Generator:
         return create_pdarray(rep_msg)
 
     def normal(self, loc=0.0, scale=1.0, size=None):
-        """
-        Draw samples from a normal distribution
+        r"""
+        Draw samples from a normal (Gaussian) distribution
 
         Parameters
         ----------
@@ -223,6 +223,16 @@ class Generator:
 
         size: numeric_scalars, optional
             Output shape. Default is None, in which case a single value is returned.
+
+        Notes
+        -----
+        The probability density for the Gaussian distribution is:
+
+        .. math::
+           p(x) = \frac{1}{\sqrt{2\pi \sigma^2}} e^{-\frac{(x-\mu)^2}{2\sigma^2}}
+
+        where :math:`\mu` is the mean and :math:`\sigma` the standard deviation.
+        The square of the standard deviation, :math:`\sigma^2`, is called the variance.
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


stress-tess:
converting to draft because I forgot to add documentation

pydoc/usage/random.rst
```diff
@@ -17,10 +17,26 @@ To create a ``Generator`` use the ``default_rng`` constructor
 Features
 ==========
 
+choice
+---------
+.. autofunction:: arkouda.random.Generator.choice
+
```
ajpotts, line 23: So why are some functions added to the usage section of the docs and not others?  What is the intended difference between the usage and api sections?

arkouda/random/_generator.py
```diff
@@ -209,6 +209,55 @@ def integers(self, low, high=None, size=None, dtype=akint64, endpoint=False):
         self._state += size
         return create_pdarray(rep_msg)
 
+    def normal(self, loc=0.0, scale=1.0, size=None):
+        r"""
+        Draw samples from a normal (Gaussian) distribution
```
ajpotts, line 214: The doc string for this one isn't rendering for me with `make doc`.

ajpotts (approved):
Looks good!  Just the two questions about the docs.

jaketrookman (approved):

pydoc/usage/random.rst
stress-tess: to be honest, this predates me. I don't really know. I think the api lists everything and usage is just meant for the most common stuff

arkouda/random/_generator.py
stress-tess: oh really?? that's odd. I might swing by to see what it's showing you

stress-tess: yah I found the problem... just had a random `latex` keyword in the usage section `.. autofunction:: latex arkouda.random.Generator.normal`... should be fixed now!

PR #3189, opened 2024-05-15
Add benchmark for for CSV Read and write perf

ShreyasKhandekar:
This PR does the following:
- Updates the `time_ak_write`, `time_ak_read`, and `check_correctness` functions in IO.py to take a `fileFormat` argument instead of a the old `parquet` argument.
- `fileFormat` is an enum to distinguish between the three different filetypes, namely- hdf5, parquet, and csv.
- Add support for testing performance of csv reads and write in the above functions
- Remove the unused `seed` argument from `time_ak_read`
- Add a new benchmark `csvIO.py`. Currently CSV read performance is pretty bad and the benchmark just says 0.0 because of rounding. I have a follow up task to improve CSV read performance.
- Update `multiIO.py`, `parquetIO.py`, and `multiParquetIO.py` to work with the new changes to `IO.py` described in the first bullet
- Add graph infrastructure to plot the new benchmark
- Small change to contributing.rst (we omit the `.py`)

### DIFF
--- a/benchmarks/IO.py
+++ b/benchmarks/IO.py
@@ -174,9 +174,9 @@ def check_correctness(dtype, path, seed, fileFormat, multifile=False):
         raise ValueError(f"Invalid dtype: {dtype}")
 
     file_format_actions = {
-        FileFormat.HDF5: (a.to_hdf, b.to_hdf if b else None, ak.read_hdf),
-        FileFormat.PARQUET: (a.to_parquet, b.to_parquet if b else None, ak.read_parquet),
-        FileFormat.CSV: (a.to_csv, b.to_csv if b else None, ak.read_csv)
+        FileFormat.HDF5: (a.to_hdf, b.to_hdf if b is not None else None, ak.read_hdf),
+        FileFormat.PARQUET: (a.to_parquet, b.to_parquet if b is not None else None, ak.read_parquet),
+        FileFormat.CSV: (a.to_csv, b.to_csv if b is not None else None, ak.read_csv)
     }
 
     if fileFormat in file_format_actions:
--- a/benchmarks/IO.py
+++ b/benchmarks/IO.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import argparse
+from enum import Enum
 import os
 import time
 from glob import glob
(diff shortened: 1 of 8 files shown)


benchmarks/IO.py
```diff
@@ -232,9 +286,11 @@ def create_parser():
     comp_str = args.compression
     comp_types = COMPRESSIONS if comp_str == "" else comp_str.lower().split(",")
 
+    fileFormat = FileFormat.CSV if args.csv else FileFormat.PARQUET if args.parquet else FileFormat.HDF5
```
bmcdonald3, line 289: I think this is an issue if both Parquet and CSV are set.

I think maybe what I would do is run both CSV and Parquet if they are both set? But that also seems odd, so I'm not sure... 

Anyways, I think this is fine for now, as it is mainly run by our nightly performance testing, but I think that could potentially cause confusion

bmcdonald3 (approved):
Looks good, just one comment I don't think should hold this up over.

Also, have you made sure this works with a manual run? Do you have any characteristics to share about what you expect the runs to come out as on XC or single node?

benchmarks/IO.py
ShreyasKhandekar: Hi Ben, the argument parser will not allow both of them to be set. In lines 235-241, the two arguments are in a mutually exclusive group now; therefore, only one of them is allowed at a time.

ShreyasKhandekar:
I have run this manually, the benchmark and generating graphs both run successfully. The performance is unfortunately so abysmal that it is reported as being basically 0 GB/s. Here are the results from a run on ChapDL:

```
>>> arkouda int64 CSV write
numLocales = 1, N = 1,000,000, filesPerLoc = 1
write Average time CSV = 0.1532 sec
write Average rate CSV = 0.05 GiB/sec
>>> arkouda int64 CSV read
numLocales = 1, N = 1,000,000, filesPerLoc = 1
read Average time CSV = 5.1015 sec
read Average rate CSV = 0.00 GiB/sec
```

It looks like the numbers I get are about 2-3 times slower than what our nightly machines get (comparing the `IO` benchmark) so maybe in the nightly the graph will say something other than 0. 
However, the read time is a good measure to go by in my opinion since it isn't 0, and I hope that my improvements to the implementation will help bring up the graph significantly.

benchmarks/IO.py
bmcdonald3: Great, thanks for clarifying!

bradcray:
> The read performance is unfortunately so abysmal that it is reported as being basically 0 GB/s.

Should we change the units, precision of the timing being printed, or problem size to try to get it to be something non-zero?

ShreyasKhandekar:
Ah sorry for the late response, I need to redirect emails for this repo like I do for the others. I saw these when checking my personal email. Yes, this is ready to be merged.

Brad - the reason I didn't change the units or change the problem size is to that we can do an apples to apples comparison of different file formats' performance in Arkouda with just a quick look at the graphs. And hopefully these numbers will rise to be non-zero by the end of this week.

bradcray:
> we can do an apples to apples comparison of different file formats' performance in Arkouda with just a quick look at the graphs.

What about just printing more significant figures then?  E.g., if I'm doing my math right, printing 4 or 5 decimal places rather than 2 seems like it ought to yield a non-zero result?

bmcdonald3:
> What about just printing more significant figures then? E.g., if I'm doing my math right, printing 4 or 5 decimal places rather than 2 seems like it ought to yield a non-zero result?

I think that since we are getting the timing in seconds as well as the GiB/s numbers, this wouldn't add a ton to the output here. In other words, a reduction in total execution time will be reflected in the same way if we change the significant digits printed out and won't impact the final graphing.

The output we are seeing here is:
```
...
write Average time CSV = 0.1532 sec
write Average rate CSV = 0.05 GiB/sec
...
read Average time CSV = 5.1015 sec
read Average rate CSV = 0.00 GiB/sec
```
So the seconds is already "significant" enough to be meaningful in my mind.

ShreyasKhandekar:
I'm okay with changing the significant digits in the `GB/s` to also be 4 so that the graphs don't say 0. The seconds is significant enough, but it's not graphed so it would be nice to make this change.

bmcdonald3 (approved):

bmcdonald3:
I had the point of those digits mixed up, I was thinking that what we printed would not affect graph output and it was based off of the graps sig figs, but I looked into it and it turns out that what is printed is used, so the 4 digits definitely makes sense, my mistake.

PR #3193, opened 2024-05-16
Closes #3192 PROTO_tests/tests/dtypes_test.py is failing

### LINKED ISSUE #3192: PROTO_tests/tests/dtypes_test.py is failing
PROTO_tests/tests/dtypes_test.py is failing `make test-proto`.  Fix the bug so that this test suite passes.

ajpotts:
Closes #3192 PROTO_tests/tests/dtypes_test.py is failing

This corrects the failing unit test `test_check_np_dtype` that was expecting a `TypeError` because in the past `bigint` `dtype` was not a supported type.  However, that is no longer the case.

### DIFF
--- a/PROTO_tests/tests/dtypes_test.py
+++ b/PROTO_tests/tests/dtypes_test.py
@@ -27,7 +27,7 @@ class TestDTypes:
     def test_check_np_dtype(self, dtype):
         dtypes.check_np_dtype(np.dtype(dtype))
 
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good, but just a thought, should we also remove the parameterization, or are there potentially other things we would want to test against in the future?

stress-tess (approved):

PR #3209, opened 2024-05-17
Closes #3208-Index.equals

### LINKED ISSUE #3208: Index.equals
Add `Index.equals` to match pandas:
https://pandas.pydata.org/docs/reference/api/pandas.Index.equals.html

ajpotts:
Closes #3208-Index.equals

Add `Index.equals` to match pandas:
https://pandas.pydata.org/docs/reference/api/pandas.Index.equals.html

Also adds the binops `==` and `!=` to the `Index` class.

### DIFF
--- a/tests/index_test.py
+++ b/tests/index_test.py
@@ -86,6 +86,43 @@ class IndexTest(ArkoudaTest):
         m = ak.MultiIndex([ak.arange(size), ak.arange(size) * -1], names=["test", "test2"])
         self.assertEqual(m.inferred_type, "mixed")
 
+    def assert_equal(self, pda1, pda2):
+        from arkouda import sum as aksum
+
+        assert pda1.size == pda2.size
+        assert aksum(pda1 != pda2) == 0
+
+    def test_eq(self):
+        i = ak.Index([1, 2, 3])
+        i_cpy = ak.Index([1, 2, 3])
+        self.assert_equal(i == i_cpy, ak.array([True, True, True]))
+        self.assert_equal(i != i_cpy, ak.array([False, False, False]))
+        assert i.equals(i_cpy)
+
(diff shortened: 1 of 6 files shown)


tests/index_test.py
```diff
@@ -59,6 +59,42 @@ def test_multiindex_creation(self):
...
+        i3 = ak.Index(["a", "b", "c"], allow_list=True)
+        i3_cpy = ak.Index(["a", "b", "c"], allow_list=True)
+        self.assert_equal(i3 == i3_cpy, ak.array([True, True, True]))
+        self.assert_equal(i3 != i3_cpy, ak.array([False, False, False]))
+        assert i3.equals(i3_cpy)
+
+        i4 = ak.Index(["a", "d", "c"], allow_list=True)
+        self.assert_equal(i3 == i4, ak.array([True, False, True]))
+        self.assert_equal(i3 != i4, ak.array([False, True, False]))
+        assert not i3.equals(i4)
```
stress-tess, line 93: I don't really see how `i3_cpy` and `i4` are different.  Looking at the integer version it seems like we'd want to declare `i3` and it's copy without `allow_list` set

stress-tess (approved):
Look good!! one super small comment about one of the tests, but all the logic is fine

tests/index_test.py
ajpotts: The second element is "d" v.s. "b".  It was to test comparing two lists (rather than a pdarray to a list) and to do a case where the answer was not all `True` or all `False`.  I changed the "d" to "x" to make it more obvious they were different.

stress-tess: Oh my bad! I don't know how I misses that!!

stress-tess (approved):

jaketrookman (approved):

stress-tess (approved):

PR #3210, opened 2024-05-20
Fixes #3185: strings broadcast bug

### LINKED ISSUE #3185: bug in strings broadcast
strings broadcast is not applying the permutation from the groupby
```python
>>> keys = ak.randint(0, 4, 100, int)
>>> keys
array([3 2 1 2 2 2 3 3 0 2 1 0 2 3 3 0 1 1 0 0 0 1 3 0 3 2 0 3 3 3 0 0 2 0 0 0 3 3 2 2 1 2 0 3 2 3 3 0 2 2 0 1 2 0 1 2 0 3 0 2 0 0 0 2 2 3 1 3 2 3 1 2 0 0 3 1 0 2 0 3 3 1 0 0 2 0 3 0 0 0 3 0 2 0 1 1 1 1 1 2])

>>> g = ak.GroupBy(keys)
>>> g.broadcast(ak.array([str(i) for i in range(4)]))
 array(['0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '0', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '2', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3', '3'])
```

stress-tess:
This PR fixes #3185 by removing the custom string permutation logic in favor of broadcasting the indices and then indexing into the segstring. I realized that we overcomplicating things and that optimizing this solution would essientailly be duplicating the segstring indexing code. I don't know what lead me to take this approach in the first place, but this is much cleaner and makes use of existing well-tested and optimized functions

### DIFF
--- a/PROTO_tests/tests/string_test.py
+++ b/PROTO_tests/tests/string_test.py
@@ -853,3 +853,14 @@ class TestString:
     def test_inferred_type(self):
         a = ak.array(["a", "b", "c"])
         assert a.inferred_type == "string"
+
+    def test_string_broadcast(self):
+        keys = ak.randint(0, 10, 100, int)
+        g = ak.GroupBy(keys)
+        str_vals = ak.random_strings_uniform(0, 3, 10, characters="printable")
+        str_broadcast_ans = str_vals[keys]
+
+        gb_broadcasted = g.broadcast(str_vals)
+        manual_broadcasted = ak.broadcast(g.segments, str_vals, permutation=g.permutation)
+        assert (gb_broadcasted == str_broadcast_ans).all()
+        assert (manual_broadcasted == str_broadcast_ans).all()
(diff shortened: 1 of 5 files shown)


ajpotts (approved):
Looks good!  Thanks for fixing this.

PR #3220, opened 2024-05-21
Closes #3219: Optimize old Parquet srting read code

### LINKED ISSUE #3219: Optimize old Parquet srting read code
After seeing more results gathered on different machines showing mixed results for the new Parquet string optimization, we have decided to make some changes and go back towards a simpler optimization that we are more confident in across the board optimizations, though, they may be more minor than the "best case" with the optimized version, but we shouldn't be seeing any cases get worse with this approach, which seems preferable.

bmcdonald3:
After seeing more results gathered on different machines showing mixed results for the new Parquet string optimization, we have decided to make some changes and go back towards a simpler optimization that we are more confident in across the board optimizations, though, they may be more minor than the "best case" with the optimized version, but we shouldn't be seeing any cases get worse with this approach, which seems preferable.

### DIFF
--- a/PROTO_tests/tests/io_test.py
+++ b/PROTO_tests/tests/io_test.py
@@ -118,7 +118,7 @@ def make_multi_dtype_dict():
                 ]
             ),
         ),
-        #"c_9": ak.array(["abc", " ", "xyz", ""]),
+        "c_9": ak.array(["abc", " ", "xyz", ""]),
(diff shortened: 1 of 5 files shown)


bmcdonald3:
This will pass CI after https://github.com/Bears-R-Us/arkouda/pull/3312 is merged

ajpotts (approved):

PR #3223, opened 2024-05-22
Compatibility improvements for Xarray chunk-manager

jeremiah-corrado:
Adds a `chunk_info` query to array api module for determining how an array's indices are "chunked" across locales

Also makes a couple of improvements to pdarray indexing:
* `__setitem__` can now accept numpy ndarray values
* slicing all of a multidimensional pdarray's indices with `arr[:]` is now supported (only worked for 1D arrays before)


stress-tess (approved):
looks good to me!

PR #3230, opened 2024-05-23
Closes #3226: Update proto random test

### LINKED ISSUE #3226: Update `random_test`

drculhane:
Adds parameterization to test_shuffle and test_permutation in PROTO-tests/test/random_test.py.  Other changes are cosmetic (added whitespace, expanded comments).

closes #3226


PROTO_tests/tests/random_test.py
```diff
@@ -47,70 +52,75 @@ def test_integers(self):
...
 
         # verify within bounds (lower inclusive and upper exclusive)
+
         rng = ak.random.default_rng()
         bounded_arr = rng.integers(-5, 5, 1000)
         assert all(bounded_arr.to_ndarray() >= -5)
         assert all(bounded_arr.to_ndarray() < 5)
 
-    def test_shuffle(self):
+    @pytest.mark.parametrize("data_type",INT_FLOAT)
```
ajpotts, line 61: Do we want to add `size` as a parameter to these?

PROTO_tests/tests/random_test.py
```diff
@@ -47,70 +52,75 @@ def test_integers(self):
...
-        same_seed_array_permute = rng.permutation(pda)
-        assert (array_permute == same_seed_array_permute).all()
+        # verify all the same elements are in permutation as in the original 
 
-        pda = rng.uniform(-(2**32), 2**32, 10)
-        same_seed_float_array_permute = rng.permutation(pda)
-        # verify all the same elements are in permutation as the original
-        assert np.allclose(float_array_permute.to_list(), same_seed_float_array_permute.to_list())
+        rng = ak.random.default_rng(18)
+        rnfunc = rng.integers if data_type is ak.int64 else rng.uniform
```
ajpotts, line 115: Again `rnfunc` probably doesn't need to be defined twice.

ajpotts (approved):

PROTO_tests/tests/random_test.py
```diff
@@ -47,70 +52,75 @@ def test_integers(self):
...
 
-        float_pda = rng.uniform(-(2**32), 2**32, 10)
-        pda_copy = float_pda[:]
-        rng.shuffle(float_pda)
-        # verify all the same elements are in permutation as the original
-        assert (ak.sort(float_pda) == ak.sort(pda_copy)).all()
+        # verify all the same elements are in the permutation as in the original
 
         rng = ak.random.default_rng(18)
+        rnfunc = rng.integers if data_type is ak.int64 else rng.uniform
```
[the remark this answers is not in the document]
drculhane: It does.  It has to be defined each time rng is defined, in order to reinitialize the seed.

PROTO_tests/tests/random_test.py
drculhane: But it does.

PROTO_tests/tests/random_test.py
```diff
@@ -47,70 +52,75 @@ def test_integers(self):
...
 
-        pda = rng.uniform(-(2**32), 2**32, 10)
-        float_array_permute = rng.permutation(pda)
-        # verify all the same elements are in permutation as the original
-        assert np.allclose(ak.sort(pda).to_list(), ak.sort(float_array_permute).to_list())
+        rng = ak.random.default_rng(18)
+        rnfunc = rng.integers if data_type is ak.int64 else rng.uniform
+        pda = rnfunc(-(2**32),2**32,10)
+        permuted = rng.permutation(pda)
+        assert check(ak.sort(pda),ak.sort(permuted),data_type)
```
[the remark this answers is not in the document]
drculhane: The check immediately below that is to test whether using the same seed twice gives the same permutation.  Sorting within the check would fail to check that.

PROTO_tests/tests/random_test.py
drculhane: I considered that, but I don't really think it adds anything.  In fact, there are several of our tests where it seems that way to me.

PROTO_tests/tests/random_test.py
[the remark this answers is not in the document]
drculhane: In fact, the more I looked at this, the more I realized I didn't want any sort in check (I was sorting if the data type was ak.int64), so I removed that.

PROTO_tests/tests/random_test.py
drculhane: But Tess and I were just talking, and the conversation reminded me that some of our obscure issues have only shown up on large tests, so yep, I'll add problem size to the parametrization.

drculhane: I hit "Comment" on that one too soon.  None of the tests in PROTO-tests/tests/random_test.py have size as a parameter.  All have sizes hardcoded, usually as 10, but not always.  This would be a good topic for a group discussion.

PROTO_tests/tests/groupby_test.py
stress-tess, line 1: are we intentionally deleting `groupby_test` here?

PROTO_tests/conftest.py
```diff
@@ -19,7 +19,7 @@ def pytest_addoption(parser):
              "Defaults to 2. If Arkouda is not configured for multi_locale, 1 locale is used"
     )
     parser.addoption(
-        "--size", action="store", default="10**8",
+        "--size", action="store", default="10**4",
```
stress-tess, line 22: what the rationale for lowering the default size here? was this changed just to make testing easier and you forgot to change it back or do you think we should make `10**4` the new default size

PROTO_tests/tests/random_test.py
```diff
@@ -47,70 +53,81 @@ def test_integers(self):
...
 
-        pda = rng.integers(-(2**32), 2**32, 10)
-        array_permute = rng.permutation(pda)
-        # verify all the same elements are in permutation as the original
-        assert (ak.sort(pda) == ak.sort(array_permute)).all()
+        # verify same seed gives reproducible arrays
+
+        rng = ak.random.default_rng(18)
+        rnfunc = rng.integers if data_type is ak.int64 else rng.uniform
+        pda = rnfunc(-(2**32), 2**32, 10)
```
stress-tess, line 109: this is nuanced but doing parameterization here is not testing quite the same thing as the original. Since this is only using the same rng for a single function call but the original is making sure that subsequent calls to a generator are also consistent. But it's not actually making sure the second call differs form the first... maybe this isn't that important

I think I need to chat about this with the group since I'm struggling to decide if this slight difference actually matters or if we should create a separate test for this kinda thing

PROTO_tests/tests/random_test.py
```diff
@@ -47,70 +53,81 @@ def test_integers(self):
...
 
-        pda = rng.uniform(-(2**32), 2**32, 10)
-        same_seed_float_array_permute = rng.permutation(pda)
-        # verify all the same elements are in permutation as the original
-        assert np.allclose(float_array_permute.to_list(), same_seed_float_array_permute.to_list())
+        rng = ak.random.default_rng(18)
+        rnfunc = rng.integers if data_type is ak.int64 else rng.uniform
+        pda_p = rnfunc(-(2**32), 2**32, 10)
+        permuted_p = rng.permutation(pda_p)
+        assert check(ak.sort(pda), ak.sort(permuted), data_type)
```
stress-tess, line 125: we declare `pda_p` and `permuted_p` but the we don't run `check` on these pdarrays

stress-tess (changes requested):
couple of things to look into the biggest one being the seemingly accidental deletion of dataframe test

PROTO_tests/tests/random_test.py
drculhane: Yup.  That was my mistake.  Will fix.

PROTO_tests/tests/random_test.py
drculhane: Oh, I see it.  Without parameterization, one rng is created that both INT and FLOAT use.  With parameterization, each gets its own.  It's true  that that's a difference, but I can't see how it's important.

stress-tess: covered in team sync I think this isnt an issue for shuffle / permutation, but I think adding another test to make sure we capture this duplicated calls to a generator with different dtypes and make sure the state is carried over

stress-tess:
closing in favor of https://github.com/Bears-R-Us/arkouda/pull/3320. @drculhane if this is not what we need to do, feel free to reopen!

Issue #3229, opened 2024-05-22
 CI failures due to `indexof1d`


stress-tess:
We're seeing CI failures from the new `indexof1d`: https://github.com/Bears-R-Us/arkouda/actions/runs/9196080930/job/25293360608?pr=3223

```python
=========================== short test summary info ============================
FAILED tests/setops_test.py::SetOpsTest::test_index_of - ValueError: shape mismatch [26] [20]
====== 1 failed, 638 passed, 1 skipped, 88 warnings in 559.10s (0:09:19) =======
```

stress-tess:
The path forward if I'm unable to reproduce will likely be to revert the functionality until after the release. Then we'll add it back with a seed

PR #3232, opened 2024-05-23
Part of #3229: CI failures due to `indexof1d`

stress-tess:
This PR is part of #3229. We were seeing CI failures due to `indexof1d`. I reverted back to the original implementation until I can look into this more. I added a transmute to ensure proper handling of `nan`s. I also modified the test to write out seeds to make future failures easier to reproduce

### DIFF
--- a/PROTO_tests/tests/setops_test.py
+++ b/PROTO_tests/tests/setops_test.py
@@ -680,14 +680,24 @@ class TestSetOps:
         # index of nan (reproducer from #3009)
         s = ak.Series(ak.array([1, 2, 3]), index=ak.array([1, 2, np.nan]))
         assert ak.indexof1d(ak.array([np.nan]), s.index.values).to_list() == [2]
+        rng = np.random.default_rng()
+        seeds = [rng.choice(2**63), rng.choice(2**63), rng.choice(2**63), rng.choice(2**63)]
+        print("seeds: \n", seeds)
 
+        def are_pdarrays_equal(pda1, pda2):
+            # we first check the sizes so that we won't hit shape mismatch
+            # before we can print the seed (due to short-circuiting)
+            return (pda1.size == pda2.size) and ((pda1 == pda2).all())
+
+        count = 0
         select_from_list = [
-            ak.randint(-(2**32), 2**32, 10),
+            ak.randint(-(2**32), 2**32, 10, seed=seeds[0]),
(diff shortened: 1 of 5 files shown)


PROTO_tests/tests/setops_test.py
```diff
@@ -680,30 +680,52 @@ def test_index_of(self):
...
         select_from_list = [
-            ak.randint(-(2**32), 2**32, 10),
+            ak.randint(-(2**32), 2**32, 10, seed=seeds[0]),
             ak.linspace(-(2**32), 2**32, 10),
-            ak.random_strings_uniform(1, 16, 10),
+            ak.random_strings_uniform(1, 16, 10, seed=seeds[1]),
         ]
         for select_from in select_from_list:
-            arr1 = select_from[ak.randint(0, select_from.size, 20)]
+            count += 1
```
ajpotts, line 699: You increment count before using it in the loop, but then use it an additional time outside the loop, so I think you could use the same value twice.

PROTO_tests/tests/setops_test.py
```diff
@@ -680,30 +680,52 @@ def test_index_of(self):
         # index of nan (reproducer from #3009)
         s = ak.Series(ak.array([1, 2, 3]), index=ak.array([1, 2, np.nan]))
         assert ak.indexof1d(ak.array([np.nan]), s.index.values).to_list() == [2]
+        rng = np.random.default_rng()
+        seeds = [rng.choice(2**63), rng.choice(2**63), rng.choice(2**63), rng.choice(2**63)]
+        print("seeds: \n", seeds)
```
ajpotts, line 685: Did you leave this print statement on purpose?  Same with the ones below?

ajpotts (approved):

PROTO_tests/tests/setops_test.py
stress-tess: yep! I mentioned in the PR description that I modified the test to write out seeds to make future failures easier to reproduce

PROTO_tests/tests/setops_test.py
stress-tess: I think I see what your saying but for the time outside the loops it's `seeds[3]` not `seed[2]`, so I think having the same `count` value shouldn't matter as long as the overall seed is different

jaketrookman (approved):

PR #3233, opened 2024-05-23
Fixes #3074: Bug reading segarrays from parquet files

### LINKED ISSUE #3074: Error reading segarray from parquet file
A user ran into an error when reading a list of lists (should be read in as segarray) from a parquet file.

The error was an `unexpect end of stream` occurring on line 217 of `ParquetMsg.chpl`:
https://github.com/Bears-R-Us/arkouda/blob/a84a35f5a734e85d36dd68f27b33c6c983d5dc02/src/ParquetMsg.chpl#L214-L218

This is in `readListFilesByName`, so it is correctly identifying this as a segarray. The error seems to be happening in the cpp code `cpp_readListColumnByName`. My first guess is we are incorrectly handling empty segments

stress-tess:
We were seeing a bug where segarrays containing empty segments would fail if they exceeded ~700000 values. It seems our c++ code was mishandling the empty segments and this only appeared when values were large enough (my running theory is it has to do with the batch sizes). Reworked the logic to properly account for empty segment fixes #3074, but requires we read single elements instead of batches (which is standard when nans are present)

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -572,6 +572,7 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
       }
 
       int64_t i = 0;
+      int64_t arrayIdx = 0;
       for (int r = 0; r < num_row_groups; r++) {
         std::shared_ptr<parquet::RowGroupReader> row_group_reader =
           parquet_reader->RowGroup(r);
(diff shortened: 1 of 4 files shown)


src/ArrowFunctions.cpp
```diff
@@ -581,32 +582,37 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
...
-            if((numElems - i) < batchSize)
-              batchSize = numElems - i;
-            // Can't read directly into chpl_ptr because it is int64
-            (void)reader->ReadBatch(batchSize, nullptr, nullptr, tmpArr, &values_read);
-            for (int64_t j = 0; j < values_read; j++)
-              chpl_ptr[i+j] = (int64_t)tmpArr[j];
-            i+=values_read;
+          int32_t tmp;
+          while (reader->HasNext() && arrayIdx < numElems) {
+            (void)reader->ReadBatch(1, &definition_level, nullptr, &tmp, &values_read);
```
bmcdonald3, line 608: Breaks my heart to be going to single batch size for so many of these </3, but I think that the segarray is an obscure enough case that I am not losign sleep over it (and correctness is of course more important).

bmcdonald3 (approved):

stress-tess:
hmmmm it seems that `nan`s and empty segs might still need some looking into. I've captured that in https://github.com/Bears-R-Us/arkouda/issues/3234. But I think this is still worth getting in as is since it contains a fix for a user with int segarrays

src/ArrowFunctions.cpp
stress-tess: 💔

jaketrookman (approved):

PR #3236, opened 2024-05-24
Move some definitions from ArrowFunctions header to source

e-kayrakli:
This PR is a small fix for issues we've encountered with GPU testing.

ArrowFunctions.h header has some `map` definitions that causes multiple definitions issues as the header is used in multiple places. This only comes up in Chapel's GPU testing because it uses C++ linkage. Note that the problematic definitions are in `ifdef __cplusplus` which is only true when compiling ArrowFunctions.cpp normally. However, that's always true when using GPU support because of C++ linkage.

To me, it looks like these maps are implementations details for ArrowFunctions.cpp. So, this PR just moves them to be static globals in that file. If they need to be used in more than one source file down the road, the correct way of doing that is to, declare them as `extern`s in the header and remove the `static` qualifier in the source code.

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -1,5 +1,10 @@
 #include "ArrowFunctions.h"
 
+static std::map<int, std::shared_ptr<parquet::ParquetFileReader>> globalFiles;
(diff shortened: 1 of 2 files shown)


bmcdonald3 (approved):
Thanks Engin <3

PR #3237, opened 2024-05-24
Add implementations of `clip`, `diff`,`pad` to Array API module

jeremiah-corrado:
XArray requires its array backends to implement a few numpy functions that are not yet included in the Array API standard: [`clip`](https://numpy.org/doc/stable/reference/generated/numpy.clip.html), [`diff`](https://numpy.org/doc/stable/reference/generated/numpy.diff.html), and [`pad`](https://numpy.org/doc/stable/reference/generated/numpy.pad.html). Note, `np.repeat` also falls under this category, but was already added in [this PR](https://github.com/Bears-R-Us/arkouda/pull/3056).

This PR adds implementations of the functions to the utility_functions array-api module. The `pad` function only implements the `constant` mode, as it appears the other modes aren't used in XArray (other padding modes can be added as needed).

*also fixes a bug in `Array.__bool__`

### DIFF
--- a/ArrayAPIServerModules.cfg
+++ b/ArrayAPIServerModules.cfg
@@ -17,6 +17,7 @@ SetMsg
 SortMsg
 StatsMsg
 TransferMsg
+UtilMsg
 
 # Add additional modules located outside
 # of the Arkouda src/ directory below.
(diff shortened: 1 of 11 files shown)


stress-tess (approved):
looks good to me!

jaketrookman (approved):

PR #3238, opened 2024-05-24
Improve the CSV Read performance with new impl

ShreyasKhandekar:
This PR improves the CSV read performance, especially in multilocale and
multi-file settings where the performance previously was abysmal.

Here's the performance improvements in different settings:

This new implementation does the following:
- Create a new `read_csv_pattern` method that takes out the file reading logic from `read_files_into_dist_array`
- Creates a new `csvLine` custom record that uses Chapel deserializers to read the csv file faster and better.
- Instead of reading the entire file at all times in all locales and storing the entire file in memory we do the following:
    - Skip even opening the file when the file domains do not intersect with the local subdomains of the array for the current locale
    - Read the file line by line and only retain a single item from each line in memory, dropping the unnecessary data
    - Minimize unnecessary communication across locales by doing so
- We also read all files in parallel on all locales since there is little communication in this new implementation

These are the read time numbers from the `csvIO.py` benchmark
Please keep in mind that the numbers aren't the best that they can be for this machine, but the ratio of the speedup is what we should look at here.
The times are in seconds and file size is in number of lines of csv.

1. Single locale, single file:

| File Size | old      | new      |
|------|----------|----------|
| 1M   | 13.3876  | 9.5591   |
| 5M   | 63.3379  | 44.2360  |
| 10M  | 125.9585 | 87.3109  |
| 15M  | 186.5688 | 131.8712 |

2. Multi locale, single file per local

| 5 locales 1 file per locale N=5*value_below | old      | new      |
|---------------------------------------------|----------|----------|
| 1M                                          | 145.3038 | 18.2276  |
| 5M                                          | 711.4987 | 90.4563  |
| 10M                                         | -        | 176.2367 |
| 15M                                         | -        | 263.6036 |

3. Multi locale, multi file per locale:

| 5 locales n file per locale N=5M total | old      | new      |
|----------------------------------------|----------|----------|
| 1                                      | 145.3038 | 19.6761  |
| 5                                      | 668.5100 | 63.5184  |
| 10                                     | -        | 127.4142 |
| 15                                     | -        | 188.9333 |

The `-` for the old implementation values are because the trials were going to take too long and I didn't want to wait.

In general, this is a huge leap forward for multilocale settings.

### DIFF
--- a/src/CSVMsg.chpl
+++ b/src/CSVMsg.chpl
@@ -227,7 +227,7 @@ module CSVMsg {
 
         coforall loc in Locales do on loc {
             const localeFilename = filenames[loc.id];
-            
+
             // create the file to write to
             var csvFile = open(localeFilename, ioMode.cw);
             var writer = csvFile.writer(locking=false);
@@ -312,13 +312,13 @@ module CSVMsg {
         var hasHeader = false;
         var dtype_idx = 0;
         var column_name_idx = 0;
-        
+
         if lines[0] == CSV_HEADER_OPEN + "\n" {
             hasHeader = true;
             dtype_idx = 1;
             column_name_idx = 3;
         }
-        
+
         var columns = lines[column_name_idx].split(col_delim).strip();
         var file_dtypes: [0..#columns.size] string;
         if dtype_idx > 0 {
@@ -342,7 +342,7 @@ module CSVMsg {
                 throw getErrorWithContext(
                     msg="The dataset %s was not found in %s".doFormat(dset, filename),
                     lineNumber=getLineNumber(),
-                    routineName=getRoutineName(), 
+                    routineName=getRoutineName(),
                     moduleName=getModuleName(),
                     errorClass="DatasetNotFoundError");
             }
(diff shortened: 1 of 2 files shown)


src/CSVMsg.chpl
```diff
@@ -353,6 +353,110 @@ module CSVMsg {
...
+        var item: itemType;
+        const colIdx: int;
+        const colDelim: string;
+        const r: regex(string);
+
+        proc init(type itemType, colIdx: int, colDelim: string) {
+            this.itemType = itemType;
+            this.colIdx = colIdx;
+            this.colDelim = colDelim;
+            this.r = try! new regex("[\n"+this.colDelim+"]");
```
bmcdonald3, line 443: We only need this regex when it's a string type, right? I'm not sure it's a big deal at all, just wondering if we could potentially avoid creating this everytime.

Also, I am seeing the `try!` syntax here, which we try to avoid in Arkouda. Even if it is very unlikely to throw, we never want to create a scenario where we can crash the server. We could possibly move it to after a `init this` (I think that's the syntax...) or something like that to be able to throw?

src/CSVMsg.chpl
```diff
@@ -361,47 +465,13 @@ module CSVMsg {
             var locFiledoms = filedomains;
             /* On this locale, find all files containing data that belongs in
                 this locale's chunk of A */
-            for (filedom, filename, file_idx) in zip(locFiledoms, locFiles, 0..) {
+            forall (filedom, filename, file_idx) in zip(locFiledoms, locFiles, 0..) with (ref A, | reduce hadError) {
```
bmcdonald3, line 468: 🎉

bmcdonald3 (commented):
Awesome work Shreyas, glad to see this getting in!

Only real comment is on the use of `try!`, we want to avoid there being any way to crash the server, but not sure off the top of my head the best way to do so in this case, so curious on your thoughts there.

bmcdonald3:
Pinging @stress-tess in case they want to check this out prior to getting this merged.

src/CSVMsg.chpl
ShreyasKhandekar: Yes, we would only need this when the type being read in is a `string`. I can make it so that it doesn't compile the regex when we aren't read a `string`.

I will make the regex compilation throwing as well.

stress-tess (approved):
This is awesome!!! 🚀 

Thanks so much @ShreyasKhandekar! I don't see any issues, so I'm good with this being merged whenever

bmcdonald3 (approved):

PR #3239, opened 2024-05-24
Reduce file size for csvIO benchmark

ShreyasKhandekar:
The csvIO benchmark is timing out in multilocale runs due to bad performance of the current implementation. There is a PR up
(https://github.com/Bears-R-Us/arkouda/pull/3238) which will improve this implementation, but it may take a while to merge based on feedback. 

Therefore we will reduce the file size here to 1/5th of the current and bring it back up later once the performance improves. The timeout happens after 300 seconds and this size takes ~100 seconds so we should be safe 🤞 
For reference with the new implementation `./benchmarks/run_benchmarks.py csvIO -nl 16` takes `46.4425` seconds, so if the other PR is merged before this one, we don't need to merge this.

### DIFF
--- a/benchmarks/csvIO.py
+++ b/benchmarks/csvIO.py
@@ -17,7 +17,7 @@ def create_parser():
     parser.add_argument("hostname", help="Hostname of arkouda server")
     parser.add_argument("port", type=int, help="Port of arkouda server")
     parser.add_argument(
-        "-n", "--size", type=int, default=10**6, help="Problem size: length of array to read/write"
+        "-n", "--size", type=int, default=2*10**5, help="Problem size: length of array to read/write"
     )
     parser.add_argument(
         "-t", "--trials", type=int, default=1, help="Number of times to run the benchmark"


bmcdonald3:
Looks like CI is failing here, if you can get that fixed, I can merge this and then we get testing back over the weekend.

ShreyasKhandekar:
Hmm I don't think my change caused the CI failure.
The test that failed is `tests/setops_test.py::SetOpsTest::test_index_of`
``` 
FAILED tests/setops_test.py::SetOpsTest::test_index_of - ValueError: shape mismatch [23] [20]
```
And the error from the CI log is:
```
>               self.assertTrue((idx_of_first_in_second == ak.find(arr1, arr2, remove_missing=True)).all())

tests/setops_test.py:356: 
arkouda/pdarrayclass.py:733: in __eq__
    return self._binop(other, "==")

self = <[RuntimeError('client is not connected to a server') raised in repr()] pdarray object at 0x7f68e463e880>
other = <[RuntimeError('client is not connected to a server') raised in repr()] pdarray object at 0x7f68e4653430>
```

Which makes it seem like the CI failed due to some sort of server disconnect error. Could we try running the checks again to see if they pass?

bmcdonald3:
Interesting... I've restarted

bmcdonald3 (approved):

PR #3240, opened 2024-05-27
Fixes #3222: series of segarray print bug

### LINKED ISSUE #3222: failure printing series containing a segarray
I discovered this when working with dataframes containing segarrays since the recent change in df indexing (see https://github.com/Bears-R-Us/arkouda/pull/3109) has changed what is returned
```python
>>> test_df = ak.DataFrame({'test': ak.SegArray(ak.arange(10), ak.arange(10))})
>>> sa = test_df['test']
>>> sa
...
ValueError: 2

>>> sa.values
SegArray([
[0]
[1]
[2]
...
[7]
[8]
[9]
])
```

stress-tess:
This PR fixes #3222, the series of segaray print bug. The bug was due to pandas series erroring an ndarray where each element is a list, but it handles a list of lists just fine. So we just call `to_list` instead of `to_ndarray` for the segarray case

output is now what I would expect
```python
>>> test_df = ak.DataFrame({'test': ak.SegArray(ak.arange(0, 30, 3), ak.arange(30))})
>>> test_df['test']
0       [0, 1, 2]
1       [3, 4, 5]
2       [6, 7, 8]
3     [9, 10, 11]
4    [12, 13, 14]
5    [15, 16, 17]
6    [18, 19, 20]
7    [21, 22, 23]
8    [24, 25, 26]
9    [27, 28, 29]
dtype: object
```

### DIFF
--- a/PROTO_tests/tests/series_test.py
+++ b/PROTO_tests/tests/series_test.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pandas as pd
 import pytest
-from pandas.testing import assert_frame_equal
+from pandas.testing import assert_frame_equal, assert_series_equal
 
 import arkouda as ak
 from arkouda.series import Series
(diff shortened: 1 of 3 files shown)


jaketrookman (approved):

stress-tess:
> Looks good!
> 
> If this is a bug fix, should we have a unit test to cover the case that was causing the bug?

yah thats a good point. my initial reaction was it would be kinda hard since it was an error with it printing. but since the actual issue is in `to_pandas` we can just call that and compare against pandas

stress-tess:
@ajpotts, I added a test! I verified it fails without the fix and passes now. It should be ready for re-review

ajpotts (approved):
Thanks!

PR #3241, opened 2024-05-28
Fixes #3234: segarray with empty segments and nans parquet bug

### LINKED ISSUE #3234: segarray with empty segments and `nan`s
The fix from https://github.com/Bears-R-Us/arkouda/pull/3233 doesn't seem to handle empty segments and `nan`s (despite this working locally). I'll dig into this more but want the fix for ints to be merged asap since it was a user request

stress-tess:
Fixes #3234 

We found some odd behavior with the existing implementation, so I've opted not to squash my commits in order to make it easier to reproduce in case we want to dig into it more. With a set seed running the same code and versions, it seemed to consistently pass on arm64 macs and fail on x86 linux machines. It not immediately obvious what about the implementation was causing this seemingly architecture specific failure

To sidestep this, I opted for simpler single batch read logic for float/double segarrays. This will be slower but obviously correctness is most important. This is the latest in a string of read logic that I've changed to be single batched, so it might be worth revisiting in the future to see if we can reenable batch reads

### DIFF
--- a/src/ArrowFunctions.cpp
+++ b/src/ArrowFunctions.cpp
@@ -585,20 +585,6 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
         int16_t definition_level; // needed for any type that is nullable
 
         std::shared_ptr<parquet::ColumnReader> column_reader = row_group_reader->Column(idx);
-
-
-
-
-
-
-
-
-
-
-
-
-
-
         if(lty == ARROWINT64 || lty == ARROWUINT64) {
           int16_t definition_level; // nullable type and only reading single records in batch
           auto chpl_ptr = (int64_t*)chpl_arr;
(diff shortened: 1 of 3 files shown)


src/ArrowFunctions.cpp
```diff
@@ -644,85 +644,57 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
...
-                chpl_ptr[i+val_idx] = NAN;
-                tmp_offset++; // adjustment for values array since Nulls are not included
-              }
-              else if (def_lvl[j] == 3){ // defined value to write
-                chpl_ptr[i+val_idx] = (double)tmpArr[val_idx-tmp_offset];
+          while (reader->HasNext() && arrayIdx < numElems) {
+            (void)reader->ReadBatch(1, &definition_level, nullptr, &chpl_ptr[arrayIdx], &values_read);
+            // if values_read is 0, that means that it was a null value or empty seg
+            if (values_read != 0) {
+              arrayIdx++;
```
ajpotts, line 688: Why is this different than the `ARROWFLOAT` version?  It's missing this line: `chpl_ptr[arrayIdx] = (double) tmp;`

ajpotts (approved):
Other than the one comment, it looks OK to me!  I'm not familiar with the Arrow API though :)

src/ArrowFunctions.cpp
```diff
@@ -644,85 +644,57 @@ int cpp_readListColumnByName(const char* filename, void* chpl_arr, const char* c
...
-              
-              // Null values treated as NaN
-              if (def_lvl[j] == 2) {
-                chpl_ptr[i+val_idx] = NAN;
-                tmp_offset++; // adjustment for values array since Nulls are not included
-              }
-              else if (def_lvl[j] == 3){ // defined value to write
-                chpl_ptr[i+val_idx] = (double)tmpArr[val_idx-tmp_offset];
+          while (reader->HasNext() && arrayIdx < numElems) {
+            (void)reader->ReadBatch(1, &definition_level, nullptr, &chpl_ptr[arrayIdx], &values_read);
```
bmcdonald3, line 685: I am crying every time I see one of these batch sizes get reduced to 0. I think I know a way around that so we can get the better performance while still checking for nulls (it's similar to my string PR up right now).

We should queue that up for a discussion after ChapelCon gets over.

bmcdonald3 (approved):
Looks good, unfortunate we are reducing these batch sizes, we can try to fix that in the future.

src/ArrowFunctions.cpp
stress-tess: yah so float is like the 32 bit version of real and double is the 64 bit version.  So we turn the floats into doubles. You can see something similar done to convert the `(u)int32`s into `(u)int64`s

src/ArrowFunctions.cpp
stress-tess: I know 😭

jaketrookman (approved):

PR #3242, opened 2024-05-28
Array API Documentation

jeremiah-corrado:
Get the Array API module's documentation up to date

### DIFF
--- a/arkouda/array_api/__init__.py
+++ b/arkouda/array_api/__init__.py
@@ -26,8 +26,8 @@ from .creation_functions import (
 from .data_type_functions import (
     astype,
     can_cast,
-    # finfo,
-    # iinfo,
+    finfo,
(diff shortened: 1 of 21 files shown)


ajpotts (approved):
Looks good.  Thanks for adding these doc strings!

I did notice a lot of them are missing Return and/or Parameter sections.  Is the idea to add those at a later time?

stress-tess (approved):
Looks great! Also it was a nice opportunity to catch up on all the functionality you've added

jaketrookman (approved):

jeremiah-corrado:
> I did notice a lot of them are missing Return and/or Parameter sections. Is the idea to add those at a later time?
@ajpotts , yeah, I was focusing on getting some decent docs in place for the time being, and then I think they can improve over time.

PR #3248, opened 2024-05-29
Remove single-column cases from multi-col-merge test. 

brandon-neth:
The single-column cases are handled by `test_merge`, so I removed them from `test_multi_col_merge`. This should improve overall test duration, as `test_multi_col_merge` is far and away the longest test case. On my mac, the test went from ~5 minutes to ~2.

### DIFF
--- a/PROTO_tests/tests/dataframe_test.py
+++ b/PROTO_tests/tests/dataframe_test.py
@@ -1016,9 +1016,6 @@ class TestDataFrame:
 
             for how in "inner", "left", "right":
                 for on in (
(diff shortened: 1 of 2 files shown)


ajpotts:
It's OK with me, but the change should be to the PROTO_tests version as the tests folder will be deprecated soon.

brandon-neth:
> It's OK with me, but the change should be to the PROTO_tests version as the tests folder will be deprecated soon.

Done!

stress-tess (approved):

PR #3253, opened 2024-05-30
Closes #3245: Add `poisson` distribution to random number generators

### LINKED ISSUE #3245: add `poisson` to random number generators
add poisson to random number generators 

Relevant docs:
[numpy.random.Generator.poisson](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.poisson.html)

stress-tess:
This PR (closes #3245) adds the `poisson` distribution to random number generators. This also adds the docs and testing including hypothesis testing against the probabilities expected from the poisson probability mass function. This PR does a coforall over the locals and tasks so I can ensure each task gets a unique seed that is reproducable

I also moved one of the other hypothesis tests so they are all together

### DIFF
--- a/PROTO_tests/tests/random_test.py
+++ b/PROTO_tests/tests/random_test.py
@@ -1,3 +1,5 @@
+from collections import Counter
+
 import numpy as np
 import pytest
 from scipy import stats as sp_stats
@@ -128,31 +130,6 @@ class TestRandom:
         assert all(bounded_arr.to_ndarray() >= -5)
         assert all(bounded_arr.to_ndarray() < 5)
 
-    def test_choice_hypothesis_testing(self):
-        # perform a weighted sample and use chisquare to test
-        # if the observed frequency matches the expected frequency
-
-        # I tested this many times without a set seed, but with no seed
-        # it's expected to fail one out of every ~20 runs given a pval limit of 0.05.
-        rng = ak.random.default_rng(43)
-        num_samples = 10**4
-
-        weights = ak.array([0.25, 0.15, 0.20, 0.10, 0.30])
-        weighted_sample = rng.choice(ak.arange(5), size=num_samples, p=weights)
-
-        # count how many of each category we saw
-        uk, f_obs = ak.GroupBy(weighted_sample).size()
-
-        # I think the keys should always be sorted but just in case
-        if not ak.is_sorted(uk):
-            f_obs = f_obs[ak.argsort(uk)]
-
-        f_exp = weights * num_samples
-        _, pval = akchisquare(f_obs=f_obs, f_exp=f_exp)
-
-        # if pval <= 0.05, the difference from the expected distribution is significant
-        assert pval > 0.05
-
     def test_choice(self):
         # verify without replacement works
         rng = ak.random.default_rng()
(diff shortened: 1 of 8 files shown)


src/RandMsg.chpl
```diff
@@ -10,6 +10,7 @@ module RandMsg
     use Logging;
     use Message;
     use RandArray;
+    use Unique;
```
jeremiah-corrado, line 13: I'm guessing this is no longer needed?

src/RandMsg.chpl
```diff
@@ -547,6 +548,101 @@ module RandMsg
...
+        // https://en.wikipedia.org/wiki/Poisson_distribution#Random_variate_generation
+        // generates values drawn from poisson distribution using a stream of uniformly distributed random numbers
+
+        const nTasksPerLoc = here.maxTaskPar; // tasks per locale based on locale0
+        const Tasks = {0..#nTasksPerLoc}; // these need to be const for comms/performance reasons
+
+        const generatorSeed = (rng.next() * 2**62):int;
+        var poissonArr = makeDistArray(size, int);
+
+        // I hate the code duplication here but it's not immediately obvious to me how to avoid it
```
jeremiah-corrado, line 583: Not sure if it's worth all the overhead, but you could use a record type that can store either a single lambda value, or an array of lambdas s.t. `const lam = lamArr[locSubDom.low + i];` always works:
```chapel
record lambda {
  var isArray: bool;
  var arr;
  var val: real;

  proc init(arr: [] real) {
    this.isArray = true;
    this.arr = arr;
  }

  proc init(val) {
    this.isArray = false;
    this.arr = makeDistArray(0..0);
    this.val = val:real;
  }

  proc this(idx) {
    return if isArray then this.arr[idx] else this.val;
  }
}
```

And then initialize it differently depending on the value of `isSingleLam`:
```chapel
var lamArr = if isSingleLam then new lambda(lamStr)
  else new lambda(toSymEntry(getGenericTypedArrayEntry(lamStr, st),real).a);
```

jeremiah-corrado (approved):
Looks good! Just a couple of suggestions:

src/RandMsg.chpl
stress-tess: oops! yep ur right!

src/RandMsg.chpl
stress-tess: oh yah I like this a lot! thanks for this! I bet theres quite a few places in the codebase that benefit from something like this

jaketrookman (approved):

PROTO_tests/tests/random_test.py
```diff
@@ -246,6 +264,33 @@ def test_normal_hypothesis_testing(self):
...
+        while (1 - sp_stats.poisson.cdf(num_elems, mu=lam)) > tol:
+            num_elems += 5
+
+        obs_counts = np.array([0] * num_elems)
+        for k, v in count_dict.items():
+            obs_counts[k] = v
+
+        # use the probability mass function to get the probability of seeing each value
+        # and multiply by num_samples to get the expected counts
+        exp_counts = np.array([sp_stats.poisson.pmf(v, mu=lam) for v in range(num_elems)]) * num_samples
```
ajpotts, line 290: This could also be written `sp_stats.poisson.pmf(range(num_elems), mu=lam) * num_samples`.

ajpotts (approved):
Looks good.  Thanks!

PR #3258, opened 2024-05-31
3231 unique unit tests

### LINKED ISSUE #3231: unit tests for unique
Add unit test coverage for `unique` in groupbyclass.py (this is different than `GroupBy.unique`).

Here is the relevant chapel code block:
https://github.com/Bears-R-Us/arkouda/blob/41fec72befe88697360e05b023e5b2e6abee00ec/src/UniqueMsg.chpl#L132

In the unit tests, make sure the verify that the returned keys are sorted in the `int64` case.

drculhane:
Adds PROTO-tests unit test for ak.unique.

Closes #3231

As discussed over Slack and in-person, it took me a bit of study to realize this was much simpler than I was making it out to be.  Although the 3 booleans make for 8 possible test cases, they're not independent.  As noted in the comments, these two cases are supersets of all the others:

return_groups = True, assume_sorted = True, return_indices = False
return_groups = True, assume_sorted = False, return_indices = False

So I am only testing those.  Strangely, the indices are returned regardless of the value of return_indices.  In fact, more values are returned when it's False than when it's True.

The test covers int, float, str and categorical, and makes use of parameterization.  There are (hopefully) ample comments to make it clear what I'm testing.

### DIFF
--- a/PROTO_tests/tests/groupby_test.py
+++ b/PROTO_tests/tests/groupby_test.py
@@ -5,7 +5,36 @@ import pytest
 import arkouda as ak
 from arkouda.groupbyclass import GroupByReductionType
 from arkouda.scipy import chisquare as akchisquare
+from arkouda.dtypes import npstr
 
+#  block of variables and functions used in test_unique
+
+UNIQUE_TYPES = [ak.categorical,ak.int64,ak.float64,npstr]
+VOWELS_AND_SUCH = ['a','e','i','o','u','AB',47,2,3.14159]
+PICKS = np.array([f"base {i}" for i in range(10)])
+
+isSorted = lambda x : np.all(x[:-1] <= x[1:])  # short for is x[i] <= x[i+1] for all i
+
+#  This function (almost) guarantees both a sorted and unsorted version of
+#  a 1d array.  The only exception is an array of all identical values.
+#  The first "if" block skips the whole function in that case.  Otherwise,
+#  if the sample is already sorted, a non-sorted permutation is generated,
+#  and the two are returned.  If it isn't, a sorted version is created,
+#  and those two are returned.
+
+def make_sorted_and_unsorted_data (sample) :
+    if np.all(sample == sample[0]) : return sample, sample
+    if isSorted(sample) :
+        s_a = sample[:]
+        us_a = np.random.permutation(sample)
+        while isSorted(us_a) :
+            us_a = np.random.permutation(us_a)
+    else :
+        s_a = np.sort(sample)
+        us_a = sample[:]
+    return s_a, us_a
+
+#  end of block
 
 def to_tuple_dict(labels, values):
     # transforms labels from list of arrays into a list of tuples by index and builds a dictionary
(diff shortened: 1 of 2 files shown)


PROTO_tests/tests/groupby_test.py
```diff
@@ -654,6 +687,82 @@ def test_large_mean_aggregation(self):
...
+        assert check1 and check2
+
+        # for unsorted data, a bit more work is required.  A reordered
+        # copy of the pdarray is created based on the returned permutation,
+        # and the indices are used to confirm that elements within a segment
+        # match.
+
+        aku = ak.unique(us_pda).to_ndarray()
+        reordering = ak_TFF[1].to_ndarray()
+        reordered = [nda[i] for i in reordering]
```
ajpotts, line 752: If you don't convert to ndarray first, this becomes: 

```
aku = ak.unique(us_pda)
reordered = us_pda[ak_TFF[1]]
ak.all (reordered == ak.broadcast(indices, aku, ak_TFF[1].size) )
```

So 15 lines can be done in 3 lines....

ajpotts: You might also be able to simplify the one above with something similar:
`ak.broadcast(indices, ak.array(np_unique), len(nda_list))`

drculhane: In your first comment above, do you mean "assert ak.all (...)"

drculhane: Writing it that way (assert ak.all(...) ) passes on categoricals, ints and floats, but fails on strings.  I'm going to revert for now.

ajpotts (approved):

jaketrookman (approved):

PROTO_tests/tests/groupby_test.py
drculhane: Okay, my mistake.  I had one lousy typo.  Your suggestions both work, and make the code much more concise.

ajpotts (approved):

PROTO_tests/tests/groupby_test.py
```diff
@@ -654,6 +687,73 @@ def test_large_mean_aggregation(self):
...
+    #      Combinations TTF and TFF are supersets of all other possible
+    #      combinations, so only those are tested below.
+
+    @pytest.mark.parametrize("data_type", UNIQUE_TYPES)
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_unique(self, data_type, prob_size):
+        Jenny = pytest.seed if pytest.seed is not None else 8675309
+        T = True
+        F = False
+        np.random.seed(Jenny)
```
stress-tess, line 702: since we only use `Jenny` in one place I think it makes sense to just do:
```python
np.random.seed(pytest.seed if pytest.seed is not None else 8675309)
```
Also I think it's not super necessary to alias True and False, but this is all just stylistic

stress-tess (approved):
looks good! A stylistic comment but I don't think matters enough to hold up a PR especially since it's in a test

PR #3280, opened 2024-06-04
poisson cleanup

stress-tess:
small updates to https://github.com/Bears-R-Us/arkouda/pull/3253 found during team code share / knowledge transfer:
* changed a `for` loop to a `coforall`
* make use of the inverse survival function instead of iterating `1-cdf` to ensure we're within the tolerance required by `chisquare`


ajpotts (approved):
Looks great!

PR #3285, opened 2024-06-05
Reduces some code duplication in poisson implementation

drculhane:
As the title says, this uses the ziggurat function I shared over Slack to reduce _some_ of the code duplication in RandMsg.chpl.  This felt like a worthy thing to look into, especially with Tess's comments about not liking the amount of code duplication.  It was a bit of a challenge.  The two cases where ziggurat is used are slightly different, but there was enough similarity to create a function that both cases can use.  So if this is worthy, that's terrific, and if not, well, it gave me a chance to collaborate with Jake, and to revisit how to pass objects in Chapel.

### DIFF
--- a/src/RandMsg.chpl
+++ b/src/RandMsg.chpl
@@ -549,34 +549,34 @@ module RandMsg
 
     proc poissonGeneratorMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
 
-	// this is an idea
-	proc ziggurat (lam : real, ref rs : ? ) : int {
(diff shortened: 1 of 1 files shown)


stress-tess:
thanks so much for looking into this! I don't think it's needed any longer since https://github.com/Bears-R-Us/arkouda/pull/3352 has been merged removes the need for duplicating the code in the first place

stress-tess:
thanks so much for looking into this! I don't think it's needed any longer since https://github.com/Bears-R-Us/arkouda/pull/3352 has been merged removes the need for duplicating the code in the first place

PR #3303, opened 2024-06-06
Drop support for Chapel 131

bmcdonald3:
Chapel 1.31 is no longer in need for support and was causing some compatability changes with some new code, so we've made the decision to drop 1.31.


ajpotts (approved):
Looks OK to me.  Thanks for taking care of this!

PR #3306, opened 2024-06-07
Clean up Compat modules

jeremiah-corrado:
Many of the compatibility modules are no longer needed because they are either empty or contain the same code for all supported versions of Chapel. This PR removes the unnecessary compat modules and updates the codebase accordingly.

With this change, we only have the following four compat modules:
* ArkoudaAryUtilCompat (can be removed when 1.32 support is dropped)
* AkoudaIOCompat
* ArkoudaRandomCompat
* ArkoudaSortCompat

- [x] tests passing w/ gasnet build


src/StatsMsg.chpl
```diff
@@ -343,7 +343,7 @@ module StatsMsg {
             var eOut = st.addEntry(rname, outShape, t);
 
             const DD = domOffAxis(eIn.a.domain, axis);
-            forall idx in DD {
```
stress-tess, line 346: just want to make sure this bit is intentional since it seems unrelated to the compat modules

stress-tess: I didn't highlight all of it apparently but the `forall` is being turned into a `for`

jeremiah-corrado: yeah, that change fixed an error I was seeing for nd-array builds. Just happened to notice it while working on this.

PR #3310, opened 2024-06-10
Closes #3183, #3364: Add `exponential` distribution and aggregation to random generator loop

### LINKED ISSUE #3183: Add `exponential` and `standard_exponential` to random number generators
Add `exponential` and `standard_exponential` to our random number generators

Relevant numpy docs:
- [numpy.random.Generator.standard_exponential](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.standard_exponential.html)
- [numpy.random.Generator.exponential](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.exponential.html)

### LINKED ISSUE #3364: Add copy aggregation to `uniformStreamPerElem`
I think `uniformStreamPerElem` would benefit from copy aggregation. Especially in the case where a chunk a random stream is assigned to bleeds over to the next locale bc we solved this by having the current locale "steal" those indices. This at the very least would result in some PUTs to write those results to the correct locale. This may be the only communication we would expect, in which case we'd only need to create an aggregator for the last chunk on a locale, but I'd need to do some digging before I'm convinced of that.

I'd also definitely need feedback on this section from someone who understands aggregation a bit better than me since the only example I see that uses it with coforalls is radixSortLSD. But that creates an aggregator per task and we'd be creating one per set number of elements (same as when we create a random stream) so idk how this would change the worthwhileness of it

stress-tess:
This PR (closes #3183) adds `exponential` and `standard_exponential` to random number generators implementing both the inverse cdf and ziggurat methods. Adds testing and necessary constants for ziggurat method as tuples (which is necessary for reducing comm, see #2354)

The constants are from: https://github.com/numpy/numpy/blob/main/numpy/random/src/distributions/ziggurat_constants.h

and the implementation relied heavily on numpy's implementation found here:
https://github.com/numpy/numpy/blob/92412e93e291392b507d0d49914a2bc1866d308d/numpy/random/src/distributions/distributions.c#L46-L73

Also (closes #3364) by adding copy aggregation to `uniformStreamPerElem`. When a chunk a random stream is assigned to overflows onto the next locale, we have the current locale "steal" those indices. This results in some PUTs to write those results to the correct locale after they are generated.

This seems to be the only communication based on my local testing. At first, I created a copy aggregator every time I created a randomStream (every `elemsPerStream` number of elements). I then compared these communication numbers to creating a copy aggregator for only the last chunk on a locale and they didn't change. So I left that implementation, this will hopefully reduce start up costs since we only create `numLocales` aggregators instead of `total_size / elemsPerStream` aggregators

I'd also definitely like feedback from someone who understands aggregation a bit better than me to sanity check that I'm not doing anything dumb since I normally only use them with `forall` loops. I based this off radixSortLSD since it also uses aggregations with coforalls. But that creates an aggregator per task so idk how this would change things


ajpotts (approved):
Looks great!

jaketrookman (approved):
Looks good

src/RandMsg.chpl
drculhane, line 1: I didn't find the "tip" part of the "fallback for the tail" at the Wikipedia site.  Do you have a reference for that?

stress-tess:
converting to a draft until https://github.com/Bears-R-Us/arkouda/issues/3322 can be resolved

src/RandMsg.chpl
```diff
@@ -332,6 +333,128 @@ module RandMsg
...
+        https://blogs.mathworks.com/cleve/2015/05/18/the-ziggurat-random-normal-generator/
+
+        This implementation based on numpy's:
+        https://github.com/numpy/numpy/blob/main/numpy/random/src/distributions/distributions.c
+
+        Uses constants from lists of size 256 (number of rectangles) found in ZigguratConstants.chpl
+    */
+    inline proc standardExponentialZig(ref realRng, ref uintRng): real {
+        // modified from numpy to use while loop instead of recusrsion so we can inline the proc
+        while true {
```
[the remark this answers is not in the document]
stress-tess: absolutely! exiting at the core should be a probability of `98.9%` so we could just throw a cap of like 100 or something. The odds of the sampling failing 100 times in a row would be 

$$ 1 - (.989)^{100} = 1.378061 * 10^{-196}$$

which I consider to be sufficiently unlikely :)

src/RandMsg.chpl
stress-tess: yah! we talked about it a bit in person, the tip part didn't come from the wikipedia but from here:
https://blogs.mathworks.com/cleve/2015/05/18/the-ziggurat-random-normal-generator/

it doesn't actually give the computation but just say `r = tip_computation`. We use the computation from numpy's `standard_exponential_unlikely` function found here:
https://github.com/numpy/numpy/blob/882611cf11e1925bd8f3a8d7aa00d74675a7601c/numpy/random/src/distributions/distributions.c#L46

let me know if these links are helpful or if it would be useful to chat about it more!

stress-tess: the computaiton for the non-tip part (the core) can be found at the same numpy link in the `standard_exponential_likely` function

stress-tess:
@Bears-R-Us/arkouda-core-dev this is ready for re-review!

src/ZigguratConstants.chpl
```diff
@@ -0,0 +1,721 @@
...
+            ke_double
+            we_double
+            fe_double
+    */
+
+    const ziggurat_nor_r = 3.6541528853610087963519472518;
+    const ziggurat_nor_inv_r = 0.27366123732975827203338247596; // 1.0 / ziggurat_nor_r;
+    const ziggurat_exp_r = 7.6971174701310497140446280481;
+
+    const ki_double = (
```
bmcdonald3, line 26: I wonder if making these large tables `param` values would be beneficial, though, the compiler is likely already doing this since it's just literal values.

bmcdonald3 (approved):
Very cool, learning some stuff about science here, but not a ton of feedback on the code.

jaketrookman (approved):
Looks good

jeremiah-corrado (approved):

src/ZigguratConstants.chpl
stress-tess: I went ahead and converted the ones I could. The compiler said the float tuples is not a supported param type

PR #3314, opened 2024-06-11
Fix CSV Reading Bugs

ShreyasKhandekar:
Resolves failures with CSV testing in multilocale configurations where the files do not correspond to a single locale's data.

Some files have data meant to be distributed across multiple locales
therefore we have to be aware of the lower bound of the line indices to
be read in. In order to do this, we skip over the lines until this lower
bound of the intersection is reached. 

Since the domains are continuous, we don't need to worry about any other bounds.


bmcdonald3 (approved):
Looks good, thanks Shreyas. Just to confirm, you validated that this now passes the unit testing for multi locale?

PR #3316, opened 2024-06-12
Part of #3229: Re-enable find implementation of `indexof1d`

stress-tess:
This PR is part of #3229. Re-enable the `find` implementation. I modified it to have a more optimized way of finding all occurrences. Part of the trade off for the more optimized approach is it's not as easy to replace missing elements with `-1`. Since this isn't a common case and no one is asking for this functionality I decided to have `all_occurrences` being set automatically enable `replace_missing`. I tried to make this clear in the doc string. While I'm not sure if this is the best way of calculating this, it's no longer the bottleneck and it gets better performance than the current implementation.

I'm hopeful the updated find way will also solve the issue we were seeing in the CI. But previously I added seeds to the `indexof1d` test, so it will be easier to reproduce if any remain failures. I'm not closing the issue yet since there's still some code cleanup to be done, and I want to leave existing code to make it easier to fall back to a similar approach in the event that the `find` way still has issues

### DIFF
--- a/arkouda/alignment.py
+++ b/arkouda/alignment.py
@@ -126,18 +126,19 @@ def find(query, space, all_occurrences=False, remove_missing=False):
         occurrences as a pdarray. Defaults to only finding the first occurrence.
         Finding all occurrences is not yet supported on sequences of arrays
     remove_missing: bool
+        If all_occurrences is True, remove_missing is automatically enabled.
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):

ajpotts (approved):

brandon-neth (approved):
Doc string is clear to me!

Issue #3297, opened 2024-06-05
array_equal


ajpotts:
Implement an `array_equal` function to mirror numpy:
https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html

PR #3319, opened 2024-06-12
Adds array_equal to match numpy

drculhane:
Implements an array_equal function for two pdarrays, matching what numpy.array_equal does.

Also adds a utility function "matches_shape" which for now only compares pdarray sizes, but which will serve as a placeholder for when we expand pdarrays to n-dimensions.

### DIFF
--- a/PROTO_tests/tests/numeric_test.py
+++ b/PROTO_tests/tests/numeric_test.py
@@ -636,6 +636,14 @@ class TestNumeric:
         assert h1.to_list() == h3.to_list()
         assert h2.to_list() == h4.to_list()
 
+    @pytest.mark.parametrize("prob_size",pytest.prob_size)
+    def test_matching_shapes(self,prob_size) :
+        a = ak.randint(0,10,prob_size)
+        b = ak.randint(0,10,prob_size)
+        assert ak.matching_shapes(a,b)
+        b = ak.randint(0,10,prob_size+1)
+        assert not(ak.matching_shapes(a,b))
+
     # Notes about array_equal:
     #   Strings compared to non-strings are always not equal.
     #   nan handling is (of course) unique to floating point
@@ -654,6 +662,8 @@ class TestNumeric:
             pda_a = ak.array(temp)
             pda_b = ak.array(temp)
             assert ak.array_equal(pda_a, pda_b)  # matching string arrays
+            pda_c = pda_b[:-1]
+            assert not (ak.array_equal(pda_a,pda_c))   # matching except c is shorter by 1
             temp = np.random.choice(VOWELS_AND_SUCH, prob_size)
             pda_b = ak.array(temp)
             assert not (ak.array_equal(pda_a, pda_b))  # mismatching string arrays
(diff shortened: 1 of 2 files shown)


arkouda/numeric.py
```diff
@@ -1722,6 +1725,89 @@ def where(
     return create_pdarray(type_cast(str, repMsg))
```
ajpotts, line 1727: Before I review these, can you add unit tests to the PROTO_tests directory?

drculhane: Brainfade.  I had done that, and I think I lost it in the rebasing.  I'll reconstruct those tests later today.

arkouda/numeric.py
```diff
@@ -1722,6 +1725,89 @@ def where(
...
+    """
+    For now, this will only check that two pdarrays have matching sizes.
+    It's included so that we can put placeholders in for the check as
+    we implement n-dimensional array computation.
+
+    Parameters:
+        pda_a : pdarray
+        pda_b : pdarray
+
+    Returns:
```
ajpotts, line 1738: Same comment as below about the Return section starting with the return type.

arkouda/numeric.py
```diff
@@ -1722,6 +1725,89 @@ def where(
...
+    If equal_Nan is False, then any nan element in either array gives a False return.
+    If equal_Nan is True, then pairwise-corresponding nans are considered equal.
+
+    Parameters:
+        pda_a : pdarray
+        pda_b : pdarray
+        equal_nan : boolean to determine how to handle nans, default False
+
+    Returns:
+        With string data: False if one array is of type npstr and the other isn't,
```
ajpotts, line 1772: Typically we let the reviewer resolve the conversation so they can check what was changed.

arkouda/numeric.py
```diff
@@ -1722,6 +1725,89 @@ def where(
...
+                            True is both are npstr and they match.
```
ajpotts, line 1773: So, I can tell the Return section is not quite right because when I `make doc` on this and then navigate in the web browser, the `Return Type` section is missing.  Also, `and all nans in pda_a correspond to nans in pda_b` is highlighted in a pale blue.  I'm sorry, I know that sphinx is really fussy and it can be frustrating to deal with.

arkouda/numeric.py
ajpotts: This is still not rendering correctly for me.  The correct formatting is:
```
    Returns
    -------
    boolean
        True if pda_a.size == pda_b.size, else False
```

arkouda/numeric.py
ajpotts: This is still not rendering correctly for me.  Perhaps you were going for something like this:
```
    Returns
    -------
    boolean

        With string data:
            False if one array is of type npstr and the other isn't,

            True is both are npstr and they match.

        With numeric data:
            True if neither array has any nan elements, and all elements pairwise equal.

            False if equal_Nan is False, and either array has any nan element.

            True if equal_Nan is True, all non-nan elements are pairwise equal,
            and all nans in pda_a correspond to nans in pda_b

```

arkouda/numeric.py
```diff
@@ -1722,6 +1725,87 @@ def where(
     return create_pdarray(type_cast(str, repMsg))
 
 
+def matching_shapes(pda_a: pdarray, pda_b: pdarray):
+    """
+    For now, this will only check that two pdarrays have matching sizes.
+    It's included so that we can put placeholders in for the check as
+    we implement n-dimensional array computation.
```
stress-tess, line 1732: is this a function in numpy? If not, I'm not sure how I feel about it being exposed to the user. We could make this a hidden helper by appending a `_`, so `_matching_shapes`

But I think instead of having a function for this we could just do `pda_a.shape == pda_b.shape` and cover the single and multi-dim cases

arkouda/numeric.py
```diff
@@ -1722,6 +1725,87 @@ def where(
...
+    False
+    >>> ak.array_equal(a,b,True)
+    True
+    """
+    if not matching_shapes(pda_a, pda_b):
+        return False
+    elif (pda_a.dtype is npstr and not (pda_b.dtype is npstr)) or (
+        pda_b.dtype is npstr and not (pda_a.dtype is npstr)
+    ):
+        return False
```
stress-tess, line 1802: Could we just combine these two to be:
```python
if (pda_a.shape() != pda_b.shape()) or (pda_a.dtype is npstr ^ pda_b.dtype is npstr):
    return False
```

PROTO_tests/tests/numeric_test.py
```diff
@@ -639,6 +636,61 @@ def test_hash(self):
...
+            pda_b = ak.array(nda_b) if same_size else ak.array(nda_b[:-1])
+            assert ak.array_equal(pda_a, pda_b, nan_handling) == (matching and same_size)
+        else:  # other types have simpler tests
+            pda_a = ak.random.randint(0, 100, prob_size, dtype=data_type)
+            if matching:  # known to match?
+                pda_b = pda_a if same_size else pda_a[:-1]
+                assert (ak.array_equal(pda_a, pda_b) == (matching and same_size))
+            #elif same_size:  # not matching, but same size?
+            #    pda_b = ak.random.randint(0, 100, prob_size, dtype=data_type)
+            #    assert not (ak.array_equal(pda_a, pda_b))
```
stress-tess, line 689: I noticed this bit was commented out. I wanted to make sure it shouldn't be uncommented or deleted

stress-tess:
It's helpful when keeping track of things to use closing keywords in your title and description of your PR. So add `closes #3297` or something along those lines. 

Having it in the description links the PR with the pull request so the issue will automatically be closed when the PR is merged. And having it in the title makes it easy to tell at a glance which issue it is associated with and is helpful when writing release notes

ajpotts:
`array_equal` was added in this PR: https://github.com/Bears-R-Us/arkouda/pull/3725

PR #3323, opened 2024-06-13
Reverts DataFrame Indexing Changes

brandon-neth:
Because of performance regressions, Ben and I determined that reverting the dataframe indexing updates was the best path forward to get a release out this week. This PR reverts those updates. 

This should both fix the performance regressions and return the code base to a state where a `.values` call is not needed on each DataFrame indexing call. 

Tested with CHPL_COMM={gasnet,none} for regular tests, proto tests, and benchmarks.


stress-tess (approved):

bmcdonald3 (commented):
Looks good, thanks for putting this together.

I just started a build on Horizon to make sure the performance was restored, but will merge once that comes back good 👍

bmcdonald3 (approved):
```
array size = 10,000
number of trials =  1
>>> arkouda dataframe display
numLocales = 16, N = 160,000
  _get_head_tail_server Average time = 0.9649 sec
  _get_head_tail_server Average rate = 0.03 GiB/sec
  _get_head_tail Average time = 1.9099 sec
  _get_head_tail Average rate = 0.01 GiB/sec
```
Looks good

PR #3333, opened 2024-06-14
Closes #3332: Split Parquet code into multiple files

### LINKED ISSUE #3332: Split Parquet code into multiple files
The Parquet code currently exists as a single monolithic file, which makes it difficult to manage and is not good practice.

By splitting the Parquet code into multiple files, each limited to a distinct functionality, we can make the code easier to maintain/keep track of/modify. Initially, I am splitting the code into three files (1) read (2) write (3) utility.

There will be follow up PRs to further improve the readability/maintainability of this code.

bmcdonald3:
The Parquet code currently exists as a single monolithic file, which makes it difficult to manage and is not good practice.

By splitting the Parquet code into multiple files, each limited to a distinct functionality, we can make the code easier to maintain/keep track of/modify. Initially, I am splitting the code into three files (1) read (2) write (3) utility.

There will be follow up PRs to further improve the readability/maintainability of this code.

Closes #3332


src/parquet/UtilParquet.cpp
```diff
@@ -0,0 +1,1083 @@
...
+      } else {
+        fields.push_back(parquet::schema::PrimitiveNode::Make(cname_ptr[i], parquet::Repetition::OPTIONAL, parquet::Type::BYTE_ARRAY, parquet::ConvertedType::NONE));
+      }
+    }
+  }
+  return std::static_pointer_cast<parquet::schema::GroupNode>(
+      parquet::schema::GroupNode::Make("schema", parquet::Repetition::REQUIRED, fields));
+}
+
+int cpp_writeMultiColToParquet(const char* filename, void* column_names,
```
stress-tess, line 345: should this be in `WriteParquet` instead of `UtilParquet`? It seems to be in both the write and util header files with the `c_writeMultiColToParquet` being in `WriteParquet`

stress-tess (approved):
Looks good other than one function which it seems was meant to be moved to write

src/parquet/UtilParquet.cpp
bmcdonald3: Good catch! Should be fixed now.

stress-tess (approved):

PR #3335, opened 2024-06-14
Closes #3334, #3351: Simplify server side string code and added fixed length

### LINKED ISSUE #3334: Simplify server side string code and added fixed length
In an effort to simplify the Parquet string reading code
to better understand the performance implications of
changes to that code, this PR switches back to the simpler
way of doing things and also adds a "fixed_len" arg on
the client side to skip byte calculation in cases where
the size of each string is known at runtime and
consistent in the entire file.

### LINKED ISSUE #3351: Multi-batch read test failures
We're seeing failures in the nightly for the multi-batch read tests added in #3350:
```python
=========================== short test summary info ============================
FAILED tests/parquet_test.py::ParquetTest::test_multi_batch_reads - AssertionError: assert False
 +  where False = <bound method pdarray.all of array([True True True ... True True True])>()
 +    where <bound method pdarray.all of array([True True True ... True True True])> = array(['HH', 'JME', 'EHG', ... , '', 'GRG', '']) == array(['HH', 'JME', 'EHG', ... , '', 'GRG', '']).all
====== 1 failed, 640 passed, 1 skipped, 54 warnings in 2154.92s (0:35:54) ======
```

bmcdonald3:
In an effort to simplify the Parquet string reading code
to better understand the performance implications of
changes to that code, this PR switches back to the simpler
way of doing things and also adds a "fixed_len" arg on
the client side to skip byte calculation in cases where
the size of each string is known at runtime and
consistent in the entire file.

Additionally, the IO benchmark is updated to properly
handle strings for byte calculation.

Closes #3334 
Fixes #3351

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -814,6 +814,8 @@ module ParquetMsg {
     var tagData: bool = msgArgs.get("tag_data").getBoolValue();
     var strictTypes: bool = msgArgs.get("strict_types").getBoolValue();
 
+    var fixedLen = msgArgs.get('fixed_len').getIntValue() + 1;
+
     var allowErrors: bool = msgArgs.get("allow_errors").getBoolValue(); // default is false
     var hasNonFloatNulls: bool = msgArgs.get("has_non_float_nulls").getBoolValue();
     if allowErrors {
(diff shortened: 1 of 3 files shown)


bmcdonald3:
Noting that this PR should be merged after https://github.com/Bears-R-Us/arkouda/pull/3333 and rebased on top of that

src/ParquetMsg.chpl
```diff
@@ -983,57 +985,20 @@ module ParquetMsg {
...
+          if fixedLen < 2 {
             byteSizes = calcStrSizesAndOffset(entrySeg.a, filenames, sizes, dsetname);
             entrySeg.a = (+ scan entrySeg.a) - entrySeg.a;
-            entryVal = createSymEntry((+ reduce byteSizes), uint(8));
-            readStrFilesByName(entryVal.a, whereNull, filenames, byteSizes, dsetname, ty);
           } else {
-            entryVal = createSymEntry(totalBytes, uint(8));
-            entrySeg.a = (+ scan entrySeg.a) - entrySeg.a;
-            copyValuesFromC(entryVal, distFiles, externalData, valsRead, numRowGroups, rgSubdomains, maxRowGroups, sizes, entrySeg.a, startIdxs);
+            entrySeg.a = fixedLen;
```
stress-tess, line 995: don't we still need to do something like `entrySeg.a = (+ scan entrySeg.a) - entrySeg.a;`? Cause to me this reads as the lengths not the segs (so every val in `entrySeg.a` would be like 5 or whatever)

stress-tess (commented):
maybe I'm misreading or it's just the name not aligning with what I'm thinking, but I don't understand how the segments ends up being what we want in the fixed length case

src/ParquetMsg.chpl
bmcdonald3: Whoops, yup, you got me. I was experimenting with some performance stuff and guess that slipped by. For context, this is what I was seeing for performance:

regular
| nl | Single file per locale | 10 files per locale |
| -- | ----------------------:| -------------------:|
| 1  |                0.13s   |            1.91s    |
| 16 |                1.00s   |            16.0s    |

fixed length
| nl | Single file per locale | 10 files per locale |
| -- | ----------------------:| -------------------:|
| 1  |                0.08s   |            1.35s    |
| 16 |                0.95s   |            23.9s    |

So, even though we are doing strictly less work with the fixed length approach, the performacne is still worse, I suspect due to some file caching problems, but I am still working to pin down the root of the issue there.

What that means is that the fixed length arg doesn't result in a perf improvement all the time, but having it in place still allows you to get it in the cases where it does (but really, will help make my performance experiments easier since I can make more isolated changes 😄)

stress-tess (approved):
looks good! thanks ben

PR #3338, opened 2024-06-17
Closes #3337: Fix can_cast with NumPy 2.0 breaking changes

### LINKED ISSUE #3337: Fix can_cast with NumPy 2.0 breaking changes
In NumPy 2.0, the behavior of `can_cast` changed to no
    longer on values, only types, so by querying the dtype
    of the pdarray, we can work with NumPy 2.0.

bmcdonald3:
In NumPy 2.0, the behavior of `can_cast` changed to no longer on values, only types, so by querying the dtype of the pdarray, we can work with NumPy 2.0.


bmcdonald3:
Closing as this was fixed by pinning NumPy with version <2.0 instead.

PR #3345, opened 2024-06-18
Simplify Command Map

jeremiah-corrado:
Prior to these changes, Arkouda had three command maps: `commandMap` handled the vast majority of commands, `commandMapBinary` handled commands that return a `bytes` instead of a `msgTuple`, and `commandMapArray` handled one command that took a binary payload as an extra argument. 

This PR unifies the command maps by folding the `bytes` return value into `msgTuple` and the binary payload argument into `MessageArgs`. As such, there is only one command map and all message handlers have the same signature. No changes to message serialization/deserialization on the frontend were required.

Also makes some ergonomic improvements for initializing a `MsgTuple` and simplifies error handling in ServerDaemon.


bmcdonald3 (approved):
Looks good to me, great savings of ~350 lines, nice work. Especially like the new way that binary payloads are handled.

stress-tess (approved):
This looks great!! nice work jeremiah

Issue #3346, opened 2024-06-18
Intermittent CI failures with python `3.x/3.12` due to `ForwardRef`


stress-tess:
Even after pinning the numpy version to `<2.0` I'm still seeing CI failures:https://github.com/Bears-R-Us/arkouda/actions/runs/9572308279/attempts/1

It seems only the `3.x` python version is affected. The python version was `3.12.4` with an error along the lines of
```python
TypeError: type of argument "pda1" must be one of (pdarray, Strings, ForwardRef('Categorical'), Sequence); got arkouda.categorical.Categorical instead
```
So this appears to be a forward ref issue

The interesting bit is that python 3.12.0 didn't have this issue and python `3.12.4` was released on June 6th but we've had successful runs since then...

I think the most straightforward plan is to pin the python version as well until we can dig into it more

stress-tess:
it seems rerunning the test worked? I'm going to leave this issue open since this doesn't seem like a one off kinda problem

stress-tess:
I'm working on eliminating the `ForwardRef` since that seems to be the main source of the issue. This reddit comment seems to give a much cleaner solution for avoiding circular imports

stress-tess:
Testing locally this seems to only be breaking for python `3.12.4`. I pinned python to be less than this and disabled the `3.12/3.x` CI checks for this release until we can get to the bottom of this

stress-tess:
python 3.12.5 was released 2 days ago https://www.python.org/downloads/release/python-3125/

It seems to make some changes to how they handle `ForwardRef` in particular the `__eq__` function: 
https://docs.python.org/release/3.12.5/whatsnew/changelog.html#python-3-12-5

I'm hoping this will fix the problems we were having. It's not yet available through conda, but I'm impatient so I'm going to build from source

stress-tess:
okay that didn't fix our problem, but I did find this:
https://typeguard.readthedocs.io/en/stable/userguide.html#permanently-suppressing-type-checks-for-selected-functions

indicating we needed to switch from `@no_type_check` to `@typeguard_ignore`, but that caused errors. I realized the version of typeguard is way behind. Updating that gives the following error:
```python
TypeError: ForwardRef._evaluate() missing 1 required keyword-only argument: 'recursive_guard'
```

This one is seen by others when moving from 3.12.3 to 3.12.4, so I feel like we're on the right track. It seems most other ppl are using pydantic so this was an upstream error for them

PR #3348, opened 2024-06-19
Resolves #3347: Eliminate csv file reads fully into memory

### LINKED ISSUE #3347: CSV Reading uses too much memory due to full file reads
In the CSVMsg module, `read_csv` uses too much memory which causes Segmentation fault when the program runs out of memory.

This is arising due to the following pattern which appeared in many places in the module:
```chapel
var lines = reader.lines().strip();
```

This reads the entire file into memory, and when the file sizes are huge relative to the memory available, the server crashes.
https://github.com/Bears-R-Us/arkouda/pull/3238 improved the situation here by eliminating this pattern from `read_files_into_dist_array`, but `lsCSVMsg` and `get_info` still use this pattern and cause crashes.

ShreyasKhandekar:
Resolves #3347

Replace `reader.lines()` with more granular reads.

We were seeing nighly failures in the `csvIO` benchmark with Segmentation Faults and timeouts.
This was hard to diagnose due to the nature of the Segfault not allowing some good old printf debugging.

After going through the git history to nail down the offending commit, I realized that this issue was not introduced in the new code; rather, it had been there all along.
In the CSVMsg module, `read_csv` uses too much memory which causes Segmentation fault.

This was arising due to the following pattern which appeared in many places in the module:

```chapel
var lines = reader.lines().strip();
```

This reads the entire file into memory, and when the file sizes are huge relative to the memory available, the server crashes.
https://github.com/Bears-R-Us/arkouda/pull/3238 improved the situation here by eliminating this pattern from `read_files_into_dist_array`, but `lsCSVMsg` and `get_info` still used this pattern and cause crashes.

I've replaced the offending pattern in the two methods which still used it and now the performance is better than ever and the Segfaults are gone. Hopefully this will also resolve the nightly failures without having to bump down the problem size again.

Multi locale, multi file per locale (time in seconds)

| 5 locales n file per locale N=5M total | old      | new (#3238)     | new (this PR)
|----------------------------------------|----------|----------|----------|
| 1                                      | 145.3038 | 19.6761  | 3.9757 |
| 5                                      | 668.5100 | 63.5184  | 8.6695 |
| 10                                     | -        | 127.4142 | 18.6796 |
| 15                                     | -        | 188.9333 | 27.1377 |

In general, this is **another** huge leap forward for multilocale settings (I didn't even mean to make a huge perf improvement, just wanted to make the nightly tests pass :P )

### DIFF
--- a/src/CSVMsg.chpl
+++ b/src/CSVMsg.chpl
@@ -61,18 +61,34 @@ module CSVMsg {
             var errorMsg = "File %s does not exist in a location accessible to Arkouda".doFormat(filename);
             csvLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
             return new MsgTuple(errorMsg,MsgType.ERROR);
-        } 
+        }
 
         // open file and determine if header exists.
         var idx = 0;
         var reader = openReader(filename);
-        var lines = reader.lines().strip();
-        if lines[0] == CSV_HEADER_OPEN {
-            idx = 3; // set to first line after header
+        var line = reader.readLine(stripNewline=true);
+        var hasHeader = false;
+        if line == CSV_HEADER_OPEN {
+            hasHeader=true;
+            // The first three lines are headers we don't care about
+            // Advance through them without reading them in
+            for param i in 1..<3 {
+                try {reader.advanceThrough(b'\n');}
+                catch {
+                    throw getErrorWithContext(
+                        msg="This CSV file is missing header values when headers were detected.",
+                        lineNumber=getLineNumber(),
+                        routineName=getRoutineName(),
+                        moduleName=getModuleName(),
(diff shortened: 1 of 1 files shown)


ShreyasKhandekar:
The CI failures are due to the numpy version issues, not this PR. See https://github.com/Bears-R-Us/arkouda/issues/3346

stress-tess:
@ShreyasKhandekar I merged the PR pinning numpy to `<2.0` which should resolve most of the CI failures. I was still seeing intermittent failures from `3.x/3.12` python tests with a `ForwardRef` issue. I'm working on eliminating the need for that now

stress-tess (approved):
a bug fix and a perf increase!! 🚀  nice job @ShreyasKhandekar!

PR #3349, opened 2024-06-19
Related to #3346: Disable `3.12` / `3.x` workflows

### LINKED ISSUE #3346: Intermittent CI failures with python `3.x/3.12` due to `ForwardRef`

stress-tess:
Related to #3346, we were seeing some intermittent CI failures with python `3.x/3.12` due to `ForwardRef`. I tried a couple different things to resolve this but wasn't able to get it working. For the time being disable the typechecking on these functions

The biggest thing I tried was to try and use `TYPE_CHECKING` as shown here [this stackoverflow post](https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports). While this did get past mypy it would raise a type error when functions with this type annotation were called.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -63,7 +63,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.x']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
     container:
       image: chapel/chapel:1.33.0
     steps:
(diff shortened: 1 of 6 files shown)


stress-tess:
I pulled this down locally and it's working for all python versions `<3.12.4`. So for the time being, I'm disabling the `3.12` and `3.x` checks

PR #3350, opened 2024-06-20
Closes #3339: Add multi-batch parquet read tests

### LINKED ISSUE #3339: Add parquet read tests which require more than one batch
We've encountered a number of examples where we run into issues where some new parquet logic to avoid single batches seems to work until we exceed ~700,000 elements (which seems to be around where a second batch is needed). This mostly crops up when we have some combination of nans, empty segs, and empty strings. To verify an future attempts to improve our batch writes in parquet don't run into the same problem, we should add tests for these "large" reads

stress-tess:
This PR (closes #3339) adds testing for parquet reads or arrays and strings large enough to trigger more than one batch. We also add testing of a segarray of segstrings containing empty segs and empty strings

### DIFF
--- a/PROTO_tests/tests/io_test.py
+++ b/PROTO_tests/tests/io_test.py
@@ -527,11 +527,13 @@ class TestParquet:
         df["a"] = df["a"].export_uint()
         assert ak.arange(10).to_list() == df["a"].to_list()
 
-    def test_empty_segs_segarray(self):
+    def test_multi_batch_reads(self):
         # verify reproducer for #3074 is resolved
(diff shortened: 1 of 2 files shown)


drculhane (commented):
This just writes some data, reads it back in, and verifies that read=written, correct?

stress-tess:
> This just writes some data, reads it back in, and verifies that read=written, correct?

@drculhane yep! it just we've seen issues where bugs only appear above a certain number of elements (see #3074) because they aren't testing when a read requires more than one batch based on our batch size

PR #3352, opened 2024-06-21
Closes #3336, #3362: Reuse random number generation loop structure

### LINKED ISSUE #3336: Reuse and improve structure for random generation methods
- creating a record similar to what jeremiah suggested [here](https://github.com/Bears-R-Us/arkouda/pull/3253#discussion_r1621285128) to avoid having the duplicated code for single value and array versions of distribution parameters 
- pulling out the structure into a proc (like  which takes the actual distribution method as an inline proc. This will be along the lines of what is done in [`SegmentedComputation.chpl`](https://github.com/Bears-R-Us/arkouda/blob/master/src/SegmentedComputation.chpl)
  - A dry run of something similar (with an outdated stream per task structure) can be seen with exponential [here](https://github.com/Bears-R-Us/arkouda/pull/3310)

I just wanted to get this fixed as soon as possible and not complicate the review

_Originally posted by @stress-tess in https://github.com/Bears-R-Us/arkouda/issues/3331#issuecomment-2168411183_

### LINKED ISSUE #3362: small bug in `thisLocsNumChunks` calculation in `uniformStreamPerElem`
While presenting on adapting the looping structure of the random number generation functions to calculate a fixed number of elements per stream. I saw this line
https://github.com/Bears-R-Us/arkouda/blob/522327336c23c9718e11a355013673db048bd83d/src/RandMsg.chpl#L596

but I think it should actually be:
```chapel
thisLocsNumChunks = ceil((locSubDom.high + 1 - (locSubDom.low + offset)) / elemsPerStream:real):int;
```

because we want the number of elements we're handling which is our end `locSubDom.high + 1 ` minus our  start `(locSubDom.low + offset)`  which we then divide by `elemsPerStream` to get the number of chunks. So the current code is adding `offset` when I think it should be subtracting it

stress-tess:
This PR (closes #3336 and fixes #3362)

Update to reuse the random number generation loop structure for different algorithms. Also added the record which removes the need for duplicating the loops for a scalar lambda and array lambda

### DIFF
--- a/src/RandUtil.chpl
+++ b/src/RandUtil.chpl
@@ -1,0 +1,90 @@
+module RandUtil {
+    use MultiTypeSymbolTable;
+    use MultiTypeSymEntry;
+    use RandMsg;
+    use ArkoudaRandomCompat;
+
+    const minPerStream = 256; // minimum number of elements per random stream
+
+    record scalarOrArray {
+        var isArray: bool;
+        var sym;  // TODO figure out type hint here to avoid generic
+        var val: real;
+
+        proc init(scalarOrArrayString: string, isArray: bool, st: borrowed SymTab) {
+            // I'm not sure if there's a good way to remove these try!
+            this.isArray = isArray;
+            if isArray {
+                try! st.checkTable(scalarOrArrayString);
+                this.sym = try! toSymEntry(getGenericTypedArrayEntry(scalarOrArrayString, st),real).a;
+            }
+            else {
+                // prob not the smartest way of doing this
+                // we just want to avoid unnecessarily creating a large array
+                this.sym = try! makeDistArray([0.0]);
+                val = try! scalarOrArrayString:real;
+            }
+        }
+
+        proc this(idx): real {
+            return if isArray then this.sym[idx] else this.val;
+        }
+    }
+
+    enum GenerationFunction {
+      PoissonGenerator,
+    }
+
+    proc uniformStreamPerElem(ref randArr: [?D] ?t, param function: GenerationFunction, hasSeed: bool, const lam: scalarOrArray(?), ref rng) throws {
+        if hasSeed {
+            // use a fixed number of elements per stream instead of relying on number of locales or numTasksPerLoc because these
+            // can vary from run to run / machine to mahchine. And it's important for the same seed to give the same results
+            const generatorSeed = (rng.next() * 2**62):int,
+                elemsPerStream = max(minPerStream, 2**(2 * ceil(log10(D.size)):int));
+
+            // using nested coforalls over locales and tasks so we know how to generate taskSeed
+            coforall loc in Locales do on loc {
+                const locSubDom = randArr.localSubdomain(),
+                    offset = if loc.id == 0 then 0 else elemsPerStream - (locSubDom.low % elemsPerStream);
+
+                // skip if all the values were pulled to previous locale
+                if offset <= locSubDom.high {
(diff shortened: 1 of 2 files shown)


src/RandUtil.chpl
```diff
@@ -0,0 +1,90 @@
...
+        proc this(idx): real {
+            return if isArray then this.sym[idx] else this.val;
+        }
+    }
+
+    enum GenerationFunction {
+      PoissonGenerator,
+    }
+
+    proc uniformStreamPerElem(ref randArr: [?D] ?t, param function: GenerationFunction, hasSeed: bool, const lam: scalarOrArray(?), ref rng) throws {
```
jeremiah-corrado, line 38: If it would make things simpler when you add the other generators, I think you could adjust this s.t. the generator function can be passed in as an argument directly so you don't have to use the enum+select below: 
```chapel
proc uniformStreamPerElem(ref randArr: [?D] ?t, function: proc(_: real, ref _: randomStream(real)), ...
```

(although I'm not sure if all the generator functions would have the same signature?)

src/RandUtil.chpl
```diff
@@ -0,0 +1,90 @@
...
+
+    const minPerStream = 256; // minimum number of elements per random stream
+
+    record scalarOrArray {
+        var isArray: bool;
+        var sym;  // TODO figure out type hint here to avoid generic
+        var val: real;
+
+        proc init(scalarOrArrayString: string, isArray: bool, st: borrowed SymTab) {
+            // I'm not sure if there's a good way to remove these try!
```
jeremiah-corrado, line 15: Throwing initializers are currently somewhat incomplete as a feature, but you can call a throwing helper procedure from within an initializer after `init this`. Ex:

```chapel
proc init() throws {
  this.x = 1;
  init this;
  this.y = throwingHelper();
}
```

Which isn't the cleanest pattern, but could allow you to remove the `try!`'s here.

jeremiah-corrado (approved):
Looks good! I have a couple of minor suggestions:

src/RandUtil.chpl
stress-tess: ah yah I didn't consider this! but yah I don't think they would have the same signature

src/RandUtil.chpl
stress-tess: that's what I thought about throwing initializers. And I like this better than using `try!` so I'll update that!

stress-tess: Hmmm I tried this and it wasn't very happy

```chapel
/proj/arkouda//src/RandUtil.chpl:14: In initializer:
/proj/arkouda//src/RandUtil.chpl:19: error: call to throwing function throwingSymEntryHelper without throws, try, or try! (relaxed mode)
/proj/arkouda//src/RandUtil.chpl:33: note: throwing function throwingSymEntryHelper defined here
```

stress-tess: ooops I missed that ur `init` had a `throws` after it! retyring now

stress-tess: I couldn't figure out the type hint that `sym` should have and it was angry that `sym` hadn't be initialized since it has no type
```
proj/arkouda//src/RandUtil.chpl:14: In initializer:
proj/arkouda//src/RandUtil.chpl:17: error: can't omit initialization of field "sym", no type or default value provided
```

I tried defualting it to `makeDistArray([0.0])` but that caused this error with runtime checks enabled:
`error: halt reached - assigning between arrays of different shapes in dimension 0: 1 vs. 5`

so it seems like the only one I could easily add a throwing helper to is the cast to real. I think I'm just gonna leave this as is for now

jaketrookman (approved):
Looks good

PR #3361, opened 2024-06-25
Closes #3360: Reduce code duplication in Parquet read code with templates

### LINKED ISSUE #3360: Reduce code duplication in Parquet read code with templates
The Parquet code previously had been duplicating blocks of code with only changing the types. To reduce this duplication to improve readability and maintainability, template functions are being added in place of the duplicate code.

bmcdonald3:
The Parquet code previously had been duplicating blocks of code with only changing the types. To reduce this duplication to improve readability and maintainability, template functions are being added in place of the duplicate code.

This reduces the readColumnByName function from 198 lines to 100 lines, so about 50% reduction in function size.

Additionally, since the code is no longer duplicated, changes won't need to be put in multiple places.

### DIFF
--- a/src/parquet/ReadParquet.cpp
+++ b/src/parquet/ReadParquet.cpp
@@ -1,12 +1,12 @@
 #include "ReadParquet.h"
 #include "UtilParquet.h"
 
-template <typename ReaderType, typename RetType>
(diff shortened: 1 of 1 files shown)


stress-tess (approved):
This looks great!! I like this a lot

PR #3363, opened 2024-06-26
Ignore new Parquet object files

bmcdonald3:
After splitting the Parquet code into multiple, distinct files, we need to add those to the gitignore in the case that someone adds the src directory non-descriminantly.


stress-tess (approved):
thanks ben!

PR #3365, opened 2024-06-26
Remove string.doFormat, replacing with string.format

jeremiah-corrado:
Remove the `doFormat` method from the compatibility modules as it is no longer needed.


bmcdonald3 (approved):

bmcdonald3:
Looks like there are some merge conflicts here now.

jeremiah-corrado:
Those should be fixed now.

bmcdonald3 (approved):

PR #3370, opened 2024-06-28
Closes #3293: Add `putmask`

### LINKED ISSUE #3293: putmask
Implement a `putmask` function to mirror numpy:
https://numpy.org/doc/stable/reference/generated/numpy.putmask.html

drculhane:
This is the putmask discussed in a recent group meeting, with the improvements we noted for making the pdarray of masked-in values the same length as the source data.

closes #3293


arkouda/numeric.py
```diff
@@ -2182,3 +2183,59 @@ def count_nonzero(pda):
...
+
+    # if values is not the same size as pda, repeat it and/or prune it as needed
+
+    growth = pda.size // values.size + (0 if pda.size % values.size == 0 else 1)
+    result = concatenate(growth * [values])
+    if result.size > pda.size:
+        reduction = result.size % pda.size
+        result = result[:-(reduction)]
+
+    return where(mask, result, pda)
```
ajpotts, line 2241: Instead of returning a new array, can we make the change in place to match numpy?

ajpotts (changes requested):
Looks great!  I think maybe we should have it work in place though, to match numpy more exactly.

arkouda/numeric.py
drculhane: Made that change, and pushed it.  Also fixed some stuff flake8 didn't like.  The PR is still failing some checks that involve the tests directory (not PROTO-tests).

drculhane: And a "make test" causes the same error.  Is there a way I can tell git not to run "tests" tests?

ajpotts (approved):
Looks good.  Thanks!

jaketrookman (approved):
Looks good

arkouda/numeric.py
```diff
@@ -2182,3 +2183,58 @@ def count_nonzero(pda):
...
+    from arkouda.pdarraysetops import concatenate
+
+    # check for matching types
+
+    if values.dtype != pda.dtype:
+        raise TypeError("ak.putmask requires arrays of matching type")
+
+    # if values is not the same size as pda, repeat it and/or prune it as needed
+
+    growth = pda.size // values.size + (0 if pda.size % values.size == 0 else 1)
```
stress-tess, line 2234: bools can be interpreted as ints, so 
```python
+ (0 if pda.size % values.size == 0 else 1)
# is equivalent to
+ (pda.size % values.size != 0)
```

arkouda/numeric.py
```diff
@@ -2182,3 +2183,58 @@ def count_nonzero(pda):
...
+    result = concatenate(growth * [values])
+    if result.size > pda.size:
+        reduction = result.size % pda.size
+        result = result[:-(reduction)]
```
stress-tess, line 2238: I feel like this could be more space and time efficient if we did this on the chpl side. Because if `values.size == 2`, we could just loop over these 2 over and over and write them into place instead of allocating `pda.size` elements and concatenating them together.

We could also do some sort of thresholding where if `values` is quite small we make a copy of it on each locale. I'm thinking something similar to how we have 3 different methods for histogram depending on the bucket size. I think the same thing applies here but instead of writing into, we are reading from (so we wouldn't need atomics, we can just use `const`)

stress-tess (changes requested):
This all looks correct to me, but I think we can get a perf boost by writing a chpl kernel to handle this

arkouda/numeric.py
drculhane: I'll look into pushing this chapel-side.

arkouda/numeric.py
drculhane: I'm inclined to resist that change.  While it's true that these are equivalent, I don't think the meaning of "+ (pda.size % values.size !=0) is as clear.

arkouda/numeric.py
drculhane: Had a very good and short conversation with Amanda about this, the upshot of which is that I'm going to look at arkouda-izing np.tiles and np.repeat, and then return to this.

stress-tess: we talked about this a bit. I think let's merge this to get some unit tests in it and then we'll do 2 phases of adding more optimized / complex chapel implementations

arkouda/numeric.py
stress-tess: I would argue that adding the boolean rather than the explicit `1/0` is more pythonic and more closely resembles the style of the project as a whole. But ultimately this is just a matter of taste, so im totally down with leaving it as is!

arkouda/numeric.py
stress-tess: see https://github.com/Bears-R-Us/arkouda/issues/3428 for more info on the future work

PROTO_tests/tests/numeric_test.py
```diff
@@ -744,3 +744,115 @@ def test_clip(self, prob_size):
...
+            # finally try to raise the error
+
+            pda = ak.random.randint(0, 10, 10).astype(ak.float64)
+            values = np.arange(10)
+            with pytest.raises(TypeError):
+                ak.putmask(pda, pda > 3, values)
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_clip(self, prob_size):
+        seed = pytest.seed if pytest.seed is not None else 8675309
```
stress-tess, line 801: just double checking we are intentionally adding `test_clip` here since it's not immediately obvious how that ties into `putmask`

drculhane: Okay, that puzzles me, too.  I'll have to check why that wasn't already there.

drculhane: This looks like I made a cut&paste error.  test_clip is in there twice.  I'm surprised testing didn't flag that, but I guess it doesn't.  I'll compare the two versions, and if they're identical, I'll remove one.  If they're not, I'll have to figure out why.

ServerModules.cfg
stress-tess, line 1: seems we accidentally included so changes from playing around with multidim stuff

serverConfig.json
stress-tess, line 1: here too

src/ManipulationMsg.chpl
stress-tess, line 1: this one is just a comment so not terribly important that if it's included or not

ServerModules.cfg
```diff
@@ -19,7 +19,7 @@ IndexingMsg
 JoinEqWithDTMsg
 KExtremeMsg
 LogMsg
-# ManipulationMsg
+ManipulationMsg
```
stress-tess, line 22: this one seems to have still slipped thru the cracks

stress-tess (approved):
looks good!

jaketrookman (approved):
Great updates

PR #3371, opened 2024-06-28
Automated command registration

jeremiah-corrado:
Adds support for two new annotations for registering server commands
* `arkouda.instantiateAndRegister`
* `arkouda.registerCommand`

The `arkouda.registerND` annotation is still supported but is intended to be replaced by the above annotations (it is a strictly less powerful version of  `instantiateAndRegister`).

- [x] multi-locale testing
- [x] array-api testing

## Annotations

| annotation | purpose | restrictions |
| :---  | :--- | :--- |
| `instantiateAndRegister` | create commands from a generic message handler proc. Generally useful for more complex message handlers. | must have the signature: `proc <name>(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, <generic args>): MsgTuple throws` |
| `registerCommand` | create command(s) from a normal chapel proc with argument types. Intended for most use cases. | arguments can be: arrays (corresponding to `SymEntry`'s), scalars (int, bool, string, etc.), or `list`'s of scalars |

<details>

<summary>instantiateAndRegister details</summary>

Generic formals (`type` and `param` formals coming after the `st: borrowed SymTab,` formal)  should have names in the following form: `<parameter_class>_<parameter_name>[_][...]`

The parameter class and name correspond to lists in the configuration file that tell the registration script which values to use for that generic formal. Today, the only parameter class defined in the configuration file is `array`, with the two parameters: `nd`, and `dtype`.

In the future, we may want to add other parameter classes needed for other generic symbol entry types or other groups of generic commands that aren't array-focused.

As an example, `reshapeMsg` uses `instantiateAndRegister` with a generic `type` formal corresponding to the array argument's element type, an input array rank formal corresponding to the array arguments rank, and an output rank formal corresponding to the returned array's rank:
```chapel
@arkouda.instantiateAndRegister(prefix='reshape')
proc reshapeMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab,
  type array_dtype,
  param array_nd_in: int,
  param array_nd_out: int
): MsgTuple throws { 
  ...
  const name = msgArgs["name"],
          rawShape = msgArgs["shape"].toScalarTuple(int, array_nd_out);

  var eIn = st[name]: borrowed SymEntry(array_dtype, array_nd_in);
  ...
}
```

This gets instantiated for all permutations of the `dtype` and `nd` values in the config file, creating commands like the following:

* "reshape<int, 1, 1>"
* "reshape<int, 1, 2>" 
* "reshape<int, 2, 1>"
* ...
* "reshape<float64, 3, 2>"
* ...   

</details>

<details>

<summary>registerCommand details </summary>

This annotation will generate commands from a chapel procedure (that adheres to a few constraints) avoiding the need to manually write argument parsing and client response code.

Example:

```chapel
@arkouda.registerCommand()
proc meanReduce(const ref x: [?d] ?t, skipNan: bool, axes: list(int)): [] real throws {
  var meanArr = makeDistArray(domOffAxis(d, axes), real);
  forall (slice, sliceIdx) in axisSlices(d, axes) {
    if canBeNan(t) && skipNan
      then meanArr[sliceIdx] = meanSkipNan(x, slice);
      else meanArr[sliceIdx] = meanOver(x, slice);
  }
  return meanArr;
}
```

Note: one should still pass a pdarray to the first argument on the client side (not a Python list or other array-like data structure). The argument parsing code assumes that you want that array argument to be extracted from the symbol table as a `SymEntry`. This means that the above procedure should be called as follows:
```python
create_pdarray(
    generic_msg(
        cmd=f"meanReduce<{x.dtype},{x.ndim}>",
        args={
            "x": x,
            "axes": [1, 2],
            "skipNan": False, 
        },
    )
)
```

Procedures annotated with `registerCommand` that have array arguments will be instantiated for all permutations of the array argument types and ranks. This means that the above proc would register a list of commands like the following because it has one array argument:

* "meanReduce<int, 1>"
* "meanReduce<int, 2>" 
* ...
* "meanReduce<float64, 3>"
* ... 

If there are multiple array arguments, then each will be permuted separately. We may want to change this behavior going forward to avoid creating too many commands for simple procedures (see `corr` in `statsMsg` for an example of how this is currently handled).

</details>

## Build System Adjustments

**TLDR:** ❗️❗️❗️ If you make any changes to `server_config.json` or `ServerModules.cfg` you need to run `make register-commands` **before** running `make` to build the arkouda server. 
(or, if you want to use a different config file e.g., `ARKOUDA_CONFIG_FILE=ArrayAPIServerModules.cfg make`, then run `ARKOUDA_CONFIG_FILE=ArrayAPIServerModules.cfg make register-commands` first).

Adds a new command  to the makefile: `make register-commands` which invokes a python script `src/registry/register_commands.py`. This script looks for the above annotations in all the server's message handler modules and generates Chapel code to register the annotated functions as commands/message-handlers. The generated code is put in `src/registry/Commands.chpl`.

The script is currently not invoked at build time because it relies on a 2.0+ version of the Chapel frontend compiler bindings. This can be installed, by navigating to `$CHPL_HOME/tools/cd tools/chapel-py` and `pip install -e .`-ing.

The intention is to eventually run register-commands at build time. For now, the `Commands.chpl` file contains all the code needed to build a default arkouda server (i.e., 1-D arrays with support for the standard scalar types). For multi-dimensional support, or when adding new commands, it needs to be run (where the new `Commands.chpl` should be included in PRs that add new commands).

## Config File Adjustments

The `server-config.json` file is replaced by `registration-config.json` (name change to avoid confusion with `ARKOUDA_CONFIG_FILE` variable). Going forward, this portion of the file will be used to adjust supported array dimensionality and dtypes:
```json
"array": {
    "nd": [1],
    "dtype": [
        "int",
        "uint",
        "real",
        "bool"
    ]
}
```

the "nd" option can accept a list of numbers, or an arbitrary piece of python code that evaluates to a list of numbers (as a string).

## Code Changes

The following files were updated to use the new annotations, and can be used as examples of how to go translating other modules to use the updated framework. The relevant frontend code was also updated to take advantage of the new command names.

* manipulationMsg.chpl
* statsMsg.chpl


bmcdonald3 (commented):
Super cool, love the new annotations. I haven't looked deeply at the code yet, but reading the description, I'm seeing:
> For now, the Commands.chpl file contains all the code needed to build a default arkouda server (i.e., 1-D arrays with support for the standard scalar types). For multi-dimensional support, or when adding new commands, it needs to be run (where the new Commands.chpl should be included in PRs that add new commands).

Based on that description and looking at `src/registry/Commands.chpl`, it seems that we are giving up the current `ServerModules.cfg` file, if I'm understanding correctly. Given this is a 2.0+ requirement, could we just special case it in the `Makefile` (like we do for comapt files) to run the old code, rather than require people to update the Commands file? I feel like that would be difficult to disseminate knowledge about, especially with how fast this is changing.

Alternatively, we only "require" support of the previous release (and aren't keeping any around for perf reasons anymore), so we could just remove those old Chapel compiler versions.

In summary, I don't think I'd want to make a change that is going to temporarily make some things a little bit harder if it's only based on the Chapel compiler version. 

What do you think about that? Would you rather special case in the Makefile or start conversations about dropping the older `chpl` versions or somethign else? I'd probably lean towards dropping older `chpl` versions.

bmcdonald3:
Is this ready for another review pass Jeremiah?

jeremiah-corrado:
Yep, it should be good to go, thanks!

Also to document the result of the offline discussion about your previous comment: we tentatively decided to drop support for pre-2.0 versions of Chapel in a separate PR after this gets merged. This way, `src/registry/register_commands.py` could be run automatically whenever the server is built.

bmcdonald3:
Oh good point, sorry to drag you into this, but @stress-tess have you been able to confirm if we can drop support for pre 2.0 releases of Chapel?

arkouda/array_api/_typing.py
```diff
@@ -44,11 +44,9 @@
 
 
 class NestedSequence(Protocol[_T_co]):
-    def __getitem__(self, key: int, /) -> _T_co | NestedSequence[_T_co]:
-        ...
+    def __getitem__(self, key: int, /) -> _T_co | NestedSequence[_T_co]: ...
 
-    def __len__(self, /) -> int:
-        ...
+    def __len__(self, /) -> int: ...
```
bmcdonald3, line 49: Why were these `...`s added?

src/CommandMap.chpl
```diff
@@ -28,13 +28,15 @@ module CommandMap {
...
    * function signature & MsgTuple return type
    */
-  proc registerFunction(cmd: string, fcf: f.type) {
-    commandMap.add(cmd, fcf);
-  }
+  proc registerFunction(cmd: string, fcf: f.type, modName: string = "", line: int = -1) {
+    if commandMap.contains(cmd) {
+      writef("Warning: Command '%s' already registered. Ignoring registration from [%s:%d]", cmd, modName, line);
+    } else {
+      commandMap.add(cmd, fcf);
```
bmcdonald3, line 36: Ah, this actually solves the same problem that I was fixing with https://github.com/Bears-R-Us/arkouda/pull/3424, looks good 👍

bmcdonald3 (approved):
Looks good mostly, I think there were a lot of whitespace, formatting, etc. changes here that made it sort of difficult to sort out the meaningful code from the formatting differences, but I'd say this is ready to go. I've asked a question about some dots in the client side, but not important enough to hold up merging.

I'm going to wait til Monday to merge just because testing has been noisy with the array view ellision stuff lately.

jeremiah-corrado:
>  I think there were a lot of whitespace, formatting, etc. changes here that made it sort of difficult to sort out the meaningful code from the formatting differences

Sorry about that. I ran `black` on the array api modules as a part of this but probably should have put that in a separate PR.

arkouda/array_api/_typing.py
jeremiah-corrado: I think the ellipses just got moved up from the next line by the auto-formatter I used.

src/ManipulationMsg.chpl
```diff
@@ -58,87 +58,84 @@ module ManipulationMsg {
...
+    return st.insert(eOut);
+  }
+
+  proc broadcastToMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab,
+    type array_dtype,
+    param array_nd_in: int,
+    param array_nd_out: int
+  ): MsgTuple throws
+    where array_nd_in > array_nd_out
+  {
```
stress-tess, line 134: can version of `broadcastToMsg` also get called even though it doesn't have the `@arkouda.instantiateAndRegister`?

src/ServerDaemon.chpl
```diff
@@ -774,6 +773,7 @@ module ServerDaemon {
             this.socket = this.context.socket(ZMQ.REP); 
             this.port = try! getEnv('METRICS_SERVER_PORT','5556'):int;
 
+            try! this.socket.bind("tcp://*:%?".format(this.port));
             try! this.socket.bind("tcp://*:%?".format(this.port));
```
stress-tess, line 777: I think this is the second time this has snuck in lol

src/registry/Commands.chpl
stress-tess, line 1: was this file committed on purpose?

stress-tess (commented):
Looks great! Just a few quick questions

src/registry/Commands.chpl
jeremiah-corrado: Yes, it supports building the server in the default config (without having to manually run `make register-comands`). After support for older versions is removed, this file can be removed from the tree and added to .gitignore (because it will be generated automatically whenever the server is built).

src/ServerDaemon.chpl
jeremiah-corrado: Thanks for catching this! (merge conflicts have not been my favorite lately)

src/ManipulationMsg.chpl
jeremiah-corrado: yeah, the generated code will just call `ManipulationMsg.broadcastToMsg(...)`, and then it's up to the compiler to decide which overload to call from there.

As a side note, I'd like to add support for where-clause interpretation to `instantiateAndRegister`, s.t. it wouldn't even try to instantiate a function whenever the where-clause would evaluate to false. In this case, that would make this second overload unnecessary.

stress-tess (approved):
looks great! thanks jeremiah!

PR #3393, opened 2024-07-08
Updates PROTOs pdarray_creation_test

drculhane:
This updates PROTO-tests/tests/pdarray_creation_test.py to ensure that it covers all of the same tests that the non-PROTOs version did.

Several tests that are done in one function in tests are done in multiple ones in PROTO-tests (e.g. test_randint vs test_randint_array_dtype, test_randint_num_dtype, test_randint_misc).  This makes for a challenging crosswalk, but I think that everything has been updated.

I made the following changes to ones_like, zeros_like, and full_like:

- added an assert that the new array is the same size as the original (this will change to shape with n-d arrays),
- changed the creation of the original array so that the "_like" array would have different content.
- as in, old version was zeros_arr = ak.zeros(size) ; zeros_like_arr = ak.zeros_like(zeros_arr).
- new version is ran_arr = ak.array(arange(size)) ; zeros_like_arr = ak.zeros_like(ran_arr)
- as it was, the "_like" functions could have just been doing a copy, and we wouldn't have caught it

In addition to the note above about shape, several other functions will have to be updated when we implement n-d arrays.  For example, it currently asserts that n-d arrays have ArrayView type.

### DIFF
--- a/PROTO_tests/tests/pdarray_creation_test.py
+++ b/PROTO_tests/tests/pdarray_creation_test.py
@@ -206,6 +206,10 @@ class TestPdarrayCreation:
 
     @pytest.mark.parametrize("size", pytest.prob_size)
     def test_randint_misc(self, size):
+
+        # Test that int_scalars covers uint8, uint16, uint32
+        ak.randint(low=np.uint8(1), high=np.uint16(100), size=np.uint32(100))
+
         # test resolution of modulus overflow - issue #1174
         assert (ak.randint(-(2**63), 2**63 - 1, size) != ak.full(size, -(2**63))).any()
 
@@ -343,11 +347,12 @@ class TestPdarrayCreation:
     @pytest.mark.parametrize("size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", [ak.int64, ak.float64, ak.bool_, ak.bigint])
     def test_ones_like(self, size, dtype):
-        ones_arr = ak.ones(size, dtype)
-        ones_like_arr = ak.ones_like(ones_arr)
+        ran_arr = ak.array(np.arange(size), dtype=dtype)
+        ones_like_arr = ak.ones_like(ran_arr)
         assert isinstance(ones_like_arr, ak.pdarray)
         assert dtype == ones_like_arr.dtype
         assert (1 == ones_like_arr).all()
+        assert ones_like_arr.size == ran_arr.size
 
     @pytest.mark.parametrize("size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", [int, ak.int64, ak.uint64, float, ak.float64, bool, ak.bool_])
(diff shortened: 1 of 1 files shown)


PROTO_tests/tests/pdarray_creation_test.py
```diff
@@ -343,11 +347,12 @@ def test_ones_misc(self):
     @pytest.mark.parametrize("size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", [ak.int64, ak.float64, ak.bool_, ak.bigint])
     def test_ones_like(self, size, dtype):
-        ones_arr = ak.ones(size, dtype)
-        ones_like_arr = ak.ones_like(ones_arr)
+        ran_arr = ak.array(np.arange(size), dtype=dtype)
```
ajpotts, line 350: `ran_arr = ak.arange(size, dtype=dtype)` would be more performant.

PROTO_tests/tests/pdarray_creation_test.py
```diff
@@ -389,20 +394,22 @@ def test_full_misc(self):
...
         assert dtype == full_like_arr.dtype
         assert (full_like_arr == 1).all()
+        assert full_like_arr.size == ran_arr.size
 
     @pytest.mark.parametrize("size", pytest.prob_size)
     @pytest.mark.parametrize("dtype", [int, ak.int64, ak.uint64, float, ak.float64, bool, ak.bool_])
     def test_zeros_like(self, size, dtype):
-        zeros_arr = ak.zeros(size, dtype)
-        zeros_like_arr = ak.zeros_like(zeros_arr)
+        ran_arr = ak.array(np.arange(size), dtype=dtype)
```
ajpotts, line 407: Same comment as above.

PROTO_tests/tests/pdarray_creation_test.py
```diff
@@ -206,6 +206,10 @@ def test_randint_num_dtype(self, dtype):
 
     @pytest.mark.parametrize("size", pytest.prob_size)
     def test_randint_misc(self, size):
+
+        # Test that int_scalars covers uint8, uint16, uint32
+        ak.randint(low=np.uint8(1), high=np.uint16(100), size=np.uint32(100))
```
ajpotts, line 211: I see several tests like this, where it instantiates an array without checking the output.  Is it just to test there is no error thrown?  I think we should also assert the output is reasonable.

PROTO_tests/tests/pdarray_creation_test.py
```diff
@@ -560,6 +567,9 @@ def test_random_strings_lognormal_errors(self):
         with pytest.raises(TypeError):
             ak.random_strings_lognormal(2, 0.25, 100, 1000000)
 
+        # Test that int_scalars covers uint8, uint16, uint32
+        ak.random_strings_lognormal(np.uint8(2), 0.25, np.uint16(100))
```
ajpotts, line 571: I think if we instantiate an array we should check the output is reasonable.

PROTO_tests/tests/pdarray_creation_test.py
```diff
@@ -658,6 +668,11 @@ def test_fill(self, size, dtype):
         ones.fill(dtype(2))
         assert (dtype(2) == ones).all()
 
+        # Test that int_scalars covers uint8, uint16, uint32
+        ones.fill(np.uint8(2))
+        ones.fill(np.uint16(2))
+        ones.fill(np.uint32(2))
```
ajpotts, line 674: I think if we instantiate an array we should check the output is reasonable.

ajpotts (approved):
It looks good.   I just have the comments about not using numpy unless necessary (for performance reasons on the larger tests on the server) and checking the output of all the arrays that are generated.

PROTO_tests/tests/pdarray_creation_test.py
drculhane: Honestly, I don't know.  It was in nearly all of the non-PROTOs tests, and was missing from some of the PROTOs ones.  In the interests of ensuring that PROTOs covers the same conditions that non-PROTOs does, I included them where they were missing.

PROTO_tests/tests/pdarray_creation_test.py
drculhane: I agree, but as above, I'm copying what was in the soon-to-be-deprecated version.  I suspect this is just there to see if an error is thrown.

PROTO_tests/tests/pdarray_creation_test.py
drculhane: Same response as above.

PROTO_tests/tests/pdarray_creation_test.py
drculhane: Same reply as above :)

PROTO_tests/tests/pdarray_creation_test.py
ajpotts: OK, I see.  Sorry, I should have read the ticket description more closely.

jaketrookman (approved):

PR #3394, opened 2024-07-08
Fixes #3392: Fix mypy CI failures 

### LINKED ISSUE #3392: mypy CI failures
We're seeing some CI failures due to mypy that popped up in https://github.com/Bears-R-Us/arkouda/pull/3386 but also appear when [rerunning the CI on master](https://github.com/Bears-R-Us/arkouda/actions/runs/9753222168/job/27175506781) which was previously passing:

```
arkouda/plotting.py:98: error: Item "Axes" of "Union[Axes, ndarray[Any, Any]]" has no attribute "flatten"
arkouda/plotting.py:100: error: Incompatible types in assignment (expression has type "List[Union[Axes, ndarray[Any, Any]]]", variable has type "Union[Axes, ndarray[Any, Any]]")
arkouda/plotting.py:104: error: Value of type "Union[Axes, ndarray[Any, Any]]" is not indexable
Found 3 errors in 1 file (checked 64 source files)
```

stress-tess:
Fixes https://github.com/Bears-R-Us/arkouda/issues/3392 by refactoring the code slighty to handle if `plt.subplots` returns type `plt.Axes` to satisfy `mypy`

### DIFF
--- a/arkouda/plotting.py
+++ b/arkouda/plotting.py
@@ -94,7 +94,9 @@ def hist_all(ak_df: DataFrame, cols: list = []):
     fig, axes = plt.subplots(num_rows, num_cols, figsize=(10, 10))
     fig.tight_layout(pad=2.0)
 
-    if num_rows > 1:
+    if isinstance(axes, plt.Axes):
+        axes = np.array(axes).flatten()
+    elif isinstance(axes, np.ndarray):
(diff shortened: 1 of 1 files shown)


bmcdonald3 (approved):
Looks good

PR #3396, opened 2024-07-08
Closes #3395 align to numpy scalar types

### LINKED ISSUE #3395: align to numpy scalar types
Align the arkouda scalar types to match numpy:

```
uint8 = np.uint8
uint16 = np.uint16
uint32 = np.uint32
uint64 = np.uint64
int8 = np.int8
int16 = np.int16
int32 = np.int32
int64 = np.int64
float32 = np.float32
float64 = np.float64
complex64 = np.complex64
complex128 = np.complex128
bool_ = np.bool_
str_ = np.str_
```

ajpotts:
Aligns the arkouda scalar types to match numpy:

```
uint8 = np.uint8
uint16 = np.uint16
uint32 = np.uint32
uint64 = np.uint64
int8 = np.int8
int16 = np.int16
int32 = np.int32
int64 = np.int64
float32 = np.float32
float64 = np.float64
complex64 = np.complex64
complex128 = np.complex128
bool_ = np.bool_
str_ = np.str_
```
Also adjusts the necessary type checking statements.

Closes #3395 align to numpy scalar types


jaketrookman (approved):

jaketrookman (approved):

arkouda/pdarrayclass.py
```diff
@@ -764,7 +764,7 @@ def __invert__(self):
             return self._binop(~0, "^")
         if self.dtype == akuint64:
```
stress-tess, line 765: it's not exactly clear to me when we should use `akuint64` vs `akdtype("int64")`

arkouda/random/_generator.py
```diff
@@ -56,6 +59,20 @@ def __str__(self):
...
+    def _generator_msg(self, dtype):
+        return generic_msg(
+            cmd="createGenerator",
+            args={
+                "dtype": akdtype(dtype).name,
+                "has_seed": self._has_seed,
+                "seed": self._initial_seed,
+                "state": self._initial_state,
+            },
+        )
```
stress-tess, line 74: I think this creates a new generator every time and adds it to the symbol table. I think right now we declare a generator for each type and fetch the type we need from the symbol table? 

I think this way would work cause we fast forward the state, but it seems to make unnecessary copies

on master:
```python
>>> rng = ak.random.default_rng(11)

# one per type
>>> ak.list_symbol_table()
['id_RjcKYUF_2', 'id_RjcKYUF_3', 'id_RjcKYUF_1', 'id_RjcKYUF_4']

# call integers repeatedly (this shouldn't add any new generators
>>>  pda_names = []
   ...: for i in range(10):
   ...:     pda = rng.integers(20, size=3)
   ...:     pda_names.append(pda.name)
   ...:

# symbol table ignoring the names of pdarray created
>>> set(ak.list_symbol_table()) -  set(pda_names)
{'id_RjcKYUF_1', 'id_RjcKYUF_2', 'id_RjcKYUF_3', 'id_RjcKYUF_4'}
```

with this PR:
```python
>>> rng = ak.random.default_rng(11)

# initally no generators are in symbol table
>>> ak.list_symbol_table()
[]

>>> pda_names = []
   ...: for i in range(10):
   ...:     pda = rng.integers(20, size=3)
   ...:     pda_names.append(pda.name)
   ...:

# symbol table ignoring pda names shows many more generator entries have been added to the sym table
>>> set(ak.list_symbol_table()) -  set(pda_names)
{'id_DYklbur_1',
 'id_DYklbur_11',
 'id_DYklbur_13',
 'id_DYklbur_15',
 'id_DYklbur_17',
 'id_DYklbur_19',
 'id_DYklbur_3',
 'id_DYklbur_5',
 'id_DYklbur_7',
 'id_DYklbur_9'}
```

stress-tess (commented):
I have some concerns about the refactor of how generators are handled in `random`. If you're okay with it, I think we should leave that out for now since it seems a bit outside the scope of this PR

Other than that this looks good to go!

arkouda/pdarrayclass.py
ajpotts: Technically, pdarray.dtype is a dtype, and therefore it is equal to `akdtype("int64")`.  However, it will be satisfy equality with `'int64' `and `int64`, which is good for backward compatibility.

```
In [11]: x = ak.arange(10)

In [12]: x.dtype == 'int64'
Out[12]: True

In [13]: x.dtype == ak.dtypes.dtype('int64')
Out[13]: True

In [14]: x.dtype == ak.dtypes.int64
Out[14]: True

In [15]: type(x.dtype)
Out[15]: numpy.dtypes.Int64DType

In [16]: type(ak.dtypes.dtype('int64'))
Out[16]: numpy.dtypes.Int64DType

In [17]: type( ak.dtypes.int64)
Out[17]: type
```

arkouda/random/_generator.py
ajpotts: It should be fixed now.  Sorry about that.

stress-tess (approved):
looks good to me!

PR #3418, opened 2024-07-09
Closes #3417: Separate Parquet string read code from generic read function

### LINKED ISSUE #3417: Separate Parquet string read code from generic read function
Previously, the readColumnByName function included logic for handling strings directly within the function itself. This approach made it difficult to modify string-specific behavior because changes required updates to the entire function header across all call sites. Additionally, this approach made the code less clear as string handling logic was mixed with generic data type handling.

This pull request separates the string read code into a dedicated function. This improves code maintainability and readability by:
- Isolating string-specific logic, making it easier to modify without affecting other data types.
- Enhancing clarity by separating generic data type handling from string handling.

bmcdonald3:
Previously, the readColumnByName function included logic for handling strings directly within the function itself. This approach made it difficult to modify string-specific behavior because changes required updates to the entire function header across all call sites. Additionally, this approach made the code less clear as string handling logic was mixed with generic data type handling.

This pull request separates the string read code into a dedicated function. This improves code maintainability and readability by:
- Isolating string-specific logic, making it easier to modify without affecting other data types.
- Enhancing clarity by separating generic data type handling from string handling.

closes #3417

### DIFF
--- a/src/parquet/ReadParquet.cpp
+++ b/src/parquet/ReadParquet.cpp
@@ -95,6 +95,65 @@ void readColumnIrregularBitWidth(void* chpl_arr, int startIdx, std::shared_ptr<p
   }
 }
 
+int cpp_readStrColumnByName(const char* filename, void* chpl_arr, const char* colname, int64_t batchSize, char** errMsg) {
+  try {
+    int64_t ty = cpp_getType(filename, colname, errMsg);
+  
+    std::unique_ptr<parquet::ParquetFileReader> parquet_reader =
+      parquet::ParquetFileReader::OpenFile(filename, false);
+
+    std::shared_ptr<parquet::FileMetaData> file_metadata = parquet_reader->metadata();
+    int num_row_groups = file_metadata->num_row_groups();
+
+    int64_t i = 0;
+    for (int r = 0; r < num_row_groups; r++) {
+      std::shared_ptr<parquet::RowGroupReader> row_group_reader =
+        parquet_reader->RowGroup(r);
+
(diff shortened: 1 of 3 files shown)


src/parquet/ReadParquet.cpp
```diff
@@ -95,6 +95,65 @@ void readColumnIrregularBitWidth(void* chpl_arr, int startIdx, std::shared_ptr<p
...
+      }
+    }
+    return 0;
+  } catch (const std::exception& e) {
+    *errMsg = strdup(e.what());
+    return ARROWERROR;
+  }
+}
+
 int cpp_readColumnByName(const char* filename, void* chpl_arr, bool* where_null_chpl, const char* colname, int64_t numElems, int64_t startIdx, int64_t batchSize, int64_t byteLength, bool hasNonFloatNulls, char** errMsg) {
```
stress-tess, line 157: we should be able to remove the `if(ty == ARROWSTRING)` case of `cpp_readColumnByName` now right?

stress-tess (approved):
looks good to me!

src/parquet/ReadParquet.cpp
bmcdonald3: I think still want it as an extra line of defense. We get the type just prior to this and if by some error is actually not a string column, then that would crash the code, where here it would throw an error. I don't see much of a downside, so that's why I left it in, even though the upside is probably incredibly unrealistic

stress-tess: oh I didn't even think of that, makes sense to me!

PR #3420, opened 2024-07-10
Closes #3419: Remove intertwined list column and string column byte calculation logic

### LINKED ISSUE #3419: Remove intertwined list column and string column byte calculation logic
This pull request addresses a code structure that previously intertwined string list handling with regular string handling within the readColumnByName function. This tight coupling made it difficult to isolate the impact of changes on string read performance, which is a current focus area.

By separating string handling logic for lists from regular strings, we achieve the following benefits:

- Enhanced Performance Isolation: Modifications to either string type (list or regular) become more targeted, allowing for clearer performance evaluation of each change.
- Simplified Code Maintenance: Decoupling the logic improves code maintainability and reduces the risk of unintended side effects when modifying either string type.
This improved isolation will be crucial as we continue to optimize string read performance.

bmcdonald3:
This pull request addresses a code structure that previously intertwined string list handling with regular string handling within the readColumnByName function. This tight coupling made it difficult to isolate the impact of changes on string read performance, which is a current focus area.

By separating string handling logic for lists from regular strings, we achieve the following benefits:

- Enhanced Performance Isolation: Modifications to either string type (list or regular) become more targeted, allowing for clearer performance evaluation of each change.
- Simplified Code Maintenance: Decoupling the logic improves code maintainability and reduces the risk of unintended side effects when modifying either string type.

This improved isolation will be crucial as we continue to optimize string read performance.

Closes #3419

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -276,6 +276,29 @@ module ParquetMsg {
     return byteSizes;
   }
 
+  proc calcStrListSizesAndOffset(offsets: [] ?t, filenames: [] string, sizes: [] int, dsetname: string) throws {
+    var (subdoms, length) = getSubdomains(sizes);
+
+    var byteSizes: [filenames.domain] int;
+
+    coforall loc in offsets.targetLocales() with (ref byteSizes) do on loc {
+        var locFiles = filenames;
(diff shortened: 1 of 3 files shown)


stress-tess (approved):
looks good to me!

PR #3424, opened 2024-07-10
Improve Msg Function Registration for Module Tracking

bmcdonald3:
This pull request refines the logic for registering message-related functions (Msg functions) within the context of the saveUsedModules flag.

The `saveUsedModules` flag tracks modules that can be selectively included or excluded during the build process. However, including functions whose corresponding modules are always included by default creates unnecessary duplication in the generated function headers.

This PR introduces an explicit check to identify Msg functions with module names containing "Msg". These functions are skipped during registration, as their corresponding modules are inherently part of the build.

This will:
- Eliminates redundant function headers
- Ensure `saveUsedModules` accurately reflects modules subject to inclusion/exclusion
- Addresses build script errors caused by registering functions from always-included modules like GenSymIO


jeremiah-corrado (approved):
Looks good!

stress-tess (approved):

PR #3470, opened 2024-07-12
Closes #3441 deprecate import_export_test.py

### LINKED ISSUE #3441: deprecate import_export_test.py

ajpotts:
None of the tests from this module were in PROTO_tests, so I had to move the entire file and just did a minimal refactor so the tests would run.

Because they share a temp directory, I had to refactor io_test.py a bit.

Closes #3441 deprecate import_export_test.py


PROTO_tests/tests/import_export_test.py
```diff
@@ -0,0 +1,135 @@
...
+    def finalizer():
+        # Clean up any resources if needed
+        io_util.delete_directory(ie_test_base_tmp)
+
+    # Register the finalizer to ensure cleanup
+    request.addfinalizer(finalizer)
+    return ie_test_base_tmp
+
+
+class TestImportExport:
```
stress-tess, line 28: I thought these were covered in `io_test` (we tried to refactor to have all the io tests in the same file using class names to differentiate them)

PROTO_tests/tests/io_test.py
```diff
@@ -1523,8 +1575,6 @@ def test_csv_read_write(self):
 
 
 class TestImportExport:
```
stress-tess, line 1577: do this tests in this class cover the tests from `import_export_test.py`? or do they need to be updated

ajpotts: Thanks for pointing this out, I deleted the `import_export_test.py` file.

PROTO_tests/tests/import_export_test.py
ajpotts: Removed :)

stress-tess (approved):
looks great!

PR #3471, opened 2024-07-12
Closes #3446 deprecate logger_test.py

### LINKED ISSUE #3446: deprecate logger_test.py

ajpotts:
I verified all the tests had been moved over to PROTO_tests and then deprecated the tests.

Closes #3446 deprecate logger_test.py


drculhane (commented):
tests/logger_test.py and PROTO-tests/test/logger_test.py look almost identical, but I'm a bit puzzled that one of them tests for the name "LoggerTest" and the other for the name "TestLogger."  Since they both pass, this probably doesn't matter, but it does seem odd.

stress-tess (approved):

stress-tess:
> tests/logger_test.py and PROTO-tests/test/logger_test.py look almost identical, but I'm a bit puzzled that one of them tests for the name "LoggerTest" and the other for the name "TestLogger." Since they both pass, this probably doesn't matter, but it does seem odd.

@drculhane yah that was just a choice the previous team lead made during the switch over-- I can't remember the reasoning. But I believer that it should be the same for all the classes names in both test dirs

jaketrookman (approved):

PR #3474, opened 2024-07-15
Remove legacy 'supported_scalar_types' section from config file

jeremiah-corrado:
Removes the supported_scalar_types section from the registration-config.json file. With the new registration framework, this is replaced by the `dtype` section in the same json file.

As a part of this effort, the following commands that relied on the old type configurability system were refactored to use the new framework:
* cast
* set
* create
* randint

### DIFF
--- a/src/CastMsg.chpl
+++ b/src/CastMsg.chpl
@@ -15,345 +15,65 @@ module CastMsg {
   private config const logChannel = ServerConfig.logChannel;
   const castLogger = new Logger(logLevel, logChannel);
 
-  @arkouda.registerND
-  proc castMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, param nd: int): MsgTuple throws {
-    param pn = Reflection.getRoutineName();
-    const name = msgArgs.getValueOf("name"),
-          objtype = msgArgs.getValueOf("objType").toUpper(): ObjType,
-          targetDtype = str2dtype(msgArgs.getValueOf("targetDtype")),
-          opt = msgArgs.getValueOf("opt");
-    castLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),
-          "name: %s obgtype: %? targetDtype: %? opt: %?".format(
-                                                 name,objtype,targetDtype,opt));
+  proc isFloatingType(type t) param : bool {
+    return isRealType(t) || isImagType(t) || isComplexType(t);
+  }
 
-    select objtype {
-      when ObjType.PDARRAY {
-        var gse: borrowed GenSymEntry = getGenericTypedArrayEntry(name, st);
+  @arkouda.instantiateAndRegister(prefix="cast")
+  proc castArray(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab,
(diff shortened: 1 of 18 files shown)


src/CastMsg.chpl
```diff
@@ -15,345 +15,65 @@ module CastMsg {
   private config const logChannel = ServerConfig.logChannel;
   const castLogger = new Logger(logLevel, logChannel);
 
-  @arkouda.registerND
```
bmcdonald3, line 18: Wow, this is awesome being able to cut out all this code.

src/registry/Commands.chpl
```diff
@@ -280,21 +577,35 @@ proc ark_repeatFlat_bool_1(cmd: string, msgArgs: borrowed MessageArgs, st: borro
   return ManipulationMsg.repeatFlatMsg(cmd, msgArgs, st, array_dtype=bool, array_nd=1);
 registerFunction('repeatFlat<bool,1>', ark_repeatFlat_bool_1, 'ManipulationMsg', 903);
 
-import OperatorMsg;
-
-import ParquetMsg;
```
bmcdonald3, line 285: It isn't clear to me why we can get rid of all these imports in this file now and I'm not seeing where that is correspondingly changed in the Python scripts. What am I missing here?

bmcdonald3 (approved):
Mostly looks good, super cool to see all that duplicate code getting removed, just one question on the command registry.

src/registry/Commands.chpl
jeremiah-corrado: These can be removed because those modules don't have any `registerCommand` or `instantiateAndRegister` annotations yet. This was mostly fixed by: https://github.com/Bears-R-Us/arkouda/pull/3371/commits/f08fdf6d436871e845e06081c8b54ad4c21ffcae (in the previous PR), but didn't take affect because I also needed the change on line `619` of `register_commands.py` (added in this PR).

bmcdonald3 (approved):

src/RandMsg.chpl
```diff
@@ -29,137 +32,53 @@ module RandMsg
...
-                              "compute time = %i sec".format(Time.timeSinceEpoch().totalSeconds() - t2));
+                         "cmd: %s len: %i dtype: %s aMin: %?: aMax: %?".format(
+                         cmd,len,type2str(array_dtype),low,high));
 
-            const repMsg = "created " + st.attrib(rname);
-            randLogger.debug(getModuleName(),getRoutineName(),getLineNumber(),repMsg);
-            return new MsgTuple(repMsg, MsgType.NORMAL);
-        }
+        var t = new stopwatch();
+        t.start();
```
stress-tess, line 54: first time seeing the stopwatch. I've still been using `timeSinceEpoch().totalSeconds()`, I'll have to remember this

stress-tess (approved):
Looks great!! Really showing off the power of the refactor in this one! 🎉

src/RandMsg.chpl
stress-tess: in fact I'll go update the chpl tutorial to use it now!

PR #3477, opened 2024-07-16
Remove support for pre-2.0 versions of Chapel

jeremiah-corrado:
Removes support for Chapel 1.32 and 1.33. CI is updated to run on version 2.1 with compatibility checking for version 2.0.

Modifies the makefile to run `register-commands` (added [here](https://github.com/Bears-R-Us/arkouda/pull/3371)) when building the server. This requires installing the chapel-frontend python bindings. They can be installed with `(cd $CHPL_HOME/tools/chapel-py && pip install .)`. The CI script was modified to include this step before building the server.

`Commands.chpl` is removed from git tracking because it is now automatically generated when running `make`.


bmcdonald3 (approved):
🎉 very happy to see those old compat modules going away.

Do we still need the 2.0+ compat modules? Are there differences between 2.0 and 2.1/2.2 that we are tracking?

I know that I have seen a couple of deprecations popping up, but haven't looked deeply enough to know if they are deprecations requiring compat modules or ones where we can jsut make global changes. Have you looked into it enough to know the answer to that?

jeremiah-corrado:
> Do we still need the 2.0+ compat modules? Are there differences between 2.0 and 2.1/2.2 that we are tracking?

I don't think we need them currently, but I wanted to keep the framework in place for future changes. There may be some changes to the `Sort` module coming in with 2.2 but I'm not sure whether they'll affect Arkouda.

stress-tess (approved):
This looks great! Thanks so much @jeremiah-corrado!

PR #3503, opened 2024-07-16
Closes #3475 make fails when lib and lib64 directories are both present

### LINKED ISSUE #3475: make fails when lib and lib64 directories are both present
After this PR:
https://github.com/Bears-R-Us/arkouda/pull/3423

`make` fails if both `lib` and `lib64` are present and the necessary dependencies are in `lib`.  

@stress-tess @brandon-neth @bmcdonald3

ajpotts:
This fixes the issue described in #3475.  The solution was to revert changes in PR #3423 .

Closes #3475 make fails when lib and lib64 directories are both present


bmcdonald3 (approved):
Thanks for putting a fix together for this Amanda, sorry about the extra work!

stress-tess (approved):

PR #3506, opened 2024-07-17
Adds skip configuration for multidimensional histogram test

brandon-neth:
The multidimensional histogram test was failing for unknown reason on Horizon. This PR adds a skip condition so it doesn't run on Horizon.


bmcdonald3 (approved):
For context, the machine that this is failing on is a 16-node-xc that is being phased out and won't be around for too much longer. We don't know for sure what it is exactly causing the failure, but it is not showing up in other configurations and this will help us make more sense out of nightly testing results if we don't have this sporadic failure popping up.

@stress-tess are you alright disabling this test in this specific environment for now? We should have an alternate multilocale machine with more modern hardware (and IB interconnect) online soon and if this shows up there, that will trigger the alarm bells that we need to figure this out, but this specific configuration does not.

stress-tess:
> For context, the machine that this is failing on is a 16-node-xc that is being phased out and won't be around for too much longer. We don't know for sure what it is exactly causing the failure, but it is not showing up in other configurations and this will help us make more sense out of nightly testing results if we don't have this sporadic failure popping up.
> 
> @stress-tess are you alright disabling this test in this specific environment for now? We should have an alternate multilocale machine with more modern hardware (and IB interconnect) online soon and if this shows up there, that will trigger the alarm bells that we need to figure this out, but this specific configuration does not. edit: well, it doesn't trigger the bells for me at least 😄

Yah that sounds good to me!

PR #3507, opened 2024-07-17
Closes #3378: Update chpl tutorial

### LINKED ISSUE #3378: Update chpl tutorial and improve copy aggregation section
Update our chapel tutorial to make sure everything still runs as expected and update the explanation of copy aggregation

stress-tess:
This PR (closes #3378) is a rather substational update of the chpl tutorial since it's been quite a while since anyone has looked at it. I will say never revisit an explanation you wrote years ago 😬 

Given there's so many changes and moving things around, I think it probably easier to just look at the md file on my fork than the diff:
https://github.com/stress-tess/arkouda/blob/3378_tutorial_update/training/CHAPEL_TUTORIAL.md

___
I added sections, changed explanations, removed unnecessarily complicated sections, and attempted to better motivate concepts. I tried my best to address any spots where new chapel is functionality is introduced with no explanation / link. I also removed the numbering since it made it really hard to update anything (who's idea was that anyway! 🫠 )

I think this is a big improvement from what we had, but I did lose steam towards the end so I won't be surprised if there are lots of mistakes.

I've added some chapel devs as reviews but don't feel like you need to look at this! I would appreciate it if anyone gets a chance, especially to make sure I'm not saying anything that's flat out wrong


training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1072,613 @@ Expected Output:
...
+
+For the best performance we want to minimize how often we move data between locales. But there are situations
+where you need to move lots of data between locales. For example assigning local values into a distributed array
+at random indices. In this case, one locale might be copying multiple elements to every other locale.
+Sending each of those elements one at a time results in a lot of communication. Aggregation is used to mitigate
+this.
+
+I like to think of this in terms of transporting people, where our goal is minimizing the total number of trips.
+If a bunch of people are going to the same place, they could all travel there individually. But we'd have
+fewer total trips if we wait until we have enough people to fill a bus and send a whole group at once.
```
bmcdonald3, line 1269: ❤️

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1072,613 @@ Expected Output:
...
 
-B = 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15
+<a id="compile-multiloc"></a>
+### Enabling multiple locales
+For the rest of this tutorial, we will be using Chapel with
+[multiple locales](https://chapel-lang.org/docs/users-guide/locality/compilingAndExecutingMultiLocalePrograms.html).
+
+If you're not running on a distributed system, Chapel will simulate running in a distributed way as if you had
+multiple distinct machines. This is useful for running diagnostics and finding bugs that only show up
+in a distributed setting.
```
bmcdonald3, line 1173: Maybe note that the performance is pretty much meaningless when running with gasnet on a non-distributed achine?

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1072,613 @@ Expected Output:
...
 
-| locale | execute_on_nb |
-| -----: | ------------: |
-|      0 |             3 |
-|      1 |             0 |
-|      2 |             0 |
-|      3 |             0 |
-looping over arrays directly time = 0.0310571
+Check out this [actual discussion](https://github.com/Bears-R-Us/arkouda/pull/2159#discussion_r1113699483)
+in arkouda that highlights a lot of these points.
```
bmcdonald3, line 1676: **A**rkouda <- capital!

bmcdonald3 (approved):
I must admit, I didn't read through most of the examples, but did real (I believe) all of the text and no major hiccups for me there! This is awesome, probably one of the best Chapel learning materials out there!

bmcdonald3:
I had a couple more hearts on sections that I liked, I don't know why those didn't make it through...

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1072,618 @@ Expected Output:
...
-// use a var if you expect a value to change
-var myVar = [0, 1, 2];
-// we use a const because we don't know the value at compilation time
-const myConst = helper(myBool);
-// we use a param becasue we know what the value is at compilation time
-param myParam = 17;
+> Say `x` and `y` are both stored on `locale_i`. From the perspective of `locale_i`, we say `x` and `y` are
+both local. There no significant difference in access times.
+> 
+> But if `x` is on `locale_i` and `y` is on `locale_j`. From the perspective of `locale_i`, we say `x` is local
```
jaketrookman, line 1103: I think it would be helpful to also add the third case of `y` being local and `x` being remote. Not sure if it is too far of a leap at this point to make a reference to how these are used in Source and Destination Aggregators, or at least maybe link the sections.

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1072,618 @@ Expected Output:
...
+[Copy aggregation](https://chapel-lang.org/docs/modules/packages/CopyAggregation.html#copyaggregation)
+is the same idea. Instead of copying each value individually, we wait until we have a bunch
+that are all going to the same locale and copy them over all at once.
+
+There are two types of copy aggregation:
+* copying local values into remote variables (`DstAggregator`)
+* copying remote values into local variables (`SrcAggregator`)
+
+It's important to note copy aggregation will only work if at least one side is local.
+Both sides being remote (remote-to-remote aggregation) is not supported.
```
jaketrookman, line 1285: I really do think it is helpful to add the description of the different perspectives of how deciding when we are using Source versus Destination Aggregators, based on the relative locale versus remote locale for each case.

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1072,618 @@ Expected Output:
...
 
-In this example we will take `hashes`, a distributed array contain a tuple of uints `(uint, uint)`. And copy the values
-into `upper` and `lower`, 2 distributed `uint` arrays.
-```Chapel
-use BlockDist, CommDiagnostics, Time;
+When you do a `forall` loop over a distributed array, each locale will handle the values that are local to it.
+So we can go about this 2 different ways:
+* We could loop over the `src` array, so the source values will always be local. We then copy our local values
+into a remote position of the `dst` array using a `DstAggregator`
+* Or we loop over the `dst` array, so destination values are local. We then want to copy remote values from
```
jaketrookman, line 1311: You definitely state them here, but it might be worth regrouping them with where they are defined up above and/or adding the more detailed `x` on `locale_i` and 'y` on `local_j` example you previously used.

jaketrookman (commented):
Absolutely love the overhaul and everything is significantly more clear and understandable. I think there are just a few places where we can clean up the explanation of aggregation based on how you walked through it to us as a group that might add more insight and be more helpful to new users.

training/CHAPEL_TUTORIAL.md
stress-tess: I appreciate the feedback! It certainly wouldn't be hard to update, but I think it'd just be duplicating the second case but with ther variable names swapped

like in my head the three possible cases are:
- both local
- one local, one remote
- both remote

so since we chose `x` and `y` arbitrarily, I think these both fall under the second

I def see why you jumped to copy aggregation since you're already primed to think about that. But here we're just defining the terms local and remote

training/CHAPEL_TUTORIAL.md
stress-tess: Yah I think this is a good idea! I could add a bit talking about how to the same action can be seen as either a `GET/SrcAggregator ` or a `PUT/DstAggregator` depending on which locale's perspective we are viewing it from (i.e. which locale is doing the computation

training/CHAPEL_TUTORIAL.md
stress-tess: tbh I could prob just copy this section and say "in the following example ..."

stress-tess: yah! we could def move this up

training/CHAPEL_TUTORIAL.md
jaketrookman: That makes sense to me

training/CHAPEL_TUTORIAL.md
stress-tess: updated! I like this much better with extra explanation switching between the locale's perspectives. Thanks @jaketrookman!

jaketrookman: Love the update.

jaketrookman (approved):
Great work!

training/CHAPEL_TUTORIAL.md
stress-tess: we chatted offline and decided to leave this bit where it is. Instead we decided to add an example write below the definitions that's structured similarly to the one used to define `remote` and `local`

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1073,635 @@ Expected Output:
...
+Since we're viewing this from perspective of `locale_i`, positions on `locale_j` are remote.
+So we need a `DstAggregator` because we are putting a local value into a remote position.
+>
+> If instead `locale_j` is the one doing this computation, the position
+we want to write into is local.
+And from `locale_j`'s perspective `x` is remote. 
+Since we're getting a remote value and writing it into a local position, we need a `SrcAggregator`.
+
+It's important to note copy aggregation will only work if at least one side is local.
+Both sides being remote (remote-to-remote aggregation) is not supported.
```
bradcray, line 1303: You could say "not currently supported" here, if desired (in the sense that I believe it could be added if someone really wanted/needed it in the future).

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1073,635 @@ Expected Output:
...
+
+src: 0 1 2 3 4 5
+dst: 3 4 5 0 1 2
+```
 
-var upperEntry = new SymEntry(SIZE, uint);
-var lowerEntry = new SymEntry(SIZE, uint);
+<a id="perf"></a>
+## Performance and Diagnostics
+Now that you're a pro at writing chapel code, let's talk about how to make the code you write more efficient!
```
bradcray, line 1500: This makes me think:  What do you think about writing your aggregator examples without the aggregators first, to show the simpler version of the computation you're expressing followed by the more complex (but efficient) form?  (in which case, maybe this mention of efficiency should come just before showing the aggregators).

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1073,635 @@ Expected Output:
...
+      // since dst is remote, we use a dst aggregator
+      // assignment without aggregation would look like:
+      // dst[ (i + 3) % 6 ] = s
+      agg.copy(dst[ (i + 3) % 6 ], s);
+    }
+    writeln();
+    writeln("src: ", src);
+    writeln("dst: ", dst);
+}
+else {
```
bradcray, line 1380: Showing both the source and destination aggregator in one code example results in a lot of code.  What do you think about picking just one of them to start with (does Arkouda use one more than the other?) and then showing the second example in a details section or something like that?

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1073,635 @@ Expected Output:
...
-// the leading iterator is a range, so all the computation happens on locale 0
-forall (i, (up, low)) in zip(0..#SIZE, hashes) {
-  upperEntry.a[i] = up;
-  lowerEntry.a[i] = low;
-}
+[`const`](https://chapel-lang.org/docs/users-guide/base/constParam.html):
+* can be used when we know a variable should never be changed
+* a program execution time constant. This means the compiler doesn't need to know its value, and it can be 
+set at runtime (i.e. passed in by a user)
+* knowing this value won't change can help the compiler to make optimizations
```
bradcray, line 1511: You could add something in this list about it being a way for a programmer to keep themselves honest / check themselves.  I.e., "I don't want this to ever change, and want the compiler to complain at me if I do try to re-assign it."

I also wonder whether the final bullet should be strengthened by saying something like "programmers should always reach for `const` when they have variables that they know won't be changing [for both the performance and code safety benefits]"

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1073,635 @@ Expected Output:
...
-writeln();
-printCommDiagnosticsTable();
+[`const ref`](https://chapel-lang.org/docs/language/spec/variables.html?highlight=ref#ref-variables)
+* same as a `ref` but cannot be changed
+
+<a id="diagnostics"></a>
+### Diagnostics
+
+There are several chapel modules available to aid in optimizing your code. The most common ones
+used by the arkouda team are [comm](https://chapel-lang.org/docs/modules/standard/CommDiagnostics.html)
```
bradcray, line 1530: Maybe 'CommDiagnostics'?  (and 'Time' rather than 'time'?)

Or change "are comm and time" to "measure communication and time"?

training/CHAPEL_TUTORIAL.md
```diff
@@ -993,358 +1073,635 @@ Expected Output:
...
+|      3 |   7 | 250000 |             0 |
+
+copy with aggregation time = 0.159691
+communication using aggregation:
+| locale | get | put | execute_on | execute_on_nb |
+| -----: | --: | --: | ---------: | ------------: |
+|      0 |  80 |  40 |         40 |             3 |
+|      1 |  80 |  40 |         40 |             0 |
+|      2 |  80 |  40 |         40 |             0 |
+|      3 |  80 |  40 |         40 |             0 |
```
bradcray, line 1630: Do you have any interest in defining what these different categories are?  It's a bit of a rathole, but also natural to wonder about.  My high-level explanation would be:

* "get": reading a value from a remote locale's memory
* "put": storing a value into a remote locale's memory
* "execute_on": creating a task on a remote locale, for example via an `on`-clause
* "_nb": non-blocking (doesn't prevent the original task from continuing)

bradcray (commented):
This is looking great, Tess, and you're welcome to write tutorial examples for us anytime!  :D

I only found one actual bug and many of my comments are aimed at improving clarity or connecting the dots slightly better for a new reader.  Some are providing information that is probably too much for the tutorial itself, but may be helpful to you if you don't already know.  Some may be more work than are worthwhile at this time (e.g., few are things I'd hold up the PR over).  Some are just nits, some of which can be ignored.

Thanks for putting this together!  I really like the concept of tutorials aimed at a specific application's developer community as a means of focusing down on "things we use" vs. "things Chapel has but that we don't."

training/CHAPEL_TUTORIAL.md
```diff
@@ -365,65 +477,57 @@ Expected Output:
...
-The most common use case of `coforalls` will be introduced in a later section.
+Notice we have an [unbounded range](https://chapel-lang.org/docs/primers/ranges.html#variations-on-basic-ranges),
+`1..`, so the end bound is determined by the size of the other iterables.
+Since in this case the other iterables are length 5, `1..` is equivalent to `1..5`.
 
 <a id="ternary"></a>
 ## Ternary
-A ternary statement is where a variable can have one of two possible values depending on whether a condition is True or False.
+A ternary statement can have one of two possible values depending on whether a condition is true or false.
+Because Chapel is a strongly typed language, both values must be the same type.
```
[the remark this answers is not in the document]
stress-tess: oh cool I didn't know that!

ajpotts:
I know this is closed now, but I did review and it's great!  Thanks for updating this!

PR #3509, opened 2024-07-18
Closes #3421 testing equivalence module

### LINKED ISSUE #3421: testing equivalence module
Create a module that will have set of asserters to test whether an arkouda object is equivalent to a pandas or numpy object.

ajpotts:
This creates a new testing module that has asserters to check whether an arkouda object is equivalent to a matching numpy or pandas object.

Closes #3421 testing equivalence module


stress-tess (approved):
looks good!!

PR #3516, opened 2024-07-18
Add initial implementation for Sparse Matrix Mult

ShreyasKhandekar:
This PR does the following:

1. Add new types to the Symbol Table:
  - `SparseGenSymEntry` and `SparseSymEntry` are added to support sparse matrices.
2. Add a new type to the client
  - `sparray`  is the client representation of sparse matrices
3. Add functionality for this new type
  - New SparseMatrix module copied from https://github.com/chapel-lang/chapel/pull/25269/ which has:
  - Factory function to create a random 2D sparse array: `ak.random_sparse_matrix`
  - Function to multiply two sparse arrays: `ak.sparse_matrix_matrix_mult`

I am not adding the new SparseMatrixMsg to ServerModule.cfg by default today because it is an unstable prototype and may change in the future.

Future work here is to add support for printing the sparrays and add a test for the new functions.

### DIFF
--- a/arkouda/sparrayclass.py
+++ b/arkouda/sparrayclass.py
@@ -8,13 +8,11 @@ import numpy as np
 from typeguard import typechecked
 
 from arkouda.client import generic_msg
-from arkouda.dtypes import dtype
-from arkouda.dtypes import (
-    int_scalars
-)
+from arkouda.dtypes import dtype, int_scalars
 from arkouda.logger import getArkoudaLogger
 
 logger = getArkoudaLogger(name="sparrayclass")
+
 
 class sparray:
     """
@@ -82,6 +80,9 @@ class sparray:
     def __len__(self):
         return reduce(lambda x, y: x * y, self.shape)
 
+    def __getitem__(self, key):
+        raise NotImplementedError("sparray does not support __getitem__")
+
     # def __str__(self): # This won't work out of the box for sparrays need to add this in later
     #     from arkouda.client import pdarrayIterThresh
 
@@ -91,6 +92,7 @@ class sparray:
     #     from arkouda.client import pdarrayIterThresh
 
     #     return generic_msg(cmd="repr", args={"array": self, "printThresh": pdarrayIterThresh})
+
 
 # creates sparray object
 #   only after:
@@ -128,7 +130,6 @@ def create_sparray(repMsg: str, max_bits=None) -> sparray:
         mydtype = fields[2]
         size = int(fields[3])
         ndim = int(fields[4])
-        
 
         if fields[5] == "[]":
             shape = []
--- a/arkouda/sparrayclass.py
+++ b/arkouda/sparrayclass.py
@@ -1,0 +1,146 @@
+from __future__ import annotations
+
+import builtins
+from functools import reduce
+from typing import Optional, Sequence, Union
+
+import numpy as np
+from typeguard import typechecked
+
+from arkouda.client import generic_msg
+from arkouda.dtypes import dtype
+from arkouda.dtypes import (
+    int_scalars
+)
+from arkouda.logger import getArkoudaLogger
+
+logger = getArkoudaLogger(name="sparrayclass")
+
+class sparray:
+    """
+    The class for sparse arrays. This class contains only the
+    attributies of the array; the data resides on the arkouda
+    server. When a server operation results in a new array, arkouda
+    will create a sparray instance that points to the array data on
+    the server. As such, the user should not initialize sparray
+    instances directly.
+
+    Attributes
+    ----------
+    name : str
+        The server-side identifier for the array
+    dtype : dtype
+        The element type of the array
+    size : int_scalars
+        The size of any one dimension of the array (all dimensions are assumed to be equal sized for now)
+    ndim : int_scalars
+        The rank of the array (currently only rank 2 arrays supported)
+    shape : Sequence[int]
+        A list or tuple containing the sizes of each dimension of the array
+    layout: str
+        The layout of the array ("CSR" or "CSC" are the only valid values)
+    itemsize : int_scalars
+        The size in bytes of each element
+    """
+
+    def __init__(
+        self,
+        name: str,
+        mydtype: Union[np.dtype, str],
+        size: int_scalars,
+        ndim: int_scalars,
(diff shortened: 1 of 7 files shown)


src/SparseMatrix.chpl
```diff
@@ -0,0 +1,346 @@
...
+
+    var C = makeSparseMat(A.domain.parentDom, spsData);
+
+    const elapsed = time.elapsed();
+
+    if countComms {
+      stopCommDiagnostics();
+      printCommDiagnosticsTable();
+      writeln();
+    }
```
bmcdonald3, line 29: We should probably remove this debug output for production (along with the timings)

src/SparseMatrix.chpl
```diff
@@ -0,0 +1,346 @@
...
+  proc sparseMatMatMult(A, B) where (!A.chpl_isNonDistributedArray() &&
+                                     !B.chpl_isNonDistributedArray()) {
+    var CD = emptySparseDomLike(B);  // For now, hard-code C to use CSR, like B
+    var C: [CD] int;
+
+    ref targLocs = A.targetLocales();
+  
+    if countComms then startCommDiagnostics();
+    var time: stopwatch;
+    time.start();
```
bmcdonald3, line 56: Same here, I don't think we want this in production.

src/SparseMatrix.chpl
```diff
@@ -0,0 +1,346 @@
...
+    config const seed = 0,
+                printSeed = seed == 0;
+
+    var rands = if seed == 0 then new randomStream(real)
+                          else new randomStream(real, seed);
+
+    // print library-selected seed, for reproducibility
+    //
+    if printSeed then
+      writeln("Using seed: ", rands.seed);
```
bmcdonald3, line 176: Won't keep calling them out, but all these debug prints should go.

src/SparseMatrixMsg.chpl
```diff
@@ -0,0 +1,89 @@
...
+    proc randomSparseMatrixMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
+        var repMsg: string; // response message with the details of the new arr
+
+        var vName = st.nextName(); // symbol table key for resulting sparse arr
+
+        var size: int = msgArgs.get("size").getIntValue();
+        var density: real = msgArgs.get("density").getRealValue();
+        // var shape: always 2d?
+        var l: string = msgArgs.getValueOf("layout");
+        // var distributed: bool = msgArgs.getBoolValueOf("distributed"); TODO does this need to be configurable?
```
bmcdonald3, line 34: Let's remove these comments -- arrays are always distributed in Arkouda, it does not need to be configurable.

src/SparseMatrix.chpl
```diff
@@ -0,0 +1,346 @@
...
+      writeln(msg);
+
+      for r in DenseDom.dim(0) {
+        for c in DenseDom.dim(1) {
+          write(Arr[r,c], " ");
+        }
+        writeln();
+      }
+      writeln();
+    }
```
bmcdonald3, line 258: We could just have the Arkouda `__str__` function call this in a truncated format for the prints.

arkouda/sparrayclass.py
```diff
@@ -0,0 +1,146 @@
...
+from arkouda.client import generic_msg
+from arkouda.dtypes import dtype
+from arkouda.dtypes import (
+    int_scalars
+)
+from arkouda.logger import getArkoudaLogger
+
+logger = getArkoudaLogger(name="sparrayclass")
+
+class sparray:
```
bmcdonald3, line 19: @stress-tess This is the effort that MikeO and Brad have been taking about regarding sparse matrix multiplication in Arkouda. What's going on here is that we've added a sparse matrix class to the Python side and the name chosen here is `sparray`.

I am wondering your thoughts on the naming here, are you alright with `sparray` for that? I wonder if having matrix, rather than array would be good, but naming has never been a strength of mine... 

I think the reason why this feels odd to me is even the creation functions are called like `random_sparse_matrix`, which seems it should logically return something with matrix in the name, but instead you get an array.

bmcdonald3 (approved):
Mainly looks good, I think we want to remove the debug output though. Also, one question on naming that I've pinged Tess about.

bmcdonald3 (commented):
Whoops, missed this first time around, but looks like the failing CI checks will need to be resolved as well.

arkouda/sparrayclass.py
ShreyasKhandekar: The reason I decided to call it `sparray` here is a parallel to `pdarray` but it's `sp`arse.
The class also has an `ndim` field, which right now is always 2, so it is a 2D matrix, but there isn't anything from people creating more functions that create just a 1D sparse array and operate on that.

src/SparseMatrix.chpl
ShreyasKhandekar: I think we discussed that we should not print the array as if it were dense (Which is what this function does), rather we want to represent the arrays which have the non IRV values, and the corresponding indices (depending on if the matrix is CSC or CSR for the right representation there).

ShreyasKhandekar:
The error with `make doc` might be an issue with `chpldoc` I've pinged the Chapel team to see if thats the case.

arkouda/sparrayclass.py
ShreyasKhandekar: As a side note scipy also calls their sparse arrays `sparray` (`spmatrix` for 2D), the fact that I also chose `sparray` is completely coincidental, but a nice alignment with the existing name out there.

https://docs.scipy.org/doc/scipy/reference/sparse.html#sparse-matrix-classes

bradcray: >  but a nice alignment with the existing name out there.

I wonder if we'd want to call ours something like sppdarray or pdsparray to distinguish the two and nod to Arkouda's pd ("parallel distributed") arrays—not a name I'm crazy about, but it's a precedent.

bmcdonald3: After discussing with Shreyas, I think that we are going to merge this with the name as-is for now with the current name and then leave it open for further discussion as we continue to make improvements here.

bmcdonald3 (approved):

PR #3517, opened 2024-07-18
Prototype arkouda.for_each method.

wlav:
This PR is a WIP, added on request; it's an initial prototype showing how an `arkouda.for_each` method could be implemented. A true production version should take LLVM IR on the client side and send that instead of pickling the functor, then JIT it using ORCJit on the server. (Note that LLVM IR is not as stable as Python's pickle and thus there may be deviations between LLVM used in Chapel and LLVM used in Numba.) Doing so would remove the dependency on Python for the server.

There is a client-side dependency added on `cloudpickle`. That module allows for complete packaging of what is pickled as opposed to the standard pickle which only adds references to modules that are subsequently imported when unpickling. This is necessary since it can not be expected that client-side Python modules are available server-side. Going to LLVM IR instead would remove the need of this dependency.

`PythonMsg` needs to be uncommented in `ServerModules.cfg` to enable it, and building/linking with Python requires `ARKOUDA_PYTHON_SERVER=1` when running `make`.

Some rough numbers as to why we're doing this; single node, all local, 10^10 float64s, a 1D p3:
```
numpy, externalized loop:      1hr 50min (est.)
arkouda, externalized loop:    178days (est.)
numpy.poly1d:                  92s
C++ vector -O3:                8s
arkouda.for_each:              3s
```
I'm somewhat new to Chapel; I'm presuming that it outperforms C++ b/c the code is either vectorized or threaded. `numpy.poly1d` is that much slower because of the creation of temporary arrays and subsequent memory copies. The `arkouda.for_each` timing includes running the JIT (no warmup; Numba itself also adds a one-off overhead of ~0.3s here).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,9 @@ CHPL := chpl
 CHPL_FLAGS += --ccflags="-DH5_USE_110_API"
 
 # Python
+ifdef ARKOUDA_PYTHON_SERVER
 CHPL_FLAGS += $(shell python3-config --includes) $(shell python3-config --ldflags) -lpython3.$(shell python -c "import sys; print(sys.version_info.minor)")
+endif
 
 # silence warnings about '@arkouda' annotations being unrecognized
 CHPL_FLAGS += --using-attribute-toolname arkouda
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ CHPL := chpl
 # We need to make the HDF5 API use the 1.10.x version for compatibility between 1.10 and 1.12
 CHPL_FLAGS += --ccflags="-DH5_USE_110_API"
 
+# Python
+CHPL_FLAGS += $(shell python3-config --includes) $(shell python3-config --ldflags) -lpython3.$(shell python -c "import sys; print(sys.version_info.minor)")
+
 # silence warnings about '@arkouda' annotations being unrecognized
 CHPL_FLAGS += --using-attribute-toolname arkouda
 
--- a/setup.py
+++ b/setup.py
@@ -130,6 +130,7 @@ setup(
         "tables>=3.7.0",
         "pyarrow",
         "scipy<=1.13.1",
+        "cloudpickle",
     ],
     # List additional groups of dependencies here (e.g. development
     # dependencies). Users will be able to install these using the "extras"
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -130,6 +130,7 @@
         "tables>=3.7.0",
         "pyarrow",
         "scipy<=1.13.1",
+        "cloudpickle",
```
ajpotts, line 133: `cloudpickle` would have to be added to `arkouda-env-dev.yml`,  `arkouda-env.yml`, and `pydoc/requirements.txt`.

wlav: Yes.

Aside, I'm currently working on an alternative implementation that sends LLVM IR and thus doesn't require an embedded Python interpreter on the server side (and hence no cloudpickle on the client side). If successful (it's looking good), that would supersede this prototype implementation. Thanks!

ajpotts:
Just a note, to compile this PR you need to set `export ARKOUDA_PYTHON_SERVER=True`.

Makefile
```diff
@@ -17,6 +17,11 @@ CHPL := chpl
 # We need to make the HDF5 API use the 1.10.x version for compatibility between 1.10 and 1.12
 CHPL_FLAGS += --ccflags="-DH5_USE_110_API"
 
+# Python
+ifdef ARKOUDA_PYTHON_SERVER
+CHPL_FLAGS += $(shell python3-config --includes) $(shell python3-config --ldflags) -lpython3.$(shell python -c "import sys; print(sys.version_info.minor)")
```
stress-tess, line 22: A note that running `python3-config --ldflags` on mac ends with `'-framework CoreFoundation'` and chpl doesn't know how to handle that. It causes the make to error with `Unrecognized flag: '-f'`

This is prob not the best way to do this, but I modified the make file to remove this suffix. That made it so I could build the server
```
python_ld_flags=$(shell python3-config --ldflags)

# mac has '-framework CoreFoundation' and chpl doesn't know how to deal with it
# so we're gonna remove it
mac_framework_suffix="-framework CoreFoundation"

CHPL_FLAGS += $(shell python3-config --includes) ${python_ld_flags%"$mac_framework_suffix"}  -lpython3.$(shell python -c "import sys; print(sys.version_info.minor)")
```

PR #3518, opened 2024-07-18
Stop requiring manual installation of `chapel-py` to register commands

jeremiah-corrado:
Following https://github.com/Bears-R-Us/arkouda/pull/3477, the chapel frontend python bindings (`chapel-py`) are required to build Arkouda. Because of this, chapel-py needed to be installed in the local python environment before running `make`.

This PR adjusts the makefile s.t., the prebuilt bindings shipped with Chapel are used to run `register-commands`. This avoids the need to manually install/build `chapel-py` before building Arkouda.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -566,7 +566,7 @@ version:
 #####################
 
 register-commands:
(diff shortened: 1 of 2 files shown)


stress-tess (approved):
looks great! thanks jeremiah!

jaketrookman (approved):

PR #3520, opened 2024-07-18
Closes #3519, #3522: unpin hdf5 library

### LINKED ISSUE #3519: unpin hdf5 library
The pandas io library is throwing the following error: 
```
E                   ImportError: Pandas requires version '3.8.0' or newer of 'tables' (version '3.7.0' currently installed).
```

But in order to upgrade `pytables`, `hdf5` needs to be upgraded as well, otherwise conda fails to solve environment.

### LINKED ISSUE #3522: local failing of test_import_hdf
local failing of test_import_hdf in `tests\impor_export_tests.py`
```
E                   ImportError: Pandas requires version '3.8.0' or newer of 'tables' (version '3.7.0' currently installed).
```
This was on ubuntu using the `arkouda-dev` conda environment.

ajpotts:
This changes was implemented in order to address the following problem:
local failing of test_import_hdf in `tests\impor_export_tests.py`:
```
E                   ImportError: Pandas requires version '3.8.0' or newer of 'tables' (version '3.7.0' currently installed).
```

The solution is to upgrade `tables` to `3.8.0`, which requires a newer version of `hdf5`.

Closes #3519 unpin hdf5 library
Closes #3522 local failing of test_import_hdf

### DIFF
--- a/arkouda-env-dev.yml
+++ b/arkouda-env-dev.yml
@@ -12,10 +12,10 @@ dependencies:
   - versioneer
   - matplotlib>=3.3.2
   - h5py>=3.7.0
-  - hdf5==1.12.2
+  - hdf5>=1.12.2
   - pip
   - types-tabulate
-  - pytables>=3.7.0
+  - pytables>=3.8.0
   - pyarrow
   - libiconv
   - libidn2
--- a/Makefile
+++ b/Makefile
@@ -127,8 +127,8 @@ install-zmq:
 	rm -r $(ZMQ_BUILD_DIR)
 	echo '$$(eval $$(call add-path,$(ZMQ_INSTALL_DIR)))' >> Makefile.paths
 
-HDF5_MAJ_MIN_VER := 1.12
-HDF5_VER := 1.12.1
+HDF5_MAJ_MIN_VER := 1.14
+HDF5_VER := 1.14.4
 HDF5_NAME_VER := hdf5-$(HDF5_VER)
 HDF5_BUILD_DIR := $(DEP_BUILD_DIR)/$(HDF5_NAME_VER)
 HDF5_INSTALL_DIR := $(DEP_INSTALL_DIR)/hdf5-install
--- a/arkouda-env.yml
+++ b/arkouda-env.yml
@@ -12,10 +12,10 @@ dependencies:
   - versioneer
   - matplotlib>=3.3.2
   - h5py>=3.7.0
-  - hdf5==1.12.2
+  - hdf5>=1.12.2
   - pip
   - types-tabulate
-  - pytables>=3.7.0
+  - pytables>=3.8.0
   - pyarrow
   - libiconv
   - libidn2
--- a/pydoc/requirements.txt
+++ b/pydoc/requirements.txt
@@ -9,10 +9,10 @@ pyfiglet
 versioneer
 matplotlib>=3.3.2
 h5py>=3.7.0
-hdf5==1.12.2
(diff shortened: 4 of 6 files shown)


ajpotts:
This lists the available versions of hdf5: https://www.hdfgroup.org/downloads/hdf5

ajpotts:
When I search for the python version of the `hdf5` version in conda:
```
$ conda search hdf5
Loading channels: done
# Name                       Version           Build  Channel             
hdf5                          1.8.18      h525d4c3_0  pkgs/main           
hdf5                          1.8.18      h6792536_1  pkgs/main           
hdf5                          1.8.20      hba1933b_1  pkgs/main           
hdf5                          1.10.1      h9caa474_1  pkgs/main           
hdf5                          1.10.1      hb0523eb_0  pkgs/main           
hdf5                          1.10.2      hba1933b_1  pkgs/main           
hdf5                          1.10.4      hb1b8bf9_0  pkgs/main           
hdf5                          1.10.6      h3ffc7dd_1  pkgs/main           
hdf5                          1.10.6      hb1b8bf9_0  pkgs/main           
hdf5                          1.10.7      hb1b8bf9_0  pkgs/main           
hdf5                          1.12.0      hc3cf35f_0  pkgs/main           
hdf5                          1.12.1      h2b7332f_3  pkgs/main           
hdf5                          1.12.1      h69dfa17_1  pkgs/main           
hdf5                          1.12.1      h70be1eb_2  pkgs/main 
```

However, everything seems to match if with the python version of `hdf5` library is `12.2.2`, so long as the version in the Makefile is `14.1.4`.

arkouda-env-dev.yml
```diff
@@ -12,10 +12,10 @@ dependencies:
   - versioneer
   - matplotlib>=3.3.2
   - h5py>=3.7.0
-  - hdf5==1.12.2
+  - hdf5>=1.12.2
```
stress-tess, line 15: I'm not sure if we should unpin hdf5 or pin to the new makefile version. My guess is this was done to ensure the `make deps` and conda env versions don't get out of step

stress-tess (commented):
If don't care about the conda env and `make deps` versions being out of sync this looks great!

I'm not really sure what's the right thing to do here. When I bumped up the arrow version, I kept it pinned to the same version as the makefile:
https://github.com/Bears-R-Us/arkouda/pull/2806/files

but then it seems I unpinned it fix some python 3.12 errors in this https://github.com/Bears-R-Us/arkouda/commit/58bf71f10c83d1d77dbb7fc42f2c91f5b39daa82

based on my commit message it seems like this was on my mind but I forgot to revisit it
> Unpinning tables and pyarrow seems to fix our dependency issues with the 3.12 check... I need to double check if the arrow version in the makefile has to match the one in yaml files

@bmcdonald3 u've been around for a while lol. do u remember any other reasons why these deps were forced to be in lock step? Or do u have any opinions on keeping the conda env version pinned or not

bmcdonald3:
We pinned HDF5 a long time ago because there were breaking changes in newer HDF5 versions. If we unpin HDF5, we may need to make some compatibility changes to make sure it continues to work, but it's also possible that it was just a single release that had a bug in it or something, my memory is a bit hazy, but I know there were changes from one release to the next that changed some HDF5 functions.

If it is working unpinned, I'd say that is better than pinning a specific version

stress-tess (approved):
This looks good to me! ben let me know the `make install-deps` versions are tested by the nightly so we'll still see any errors before they pop up

Issue #3383, opened 2024-07-02
function to return compiled array dimensions.


ajpotts:
**Feature Description**
A function on the chapel side that returns the value/values of `max_array_dims` that were used at compilation time.

**Example Behavior**
If `max_array_dims = 3`, then acceptable outputs would be: `proc maxArrayDims(){ return 3;}` or `proc allowedArrayDims(){return [0, 1, 2, 3];}`.

**Describe Workarounds**
Right now developers can read the json, but this may not be available to users, or could be altered after compile and give the incorrect result.

**Additional context**
N/A

jeremiah-corrado:
Good idea. I can include that in https://github.com/Bears-R-Us/arkouda/pull/3371 since I'm changing how the config file works there anyway.

PR #3521, opened 2024-07-18
Add function to retrieve server's maximum supported pdarray rank

jeremiah-corrado:
Adds a function `client.get_max_array_rank` to return the maximum array rank that the server was compiled to support.

resolves: https://github.com/Bears-R-Us/arkouda/issues/3383

### DIFF
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -257,6 +257,7 @@ module ServerConfig
             const byteorder: string;
             const autoShutdown: bool;
             const serverInfoNoSplash: bool;
+            const maxArrayDims: int;
         }
 
         var (Zmajor, Zminor, Zmicro) = ZMQ.version;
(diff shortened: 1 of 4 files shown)


src/ServerConfig.chpl
```diff
@@ -282,7 +283,8 @@ module ServerConfig
             regexMaxCaptures = regexMaxCaptures,
             byteorder = try! getByteorder(),
             autoShutdown = autoShutdown,
-            serverInfoNoSplash = serverInfoNoSplash
+            serverInfoNoSplash = serverInfoNoSplash,
+            maxArrayDims = MaxArrayDims
```
stress-tess, line 287: just a note for other reviewers, it wasn't immediately obvious where `MaxArrayDims` was originally defined but I found it as a config var on line 45. and if i'm to trust the comment about it, it should get set by one of the python scrips

stress-tess (approved):
looks good to me!

PR #3525, opened 2024-07-23
Soften `chapel-py` requirement for building server

jeremiah-corrado:
This PR re-includes the generated `Commands.chpl` file in the tree as a fallback when `make register-commands` fails due to difficulties installing the Chapel compiler's frontend bindings. 

With this change chapel-py is still required to build a non-default server (non-default modules, array dtypes, or multidimensional array support).

With some improvements to chapel-py's installation process, the file can be removed again.

### DIFF
--- a/registration-config.json
+++ b/registration-config.json
@@ -1,6 +1,5 @@
 {
     "parameter_classes": {
-        "NOTE!!!": "run 'make register-commands' after changing any of these values",
         "array": {
             "nd": [1],
             "dtype": [
--- a/src/registry/.gitignore
+++ b/src/registry/.gitignore
@@ -1,1 +1,0 @@
-Commands.chpl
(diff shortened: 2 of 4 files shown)


bmcdonald3 (approved):
Just to confirm, you did make sure this works in an environment without the Python bindings, right?

jeremiah-corrado:
I still need to double check that. Also are you familiar with these CI failures?

bmcdonald3:
Just looks like incorrect syntax in the changes to me:
```
/__w/arkouda/arkouda//src/registry/register_commands.bash: 22: Syntax error: "else" unexpected (expecting "then")
```

jeremiah-corrado:
Okay, tried building without chapel-py installed and it's working

bmcdonald3 (approved):

PR #3590, opened 2024-07-26
Refactor IndexingMsg

jeremiah-corrado:
Refactor IndexingMsg to use `registerCommand` and `instantiateAndRegister` annotations.

Also adds some new functionality to the `registerCommand` annotation:

1. support for homogenous tuple formal types

2. querying the domain and element type of an array argument to use in other parts of the signature are now supported. For example, these are both used in `intIndex`:

```chapel
@arkouda.registerCommand("[int]")
proc intIndex(const ref array: [?d] ?t, idx: d.rank*int): t {
    return array[idx];
}
```

3. If needed, procs can return a `SymEntry` instead of a plain array (this is useful for setting `max_bits` before returning — in most cases, it makes more sense to just return an array (`: [] t`)). For example, used in `sliceIndex`:

```chapel
@arkouda.registerCommand("[slice]")
proc sliceIndex(const ref array: [?d] ?t, starts: d.rank*int, stops: d.rank*int, strides: d.rank*int, max_bits: int): SymEntry(t, d.rank) throws {
  ...
   return new shared SymEntry(arraySlice, max_bits);
}
```


bmcdonald3 (approved):
The amount of code reduction here is crazy, nice work.

stress-tess (approved):
This looks great!! Thanks jeremiah!

PR #3594, opened 2024-07-29
Closes #3508: Update installation docs

### LINKED ISSUE #3508: Update recommended versions in install docs
Update recommended versions in install docs especially the chapel version now that we are only compatible with chpl 2.0 or higher

stress-tess:
This PR (closes #3508) updates the installation docs. This mostly just increases version numbers. To make sure the quickstart instructions worked properly, I downloaded a zip of the latest tag. I ran into some issue and in order to make it work, I needed to add some dependencies to the arkouda user yaml and add more instructions.

I also:
- made the quickstart link to the latest tag rather than needing us to manually update it each release
- provided a one liner for adding your conda environment to `Makefile.paths` that doesn't require the person installing to search for the right path. This will hopefully make the common case easier

Note to reviewers:
it appears as if I remove the dependencies for `sphinx-design` and `sphinx-autodoc-typehints`. But these are duplicated in the yaml file so I'm just removing the duplicate

### DIFF
--- a/arkouda-env-dev.yml
+++ b/arkouda-env-dev.yml
@@ -21,11 +21,11 @@ dependencies:
   - libidn2
   - jupyter
   - scipy<=1.13.1
+  - pytest>=6.0
+  - pytest-env
 
   # Developer dependencies
   - pexpect
-  - pytest>=6.0
-  - pytest-env
   - Sphinx>=5.1.1
   - sphinx-argparse
   - sphinx-autoapi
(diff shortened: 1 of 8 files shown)


bmcdonald3 (approved):

jaketrookman (dismissed):

jaketrookman (approved):
Looks good and they updated the gnu site and now correctly passes CI

PR #3596, opened 2024-07-29
Closes #3278: Add ziggurat method to standard normal

### LINKED ISSUE #3278: Add Ziggurat as a method for `standard_normal`
Right now our implementation of `standard_normal` does a [box-muller transform](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform), but numpy uses the [Ziggurat algorithm](https://en.wikipedia.org/wiki/Ziggurat_algorithm). Ziggurat algorithm's wikipedia page says:

> Nevertheless, the algorithm is computationally much faster than the two most commonly used methods of generating normally distributed random numbers, the [Marsaglia polar method](https://en.wikipedia.org/wiki/Marsaglia_polar_method) and the [Box–Muller transform](https://en.wikipedia.org/wiki/Box%E2%80%93Muller_transform), which require at least one logarithm and one square root calculation for each pair of generated values. However, since the ziggurat algorithm is more complex to implement it is best used when large quantities of random numbers are required.

considering it's used by numpy and we're likely to need to generate large quantities of numbers, I think it would be worthwhile. It's also used in numpy's implementation of other distributions such as `exponential` which will be added in #3183 

Numpy's c based implementation can be found [here](https://github.com/numpy/numpy/blob/a2d1972504493fe6ef00ec189924e0a0a0c9c006/numpy/random/src/distributions/distributions.c#L137)

This will require saving off [some constants ](https://github.com/numpy/numpy/blob/main/numpy/random/src/distributions/ziggurat_constants.h). The nature of chpl means we'll want these constants on each locale or else we'll kill performance by going back and forth to the lookup tables on locale 0. This has caused us issues previously (see https://github.com/Bears-R-Us/arkouda/pull/2354)

stress-tess:
This PR (closes #3278) adds `ziggurat` method for calculating standard normal based on [numpy's implemenation](https://github.com/numpy/numpy/blob/a2d1972504493fe6ef00ec189924e0a0a0c9c006/numpy/random/src/distributions/distributions.c#L137).

I also realized that even though the zig method for both `standard_normal` and `standard_exponential` can't handle multi dimensional arrays yet, their other methods can. I created two versions of each proc with `where array_nd > 1` and `where array_nd == 1`, so now we can generate multi-dim exponentially and normally distributed arrays if we use the compatible method. Using the incompatible method fails gracefully. I also added parameterization for the methods in the tests


src/RandMsg.chpl
```diff
@@ -157,32 +157,167 @@ module RandMsg
...
+        var count = 0;
+        // to guarantee completion this should be while true, but limiting to 100 tries will make the chance
+        // of failure near zero while avoiding the possibility of an infinite loop
+        // the odds of failing 100 times in a row is (1 - .993)**100 = 3.2345e-216
+        while count <= 100 {
+            var ri = uintRng.next();
+
+            // AND with 0xFF (255) to get our index into our 256 long const arrays
+            var idx = ri & 0xFF;
+            idx = ri & 0xff;
```
ajpotts, line 189: It seems like you define `idx` and then immediately reset it.
```
var idx = ri & 0xFF;
idx = ri & 0xff;
```

ajpotts (approved):
Looks good!

src/RandMsg.chpl
stress-tess: ah yes that was not intentional... lol good catch!!

PR #3605, opened 2024-08-01
Closes #3372: Add logistic to random number generators

### LINKED ISSUE #3372: add `logistic` to random number generators
add logistic to random number generators

Relevant docs:
[numpy.random.Generator.logistic](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.logistic.html)

stress-tess:
This PR (closes #3372) adds logistic to generators. This doesn't add multi-dim bc it uses `uniformStreamPerElem`

### DIFF
--- a/arkouda/random/_generator.py
+++ b/arkouda/random/_generator.py
@@ -220,6 +220,59 @@ class Generator:
         self._state += full_size if method.upper() == "INV" else 1
         return create_pdarray(rep_msg)
 
+    def logistic(self, loc=0.0, scale=1.0, size=None):
+        if size is None:
+            # delegate to numpy when return size is 1
+            return self._np_generator.logistic(loc=loc, scale=scale, size=size)
+
+        mu = loc
+        if _val_isinstance_of_union(mu, numeric_scalars):
+            is_single_mu = True
+            if not _val_isinstance_of_union(mu, float_scalars):
+                mu = float(mu)
+        elif isinstance(mu, pdarray):
+            is_single_mu = False
+            if size != mu.size:
+                raise TypeError("array of locs must have same size as return size")
+            if mu.dtype != akfloat64:
+                from arkouda.numeric import cast as akcast
+
+                mu = akcast(mu, akfloat64)
+        else:
+            raise TypeError("logistic only accepts a pdarray or float scalar for loc")
+
+        if _val_isinstance_of_union(scale, numeric_scalars):
+            is_single_scale = True
+            if not _val_isinstance_of_union(scale, float_scalars):
+                scale = float(scale)
+        elif isinstance(scale, pdarray):
+            is_single_scale = False
+            if size != scale.size:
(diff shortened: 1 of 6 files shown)


arkouda/random/_generator.py
```diff
@@ -213,6 +213,59 @@ def standard_exponential(self, size=None, method="zig"):
...
+            if not _val_isinstance_of_union(scale, float_scalars):
+                scale = float(scale)
+        elif isinstance(scale, pdarray):
+            is_single_scale = False
+            if size != scale.size:
+                raise TypeError("array of scales must have same size as return size")
+            if scale.dtype != akfloat64:
+                from arkouda.numeric import cast as akcast
+
+                scale = akcast(scale, akfloat64)
```
drculhane, line 248: Two observations and/or comments: (1) shouldn't loc just be called mu in the first place? (2) since the blocks of code that process mu and scale are identical, this would be a good candidate for moving that code into a function, and calling it for both mu and scale.  This is either (a) a very good comment, or (b) just a case of two people having different programming styles.

src/RandMsg.chpl
```diff
@@ -402,6 +402,38 @@ module RandMsg
...
+        if state != 1 then rng.skipTo(state-1);
+
+        var logisticArr = makeDistArray(size, real);
+        const mu = new scalarOrArray(muStr, !isSingleMu, st),
+              scale = new scalarOrArray(scaleStr, !isSingleScale, st);
+
+        uniformStreamPerElem(logisticArr, rng, GenerationFunction.LogisticGenerator, hasSeed, mu=mu, scale=scale);
+        return st.insert(createSymEntry(logisticArr));
+    }
+
```
drculhane, line 436: The code is solid.  I'd love a tutorial on logistic random number generation at one of our aperiodic team meetings.

drculhane (approved):

arkouda/random/_generator.py
stress-tess: This is a good comment! 

(1) I kept it `loc` in the first place because that's what the argument is called in the numpy version of `logistic`. The reason I don't keep it `loc` is because I was hitting name collisions in the chapel code with  variable for a locale. I figured `mu` was a good alternative since that's the symbol that loc represents in the formula

(2) I agree I think pulling this out as a helper is a good idea. It's definitely going to come up again when I add another distribution so I'll make that change then

src/RandMsg.chpl
stress-tess: Thanks!! and absolutely! :)

stress-tess:
still needs a docstring and I'll go ahead and add the helper method for scalar or array

stress-tess:
@drculhane do you mind rereviewing? I added tests, docs, and the helper we talked about

jaketrookman (approved):

drculhane (approved):

drculhane:
Looks good!

PR #3615, opened 2024-08-05
Revised draft3422

drculhane:
TDLR: although not everything mentioning ArrayViews has been deleted, I think it's now impossible to create one.  But there's some challenging territory in here, so I'm not ready to make this more than a Draft at this point.

A bit more detail:

- pdarraycreation now supports multidimensionality, and does not return ArrayViews.
- in pdarrayclass, the reshape method has been modified to return pdarrays, not ArrayViews.

- CREDIT TO TESS => all of the random generators now return multidimensional arrays, per PR 3524

So what about eliminating the ArrayView class?

- io routines, both python and chapel, have many references to ArrayView, however ...
- afaik, the only place that actually makes them is the arrayView_readhdfMsg function in HDF5Msg.chpl.
- I altered its return statement so that the calling function will interpret the response as a pdarray, rather than an ArrayView.  I am optimistic that it's that simple, but am not sure.  I need to study the io_tests more before I'm confident.

- worth mentioning: the io routines are still full of functions that will write an ArrayView to a file, if they're given one.  I'm not concentrating on those for now, since they don't actually create ArrayViews.  When we're sure none are being created, it should be simple to delete these functions.

- there were tests in PROTO-tests/tests that specifically exercised the ArrayView class; I've rewritten some of those, and deleted the array_view_test.py file.
- there were other tests that enforced a hardcoded max array rank of 1.  I've rewritten those, too.

Finally, I'm having some inconsistent results when running pytest.  That's why one of the coargsort tests is skipped.

Before I'll be ready to turn this into a PR, I need to push out this draft for your review and comments.  I'm also going to dive back into the many items that are still unfinished.


drculhane:
Ah, I see that array_view_test.py was already deprecated an hour ago :)

ajpotts:
No longer necessary as changes were covered by:

- https://github.com/Bears-R-Us/arkouda/pull/3715
- https://github.com/Bears-R-Us/arkouda/issues/3702
- https://github.com/Bears-R-Us/arkouda/pull/3701
- https://github.com/Bears-R-Us/arkouda/pull/3633

PR #3633, opened 2024-08-07
Closes #3632: refactor LinalgMsg to remove registerND annotation

### LINKED ISSUE #3632: refactor LinalgMsg to remove registerND annotation
refactor LinalgMsg to remove registerND annotation.

ajpotts:
Refactors LinalgMsg to remove registerND annotation and replace with @arkouda.instantiateAndRegister.

@drculhane did all the work on this one, I'm just helping to submit the PR.

Closes #3632: refactor LinalgMsg to remove registerND annotation

### DIFF
--- a/arkouda/numeric.py
+++ b/arkouda/numeric.py
@@ -10,6 +10,7 @@ from typeguard import typechecked
 from arkouda.client import generic_msg
 from arkouda.dtypes import DTypes, bigint
 from arkouda.dtypes import dtype as akdtype
+from arkouda.dtypes import int64 as akint64
 from arkouda.dtypes import (
     int_scalars,
     isSupportedNumber,
@@ -34,6 +35,7 @@ __all__ = [
     "ceil",
     "clip",
     "count_nonzero",
+    "eye",
     "floor",
     "trunc",
     "round",
@@ -48,6 +50,11 @@ __all__ = [
     "exp",
     "expm1",
     "square",
+    "matmul",
+    "triu",
+    "tril",
+    "transpose",
+    "vecdot",
     "cumsum",
     "cumprod",
     "sin",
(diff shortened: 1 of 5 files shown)


drculhane:
(1) You made excellent improvements; thanks.
(2) I'm having unusual problems making and testing this.  I'm assuming that once again, it's most likely my VM.  I'll probably post another comment with a summary later today.

drculhane:
Okay, here's my follow-up.

(1) Had to increase my VM RAM to 32GB to make with max-array-rank of 3.  My RevisedDraft3422 would make in 24GB (but not 16).  I have no explanation for this.

(2) Some tests hung with max-array-rank < 3.  With 3, all pass (or skip) except for two (in client_test.py): test_client_get_config and test_client_array_dim_cmd_error , which are hardcoded to expect max-array-rank of 1.

Those client_test.py tests are fixed in RevisedDraft3422, but when Amanda suggested splitting it into smaller PRs for evaluation, it slipped my mind that client_test.py needed to be part of the first one (and I'm not sure that this belongs in client_test anyway, but I don't have strong feelings about it).

ajpotts:
I did end up making some minor changes to `LinalgMsg` to streamline code a little.  I also changed one of the tests in `client_test.py`.  Let me know what you think.

src/LinalgMsg.chpl
```diff
@@ -307,60 +305,53 @@ module LinalgMsg {
...
-            const errorMsg = "Matrix transpose with arrays of dimension < 2 is not supported";
-            linalgLogger.error(getModuleName(),getRoutineName(),getLineNumber(),errorMsg);
-            return new MsgTuple(errorMsg, MsgType.ERROR);
-        }
+    // Transpose an array.
 
-        const name = msgArgs.getValueOf("array"),
-              rname = st.nextName();
+    @arkouda.instantiateAndRegister
+    proc transpose(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, type array_dtype, param array_nd: int): MsgTuple throws
```
jeremiah-corrado, line 318: Some of these could use the `registerCommand` annotation instead of `instantiateAndRegister` to simplify things further. I don't think it's necessary to change what you have here, just pointing it out for future reference.

```chapel
@arkouda.registerCommand
proc transpose(array: [?d] ?t): [] t
  where d.rank >= 2
{
  var outShape = array.shape;
  outShape[outShape.size-2] <=> outShape[outShape.size-1];
  var ret = makeDistArray(outShape, t);
  doTranspose(array, ret);
  return ret;
}
```

The argument parsing and symbol creation code will be generated automatically.

src/LinalgMsg.chpl
```diff
@@ -181,71 +160,89 @@ module LinalgMsg {
...
-        }
+        //  The select block below (and in vecdot) implements the rule:
+        //     if either input is real, the output is real
+        //     else if either input is integer, the output is integer
+        //     else if either input is uint(8), the output is uint(8)
+        //     else if both inputs are bool, the output is bool
+        //     else error condition.
 
         select (x1G.dtype, x2G.dtype) {
+            // real cases
```
jeremiah-corrado, line 188: The type logic could be lifted into the signature here:

```chapel
@arkouda.instantiateAndRegister
proc matmul(cmd: ..., type array_dtype_x1, type array_dtype_x2, param array_nd: int): MsgTuple throws {
  const x1 = st[msgArgs["x1"]]: borrowed SymEntry(array_dtype_x1, array_nd);
  const x2 = st[msgArgs["x2"]]: borrowed SymEntry(array_dtype_x2, array_nd);

  type result_type = compute_result_type(array_dtype_x1, array_dtype_x2);

  ...

}

proc compute_result_type(type t1, type t2) type {
  if t1 == real || t2 == real then return real;
  if t1 == int || t2 == int then return int;
  if t1 == uin(8) || t2 == uint(8) then return uint(8);
  return bool;
}

**Review by jeremiah-corrado — APPROVED** [review]
looks good to me! I have a couple of (very optional) suggestions on how to further simplify the server code:

File: src/LinalgMsg.chpl (hunk 2 above)
Comment (stress-tess) [reply src/LinalgMsg.chpl:188]: hey @jeremiah-corrado, I'm struggling a bit to create a function that returns a type in chpl. I did this off to the side:
```chapel
proc myHelper(retFloat:bool) {
    if retFloat {
        return real;
    }
    else {
        return int;
    }
}

type t = myHelper(true);
```
and got a compiler error `error: illegal return of type where value is expected`

jeremiah-corrado: For that proc to work, the argument needs to be a `param` s.t. the if-block can be evaluated at compile time, and it needs a type "return intent":
```chapel
proc myHelper(param retFloat: bool) type { ... }
```

stress-tess: ah! I tried each of these individually but never combined them

ajpotts: This is a great suggestion.  I  made the change.

src/LinalgMsg.chpl
ajpotts: I couldn't get this to work easily so I made a separate ticket: https://github.com/Bears-R-Us/arkouda/issues/3656

arkouda/numeric.py
```diff
@@ -2248,3 +2255,305 @@ def putmask(pda: pdarray, mask: Union[bool, pdarray], values: pdarray):
         result = result[:-(reduction)]
 
     pda[:] = where(mask, result, pda)  # pda[:] = allows us to return modified value
+
+
+# after this is working, add comments and docstring
```
drculhane, line 2260: Oops.  Line 2260 is one of those infamous "comments I forgot to remove."

drculhane (approved):
There's one of my "oops, I should have removed that" comments at line 2260 in numeric.py, but if that makes it through this PR, we'll take it out in the next one.

arkouda/numeric.py
drculhane, line 2260:
```suggestion
```

ajpotts: Removed it.

src/LinalgMsg.chpl
```diff
@@ -392,9 +351,41 @@ module LinalgMsg {
...
+        // not allowed.
+
+        select (x1G.dtype, x2G.dtype) {
+        // real cases
+            when (DType.Float64, DType.Int64)   do return doVecdot(real,    int,     real);
+            when (DType.Float64, DType.UInt8)   do return doVecdot(real,    uint(8), real);
+            when (DType.Float64, DType.Float64) do return doVecdot(real,    real,    real);
+            when (DType.Float64, DType.Bool)    do return doVecdot(real,    bool,    real);
+            when (DType.Int64, DType.Float64)   do return doVecdot(int,     real,    real);
+            when (DType.UInt8, DType.Float64)   do return doVecdot(uint(8), real,    real);
```
stress-tess, line 366: we don't have to make this change, but is this the same use case for `compute_result_type` as matrix multiplication

stress-tess (approved):
A few small suggestions, but I don't see any logic issues!

PROTO_tests/tests/numeric_test.py
```diff
@@ -828,3 +832,169 @@ def test_putmask(self, prob_size):
...
+                ppa = ak.tril(pda, diag).to_ndarray()
+                assert check(npa, ppa, data_type)
+
+    # triu works on ints, floats, or bool
+
+    @pytest.mark.parametrize("data_type", INT_FLOAT_BOOL)
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_triu(self, data_type, prob_size):
+        if get_max_array_rank() < 2:
+            pytest.skip()
```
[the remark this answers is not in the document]
ajpotts: The reason I changed it to this is that `get_max_array_rank()` cannot be called before the test because the arkouda server needs to connect first.  `get_server_max_array_dims()` depends on the config file which is not as accurate.

PROTO_tests/tests/numeric_test.py
```diff
@@ -828,3 +832,169 @@ def test_putmask(self, prob_size):
...
+
+        size = int(sqrt(prob_size))
+
+        # ints and bools are checked for equality; floats are checked for closeness
+
+        check = lambda a, b, t: (
+            np.allclose(a.tolist(), b.tolist()) if akdtype(t) == "float64" else (a == b).all()
+        )
```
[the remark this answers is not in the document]
drculhane: That was a new one to me, but yes, it looks like it is much the same as assert_almost_equal.

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
drculhane: Good suggestion, definitely.

PROTO_tests/tests/numeric_test.py
```diff
@@ -828,3 +832,169 @@ def test_putmask(self, prob_size):
...
+
+        # test on one square and two non-square products
+
+        for rows, cols in [(size, size), (size + 1, size - 1), (size - 1, size + 1)]:
+            pdaLeft = ak.randint(0, 10, (rows, size), dtype=data_type1)
+            ndaLeft = pdaLeft.to_ndarray()
+            pdaRight = ak.randint(0, 10, (size, cols), dtype=data_type2)
+            ndaRight = pdaRight.to_ndarray()
+            akProduct = ak.matmul(pdaLeft, pdaRight)
+            print(akProduct.dtype, "******************************************")
```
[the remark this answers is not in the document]
drculhane: Yeah, sigh.  I do try to catch these.

PROTO_tests/tests/numeric_test.py
```diff
@@ -828,3 +832,169 @@ def test_putmask(self, prob_size):
...
+        check = lambda a, b, t: (
+            np.allclose(a.tolist(), b.tolist()) if akdtype(t) == "float64" else (a == b).all()
+        )
+
+        pda_a = ak.randint(0, 10, (depth, width), dtype=data_type1)
+        nda_a = pda_a.to_ndarray()
+        pda_b = ak.randint(0, 10, (depth, width), dtype=data_type2)
+        nda_b = pda_b.to_ndarray()
+        akProduct = ak.vecdot(pda_a, pda_b)
+        print(akProduct.dtype, "******************************************")
```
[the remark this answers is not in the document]
drculhane: Same same.

src/LinalgMsg.chpl
drculhane: It's almost identical, except that vecdot doesn't allow both inputs to be bool.

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
stress-tess: @ajpotts ah! well in that case, should we change other spots that use this `skipif` to do the same `get_max_array_rank` check? either way that would be outside the scope of this PR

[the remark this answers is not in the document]
ajpotts: I made a separate ticket for this question: https://github.com/Bears-R-Us/arkouda/issues/3664

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
ajpotts: This is a great idea but I need to update the testing library to support multi-dimensional arrays.  I made a separate ticket: https://github.com/Bears-R-Us/arkouda/issues/3665

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
ajpotts: This should be fixed now.

PROTO_tests/tests/numeric_test.py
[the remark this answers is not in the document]
ajpotts: This should be fixed now.

src/LinalgMsg.chpl
ajpotts: OK, I refactored this one too.

stress-tess (approved):

jaketrookman (approved):

PR #3634, opened 2024-08-07
Fix [slice] indexing bug on multi-locale builds

jeremiah-corrado:
Adjustments to indexingMsg in https://github.com/Bears-R-Us/arkouda/pull/3590 caused non-distributed domains to be used in distributed indexing expressions. This caused some index-out-of-bounds errors.

The errors were caused by a [Chapel bug](https://github.com/chapel-lang/chapel/issues/25727), which this PR works around by making the slicing domains distributed.


bmcdonald3 (approved):
Looks good to me, cool you found a way to do that without reverting to the bulk assignment.

stress-tess (approved):

PR #3635, opened 2024-08-07
Closes #3049: Remove mypy version upper bound

### LINKED ISSUE #3049: Update supported mypy versions
We are seeing failures with mypy versions in the supported range specifically ` 0.982` when the range is `mypy>=0.931,<0.990`

We should :
- figure out what the new minimum supported mypy version is
- go through and address the mypy errors so we can remove the upper limit

stress-tess:
This PR (closes https://github.com/Bears-R-Us/arkouda/issues/3049) removes the upper bound of our `mypy` version since the latest works just fine. Also related to https://github.com/Bears-R-Us/arkouda/issues/3346, I changed the python restriction to be `!=3.12.4` since `3.12.3` works just fine. It seems version `3.12.5` was just released, so we'll have to see if the forward refs are still an issue

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -69,7 +69,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v5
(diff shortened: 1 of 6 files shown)


jaketrookman (approved):
Looks good

PR #3636, opened 2024-08-07
Fixes #3627: multilocale choice failures

### LINKED ISSUE #3627: Multidim random causing failures with multilocale
PR #3524 is causing failures in the nightly. It seems my conversion over to use the new framework makes it so even 1 dimensional random generation is causing failures with multi-locale

stress-tess:
This PR (fixes #3627) I leftover some logic that I used to call `orderToIndex` when I was tryig to make `choice` work for multi-dimensional arrays. This was causing an assertion error. Since `choice` doesn't support multidim yet, I just switched back to looping over the array directly


jaketrookman (approved):

bmcdonald3 (approved):
My hero

PR #3647, opened 2024-08-12
Closes #3646: Config changes for `make test-proto`

### LINKED ISSUE #3646: Config changes for `make test-proto`
I was digging into the proto tests to see how we fared with a multi-locale server with runtime checks enabled. While investigating this I found a few small errors and some quality of life changes

stress-tess:
This PR closes #3646. I was digging into the proto tests to see how we fared with a multi-locale server with runtime checks enabled. While investigating this, I found a few small errors and some quality of life changes. These came up because I wanted to use the client running mode to bring my own server (one built with multi-locale and runtime checks) and attempting to skip tests that contained failures to see if the subsequent files had any issues

The changes made:
* `make test-proto` command included `PROTO_tests/` in addition to `pytest_PROTO.ini `. This caused test files to always be run even if they were removed from the `.ini` file
* The environment variables set in `pytest_PROTO.ini ` were being picked up when you run the tests in pycharm. It seems pycharm looks for a file named `pytest.ini` for it's configuration, so I renamed our `ini` to that since we're making the proto tests our main/only testing suite
* There's one test in `client_test`, which shuts down the server and calls `start_arkouda_server` to restart it, but this doesn't work when running mode is client. So I added `pytest.test_running_mode` and created a `skipif` for this test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -557,7 +557,7 @@ print-%:
 
 size=100
 test-python-proto:
-	python3 -m pytest -c pytest_PROTO.ini PROTO_tests/ --size=$(size) $(ARKOUDA_PYTEST_OPTIONS)
+	python3 -m pytest -c pytest.ini --size=$(size) $(ARKOUDA_PYTEST_OPTIONS)
 
 CLEAN_TARGETS += test-clean
 .PHONY: test-clean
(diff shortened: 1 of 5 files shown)


jaketrookman (approved):
Looks good

PR #3651, opened 2024-08-12
Fixes #3648: Fix empty bounding box error

### LINKED ISSUE #2648: `logger_test.py` Conversion for new test framework
Convert `logger_test.py` to work in the new test framework. The existing file should be maintained since we will be doing a complete swap once all tests are converted. Only the existing tests need to be converted, there should not be any additional testing that needs to be added.

### LINKED ISSUE #3648: multi-locale failure `blockDist() requires a non-empty boundingBox`
A couple of our tests were failing with the following error message:
```
$CHPL_HOME/modules/dists/BlockDist.chpl:673: error: halt reached - blockDist() requires a non-empty boundingBox
```

one reproducer with a 2 locale server is:
```
ARKOUDA_RUNNING_MODE=CLIENT python3 -m pytest PROTO_tests/tests/categorical_test.py::TestCategorical::test_sort
```

stress-tess:
This PR (closes #3648) fixes empty bounding box error. It seems this little bit got lost in #3634, so this is just adding it back in

### DIFF
--- a/src/SymArrayDmap.chpl
+++ b/src/SymArrayDmap.chpl
@@ -31,7 +31,13 @@ module SymArrayDmap {
           return dom;
         }
         when Dmap.blockDist {
-          return blockDist.createDomain(dom);
+          if dom.size > 0 {
+              return blockDist.createDomain(dom);
+          }
+          // fix the annoyance about boundingBox being empty
+          else {
+            return dom dmapped new blockDist(boundingBox=dom.expand(1));
+          }
         }
       }
     }


jeremiah-corrado (approved):
My bad, thanks for fixing this! Out of curiosity, what test causes the error to come up?

stress-tess:
> My bad, thanks for fixing this! Out of curiosity, what test causes the error to come up?

quite a few of them actually but the one I was using a reproducer was `PROTO_tests/tests/categorical_test.py::TestCategorical::test_sort`

Also I'm really glad you asked because it made realized I'd linked the wrong issue against this PR 🤦‍♀️

jaketrookman (approved):

PR #3653, opened 2024-08-12
Fixes #3650: incorrect parquet reads with larger sizes

### LINKED ISSUE #3650: incorrect parquet reads with larger sizes
Our parquet tests pass when running `make test-proto` but that changes the problem size from `10**8`, the default, to `100`

Reproducer:
```
python3 -m pytest PROTO_tests/tests/io_test.py::TestParquet
```

stress-tess:
This PR fixes #3650 which caused incorrect results when reading a large array using parquet

The issue was caused by moving the `readColumn` logic, which relies on `i` and updates `i`, into a function. The updates made to `i` passed into the function never made their way to `i` from the outer scope. This function is called inside a for loop over `num_row_groups`, so if there's more than one row group, the value of `i` gets reset. I resolved this by having the helpers return the updated value of `i`

### DIFF
--- a/src/parquet/ReadParquet.cpp
+++ b/src/parquet/ReadParquet.cpp
@@ -2,7 +2,7 @@
 #include "UtilParquet.h"
 
 template <typename ReaderType, typename ChplType>
-void readColumn(void* chpl_arr, int startIdx, std::shared_ptr<parquet::ColumnReader> column_reader,
+int64_t readColumn(void* chpl_arr, int64_t startIdx, std::shared_ptr<parquet::ColumnReader> column_reader,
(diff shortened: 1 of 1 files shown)


bmcdonald3 (approved):
This is great, thanks for tracking this down. I think you might also be able to make it a pass by reference, rather than having it return the value, but this works for me and is probably less confusing. My mistake on introducing this one!

jaketrookman (approved):

stress-tess:
> This is great, thanks for tracking this down. I think you might also be able to make it a pass by reference, rather than having it return the value, but this works for me and is probably less confusing. My mistake on introducing this one!

yah! that crossed my mind and I was like the less I need to think about pointers / references the better lol

PR #3659, opened 2024-08-13
Fixes #3654: Out of bounds bug writing float segarray

### LINKED ISSUE #3654: index OOB for float segarray in `io_test`
Running `io_test.py` with multilocale and runtime checks, I found an out of bounds error
Error:
```
arkouda//src/ParquetMsg.chpl:1712: error: halt reached - array index out of bounds
note: index was 0 but array bounds are 0..-1
```
Based on the line number it seems to be for float segarrays. It seems to be indexing into an empty domain so hopefully it won't be too awful hard to fix

Reproducer:
```
ARKOUDA_RUNNING_MODE=CLIENT python3 -m pytest PROTO_tests/tests/io_test.py --size=100
```

stress-tess:
This PR (fixes #3654) bug with writing float segarray containing empty segments and nans with parquet

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -1683,7 +1683,9 @@ module ParquetMsg {
                   var valIdxRange = startValIdx..endValIdx;
                   ref olda = values.a;
                   int_vals[ui..#valIdxRange.size] = olda[valIdxRange];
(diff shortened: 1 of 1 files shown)


bmcdonald3 (approved):

jaketrookman (approved):
Looks good

PR #3359, opened 2024-06-25
Refactor SymbolTable and error handling

jeremiah-corrado:
This PR modifies the symbol table interface to move some boilerplate code out of message handlers:

* adds `this(name: string)` access of symbol table
* adds `insert` method, which generates and returns a message response with the typical `"created" + st.attrib(rname)` information
* folds some error-handling into `executeCommand` to simplify the serverDaemon loop

PR #3671, opened 2024-08-15
Array transfer perf fix

jeremiah-corrado:
Fix recent [regression](https://chapel-lang.org/perf/arkouda/16-node-xc/?startdate=2024/02/19&enddate=2024/08/13&graphs=arraytransferperformance,bigintarraytransferperformance) in array transfer performance. The regression was caused by:
* an `in`-intent in `sendRepMsg` (added [here](https://github.com/Bears-R-Us/arkouda/pull/3345/files#diff-70e824ad7029c3097d23ee72e6d9ffc1208500254319bef0045bab0e7125b9fd)) leading to an extra copy of the `bytes` payload. Fixed by not referencing `repMsg` after it gets handed off to `sendRepMsg`.
* error handling that got moved into `executeCommand` [here](https://github.com/Bears-R-Us/arkouda/pull/3359/files#diff-f76841cd12e3d075d3dd1851ac65b14a688efd74ce116190ea6e81fc7a4306df). Fixed by moving error handling back out into serverDaemon.

Also converts "tondarray" and "array" commands to use instantiateAndRegister annotation.


PROTO_tests/tests/client_test.py
stress-tess, line 1: was this file meant to be committed?

stress-tess (commented):
looks good! My only question is whether the proto test client file was committed on purpose. If so, we actually moved the tests to the `tests` directory, so if you could updated that version of `client_test` instead that would be great

stress-tess (approved):

PR #3680, opened 2024-08-19
Closes #3679: Flatten and unflatten runtime checks failure

### LINKED ISSUE #3679: `flatten` and `unflatten` runtime checks failure
when building with with gasnet, runtime checks, and multi-dim, I was seeing some test failures with the following error 

```
arkouda//src/AryUtil.chpl:966: error: references to remote data cannot be passed to external routines like 'c_pointer_return'
```

reproducer:
```
ARKOUDA_RUNNING_MODE=CLIENT python3 -m pytest tests/array_api/array_manipulation.py
```

stress-tess:
when building with with gasnet, runtime checks, and multi-dim, I was seeing some test failures with the following error:

```
arkouda//src/AryUtil.chpl:966: error: references to remote data cannot be passed to external routines like 'c_pointer_return'
```

This was caused by some `c_ptrTo` called on some remote values. This was definitely what was intended because we were passing the addresses to `get/put`. I looked into how this was handled in the aggregation code and found `getAddr` in `CommPrimitives`
https://github.com/Bears-R-Us/arkouda/blob/c452ea3f9375f0d40a44fd60cfffddae7ed995b8/src/CommPrimitives.chpl#L7-L10

I think this does the same thing as `c_ptrTo` but avoids the runtime check? I'm not sure but making the substitution seems to have fixed the issue

### DIFF
--- a/src/AryUtil.chpl
+++ b/src/AryUtil.chpl
@@ -14,6 +14,8 @@ module AryUtil
     use OS.POSIX;
     use List;
     use CommAggregation;
+    use CommPrimitives;
+
 
     param bitsPerDigit = RSLSD_bitsPerDigit;
     private param numBuckets = 1 << bitsPerDigit; // these need to be const for comms/performance reasons
@@ -902,7 +904,7 @@ module AryUtil
             // flat region sits within a single locale, do a single get
             get(
               c_ptrTo(unflat[low]),
-              c_ptrToConst(a[flatSlice.low]):c_ptr(t),
+              getAddr(a[flatSlice.low]),
               locInStart,
               c_sizeof(t) * flatSlice.size
             );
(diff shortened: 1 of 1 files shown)


ajpotts (approved):

jaketrookman (approved):

jeremiah-corrado (approved):
Looks good!

> I think this does the same thing as c_ptrTo but avoids the runtime check?

I believe `c_ptrTo` and `c_ptrToConst` are only designed to work for local memory, but `__primitive("_wide_get_addr",...` can get the address for a remote memory region (which I probably didn't know when I wrote this). So I'm kinda surprised that it worked for me when I originally tested it. Thanks for fixing this!

PR #3690, opened 2024-08-21
Fixes #3676 - index order bug in `nonzero`

### LINKED ISSUE #3676: nonzero in ReductionMsg to preserve order
nonzero in ReductionMsg to preserve order

To match numpy,
https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html
the order of the nonzero indices returned should be row-major, C-style order.  However, when running `test_nonzero` locally,
```
print(nz)
(Arkouda Array ((4,), int64)[0 1 2 3], Arkouda Array ((4,), int64)[1 2 2 2], Arkouda Array ((4,), int64)[0 3 2 1])
```
whereas in the CI
```
print(nz)
(Arkouda Array ((4,), int64)[0 2 3 1], Arkouda Array ((4,), int64)[1 2 2 2], Arkouda Array ((4,), int64)[0 2 1 3])
```

jeremiah-corrado:
Fix `nonzero` s.t. it maintains row-major order in the output regardless of the number of tasks used. #3676  pointed out that it was returning indices in an incorrect order when the number of tasks is much smaller than the array size (e.g., in CI testing with 2 cpu cores, or any real-world use case with large arrays).

Also converts `nonZeroMsg` to use `instantiateAndRegister`.

resolves #3676


stress-tess (approved):
looks great! Thanks @jeremiah-corrado!

jeremiah-corrado:
Thanks, I realized my change did not fix the problem for multi-locale though. I just got something working but it's still draft code, so this isn't quite ready to merge yet FYI.

jeremiah-corrado:
Ready for review/merge again @stress-tess.

stress-tess (approved):

PR #3693, opened 2024-08-26
Closes #3656: Refactor LinalgMsg.transpose to use registerCommand

### LINKED ISSUE #3656: Refactor LinalgMsg.transpose to use registerCommand
Refactor LinalgMsg.transpose to use registerCommand

@jeremiah-corrado pointed out that this might be more efficient:

```
@arkouda.registerCommand
proc transpose(array: [?d] ?t): [] t
  where d.rank >= 2
{
  var outShape = array.shape;
  outShape[outShape.size-2] <=> outShape[outShape.size-1];
  var ret = makeDistArray(outShape, t);
  doTranspose(array, ret);
  return ret;
}
```

ajpotts:
Refactors `LinalgMsg.transpose` to use `@registerCommand` which is slightly more efficient.

Closes #3656: Refactor LinalgMsg.transpose to use registerCommand


src/LinalgMsg.chpl
```diff
@@ -285,39 +285,23 @@ module LinalgMsg {
...
-        return MsgTuple.error("Matrix transpose with arrays of dimension < 2 is not supported");
+    proc transpose(array: [?d] ?t): [d] t throws
+    where d.rank < 2 {
+      throw new Error("Matrix transpose with arrays of dimension < 2 is not supported");
     }
 
-    // TODO: performance improvements. Should use tiling to keep data local
-
+    // // TODO: performance improvements. Should use tiling to keep data local
+
```
jeremiah-corrado, line 304: the body of `doTranspose` could be moved into `transpose` imo (assuming it's not also being used somewhere else in this module)

jeremiah-corrado (approved):
Looks good! One minor suggestion:

jaketrookman (approved):

src/LinalgMsg.chpl
ajpotts: Good idea.  I made the change.  Thanks!

stress-tess (approved):

PR #3701, opened 2024-08-28
Closes #3699 zeros, ones, full to return Array

### LINKED ISSUE #3699: zeros, ones, full to return Array
The functions `zeros`, `ones`, `full` in `pdarrayclass` to return Array type when the dimension is greater than one.

ajpotts:
This changes the functions zeros, ones, and full in pdarrayclass to return Array type in the multi dimensional case.  It also adds some extra unit tests.  

Closes #3699 zeros, ones, full to return Array

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -26,7 +26,7 @@ from arkouda.numpy.dtypes import (
 from arkouda.numpy.dtypes import uint64 as akuint64
 from arkouda.pdarrayclass import create_pdarray, pdarray
 from arkouda.strings import Strings
-from arkouda.util import _infer_shape_from_size
+
 
 __all__ = [
     "array",
@@ -283,6 +283,7 @@ def array(
         except TypeError:
             raise RuntimeError(f"Unhandled dtype {a.dtype}")
     else:
+        from arkouda.util import _infer_shape_from_size
         shape, ndim, full_size = _infer_shape_from_size(a.shape)
 
         # Do not allow arrays that are too large
@@ -476,6 +477,7 @@ def zeros(
     # check dtype for error
     if dtype_name not in NumericDTypes:
         raise TypeError(f"unsupported dtype {dtype}")
+    from arkouda.util import _infer_shape_from_size
     shape, ndim, full_size = _infer_shape_from_size(size)
 
     if ndim > get_max_array_rank():
@@ -535,6 +537,7 @@ def ones(
     # check dtype for error
     if dtype_name not in NumericDTypes:
         raise TypeError(f"unsupported dtype {dtype}")
+    from arkouda.util import _infer_shape_from_size
     shape, ndim, full_size = _infer_shape_from_size(size)
 
     if ndim > get_max_array_rank():
@@ -603,6 +606,7 @@ def full(
     # check dtype for error
     if dtype_name not in NumericDTypes:
         raise TypeError(f"unsupported dtype {dtype}")
+    from arkouda.util import _infer_shape_from_size
     shape, ndim, full_size = _infer_shape_from_size(size)
 
     if ndim > get_max_array_rank():
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -1,13 +1,10 @@
-from __future__ import annotations
-
 import itertools
 from typing import Any, Iterable, List, Optional, Tuple, Union, cast
 
 import numpy as np
 import pandas as pd
 from typeguard import typechecked
-import arkouda.array_api as xp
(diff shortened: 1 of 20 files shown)


arkouda/pdarraycreation.py
```diff
@@ -285,24 +301,30 @@ def array(
...
+            rep_msg = generic_msg(
+                cmd=f"array<{a.dtype.name},{ndim}>",
+                args={"shape": full_size},
+                payload=aview,
+                send_binary=True,
+            )
+            return create_pdarray(rep_msg) if dtype is None else akcast(create_pdarray(rep_msg), dtype)
+
+        else:
+            return xp.asarray(a, dtype=dtype)  # type: ignore
```
jeremiah-corrado, line 327: Why not return a multidimensional pdarray here? 

I don't think the array-api's `Array` type needs to be a part of the regular arkouda API. As far as I understand, the vision was for `Array` to be a thin wrapper around a `pdarray` that implements the Array API specification, while `pdarray` itself could implement something more like numpy's API).

arkouda/pdarraycreation.py
```diff
@@ -472,23 +494,29 @@ def zeros(
...
-    repMsg = generic_msg(cmd=f"create<{dtype_name},1>", args={"shape": size})
+        raise TypeError(f"unsupported dtype {dtype}")
+    shape, ndim, full_size = _infer_from_size(size)
+
+    if ndim > get_max_array_rank():
+        raise ValueError(f"array rank {ndim} exceeds maximum of {get_max_array_rank()}")
+
+    if ndim > 1:
+        return xp.zeros(shape, dtype=dtype)  # type: ignore
+    repMsg = generic_msg(cmd=f"create<{dtype_name},{ndim}>", args={"shape": shape})
```
jeremiah-corrado, line 509: along the same lines, this block of code could just be:

```chapel
return create_pdarray(
  generic_msg(cmd=f"create<{dtype_name},{ndim}>", args={"shape": shape})
)
```

no `if ndim > 1` conditional needed, since the `create` command is already generic wrt `ndim`.

And then at some point, `zeros` in the array api module could/should be a wrapper around this function (instead of calling `f"create<{dtype_name},{ndim}>"` itself):

```chapel
def zeros(
    shape: Union[int, Tuple[int, ...]],
    /,
    *,
    dtype: Optional[Dtype] = None,
    device: Optional[Device] = None,
) -> Array:
    """
    Return a new array with the specified shape and type, filled with zeros.
    """
    from .array_object import Array

   ...

    if dtype is None:
       dtype = ak.Int64

    return Array._new(ak.zeros(shape, dtype))
```

jeremiah-corrado (commented):
I'm not seeing the motivation for importing the array-api module into arkouda propper to implement the multi-dim stuff. I would have thought that doing things the other way around would make more sense (i.e., make arkouda-propper support multi-dim pdarrays natively, and then build the Array API module on top of that (or, leave the Array API module as a separate thing for now)).

My mental model for the array-api module was for `Array` to be a thin wrapper around a `pdarray` that implements the Array API specification, while `pdarray` itself could implement something more like numpy's API. (`Array` needed to be a separate type, because there are places where the Array API deviates from numpy's API.)

I left a couple of comments about this, but haven't gone over all the changes yet:

arkouda/pdarraycreation.py
ajpotts: I made some changes, let me know what you think.

jeremiah-corrado (approved):
Looks good to me!

arkouda/pdarraycreation.py
jeremiah-corrado: sounds good, definitely doesn't need to happen any time soon, but would be nice from a software maintainability standpoint.

jaketrookman (approved):
Looks good

arkouda/pdarraycreation.py
```diff
@@ -49,6 +49,21 @@
 ]
 
 
+def _infer_from_size(size):
+    shape: Union[int_scalars, Tuple[int_scalars, ...]] = 1
+    if isinstance(size, tuple):
```
stress-tess, line 54: this seems to do the same thing as `_calc_shape` that I added to `util` a while back, we can probably combine them 
https://github.com/Bears-R-Us/arkouda/blob/844ad2fd73cd49a93ed5057f2ef5f5e7df061d8a/arkouda/util.py#L613-L615

I actually like the name of urs a lot though. Maybe `_infer_shape_from_size`?

stress-tess (approved):
Looks good! only comment is on a duplicated helper function

stress-tess (approved):

PR #3715, opened 2024-08-29
Closes #3709: reshape to return a multi-dimensional array

### LINKED ISSUE #3709: reshape to return a multi-dimensional array
`pdarrayclass.reshape` to return a multi-dimensional `pdarray`.  This is part of the effort to remove support for `ArrayView`.

ajpotts:
Modifies the reshape function to return a multi-dimensional `pdarray` rather than `ArrayView`.  

Also strips out the `ArrayView` class and module.

Closes #3709: reshape to return a multi-dimensional array

Co-authored with @drculhane .


arkouda/pdarrayclass.py
```diff
@@ -1754,18 +1737,24 @@ def reshape(self, *shape, order="row_major"):
...
         """
-        from arkouda.array_view import ArrayView
-
         # allows the elements of the shape parameter to be passed in as separate arguments
         # For example, a.reshape(10, 11) is equivalent to a.reshape((10, 11))
         if len(shape) == 1:
             shape = shape[0]
         elif not isinstance(shape, pdarray):
             shape = [i for i in shape]
-        return ArrayView(base=self, shape=shape, order=order)
```
jeremiah-corrado, line 1768: can the `order` argument be removed from reshape?

jeremiah-corrado (approved):
Looks good! One question:

arkouda/pdarrayclass.py
ajpotts: Good catch!  

I think we should remove it temporarily, but add it back in later.  It does seem like it will be used in some of the libraries we want to integrated with, like `scipy`.  

I made a separate ticket.  https://github.com/Bears-R-Us/arkouda/issues/3729

stress-tess (approved):
Looks good! thanks!

PR #3724, opened 2024-08-29
Update Commands.chpl to reflect recent changes to linalg module

jeremiah-corrado:
Updates `Commands.chpl` (ran `make register-commands` on master) to reflect changes made in https://github.com/Bears-R-Us/arkouda/pull/3693. This fixes a build failure when `chapel-py` is not available and build scripts need to fall back to the tracked `Commands.chpl` file.


ajpotts (approved):
Thanks for fixing this!

PR #3730, opened 2024-08-30
Option for running each benchmark in its own server instance

jeremiah-corrado:
Add options to `run_benchmarks.py` that allow each benchmark to be run in it's own server instance. This can be useful for ensuring that performance regressions/improvements aren't related to interactions between benchmarks.


stress-tess (approved):
I don't know much about slurm run commands but this all looks good to me!

ajpotts (approved):
Looks good to me.

PR #3745, opened 2024-08-30
Sparse matrix testing and added features

ShreyasKhandekar:
This PR adds more features to the SparseMatrix Module:
1. `fill_vals` : Populates the sparse matrix with a 1D array which has an equal number of elements as the number of non zero values in the sparse matrix.  The order in which the values are assigned is the natural iteration order of the sparse matrix (CSC vs CSR)
2. `to_pdarray`: Converts a sparse matrix to 3 pdarrays: row indices, col indices, and the values at the indices. The size of each array is equal and is equal to the nnz in the sparse array.

This PR also adds a few simple tests for the above utilities, as well as for `sparse_matrix_matrix_mult`

We also add a new compatibility shim `ArkoudaSparseMatrixCompat.chpl`, to make this work with 2.0. Note that even though this works with 2.0, performance will be worse than later releases. This is because https://github.com/chapel-lang/chapel/pull/24391 and https://github.com/chapel-lang/chapel/pull/25243 were two PRs that improved the performance of sparse computations which are not a part of 2.0

We also change the SparseMatrix module to be compiled by default in Arkouda.

Tested single and multilocale


ShreyasKhandekar:
I see the chapel portability tests failing, given that a lot of the new work done by Brad in order to support these sparse matrix computations in Arkouda came after 2.0, I'm not sure what we can do to resolve this because it's not compatible AFAIK. I may be mistaken here and @bradcray might have a better idea.

ShreyasKhandekar:
Yup, testing passes with multilocale. Although if we scale it up to more than 8 locales, it will run out of memory due to the inefficient implementations used for now.

Also, I think there is something up with the CI:
```
tests/alignment_test.py ...FFF.....                                      [  4%]
tests/bigint_agg_test.py F....                                           [  4%]
tests/bitops_test.py FFFFFFFF..                                          [  5%]
tests/categorical_test.py ...........................                    [  6%]
tests/client_dtypes_test.py ..........F                                  [  7%]
tests/client_test.py ...........                                         [  7%]
```
It's been stuck here for over 1.5 hours, and many tests are failing that I didn't modify in any way.

Could we restart it?

bmcdonald3:
> Although if we scale it up to more than 8 locales, it will run out of memory

Is this on your laptop or in all environments? If we run out of memory at 8 locales, this will fail in nightly testing, which does 16 nodes on an XC

ShreyasKhandekar:
I thought only the benchmarks did 16 locale runs, not the correctness tests?

bmcdonald3:
No, the correctness tests run at 16 locales as well

ShreyasKhandekar:
Ah, okay. I will try to reduce the problem size or the memory requirement (or both) to see if I can make it work on 16 locales then. Thanks for the catch!

ShreyasKhandekar:
Okay I think I've convinced myself that my previous tests where I was running out of memory and getting a segfault were because somehow the wrong (non-distributed) version of the sparse matrix matrix multiplication overload was being chosen. I've commented out that version so it doesn't get accidentally chosen. I've reduced the problem size anyway, just to be safe.

If the CI passes, I think we should be good to merge this.

bmcdonald3 (approved):

src/SparseMatrixMsg.chpl
```diff
@@ -80,9 +79,78 @@ module SparseMatrixMsg {
...
+        var repMsg: string; // response message with the details of the new arr
+
+        var gEnt = getGenericSparseArrayEntry(msgArgs.getValueOf("matrix"), st);
+        var gEntVals: borrowed GenSymEntry = getGenericTypedArrayEntry(msgArgs.getValueOf("vals"), st);
+
+        //Hardcode int for now
+        var vals = toSymEntry(gEntVals,int);
+        if gEnt.layoutStr=="CSC" {
+            // Hardcode for int right now
+            var sparrayEntry = gEnt.toSparseSymEntry(int, dimensions=2, layout.CSC);
```
jeremiah-corrado, line 129: I don't think it needs to change for this PR, but for future reference, the new annotations will allow you to make this procedure generic wrt array element type without changing too much code:

```chapel
@arkouda.instantiateAndRegister
proc fillSparseMatrixMsg(..., type array_dtype): MsgTuple throws {
  var vals = st[msgArgs["vals"]]: borrowed SymEntry(array_dtype, 1);
  var gEnt = getGenericSparseArrayEntry(msgArgs["matrix"].val, st);
  if gEnt.layoutStr=="CSC" {
    var sparrayEntry = gEnt.toSparseSymEntry(array_dtype, dimensions=2, layout.CSC);
  } ...
}
```

arkouda/sparrayclass.py
```diff
@@ -94,6 +100,56 @@ def __str__(self):  # This won't work out of the box for sparrays need to add th
...
+    -------
+    List[ak.pdarray]
+        A list of 3 pdarrays which contain the row indices, the column indices,
+        and the values at the respective indices within the sparse matrix.
+
+    Raises
+    ------
+    RuntimeError
+        Raised if there is a server-side error thrown, if the pdarray size
+        exceeds the built-in client.maxTransferBytes size limit, or if the bytes
```
jeremiah-corrado, line 115: Is this true? Afaik, the `maxTransferBytes` limit should only apply to converting to/from numpy arrays because the data have to be sent to/from the client. In this case, aren't you only creating three new array symbols on the server side and then sending their handles back to the client?

src/SparseMatrix.chpl
```diff
@@ -1,17 +1,57 @@
...
+                        moduleName=getModuleName(),
+                        errorClass="IllegalArgumentError"
+                        );
+    for((i,j), idx) in zip(spsMat.domain,A.domain) {
+        spsMat[i,j] = A[idx];
+    }
+  }
 
+  proc sparseMatToPdarray(const ref spsMat, ref rows, ref cols, ref vals){
+    for((i,j), idx) in zip(spsMat.domain,0..) {
```
jeremiah-corrado, line 39: note: this could be a `forall` (though I'm not sure how efficient it would be).

```chapel
const m = spsMat.domain.dim(1).size;
forall (i, j) in spsMat.domain {
  const idx = j + i * m; // don't quote me on this
  ...
}

```

jeremiah-corrado (approved):
Looks good, glad to see this going in!

Overall, I think it would be nice to get the server-side code converted over to the new function-registration framework at some point, but I definitely don't think that needs to happen for this PR.

Other than that, I just had a few minor notes:

arkouda/sparrayclass.py
ShreyasKhandekar: Ah, that's a copy paste error, nice catch.

src/SparseMatrix.chpl
jeremiah-corrado: Oh, good point! nvm

src/SparseMatrixMsg.chpl
ShreyasKhandekar: Yeah, I'm looking forward to making that change as well in a subsequent PR.

stress-tess (approved):
looks good to me! thanks sheryas!

PR #3747, opened 2024-09-04
Closes #3746 remove ForwardRef

### LINKED ISSUE #3746: remove ForwardRef
Remove `ForwardRef` from the code.

ajpotts:
Closes #3746 remove ForwardRef

### DIFF
--- a/arkouda/pdarraysetops.py
+++ b/arkouda/pdarraysetops.py
@@ -1,24 +1,29 @@
 from __future__ import annotations
 
-from typing import ForwardRef, Sequence, Union, cast
+from typing import Sequence, TypeVar, Union, cast
 
 import numpy as np
 from typeguard import typechecked
 
+
 from arkouda.client import generic_msg
 from arkouda.client_dtypes import BitVector
+from arkouda.groupbyclass import GroupBy, groupable, groupable_element_type, unique
+from arkouda.logger import getArkoudaLogger
 from arkouda.numpy.dtypes import bigint
 from arkouda.numpy.dtypes import bool_ as akbool
 from arkouda.numpy.dtypes import int64 as akint64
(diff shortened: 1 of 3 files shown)


arkouda/pdarraysetops.py
```diff
@@ -90,9 +95,9 @@ def _in1d_single(
         # this causes an issue with mypy for unknown reasons.
         if pda1.size == 0:
             return zeros(0, dtype=akbool)
-    if isinstance(pda2, pdarray) or isinstance(pda2, Strings) or isinstance(pda2, Categorical_):
-        if pda2.size == 0:
-            return zeros(pda1.size, dtype=akbool)
+        if isinstance(pda2, pdarray) or isinstance(pda2, Strings) or isinstance(pda2, Categorical_):
+            if pda2.size == 0:
+                return zeros(pda1.size, dtype=akbool)
```
stress-tess, line 100: was this meant to be indented?

stress-tess (commented):
This is awesome!!! I've tried a couple times to get rid of those forward refs and never made it happen! `TypeVar` seems pretty cool! I read up on it a bit here:
https://dev.to/decorator_factory/typevars-explained-hmo

The one thing I saw was we indented an if block and I want to make sure it was on purpose

arkouda/pdarraysetops.py
ajpotts: Yes, if you look at the logic it only makes sense if both `pda1` and `pda2` are `(pdarray, Strings, Categorical)` (not a Sequence), which mypy will complain about.  Indenting was the easy way of achieving this.

stress-tess (approved):
Looks good!

arkouda/pdarraysetops.py
stress-tess: makes sense! I like the change!

PR #3751, opened 2024-09-06
Closes #3665: add multi-dim support to arkouda.testing module

### LINKED ISSUE #3665: add multi-dim support to arkouda.testing module
add multi-dim support to `arkouda.testing` module.  For example, `assert_almost_equal` should work for multi dimensional arrays.

ajpotts:
Multi-dim seems to be working, resulting from the other recent PRs.  This ticket verifies it's working by adding some multi-dim unit tests.

Closes #3665: add multi-dim support to arkouda.testing module

### DIFF
--- a/tests/testing/asserters_test.py
+++ b/tests/testing/asserters_test.py
@@ -166,6 +166,38 @@ class TestDataFrame:
                 assert_almost_equal(df, df3, atol=atol, rtol=rtol)
             assert_almost_equivalent(convert_left(df), convert_right(df3), atol=atol, rtol=rtol)
 
+    @pytest.mark.skip_if_max_rank_less_than(3)
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    @pytest.mark.parametrize("left_as_arkouda", [True, False])
+    @pytest.mark.parametrize("right_as_arkouda", [True, False])
+    def test_assert_almost_equal_multi_dim(self, size, left_as_arkouda, right_as_arkouda):
+
+        both_ak = left_as_arkouda and right_as_arkouda
+        convert_left = self.get_converter(left_as_arkouda)
+        convert_right = self.get_converter(right_as_arkouda)
+
+        shape = (2, 2, size)
+
+        rng = ak.random.default_rng()
+        atol = 0.001
+        rtol = 0.001
+        a = ak.arange(size * 4, dtype="float64")
+        a2 = self.perturb(a, atol=atol, rtol=rtol, rng=rng)
+        a3 = a + rtol + atol
+
+        a = a.reshape(shape)
+        a2 = a2.reshape(shape)
+        a3 = a3.reshape(shape)
+
+        if both_ak:
+            assert_almost_equal(a, a2, atol=atol, rtol=rtol)
+        assert_almost_equivalent(convert_left(a), convert_right(a2), atol=atol, rtol=rtol)
+        if both_ak:
+            with pytest.raises(AssertionError):
+                assert_almost_equal(a, a3, atol=atol, rtol=rtol)
+        with pytest.raises(AssertionError):
+            assert_almost_equivalent(convert_left(a), convert_right(a3), atol=atol, rtol=rtol)
+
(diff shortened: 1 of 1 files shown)


tests/testing/asserters_test.py
```diff
@@ -166,6 +166,37 @@ def test_assert_almost_equal(self, size, left_as_arkouda, right_as_arkouda):
...
+        a2 = a2.reshape(shape)
+        a3 = a3.reshape(shape)
+
+        if both_ak:
+            assert_almost_equal(a, a2, atol=atol, rtol=rtol)
+        assert_almost_equivalent(convert_left(a), convert_right(a2), atol=atol, rtol=rtol)
+        with pytest.raises(AssertionError):
+            if both_ak:
+                assert_almost_equal(a, a3, atol=atol, rtol=rtol)
+            assert_almost_equivalent(convert_left(a), convert_right(a3), atol=atol, rtol=rtol)
```
stress-tess, line 198: I was gonna point out that when `both_ak`, I think this will only test `assert_almost_equal` and not `assert_almost_equivalent`. Because we'd exit the `with` block once the first `AssertionError` is raised. They both get called for the `a` and `a2` comparison, so I wanted to make sure this is what we wanted

But looking into it more it seems that `assert_almost_equivalent` just calls `assert_almost_equal` anyway, so I don't this really matters

tests/testing/asserters_test.py
```diff
@@ -166,6 +166,37 @@ def test_assert_almost_equal(self, size, left_as_arkouda, right_as_arkouda):
                 assert_almost_equal(df, df3, atol=atol, rtol=rtol)
             assert_almost_equivalent(convert_left(df), convert_right(df3), atol=atol, rtol=rtol)
 
+    @pytest.mark.skip_if_max_rank_less_than(2)
```
stress-tess, line 169: should this be `skip_if_max_rank_less_than(3)` since `shape = (2, 2, size)`?

tests/testing/asserters_test.py
ajpotts: This is a good catch.  I made the change here and will put in a separate PR to fix the rest of the module.

tests/testing/asserters_test.py
ajpotts: Good catch.  It should be fixed now.

jaketrookman (approved):
Looks good

tests/testing/asserters_test.py
```diff
@@ -889,6 +920,28 @@ def test_assert_equal(self, size, left_as_arkouda, right_as_arkouda):
                 assert_equal(df, df2)
             assert_equivalent(convert_left(df), convert_right(df2))
 
+    @pytest.mark.skip_if_max_rank_less_than(2)
+    @pytest.mark.parametrize("size", pytest.prob_size)
```
stress-tess, line 924: I think this need to be `@pytest.mark.skip_if_max_rank_less_than(3)` for the same reason about `shape=(2, 2, size)`

tests/testing/asserters_test.py
```diff
@@ -981,6 +1033,27 @@ def test_assert_arkouda_array_equal(self, size, left_as_arkouda, right_as_arkoud
                 assert_arkouda_array_equal(s, c)
             assert_arkouda_array_equivalent(convert_left(s), convert_right(c))
 
+    @pytest.mark.skip_if_max_rank_less_than(2)
```
stress-tess, line 1036: same here

tests/testing/asserters_test.py
ajpotts: Sorry about that, it should be fixed now.

tests/testing/asserters_test.py
ajpotts: Fixed.

tests/testing/asserters_test.py
stress-tess: no worries!

stress-tess (approved):
looks great!

PR #3752, opened 2024-09-06
Closes #2978:  Strip out ArrayView

### LINKED ISSUE #2978: Strip out ArrayView
Strip out the array view code. This was never fully implemented, so it's presence will just be confusing for users who think we currently have multi-dimensional array support. The idea to view a single dimensional array as multi-dimensional was based on numpy but doesn't really make sense in a distributed setting. With the progress on the array api work, there's a possibility we will be able to add support using chapel multi dimensional arrays eventually

ajpotts:
This PR strips out the remaining ArrayView code, as that object is no longer supported.

Closes #2978:  Strip out ArrayView


stress-tess (approved):
looks good thanks!

PR #3761, opened 2024-09-11
Closes #3757 ak.array gives unexpected results on a transposed numpy multi dimensional array

### LINKED ISSUE #3757: `ak.array` gives unexpected results on a transposed numpy multi-dimensional array.
**Describe the bug**
`ak.array` gives unexpected results on a transposed numpy multi-dimensional array.

**To Reproduce**
```
In [81]: 
    ...: import numpy as np
    ...: rows = 5
    ...: cols = 5
    ...: nda = np.random.randint(1, 10, (rows, cols))
    ...: nda
Out[81]: 
array([[7, 7, 7, 8, 6],
       [9, 4, 9, 6, 2],
       [2, 2, 3, 1, 9],
       [2, 7, 1, 3, 7],
       [4, 2, 7, 5, 1]])

In [82]: 
    ...: nda_to_ak = ak.array(nda)
    ...: nda_to_ak
    ...: 
Out[82]: array([array([7 7 7 8 6]) array([9 4 9 6 2]) array([2 2 3 1 9]) array([2 7 1 3 7]) array([4 2 7 5 1])])

In [83]: 
    ...: npa = np.transpose(nda)
    ...: npa
Out[83]: 
array([[7, 9, 2, 2, 4],
       [7, 4, 2, 7, 2],
       [7, 9, 3, 1, 7],
       [8, 6, 1, 3, 5],
       [6, 2, 9, 7, 1]])

In [84]: 
    ...: npa_to_ak = ak.array(npa)
    ...: npa_to_ak
    ...: 
    ...: 
Out[84]: array([array([7 7 7 8 6]) array([9 4 9 6 2]) array([2 2 3 1 9]) array([2 7 1 3 7]) array([4 2 7 5 1])])


```

ajpotts:
Closes #3757 ak.array gives unexpected results on a transposed numpy multi dimensional array

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -291,6 +291,8 @@ def array(
             raise RuntimeError(
                 "Array exceeds allowed transfer size. Increase ak.client.maxTransferBytes to allow"
             )
+        #   Make a copy to avoid error #3757
+        a = a.copy()
         # Pack binary array data into a bytes object with a command header
         # including the dtype and size. If the server has a different byteorder
         # than our numpy array we need to swap to match since the server expects
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -8,6 +8,7 @@ import pandas as pd
 import pytest
 
 import arkouda as ak
+from arkouda.testing import assert_arkouda_array_equal
 
 INT_SCALARS = list(ak.dtypes.int_scalars.__args__)
 NUMERIC_SCALARS = list(ak.dtypes.numeric_scalars.__args__)
@@ -103,6 +104,17 @@ class TestPdarrayCreation:
 
         with pytest.raises(TypeError):
             ak.array(list(list(0)))
+
+    @pytest.mark.skip_if_max_rank_less_than(2)
+    def test_array_creation_transpose_bug_reproducer(self):
+
+        import numpy as np
+
+        rows = 5
+        cols = 5
+        nda = np.random.randint(1, 10, (rows, cols))
+
+        assert_arkouda_array_equal(ak.transpose(ak.array(nda)), ak.array(np.transpose(nda)))
 
     def test_infer_shape_from_size(self):
         from arkouda.util import _infer_shape_from_size


arkouda/pdarraycreation.py
```diff
@@ -291,6 +291,8 @@ def array(
             raise RuntimeError(
                 "Array exceeds allowed transfer size. Increase ak.client.maxTransferBytes to allow"
             )
+        #   Make a copy to avoid error #3757
+        a = a.copy()
```
stress-tess, line 295: Do we know why this fixes the error? I haven't thought about it much but if we don't have to create a copy, that's prob better right?

stress-tess: I mean, I guess it's not that a big of a deal bc if it was passed in it was relatively small

stress-tess: yah it's exactly what I thought. when numpy does a transpose, it doesn't actually move any data around, it just takes the same base array in memory and switches it from `column_major` to `row_major` or vice versa (I forget which one C used vs Fortran)

```
>>> nda.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

>>> npa = np.transpose(nda)

>>> npa.flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

>>> npa.copy().flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
```

When we do a deep copy, it actually allocates new memory and interprets it correctly. We don't have support for different orders with the current impl. We used to have it with `arrayview` but that had it's own mess of problems. I guess we could try to check if the `OWNDATA` flag is false and then deep copy but idk if it's worth it

stress-tess: I think I'll just reference this comment in the issue for future reference and approve for now. I'll leave this unresolved so other reviewers can see it bc I think it kinda interesting. And i'd like to get other opinions

stress-tess (approved):
nice work!! this must've been a pain to track down

stress-tess (approved):
I was hoping to get a second opinion on this. But it fixes the problem, so im good with merging this

PR #3763, opened 2024-09-11
Closes #3283: histogram2d between different dtypes

### LINKED ISSUE #3283: histogram2d between different dtypes
the following code:
`import arkouda as ak`
`ak.histogram2d(ak.array([1, 2, 3]), ak.array([1.5, 2.5, 3.5]))`

gives the error:
`RuntimeError: Error: histogram2DMsg: (int64,float64) not implemented`

optimally, I would like the ability for all combinations of numeric dtypes to be _histogrammed_ against one another

jaketrookman:
This PR (Closes #3283) by adding mixed types to work with histogram2d and matches the result dtypes with Numpy.

### DIFF
--- a/src/Histogram.chpl
+++ b/src/Histogram.chpl
@@ -61,7 +61,7 @@ module Histogram
         return hist;
     }
 
-    proc histogramGlobalAtomic(x: [?aD] ?etype, y: [aD] etype, xMin: etype, xMax: etype, yMin: etype, yMax: etype, numXBins: int, numYBins: int, xBinWidth: real, yBinWidth: real) throws {
+    proc histogramGlobalAtomic(x: [?aD] ?etype1, y: [aD] ?etype2, xMin: etype1, xMax: etype1, yMin: etype2, yMax: etype2, numXBins: int, numYBins: int, xBinWidth: real, yBinWidth: real) throws {
         const totNumBins = numXBins * numYBins;
         var hD = makeDistDom(totNumBins);
         var atomicHist: [hD] atomic int;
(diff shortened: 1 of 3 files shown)


src/Histogram.chpl
```diff
@@ -61,7 +61,7 @@ module Histogram
         return hist;
     }
```
ajpotts, line 63: Could you please squash the commits?

ajpotts (approved):
Looks good!  

Please remember to squash the commits when you get a chance.

src/Histogram.chpl
ajpotts: Thanks!

src/Histogram.chpl
```diff
@@ -61,7 +61,7 @@ module Histogram
...
 
-    proc histogramGlobalAtomic(x: [?aD] ?etype, y: [aD] etype, xMin: etype, xMax: etype, yMin: etype, yMax: etype, numXBins: int, numYBins: int, xBinWidth: real, yBinWidth: real) throws {
+    proc histogramGlobalAtomic(x: [?aD] ?etype1, y: [aD] ?etype2, xMin: etype1, xMax: etype1, yMin: etype2, yMax: etype2, numXBins: int, numYBins: int, xBinWidth: real, yBinWidth: real) throws {
         const totNumBins = numXBins * numYBins;
         var hD = makeDistDom(totNumBins);
         var atomicHist: [hD] atomic int;
```
stress-tess, line 67: all the other `atomicHist` were turned into `real`, should this one be too?

stress-tess (commented):
question about the `histogramGlobalAtomic` histogram type. It's worth noting this never gets run by our tests bc the problem size has to be quite large to take this code path

src/Histogram.chpl
jaketrookman: Added this and confirmed I was not using the right build to check multidim

stress-tess (approved):

PR #3764, opened 2024-09-11
Refactor `OperatorMsg`

### LINKED ISSUE #3719: Remove `@arkouda.registerND` from `OperatorMsg.chpl`.
Remove `@arkouda.registerND` from `OperatorMsg.chpl`.

jeremiah-corrado:
Refactors OperatorMsg to use `instantiateAndRegister` with the goal of making the code clearer and more readable. Closes #3719

Previously, this module had a few monolithic message handlers that selected over all pairs of `dytpes`. Within each `when`-clause, the implementation would figure out the result type based on the kind of operator, and then call out to helper procedures in `binOps.chpl` that had various conditionals over the array types and/or the result array type to select the correct implementation of the operator (or return a not supported message) based on the types involved.

I found this scheme somewhat difficult to follow when reading/modifying the operatorMsg code. As such, the refactored implementation, has one message-handler for each kind of operator — where the goal is to have a consistently defined result type within that procedure. In some cases the procedures are overloaded for array element types like `bool` or `bigint` that require special implementations.

I'm hoping that the simplified layout leads to faster compilation and/or reduced memory consumption when compiling the module. I have yet to test this.


jeremiah-corrado:
Note: `tests/numpy/numeric_test.py::test_putmask` appears to be hanging in CI. It's not obvious why, and I've been unable to reproduce this locally so far (even with `CHPL_RT_NUM_THREADS_PER_LOCALE=2`). I'm still investigating.

PR #3766, opened 2024-09-11
Fixes #3762: Fix dataframe groupby aggregations when keys contain `NaN`s

### LINKED ISSUE #3762: mask `values` by `isnan` for dataframe groupby with `dropna=True`
```python
In [32]: df = ak.DataFrame({"A":[1,2,2,np.nan],"B":[3,4,5,6]})

In [33]: df.groupby("A").count()

ValueError: Attempt to group array using key array of different length

>>> df.groupby("A",dropna=False).count()
     B
A
1.0  1
2.0  2
NaN  1 (3 rows x 1 columns)
```
doing a dataframe groupby on a column containing `NaN`s causes an error because we are trying do the aggregation on the full column length. We need to be masking out the values in the `NaN` segment

```python
>>> df.to_pandas().groupby("A").count()
     B
A     
1.0  1
2.0  2
```

stress-tess:
This PR (fixes #3762) using dataframe groupby with keys that contain `NaN`s would cause the aggregations to fail. To resolve this, we mask out the values that belong to the `NaN` segment

Before this PR:
```python
>>> df = ak.DataFrame({"A":[1,2,2,np.nan],"B":[3,4,5,6], "C":[1,np.nan,2, 3]})
>>> df.groupby("A").count()
ValueError
...
ValueError: Attempt to group array using key array of different length
```

After this PR:
```python
>>> df = ak.DataFrame({"A":[1,2,2,np.nan],"B":[3,4,5,6], "C":[1,np.nan,2, 3]})
>>> df.groupby("A").count()
     B  C
A
1.0  1  1
2.0  2  1 (2 rows x 2 columns)

>>> df.to_pandas().groupby("A").count()
     B  C
A
1.0  1  1
2.0  2  1

>>> df.groupby(["A","C"],as_index=False).count()
     A    C  B
0  1.0  1.0  1
1  2.0  2.0  1 (2 rows x 3 columns)

>>> df.to_pandas().groupby(["A","C"],as_index=False).count()
     A    C  B
0  1.0  1.0  1
1  2.0  2.0  1
```

### DIFF
--- a/arkouda/dataframe.py
+++ b/arkouda/dataframe.py
@@ -4,6 +4,7 @@ import json
 import os
 import random
 from collections import UserDict
+from functools import reduce
 from typing import Callable, Dict, List, Optional, Tuple, Union, cast
 from warnings import warn
 
@@ -24,10 +25,11 @@ from arkouda.index import Index, MultiIndex
 from arkouda.join import inner_join
 from arkouda.numpy import cast as akcast
 from arkouda.numpy import cumsum, where
-from arkouda.numpy.dtypes import bigint
+from arkouda.numpy.dtypes import _is_dtype_in_union, bigint
 from arkouda.numpy.dtypes import bool_ as akbool
 from arkouda.numpy.dtypes import float64 as akfloat64
 from arkouda.numpy.dtypes import int64 as akint64
+from arkouda.numpy.dtypes import numeric_scalars
 from arkouda.numpy.dtypes import uint64 as akuint64
 from arkouda.pdarrayclass import RegistrationError, pdarray
 from arkouda.pdarraycreation import arange, array, create_pdarray, full, zeros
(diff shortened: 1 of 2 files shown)


arkouda/dataframe.py
```diff
@@ -104,14 +106,48 @@ class DataFrameGroupBy:
...
+    def __init__(self, gb, df, gb_key_names=None, as_index=True, dropna=True):
+
         self.gb = gb
         self.df = df
         self.gb_key_names = gb_key_names
         self.as_index = as_index
         for attr in ["nkeys", "permutation", "unique_keys", "segments"]:
             setattr(self, attr, getattr(gb, attr))
 
+        self.dropna = dropna
```
ajpotts, line 118: You don't have to pass in `dropna` if you don't want to b/c `self.gb.dropn`a is already set.

arkouda/dataframe.py
```diff
@@ -104,14 +106,48 @@ class DataFrameGroupBy:
...
+        if dropna:
+            from arkouda import all as ak_all
+            from arkouda import isnan
+
+            # calculate ~isnan on each key then & them all together
+            # keep up with if they're all_non_nan, so we can skip indexing later
+            key_cols = (
+                [df[k] for k in gb_key_names] if isinstance(gb_key_names, List) else [df[gb_key_names]]
+            )
+            where_key_not_nan = [
```
ajpotts, line 131: `where_key_not_nan` is already computed in `groupbyclass.py` at line 272.  I guess it's one of those time memory trade offs whether it's better to recompute or save it the first time.

https://github.com/Bears-R-Us/arkouda/blob/master/arkouda/groupbyclass.py#L272

arkouda/dataframe.py
stress-tess: oh yah that's a good point!

arkouda/dataframe.py
stress-tess: yaaah I saw this and it did annoy me. but I couldn't think of a great way to take advantage of this without having the list of arrays persist indefinitely (by making it a field of groupby). But I also didn't try too hard to think of a way. The recalculation shouldn't be too expensive tho

tests/dataframe_test.py
```diff
@@ -664,6 +662,35 @@ def test_gb_aggregations_with_nans(self, agg):
...
+        # verify reproducer for #3762 is fixed
+        df = ak.DataFrame({"A": [1, 2, 2, np.nan], "B": [3, 4, 5, 6], "C": [1, np.nan, 2, 3]})
+        ak_result = df.groupby("A").count()
+        pd_result = df.to_pandas().groupby("A").count()
+        assert_frame_equal(ak_result.to_pandas(retain_index=True), pd_result)
+
+        ak_result = df.groupby(["A", "C"], as_index=False).count()
+        pd_result = df.to_pandas().groupby(["A", "C"], as_index=False).count()
+        assert_frame_equal(ak_result.to_pandas(retain_index=True), pd_result)
+
```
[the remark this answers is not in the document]
stress-tess: great idea! I'll make that change!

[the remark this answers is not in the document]
ajpotts: Thanks!

ajpotts (approved):

PR #3778, opened 2024-09-18
Closes #3777: Fix testing logic in tests/testing/asserters_test.py

### LINKED ISSUE #3777: Fix testing logic in tests/testing/asserters_test.py
Here is an example where the logic is not quite right:
```
        with pytest.raises(AssertionError):
            if both_ak:
                assert_arkouda_array_equal(c, c2)
            assert_arkouda_array_equivalent(convert_left(c), convert_right(c2))
```
In this case, the second `assert` will not execute when `both_ak=True`.  This ticket is to fix the issue throughout the file.

ajpotts:
Closes #3777: Fix testing logic in tests/testing/asserters_test.py

Here is an example where the logic was not quite right:
```
        with pytest.raises(AssertionError):
            if both_ak:
                assert_arkouda_array_equal(c, c2)
            assert_arkouda_array_equivalent(convert_left(c), convert_right(c2))
```
In this case, the second `assert` will not execute when `both_ak=True`.  This PR fixes the issue throughout the file.

### DIFF
--- a/tests/testing/asserters_test.py
+++ b/tests/testing/asserters_test.py
@@ -125,9 +125,10 @@ class TestDataFrame:
         if both_ak:
             assert_almost_equal(a, a2, atol=atol, rtol=rtol)
         assert_almost_equivalent(convert_left(a), convert_right(a2), atol=atol, rtol=rtol)
-        with pytest.raises(AssertionError):
-            if both_ak:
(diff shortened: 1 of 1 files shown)


stress-tess (approved):
looks great! thanks

PR #3602, opened 2024-07-31
Closes #3526: refactor argsortMsg to remove registerND annotation

ajpotts:
This PR refactors `argsortMsg` to use `@arkouda.instantiateAndRegister`.

PR #3779, opened 2024-09-18
Argsort performance fix for float64

jeremiah-corrado:
Fixes the drop in float64 MSD radix sort performance observed [here](https://chapel-lang.org/perf/arkouda/16-node-xc/?startdate=2024/07/20&enddate=2024/08/22&configs=release&graphs=uniformlydistributeddata,powerlawdistributeddata).

### DIFF
--- a/arkouda/sorting.py
+++ b/arkouda/sorting.py
@@ -94,7 +94,6 @@ def argsort(
             args={
                 "name": pda.name,
                 "algoName": algorithm.name,
-                "objType": pda.objType,
                 "axis": axis,
             },
         )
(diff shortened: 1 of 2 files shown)


jeremiah-corrado:
Nvm, turns out the [change](https://github.com/Bears-R-Us/arkouda/pull/3602/files#diff-ada886379a6a491c2ce7b7df03c64366fbd377fe9014e3fb1cdfe94e54de28a9) that caused the performance regression was actually just a bug fix. The `algorithm` argument [used to be](https://github.com/Bears-R-Us/arkouda/blob/c19a5f5e84b8892e51a74b9960e47f7a74f642d9/src/ArgSortMsg.chpl#L455) ignored when argsorting real arrays. #3602 fixed that, causing the MSD **and** LSD results to be plotted instead of plotting LSD twice.

PR #3786, opened 2024-09-19
Improved domain and type query support for `registerCommand`

jeremiah-corrado:
Made the following improvements to `registerCommand`:
* array formals with static element types aren't instantiated for all dtypes
* array formal types can reference domain and type queries from other formal types
  * allows two array arguments to always have the same rank or type (preventing separate instantiation for both formals)  

**Examples:**

The following proc is only instantiated for supported array ranks (e.g., "asdf<1>", "asdf<2>", ...) because the element type is static:
```chapel
@arkouda.registerCommand()
proc asdf(ref x1: [?d] real) throws {
    ...
}
```

The following proc is only instantiated for one set of dtype and rank arguments (e.g., "pqrs<int,1>", "pqrs<real,1>", ...) because the domain and element type are shared:
```chapel
@arkouda.registerCommand()
proc pqrs(ref x1: [?d] ?t, x2: [d] t) throws {
    ...
}
```

### DIFF
--- a/src/registry/register_commands.py
+++ b/src/registry/register_commands.py
@@ -139,10 +139,6 @@ def get_formals(fn, require_type_annotations):
                     ten = "<array>"
                     extra_info = [None, None]
 
-                    # TODO: support referencing a domain from another array's domain query
-                    # (e.g., 'proc foo(x: [?d], y: [d])')
-                    # to avoid instantiating all combinations of array ranks for the 2+ arrays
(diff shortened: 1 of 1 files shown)


ajpotts (approved):
This is awesome!  Thanks!

stress-tess (approved):
looks great!! thanks jeremiah!

PR #3787, opened 2024-09-20
Parallel sparseMatrixtoPdarray implementation

jeremiah-corrado:
Adds a parallel and distributed implementation of the sparse-matrix to pdarray conversion algorithm that creates three pdarrays populated with: rows, columns, and values.

This hasn't been fully tested for distributed sparse arrays yet, but the algorithm should be ready to work with them once `SparseSymEntry` is updated to be distributed.

### DIFF
--- a/src/SparseMatrix.chpl
+++ b/src/SparseMatrix.chpl
@@ -38,9 +38,9 @@ module SparseMatrix {
 
   /*
     Fill the rows, cols, and vals arrays with the non-zero indices and values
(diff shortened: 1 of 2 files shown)


ShreyasKhandekar (approved):
This looks good to me. I think we can definitely use more of the information in the `CSDom` backing the matrix to make some operations (especially the nested loops as you noted) more efficient here, but that can go in in subsequent PRs.

stress-tess (approved):
looks great!! thanks jeremiah!

PR #3814, opened 2024-10-07
Fix performance regression in array transfer performance

jeremiah-corrado:
Fix a recent [regression](https://chapel-lang.org/perf/arkouda/chapcs/?startdate=2024/09/01&enddate=2024/10/06&graphs=arraytransferperformance) in array transfer performance caused by an extra array copy added [here](https://github.com/Bears-R-Us/arkouda/pull/3761/files).

The copy was included to ensure the array data are properly transposed on the server side following a shallow numpy transpose (where the buffer remains the same and an array-metadata flag is set to indicate that the array was transposed). The change fixed https://github.com/Bears-R-Us/arkouda/issues/3757, but added overhead. This PR only makes the extra copy for the shallow-transpose case, which should restore performance for the common case.

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -291,16 +291,19 @@ def array(
             raise RuntimeError(
                 "Array exceeds allowed transfer size. Increase ak.client.maxTransferBytes to allow"
             )
-        #   Make a copy to avoid error #3757
-        a = a.copy()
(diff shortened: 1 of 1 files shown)


bmcdonald3 (approved):

ajpotts (approved):
This is awesome.  Thanks for fixing this!

PR #3828, opened 2024-10-09
Closes #3827:  rename flatten to split

### LINKED ISSUE #3827: rename flatten to split
flatten is should be renamed to split to better match with the numpy definitions of the two functions:
https://numpy.org/doc/2.0/reference/generated/numpy.ndarray.flatten.html
https://numpy.org/doc/stable/reference/generated/numpy.split.html

ajpotts:
This pr renames `flatten` to `split` and `split` to `regex_split`.

Closes #3827:  rename flatten to split

### DIFF
--- a/tests/regex_test.py
+++ b/tests/regex_test.py
@@ -52,7 +52,7 @@ class TestRegex:
     def test_empty_string_patterns(self):
         lit_str = ["0 String 0", "^", " "]
         ak_str = ak.array(lit_str)
-        has_regex_arg = ["contains", "startswith", "endswith", "peel", "flatten"]
+        has_regex_arg = ["contains", "startswith", "endswith", "peel", "split"]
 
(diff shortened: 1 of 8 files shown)


tests/regex_test.py
```diff
@@ -251,7 +251,7 @@ def test_regex_peel(self):
         assert ["", "", "1f2g"] == d_right.to_list()
         assert ["", "", "__f____g"] == u_right.to_list()
 
-    def test_regex_flatten(self):
+    def test_regex_split(self):
```
drculhane, line 254: You've renamed two different functions to "regex_split."  Is that intentional?

ajpotts: I renamed the second one to t`est_regex_regex_split`, since is is a `regex` test for `regex_split`.

jaketrookman (approved):
Looks good

PR #3829, opened 2024-10-09
Closes 3818 -- eliminates warning messages about tilde vs not

### LINKED ISSUE #3819: Update ~ (deprecated) to not in tests
The make test gives about a half dozen warnings that the use of ~ with bools is deprecated, and that it should be replaced with not.  These instances are in series_test.py, util_test.py, and numpy/dtypes_test.py.

drculhane:
Closes #3819 -- this is really minimal.  Changes maybe 7 lines in 3 files.

### DIFF
--- a/tests/array_api/util_functions.py
+++ b/tests/array_api/util_functions.py
@@ -27,12 +27,12 @@ class TestUtilFunctions:
         assert xp.all(a)
 
         a[3, 4] = False
-        assert ~xp.all(a)
+        assert not xp.all(a)
 
     @pytest.mark.skip_if_max_rank_less_than(2)
     def test_any(self):
         a = xp.zeros((10, 10), dtype=ak.bool_)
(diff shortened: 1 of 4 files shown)


jaketrookman (commented):
This looks good, but it looks like there are 2 more examples of `assert ~` in `tests/util_functions.py`.

drculhane:
> This looks good, but it looks like there are 2 more examples of `assert ~` in `tests/util_functions.py`.

Ah, thanks.  Not sure how i missed those.  I'll change those, too.

drculhane:
Since the change to array_api tests requires multi-dim to test, I'm running another make before I push those changes up, even though the changes themselves were only in the python.

jaketrookman (approved):

ajpotts (approved):
Thanks Andy!

Issue #3837, opened 2024-10-11
argTypeReductionMessage axis argument as list


ajpotts:
Refactor argTypeReductionMessage to have axis argument as list.

Per @jeremiah-corrado suggestion:
One suggestion, perhaps for a future PR, I think the axis argument could be replaced with a `list(int)` to avoid the `axis.rank == 1` checks and extra instantiations.

PR #3841, opened 2024-10-15
Support where-clause evaluation in registration annotations

jeremiah-corrado:
Add (primitive) support for where-clause evaluation to `registerCommand` and `instantiateAndRegister`. Both annotations now evaluate the procedure's where-clause and will only generate instantiations for sets of generic parameters where the where-clause would evaluate to `true`.

The primary goal of this change is to be able to remove extraneous overloads of command procedures that simply return an error message indicating that some set of types and/or array ranks are not supported for that command. For example, in the following code snippet, the second overload of `asdf` can now be removed because the build system won't attempt to generate an instantiation for it:

```chapel
@arkouda.registerCommand
proc asdf(x: [?d] ?t): [d] t throws
  where t != bigint
{
  // ...
}

proc asdf(x: [?d] ?t): [d] t throws
  where t == bigint
    do throw new Error("'asdf' does not support 'bigint'");
```

This feature is not totally general-purpose yet. **Current limitations:**
* only some `param`-function calls are currently supported. Specifically, support for`isIntegralType` and other related procedures from Chapel's Types module is hard-coded in, but any other function calls within a where-clause will not work
* references to queried array types and queried domains are supported; however:
  * only the rank of a queried domain can be referenced (as in the code example below), none of the domain's other `param` or `type` methods/fields are accessible
  * referencing an array's `etype` or `rank` directly is not yet supported (ex: `where x.etype != bool && x.rank < 3`)
 * referencing `param` or `type` fields of a non-array argument is not yet supported (ex: `where mySparseSym.matLayout == Layout.CSR`)

Eventually, `registerCommands.py` should be modified to use the Compiler's resolver to speculatively evaluate where-clauses in a much more robust manner. This would in principle remove the above limitations, but is likely a more significant development effort than the solution in this PR.

Limitations aside, this PR allows most of the extraneous procedures currently in Arkouda to be removed.

Note: one can opt out of this feature by setting the `ignoreWhereClause` argument in the annotation call. This could be useful in an example like the one below, or to get around one of the limitations listed above.

```chapel
@arkouda.registerCommand(ignoreWhereClause=true)
proc jkl(x: [?d] ?t): [d] t throws
  where d.rank == 1 {
    // special 1D implementation ... 
}

proc jkl(x: [?d] ?t): [d] t throws
  where d.rank > 1 {
    // general ND implementation ... 
}
```

(resolves: https://github.com/Bears-R-Us/arkouda/issues/3837)

### DIFF
--- a/src/registry/register_commands.py
+++ b/src/registry/register_commands.py
@@ -1106,36 +1106,34 @@ class WCFunc(WCNode):
 
     def eval(self, args, translation_table=None):
         # TODO: this is a really bad way to do this. the compiler should be leveraged much more heavily here
+        arg = self.actuals[0].eval(args, translation_table)
         if self.name == "isIntegralType":
-            return self.actuals[0].eval(args, translation_table) in [
-                "int",
-                "int(8)",
-                "int(16)",
-                "int(32)",
-                "int(64)",
-                "uint",
-                "uint(8)",
-                "uint(16)",
-                "uint(32)",
-                "uint(64)",
+            return arg in [
+                "int8",
+                "int16",
+                "int32",
+                "int64",
+                "uint8",
+                "uint16",
+                "uint32",
+                "uint64",
             ]
-        if self.name == "isRealType":
-            return self.actuals[0].eval(args, translation_table) in [
-                "real",
-                "real(32)",
-                "real(64)",
+        elif self.name == "isRealType":
+            return arg in [
+                "float32",
+                "float64",
             ]
-        if self.name == "isComplexType":
-            return self.actuals[0].eval(args, translation_table) in [
+        elif self.name == "isComplexType":
+            return arg in [
                 "complex",
-                "complex(64)",
-                "complex(128)",
+                "complex64",
+                "complex128",
             ]
-        if self.name == "isImagType":
-            return self.actuals[0].eval(args, translation_table) in [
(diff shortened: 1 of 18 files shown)


bmcdonald3 (approved):

ajpotts (approved):
Thanks!  Look forward to using this!

src/registry/register_commands.py
```diff
@@ -987,11 +1008,192 @@ def gen_command_proc(name, return_type, formals, mod_name, config):
...
+        # note: scalar type names are canonicalized to ensure 'int' == 'int(64)' (for example)
+        if width is not None:
+            self.value = canonicalize_type_name(f"{ast}({width})")
+        elif isinstance(ast, chapel.Identifier):
+            self.value = canonicalize_type_name(ast.name())
+        elif isinstance(ast, chapel.IntLiteral):
+            self.value = ast.text()
+        elif isinstance(ast, chapel.Dot):
+            self.value = ast.receiver().name() + "." + ast.field()
+            # 🥲
```
stress-tess, line 1167: 🥲 

I love this comment so much lol

stress-tess (approved):
Thanks for doing this jeremiah!

stress-tess (approved):

PR #3834, opened 2024-10-10
refactor argTypeReductionMessage

ajpotts:
This PR refactors `argTypeReductionMessage` to remove `@arkouda.registerND`.  `sum`, `prod`, `min`, and `max` are separated into distinct functions.  Also, the `axis` argument is added to `sum`, `prod`, `min` and `max` in `pdarrayclass.py`, and corresponding unit tests are added.

PR #3845, opened 2024-10-16
Part of argTypeReductionMessage refactor

ajpotts:
Part of argTypeReductionMessage refactor

Adds  `sum`, `prod`, `min`, and `max` functions.  Also, the `axis` argument is added to `sum`, `prod`, `min` and `max` in `pdarrayclass.py`, and corresponding unit tests are added.

This PR has the additions from PR  #3834, without stripping out the old code.  This is to isolate a problem that was causing the CI to hang.


stress-tess (approved):
looks good!

PR #3857, opened 2024-10-22
Fix broken error reporting for Chapel 2.0 in `register-commands.py`

jeremiah-corrado:
Fix a bug in `register-commands.py`'s error reporting, where the code was only working with versions of chpl >=2.1. Errors should now be reported correctly for versions 2.0 - 2.2.


stress-tess (approved):
ah I'm running 2.2. no wonder I didn't see it. Thanks for finding this @drculhane and thanks for fixing it @jeremiah-corrado

ajpotts (approved):
Thanks for fixing this!

PR #3809, opened 2024-10-03
3807 update trig and hyp functions

drculhane:
Closes #3807, updating of trig and hyperbolic functions to new interface.

Only two files are changed:

arkouda/numpy/_numeric.py has two new functions at the top (where_helper and datatype_check), but the primary change is inside _trig_helper, which now calls the trig and hyp functions directly, instead of going through "efunc".

src/EfuncMsg.chpl has new procs for all of these functions, written with the new interface.  All of the trig and hyp functions have been removed from the long Select-When clauses that use the old interface.

PR #3863, opened 2024-10-23
Closes 3809 moves trig and hyp fns to new interface

### LINKED ISSUE #3860: set max rank for test_tile to 3
Set max rank for `test_tile` to 3, otherwise the test will fail when the maximum rank is 2.

drculhane:
Closes #3809, closes #3860


ajpotts (approved):
Thanks for updating the PR!

PR #3880, opened 2024-11-01
Eliminates duplicates in tests/numpy/numeric_test.py

### LINKED ISSUE #3879: Eliminate duplicate test functions in numeric_test.py
Somehow multiple tests have been duplicated.  The tests should be analyzed for whether they are complete duplicates.  Where they are, the extra one should be deleted.  Where they differ, both should be deleted, and in their place, one test that captures our intent.

drculhane:
Closes #3879.


ajpotts (approved):
Thanks for taking care of this!

jaketrookman (approved):

PR #3890, opened 2024-11-13
Gamma collab

jaketrookman:
Work in progress of implementing Gamma function distribution into the random module
Need to look into setting seed, linter error, testing failure


ajpotts:
@jaketrookman sent me this link to the numpy version: https://github.com/numpy/numpy/blob/main/numpy/random/src/distributions/distributions.c#L220

src/RandMsg.chpl
```diff
@@ -438,6 +438,100 @@ module RandMsg
...
+            writeln("OMG FIRST LOOP IS BROKE!!");
+
+            return -1.0;  // we failed 10000 times in a row which should practically never happen
+
+        } 
+        else {
+            var b = kArg - 1/3;
+            var c = 1/sqrt(9 * b);
+            var X = 0.0;
+            var V = -1.0;
```
ajpotts, line 485: I think this is the bug.  `var V = -1.0;` needs to go inside the while loop.  Otherwise, it does `V * V * V` over and over again which eventually goes to infinity.

drculhane: On a first analysis, it looks like the V = -1.0 line was inserted so that the "while V <= 0" loop almost immediately below will execute at least once.  But that could be done via a do-while loop, as is done in the python.

src/RandMsg.chpl
```diff
@@ -438,6 +438,100 @@ module RandMsg
...
+                count+= 1;
+            }
+            writeln("\n\n\n\n\n\n\n\n\n\n\n\n\n\n");
+            writeln("OMG FIRST LOOP IS BROKE!!");
+
+            return -1.0;  // we failed 10000 times in a row which should practically never happen
+
+        } 
+        else {
+            var b = kArg - 1/3;
```
ajpotts, line 482: 1/3 needs to be 1/3.0 for correct floating point division.  Same for any other computations in this function.  This is
Screenshot from 2024-12-19 16-25-04
 causing the distribution to be slightly off.

ajpotts:
Covered by PR https://github.com/Bears-R-Us/arkouda/pull/4089

PR #3892, opened 2024-11-14
Causes error in registration of cumsum and cumprod functions

drculhane:
As documented in the comments starting at line 179 of src/EfuncMsg.chpl, this causes an error in the registration of the cumsum and cumprod functions, using the new user interface.

Each has two procs defined: one for input types int, uint and real, where output type = input type, and one for input type bool, where output = int.

Only the first of the procs is registered.  In this example, one has the "bool" version first, and only that is registered, and the other has the "bool" version second, and it 's not registered.

I can get around this by making one proc for each, and using a separate function to assign the output type, but there is definitely an error somewhere in here.


ajpotts:
See comments on #3893

Issue #3839, opened 2024-10-11
new benchmarks to output performance graph format


ajpotts:
The new benchmarks should have an option to output the performance graph format used by the chapel team.

PR #3894, opened 2024-11-14
Part of #3839 new benchmarks to output performance graph format

ajpotts:
This PR adds a script `benchmark_v2/reformat_benchmark_results.py` to convert the benchmark results from json to a format that matches the chapel team viz.  It only processes this for `argsort`, `coargsort`, and `aggregate`.  The other benchmarks will be added in a future PR.

Part of #3839 new benchmarks to output performance graph format


benchmark_v2/reformat_benchmark_results.py
```diff
@@ -0,0 +1,415 @@
...
+    parser.add_argument("--annotations", default="", help="File containing annotations")
+    parser.add_argument("--configs", help="comma seperate list of configurations")
+    parser.add_argument("--start-date", help="graph start date")
+    parser.add_argument("--benchmark-data", help="the benchnmark output data in json format.")
+
+    return parser
+
+
+def get_header_dict(directory_path):
+    """Get a dictionary there the keys are the benchmark name and the values are list of header fields for the .dat file."""
```
drculhane, line 136: This docstring appears below also, and looks like it was probably cut-and-pasted from somewhere.  It's just a bit unclear and ungrammatical.  Not a show-stopper, but could you look at this and maybe improve it?

benchmark_v2/reformat_benchmark_results.py
```diff
@@ -0,0 +1,415 @@
...
+
+def get_nested_value(data: dict, keys: list):
+    """Look up a value in a dictionary using a list of keys."""
+    for key in keys:
+        if isinstance(data, dict):
+            data = data.get(key, {})
+        else:
+            return None
+    return data if data != {} else None
+
```
drculhane, line 187: This feels a bit tricky to me, so let me see if I understand it.
Once you find a key, you change data to the value data.get(key,{}), and it's no longer a dictionary.
That effectively ends your for loop.
Is that accurate?
If so, (1) it's not the way I would do it, but (2) approval isn't a matter of "did you do it the way I would?"

But since I find this a bit confusing, I'd appreciate maybe some comments making it just a bit clearer.

drculhane (approved):
I've made a couple of comments in the code.  Other than that, (1) there's a lot in here, and while the code is straightforward, I'm like a member of a doctoral defense committee -- I just don't know this stuff well enough to confirm it, and I'm not able to generate test data to check it, and there aren't any tests included, however (2a) this is a utility to be used on the output of arkouda, as I understand it, so tests aren't necessarily needed, and (2b) the code is well-written with clear intent.  In my judgment, it merits approval, although I'd appreciate some feedback on my comments below, and a walkthrough in our next team meeting, if that's feasible, would be nice.

jaketrookman (approved):
Looks good to me, great work

PR #3901, opened 2024-11-15
Closes #3864 max and min of bool to return bool like numpy

### LINKED ISSUE #3864: max and min of bool to return bool like numpy
`max` and `min` functions on pdarrays of type `bool` should return `bool` to better align with numpy:
```
In [7]: pda = ak.zeros(size, dtype=dtype)

In [8]: ak.max(pda)
Out[8]: 0

In [9]: type(ak.max(pda))
Out[9]: numpy.int64

In [10]: type(np.max(pda.to_ndarray()))
Out[10]: numpy.bool_
```

ajpotts:
Closes #3864 max and min of bool to return bool like numpy


jaketrookman (approved):
Looks good to me

drculhane (approved):

PR #3913, opened 2024-11-21
Closes #3912:  failing unit test test_is_locally_sorted_multidim unde…

### LINKED ISSUE #3912: failing unit test test_is_locally_sorted_multidim under gasnet nl==2

ajpotts:
This fixes a bug in the unit tests that fail when the number of locales is 2.  The reason is that a small random array will occasionally have a sorted dimension, causing `is_locally_sorted(a, axis=axis)` to have `True` values.  The solution is to increase the array dimension, and also skip the test for a larger number of locales.

Closes #3912:  failing unit test test_is_locally_sorted_multidim under gasnet nl==2

### DIFF
--- a/tests/pdarrayclass_test.py
+++ b/tests/pdarrayclass_test.py
@@ -113,12 +113,13 @@ class TestPdarrayClass:
         assert not is_sorted(a)
 
     @pytest.mark.skip_if_max_rank_less_than(3)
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):

drculhane (approved):
Thanks for fixing this.

Issue #3931, opened 2024-12-09
bug in make install-deps


ajpotts:
**Describe the bug**
`make install-deps` runs into an error

**To Reproduce**
I cloned a fresh copy of master and ran `make install-deps` .

**Expected behavior**
It completes without error

**Error Message**
```

 /usr/bin/mkdir -p '/home/amandapotts/tmp/arkouda//dep/zeromq-install/include'
 /usr/bin/install -c -m 644 include/zmq.h include/zmq_utils.h '/home/amandapotts/tmp/arkouda//dep/zeromq-install/include'
 /usr/bin/mkdir -p '/home/amandapotts/tmp/arkouda//dep/zeromq-install/lib/pkgconfig'
 /usr/bin/install -c -m 644 src/libzmq.pc '/home/amandapotts/tmp/arkouda//dep/zeromq-install/lib/pkgconfig'
make[3]: Leaving directory '/home/amandapotts/tmp/arkouda/dep/build/zeromq-4.3.5'
make[2]: Leaving directory '/home/amandapotts/tmp/arkouda/dep/build/zeromq-4.3.5'
make[1]: Leaving directory '/home/amandapotts/tmp/arkouda/dep/build/zeromq-4.3.5'
rm -r /home/amandapotts/tmp/arkouda//dep/build/zeromq-4.3.5
echo '$(eval $(call add-path,/home/amandapotts/tmp/arkouda//dep/zeromq-install))' >> Makefile.paths
Installing HDF5
rm -rf /home/amandapotts/tmp/arkouda//dep/build/hdf5-1.14.4 /home/amandapotts/tmp/arkouda//dep/hdf5-install
mkdir -p /home/amandapotts/tmp/arkouda//dep /home/amandapotts/tmp/arkouda//dep/build
cd /home/amandapotts/tmp/arkouda//dep/build && curl -sL https://support.hdfgroup.org/releases/hdf5/v1_14/v1_14_4/downloads/hdf5-1.14.4-3.tar.gz | tar xz
cd /home/amandapotts/tmp/arkouda//dep/build/hdf5-1.14.4 && ./configure --prefix=/home/amandapotts/tmp/arkouda//dep/hdf5-install --enable-optimization=high --enable-hl && make && make install
/bin/sh: 1: cd: can't cd to /home/amandapotts/tmp/arkouda//dep/build/hdf5-1.14.4
make: *** [Makefile:149: install-hdf5] Error 2

```

PR #3932, opened 2024-12-10
Part of #3931 bug in make install deps

ajpotts:
This resolves the initial error documented in #3931.

Part of #3931 bug in make install deps


jaketrookman (approved):
Looks good to me

jaketrookman (approved):
Reaproving

PR #3940, opened 2024-12-12
Closes #3939: install-deps to work offline

### LINKED ISSUE #3939: install-deps to work offline
Some customers will need to install `arkouda` without internet connection, so `install-deps` needs to work offline.

ajpotts:
This PR modifies the Makefile such that
`make install-deps DEP_INSTALL_DIR=<directory>`
can be called, and the install script will attempt to look for the necessary files in `DEP_INSTALL_DIR` before calling out to the network.  This will be useful for offline builds.

`make install-arrow` is being handled under #3933.

Closes #3939: install-deps to work offline

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -133,7 +133,9 @@ jobs:
     - name: Make install-idn2
       run: |
         make install-idn2        
-     
+    - name: Make install-blosc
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good

Issue #3957, opened 2025-01-02
simplify offline builds


ajpotts:
Some users will need to be able to install arkouda in an offline environment.  Investigate ways to streamline this process.

PR #3959, opened 2025-01-04
Part 2 of #3957: simplify offline builds 

ajpotts:
This PR separates the target `arrow-download-source` from the target `install-arrow` in the Makefile.  It also makes the corresponding change to the other installs.  This is in order to have the ability to download the dependencies on one system, move the dependencies to another system, and then install them on the second system.  Otherwise, when the install script is run in the build directory, text files are created which have some paths hard coded.

Part of #3957


jaketrookman (approved):
Looks good to me

vasslitvinov (approved):

PR #3962, opened 2025-01-06
Offline build scripts pr2

ajpotts:


drculhane:
As we discussed in Slack, I had some difficulties in getting this to work, and I think they all trace to the sed in line 2.  I'm not sure how to do that better, so I'll just skip ahead to "when I made those edits manually, everything worked."  So once again, kudos for some outstanding docker work.

PR #3963, opened 2025-01-08
Add prototype support for an `ak.apply` function

jabraham17:
Adds prototypical support for calling an arbitrary python function on a distributed Arkouda array.

For example:

```python
import arkouda as ak
ak.connect()
arr = ak.array([1,2,3])
res = ak.apply(arr, "lambda x,: x+1")
# res is now [2, 3, 4]
``` 

This is a work in progress, and the usage of a string rather than a python function is a temporary limitation.

### DIFF
--- a/ServerModules.cfg
+++ b/ServerModules.cfg
@@ -1,6 +1,7 @@
 # Comment out modules using a `#` to exclude
 # a module from a build
 
+ApplyMsgFunctions
 ArgSortMsg
 ArraySetopsMsg
 AryUtil
--- a/arkouda/apply.py
+++ b/arkouda/apply.py
@@ -1,0 +1,38 @@
+from typeguard import typechecked
+
+from arkouda.client import generic_msg
+from arkouda.pdarrayclass import pdarray
+from arkouda.pdarraycreation import create_pdarray
+
+__all__ = [
+    "apply",
+]
+
+@typechecked
+def apply(arr: pdarray, func: str) -> pdarray:
+    """
+    Apply a lambda function to a pdarray. The lambda function should be a string
+
+    For example,
+    >>> apply(ak.array([1, 2, 3]), "lambda x,: x+1")
+
+    Parameters
+    ----------
+    arr : pdarray
+        The pdarray to which the function is applied
+
+    func : str
+
+        The lambda function to apply to the array. This should be a string that
+        has the following format: "lambda: x,: ...", where the "..." can be a
+        valid python expression.
+
+    Returns
+    -------
+    pdarray
+        The pdarray resulting from applying the lambda function to the input array
+    """
+
+    repMsg = generic_msg(cmd=f"apply<{arr.dtype},{arr.ndim}>",
+                            args={"x" : arr, "funcStr" : func})
+    return create_pdarray(repMsg)
--- a/src/ApplyMsgFunctions.chpl
+++ b/src/ApplyMsgFunctions.chpl
@@ -1,0 +1,21 @@
+module ApplyMsgFunctions {
+
+  use Python;
+
+  @arkouda.registerCommand
+  proc apply(const ref x: [?d] ?t, funcStr: string): [] t throws {
+    var ret = makeDistArray(d, t);
+    coforall l in d.targetLocales() do on l {
+      var interp = new Interpreter();
+      var func = new Function(interp, funcStr);
+      for sd in d.localSubdomains() {
+        for i in sd {
+          ret[i] = func(t, x[i]);
+        }
+      }
+    }
+    return ret;
+  }
+
+
+}
--- a/tests/apply_test.py
+++ b/tests/apply_test.py
@@ -1,0 +1,29 @@
+import pytest
+
+import arkouda as ak
+
+"""
+Encapsulates a variety of arkouda apply test cases.
+"""
+
+
+class TestApply:
+
+    @classmethod
+    def setup_class(cls):
+        cls.size = 1000
+
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64])
+    def test_apply_inc(self, dtype):
+        a = ak.arange(TestApply.size, dtype=dtype)
+        b = ak.apply(a, "lambda x,: x+1")
+        assert ak.all(b == a + 1)
+        assert b.dtype == dtype
+
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64])
+    def test_apply_cube(self, dtype):
+        a = ak.arange(TestApply.size, dtype=dtype)
+        b = ak.apply(a, "lambda x,: x*x*x")
+        assert ak.all(b == a * a * a)
+        assert b.dtype == dtype
+
--- a/arkouda/__init__.py
+++ b/arkouda/__init__.py
@@ -43,3 +43,4 @@ from arkouda.util import (
 from arkouda.scipy.special import *
 from arkouda.scipy import *
 from arkouda.testing import *
+from arkouda.apply import *
--- a/pytest.ini
+++ b/pytest.ini
@@ -14,6 +14,7 @@ testpaths =
     tests/array_api/stats_functions.py
     tests/array_api/util_functions.py
     tests/alignment_test.py
+    tests/apply_test.py
     tests/bigint_agg_test.py
     tests/bitops_test.py
     tests/categorical_test.py


setup.py
```diff
@@ -129,6 +129,7 @@
         "tables>=3.8.0",
         "pyarrow",
         "scipy<=1.13.1",
+        "cloudpickle"
         # chapel-py
```
ajpotts, line 133: cloudpickle also needs to be added to:  `arkouda-env.yml`, `arkouda-dev-env.yml`, and `pydoc/requirements.txt`.

tests/apply_test.py
```diff
@@ -0,0 +1,64 @@
...
+            return x*math.pi
+
+        a = ak.arange(TestApply.size, dtype=dtype)
+        b = ak.apply(a, times_pi, "float64")
+
+        res = ak.cast(a, "float64") * math.pi
+
+        assert ak.all(b == res)
+        assert b.dtype == "float64"
+
```
ajpotts, line 64: Would you please squash your commits before this PR is merged in.

jabraham17:
The CI failures appear unrelated to the contents of this PR

src/ApplyMsgFunctions.chpl
```diff
@@ -0,0 +1,88 @@
...
+
+      var xSym = st[msgArgs["x"]]: SymEntry(array_dtype, array_nd);
+      ref x = xSym.a;
+
+      var d = x.domain;
+      var ret = makeDistArray(d, array_dtype_to);
+
+      coforall l in d.targetLocales() do on l {
+        var interp = new Interpreter();
+        var func = new Value(interp, pickleData);
```
ajpotts, line 78: I just pulled the latest version and I am unable to get it to compile locally.  I'm getting the following error:
```
/home/amandapotts/git/arkouda//src/ApplyMsgFunctions.chpl:41: In function 'applyPickle':
/home/amandapotts/git/arkouda//src/ApplyMsgFunctions.chpl:78: error: Attempt to 'new' a function or undefined symbol
/home/amandapotts/git/arkouda//src/ApplyMsgFunctions.chpl:78: error: Attempt to 'new' a function or undefined symbol
make: *** [Makefile:540: arkouda_server] Error 1

```

jabraham17: With Chapel main? This PR relies on new features on main

Makefile
```diff
@@ -343,6 +343,14 @@ ifdef LD_LIBRARY_PATH
 CHPL_FLAGS += $(patsubst %,-L%,$(strip $(subst :, ,$(LD_LIBRARY_PATH))))
 endif
 
+# flags for python interop
+CHPL_FLAGS += --ccflags -isystem$(shell python3 -c "import sysconfig; print(sysconfig.get_paths()['include'])")
+CHPL_FLAGS += -L$(shell python3 -c "import sysconfig; print(sysconfig.get_config_var('LIBDIR'))")
+CHPL_FLAGS += --ldflags -Wl,-rpath,$(shell python3 -c "import sysconfig; print(sysconfig.get_config_var('LIBDIR'))")
+CHPL_FLAGS += -lpython$(shell python3 -c "import sysconfig; print(sysconfig.get_config_var('LDVERSION'))")
+CHPL_FLAGS += --ccflags -Wno-macro-redefined
```
e-kayrakli, line 351: Can you add comments here as to why these were necessary, and what they do?

arkouda/apply.py
```diff
@@ -0,0 +1,99 @@
...
+# TODO: it would be nice to typecheck that func takes and returns arr.dtype,
+# but that likely requires making pdarray generic over its dtype
+@typechecked
+def apply(
+    arr: pdarray,
+    func: Union[Callable, str],
+    result_dtype: Optional[Union[np.dtype, str]] = None,
+) -> pdarray:
+    """
+    Apply a python function to a pdarray. The function should take 1 argument
```
e-kayrakli, line 27: s/1/one

arkouda/apply.py
```diff
@@ -0,0 +1,99 @@
...
+    >>> import math
+    >>> arr = ak.randint(0, 10, 10_000)
+    >>> def times_pi(x):
+            return x*math.pi
+    >>> ak.apply(arr, times_pi, "float64")
+
+    Warning: This function is experimental and may not work as expected.
+    Known limitations:
+    - The use of some python modules inside of the function may crash the server.
+    Known modules know to crash the server are:
```
e-kayrakli, line 44: "Modules known to crash the server"

arkouda/apply.py
```diff
@@ -0,0 +1,99 @@
...
+        - numpy
```
e-kayrakli, line 45: Based on your comment in the PR message, I was under the impression that this was more about using numpy from inside Jupyter. But I guess the issue is more general?

arkouda/apply.py
```diff
@@ -0,0 +1,99 @@
...
+    ----------
+    arr : pdarray
+        The pdarray to which the function is applied
+
+    func : Union[Callable, str]
+
+        The function to apply to the array. This can be a callable function or
+        a string, but either way it should take a single argument and return a
+        single value. If a string, it should be a lambda function that takes a
+        single argument, e.g. "lambda x,: x+1".
```
e-kayrakli, line 60: I remember dangling comma being necessary for a reason I can't recall. Is that the case? If so, could you add that somewhere here?

src/ApplyMsgFunctions.chpl
```diff
@@ -0,0 +1,88 @@
...
+        getRoutineName(),
+        getModuleName()
+      );
+    } else {
+      const pickleDataStr = msgArgs["pickleData"].toScalar(string);
+      var pickleData: bytes;
+      {
+        // TODO: this is a big hack, ideally we properly implement the base64 module
+        // instead of using Python as a polyfill.
+        // even better, we should have a way to just send bytes from the client to the server
```
e-kayrakli, line 63: can we not send a Python `bytes`?

ServerModules.cfg
```diff
@@ -1,6 +1,7 @@
 # Comment out modules using a `#` to exclude
 # a module from a build
 
+ApplyMsgFunctions
```
e-kayrakli, line 4: While I see at least one exception, could you rename this as `ApplyMsg` for better consistency? `ApplyMsgFunctions` sounds like internal helpers to me (Sparse support has a similar module structure IIRC)

e-kayrakli (commented):
@jabraham17 -- thanks for this PR. I left some comments.
@ajpotts -- I see that Jade has some questions for you in the PR, could you take a look?

e-kayrakli:
> Some libraries seem to not play with being cloudpickled multiple times. For example, using numpy inside a function applied to a pdarray is fine. However, if running from a jupter notebook and the apply is run from a cell twice in a row, the server will segfault

Do you have an idea how to mitigate this? It sounds like the second time you cloudpickle a numpy function, it is not playing well with the server, is that correct? Could the server cache functions? Arguably, that will not prevent the client from resending them, but they can be ignored. Or ideally, client can maintain a state, in which the fact that a given function is cached by the server can be stored.

arkouda/apply.py
jabraham17: Its not just numpy. I saw other modules crash it like matplotlib.

jabraham17: The more general problem comes from cloudpickle. Its not clear if the problem is in how we use cloudpickle or in cloudpickle itself

jabraham17:
> Do you have an idea how to mitigate this? It sounds like the second time you cloudpickle a numpy function, it is not playing well with the server, is that correct? Could the server cache functions? Arguably, that will not prevent the client from resending them, but they can be ignored. Or ideally, client can maintain a state, in which the fact that a given function is cached by the server can be stored.

Its not just the use of the same function, its about the use of a module in a function. This snippet replicates the issue

```python
import arkouda as ak
import numpy as np

ak.connect()

n = 100_000
arr = ak.randint(0, 10, n)

def foo(x):
    np
    return 2.0
res = ak.apply(arr, foo, "float64")
print(res)

def bar(x):
    np
    return 1.0
res = ak.apply(arr, bar, "float64")
print(res)
```

jabraham17:
I have tracked down the crash to an issue with numpy and multiple interpreters: https://github.com/numpy/numpy/issues/28271. This error should not crash the server (only throw an exception), so there is something on the Chapel Python interop side to fix. But this will just make a nicer error message, it will not fix the problem. For that, we either need to reusue the interpreter for each locale or numpy needs to be fixed.

jabraham17:
> This error should not crash the server (only throw an exception), so there is something on the Chapel Python interop side to fix

This is a general [Chapel bug](https://github.com/chapel-lang/chapel/issues/26652), there is nothing that can be done within the Python module to fix it

src/ApplyMsgFunctions.chpl
jabraham17: Understood, I have pushed a commit that makes the PR buildable 2.3 (by adding compat modules for eq-23 and ge-24)

ajpotts: I was able to build it with the current version of chapel main.

arkouda/apply.py
```diff
@@ -0,0 +1,100 @@
...
+    For example,
+    >>> apply(ak.array([1, 2, 3]), lambda x: x+1)
+
+    Or,
+    >>> import math
+    >>> arr = ak.randint(0, 10, 10_000)
+    >>> def times_pi(x):
+            return x*math.pi
+    >>> ak.apply(arr, times_pi, "float64")
+
```
ajpotts, line 40: When I try this example, it works great the first time and then crashes the server on a second call.

```python

In [7]: ak.connect()

In [8]: import math

In [9]: arr = ak.randint(0, 10, 10_000)

In [10]: def times_pi(x):
    ...:             return x*math.pi
    ...: 

In [11]: ak.apply(arr, times_pi, "float64")
Out[11]: array([25.132741228718345 6.2831853071795862 21.991148575128552 ... 18.849555921538759 18.849555921538759 25.132741228718345])

In [12]: ak.apply(arr, times_pi, "float64")
^C---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
Cell In[12], line 1
----> 1 ak.apply(arr, times_pi, "float64")

File ~/anaconda3/envs/arkouda-dev/lib/python3.12/site-packages/typeguard/__init__.py:891, in typechecked.<locals>.wrapper(*args, **kwargs)

```

I know this may be the same error mentioned in other comments.  I'm not sure it's a good idea to merge this in when it can crash the server on basic examples.  In practice there may be multiple users sharing a server.

ajpotts (changes requested):
I love this so much!  

Unfortunately I don't think we can merge it in until we can reduce the potential for server crashes.

arkouda/apply.py
jabraham17: If you were using `numpy`, I would call it the same error as mentioned above and in the docs. But the server should not crash on this example.

```python
import arkouda as ak
ak.connect()

import math
arr = ak.randint(0, 10, 10_000)
def times_pi(x):
  return x*math.pi

ak.apply(arr, times_pi, "float64")
ak.apply(arr, times_pi, "float64")
for i in range(100):
  ak.apply(arr, times_pi, "float64")
```

This works fine for me. Can you provide more details so I can replicate this and fix it?

ajpotts: [log.txt](https://github.com/user-attachments/files/18694485/log.txt)

Here are the logs I am getting for this example.  Here is my configuration:
```python
In [5]: ak.get_config()
Out[5]: 
{'arkoudaVersion': 'v2025.01.13+9.ga60c66f2b',
 'chplVersion': '2.4.0 (351d4d59a5)',
 'ZMQVersion': '4.3.5',
 'HDF5Version': '1.14.4',
 'serverHostname': 'pop-os',
 'ServerPort': 5555,
 'numLocales': 1,
 'numPUs': 8,
 'maxTaskPar': 8,
 'physicalMemory': 67257028608,
 'distributionType': 'domain(1,int(64),one)',
 'LocaleConfigs': [{'id': 0,
   'name': 'pop-os',
   'numPUs': 8,
   'maxTaskPar': 8,
   'physicalMemory': 67257028608}],
 'authenticate': False,
 'logLevel': 'INFO',
 'logChannel': 'CONSOLE',
 'regexMaxCaptures': 20,
 'byteorder': 'little',
 'autoShutdown': False,
 'serverInfoNoSplash': False,
 'maxArrayDims': 1,
 'ARROW_VERSION': '19.0.0'}


```

ajpotts: @e-kayrakli asked me to check if the python versions matched.  I tried this snippet from python interpreter:
```
$ python3
Python 3.12.3 | packaged by conda-forge | (main, Apr 15 2024, 18:38:13) [GCC 12.3.0] on linux
```

and checked it matches the conda environment I compiled with:
```
$ conda env list | grep -v "^$\|#" |awk '{print $1;}'|xargs -I{} -d "\n" sh -c 'printf "Env: {}\t"; conda list -n {} |grep "^python\s";'
Env: base	
python                    3.9.20               he870216_1  
Env: arkouda-dev	python                    3.12.3          hab00c5b_0_cpython    conda-forge
```
They appear to match, but I still get the same seg fault.

jabraham17: I've made some improvements here. The server will now check for python version mismatches.

Additionally, the crashes from using numpy are avoided by reusing the interpreters for each call to `apply` 

While doing this, I also found a bug in register_commands.py which I am opening a separate PR for

arkouda/apply.py
```diff
@@ -0,0 +1,112 @@
...
+    vers_supported = getattr(apply, "is_version_supported", None)
+    if vers_supported is None:
+        interp_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        cmd_res = generic_msg("isVersionSupported", args={"versionString": interp_version})
+        vers_supported = parse_single_value(cast(str, cmd_res))
+        setattr(apply, "is_version_supported", vers_supported)
+    if not vers_supported:
+        interp_version = f"{sys.version_info.major}.{sys.version_info.minor}"
+        raise RuntimeError(
+            f"The current interpreter version ({interp_version}) is not supported by the server"
```
ajpotts, line 82: How difficult would it be to have the error recommend which version is supported?

jabraham17: https://github.com/Bears-R-Us/arkouda/pull/3963/commits/e6b97035971b87ef393d082c965898b22be94008 should do that

arkouda/apply.py
ajpotts: I can run the example without it crashing.  Thanks for all the fixes!

ajpotts (approved):
Looks good!  Thanks for all the hard work on this one.

tests/apply_test.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+
+import arkouda as ak
+
+"""
+Encapsulates a variety of arkouda apply test cases.
+"""
+
```
ajpotts, line 8: Would you please add a multi dimensional test?

arkouda/apply.py
```diff
@@ -0,0 +1,114 @@
...
+        a valid numpy dtype string, e.g. "float64". If a numpy dtype, it should
+        be a valid numpy dtype object, e.g. np.float64. This is not supported
+        for functions passed as strings.
+
+    Returns
+    -------
+    pdarray
+        The pdarray resulting from applying the function to the input array
+    """
+
```
ajpotts, line 72: When I run this with chapel 2.3.0 I get this error:
`RuntimeError: Error executing command: ErrorWithContext Line 55 In ApplyMsg.isVersionSupported: Python module not supported with this version of Chapel`
Is it possible to more gracefully throw an error?

ajpotts: Ideally it would be on the python side, so that it doesn't show a stack trace.  It could say something, like "The apply module is not supported for chapel versions <2.4.0"

tests/apply_test.py
ajpotts: There are lots of examples in `numeric_test`, if that helps.

tests/apply_test.py
```diff
@@ -0,0 +1,78 @@
...
+        assert ak.all(b == res)
+        assert b.dtype == "float64"
+
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.uint64])
+    def test_apply_shapes(self, prob_size, dtype):
+        max_rank = pytest.max_rank
+        for r in range(1, max_rank + 1):
```
ajpotts, line 73: Can you use `get_array_ranks` to get the list of available ranks?  Also, could you use `probsize/r`?  We've had a problem where the cube of a large number is a bit too large to be practical.

jabraham17: done in https://github.com/Bears-R-Us/arkouda/pull/3963/commits/38bf25900a53da3971e8daddf689544abc6097f1

arkouda/apply.py
jabraham17: https://github.com/Bears-R-Us/arkouda/pull/3963/commits/5c46f4417ea250a6bd51cbebfd2328755896444a adds this, giving an error like:
```
>>> ak.apply(arr, lambda x: x+1)
  File "<stdin>", line 1, in <module>
  File ".../lib/python3.12/site-packages/typeguard/__init__.py", line 891, in wrapper
    retval = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File ".../arkouda/apply.py", line 79, in apply
    raise RuntimeError(
RuntimeError: The apply module is not supported by the version of Chapel this server was built with.
```

ajpotts: I like this better.  Thanks!

e-kayrakli (approved):

1RyanK (approved):

ajpotts (approved):

PR #3950, opened 2024-12-19
Read multiple row groups in Parquet files correctly

jhh67:
Columns with more than one row group were not read correctly, which could lead to server crashes and perhaps memory corruption. This fix iterates through the column's row groups while maintaining a count of the total items read, and terminates the loop when the specified number of items have been read.

To do (see https://github.com/Bears-R-Us/arkouda/blob/master/CONTRIBUTING.md#writing-pull-requests)

- [x] make test
- [x] make pypy
- [x] flake8 arkouda

Closes #3951

PR #3969, opened 2025-01-10
Revert "Read multiple row groups in Parquet files correctly"

ajpotts:
PR #3950 caused `benchmarks/parquetMultiIO.py` to time out, so reverting this temporarily.

Reverts Bears-R-Us/arkouda#3950


drculhane (approved):
That's certainly odd.  But if it's causing a timeout, absolutely let's revert.

PR #3971, opened 2025-01-13
Closes #3970: comm_diagnostics function to return a DataFrame

### LINKED ISSUE #3970: comm_diagnostics function to return a DataFrame
It would be very useful to have a function in the `comm_diagnostics` module that would return the comm diagnostics statistics as a `DataFrame` object.

ajpotts:
This PR adds a `get_comm_diagnostics()` function that returns a `DataFrame` of diagnostic statistics.  It also modifies `print_comm_diagnostics` to print to the python interpreter in addition to the chapel logs.

Closes #3970: comm_diagnostics function to return a DataFrame

### DIFF
--- a/arkouda/comm_diagnostics.py
+++ b/arkouda/comm_diagnostics.py
@@ -2,6 +2,7 @@ from typing import cast as typecast
 
 from arkouda.client import generic_msg
 from arkouda.pdarrayclass import create_pdarray
+from arkouda.dataframe import DataFrame
 
 __all__ = [
     "start_comm_diagnostics",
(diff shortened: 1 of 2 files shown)


drculhane (approved):
I pulled and ran this this morning, and it certainly appears to work.  Next time we have a team meeting, can you explain why the test does a print?  That puzzles me, but I don't see it as a reason not to approve.

1RyanK (approved):

1RyanK:
@drculhane If I understand correctly, lines 61 and 62 of tests/comm_diagnostics_test.py are
```
print_comm_diagnostics_table(print_empty_columns=True)
print_comm_diagnostics_table(print_empty_columns=False)
```
These were not edited, so they are easy to miss.
Anyways, because of the edits to that function (specifically lines 114 - 117 of the update):
```
    if print_empty_columns:
        print(df.to_markdown())
    else:
        print(df[[col for col in df.columns if ak_sum(df[col]) != 0]].to_markdown())
```
The tests will cause prints.

PR #3985, opened 2025-01-24
Closes #3984 -- fixes error in ak.trunc and ak.ceil, also adds unit tests

### LINKED ISSUE #3984: Fix bug in ak.trunc
Python side calls the argument "array," but chapel calls it "pda".

drculhane:
Ryan found typos in ak.trunc and ak.ceil python side, which also served to surface the issue that there were no unit tests for either of these.  The typos have been fixed, and the unit tests implemented.


1RyanK (approved):

tests/numpy/numeric_test.py
```diff
@@ -1170,3 +1170,12 @@ def test_array_equal(self, prob_size, data_type, same_size, matching, nan_handli
                     0, 100, (prob_size if same_size else prob_size - 1), dtype=data_type
                 )
                 assert not (ak.array_equal(pda_a, pda_b))
+
+    @pytest.mark.parametrize("prob_size", pytest.prob_size)
+    def test_trunc_and_ceil (self, prob_size) :
```
ajpotts, line 1175: Would you be willing to parameterize the functions and then combine with the `floor` test?  We also need multi-dim tests for these functions.

drculhane: Sure.  I'll do both of those later today.

drculhane: I've added multi-dim (and removed the separate test for floor), and also included code to skip any ranks between 1 and the max that aren't included.  At the moment, two of the tests were cancelled, so I'll rerun them.

drculhane: Now they've passed.

jaketrookman (approved):

tests/numpy/numeric_test.py
ajpotts: Additionally, I was able to remove the `if` statement involving the rank by using `for rank in get_array_ranks()`.

drculhane: That change works locally, but at the moment, I'm encountering some issues with CI.  That happens sometimes, so I'm going to retry as needed.

ajpotts (approved):
Thanks for this!

PR #3987, opened 2025-01-27
Closes 3548 - updates and corrects docstrings in pdarraycreation.py

drculhane:
This updates the docstrings in pdarraycreation.py.

Summary:

- some Raises entries were outdated.  I fixed them.
- some Examples involved random numbers, so without adding a seed, they weren't reproducible.  I added a seed.
- some Examples gave results as integers when arkouda gives floating point.  I corrected that.
- most Examples had commas between array elements in the output, and arkouda mostly doesn't.  I removed the commas.
- because ones, zeros and full were almost identical code, I removed the code from ones and zeros, and replaced it with a single line call to full, with a fill_value of 1 or 0 as appropriate.
- added @typechecked to promote_to_common_type.

As to Ryan's comment below re ak.standard_normal, I'm puzzled at how the current version got into the code.  When I check it myself, I get what docdoc thinks I should get.  So I updated it.

### DIFF
--- a/arkouda/pdarraycreation.py
+++ b/arkouda/pdarraycreation.py
@@ -78,33 +78,33 @@ def from_series(series: pd.Series, dtype: Optional[Union[type, str]] = None) -> 
 
     Examples
     --------
+    >>> np.random.seed(1701)
     >>> ak.from_series(pd.Series(np.random.randint(0,10,5)))
-    array([9, 0, 4, 7, 9])
+    array([4 3 3 5 0])
 
     >>> ak.from_series(pd.Series(['1', '2', '3', '4', '5']),dtype=np.int64)
-    array([1, 2, 3, 4, 5])
+    array([1 2 3 4 5])
 
+    >>> np.random.seed(1701)
     >>> ak.from_series(pd.Series(np.random.uniform(low=0.0,high=1.0,size=3)))
-    array([0.57600036956445599, 0.41619265571741659, 0.6615356693784662])
+    array([0.089433234324597599 0.1153776854774361 0.51874393620990389])
 
     >>> ak.from_series(pd.Series(['0.57600036956445599', '0.41619265571741659',
                        '0.6615356693784662']), dtype=np.float64)
-    array([0.57600036956445599, 0.41619265571741659, 0.6615356693784662])
+    array([0.57600036956445599 0.41619265571741659 0.6615356693784662])
 
+    >>> np.random.seed(1864)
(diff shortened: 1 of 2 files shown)


arkouda/pdarraycreation.py
```diff
@@ -509,22 +503,28 @@ def ones(
...
     ----------
     size : int_scalars or tuple of int_scalars
-        Size of the array (only rank-1 arrays supported)
+        Size or shape of the array
     dtype : Union[float64, int64, bool]
-        Resulting array type, default float64
+        Resulting array type, default ak.float64
     max_bits: int
         Specifies the maximum number of bits; only used for bigint pdarrays
+        Included for consistency, as ones are ones
```
ajpotts, line 511: I don't understand this comment.

arkouda/pdarraycreation.py
```diff
@@ -768,17 +773,18 @@ def ones_like(pda: pdarray) -> pdarray:
...
-
-    >>> ones = ak.ones(5, dtype=ak.float64)
-    >>> ak.ones_like(ones)
-    array([1, 1, 1, 1, 1])
-
-    >>> ones = ak.ones(5, dtype=ak.bool_)
-    >>> ak.ones_like(ones)
-    array([True, True, True, True, True])
+    >>> sample = ak.zeros(5, dtype=ak.int64)
+     >>> ak.ones_like(sample)
```
ajpotts, line 777: Indent seems off on this one.

drculhane: Fixed.  docdoc caught that too.

arkouda/pdarraycreation.py
drculhane: I'll add more detail to that one, and to the similar comment in ak.zeros

1RyanK (approved):
The only semi-valid complaint that pops up for me in docdoc is
```
def standard_normal(size: int_scalars, seed: Union[None, int_scalars] = None) -> pdarray:
...
    >>> ak.standard_normal(3,1)
NEEDS FIX
Expected: array([-0.68586185091150265 1.1723810583573375 0.567584107142031])
Actual  : array([-0.68586185091150265 1.1723810583573377 0.567584107142031])
```
(If you're having trouble seeing it like I did, look at the last digit of the middle value)
But I imagine that may just be a hardware thing? So I can't really fault you for that.

arkouda/pdarraycreation.py
ajpotts: Thanks for humoring me on this one.

ajpotts (approved):

PR #4078, opened 2025-02-05
Closes #4096, adding `range` parameter to `histogram` functions

### LINKED ISSUE #4096: Add `range` parameter to `histogram` functions
**Feature Description**: We would like Arkouda's histogram functions to accept a `range` argument like in `numpy`

**Example Behavior**: The following should count only those elements of `a` that lie between -3 and 3, inclusive:
```python
histo, edges = histogram(a, range=(-3,3))
```

**Workarounds**: filter the contents of the array before passing it to the histogram function

**Additional context**: numpy [`histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html) [`histogram2d`](https://numpy.org/doc/stable/reference/generated/numpy.histogram2d.html) [`histogramdd`](https://numpy.org/doc/stable/reference/generated/numpy.histogramdd.html)

vasslitvinov:
Closes #4096. Required by https://github.com/Bears-R-Us/arkouda-contrib/pull/179

Adds `numpy`-style optional `range` parameter to client histogram functions:
* `histogram()`
* `histogram2d()`
* `histogramdd()`

Switches the histograms produced by `histogramdd()` to contain `float` counts instead of integer counts, to match `numpy`.

While there:

* Makes `histogramdDMsg()` more efficient.
* Factors out some histogram server code.
* Removes the dependence of `histogramdd()` on `cumprod()`.
* Cleans up Makefile to remove an extraneous `/` and trailing whitespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # Makefile for Arkouda
 ARKOUDA_PROJECT_DIR := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
+ARKOUDA_PROJECT_DIR := $(patsubst %/,%,$(ARKOUDA_PROJECT_DIR))
 
 PROJECT_NAME := arkouda
 ARKOUDA_SOURCE_DIR := $(ARKOUDA_PROJECT_DIR)/src
@@ -114,7 +115,7 @@ deps-download-source: zmq-download-source hdf5-download-source arrow-download-so
 
 DEP_DIR := dep
 DEP_INSTALL_DIR := $(ARKOUDA_PROJECT_DIR)/$(DEP_DIR)
-DEP_BUILD_DIR := $(ARKOUDA_PROJECT_DIR)$(DEP_DIR)/build
+DEP_BUILD_DIR := $(ARKOUDA_PROJECT_DIR)/$(DEP_DIR)/build
 
 ZMQ_VER := 4.3.5
 ZMQ_NAME_VER := zeromq-$(ZMQ_VER)
@@ -128,8 +129,8 @@ zmq-download-source:
     ifeq (,$(wildcard ${ZMQ_BUILD_DIR}*/.*))
         #   If the tar.gz not found, download it
         ifeq (,$(wildcard ${DEP_BUILD_DIR}/${ZMQ_NAME_VER}*.tar.gz))
-			cd $(DEP_BUILD_DIR) && curl -sL $(ZMQ_LINK) | tar xz		
-		#   Otherwise just unzip it
+			cd $(DEP_BUILD_DIR) && curl -sL $(ZMQ_LINK) | tar xz
+        #   Otherwise just unzip it
         else
 			cd $(DEP_BUILD_DIR) && tar -xzf $(ZMQ_NAME_VER)*.tar.gz
         endif
@@ -166,18 +167,18 @@ hdf5-download-source:
     ifeq (,$(wildcard ${HDF5_BUILD_DIR}*/.*))
         #   If the tar.gz not found, download it
         ifeq (,$(wildcard ${DEP_BUILD_DIR}/$(HDF5_NAME_VER)*tar.gz))
-			cd $(DEP_BUILD_DIR) && curl -sL $(HDF5_LINK) | tar xz		
-		#   Otherwise just unzip it
+			cd $(DEP_BUILD_DIR) && curl -sL $(HDF5_LINK) | tar xz
+        #   Otherwise just unzip it
         else
 			cd $(DEP_BUILD_DIR) && tar -xzf $(HDF5_NAME_VER)*.tar.gz
         endif
-    endif    
+    endif
 
 install-hdf5: hdf5-download-source
 	@echo "Installing HDF5"
 	rm -rf $(HDF5_INSTALL_DIR)
 	mkdir -p $(DEP_INSTALL_DIR) $(DEP_BUILD_DIR)
-	
+
 	cd $(HDF5_BUILD_DIR)* && ./configure --prefix=$(HDF5_INSTALL_DIR) --enable-optimization=high --enable-hl && make && make install
 	echo '$$(eval $$(call add-path,$(HDF5_INSTALL_DIR)))' >> Makefile.paths
 
@@ -218,9 +219,8 @@ arrow-download-source:
 		mkdir -p $(ARROW_DEP_DIR)
 		cd $(ARROW_BUILD_DIR)/cpp/thirdparty/ && ./download_dependencies.sh $(ARROW_DEP_DIR) > $(DEP_BUILD_DIR)/arrow_exports.sh
     endif
-    
+
 	rm -fr $(ARROW_BUILD_DIR)
-    
 
 install-arrow: arrow-download-source
 	@echo "Installing Apache Arrow/Parquet"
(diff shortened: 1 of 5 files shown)


arkouda/numpy/_numeric.py
```diff
@@ -1909,11 +1960,22 @@ def histogram2d(
...
-    y_bin_boundaries = linspace(y.min(), y.max(), y_bins + 1)
-    repMsg = generic_msg(cmd="histogram2D", args={"x": x, "y": y, "xBins": x_bins, "yBins": y_bins})
+
+    if range:
+        (xMin, xMax), (yMin, yMax) = range
+        xMin, xMax = _convToDtp(x.dtype, xMin, xMax)
+        yMin, yMax = _convToDtp(y.dtype, yMin, yMax)
+    else:
+        xMin, xMax, yMin, yMax = x.min(), x.max(), y.min(), y.max()
+
```
ajpotts, line 1973: To help with the mypy errors: 

```python

from typing import get_args`
from arkouda.numpy.dtypes import  numpy_scalars

    if range:
        (xMin, xMax), (yMin, yMax) = range
        xMin, xMax = _convToDtp(x.dtype, xMin, xMax)
        yMin, yMax = _convToDtp(y.dtype, yMin, yMax)
    else:
        pda_xMin, pda_xMax, pda_yMin, pda_yMax = x.min(), x.max(), y.min(), y.max()
        if (
            isinstance(pda_xMin, get_args(numpy_scalars))
            and isinstance(pda_xMax, get_args(numpy_scalars))
            and isinstance(pda_yMin, get_args(numpy_scalars))
            and isinstance(pda_yMax, get_args(numpy_scalars))
        ):
            xMin, xMax, yMin, yMax = pda_xMin, pda_xMax, pda_yMin, pda_yMax
        else:
            raise TypeError("Some message...")
```

arkouda/numpy/_numeric.py
```diff
@@ -1996,17 +2064,31 @@ def histogramdd(
     if any(b < 1 for b in bins):
         raise ValueError("bins must be 1 or greater")
 
+    if not range:
+        range = [None for pda in sample]
```
ajpotts, line 2068: Right here you allow range to contain the value `None`.

arkouda/numpy/_numeric.py
```diff
@@ -1996,17 +2064,31 @@ def histogramdd(
...
+        else:
+            return (sampleDim.min(), sampleDim.max())
+    range = [convDim(sample[i], range[i]) for i in _pyrange(num_dims)]
+
     bins = list(bins) if isinstance(bins, tuple) else bins
     sample = list(sample) if isinstance(sample, tuple) else sample
-    bin_boundaries = [linspace(a.min(), a.max(), b + 1) for a, b in zip(sample, bins)]
-    bins_pda = array(bins)[::-1]
-    dim_prod = (cumprod(bins_pda) // bins_pda)[::-1]
+    bin_boundaries = [linspace(r[0], r[1], b + 1) for r, b in zip(range, bins)]
```
ajpotts, line 2081: But down here, the value `None` is not indexable as `r[0]` and `r[1]`.

arkouda/numpy/_numeric.py
vasslitvinov: I resolved this by introducing a different variable,  `range2`, instead of reusing `range` for different things.

BTW it turns out the same trick works in the other case with `xMin` et al. so I applied it there as well by introducing `xMin0` et al.

Makefile
```diff
@@ -1,5 +1,6 @@
 # Makefile for Arkouda
 ARKOUDA_PROJECT_DIR := $(dir $(realpath $(firstword $(MAKEFILE_LIST))))
+ARKOUDA_PROJECT_DIR := $(patsubst %/,%,$(ARKOUDA_PROJECT_DIR))
```
ajpotts, line 3: Why is `ARKOUDA_PROJECT_DIR` defined twice?

vasslitvinov: The second definition removes the trailing `/` that the first definition produces. This avoids double-slashes in many cases where ARKOUDA_PROJECT_DIR is used.

arkouda/numpy/_numeric.py
```diff
@@ -1772,8 +1772,40 @@ def where(
...
 
 
+# histogram helper
+def _convToDtp(dtp, minVal, maxVal):
+    """
+    Convert minVal and maxVal to dtp because Chapel's histogram code
+    expects min/max of the same type as array elements.
+
+    Given an integer array and a floating-point bound, allow the margin of 0.3
+    when converting the bound to an integer.
```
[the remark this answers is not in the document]
vasslitvinov: 0.3 is "not too small, not too big". I am open to suggestions.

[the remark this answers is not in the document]
e-kayrakli: Can't we truncate, or do round up/down here?

If we need to have a bound, we can make it an argument with a default value.

src/Histogram.chpl
```diff
@@ -14,6 +11,27 @@ module Histogram
     private config const logChannel = ServerConfig.logChannel;
     const hgLogger = new Logger(logLevel, logChannel);
 
+    /* Helper: is this value within the requested range? */
+    inline proc histValWithinRange(val, aMin, aMax): bool {
+      return aMin <= val && val <= aMax;
+    }
+
+    /* Helper: computes the 1-d bin number for the value, assuing it is within the range. */
```
e-kayrakli, line 19: assuming. typo

src/Histogram.chpl
```diff
@@ -44,12 +62,10 @@ module Histogram
 
         // count into atomic histogram
         forall v in a {
-            var vBin = ((v - aMin) / binWidth):int;
-            if v == aMax {vBin = bins-1;}
-            if (vBin < 0) || (vBin > (bins-1)) {
-                try! hgLogger.error(getModuleName(),getRoutineName(),getLineNumber(),"OOB");
-            }
```
e-kayrakli, line 51: Did we drop this OOB check?

src/Histogram.chpl
```diff
@@ -68,14 +84,10 @@ module Histogram
 
         // count into atomic histogram
         forall (xi, yi) in zip(x, y) {
-            var xiBin = ((xi - xMin) / xBinWidth):int;
-            var yiBin = ((yi - yMin) / yBinWidth):int;
-            if xi == xMax {xiBin = numXBins-1;}
-            if yi == yMax {yiBin = numYBins-1;}
-            if xiBin < 0 || yiBin < 0 || (xiBin > (numXBins-1)) || (yiBin > (numYBins-1)) {
-                try! hgLogger.error(getModuleName(),getRoutineName(),getLineNumber(),"OOB");
-            }
```
e-kayrakli, line 77: ditto

src/Histogram.chpl
vasslitvinov: Fixed. Thanks!

src/Histogram.chpl
vasslitvinov: Yes, same answer.

e-kayrakli (approved):

Makefile
ajpotts: OK, but then shouldn't the first line be removed since it is overwritten?

vasslitvinov: The second line uses the value set in the first line.

The following page contains a similar pattern: https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html

arkouda/numpy/_numeric.py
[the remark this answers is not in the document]
vasslitvinov: Sure, I can add `float_to_int_margin` as an argument to `histogram()` and variants with a default value of 0.3, for example. Making "margin" 0 will make it round the `min` bound up and the `max` bound down.

Ideally we would do what `numpy` does. Implementation would be involved though. Consider, for example:
```python
np.histogram(a, range=(1.1,4.1), bins=4)
```
The corresponding bin boundaries are:
```
[1.1 , 1.85, 2.6 , 3.35, 4.1 ]
```
The rounding counterpart for an array of integers is:
```python
np.histogram(a, range=(2,4), bins=4)
```
The corresponding bin boundaries are:
```
[2. , 2.5, 3. , 3.5, 4. ]
```
The first bin will count differently than in the first case.

To implement numpy semantics faithfully we would probably need to handle the case of int-arrays floating-bounds specially.

What do you think, Amanda and Engin?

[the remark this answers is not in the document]
e-kayrakli: > The first bin will count differently than in the first case.

What do you mean? By just looking at your examples, the logic seems to be the same to me: bin bounds are just equally separated based on the length of the range and number of bins.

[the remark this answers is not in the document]
e-kayrakli: Oh, I see. That helps, thanks!

> or to make the bounds always be floating-point, however this will increase complexity or code size or execution speed.

How so? It looks like `toScalar` just casts things in Chapel. We can always cast bounds to `real` when we are operating on those. Could you elaborate more how it can increase complexity or hurt performance? Are you worried about integer vs float arithmetic performance? If so, I wouldn't be considering all the other operations involved in computing a histogram.

I would be worried about code complexity and/or compilation performance if you have a reason to believe that either of those could happen, though.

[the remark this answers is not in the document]
vasslitvinov: Making the bounds always `real` is the easiest. My only concern here is that comparing `int` elements against `float` bounds would slow down building the histogram. If this is not an issue, I can implement this?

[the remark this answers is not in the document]
e-kayrakli: My intuition (certainly up for debate) is that atomic fetchAdds, or synchronization in other methods of histogramming would outweigh the difference between int-to-int and int-to-float comparison. Also based on the assumption that there is no big push for squeezing every cycle out of the hardware while histogramming, I would prefer the simpler implementation that always-floating bin boundaries would bring.

@ajpotts -- do you have any preferences here?

ajpotts (approved):

vasslitvinov:
Thank you @jaketrookman for the approval. I am working to implement the last open item for this PR, which is to use `float` when passing the range parameter to Chapel in all cases, even when operating on integer data. This is to implement numpy-like behavior. I will ping you once it is ready.

vasslitvinov:
@ajpotts or @jaketrookman - I have made all the changed I wanted to. Could you please review? These are in the last two commits.

I will rebase everything down to a single commit before merging.

registration-config.json
```diff
@@ -1,7 +1,7 @@
 {
     "parameter_classes": {
         "array": {
-            "nd": [1],
+            "nd": [1,2,3],
```
ajpotts, line 4: Please revert this change to registration-config.json

ajpotts:
Looks good other than the accidental changes to `registration-config.json`.

vasslitvinov:
FYI: I just closed and re-opened to restart the CI checks. The failure reported previously, due to failed fetch of https://sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz, seems transient.

ajpotts (approved):

PR #4087, opened 2025-02-07
Fix generic actual bug with `registry/register_commands.py`

jabraham17:
Fixes an issue where a non-generic message handler function was being treated as generic by the compiler.

This fixes the following case
```python
@arkouda.registerCommand
proc MyFunc() { }
```

Since this function is not generic, the `register_commands.py` script was just calling `registerFunction` on `ark_reg_MyFunc`. This is fine, but the generated `ark_reg_MyFunc` was generic due to missing class management. This PR rectifies that bug in `register_commands.py`, by always specifying the class management

### DIFF
--- a/src/registry/register_commands.py
+++ b/src/registry/register_commands.py
@@ -33,14 +33,23 @@ chapel_scalar_types = {
     "bigint": "bigint",
 }
 
+class formalKind(Enum):
+    ARRAY = 1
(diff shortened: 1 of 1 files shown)


e-kayrakli (approved):

ajpotts (approved):
Thanks for this!

PR #4095, opened 2025-02-12
Closes-4028, adding multi-dim testing to pdarraycreation

### LINKED ISSUE #4028: Increase multidimensional testing for pdarraycreation module
Increase multidimensional testing for pdarraycreationmodule

drculhane:
Closes #4028 

This adds multi-dimensional tests for most, but not all, of the functions in **pdarraycreation_test.py**. This gets a bit involved, so here are some comments.

I made a function **multi_dim_ranks()** (to return the same thing as **ak.client.get_array_ranks()**, but without the number 1) , so that the multi-dim tests could easily iterate over just the multi-dimensional cases.

I made a TODO that I'd like eventually to fold the 1D and multi-dim tests together.

Some of the functions tested either only involve 1D arrays, or the tests have nothing to do with dimensionality, so I chose not to make multi-dimensional versions. Those are:

**test_array_creation_transpose_bug_reproducer** (note: existing transpose function is getting a new issue)
all **arange** functions (arange, like its numpy counterpart, only returns 1D arrays)
**test_randint_misc** and **test_randint_with_seed** (these tests have nothing to do with dimensionality)
**test_zeros_misc** (similarly, these tests have nothing to do with dimensionality)
similarly **test_ones_misc** and **test_full_misc**
all **linspace, normal** and **uniform** functions (again, these only return 1D arrays)
**from_series** (also only 1D)
**test_bigint_from_uint_arrays** (also only returns 1D arrays)
**test_uint_greediness** (doesn't involve dimensionality)

Conversely, these functions ONLY have multi-dim testing:

**test_infer_shape_from_size** (well, mostly, since it does accept scalars)
**test_zeros_match_numpy** (also **ones, full**) -- these had been written to only test the 2D case, but I extended them
**test_zeros_error** (also **ones, full**) -- these attempt to generate errors by making an array bigger than the compiled ranks

These were anomalies:

**test_flatten**, in pdarrayclass_test.py, was defined twice. I inferred that the second was meant as multi_dim, and renamed it.
the function **randint_randomness** exists, but is not run as a test because its name doesn't start with "test_". There is a comment to that affect in the function. I left this as is.
the function **testTO_ndarray** was also not being run, but I chose to rename it so that it would, and also made a multi-dim version.

### DIFF
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -26,6 +26,12 @@ DTYPES = [
     ak.bigint,
 ]
 
+#   multi_dim_ranks is used in multi_dim_testing
+
+def multi_dim_ranks() :
+    ranks = ak.client.get_array_ranks()[:]
+    ranks.remove(1)
+    return ranks
 
 class TestPdarrayCreation:
     @pytest.mark.parametrize("dtype", DTYPES)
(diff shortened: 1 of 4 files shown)


tests/pdarray_creation_test.py
```diff
@@ -262,6 +335,20 @@ def test_randint_num_dtype(self, dtype):
...
+    def test_randint_num_dtype_multi_dim(self, size, dtype):
+        from arkouda.util import _generate_test_shape
+        for rank in multi_dim_ranks() :
+            shape, local_size = _generate_test_shape(rank, size)
+            for test_array in ak.randint(dtype(0), 100, shape), ak.randint(0, dtype(100), shape) :
+                assert isinstance(test_array, ak.pdarray)
+                assert local_size == len(test_array)
+                assert ak.int64 == test_array.dtype
+                assert shape == test_array.shape
+                assert ((0 <= test_array) & (test_array <= 1000)).all()
```
1RyanK, line 350: Should this be `test_array < 100` or is the wiggle room intentional?

tests/pdarray_creation_test.py
```diff
@@ -817,6 +959,22 @@ def test_fill(self, size, dtype):
...
+        for rank in multi_dim_ranks() :
+            shape, local_size = _generate_test_shape(rank, size)
+            ones = ak.ones(shape)
+            ones.fill(dtype(2))
+            assert (dtype(2) == ones).all()
+
+            # Test that int_scalars covers uint8, uint16, uint32
+            ones.fill(np.uint8(2))
+            ones.fill(np.uint16(2))
+            ones.fill(np.uint32(2))
```
1RyanK, line 976: I'm not sure what is being tested here. FWIW I can see that NUMERIC_SCALARS does have all the uint(2^i) for i = 3, 4, 5, 6 things.

1RyanK (approved):

tests/pdarray_creation_test.py
drculhane: I think that's unintentional wiggle room.  I'll make a note to fix that once everyone has commented.

tests/pdarray_creation_test.py
drculhane: To be honest, neither am I.  Because this test was already there, I kept it.

tests/pdarray_creation_test.py
drculhane: Yeah, not intentional.  I'll fix that.

1RyanK: I see it's also in the test above this one, would you mind fixing it there too?

drculhane: Sure thing.

tests/pdarray_creation_test.py
ajpotts: I think that should be removed.  My guess it was put in before the `dtypes` were parameterized and then not removed.

ajpotts (approved):

ajpotts:
This is a duplicate to https://github.com/Bears-R-Us/arkouda/pull/4095

PR #4100, opened 2025-02-14
Closes #4004 Test examples in arkouda.util.py

### LINKED ISSUE #4004: Test examples in arkouda.util.py
Running the script on all docstring examples

1RyanK:
Fixed up examples and some of the parameter types and return types in the file. There are some missing, but those are on functions without docstrings. Therefore this is progress on Issue 3558 but leaves some to do.

Closes #4004 Test examples in arkouda.util.py

### DIFF
--- a/arkouda/util.py
+++ b/arkouda/util.py
@@ -3,9 +3,10 @@ from __future__ import annotations
 import builtins
 import json
 from math import prod as maprod
-from typing import TYPE_CHECKING, Sequence, Tuple, Union, cast
+from typing import TYPE_CHECKING, Literal, Sequence, Tuple, TypeVar, Union, cast
 from warnings import warn
 
+import numpy as np
 from typeguard import typechecked
 
 from arkouda.categorical import Categorical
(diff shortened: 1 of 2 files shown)


arkouda/util.py
```diff
@@ -138,22 +146,24 @@ def expand(size, segs, vals):
...
-    ak.array
+    pdarray
         The inverse of the permutation array.
 
     """
-    if unique(perm).size != perm.size:
+    unique_vals = unique(perm)
+    if cast(pdarray, unique_vals).size != perm.size:
         raise ValueError("The array is not a permutation.")
     return coargsort([perm, arange(0, perm.size)])
```
ajpotts, line 168: Could you do something like this to avoid the cast, which might be inefficient and also give unpredictable results if `unique_vals` is not a `pdarray`:
```python
    unique_vals = unique(perm, return_groups=False)
    if (not isinstance(unique_vals, pdarray)) or unique_vals.size != perm.size:
        raise ValueError("The array is not a permutation.")
    return coargsort([perm, arange(0, perm.size)])

```

ajpotts (approved):
Looks good.  Thanks!

drculhane (approved):
Well done.

jaketrookman (approved):
Looks good

ajpotts (approved):

PR #4104, opened 2025-02-20
Avoid 0 in test with unsigned numbers

jabraham17:
Adjusts a test to avoid accidentally trying to convert a -1 to an unsigned number.

This test creates an array with random numbers between 0 and 100. But if a number is 0, then the result after doing `0**4-1` is a negative number, which causes exceptions when trying to cast the value to an unsigned int.

### DIFF
--- a/tests/apply_test.py
+++ b/tests/apply_test.py
@@ -66,7 +66,7 @@ class TestApply:
         for r in ak.client.get_array_ranks():
             size = int(prob_size**(1/r))
             shape = (size,) * r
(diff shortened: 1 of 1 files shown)


bradcray:
I'm popping on to advocate for this going in ASAP because it's breaking our Arkouda testing with some regularity and seems obvious / low-risk in hindsight.

bradcray (approved):

jaketrookman (approved):
Looks good

Issue #2802, opened 2023-10-03
`concatenate` and `arange` mishandling `max_bits`


pierce314159:
This reproducer shows 2 bugs, one with arange correctly identifying bigints and with `max_bits` not actually getting applied to the resulting array of the concatenation

```python
In [13]: a = ak.arange(2**200 - 1, 2**200+4)

In [14]: b = ak.arange(5, max_bits=3)

In [15]: b
Out[15]: array([0 1 2 3 4])

In [16]: a
Out[16]: array([1606938044258990275541962092341162602522202993782792835301375 1606938044258990275541962092341162602522202993782792835301376 1606938044258990275541962092341162602522202993782792835301377 1606938044258990275541962092341162602522202993782792835301378 1606938044258990275541962092341162602522202993782792835301379])

In [17]: ak.concatenate([b,a ])
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[17], line 1
----> 1 ak.concatenate([b,a ])

File /opt/homebrew/Caskroom/miniforge/base/envs/arkouda-dev/lib/python3.10/site-packages/typeguard/__init__.py:891, in typechecked.<locals>.wrapper(*args, **kwargs)
    889 memo = _CallMemo(python_func, _localns, args=args, kwargs=kwargs)
    890 check_argument_types(memo)
--> 891 retval = func(*args, **kwargs)
    892 check_return_type(retval, memo)
    894 # If a generator is returned, wrap it if its yield/send/return types can be checked

File ~/proj/arkouda/arkouda/pdarraysetops.py:331, in concatenate(arrays, ordered)
    329         dtype = a.dtype
    330     elif dtype != a.dtype:
--> 331         raise ValueError("All pdarrays must have same dtype")
    332     names.append(cast(pdarray, a).name)
    333 elif objtype == Strings.objType:

ValueError: All pdarrays must have same dtype

In [18]: b
Out[18]: array([0 1 2 3 4])

In [19]: b.dtype
Out[19]: dtype('int64')

In [20]: b = ak.arange(5, max_bits=3, dtype=ak.bigint)

In [21]: b
Out[21]: array([0 1 2 3 4])

In [22]: b.dtype
Out[22]: dtype(bigint)

In [23]: ak.concatenate([b,a ])
[...]

PR #4105, opened 2025-02-21
Closes #3329: hstack to match numpy

### LINKED ISSUE #3329: hstack to match numpy
Add hstack function to match numpy:
https://numpy.org/doc/stable/reference/generated/numpy.hstack.html

1RyanK:
Since hstack uses concatenate in numpy, I redid concatenate on server side. It seems to work with normal pdarrays but not Strings or Categorical. I made some modifications on other things that also rely on ConcatenateMsg.chpl.

I also updated vstack because I somewhat doubt that functioned properly before. I noticed that the code where the vstack and delete tests were (tests/array_manipulation_tests.py) was not included in the pytest.ini, so I added it. The tests for delete failed, but I didn't touch it, so I suspect that was ongoing prior to my adding it.

While trying to clean things up a bit, I noticed some interesting behavior. Some things rely on the concatenate with ordered=False (meaning that when concatenating two arrays, the result may have the two arrays mixed together, rather than a whole block where you have the first, followed by everything in the second). Specifically, when comparing two Categorical objects together (i.e. for equality), it would concatenate the Strings of categories and also two aranges (which were, I guess, meant to associate the categories to numbers). Basically, it seemed to rely on the fact that they would get concatenated in the same way. This seems very sketchy to me. At some point we should probably further investigate set_categories in arkouda/categorical.py and the functions it calls. It takes a shockingly long time to test equality on two Categoricals with three things, but different categories. Definitely feels like optimizations are possible. Also ConcatenateMsg.chpl could probably benefit from optimization for Strings.

Closes #3329: hstack to match numpy

### DIFF
--- a/src/ConcatenateMsg.chpl
+++ b/src/ConcatenateMsg.chpl
@@ -23,10 +23,86 @@ module ConcatenateMsg
     private config const logChannel = ServerConfig.logChannel;
     const cmLogger = new Logger(logLevel, logChannel);
 
+    use CommandMap;
+
+    // https://data-apis.org/array-api/latest/API_specification/generated/array_api.concat.html#array_api.concat
+    /* Concatenate a list of arrays together
+       to form one array
+    */
+    @arkouda.instantiateAndRegister(prefix='concatenate')
+    proc concatenateMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab, type array_dtype, param array_nd: int): MsgTuple throws {
+        param pn = Reflection.getRoutineName();
+
+        const nArrays = msgArgs["n"].toScalar(int),
+              names = msgArgs["names"].toScalarArray(string, nArrays),
+              axis = msgArgs["axis"].getPositiveIntValue(array_nd),
+              offsets = msgArgs["offsets"].toScalarArray(int, nArrays);
+
+        var dtype: DType;
+        dtype = getGenericTypedArrayEntry(names[0], st).dtype;
+        var same: bool = true;
+        for n in names {
+          if getGenericTypedArrayEntry(n, st).dtype != dtype {
+            same = false;
+          }
+        }
+
+        if !same {
+          const errMsg = "All arrays must have the same data type.";
+          cmLogger.error(getModuleName(), pn, getLineNumber(), errMsg);
+          return MsgTuple.error(errMsg);
+        }
+
+        // Retrieve the arrays from the symbol table
+        const eIns = for n in names do st[n]: borrowed SymEntry(array_dtype, array_nd),
+              shapes = [i in 0..<nArrays] eIns[i].tupShape,
+              (valid, shapeOut) = concatenatedShape(shapes, axis, array_nd);
+        var eOut = createSymEntry((...shapeOut), array_dtype);
+
+        if !valid {
+            const errMsg = "All arrays must have the same shape except in the concatenation axis.";
+            cmLogger.error(getModuleName(), pn, getLineNumber(), errMsg);
+            return MsgTuple.error(errMsg);
+        } else {
+            for (arrIdx, arr) in zip(eIns.domain, eIns) {
+                const offset = offsets[arrIdx];
+                const arrDomain = arr.a.domain;
+                var translation: array_nd * int;
+                translation[axis] = offset;
+                const domainTranslated = arrDomain.translate(translation);
+                eOut.a[domainTranslated] = arr.a;
+            }
+            return st.insert(eOut);
+        }
+    }
+
+    // Function to validate shapes and determine output shape
+    private proc concatenatedShape(shapes: [?d] ?t, axis: int, param N: int): (bool, N*int)
+        where isHomogeneousTuple(t)
+    {
+        var shapeOut: N*int,
+            firstShape = shapes[0];
+
+        // Ensure all shapes match except for the concatenation axis
+        for i in 1..d.last {
+            for param j in 0..<N {
+                if j != axis && shapes[i][j] != firstShape[j] {
+                    return (false, shapeOut);
+                }
+            }
+        }
+
+        // Compute output shape
+        shapeOut = firstShape;
+        shapeOut[axis] = + reduce [i in 0..<d.size] shapes[i][axis]; // Sum sizes along axis
+        
+        return (true, shapeOut);
+    }
+
     /* Concatenate a list of arrays together
        to form one array
      */
-    proc concatenateMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab) : MsgTuple throws {
+    proc concatenateStrMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab) : MsgTuple throws {
         param pn = Reflection.getRoutineName();
         var repMsg: string;
         var objtype = msgArgs.getValueOf("objType").toUpper(): ObjType;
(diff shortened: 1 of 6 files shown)


1RyanK:
I keep getting `arkouda/pdarraysetops.py:427: error: Cannot determine type of "shape"  [has-type]` approximately on the line `shape1 = arrays[i - 1].shape`. Wherever the first reference is to `.shape` mypy decides that it doesn't know what type that is, and that that's a problem. However, there is no problem in hstack's version. I tried printing out some info about the shape and got this:
```
(3,)
<class 'tuple'>
<class 'int'>
```
Note that I printed out the shape, type(shape), and then type(shape[0]).
I would greatly appreciate help resolving this and then after I add some tests for hstack and vstack it should be all set to be merged in.
Another thing I wasn't sure about came from here: https://github.com/Bears-R-Us/arkouda/issues/2802
Should I set `max_bits` to the minimum or maximum of the positive max_bits across the input arrays? In the example there's only one. I currently have it set to minimum, which would throw away bits. Kind of like in significant figures you don't pretend you have precision that you don't actually have.

src/ConcatenateMsg.chpl
```diff
@@ -23,10 +23,96 @@ module ConcatenateMsg
...
+                    const srcIdx = if array_nd == 1 then (idx,) else idx;
+                    const mapped = imap(srcIdx);
+                    const targetIdx = if array_nd == 1 then mapped(0) else mapped;
+                    agg.copy(eOut.a[targetIdx], arr.a[idx]);
+                    //if array_nd == 1{
+                    //    agg.copy(eOut.a[(imap(if array_nd == 1 then (idx,) else idx))(0)], arr.a[idx]);
+                    //}
+                    //else{
+                    //    agg.copy(eOut.a[imap(if array_nd == 1 then (idx,) else idx)], arr.a[idx]);
+                    //}
```
ajpotts, line 67: Probably should delete this commented out lines.

1RyanK: Done!

src/ConcatenateMsg.chpl
```diff
@@ -23,10 +23,90 @@ module ConcatenateMsg
...
+
+    // Record to map indices for concatenation
+    record concatIndexMapper {
+        param ndOut: int;
+        const axis: int;
+        const offset: int;
+
+        proc this(idx: ndOut*int): ndOut*int {
+            var ret: ndOut*int = idx;
+            //for param i in 0..<ndOut do ret[i] = idx[i];
```
ajpotts, line 76: I think this commented out line can be deleted.

1RyanK: Done!

arkouda/pdarraysetops.py
```diff
@@ -407,20 +405,55 @@ def concatenate(
...
+            m_bits = -1 if len(max_bit_list) == 0 else min(max_bit_list)
+            for a in arrays:
+                if a.dtype == bigint and isinstance(a, pdarray):
+                    a.max_bits = m_bits
+        offsets = [0 for _ in range(len(arrays))]
+        for i in range(1, len(arrays)):
+            prev_arr = arrays[i - 1]
+            if isinstance(prev_arr, pdarray):
+                shape1 = prev_arr.shape
+                # shape = cast(Tuple[int, ...], arrays[i - 1].shape)
```
ajpotts, line 426: This commented out line can be deleted.

1RyanK: Done!

tests/array_manipulation_tests.py
```diff
@@ -14,10 +15,135 @@ def test_vstack(self):
         n_vstack = np.vstack(n)
         a_vstack = ak.vstack(a)
 
-        assert n_vstack.tolist() == a_vstack.to_list()
+        assert_arkouda_array_equivalent(n_vstack, a_vstack)
+
+    @pytest.mark.skip_if_rank_not_compiled([2])
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [int, ak.int64, ak.uint64, float, ak.float64])
```
ajpotts, line 22: I'm not seeing `bigint`s being tested anywhere in this module.

arkouda/pdarraymanipulation.py
```diff
@@ -51,23 +70,173 @@ def vstack(
...
+        if a.ndim != ndim:
+            raise ValueError("all input arrays must have the same number of dimensions")
+
+    has_bigint = False
+    for a in arrays:
+        if a.dtype == bigint:
+            has_bigint = True
+            # Should this be min or max?
+            try:
+                m_bits = min([a.max_bits for a in arrays if a.dtype == bigint and a.max_bits > 0])
```
ajpotts, line 206: I think the server crash has something to do with this area of the code. 

Also, there is a loop over all a `in arrays` inside a loop over `a in arrays`.

src/ConcatenateMsg.chpl
```diff
@@ -23,10 +23,88 @@ module ConcatenateMsg
...
+                const imap = new concatIndexMapper(array_nd, axis, offset);
+                forall idx in arr.a.domain with (
+                    var agg = newDstAggregator(array_dtype)
+                ){
+                    const srcIdx = if array_nd == 1 then (idx,) else idx;
+                    const mapped = imap(srcIdx);
+                    const targetIdx = if array_nd == 1 then mapped(0) else mapped;
+                    agg.copy(eOut.a[targetIdx], arr.a[idx]);
+                }
+            }
```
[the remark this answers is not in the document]
1RyanK: I think I implemented this - does this look correct now?

src/ConcatenateMsg.chpl
```diff
@@ -23,10 +23,88 @@ module ConcatenateMsg
...
+
+        const nArrays = msgArgs["n"].toScalar(int),
+              names = msgArgs["names"].toScalarArray(string, nArrays),
+              axis = msgArgs["axis"].getPositiveIntValue(array_nd),
+              offsets = msgArgs["offsets"].toScalarArray(int, nArrays);
+
+        // Retrieve the arrays from the symbol table
+        const eIns = for n in names do st[n]: borrowed SymEntry(array_dtype, array_nd),
+              shapes = [i in 0..<nArrays] eIns[i].tupShape,
+              (valid, shapeOut) = concatenatedShape(shapes, axis, array_nd);
```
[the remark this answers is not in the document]
1RyanK: I tried to do this like
```
        var dtype: DType;
        dtype = st[names[0]].dtype;
        var same: bool = true;
        for n in names {
          if st[n].dtype != dtype {
            same = false;
          }
        }

        if !same {
          const errMsg = "All arrays must have the same data type.";
          cmLogger.error(getModuleName(), pn, getLineNumber(), errMsg);
          return MsgTuple.error(errMsg);
        }

        // Retrieve the arrays from the symbol table
        const eIns = for n in names do st[n]: borrowed SymEntry(array_dtype, array_nd),
              shapes = [i in 0..<nArrays] eIns[i].tupShape,
              (valid, shapeOut) = concatenatedShape(shapes, axis, array_nd);
```
But apparently the entries in st are just seen as AbstracSymEntries, so they don't have .dtype. But if I do it after converting them to SymEntry, wouldn't array_dtype just overwrite the dtype? Or is there another way to check the dtypes are the same?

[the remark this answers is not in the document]
1RyanK: Is what I have now correct?

ajpotts (approved):

src/ConcatenateMsg.chpl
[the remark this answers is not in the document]
e-kayrakli: Yep, this looks right to me. Thanks!

src/ConcatenateMsg.chpl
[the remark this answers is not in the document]
e-kayrakli: Yes -- the logic is what I have in mind. As long as your testing looks OK too, this looks great. Thanks for addressing this comment!

jaketrookman (approved):
Looks good

PR #4108, opened 2025-02-21
Closes 4028, addresses comments

drculhane:
This is a revised version of the Draft Closes-4028 from last week, addressing some comments.  Some were lost, though, when the draft PR was apparently closed, so there may be more revisions needed.

### DIFF
--- a/tests/pdarray_creation_test.py
+++ b/tests/pdarray_creation_test.py
@@ -26,6 +26,14 @@ DTYPES = [
     ak.bigint,
 ]
 
+#   multi_dim_ranks is used in multi_dim_testing
+
+
+def multi_dim_ranks():
+    ranks = ak.client.get_array_ranks()[:]
+    ranks.remove(1)
+    return ranks
+
 
 class TestPdarrayCreation:
     @pytest.mark.parametrize("dtype", DTYPES)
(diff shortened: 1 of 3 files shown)


tests/pdarray_creation_test.py
```diff
@@ -840,10 +981,22 @@ def test_fill(self, size, dtype):
...
+        for rank in multi_dim_ranks():
+            shape, local_size = _generate_test_shape(rank, size)
+            ones = ak.ones(shape)
+            ones.fill(dtype(2))
+            assert (dtype(2) == ones).all()
+
+            # Test that int_scalars covers uint8, uint16, uint32
+            ones.fill(np.uint8(2))
+            ones.fill(np.uint16(2))
+            ones.fill(np.uint32(2))
```
ajpotts, line 999: I think that should be removed.  My guess it was put in before the `dtypes` were parameterized and then not removed.

ajpotts (approved):

tests/pdarray_creation_test.py
drculhane: Yep.  Sorry.  Oversight.  Removed.

ajpotts: Thanks!

ajpotts (approved):

PR #4127, opened 2025-02-28
Add `xp.trapz`

ShreyasKhandekar:
Resolves: https://github.com/Bears-R-Us/arkouda/issues/4129

Add `trapz` (or `trapezoid`) function in Arkouda using the functionality from the array API.
   
In the future the entirity of this function may be implemented purely in Chapel.

See https://numpy.org/doc/1.26/reference/generated/numpy.trapz.html#numpy.trapz for reference.

Technically, `trapz` is deprecated in numpy `2.0` onwards, but Arkouda still uses numpy `1.26`. I've added both overloads. 

Also added tests.

### DIFF
--- a/tests/array_api/util_functions.py
+++ b/tests/array_api/util_functions.py
@@ -145,3 +145,82 @@ class TestUtilFunctions:
         ):
             xp.pad(a, ((1, 1)), mode="constant", constant_values=((-1, 1)))
             xp.diff(a, n=2, axis=0)
+
+    @pytest.mark.parametrize("dtype", DTYPES_WITH_BOOL)
+    def test_trapz_1D(self, dtype):
+        # 1D array test
+        x = randArr((10), dtype)
+        y = randArr((10), dtype)
+        x_np = x.to_ndarray()
+        y_np = y.to_ndarray()
+
+        # Single array version, default dx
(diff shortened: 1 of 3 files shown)


tests/array_api/util_functions.py
```diff
@@ -148,3 +149,81 @@ def test_pad_error(self):
         ):
             xp.pad(a, ((1, 1)), mode="constant", constant_values=((-1, 1)))
             xp.diff(a, n=2, axis=0)
+
+    @pytest.mark.parametrize("dtype", DTYPES_WITH_BOOL)
```
e-kayrakli, line 153: Mostly out of curiosity, is there a special meaning to doing `trapz` with a `bool` array? Other tests, didn't do anything with bool arrays so far, it seems. Why `trapz`?

e-kayrakli (approved):

tests/array_api/util_functions.py
ShreyasKhandekar: It's valid to pass a `bool` array to `trapz` and we should make sure the behavior is correct :) 
I did also add the same test case for `diff` as well.

ajpotts (approved):

Issue #4126, opened 2025-02-28
`xp.diff` is broken when called without an `axis` argument


ShreyasKhandekar:
**Describe the bug**
`xp` refers to the `arkouda.array_api`

When using `xp.diff` I saw that it uses a default `axis` argument of `-1` and passes it directly to the server.
`-1` is not a valid axis in Chapel, therefore it results in the function silently returning the wrong answer.

**To Reproduce**

```py
>>> a = ak.randint(0, 100, 12)
>>> a = xp.asarray(a)
>>> print(xp.diff(a))
[0 0 0 0 0 0 0 0 0 0 0 0]
```

**Expected behavior**
It should give the 1st order differential between elements of `a`.
Instead we get all `0`s.

**Error Message**
Silently returns the wrong answer, no error provided.

**Is this a Blocking Issue**
- Are there any known work-arounds? Yes, just do `a_.ndim+axis` when `axis < 0` to fix this
- Same as https://github.com/Bears-R-Us/arkouda/issues/4125

**Additional context**
This is a worse case of https://github.com/Bears-R-Us/arkouda/issues/4125 because this specific function misbehaves even with its default arguments.

PR #4128, opened 2025-02-28
Fix `xp.diff`

ShreyasKhandekar:
This resolves https://github.com/Bears-R-Us/arkouda/issues/4126, adds better documentation for the function, and adds more testing.

### DIFF
--- a/tests/array_api/util_functions.py
+++ b/tests/array_api/util_functions.py
@@ -8,7 +8,7 @@ SEED = 314159
 s = SEED
 
 DTYPES = [ak.int64, ak.float64, ak.uint64, ak.uint8]
-
+DTYPES_WITH_BOOL = [ak.int64, ak.float64, ak.uint64, ak.uint8, ak.bool_]
 
 def randArr(shape, dtype):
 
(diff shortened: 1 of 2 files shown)


tests/array_api/util_functions.py
```diff
@@ -69,12 +70,16 @@ def test_clip_errors(self):
         ):
             xp.clip(a, 10, 90)
 
-    @pytest.mark.parametrize("dtype", DTYPES)
+    @pytest.mark.parametrize("dtype", DTYPES_WITH_BOOL)
```
e-kayrakli, line 73: Continuing from the other PR, maybe you are just adding bool dtype to everything gradually :)

e-kayrakli (approved):

tests/array_api/util_functions.py
ShreyasKhandekar: Yes, `diff` specifically states in it's [documentation](https://numpy.org/doc/1.26/reference/generated/numpy.diff.html#numpy.diff) what the behavior should be for `bool` arrays. So I added it to the test to make sure we have that same behavior as `numpy`.

e-kayrakli: Yep. Thanks for covering that case! (Likewise in the other PR)

ShreyasKhandekar: Well it turns out I didn't actually do this correctly. I missed the check that disallows bool. I've fixed that and also removed the test case which errors out when a bool array is passed in.

ajpotts (approved):
Thanks!

PR #4146, opened 2025-03-05
Closes #3000: repeat function

### LINKED ISSUE #3000: repeat function
Replicate functionality of numpy.repeat:
https://numpy.org/doc/stable/reference/generated/numpy.repeat.html

1RyanK:
Closes #3000: repeat function

### DIFF
--- a/src/ManipulationMsg.chpl
+++ b/src/ManipulationMsg.chpl
@@ -9,6 +9,7 @@ module ManipulationMsg {
   use ServerErrorStrings;
   use CommAggregation;
   use AryUtil;
+  use BigInteger;
 
   use Reflection;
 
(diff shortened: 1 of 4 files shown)


src/ManipulationMsg.chpl
```diff
@@ -812,8 +813,58 @@ module ManipulationMsg {
     return shapeOut;
   }
 
-  // // https://data-apis.org/array-api/latest/API_specification/generated/array_api.unstack.html
-  // // unstack an array into multiple arrays along a specified axis
+  // https://data-apis.org/array-api/latest/API_specification/generated/array_api.repeat.html#array_api.repeat
+  @arkouda.registerCommand (name="repeat")
+  proc repeat (eIn : [?d] ?t, axis : int, reps : [?d2] ?t2 ) : [] t throws 
+    where ((t2 == int) &&
+           (t == int || t == real || t == bool || t == uint || t == uint(8) || t == bigint) &&
```
e-kayrakli, line 820: - `t2` could probably be just checked for `integral`. Or, if we stick to it, you can drop `?t2` from the formal argument, and just make it `int`
- Mild preference for typechecking `t` in a function:

```chpl
proc repeatInputElemTypeCheck(type t) param : bool {
  return <your check>;
}
```

src/ManipulationMsg.chpl
```diff
@@ -812,8 +813,58 @@ module ManipulationMsg {
...
+           (d2.rank == 1)){
+    param pn = Reflection.getRoutineName();
+
+    var eOut = makeDistArray((...repeatShape(eIn.shape, reps, axis)), t);
+
+    if reps.size == 1{
```
e-kayrakli, line 826: - space before `{`
- But more importantly, I am guessing this is to emulate the `int` version? Can we not make `reps` generic (i.e. no type in the formal argument), type check it similarly to make sure that it is either an `int` or array of `int`s, and replace this line with something like `if reps.type == int`?

src/ManipulationMsg.chpl
```diff
@@ -812,8 +813,58 @@ module ManipulationMsg {
...
+      forall idx in eOut.domain with (
+        var agg = newSrcAggregator(t)
+      ) {
+        var idx2 = if eIn.shape.size == 1 then (idx,) else idx;
+        idx2[axis] = idx2[axis] / reps[0];
+        agg.copy(eOut[idx], eIn[if eIn.shape.size == 1 then idx2[0] else idx2]);
+      }
+    }else{
+      var lookup: [0..<eOut.shape[axis]] int;
```
e-kayrakli, line 835: At least for 1D, we shouldn't need this. Probably the idea can be extended mathematically to multiple dimensions, but here's a quick take for 1D:

```chpl
var InArr = [1,2,3,4,5];
var Repeats = [2,6,1,5,6];

const outSize = + reduce Repeats;
var OutStarts: [InArr.domain] int;

OutStarts[1..] = + scan Repeats[..Repeats.size-2];

var OutArr: [0..#outSize] InArr.eltType;

// should use destination aggregation for better performance
forall (inElem, outStart, numRepeats) in zip(InArr, OutStarts, Repeats) {
  for destIdx in outStart..#numRepeats {
    OutArr[destIdx] = inElem;
  }
}

writeln(OutArr);
```

We could at least add a TODO comment here to specialize for 1D.

src/ManipulationMsg.chpl
```diff
@@ -812,8 +813,58 @@ module ManipulationMsg {
...
+        var inIdx = if eIn.shape.size == 1 then (idx,) else idx;
+        inIdx[axis] = lookup[inIdx[axis]];
+        agg.copy(eOut[idx], eIn[if eIn.shape.size == 1 then inIdx[0] else inIdx]);
+      }
+    }
+
+    return eOut;
+  }
+
+  proc repeatShape(shape: ?N*int, reps: [] int, axis: int): N*int {
```
e-kayrakli, line 855: this could be made generic on `reps`, as well.

src/ManipulationMsg.chpl
1RyanK: I tried putting in the typechecking function, and it's telling me `Error  evaluating where-clause :  general function calls not yet supported in where-clauses; ignoring function: repeatInputElemTypeCheck` when compiling. It doesn't seem to affect running; will this be fixed moving forward? I agree it's much cleaner looking, though.

src/ManipulationMsg.chpl
1RyanK: I wasn't sure how to do type checking for array type, but also it's convenient to have it as an array because then in the where clause I can check that the number of dimensions is 1. Should I get rid of that and check it in the body?

src/ManipulationMsg.chpl
e-kayrakli: Ah, yeah, this must be https://github.com/Bears-R-Us/arkouda/issues/4162. You can ignore that comment for the time being.

src/ManipulationMsg.chpl
e-kayrakli: > I wasn't sure how to do type checking for array type

`isArray(reps)` should do it. That function is in the `Types` standard module.

> but also it's convenient to have it as an array because then in the where clause I can check that the number of dimensions is 1

I am not sure if I am following. `reps` probably has to have the same rank (number of dimensions) as the input array, right? But in any case, `reps.rank` will give you the number of dimensions of the array

1RyanK: `reps` has to have the same number of elements as the input array in the given axis. I'll give that a shot, though! Thanks for the help.

src/ManipulationMsg.chpl
1RyanK: I think I implemented this and it seems to work! Thanks!

src/ManipulationMsg.chpl
e-kayrakli: Ugh.. I am sorry I led you down this path. You are not doing anything wrong from Chapel sense, but you are bumping into a function registration limitation it looks like. That's a new feature in Arkouda and can use some improvements. I recorded this in https://github.com/Bears-R-Us/arkouda/issues/4164. I am talking with a teammate to see if we can make some more improvements here. For now, you can ignore this comment. Sorry again.

src/ManipulationMsg.chpl
e-kayrakli: Great, thanks!

ajpotts (approved):
Great work!

jaketrookman (approved):
Looks good to me

PR #4139, opened 2025-03-03
Runs isort and black on project

ajpotts:
Runs `isort` and `black` on project.

Closes #4166

PR #4150, opened 2025-03-05
Closes #4147:  parse_single_value to handle uints represented as nega…

### LINKED ISSUE #4147: parse_single_value to handle uints represented as negative numbers
A `uint` may be sent back to the client as a negative number.  However, numpy 2.0 no longer natively supports converting a negative number to a `uint`, so arkouda must handle that case.

ajpotts:
Numpy 2.0 does not support conversion of negative numbers to `uint64`.  For example, np.uint64("-2") will not work.  Therefore arkouda needs to update the code for the `parse_single_value` to handle the case when the server sends a uint as a string representing a negative number.

Closes #4147:  parse_single_value to handle uints represented as negative numbers

### DIFF
--- a/arkouda/numpy/pdarrayclass.py
+++ b/arkouda/numpy/pdarrayclass.py
@@ -11,6 +11,7 @@ import numpy as np
 from typeguard import typechecked
 
 from arkouda.client import generic_msg
+from arkouda.dtypes import uint64 as ak_uint64
 from arkouda.infoclass import information, pretty_print_information
 from arkouda.logger import getArkoudaLogger
 from arkouda.numpy.dtypes import (
@@ -185,6 +186,16 @@ def parse_single_value(msg: str) -> Union[numpy_scalars, int]:
         if mydtype == akstr_:
             # String value will always be surrounded with double quotes, so remove them
             return mydtype.type(unescape(value[1:-1]))
+
+        #   This is a work-around to deal with the fact that the server
+        #   sometimes returns a negative value when it should return a uint.
+        #   It can be removed when issue #4157 is resolved.
+        if mydtype == ak_uint64:
+            if value.startswith("-"):
+                value = value.strip("-")
+                uint_value = np.uint64(np.iinfo(np.uint64).max) - ak_uint64(value) + ak_uint64(1)
+                return uint_value
+            return ak_uint64(value)
         return mydtype.type(value)
     except Exception:
         raise ValueError(f"unsupported value from server {mydtype.name} {value}")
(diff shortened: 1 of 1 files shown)


1RyanK (approved):
Looks good!

arkouda/pdarrayclass.py
```diff
@@ -4131,31 +4146,30 @@ def fmod(dividend: Union[pdarray, numeric_scalars], divisor: Union[pdarray, nume
...
+        else:  # this condition *should* be impossible because of the isSupportedNumber check
             raise TypeError(f"Scalar dividend type {resolve_scalar_dtype(dividend)} not allowed in fmod")
         cmdstring = f"{acmd}<{divisor.dtype},{divisor.ndim}>"  # type: ignore[union-attr]
 
     else:
         m = mod(dividend, divisor)
         return _create_scalar_array(m)
 
-#   We reach here if this was any case other than scalar & scalar
+    #   We reach here if this was any case other than scalar & scalar
```
e-kayrakli, line 4172: Style adjustments will conflict with https://github.com/Bears-R-Us/arkouda/pull/4105

arkouda/pdarrayclass.py
```diff
@@ -166,6 +173,13 @@ def unescape(s):
         if mydtype == akstr_:
             # String value will always be surrounded with double quotes, so remove them
             return mydtype.type(unescape(value[1:-1]))
+
+        if mydtype == ak_uint64:
+            if value.startswith("-"):
+                value = value.strip("-")
+                uint_value = np.uint64(np.iinfo(np.uint64).max) - ak_uint64(value) + ak_uint64(1)
+                return uint_value
+            return ak_uint64(value)
```
e-kayrakli, line 182: So, we received what looks to be an `int(64)`, but we are told that it was `uint(64)`. Are we trying to get the correct binary representation of that value? This feels like a workaround to me that at least needs to be commented and removed, but I am curious to understand the server side of the story where we don't get the right type somewhere. I commented under https://github.com/Bears-R-Us/arkouda/issues/4147 to discuss.

e-kayrakli (commented):
I am hoping that we can make this PR obsolete by fixing the server. Falling short of that, a comment explaining the workaround accompanied by a link to a bug report would be great for maintainability and tracking.

arkouda/pdarrayclass.py
ajpotts: Style changes will be addressed in this one:https://github.com/Bears-R-Us/arkouda/pull/4139

Generally, we ask everyone to run black and isort before submitting a PR, which doesn't always happen unfortunately.

arkouda/pdarrayclass.py
ajpotts: Yes, this is a work-around.  Unfortunately it needs to go in because this blocking the python library upgrades.  However, I agree the better solution would be to understand why the server is returning negative values when it should be returning a `uint`.  I put in separate ticket: https://github.com/Bears-R-Us/arkouda/issues/4157 and also added a comment into the code referencing the issue number.

drculhane (approved):
Looks good.

jaketrookman (approved):
Good fix

PR #4158, opened 2025-03-07
Fix the shutdown test by using the correct server

e-kayrakli:
`test_shutdown` shuts the server down and restarts it. But it just assumes that we'll get the same node as the head node, which is not guaranteed, and almost never happens in a new system we are establishing Arkouda testing on.


ajpotts (approved):
I suspect this has caused other problems in the past.  Thanks for fixing this!

jaketrookman (approved):

PR #4176, opened 2025-03-12
Closes #4175:  remove try! in toSymEntry

### LINKED ISSUE #4175: remove try! in toSymEntry
Remove `try!` in `toSymEntry` because it kills the server instead of throwing an Error.

ajpotts:
Remove `try!` in `toSymEntry` because it kills the server instead of throwing an Error.

Closes #4175:  remove try! in toSymEntry


e-kayrakli:
I really like this change coming. Though, I will keep a close eye on performance as making `toSymEntry` a throwing function could result in much more code to be generated and executed to be able to handle errors thrown.

jaketrookman (approved):

1RyanK (approved):

Issue #4178, opened 2025-03-12
`comm_diagnostics_test` cause too much verbose comm output causing hangs


ShreyasKhandekar:
`comm_diagnostics_test` has tests for the comm diagnostics utility for the Arkouda client.
We have been seeing failures in nightly testing for this test, and looking into it more, I saw weird server hangs when running these tests. 

They are also abysmally slow to run because of all the output from `start_verbose_comm`.

I think the culprit here is the verbose comm output.
For example, here are the comm counts from run of this test that we get from `print_comm_diagnostics_table`:
```
+----+-------+-------+-------+--------------+-----------------+
|    |   put |   get |   amo |   execute_on |   execute_on_nb |
+====+=======+=======+=======+==============+=================+
|  0 | 62908 | 31324 | 93092 |        64813 |            7620 |
+----+-------+-------+-------+--------------+-----------------+
|  1 | 63042 | 32613 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  2 | 63042 | 32620 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  3 | 63042 | 32620 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  4 | 63042 | 32627 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  5 | 63040 | 32620 |     0 |        64349 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  6 | 63042 | 32634 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  7 | 63042 | 32620 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  8 | 63046 | 32620 |     0 |        64355 |               0 |
+----+-------+-------+-------+--------------+-----------------+
|  9 | 63042 | 32613 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
| 10 | 63042 | 32634 |     0 |        64351 |               0 |
+----+-------+-------+-------+--------------+-----------------+
| 11 | 63032 | 32620 |     0 |        64341 |               0 |
+----+-------+-------+-------+--------------+-----------------+
| 12 | 63046 | 32641 |     0 |        64355 |               0 |
+----+-------+-------+-------+--------------+-----------------+
[...]

PR #4179, opened 2025-03-12
Remove verbose output from comm diag test

ShreyasKhandekar:
Resolves https://github.com/Bears-R-Us/arkouda/issues/4178

Factor out verbose comm into it's own small test


jaketrookman (approved):
Looks good

PR #4181, opened 2025-03-13
Add `pytest.temp_directory`

jabraham17:
Adds `pytest.temp_directory`, which allows overriding the default directory for temporary files for IO related tests


e-kayrakli (approved):

ajpotts (approved):
Good idea.  Thanks!

PR #4183, opened 2025-03-14
Reorg modules into our Pandas Directory Part 1

jaketrookman:
This is the first round of reorganizing modules (part 1) to match the structure of Pandas.


arkouda/dataframe.py
```diff
@@ -526,6 +541,8 @@ def sample(self, n=None, frac=None, replace=False, weights=None, random_state=No
         ]
 
     def _return_agg_series(self, values, sort_index=True):
+        from arkouda.pandas.series import Series
```
1RyanK, line 544: Is this because of the circular imports you mentioned? Because a lot of functions are doing this.

1RyanK (approved):
Looks good!

arkouda/dataframe.py
jaketrookman: Yes, Amanda was a huge help with this and this was her solution to get rid of the circular import error. We can decide if this is the best way to do it or if there is a better restructuring to get it to correctly import without the circular errors

ajpotts (approved):
Thanks Jake!

PR #4192, opened 2025-03-19
Closes #4119: Fix delete in arkouda/pdarraymanipulation.py

### LINKED ISSUE #4119: Fix delete in arkouda/pdarraymanipulation.py
I added array_manipulation_tests.py to pytest.ini and the delete test failed. It shouldn't do that.

1RyanK:
I came up with two ways of deleting along an axis. One way is by copying using a source aggregator to bring over the data that isn't deleted, one index at a time. The other way does copying by whole slices. Interested in hearing people's thoughts about which way is better. I'm thinking copying by slice may be faster if the amount copied each time is large enough, meaning that there is either enough data in the other dimensions or the proportion of data deleted does not exceed some amount.

Closes #4119: Fix delete in arkouda/pdarraymanipulation.py

### DIFF
--- a/arkouda/numpy/pdarraymanipulation.py
+++ b/arkouda/numpy/pdarraymanipulation.py
@@ -1,4 +1,4 @@
-from typing import List, Literal, Optional, Tuple, Union
+from typing import List, Literal, Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
 from typeguard import typechecked
@@ -6,6 +6,7 @@ from typeguard import typechecked
 from arkouda.client import generic_msg
 from arkouda.numpy.dtypes import dtype as akdtype
 from arkouda.numpy.pdarrayclass import create_pdarray, pdarray
+from arkouda.numpy.pdarraycreation import arange, array
 
 __all__ = ["vstack", "delete"]
 
@@ -76,7 +77,7 @@ def vstack(
 @typechecked
 def delete(
     arr: pdarray,
-    obj: Union[pdarray, slice, int],
+    obj: Union[slice, int, Sequence[int], Sequence[bool], pdarray],
     axis: Optional[int] = None,
 ) -> pdarray:
     """
@@ -86,9 +87,9 @@ def delete(
     ----------
     arr : pdarray
         The array to remove elements from
-    obj : Union[pdarray, slice, int]
+    obj : slice, int, Sequence of int, Sequence of bool, or pdarray
         The indices to remove from 'arr'. If obj is a pdarray, it must
-        have an integer dtype.
+        have an integer or bool dtype.
     axis : Optional[int], optional
         The axis along which to remove elements. If None, the array will
         be flattened before removing elements. Defaults to None.
@@ -97,37 +98,41 @@ def delete(
     -------
     pdarray
         A copy of 'arr' with elements removed
+
+    Examples
+    --------
+    >>> arr = ak.array([[1,2,3,4], [5,6,7,8], [9,10,11,12]])
+    >>> arr
+    array([array([1 2 3 4]) array([5 6 7 8]) array([9 10 11 12])])
+    >>> ak.delete(arr, 1, 0)
+    array([array([1 2 3 4]) array([9 10 11 12])])
+
+    >>> ak.delete(arr, slice(0, 4, 2), 1)
+    array([array([2 4]) array([6 8]) array([10 12])])
+    >>> ak.delete(arr, [1, 3, 5], None)
+    array([1 3 5 7 8 9 10 11 12])
     """
 
-    if axis is None:
+    shape = arr.shape
+
+    if axis is None and arr.ndim != 1:
         # flatten the array if axis is None
-        _arr = create_pdarray(
-            generic_msg(
-                cmd=f"reshape{arr.ndim}Dx1D",
-                args={
-                    "name": arr,
-                    "shape": (arr.size,),
-                },
-            )
-        )
+        _arr = arr.flatten()
         _axis = 0
+        shape = _arr.shape
+    elif axis is None:
+        _axis = 0
+        _arr = arr
     else:
         _arr = arr
         _axis = axis
-
+    slice_weight = 1
+    for i in range(_axis + 1, len(shape)):
+        slice_weight *= shape[i]
     if isinstance(obj, pdarray):
-        return create_pdarray(
-            generic_msg(
-                cmd=f"delete{_arr.ndim}D",
-                args={
-                    "arr": _arr,
-                    "obj": obj,
-                    # TODO: maybe expose this as an optional argument? Or sort the array first?
-                    "obj_sorted": False,
-                    "axis": _axis,
-                },
-            )
-        )
+        _del = obj
+    elif isinstance(obj, Sequence):
+        _del = cast(pdarray, array(obj))
     else:
         if isinstance(obj, int):
             start = obj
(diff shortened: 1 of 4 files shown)


src/ManipulationMsg.chpl
```diff
@@ -232,6 +233,219 @@ module ManipulationMsg {
...
+  // This deletes by copying by index everything that isn't deleted.
+  @arkouda.registerCommand(name="delete1")
+  proc deleteAxis1Msg (eIn: [?d] ?t, axis: int, del: [?d2] ?t2): [] t throws 
+    where ((t2 == int || t2 == bool) &&
+           (t == int || t == real || t == bool || t == uint || t == uint(8) || t == bigint) &&
+           (d2.rank == 1)) {
+    param pn = Reflection.getRoutineName();
+    
+    // Need to ensure we have a boolean array of what gets deleted.
+    var del_bool = makeDistArray(eIn.shape[axis], bool);
```
e-kayrakli, line 245: In general, it'd be great if we can stick with camelCase on the server code as that's the generally recommended style, which is also largely used in the existing server code.

src/ManipulationMsg.chpl
```diff
@@ -232,6 +233,219 @@ module ManipulationMsg {
...
+      var inIdx = if eIn.shape.size == 1 then (idx,) else idx;
+      inIdx[axis] = map_array[inIdx[axis]];
+      agg.copy(eOut[idx], eIn[if eIn.shape.size == 1 then inIdx[0] else inIdx]);
+    }
+
+    return eOut;
+  }
+
+  // This deletes by copying the slices that aren't deleted.
+  @arkouda.registerCommand(name="delete2")
```
e-kayrakli, line 316: I wouldn't be against merging both implementations, but can you make the commands and the server functions have more self-documenting names? To me, the differentiating aspects are one is doing bulk copy and the other is doing aggregated copy.

e-kayrakli (approved):
I think similarly -- bulk copy using slices should be faster in this case as the data we are sloshing around is _mostly) contiguous. However, if the deleted indices create a scenario where there are a lot of jumping around in memory, aggregated copy could be a good contender, too.

src/ManipulationMsg.chpl
1RyanK: Fixed all of them!

ajpotts (approved):
Great work!

tests/numpy/array_manipulation_tests.py
```diff
@@ -2,41 +2,202 @@
...
+        n_delete = np.delete(n, ind)
+        a_delete = ak.delete(a, ind)
+        assert_arkouda_array_equivalent(n_delete, a_delete)
+
+        delete_inds = ak.randint(-a_size, a_size, max(1, dim_size // 3))
+        n_delete = np.delete(n, delete_inds.to_list())
+        a_delete = ak.delete(a, delete_inds.to_list())
+        assert_arkouda_array_equivalent(n_delete, a_delete)
+
+        delete_bounds = sorted(ak.randint(-a_size, a_size, 2).to_list())
```
ajpotts, line 45: This doesn't seem right when negative indices are allowed.  For example:
```
In [28]: delete_bounds
Out[28]: [-47, 27]
```
Had no effect.

tests/numpy/array_manipulation_tests.py
```diff
@@ -2,41 +2,202 @@
...
+        n_delete = np.delete(n, slice(low, high))
+        a_delete = ak.delete(a, slice(low, high))
+        assert_arkouda_array_equivalent(n_delete, a_delete)
+
+        delete_bools = ak.randint(0, 2, a_size, dtype=bool)
+        n_delete = np.delete(n, delete_bools.to_ndarray())
+        a_delete = ak.delete(a, delete_bools)
+        assert_arkouda_array_equivalent(n_delete, a_delete)
+
+        for i in range(1):
```
ajpotts, line 56: `for i in range(1):` will only take on a value of `i=0`.  
Would you consider parameterizing the `axis` argument as `[0,None]` and avoid the code duplication?

tests/numpy/array_manipulation_tests.py
```diff
@@ -2,41 +2,202 @@
...
         n_vstack = np.vstack(n)
         a_vstack = ak.vstack(a)
 
         assert n_vstack.tolist() == a_vstack.to_list()
+        """
+
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [int, ak.int64, ak.uint64, float, ak.float64, ak.bigint])
+    def test_delete(self, size, dtype):
+        dim_size = size
```
ajpotts, line 26: Can we just use `size` here instead of defining `dim_size`?

tests/numpy/array_manipulation_tests.py
```diff
@@ -2,41 +2,202 @@
...
+            n_delete = np.delete(n, delete_bools.to_ndarray(), axis=i)
+            a_delete = ak.delete(a, delete_bools, axis=i)
+            assert_arkouda_array_equivalent(n_delete, a_delete)
 
     @pytest.mark.skip_if_rank_not_compiled([2])
-    def test_delete(self):
-        a = ak.randint(0, 100, (10, 10))
-        n = a.to_ndarray()
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    @pytest.mark.parametrize("dtype", [int, ak.int64, ak.uint64, float, ak.float64, ak.bigint])
```
ajpotts, line 85: Similar comments to above.  Wondering if this would be a case where it would make sense to parameterize the dimension and use something like this:
```python
    if dim not in get_array_ranks():
        pytest.skip(f"Sever not compiled for rank {dim}")
```
Then the three tests could be combined.

arkouda/numpy/pdarraymanipulation.py
```diff
@@ -99,31 +100,57 @@ def delete(
         A copy of 'arr' with elements removed
```
ajpotts, line 100: Could you add an examples section?

arkouda/numpy/pdarraymanipulation.py
```diff
@@ -99,31 +100,57 @@ def delete(
...
 
-    if axis is None:
+    shape = arr.shape
+
+    if axis is None and arr.ndim != 1:
         # flatten the array if axis is None
         _arr = create_pdarray(
             generic_msg(
-                cmd=f"reshape{arr.ndim}Dx1D",
+                cmd=f"reshape<{arr.dtype},{arr.ndim},1>",
```
ajpotts, line 109: Why not use the arkouda `reshape` or `flatten` function here?

arkouda/numpy/pdarraymanipulation.py
```diff
@@ -99,31 +100,57 @@ def delete(
...
+        if obj.dtype == int and (shape[_axis] / max(obj.size, 1)) * slice_weight >= 100:
+            alg_choice = "BulkCopy"
+        else:
+            alg_choice = "AggCopy"
+        return create_pdarray(
+            generic_msg(
+                cmd=f"delete{alg_choice}<{_arr.dtype},{_arr.ndim},{obj.dtype},{obj.ndim}>",
+                args={
+                    "eIn": _arr,
+                    "del": obj,
```
ajpotts, line 137: It seems like you could use a `_del` variable and combine some of these cases to avoid a lot of code duplication.

tests/numpy/array_manipulation_tests.py
1RyanK: I will happily admit that this is somewhat bizarre, but the same thing happens with numpy. I think I will take out the sorted part of it. That may result in an empty slice in other ways, but essentially a slice is a slice first, without regards to any array. Then when you tie it to an array, negative indices are adjusted by the size of the array. So to have an effect, the appropriate way to store this would be `[27, -47]`, in which case numpy (at least) does change the array. For `[-47, 27]`, numpy also does nothing.

ajpotts: My point wasn't that it was giving the incorrect result, but rather that it isn't actually testing anything about half the time.  Is there a way to select the bounds that will for sure delete something?

1RyanK: It should test deleting something every time now. Thanks!

tests/numpy/array_manipulation_tests.py
1RyanK: This is addressed alongside the later comment about parameterizing the number of dimensions. Thanks!

tests/numpy/array_manipulation_tests.py
1RyanK: That makes sense for one dimension, but since the number of dimensions has been parameterized, they will be distinct for some tests. Thanks!

tests/numpy/array_manipulation_tests.py
1RyanK: Done!

arkouda/numpy/pdarraymanipulation.py
1RyanK: Done!

arkouda/numpy/pdarraymanipulation.py
1RyanK: I was thinking it might be a little quicker to directly call the Chapel-side function in my code, but I can see that it might be easier to have as few instances of the call itself as possible. Now I have a `flatten` call. Thanks!

arkouda/numpy/pdarraymanipulation.py
1RyanK: Done - cut down on a lot of repeated code, too. Thanks!

PR #4197, opened 2025-03-19
Restore scipy skipifs removed in #4172

jabraham17:
Restores the scipy skipifs that were removed in #4172, as they are still needed.


drculhane (approved):
My mistake.  Thank you for catching that.

ajpotts (approved):

PR #4204, opened 2025-03-24
Closes #4203: Fix max_bits on multidimensional arrays

### LINKED ISSUE #4203: Fix max_bits on multidimensional arrays
There have been issues with trying to get the number of max_bits or set the number of max_bits on multidimensional bigint arrays. Ultimately the issues seem to stem from trying to do `toSymEntry` on a `GenSymEntry` when the number of dimensions is greater than one. There are other places where this is done on other things as well (casting a multidimensional array of one type to another type), so fixing this may open avenues for solving other problems as well.

1RyanK:
For some reason, `toSymEntry` has an input `param dimensions = 1`, but no call (that I've seen) ever changes that. Trying to rework it so that it doesn't need that and will just automatically convert it to whatever dimension is in ndim (an attribute of `GenSymEntry`) seems impossible (I tried, which is maybe not saying much). It seems to need to know at compile time what kind of `SymEntry` it is returning. Engin pointed out that this is solved by `instantiateAndRegister` (or just `registerCommand` if that's all that's necessary).

This can probably be applied to many other instances where a call to `toSymEntry` exists (e.g. I saw some in casting). This may not fix all of them, but this alone should fix most of them and some further tweaking should be possible on the others.

Since fixing all the ones in `BigIntMsg.chpl`, this also fixes the issue from before where converting multidimensional bigint arrays to and from numpy caused errors without so much reshaping.

Closes #4203: Fix max_bits on multidimensional arrays

### DIFF
--- a/arkouda/numpy/pdarraycreation.py
+++ b/arkouda/numpy/pdarraycreation.py
@@ -237,14 +237,17 @@ def array(
     if (a.dtype == bigint or dtype == bigint) and a.ndim in get_array_ranks() and a.ndim > 1:
         sh = a.shape
         try:
-            a = a.reshape(a.size)
             # attempt to break bigint into multiple uint64 arrays
             # early out if we would have more uint arrays than can fit in max_bits
             early_out = (max_bits // 64) + (max_bits % 64 != 0) if max_bits != -1 else float("inf")
             while (a != 0).any() and len(uint_arrays) < early_out:
                 low, a = a % 2**64, a // 2**64
                 uint_arrays.append(array(np.array(low, dtype=np.uint), dtype=akuint64))
-            return bigint_from_uint_arrays(uint_arrays[::-1], max_bits=max_bits).reshape(sh)
+            # If uint_arrays is empty, this will create an empty ak array and reshape it.
+            if not uint_arrays:
+                return zeros(size=sh, dtype=bigint, max_bits=max_bits)
+            else:
+                return bigint_from_uint_arrays(uint_arrays[::-1], max_bits=max_bits)
         except TypeError:
             raise RuntimeError(f"Unhandled dtype {a.dtype}")
 
@@ -436,11 +439,11 @@ def bigint_from_uint_arrays(arrays, max_bits=-1):
     if not arrays:
         return create_pdarray(
             generic_msg(
-                cmd="big_int_creation",
+                cmd="big_int_creation<bigint,1>",
                 args={
                     "arrays": arrays,
                     "num_arrays": len(arrays),
-                    "len": 0,
+                    "shape": (0,),
                     "max_bits": max_bits,
                 },
             )
@@ -461,11 +464,11 @@ def bigint_from_uint_arrays(arrays, max_bits=-1):
 
     return create_pdarray(
         generic_msg(
-            cmd="big_int_creation",
+            cmd=f"big_int_creation<bigint,{arrays[0].ndim}>",
             args={
                 "arrays": arrays,
                 "num_arrays": len(arrays),
-                "len": arrays[0].size,
+                "shape": arrays[0].shape,
                 "max_bits": max_bits,
             },
         )
@@ -543,11 +546,6 @@ def zeros(
 
     if isinstance(shape, tuple) and len(shape) == 0:
         raise ValueError("size () not currently supported in ak.zeros.")
-
-    if ndim != 1 and max_bits is not None:
-        raise ValueError(
-            f"max_bits is not currently supported for {ndim}D pdarrays in ak.zeros, ak.ones, ak.full."
-        )
 
     repMsg = generic_msg(cmd=f"create<{dtype_name},{ndim}>", args={"shape": shape})
 
@@ -686,11 +684,6 @@ def full(
 
     if isinstance(shape, tuple) and len(shape) == 0:
         raise ValueError("size () not currently supported in ak.full.")
-
-    if ndim != 1 and max_bits is not None:
-        raise ValueError(
-            f"max_bits is not currently supported for {ndim}D pdarrays in ak.zeros, ak.ones, ak.full."
-        )
 
     repMsg = generic_msg(cmd=f"create<{dtype_name},{ndim}>", args={"shape": shape})
 
--- a/src/BigIntMsg.chpl
+++ b/src/BigIntMsg.chpl
@@ -16,18 +16,24 @@ module BigIntMsg {
     private config const logChannel = ServerConfig.logChannel;
     const biLogger = new Logger(logLevel, logChannel);
 
-    proc bigIntCreationMsg(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
+    @arkouda.instantiateAndRegister("big_int_creation")
+    proc bigIntCreationMsg(cmd: string, msgArgs: borrowed MessageArgs,
+                           st: borrowed SymTab, type array_dtype,
+                           param array_nd: int): MsgTuple throws
+        where (array_dtype == bigint) {
         param pn = Reflection.getRoutineName();
         var repMsg: string;
 
         var num_arrays = msgArgs.get("num_arrays").getIntValue();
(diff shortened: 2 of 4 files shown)


e-kayrakli:
It seems like the new-ish function registration mechanism is designed for exactly this reason. Array's dimensionality is part of its type in Chapel, and thus, has to be known at compile time. Same story with the type of its elements. Historically, this requirement, and Arkouda's dynamic nature required a lot of boilerplate to be written into the server to be able to handle many combination of types. The addition of multidim capabilities was just another dimension in this exponential list of types for which one had to write code in the server. That makes things much more unmaintainable. Therefore, (and powered by the ongoing Chapel compiler rework effort, "dyno") the function registration mechanism was implemented. I recommend checking out [this documentation](https://bears-r-us.github.io/arkouda/developer/ADDING_FEATURES.html#adding-functionality-to-the-arkouda-server) for more info on that.

Meanwhile, I prototyped get_max_bits with that infrastructure. You can see my branch in https://github.com/Bears-R-Us/arkouda/compare/master...e-kayrakli:arkouda:bigint-register. As you can see, the code is actually reduced while also adding support for multidimensional arrays. That's the direction we'd like to take Arkouda server for improving its maintainability and extensibility.

You may see that that branch also modifies the registration script to enable registering BigInt functions as they are integral to the Arkouda.

I strongly recommend reverting the server changes here as they take us in the opposite direction we want to go, and applying similar modernization in my patch to other BigInt functions to be able to support the desired functionality. Needless to say, I'd be happy to help and interested in hearing whether there are any blockers along the way in the other functions.

1RyanK:
@e-kayrakli 
> It seems like the new-ish function registration mechanism is designed for exactly this reason. Array's dimensionality is part of its type in Chapel, and thus, has to be known at compile time. Same story with the type of its elements. Historically, this requirement, and Arkouda's dynamic nature required a lot of boilerplate to be written into the server to be able to handle many combination of types. The addition of multidim capabilities was just another dimension in this exponential list of types for which one had to write code in the server. That makes things much more unmaintainable. Therefore, (and powered by the ongoing Chapel compiler rework effort, "dyno") the function registration mechanism was implemented. I recommend checking out [this documentation](https://bears-r-us.github.io/arkouda/developer/ADDING_FEATURES.html#adding-functionality-to-the-arkouda-server) for more info on that.
> 
> Meanwhile, I prototyped get_max_bits with that infrastructure. You can see my branch in [master...e-kayrakli:arkouda:bigint-register](https://github.com/Bears-R-Us/arkouda/compare/master...e-kayrakli:arkouda:bigint-register). As you can see, the code is actually reduced while also adding support for multidimensional arrays. That's the direction we'd like to take Arkouda server for improving its maintainability and extensibility.
> 
> You may see that that branch also modifies the registration script to enable registering BigInt functions as they are integral to the Arkouda.
> 
> I strongly recommend reverting the server changes here as they take us in the opposite direction we want to go, and applying similar modernization in my patch to other BigInt functions to be able to support the desired functionality. Needless to say, I'd be happy to help and interested in hearing whether there are any blockers along the way in the other functions.

That makes a lot of sense. I will change things over to doing it that way.

That being said - would it make more sense to put `BigIntMsg.chpl` in the `ServerModules.cfg` rather than `DEFAULT_MODS`? I don't see a huge difference in how those two things are treated, but I see a lot of things like `<something>Msg.chpl` in `ServerModules.cfg` and `DEFAULT_MODS` sounds like it would be the bare bones of what's required.

And should we remove `MaxArrayDims` from `ServerConfig.chpl` entirely? I think it's only in that file twice and apparently in `ManipulationMsg.chpl` but I'm not sure why if it wasn't being set before. It feels like the usage in `ManipulationMsg.chpl` could be done Python-side.

e-kayrakli:
> That being said - would it make more sense to put BigIntMsg.chpl in the ServerModules.cfg rather than DEFAULT_MODS? I don't see a huge difference in how those two things are treated, but I see a lot of things like <something>Msg.chpl in ServerModules.cfg and DEFAULT_MODS sounds like it would be the bare bones of what's required.

My take (and I don't have a good historical context for this) is that the `bigint` type is an integral part of Arkouda. Whether the server supports bigints is determined by the `registration-config.json` file. Whereas `ServerModules.cfg` cover modules that are more optional features of the server. I am relatively certain that `DEFAULT_MODS` is not something that's generally important -- it is just an implementation detail of the registration scripts. All that being said, I don't have a strong opinion as to whether `BigIntMsg` should be in `ServerModules.cfg`. I wonder what @ajpotts thinks.

> And should we remove MaxArrayDims from ServerConfig.chpl entirely? I think it's only in that file twice and apparently in ManipulationMsg.chpl but I'm not sure why if it wasn't being set before. It feels like the usage in ManipulationMsg.chpl could be done Python-side.

This is an interesting tradeoff, IMO. On one hand, it is extra maintenance to carry this variable around, with what looks like a very limited usage as you noted. On the other, it can provide additional level of resilience for the server. Silent server crashes has been an issue where we are trying to reduce. Functions making error checking on the server side could be seen as increased resilience and worthwhile. We could consider using `maxArrayDims` in the function registeration mechanism where these checks are generated automatically for the functions registered, for example. Based on this, I have a slight preference for keeping it and using that variable more in the server. Note that checks based on this variable have no impact on execution time. It may increase compilation time a bit, though. I'd like to see what @ajpotts thinks on this matter, too.

ajpotts:
I think `BigIntMsg.chpl` should be in the `ServerModules.cfg`.  `OperatorMsg` is `ServerModules.cfg`, even though it contains the binops which are pretty fundamental.  Plus, I think it's easier to maintain to have them all listed in a single file.

ajpotts:
I actually think we should remove `MaxArrayDims` from `ServerConfig.chpl`, but not until we refactor the only function using it (`expand_dims`) to use the compiled ranks instead.  I suspect this variable is a hold-over from the previous registration mechanism that has been deprecated.  I agree we need more error handling for the compiled ranks in general.  The problem with using the `MaxArrayDims` is that it doesn't account for non-contiguous ranks (for example, `[1,2,5]`), so any rank checking based on this variable would be incomplete.

arkouda/numpy/pdarraycreation.py
```diff
@@ -237,14 +237,17 @@ def array(
...
             # attempt to break bigint into multiple uint64 arrays
             # early out if we would have more uint arrays than can fit in max_bits
             early_out = (max_bits // 64) + (max_bits % 64 != 0) if max_bits != -1 else float("inf")
             while (a != 0).any() and len(uint_arrays) < early_out:
                 low, a = a % 2**64, a // 2**64
                 uint_arrays.append(array(np.array(low, dtype=np.uint), dtype=akuint64))
-            return bigint_from_uint_arrays(uint_arrays[::-1], max_bits=max_bits).reshape(sh)
+            # If uint_arrays is empty, this will create an empty ak array and reshape it.
+            if not uint_arrays:
+                return bigint_from_uint_arrays(uint_arrays[::-1], max_bits=max_bits).reshape(sh)
```
ajpotts, line 248: Can you use `zeros` here and avoid the reshape?

ajpotts (approved):
Love it!  Thanks for figuring this one out.

arkouda/numpy/pdarraycreation.py
1RyanK: Done!

src/BigIntMsg.chpl
```diff
@@ -1,4 +1,5 @@
 module BigIntMsg {
+    use CommandMap;
```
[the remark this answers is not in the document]
1RyanK: Done!

arkouda/numpy/pdarraycreation.py
ajpotts: Thanks!

src/BigIntMsg.chpl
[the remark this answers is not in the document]
ajpotts: Thanks!

jaketrookman (approved):
Looks good

ajpotts (approved):

PR #4208, opened 2025-03-25
Closes 4206:  add install-pytables to Makefile

ajpotts:
This PR adds the ability to install `pytables` from source with a makefile command `make install-pytables`.  It also adds tests the CI.  In the case of almalinux, the `make install-hdf5` is moved to coincide as it is a dependency of `pytables` that cannot be installed with `dnf` on the current container version.

Closes 4206:  add install-pytables to Makefile

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -161,15 +161,6 @@ jobs:
         retry_on: error
         command: |
           make install-arrow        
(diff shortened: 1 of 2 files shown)


1RyanK (approved):
Looks good!

jaketrookman (approved):

PR #4214, opened 2025-03-31
Closes #3005: searchsorted function

### LINKED ISSUE #3005: searchsorted function
Replicate the functionality of numpy.searchsorted:
https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html

1RyanK:
Chapel already had code for `searchsorted`, so I just added different types and fixed a problem in the comparators. Also added tests. Also fixed a typo in an unrelated test.

Closes #3005: searchsorted function

### DIFF
--- a/arkouda/numpy/sorting.py
+++ b/arkouda/numpy/sorting.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from enum import Enum
-from typing import TYPE_CHECKING, Sequence, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Literal, Sequence, TypeVar, Union, cast
 
 import numpy as np
 from typeguard import check_type, typechecked
@@ -17,12 +17,12 @@ from arkouda.numpy.dtypes import (
     uint64,
 )
 from arkouda.numpy.pdarrayclass import create_pdarray, pdarray
-from arkouda.numpy.pdarraycreation import zeros
+from arkouda.numpy.pdarraycreation import array, zeros
 from arkouda.numpy.strings import Strings
 
 numeric_dtypes = {dtype(int64), dtype(uint64), dtype(float64)}
 
-__all__ = ["argsort", "coargsort", "sort", "SortingAlgorithm"]
+__all__ = ["argsort", "coargsort", "sort", "SortingAlgorithm", "searchsorted"]
 
 SortingAlgorithm = Enum("SortingAlgorithm", ["RadixSortLSD", "TwoArrayRadixSort"])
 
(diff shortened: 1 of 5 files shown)


arkouda/numpy/sorting.py
```diff
@@ -282,3 +282,66 @@ def sort(
...
+    """
+    Find indices where elements should be inserted to maintain order.
+
+    Find the indices into a sorted array `a` such that, if the corresponding
+    elements in `v` were inserted before the indices, the order of `a` would be preserved.
+
+    Parameters
+    ----------
+    a : pdarray
+        1-D input array. Must be sorted in ascending order. `sorter` is not currently supported.
```
ajpotts, line 300: I think we should throw an error if someone tries this on a multi-dimensional pdarray.

src/SortMsg.chpl
```diff
@@ -123,15 +125,19 @@ module SortMsg
     }
 
     record leftCmp: relativeComparator {
-      proc compare(a: real, b: real): int {
-        if a < b then return -1;
+      proc compare(a: ?t, b: t): int
+        where (t == int || t == real || t == uint || t == uint(8) || t == bigint)
+      {
+        if a <= b then return -1;
```
e-kayrakli, line 131: Just noting the small change in the logic here. Before, this was just `<`, now you are making it `<=`. Based on the fact that you are also making a similar change in `rightCmp`, I am guessing this is intentional. If that's the case, that's perfectly fine. This can have an impact on [sort stability](https://stackoverflow.com/questions/1517793/what-is-stability-in-sorting-algorithms-and-why-is-it-important)

e-kayrakli (approved):

src/SortMsg.chpl
1RyanK: I found that when side was left, it was giving me the result that I was supposed to get for right, and vice versa. I think this only affects this function though, right?

arkouda/numpy/sorting.py
```diff
@@ -282,3 +282,77 @@ def sort(
...
+    3
+    >>> v = ak.array([-10, 20, 12, 13])
+    >>> ak.searchsorted(a, v)
+    array([0 5 1 2])
+    """
+
+    if a.ndim > 1:
+        raise ValueError(f"a must be one dimensional, but has {a.ndim} dimensions.")
+    if a.dtype not in numeric_dtypes and a.dtype != bigint:
+        raise ValueError(f"ak.searchsorted supports int64, uint64, bigint, or float64, not {a.dtype}")
```
ajpotts, line 334: I think this should be a `TypeError` rather than a `ValueError`.

ajpotts (approved):
Great job!

src/SortMsg.chpl
ajpotts: I looked through the code and `leftCmp` is only used in `searchSorted`, so I think this is fine.

arkouda/numpy/sorting.py
1RyanK: Done. I was looking at `sort` and it has a `ValueError` for this.

jaketrookman (approved):
Looks good

arkouda/numpy/sorting.py
```diff
@@ -282,3 +282,77 @@ def sort(
...
+    Returns
+    -------
+    indices : int or pdarray
+        If `v` is an array, returns an array of insertion points with the same shape.
+        If `v` is a scalar, returns a single integer index.
+
+    Raises
+    ------
+    ValueError
+        If searchsort is attempted on a `pdarray` with an unsupported dtype such as `bool`.
```
ajpotts, line 316: This first one should be changed to a TypeError to match the code.

1RyanK: Done! Sorry about that.

ajpotts (approved):

PR #4224, opened 2025-04-07
Closes 4124, align ak.dot to np.dot.

### LINKED ISSUE #4124: Align ak.dot to numpy dot
The current implementation matches numpy for 1D, but not for multi-dimensions.

drculhane:
Closes #4124 

Summary of work:

There was a lot here, hence these comments so that you won't have to start from scratch when you review this.

Aligning ak.dot to np.dot has been time-consuming, as np.dot does much more than the conventional 1-D vector dot product.  Hopefully the comments in def dot, in pdarrayclass.py, make it clear what's going on.

In addition to the executable python code, there's a new docstring that gives examples of all use cases of ak.dot.

It's worth noting that in order to prevent mypy from freaking out, ak.dot requires several # type: ignore statements, as is typical with functions whose inputs accept a Union of types.  There's also an unreachable return None (also with a # type: ignore) which prevents mypy from claiming there's a missing return.  I considered just having mypy bypass this function.

There were test_dot functions in both pdarrayclass_test.py and numeric_test.py.  The one in numeric_test.py has been removed.  The one in pdarrayclass_test.py has been changed to cover all the possible use cases and data types of ak.dot.

A possible TODO: move dot out of pdarrayclass.py and into _numeric.py.  Similarly with test_dot.

A dotProd function was added to LinalgMsg.chpl to perform the computation (thank you, Ryan).  I modified Ryan's code to allow for the case where the two pdarrays aren't of the same type.

In LinalgMsg.chpl, both dotProd and matmul need to compute the return type based on the input types.  Both call a function named np_ret_type (which is meant to stand for "numpy return type").  There is a similar function called by vecdot (it's currently called compute_result_type_vecdot).  In a future alignment of ak.vecdot to np.vecdot, that function may also be rolled into np_ret_type.

Additional changes were needed in LinalgMsg.chpl, OperatorMsg.chpl and BinOp.chpl to enable all of the appropriate combinations of data types.

Another possible TODO: make a 1D case for dotProd, rather than piggyback onto the ND case.

A very likely TODON'T: make test cases for > 3D.  So far, I've found that making arkouda with max_dim > 3 isn't  feasible.

Ryan: I'd love if you could add some comments to the dotProd proc in LinalgMsg.chpl.


src/LinalgMsg.chpl
```diff
@@ -276,6 +271,133 @@ module LinalgMsg {
...
+    // TODO: add bigint handling
+
+    // This implements the M-D x N-D case of ak.dot, aligning its funcionality to np.dot.
+ 
+    @arkouda.registerCommand(name="dot")
+    proc dotProd(a: [?d] ?ta, b: [?d2] ?tb): [] np_ret_type(ta,tb) throws
+    where ((d.rank >=2 && d2.rank>= 2)
+        && (ta == int || ta == real || ta == bool || ta == uint )
+        && (tb == int || tb == real || tb == bool || tb == uint )
+           ) {
```
e-kayrakli, line 304: I brought it up in the internal Chapel channels that we should add distributed array support to `LinearAlgebra.dot`, which can be used by this function for at least >=2D cases. This implementation is probably OK, but I wouldn't expect it to perform well or scale.

e-kayrakli (approved):

src/LinalgMsg.chpl
```diff
@@ -276,6 +271,133 @@ module LinalgMsg {
...
+            tmp1 = a[aDom];                 
+            tmp2 = b[bDom];
+            
+            var total: np_ret_type(ta,tb) = 0:np_ret_type(ta,tb);
+
+            // Perform the dot product.
+
+            for (i, j) in zip(tmp1, tmp2) {
+                if np_ret_type(ta,tb) == bool {
+                    total |= i:bool & j:bool;
```
e-kayrakli, line 369: Sometimes `&` can be used instead of `&&` for performance-related motivations. But, to my knowledge, those cases arise when a more complicated boolean expression is involved. For this one, I recommend `&&`, logical-and, for clarity.

(As per the language specification, what you have here should be portable, though: https://chapel-lang.org/docs/main/language/spec/conversions.html#explicit-numeric-conversions)

src/LinalgMsg.chpl
drculhane: We've had some discussions about that here, too.  None of the linear algebra stuff has been optimized for distributed processing.  I lean towards merging this, and creating a separate issue for "optimize linear algebra."

src/LinalgMsg.chpl
drculhane: You're absolutely right.  We'll make that change.

arkouda/numpy/pdarrayclass.py
```diff
@@ -3359,51 +3417,128 @@ def dot(
     pda2: Union[np.int64, np.float64, np.uint64, pdarray],
 ) -> Union[numpy_scalars, pdarray]:
     """
-    Returns the sum of the elementwise product of two arrays of the same size (the dot product) or
-    the product of a singleton element and an array.
+    computes dot product of two arrays.
```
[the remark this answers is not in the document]
drculhane: Yes, it should.  Fixed.

src/LinalgMsg.chpl
e-kayrakli: Yeah. What we have in the `LinearAlgebra` module is relatively optimized but only for single locale. We should merge this as-is.

1RyanK (approved):

PR #4242, opened 2025-04-16
Fix distributed array creation and sorting issues

### LINKED ISSUE #4233: `makeDistArray` can return a non-distributed array
**Describe the bug**
`makeDistArray` has overloads which accepts a domain, and do not check if the domain is distributed. Therefore the array we get from `makeDistArray` can be non-distributed. Which can lead to surprising behavior and bad performance.

See https://github.com/Bears-R-Us/arkouda/blob/0a7850ba78eff2302994438d7e47a61c0d63101f/src/SymArrayDmap.chpl#L94-L119 for the source code of `makeDistArray`.

**Proposed Solution**

My recommendation here is that we should modify these overloads to check if `D` is distributed and if it isn't, call `makeDistDom` to get a distributed domain instead.

**Is this a Blocking Issue**
- Are there any known work-arounds? No

ShreyasKhandekar:
Ensure `makeDistArray` consistently creates distributed arrays and update sorting implementations to utilize the new `makeDistArray`. 
Address build issues caused by the previous sorting method by refactoring the sorting logic into a dedicated module.

Closes #4233

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -20,7 +20,7 @@ module MultiTypeSymbolTable
 
     private config const logLevel = ServerConfig.logLevel;
     private config const logChannel = ServerConfig.logChannel;
-    const mtLogger : owned Logger = new Logger(logLevel, logChannel);
+    const mtLogger = new Logger(logLevel, logChannel);
 
     /* symbol table */
     class SymTab
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -20,7 +20,7 @@ module MultiTypeSymbolTable
 
     private config const logLevel = ServerConfig.logLevel;
     private config const logChannel = ServerConfig.logChannel;
-    const mtLogger = new Logger(logLevel, logChannel);
+    const mtLogger : owned Logger = new Logger(logLevel, logChannel);
 
     /* symbol table */
     class SymTab
--- a/src/SymArrayDmap.chpl
+++ b/src/SymArrayDmap.chpl
@@ -92,27 +92,39 @@ module SymArrayDmap {
     }
 
     proc makeDistArray(D: domain(?), type etype) throws
-      where D.rank == 1
+      where MyDmap != Dmap.defaultRectangular && D.isDefaultRectangular()
+    {
+      return makeDistArray(makeDistDom(D), etype);
+    }
+
+    proc makeDistArray(D: domain(?), type etype) throws
+      where D.rank == 1 && (MyDmap == Dmap.defaultRectangular || !D.isDefaultRectangular())
     {
       var res = D.tryCreateArray(etype);
       return res;
     }
 
     proc makeDistArray(D: domain(?), type etype) throws
-      where D.rank > 1
+      where D.rank > 1 && (MyDmap == Dmap.defaultRectangular || !D.isDefaultRectangular())
     {
       var res: [D] etype;
       return res;
     }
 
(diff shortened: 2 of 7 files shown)


src/MultiTypeSymbolTable.chpl
```diff
@@ -20,7 +20,7 @@ module MultiTypeSymbolTable
 
     private config const logLevel = ServerConfig.logLevel;
     private config const logChannel = ServerConfig.logChannel;
-    const mtLogger = new Logger(logLevel, logChannel);
+    const mtLogger : owned Logger = new Logger(logLevel, logChannel);
```
e-kayrakli, line 23: IIRC, this was an earlier attempt to solve the issue. If that's the case, I have a mild preference for reverting this.

src/SymArrayDmap.chpl
```diff
@@ -92,27 +92,39 @@ module SymArrayDmap {
...
+    proc makeDistArray(D: domain(?), type etype) throws
+      where D.rank == 1 && (MyDmap == Dmap.defaultRectangular || !D.isDefaultRectangular())
     {
       var res = D.tryCreateArray(etype);
       return res;
     }
 
     proc makeDistArray(D: domain(?), type etype) throws
-      where D.rank > 1
+      where D.rank > 1 && (MyDmap == Dmap.defaultRectangular || !D.isDefaultRectangular())
```
e-kayrakli, line 108: This is the only case that gave me a pause. If I am not mistaken, this is for when:

- the server is configured to use local arrays
- and the domain on which we are supposed to create an array is distributed

And on this branch, it looks like that would result in this function returning a distributed array.

If that's the case my reflex is to return a local array here, as I would think that the server's configuration is what matters in terms of the behavior of `makeDistArray`. What do you think?

src/SymArrayDmap.chpl
```diff
@@ -92,27 +92,39 @@ module SymArrayDmap {
...
     proc makeDistArray(D: domain(?), initExpr: ?t) throws
-      where D.rank == 1
+      where D.rank == 1 && (MyDmap == Dmap.defaultRectangular || !D.isDefaultRectangular())
     {
       return D.tryCreateArray(t, initExpr);
     }
 
     proc makeDistArray(D: domain(?), initExpr: ?t) throws
-      where D.rank > 1
+      where D.rank > 1 && (MyDmap == Dmap.defaultRectangular || !D.isDefaultRectangular())
```
e-kayrakli, line 127: Note that this is the same case as above and I think it is obvious that whatever we decide on, the behavior should be consistent between these two.

src/SymArrayDmap.chpl
ShreyasKhandekar: I see your point, I had the same thought when implementing this, but I decided to match the behavior for other overloads of `makeDistArray` which have the same behavior:

https://github.com/Bears-R-Us/arkouda/blob/a723f6f78fc2057cb3dba811e6027b5439271338/src/SymArrayDmap.chpl#L80-L86

If we change this here, we should also change the behavior in other overloads.

At the same time, I don't know if it's really a problem. The problem would be if someone creates the "first" distributed domain/array on the server without regard for the configuration being multi-locale or not. 
This is just a secondary way to make arrays and domains, using other existing domains. So it's not a problem by itself, and in any case, this overload should be kept to cover all the cases of the `where` clauses.

e-kayrakli: https://github.com/Bears-R-Us/arkouda/issues/4243

e-kayrakli (approved):

PR #4319, opened 2025-04-23
Revert "Fix distributed array creation and sorting issues"

### LINKED ISSUE #4318: argsort bug
**Describe the bug**
`ak.argsort` on `Strings` crashes the server with gasnet and 10 locales.

**To Reproduce**
```python

import sys
import arkouda as ak
from arkouda.client import generic_msg
from arkouda.comm_diagnostics import (start_comm_diagnostics, stop_comm_diagnostics, print_comm_diagnostics_table, reset_comm_diagnostics)
import random
from arkouda import Strings
import time
ak.connect()

how_many = 1000
size_1 = 200

base_array = [str(i) for i in range(how_many)]
list_1 = []
for i in range(size_1):
    val = random.choice(base_array)
    list_1.append(val)
    base_array.remove(val)

print(list_1)

ak_arr_1 = ak.array(list_1)


ak.argsort(ak_arr_1)

ak.shutdown()

```
Crashes the server.

ajpotts:
This change needs to be reverted until #4318 is addressed.

Reverts Bears-R-Us/arkouda#4242
Closes #4318: argsort bug


ShreyasKhandekar:
Would it be enough to only revert b9449fcd846840646550b1f99c7a435476f7a9a2 instead of the whole PR in order to patch this error? It should fix the argsort bug, but it will introduce a warning during arkouda compilation, but still work afaik.

If possible we should try not to revert to the previous behavior of make dist array.

1RyanK (approved):
Looks good!

jaketrookman (approved):

ajpotts:
Closing this for now because @ShreyasKhandekar put up a fix: https://github.com/Bears-R-Us/arkouda/pull/4327

PR #4322, opened 2025-04-24
Closes #4317: union1d needs more rigorous unit tests on Strings

### LINKED ISSUE #4317: `union1d` needs more rigorous unit tests on Strings
**Feature Description**
A recent PR crashes the server when I run `union1d` with ten locales on two Strings arrays, each of size 200. The goal is to add this in as a unit test.

1RyanK:
Added a unit test that succeeds on one locale but fails on ten locales (ultimately because of #4318). Also noticed that the bool generation in that file was only generating `False` because `np.random.randint` is exclusive on the upper bound.

Closes #4317: union1d needs more rigorous unit tests on Strings

### DIFF
--- a/tests/numpy/setops_test.py
+++ b/tests/numpy/setops_test.py
@@ -3,6 +3,7 @@ import pandas as pd
 import pytest
 
 import arkouda as ak
+from arkouda.testing import assert_arkouda_array_equivalent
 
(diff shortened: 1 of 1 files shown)


ajpotts (approved):

drculhane (approved):
Looks good.

jaketrookman (approved):
Looks good

PR #4325, opened 2025-04-25
Revert part of #4242 Make dist array always dist

ShreyasKhandekar:
This change needs to be reverted until https://github.com/Bears-R-Us/arkouda/issues/4318 is addressed.

Reverts only a part of  https://github.com/Bears-R-Us/arkouda/pull/4242, specifically just b9449fcd846840646550b1f99c7a435476f7a9a2

This is the replacement for https://github.com/Bears-R-Us/arkouda/pull/4319 and takes a more nuanced approach instead.


e-kayrakli (approved):

ShreyasKhandekar:
Since this PR is reverting the change as a stopgap, this is not a permanent fix. In fact the current state is undesirable because we're still using a non-distributed sort on distributed String arrays.

This will also cause the following compilation warning to show up in the meantime:

```sh
$CHPL_HOME/modules/standard/Sort.chpl:664: In function 'unstableSort':
$CHPL_HOME/modules/standard/Sort.chpl:680: warning: twoArrayRadix sort no longer handles distributed arrays. Please use TwoArrayDistributedRadixSort.twoArrayDistributedRadixSort instead (but note that it is not stable)
  $CHPL_HOME/modules/standard/Sort.chpl:753: called as unstableSort(x: [BlockDom(1,int(64),one,unmanaged DefaultDist)] (string,int(64)), comparator: StringIntComparator, region: range(int(64),both,one)) from function 'sort'
  /Users/khandeka/arkouda/src/SegStringSort.chpl:79: called as sort(x: [BlockDom(1,int(64),one,unmanaged DefaultDist)] (string,int(64)), comparator: StringIntComparator, param stable = false) from function 'twoPhaseStringSort'
  /Users/khandeka/arkouda/src/SegmentedString.chpl:1277: called as twoPhaseStringSort(ss: borrowed SegString)
note: generic instantiations are underlined in the above callstack
```

ShreyasKhandekar:
I've created https://github.com/Bears-R-Us/arkouda/issues/4326 to track the resolution of the underlying issue.

ShreyasKhandekar:
This wasn't the fix we were looking for, closing this in favor of https://github.com/Bears-R-Us/arkouda/pull/4327

Issue #4326, opened 2025-04-25
SegStringSort is using a non-distributed sort


ShreyasKhandekar:
We use `sort` for sorting strings in SegStringSort.chpl. `sort` is not distributed, we should instead use `
The context for this issue comes from a few other issues that revealed the problem here.

1. `makeDistArray` not always creating distributed arrays. This was patched in #4242, and it revealed this issue
2. However, the patch in #4242 caused a bug to creep in causing argsort crashes as documented in #4318 
   - Specifically, the issue was in this commit where we tried to switch to a dynamically chose distributed radix sort: https://github.com/Bears-R-Us/arkouda/commit/b9449fcd846840646550b1f99c7a435476f7a9a2
3. We reverted the above commit in #4325 as stopgap to fix server crashes and nightly testing.

However, the undesirable state of using a non-distributed sort still exists. In fact, users have also reported that they've noticed the sort being slow.

We need to find a way to reapply the change that was originally made in https://github.com/Bears-R-Us/arkouda/commit/b9449fcd846840646550b1f99c7a435476f7a9a2, without causing the server crashes.

PR #4327, opened 2025-04-25
Resolve server crashes in SegStringSort

ShreyasKhandekar:
Resolves https://github.com/Bears-R-Us/arkouda/issues/4318
Resolves https://github.com/Bears-R-Us/arkouda/issues/4326

However, this started causing server crashes for argsort on strings.

This is because there were cases where we were actually relying on the buggy behavior of makeDistArray not creating a distributed array. And when this behavior was fixed, this started causing server crashes. We found the offending behavior and patched it in a current best effort to get nightly testing back to blue.

Overall this is a strict improvement from what we had before, and should also have a huge boost in performance for argsort in string since now we're using a distributed sort instead of a local sort for twoPhaseStringSort

### DIFF
--- a/src/SegStringSort.chpl
+++ b/src/SegStringSort.chpl
@@ -33,7 +33,7 @@ module SegStringSort {
       return (new defaultComparator()).keyPart(a0, i);
     }
   }
-  
+
   proc twoPhaseStringSort(ss: SegString): [ss.offsets.a.domain] int throws {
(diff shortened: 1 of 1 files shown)


ajpotts (approved):
Thanks for fixing this!

1RyanK (approved):
This definitely runs much faster on the test case I was trying.

PR #4344, opened 2025-04-29
Closes #4313: ak.merge does not support Categorical

### LINKED ISSUE #4313: `ak.merge` does not support `Categorical`
**Feature Description**
If users would like to merge Arkouda DataFrames on Categorical columns, this is not directly supported. Furthermore, just changing the check in `merge` to `isinstance(right[right_col], (pdarray, Strings, Categorical))` doesn't fix it. It breaks further down the stack trace.

**Describe Workarounds**
Theoretically users could merge on the underlying pdarray of ints for the categories. Something like this will probably be how it gets implemented.

1RyanK:
Categorical support didn't take that much reworking. Includes unit test.

Closes #4313: `ak.merge` does not support `Categorical`


ajpotts (approved):
Looks good.  Thanks!

jaketrookman (approved):

PR #4359, opened 2025-04-30
Fix arkouda commit in containers following upgrade to arrow 19.0.1

ajpotts:
Fixes arkouda commit in containers following upgrade to arrow 19.0.1

### DIFF
--- a/docker/ubuntu-with-arkouda-deps/1.0.2/Dockerfile
+++ b/docker/ubuntu-with-arkouda-deps/1.0.2/Dockerfile
@@ -96,10 +96,8 @@ RUN  mkdir -p $DEP_DIR && mkdir -p $DEP_BUILD_DIR && mkdir -p $DEP_INSTALL_DIR
 #COPY build.tar $DEP_DIR/build.tar
 #RUN cd $DEP_DIR && tar -xvf build.tar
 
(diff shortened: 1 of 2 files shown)


docker/ubuntu-with-arkouda-deps/1.0.2/Dockerfile
```diff
@@ -97,9 +97,8 @@ RUN  mkdir -p $DEP_DIR && mkdir -p $DEP_BUILD_DIR && mkdir -p $DEP_INSTALL_DIR
 #RUN cd $DEP_DIR && tar -xvf build.tar
 
 #   TODO:  Fix the correct commit from master branch.
```
1RyanK, line 99: I'll admit I'm going off of what ChatGPT said here, but my understanding is that this PR does this TODO (and the other one in the other file), so would it be worth removing the TODO?

1RyanK (approved):
Looks good!

jaketrookman (approved):

docker/ubuntu-with-arkouda-deps/1.0.2/Dockerfile
ajpotts: Good catch.  Fixed.

1RyanK (approved):

Issue #4346, opened 2025-04-29
Running Categorical tests with two locales crashes the server


1RyanK:
**Describe the bug**
In `categorical_test.py`, attempting the test `test_basic_categorical` crashes the server when running with two locales.

**To Reproduce**
Be sure to run `make` with gasnet. Then running `make test` or changing `pytest.ini` so that you can run the server with two locales, and then either `make test` or going to that specific test in PyCharm and running it crashes the server.

**Expected behavior**
Ideally the test passes.

**Error Message**
Here's the last couple lines of server output:
```
2025-04-29 12:53:40 [ServerDaemon] run Line 692 INFO [Chapel] <<< segmentedIndex took 0.045740842819213867 sec
2025-04-29 12:53:40 [ServerDaemon] run Line 701 INFO [Chapel] bytes of memory 0 used after segmentedIndex command is 0% pct of max memory 60095125094
2025-04-29 12:53:40 [ServerDaemon] run Line 578 INFO [Chapel] MessageArgs: {param_list = [(key = other, val = N/A, dtype = str), (key = objType, val = Strings, dtype = str), (key = obj, val = id_V34Xfd4_10, dtype = <U0), (key = otherType, val = str, dtype = str), (key = op, val = ==, dtype = str)], size = 5, payload = }
2025-04-29 12:53:40 [ServerDaemon] run Line 593 INFO [Chapel] >>> segmentedBinopvs ["{\"key\": \"op\", \"dtype\": \"str\", \"val\": \"==\"}", "{\"key\": \"objType\", \"dtype\": \"str\", \"val\": \"Strings\"}", "{\"key\": \"obj\", \"dtype\": \"<U0\", \"val\": \"id_V34Xfd4_10\"}", "{\"key\": \"otherType\", \"dtype\": \"str\", \"val\": \"str\"}", "{\"key\": \"other\", \"dtype\": \"str\", \"val\": \"N/A\"}"]
*** Caught a fatal signal (proc 0): SIGSEGV(11)
*** NOTICE (proc 0): Before reporting bugs, run with GASNET_BACKTRACE=1 in the environment to generate a backtrace.
*** NOTICE (proc 0): We recommend linking the debug version of GASNet to assist you in resolving this application issue.
```

**Is this a Blocking Issue**
We can run tests without gasnet.

PR #4367, opened 2025-05-02
Fix crash in segmented search due to makeDistArray misuse

ShreyasKhandekar:
Address a crash in segmented string search caused by incorrect usage of makeDistArray, which led to test failures in multilocale settings.

Resolves https://github.com/Bears-R-Us/arkouda/issues/4346


vasslitvinov (approved):
Thanks for fixing it!

Analogous to #4327

ajpotts (approved):
Thank you!!!

PR #4371, opened 2025-05-05
Closes #4355:  Remove the DAR203 errors from the docstrings

### LINKED ISSUE #4355: Remove the DAR203 errors from the docstrings
Remove the DAR203 errors from the docstrings and remove the corresponding ignore from the setup.cfg:
```
[darglint]
# How strictly to enforce: none, short, full
strictness = full
# Which docstring style: google, sphinx, or numpy
docstring_style = numpy
ignore=DAR402,DAR103,DAR101,DAR202,DAR002,DAR102,DAR201,DAR401,DAR203,
```

ajpotts:
Closes #4355:  Remove the DAR203 errors from the docstrings


arkouda/groupbyclass.py
```diff
@@ -95,12 +95,13 @@ def unique(
 
     Returns
     -------
-    unique : (list of) pdarray, Strings, or Categorical
-        The unique values. If input dtype is int64, return values will be sorted.
-    permutation : pdarray, optional
-        Permutation that groups equivalent values together (only when return_groups=True)
-    segments : pdarray, optional
-        The offset of each group in the permuted array (only when return_groups=True)
+    Union[groupable, Tuple[groupable, pdarray, pdarray, int]]
```
1RyanK, line 98: You have a lot of instances of things like pdarray|Strings, which I assume means "pdarray or Strings," would it be more consistent to make this "groupable|Tuple[groupable, pdarray, pdarray, int"? I guess that is kind of messy.

1RyanK (approved):
Looks good!

arkouda/numpy/numeric.py
```diff
@@ -149,11 +149,12 @@ def cast(
...
     -------
-    pdarray or Strings
-        Array of values cast to desired dtype
-    [validity : pdarray(bool)]
-        If errors="return_validity" and input is Strings, a second array is
-        returned with True where the cast succeeded and False where it failed.
+    Union[Union[pdarray, Strings, Categorical], Tuple[pdarray, pdarray]]
+        pdarray or Strings
+            Array of values cast to desired dtype
+        [validity : pdarray(bool)]
```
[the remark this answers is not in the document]
ajpotts: I'm sorry, I don't understand this comment.  Could you clarify what you would like?

arkouda/groupbyclass.py
ajpotts: If we want to standardize, I would say to have it exactly match the return type of the function.  Otherwise I think it will be tedious to maintain as developers to have to translate between the two.

arkouda/dataframe.py
```diff
@@ -2613,7 +2613,7 @@ def sample(self, n=5):
         +----+-----+-----+
         |    |   A |   B |
         +====+=====+=====+
-        |  0 |   0 |   0 |
+        |  0 |   0 |   0 |264
```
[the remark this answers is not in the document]
ajpotts: Fixed.

arkouda/numpy/numeric.py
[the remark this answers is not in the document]
1RyanK: Okay, I see a few different ways this is handled. Line 1291 of `strings.py` has `pdarray, int64 (optional)`, line 103 of `groupbyclass.py` has `segments : pdarray, optional`, and then there's this. The line numbers are current as of this commit. I feel like there could be a reason to have two different ways of doing it (like some sort of optional parameter vs. a different situation based on input) but I think we could probably trim it down to two.

arkouda/groupbyclass.py
1RyanK: I don't mind that. Looking at the array api website, I do see a lot of the `|` denoting `or`, and I guess another option would be that our implementation of `array_api` could match that, but everywhere else we do the `Union[...]` thing?

jaketrookman (approved):
Looks good

arkouda/numpy/numeric.py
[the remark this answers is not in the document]
1RyanK: Makes sense!

arkouda/groupbyclass.py
ajpotts: Can we fix this later?  Part of the problem is that a lot of functions to specify return types, even though they should, so it's not possible to match them at the moment.  There are other error codes that check for that, so we will be forced to figure it out.

1RyanK: Sounds good!

Issue #3964, opened 2025-01-08
Refactor benchmark handling of running mode


ajpotts:
Examine the best way to handle running mode for the benchmarks.

ajpotts:
@vasslitvinov reported the following concerns:

In the CLASS_SERVER mode, startup_teardown() indeeds starts a server, however that server is unused. Instead the tests connect to the server specified in ARKOUDA_SERVER_HOST. If there isn’t one, the tests hang.
In the GLOBAL_SERVER mode, I do not see that the server starts at all. Again, the tests hang. I think this is not the intended behavior.
benchmark.ini defines ARKOUDA_RUNNING_MODE=CLASS_SERVER whereas benchmark_v2/conftest.py defaults to GLOBAL_SERVER if ARKOUDA_RUNNING_MODE is unset. This is confusing. benchmark.ini also defines the host and the port, with benchmark_v2/conftest.py again providing the defaults, this time consistently.
My suggestion is to remove these settings from benchmark.ini, relying directly on benchmark_v2/conftest.py.

vasslitvinov:
Within Python, the information from `startup_teardown()` to `connect()` flows as a string with the host name. If two servers are launched, for example, on `localhost`, then it is easy to use one instead of the other. The issue is more obvious when servers are launched, for example, on two different compute nodes of a bigger system.

ajpotts:
I believe in GLOBAL_SERVER mode it's not supposed to start a server, but rather connect to an existing server.

ajpotts:
In regards to your concern about the serverConnectionInfo file, this section of the arkouda code:
```chapel
        /*
        Creates the serverConnectionInfo file on arkouda_server startup
        */
        proc createServerConnectionInfo() throws {
       ....
        }

        /*
        Deletes the serverConnetionFile on arkouda_server shutdown
        */
        proc deleteServerConnectionInfo() throws {
        ....
        }

```
Indicates that it will be deleted upon server shutdown.

ajpotts:
According to ChatGPT:

> Under normal, single-process pytest runs with a module-scoped fixture, you’ll only ever have one Arkouda server up at a time:
> 
> Module scope → one start/stop per module
> 
> Pytest will call your arkouda_server fixture once at the top of each test module that needs it, then tear it down before moving on.
> 
> Because teardown happens before the next module’s setup, there’s never any overlap within a single pytest worker.
> 
> Session scope → one start/stop per entire run
> 
[...]

PR #4373, opened 2025-05-06
Part 2 of #3964:  Refactor benchmark handling of running mode

ajpotts:
Adds `--nl` option to the benchmark `conftest.py`.

Part 2 of #3964:  Refactor benchmark handling of running mode

### DIFF
--- a/benchmark_v2/conftest.py
+++ b/benchmark_v2/conftest.py
@@ -34,6 +34,16 @@ def pytest_addoption(parser):
         help="arkouda server port",
     )
 
+    default_nl = get_arkouda_numlocales()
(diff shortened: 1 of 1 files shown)


vasslitvinov (approved):
Looks good. I think the default for `--nl` may not do what's intended in corner cases. For example, how will it distinguish between `--nl 1` and the default for a multilocale server.

It would make sense to provide also an environment variable like `ARKOUDA_SERVER_LOCALES` by analogy to other options, however it is fine without it.

vasslitvinov (approved):

1RyanK (approved):
Looks good!

PR #4374, opened 2025-05-06
Removed deprecated functions

ajpotts:
This PR removes the `Series.attach` function, as well as the `closed` argument from `timeclass.date_time`, both of which have been deprecated.


1RyanK (approved):
Looks good!

jaketrookman (approved):

PR #4378, opened 2025-05-07
Auto checkpoint upon exceeding memory percentage

vasslitvinov:
Initial implementation of automatic checkpointing, performed in the server before waiting for the next zmq message from the client based on a memory percentage threshold. The threshold is controlled by a config argument of the server and is off by default. Automatic checkpointing is not done if the CheckpointMsg module was included in the build.

Minimal intelligence is included that suppresses automatic checkpointing if it has already been performed or a checkpoint has been loaded, unless memory has been below the threshold after these took place.

Essential next steps: report it to the user that the checkpointing is taking place and wait for 5 seconds (for example) idle time so that checkpointing does not occur during intermediate computations. For example, `ak.ones(nnn)` takes two messages and the checkpoint will currently occur between them if threshold is exceeded.

I would also like to move `activityMutex` to a "common header" file like `MultiTypeSymbolTable` so that CheckpointMsg can do locking at a finer granularity than (currently) for the entire `autoCheckpointMsg()`.

### DIFF
--- a/src/CheckpointMsg.chpl
+++ b/src/CheckpointMsg.chpl
@@ -21,21 +21,23 @@ module CheckpointMsg {
 
   config const dsetname = "data";
 
+  /* Perform checkpointing automatically when used memory exceeds this many percent
+     of available memory. No auto-checkpointing if 0 or below. */
+  config const ckptMemPct = 0;
+
+  // Prevent further auto-checkpoints upon exceeding 'ckptMemPct'
+  // until memory usage drops below the limit, or upon loadCheckpointMsg.
+  private var skipAutoCkpt = false;
+
   private config const logLevel = ServerConfig.logLevel;
   private config const logChannel = ServerConfig.logChannel;
   const cpLogger = new Logger(logLevel,logChannel);
 
 
-  proc saveCheckpointMsg(cmd: string, msgArgs: borrowed MessageArgs,
-                         st: borrowed SymTab): MsgTuple throws {
-    var basePath = msgArgs.getValueOf("path");
-    const nameArg = msgArgs.getValueOf("name");
-    const mode = msgArgs.getValueOf("mode");
-
+  private proc saveCheckpointImpl(basePath, cpName, mode, st) throws {
     if !exists(basePath) then
       mkdir(basePath);
 
-    const cpName = if nameArg.isEmpty() then st.serverid else nameArg;
     const cpPath = Path.joinPath(basePath, cpName);
 
     if exists(cpPath) {
@@ -80,11 +82,23 @@ module CheckpointMsg {
         cpLogger.debug(M(), R(), L(), "Cannot save %s".format(name));
       }
     }
+  }
+
+  proc saveCheckpointMsg(cmd: string, msgArgs: borrowed MessageArgs,
+                         st: borrowed SymTab): MsgTuple throws {
+    const basePath = msgArgs.getValueOf("path");
+    const nameArg = msgArgs.getValueOf("name");
+    const cpName = if nameArg.isEmpty() then st.serverid else nameArg;
+    const mode = msgArgs.getValueOf("mode");
+
+    saveCheckpointImpl(basePath, cpName, mode, st);
+
     return Msg.send(cpName);
   }
 
   proc loadCheckpointMsg(cmd: string, msgArgs: borrowed MessageArgs,
                          st: borrowed SymTab): MsgTuple throws {
+    skipAutoCkpt = true;
     var basePath = msgArgs.getValueOf("path");
     const nameArg = msgArgs.getValueOf("name");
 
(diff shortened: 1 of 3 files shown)


src/CheckpointMsg.chpl
```diff
@@ -131,6 +145,45 @@ module CheckpointMsg {
     return Msg.send(nameArg);
   }
 
+  private proc needMemBasedCkpt() {
+    if ckptMemPct <= 0 then return false;
+
+    // check whether memory use exceeds ckptMemPct
+    if (getMemUsed():real / getMemLimit()) > ckptMemPct / 100.0 {
+      return if skipAutoCkpt then false else true;
```
e-kayrakli, line 153: Or simply `return !skipAutoCkpt;`

src/CheckpointMsg.chpl
```diff
@@ -131,6 +145,45 @@ module CheckpointMsg {
...
+    } else {
+      skipAutoCkpt = false;
```
e-kayrakli, line 155: So, we disable skipping if we are not using a lot of memory. Why?

src/ServerDaemon.chpl
```diff
@@ -724,6 +736,30 @@ module ServerDaemon {
                                                                             elapsed));
             this.shutdown(); 
         }
+
+        /* While waiting for a message from client, do checkpointing if desired. */
+        proc startIdleServer() throws {
+          // Do nothing if the checkpointing module was not included.
+          if commandMap.contains("auto_checkpoint") {
+            const msgArgs = new owned MessageArgs();
+            msgArgs.addPayload(b"idle start");
```
e-kayrakli, line 745: Is the motivation for using a message-based approach to be able to do `commandMap.contains`? I am on board with it in any case. However, checkpointing is a core feature of Arkouda and we can consider making it part of the server no matter what. I can't imagine it adding a ton of compilation time.

Short of that, it would be good to wire in a mechanism so that the server can check whether a module is loaded or not without having to do this. I also wanted that feature so that when the server sees a message to be not supported, it can tell you which module it belongs to. IOW, during compilation, we would still take a look at the excluded modules just to see which messages they implement, but then not compile them. That's a bit of a tangent, though.

src/ServerDaemon.chpl
```diff
@@ -514,11 +522,13 @@ module ServerDaemon {
             var startTime = timeSinceEpoch().totalSeconds();
         
             while !this.shutdownDaemon {
+                startIdleServer();
```
e-kayrakli, line 525: Did you mean to put this in a `begin`?

ajpotts (approved):
Looks OK to me.  I'm interested in how you will test this out.

src/CheckpointMsg.chpl
```diff
@@ -131,6 +145,45 @@ module CheckpointMsg {
...
+                          .format(ckptMemPct, basePath, cpName));
+            saveCheckpointImpl(basePath, cpName, "overwrite", st);
+            cpLogger.info(M(), R(), L(), "finished autoCheckpoint into %s/%s".format(basePath, cpName));
+            skipAutoCkpt = true;
+            return MsgTuple.success("autoCheckpoint: completed");
+          }
+        }
+//        when b"idle start" {
+//          // nothing for now
+//        }
```
e-kayrakli, line 177: remove?

src/ServerDaemon.chpl
e-kayrakli: If not, the mutex is not necessary I think?

src/ServerDaemon.chpl
```diff
@@ -724,6 +736,30 @@ module ServerDaemon {
...
+            // ignore the returned message, for now
+            executeCommand("auto_checkpoint", msgArgs, st);
+          }
+        }
+
+        /* Signal that idling finished. Does NOT wait for activity completion. */
+        proc stopIdleServer() throws {
+          if commandMap.contains("auto_checkpoint") {
+            const msgArgs = new owned MessageArgs();
+            msgArgs.addPayload(b"idle stop");
```
e-kayrakli, line 758: I think this is not wired in right now. In general, it is not clear to me why this is necessary. I am probably missing something about asynchrony, `begin` etc.

src/ServerDaemon.chpl
```diff
@@ -724,6 +736,30 @@ module ServerDaemon {
                                                                             elapsed));
             this.shutdown(); 
         }
+
+        /* While waiting for a message from client, do checkpointing if desired. */
+        proc startIdleServer() throws {
+          // Do nothing if the checkpointing module was not included.
+          if commandMap.contains("auto_checkpoint") {
```
e-kayrakli, line 743: I have preference towards putting this and `needMemBasedCkpt` in a helper and calling that helper in the very top level `while` loop. Because it looks like we may end up calling a lot of functions that are no-ops. Function call is OK, but I think putting checks that may result in no-op in the top-level makes the code easier to read and maintain IMO.

e-kayrakli (commented):
Nice! Thanks for the quick turnaround.

Just so that I am understanding it right, this is not about server being idle, but it is strictly about memory percentage. Is that right? I am mostly confused by the variable names. Your PR title and message are clear on that.

> wait for 5 seconds (for example) idle time so that checkpointing does not occur during intermediate computations. For example, ak.ones(nnn) takes two messages and the checkpoint will currently occur between them if threshold is exceeded.

Good point. Also considering that we want to add the feature of checkpointing after N minutes of an idle server, the logic will eventually be like:

- if memory is above threshold, set the checkpoint timeout (e.g. ZMQ recv timeout) to 5 seconds
- otherwise, set it to 5 minutes

does that sound right?

> I would also like to move activityMutex to a "common header" file like MultiTypeSymbolTable

There must be a better place to put that. Otherwise, we can create a new module.

vasslitvinov:
Superseded by #4391. Closing.

PR #4379, opened 2025-05-07
Fixes bug involving arange and kwargs -- Closes 4321

### LINKED ISSUE #4321: `arange` doesn't handle `kwargs` properly
**Describe the bug**
When keyword arguments are passed to arange, you always have to pass `start`, `stop`, `stride`. Otherwise, you hit implementation issues

**To Reproduce**
`ak.arange(start=0, stop=10)`

**Error Message**
`UnboundLocalError: cannot access local variable 'stride' where it is not associated with a value`

**Is this a Blocking Issue**
No

**Additional context**
I think `arange` shouldn't have `args` and `kwargs` but instead the documented 4 arguments explicitly. Right now, optional arguments are handled with `len(args)` which is the reason for this issue.

While here, I am also seeing that what `np.arange` calls `step`, we call `stride`. The latter is Chapel term and probably the reason why it was chosen. I think we should change it to `step` to match numpy, though.

drculhane:
Closes #4321 

This addresses the bug Engin noted in Issue 4321.  It no longer uses kwargs, and it should behave the same way numpy does.  That is, I've tested it with:

pda = ak.arange(10)         # same as if you said ak.arange(0,10,1)
pda = ak.arange(1,10)     # same as if you said ak.arange(1,10,1)
pda = ak.arange(1,10,2)

and in each case, it treats the numbers the same way numpy does.

There is still some anomalous behavior in here.  It only handles integers (I don't know the history behind that), and removing the **kwarg aspect caused mypy issues because xp.arange calls it with floats.  For that matter, xp.arange claims to return floats, but it doesn't.  In fact, test_flip and test_roll fail if it returns floats.

Anomalous behavior aside (and perhaps that's for another issue), I believe this particular bug is fixed.

### DIFF
--- a/arkouda/numpy/pdarraycreation.py
+++ b/arkouda/numpy/pdarraycreation.py
@@ -912,26 +912,31 @@ def full_like(pda: pdarray, fill_value: numeric_scalars) -> Union[pdarray, Strin
     return full(tuple(pda.shape), fill_value, pda.dtype, pda.max_bits)
 
 
-def arange(*args, **kwargs) -> pdarray:
+def arange(
+    start: Optional[int_scalars] = None,
+    stop: Optional[int_scalars] = None,
+    step: Optional[int_scalars] = None,
+    dtype: Optional[Union[np.dtype, type, bigint]] = None,
+    max_bits: Optional[int] = None,
+) -> pdarray:
     """
-    arange([start,] stop[, stride,] dtype=int64)
+    arange([start,] stop[, step,] dtype=int64)
 
     Create a pdarray of consecutive integers within the interval [start, stop).
     If only one arg is given then arg is the stop parameter. If two args are
     given, then the first arg is start and second is stop. If three args are
-    given, then the first arg is start, second is stop, third is stride.
+    given, then the first arg is start, second is stop, third is step.
 
     The return value is cast to type dtype
 
     Parameters
     ----------
-    start: int_scalars, optional
-        Starting value (inclusive)
-    stop: int_scalars
-        Stopping value (exclusive)
-    stride: int_scalars, optional
-        The difference between consecutive elements, the default stride is 1,
-        if stride is specified then start must also be specified.
+    start : int_scalars, optional
+    stop  : int_scalars, optional
+    step  : int_scalars, optional
+        if one of these three is supplied, it's used as stop, and start = 0, step = 1
+        if two of them are supplied, start = start, stop = stop, step = 1
+        if all three are supplied, start = start, stop = stop, step = step
     dtype: np.dtype, type, or str
         The target dtype to cast values to
     max_bits: int
@@ -940,14 +945,16 @@ def arange(*args, **kwargs) -> pdarray:
     Returns
     -------
     pdarray
-        Integers from start (inclusive) to stop (exclusive) by stride
+        Integers from start (inclusive) to stop (exclusive) by step
 
     Raises
     ------
+    ValueError
+        Raised if none of start, stop, step was supplied
     TypeError
-        Raised if start, stop, or stride is not an int object
+        Raised if start, stop, or step is not an int object
     ZeroDivisionError
-        Raised if stride == 0
+        Raised if step == 0
 
     See Also
     --------
@@ -955,7 +962,7 @@ def arange(*args, **kwargs) -> pdarray:
 
     Notes
     -----
-    Negative strides result in decreasing values. Currently, only int64
+    Negative steps result in decreasing values. Currently, only int64
     pdarrays can be created with this method. For float64 arrays, use
     the linspace method.
 
@@ -976,30 +983,19 @@ def arange(*args, **kwargs) -> pdarray:
     """
     from arkouda.numpy import cast as akcast
 
-    # if one arg is given then arg is stop
+    if start is None and stop is None and step is None:
+        raise ValueError("A stopping value must be supplied to arange.")
 
-    if len(args) == 1:
-        start = 0
-        stop = args[0]
-        stride = 1
+    if step is None:
+        step = 1
+        if stop is None:
+            stop = start
+            start = 0
 
-    # if two args are given then first arg is start and second is stop
-    if len(args) == 2:
-        start = args[0]
-        stop = args[1]
-        stride = 1
-
-    # if three args are given then first arg is start,
-    # second is stop, third is stride
-    if len(args) == 3:
-        start = args[0]
-        stop = args[1]
-        stride = args[2]
-
-    if stride == 0:
+    if step == 0:
         raise ZeroDivisionError("division by zero")
 
-    dtype = akint64 if "dtype" not in kwargs.keys() else kwargs["dtype"]
+    aktype = akint64 if dtype is None else dtype
 
     # check the conditions that cause numpy to return an empty array, and
     # return one also.  This includes a fix needed for empty bigint arrays.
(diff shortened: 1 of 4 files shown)


arkouda/numpy/pdarraycreation.py
```diff
@@ -895,26 +895,31 @@ def full_like(pda: pdarray, fill_value: numeric_scalars) -> Union[pdarray, Strin
     return full(tuple(pda.shape), fill_value, pda.dtype, pda.max_bits)
 
 
-def arange(*args, **kwargs) -> pdarray:
+def arange(
+    start: Optional[int_scalars] = None,
+    stop: Optional[int_scalars] = None,
+    step: Optional[int_scalars] = None,
```
e-kayrakli, line 901: To the point about the anomalous behavior, what happens if we don't limit these arguments to integers? It seems the logic should be the same if one or more of these are floating points.

e-kayrakli (approved):
I had one comment about the xarray anomaly you noted. Otherwise this looks good! Thanks for taking care of it!

1RyanK (approved):
I think this is fine - maybe we make a separate ticket to straighten out the `array_api` and `numpy` `dtype` issues. FWIW `array_api` `arange` does not infer the data type, and [it's supposed to](https://data-apis.org/array-api/latest/API_specification/generated/array_api.arange.html#array_api.arange) (well, I guess it probably assumes that we would, but then for whatever reason, the tests are apparently predicated on it not behaving in that way, and the docs suggest it should assume the dtype is `float64`). So if you want to untangle everything and fix the tests here, that's fine, but I'd also be okay with a separate ticket so you don't feel the need to immediately go back into a rabbit hole - especially since the ticket sort of suggests the scope should be limited to the `kwargs` problem.

arkouda/numpy/pdarraycreation.py
drculhane: This answer is really more long-winded than I mean it to be, so I apologize in advance.

More than once, in the year+ that I've been on arkouda, I've thought "this isn't how this function should work.  I'll just fix it."  Then there's a ripple effect of other things breaking.  That happened the last time I looked at arange.  It happened again this time, and it also happened with the abs function inside the array_api.  What I've pushed up here is something of a compromise between the way it works and the way it should work.

I don't know the history of why ak.arange was written to only handle integers.  I'm inclined to merge this PR, and create yet another one to handle the data types correctly (similar to what Ryan says in a comment below).

arkouda/numpy/pdarraycreation.py
```diff
@@ -993,34 +989,34 @@ def arange(*args, **kwargs) -> pdarray:
     # the requested dtype.
     # This matters for several tests in tests/series_test.py
 
-    if (start == stop) | ((np.sign(stop - start) * np.sign(stride)) <= 0):
-        return akcast(array([], dtype=akint64), dt=dtype)
+    if (start == stop) | ((np.sign(stop - start) * np.sign(step)) <= 0):  # type: ignore
+        return akcast(array([], dtype=akint64), dt=aktype)
```
ajpotts, line 993: You can eliminate the need for the `# type: ignore` with the following:
```
def arange(
    *args: int_scalars,
    dtype:    Optional[Union[np.dtype, type, bigint]] = None,
    max_bits: Optional[int]                       = None,
) -> pdarray:
    from arkouda.numpy import cast as akcast
    # … your imports & early error‐checks …

    # 1) declare your locals as int_scalars
    start: int_scalars
    stop:  int_scalars
    step:  int_scalars

    # 2) assign them *individually*
    if len(args) == 1:
        start = 0           # 0 is an int, which is allowed in int_scalars
        stop  = args[0]     # args[0] is int_scalars
        step  = 1           # 1 is int → int_scalars
    elif len(args) == 2:
        start = args[0]
        stop  = args[1]
        step  = 1
    else:
        start = args[0]
        stop  = args[1]
        step  = args[2]

    # from here on, mypy knows start/stop/step all have type int_scalars
    if step == 0:
        raise ZeroDivisionError("division by zero")
```

arkouda/numpy/pdarraycreation.py
```diff
@@ -958,30 +965,19 @@ def arange(*args, **kwargs) -> pdarray:
...
+        if stop is None:
+            stop = start
+            start = 0
 
-    if stride == 0:
+    if step == 0:
         raise ZeroDivisionError("division by zero")
 
-    dtype = akint64 if "dtype" not in kwargs.keys() else kwargs["dtype"]
+    aktype = akint64 if dtype is None else dtype
```
ajpotts, line 980: This should be:
```
    from arkouda.dtypes import dtype as akdtype
    aktype = akdtype( akint64 if dtype is None else dtype)
```
Also, it would eliminate the need for the other `#type: ignore`

ajpotts (approved):
Other than the comments above, looks good.  Thanks for fixing this!

arkouda/numpy/pdarraycreation.py
```diff
@@ -895,26 +895,31 @@ def full_like(pda: pdarray, fill_value: numeric_scalars) -> Union[pdarray, Strin
     return full(tuple(pda.shape), fill_value, pda.dtype, pda.max_bits)
 
 
-def arange(*args, **kwargs) -> pdarray:
+def arange(
+    start: Optional[int_scalars] = None,
```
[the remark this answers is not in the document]
drculhane: That's exactly what I coded.

arkouda/numpy/pdarraycreation.py
e-kayrakli: Makes sense. Thanks!

arkouda/numpy/pdarraycreation.py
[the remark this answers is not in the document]
ajpotts: In this part here:
```
    start: Optional[int_scalars] = None,
    stop: Optional[int_scalars] = None,
    step: Optional[int_scalars] = None,
```
the default is set to `None` and that increases the type errors.

jaketrookman (approved):

arkouda/numpy/pdarraycreation.py
drculhane: I really don't want to do this.  It's almost identical to the code I just deleted.

I'm going to do my best to put together a concise description of what I've been doing on this.

ak.arange originally used **kwargs.  Engin suggested moving away from that, and I agreed.

As written, ak.arange requires Int arguments, but a function in the array_api was sending it floats.  This wasn't flagged as an error, because evidently, mypy can't "see through" **kwargs.  When I stopped using **kwargs, errors appeared, and it took some time to figure out why.

If we switch to **args, we're back in the same position.  Mypy won't flag type mismatches, so we will have effectively have done an invisible "ignore."  I much prefer "ignores" to be explicit.

BUT ... as usual, there are ripple effects going on here.  I've drafted another version of ak.arange which doesn't limit the inputs to integers, and MANY tests are failing.

I'm inclined to pull this back to draft status and address everything at once.

ajpotts:
For references, this is how numpy handles it: https://github.com/numpy/numpy/blob/main/numpy/_core/multiarray.pyi#L974

arkouda/numpy/pdarraycreation.py
ajpotts: OK, I will put up a separate PR `# type: ignore`.

ajpotts (approved):

jaketrookman (approved):

PR #4384, opened 2025-05-08
Fix CSV reading failures due to array misuse

ShreyasKhandekar:
Address issues in CSV reading caused by incorrect usage of `makeDistArray`, replacing it with a non-distributed array creation method.

This resolves failures in CSV nightly testing

### DIFF
--- a/src/CSVMsg.chpl
+++ b/src/CSVMsg.chpl
@@ -549,7 +549,7 @@ module CSVMsg {
     proc generate_subdoms(filenames: [?D] string, row_counts: [D] int, validFiles: [D] bool) throws {
         var skips = new set(string);
         var offsets = (+ scan row_counts) - row_counts;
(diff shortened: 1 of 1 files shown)


ShreyasKhandekar:
I have a bigger discussion about the use of `makeDistArray` in general and what we can and should do to reduce how much we use it. However, I am not prepared to talk about it this week, so I want to get this PR in to at least resolve testing failures.

As a preview, I am of the opinion that we use `makeDistArray` too much in Arkouda, often forcing the creation of distributed arrays where it's not required, sometimes it causes crashes, others it just makes for bad code. We should identify a different approach that preserves the benefits of the `tryCreateArray` interface without using the heavy hammer of "all arrays should be distributed when Arkouda is built in a distributed setting".

e-kayrakli (approved):

ajpotts (approved):

PR #4385, opened 2025-05-08
Part 3 of #3964:  Refactor benchmark handling of running mode

ajpotts:
This PR separates out the handling of `CLASS_SERVER` mode and `GLOBAL_SERVER` mode in the `conftest.py` for both the unit tests and the benchmarks.  This corrects an error where the `GLOBAL_SERVER` mode did not start an arkouda server and therefore duplicated the behavior of `CLIENT` mode.

`CLIENT` mode should not start a sever, but assumes arkouda can connect to a pre-existing server.  `GLOBAL_SERVER` mode should start a server once and be used for all tests across all modules.  In `CLASS_SERVER` mode, a new server will start before the tests for each module and stop after the tests for that module are completed.

Part 3 of #3964:  Refactor benchmark handling of running mode

### DIFF
--- a/benchmark_v2/conftest.py
+++ b/benchmark_v2/conftest.py
@@ -1,5 +1,7 @@
 import importlib
+import importlib.util
 import os
+from typing import Iterator
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


vasslitvinov (approved):
Good cleanup/fix!

jaketrookman (approved):
Looks good

Issue #2384, opened 2023-04-18
Checkpoint Functionality


Ethan-DeBandi99:
This issue is intended for tracking of researching and implementing checkpoint functionality in Arkouda. There are a few possibilities here and we may need to combine various options in our final solution. The tasks below are intended to be broken into separate tasking.

- [ ] Checkpointing using HDF5/Parquet. There are a few question here as far as performance and implementation. The main idea would be to write everything in the symbol table out to HDF5/Parquet files. A restore command would then be used to pull this data back in. We will need to look into how we should handle getting this data back to the client.
- [ ] Using Binary IO to write the Chapel objects out and load them back in. This should have better performance than using HDF5 or Parquet. However, there is still the question of how to effectively get this back to the client on restore.
- [ ] Determine if the checkpoint should be written to Network attached memory or local node memory. This is a question of performance.
- [ ] Determine if checkpointing should be done automatically behind the scenes or if it should be user callable from the client. Allowing the user to call it would enable things like naming which would make it easier to restore to the client.

vasslitvinov:
We should anticipate and handle gracefully the situation where auto-checkpointing encounters an error, not completing, then continues trying again and again encountering the same error. An important part of it is letting the user know, via the client, that this is happening, see ex. #4577.

PR #4391, opened 2025-05-09
Part of #2384: Auto checkpoint upon exceeding memory percentage or idle time

vasslitvinov:
Address part of #2384. Replaces #4378.

The server now automatically saves a checkpoint if memory usage exceeded a threshold or if the server has been idle for a bit. This is disabled by default, see below. Requires `CheckpointMsg` module. Arkouda still builds without this module.

The following server options in `CheckpointMsg` module control the behavior:

* `checkpointMemPct` : Perform checkpointing automatically when memory usage exceeds this many percent of available memory. By default / if <=0, auto-checkpointing is not triggered by memory usage.

* `checkpointMemPctDelay` : When memory exceeds the `checkpointMemPct` threhsold, wait for this many seconds of idle time before checkpointing. By default / if <=0, uses 5 seconds.

* `checkpointIdleTime` : Perform checkpointing automatically when the server is idle for this many seconds. By default / if <=0, auto-checkpointing is not triggered by idle time.

* `checkpointCheckInterval` : The auto-checkpoint implementation will wake up every this many seconds to check if a checkpoint needs to be saved due to memory usage or idle time. By default / if <=0, uses min( checkpointMemPctDelay, checkpointIdleTime ).

* `checkpointInterval` : Automatic checkpointing due to memory use or idle time will wait for at least this many seconds after any completed checkpoint save or load operations, whether automatic or client-initiated. By default / if <=0, uses 3600 (one hour). This avoids overly-frequent auto-checkpointing. The server does not need not be idle during this interval. Checkpoint operations requested by the client are not subject to this delay. 

If the server receives a message from the client while auto-checkpointing is in progress, it locks until auto-checkpointing completes.

This PR also introduces a framework for client testing with customer server options and invokes it as a sibling `pytest` command in `Makefile`. Such testing is configured using `pytest.opts.ini` and `tests/optioned-server/conftest.py` . It is used to test the functionality added in this PR.

While there: switch the order of error() and critical() in `src/Logging.chpl` to match the order in the enum; an analogous change in `src/Message.chpl`; open the file in read mode, not append, in `io_util.delimited_file_to_dict()`; added `io_util.directory_exists()`.

Next features to implement:
* The previous checkpoint is saved until the new checkpoint has successfully completed.
* The server informs the client that it is auto-saving a checkpoint.
* Tune the default thresholds, esp. `checkpointInterval` .
* Do not checkpoint until the first client connects to the server and has some activity.

The module-level variable `lastCkptCompletion` in CheckpointMsg.chpl should be specific to each `DefaultServerDaemon` instance rather than be a global. This is OK for now that we have only a single instance.

### DIFF
--- a/arkouda/io_util.py
+++ b/arkouda/io_util.py
@@ -66,6 +66,7 @@ def delimited_file_to_dict(path: str, delimiter: str = ",") -> Dict[str, str]:
     Return a dictionary populated by lines from a file where
     the first delimited element of each line is the key and
     the second delimited element is the value.
+    If the file does not exist, return an empty dictionary.
 
     Parameters
     ----------
@@ -84,16 +85,21 @@ def delimited_file_to_dict(path: str, delimiter: str = ",") -> Dict[str, str]:
     ------
     UnsupportedOperation
         Raised if there's an error in reading the file
+    ValueError
+        Raised if a line has more or fewer than two delimited elements
 
     """
     values: Dict[str, str] = {}
 
-    with open(path, "a+") as f:
-        f.seek(0)
-        for line in f:
-            line = line.rstrip()
-            key, value = line.split(delimiter)
-            values[key] = value
+    try:
+        with open(path, "r") as f:
+            for line in f:
+                line = line.rstrip()
+                key, value = line.split(delimiter)
+                values[key] = value
+    except FileNotFoundError:
+        pass  # return an empty dictionary
+
     return values
 
 
@@ -152,3 +158,20 @@ def delete_directory(dir: str) -> None:
             shutil.rmtree(dir)
         except OSError as e:
             print("Error: %s - %s." % (e.filename, e.strerror))
+
+
+def directory_exists(dir: str) -> bool:
+    """
+    Return True if the directory exists.
+
+    Parameters
+    ----------
+    dir : str
+        The path to the directory
+
+    Returns
+    -------
+    True if the directory exists, False otherwise.
+
+    """
+    return isdir(dir)
--- a/src/CheckpointMsg.chpl
+++ b/src/CheckpointMsg.chpl
@@ -2,12 +2,13 @@ module CheckpointMsg {
   use FileSystem;
   use List;
   use ArkoudaJSONCompat;
-  import IO, Path;
+  import IO, Path, Time;
   import Reflection.{getModuleName as M,
                      getRoutineName as R,
                      getLineNumber as L};
 
   use ServerErrors, ServerConfig;
+  import ServerDaemon.DefaultServerDaemon;
   use MultiTypeSymbolTable, MultiTypeSymEntry;
   use Message;
   use ParquetMsg;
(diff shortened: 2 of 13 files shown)


arkouda/io_util.py
```diff
@@ -152,3 +152,16 @@ def delete_directory(dir: str) -> None:
             shutil.rmtree(dir)
         except OSError as e:
             print("Error: %s - %s." % (e.filename, e.strerror))
+
+
+def directory_exists(dir: str) -> bool:
```
ajpotts, line 157: If you add a new function to arkouda it should have a unit test.

ajpotts (approved):

src/CheckpointMsg.chpl
```diff
@@ -128,9 +180,138 @@ module CheckpointMsg {
...
 
     }
+    updateLastCkptCompletion();
     return Msg.send(nameArg);
   }
 
+  // the smaller of those argument that are positive
+  private proc minOfPos(x, y: x.type) {
+    return if x > 0 then if y > 0 then min(x, y) else x
+                    else if y > 0 then y else 0: x.type;
```
ShreyasKhandekar, line 190: Can we make this simpler to read by having 4 return statements instead of just one?
```suggestion
  if x <= 0 && y <= 0 then return 0: x.type;
  if x <= 0 then return y;
  if y <= 0 then return x;
  return min(x, y);
```

src/CheckpointMsg.chpl
```diff
@@ -128,9 +180,138 @@ module CheckpointMsg {
...
+          // Fulfill all waiting periods before a next checkpoint.
+          delay = maxRequestedDelay;
+          continue;
+        }
+
+        // Nothing came up. Check back later.
+        delay = checkpointCheckInterval;
+
+      } catch err {
+        // We can't do anything with errors, so ignore them.
```
ShreyasKhandekar, line 293: We should log these errors using the logger for this module

ShreyasKhandekar (approved):

tests/optioned-server/conftest.py
```diff
@@ -0,0 +1,104 @@
...
+    else:
+        if not pytest.module_server_launched:
+            raise RuntimeError("neither " + _module_args(r) + "nor " + _class_args(r) + "is given")
+
+        yield
+
+
+# This fixture in the parent conftest.py runs ak.connect() and ak.disconnect():
+#  @pytest.fixture(scope="class", autouse=True)
+#  def manage_connection(_class_server):
```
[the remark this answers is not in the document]
vasslitvinov: @ShreyasKhandekar I added this as a breadcrumb to help future readers understand where ak.connect/disconnect happen.

ShreyasKhandekar (approved):

arkouda/io_util.py
ajpotts: Thanks!

arkouda/io_util.py
```diff
@@ -84,6 +84,8 @@ def delimited_file_to_dict(path: str, delimiter: str = ",") -> Dict[str, str]:
     ------
     UnsupportedOperation
         Raised if there's an error in reading the file
+    ValueError
+        Raised if a line has more or fewer than two delimited elements
```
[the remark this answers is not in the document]
ShreyasKhandekar: Resolved

src/CheckpointMsg.chpl
ShreyasKhandekar: Resolved

src/CheckpointMsg.chpl
ShreyasKhandekar: Resolved

ajpotts (approved):

PR #4403, opened 2025-05-14
Detect server exit while reading connection file

vasslitvinov:
The testing helper `start_arkouda_server` will now raise a `RuntimeError` if the `arkouda_server` process exits without creating the "connection info" file. Previously it would wait indefinitely for this file to be created.


e-kayrakli (approved):
I am not sure if this code is linted, but line lengths might hit a flake8 issue.

ajpotts (approved):

PR #4406, opened 2025-05-15
Adds negative axis handling to squeeze

### LINKED ISSUE #4405: Add negative axis handling to squeeze function
The squeeze function (used in functions that compute along slices, when keepdims=False) doesn't handle negative axes, which it should, since python does.

drculhane:
Closes #4405, a sub-issue of 4125.  Numpy functions allow for negative numbers in the axis argument (e.g. -1 means the last axis, -2 the next-to-last, etc).  Arkouda has such functionality server-side, but it's not used consistently.  This PR is part of a series that aims to correct that.

### DIFF
--- a/src/ManipulationMsg.chpl
+++ b/src/ManipulationMsg.chpl
@@ -834,7 +834,6 @@ module ManipulationMsg {
     const name = msgArgs["name"],
           nAxes = msgArgs["nAxes"].toScalar(int),
           axes = msgArgs["axes"].toScalarArray(int, nAxes);
-
     var eIn = st[name]: borrowed SymEntry(array_dtype, array_nd_in),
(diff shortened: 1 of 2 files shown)


ajpotts:
Looks good.  Would you please add a unit test to reproduce the bug?

drculhane:
Re: unit test.  In fact, I had just written a unit test for squeeze, while working on the same axis issue in the repeat function.  I'll add that test to this one.

ajpotts (approved):
Thanks for fixing this!

ajpotts:
Would you please add a description?  It can be brief.

jaketrookman (approved):

drculhane:
Description added.

ShreyasKhandekar (approved):

PR #4408, opened 2025-05-15
Adds negative axis coverage to ak.repeat, ak.argmin, ak.argmax, Closes 4407 and 4421

### LINKED ISSUE #4407: Add negative axis handling to repeat
The repeat function is one of the many that currently handles negative axes wrong.

drculhane:
Closes #4407 and #4421

Numpy functions allow for negative numbers in the axis argument (e.g. -1 means the last axis, -2 the next-to-last, etc). Arkouda has such functionality server-side, but it's not used consistently. This PR is part of a series that aims to correct that.

Update: also adds negative axis coverage for argmin and argmax.  Renames the new chapel function validateNegativeAxis to validateNegativeAxes, since "Axes" already existed, and this is just a case of overloading it.


ajpotts:
Would you please add a description the PR?

tests/numpy/manipulation_functions_test.py
```diff
@@ -104,7 +104,7 @@ def test_repeat_dim_3(self, dtype, shape: Tuple[int, ...]):
         f = ak.repeat(a, reps)
         np_f = np.repeat(np_a, reps.to_ndarray())
         assert_arkouda_array_equivalent(np_f, f)
-        for axis in range(3):
+        for axis in range(-2, 3):
```
1RyanK, line 107: Should this be `for axis in range(-3, 3):`? `for axis in range(3)` was testing 0, 1, and 2. But negatively, we're only testing -2 and -1. So we should start at -3, which would correspond to axis 0, I think.

tests/numpy/manipulation_functions_test.py
```diff
@@ -133,7 +133,7 @@ def test_repeat_dim_2(self, dtype, shape: Tuple[int, ...]):
         f = ak.repeat(a, reps)
         np_f = np.repeat(np_a, reps.to_ndarray())
         assert_arkouda_array_equivalent(np_f, f)
-        for axis in range(2):
+        for axis in range(-1, 2):
```
1RyanK, line 136: Same question here.

1RyanK (approved):
Looks good!

tests/numpy/manipulation_functions_test.py
drculhane: Good catch, yes it should be.  I had a reason for using (-2, 3) that involved debugging, and then neglected to switch it back.  I'll get this and the one below on Monday.

src/ManipulationMsg.chpl
```diff
@@ -1036,7 +1036,10 @@ module ManipulationMsg {
            (d2.rank == 1)) {
     param pn = Reflection.getRoutineName();
 
-    var eOut = makeDistArray((...repeatShape(eIn.shape, reps, axis)), t);
+    const (valid, axis_) = validateNegativeAxes (axis,d.rank) ; // new
+    if !valid then throw new Error ("Invalid axis %i in repeat".format(axis)); // new
```
ajpotts, line 1040: I think the `// new` comment can be removed.

ajpotts (approved):
Thanks for taking care of this!

src/ManipulationMsg.chpl
drculhane: Oops.  Done.

ajpotts (approved):

PR #4431, opened 2025-05-20
Closes #4430:  failing numeric_tests

### LINKED ISSUE #4430: failing numeric_tests
**Describe the bug**
Failing: `python3 -m pytest -c pytest.ini tests/numpy/numeric_test.py  --size=100`

**Error Message**
```
    def send_string_message(
        self,
        cmd: str,
        recv_binary: bool = False,
        args: Optional[str] = None,
        size: int = -1,
        request_id: Optional[str] = None,
    ) -> Union[str, memoryview]:
        message = RequestMessage(
            user=username, token=self.token, cmd=cmd, format=MessageFormat.STRING, args=args, size=size
        )
        # Note - Size is a placeholder here because Binary msg not yet support json args and
        # request_id is a noop for now
        logger.debug(f"sending message {json.dumps(message.asdict())}")
    
        self.socket.send_string(json.dumps(message.asdict()))
    
        if recv_binary:
            frame = self.socket.recv(copy=False)
            view = frame.buffer
            if bytes(view[0 : len(b"Error:")]) == b"Error:":
            return view
        else:
            raw_message = self.socket.recv_string()
            try:
    
E                   RuntimeError: Error executing command: TypeError Line 243 In Message.toScalar: Json Argument '92,233,720,368,547,758' cannot be cast as a scalar int(64)

```

ajpotts:
This PR fixes a bug causing multiple tests in `numeric_tests` to fail.  

Closes #4430:  failing numeric_tests


1RyanK (approved):
Looks good!

1RyanK:
This doesn't look too bad to fix. The first assertion error only comes up because they hardcoded in the old `NUMBER_FORMAT_STRINGS` to the test in tests/numpy/dtypes_test.py. For the other problem in the unit tests, you can just go into randint and cast `low` and `high` to the dtype prior to sending the message to the server. Although the way randint sends the message to the server strikes me as bizarre and I wonder if there's a better way to do it.

jaketrookman (approved):
Looks good

1RyanK (approved):

jaketrookman (approved):
Looks good as a fix for now

PR #4433, opened 2025-05-21
Fewer 'throws' while logging

vasslitvinov:
Switch logging methods to be non-throwing, by removing their dependence on the `format()`. This eliminates the need for `try!` when calling them and reduces (slightly) the size of the generated code.

While there, clean up and factor our the implementation; remove logger calls in CheckpointMsg that were added by mistake.

### DIFF
--- a/src/Logging.chpl
+++ b/src/Logging.chpl
@@ -75,11 +75,11 @@ module Logging {
     /*
      * getOutputHandler is a factory method for OutputHandler implementations.
      */
-    proc getOutputHandler(channel: LogChannel) : owned OutputHandler throws {
+    proc getOutputHandler(channel: LogChannel) : owned OutputHandler  {
         if channel == LogChannel.CONSOLE {
             return new owned ConsoleOutputHandler();
         } else {
-            return new owned FileOutputHandler("%s/.arkouda/arkouda.log".format(here.cwd()));
+            return new owned FileOutputHandler((try! here.cwd()) + "/.arkouda/arkouda.log");
         }
     }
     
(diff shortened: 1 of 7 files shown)


src/Logging.chpl
```diff
@@ -113,82 +102,60 @@ module Logging {
...
+        proc generateLogMessage(moduleName: string, routineName: string, lineNumber: int,
+                                level: LogLevel, msg: string...) {
+            var lineStr = if lineNumber != 0 then "Line " + lineNumber:string + " " else "";
+            var dateStr = if printDate then generateDateTimeString(" ") else "";
+            return (dateStr, "[", moduleName, "] ", routineName, " ", lineStr, level:string,
+                    " [Chapel ] ", (...msg));
         }
          
-        proc generateDateTimeString() throws {
+        proc generateDateTimeString(tail = "") {
```
ShreyasKhandekar, line 151: What is the purpose of the `tail` argument and why is it only needed in the `catch` block?

ShreyasKhandekar (approved):

src/Logging.chpl
vasslitvinov: `tail` allows this helper to include a trailing space in the string it produces.

It is only needed in the catch block because I forgot to include it in the try block. I will fix this.

vasslitvinov: Fixed as of c12932834

ShreyasKhandekar: Resolved

ajpotts (approved):
Thanks for cleaning this up!  We actually have a goal of eventually removing all `try!`s, so this helps a lot!

PR #4440, opened 2025-05-22
Closes #4235: searchsorted to work with uint type

### LINKED ISSUE #4235: searchsorted to work with uint type
After the Numpy 2.0 transition search_sorted no longer works for `uint64` dtype.  `uint64` was removed from `test_searchsorted` because it was causing an error:

```
self = <arkouda.client.ZmqChannel object at 0x7e101ce377e0>
cmd = 'searchSorted<float64,1,1>', recv_binary = False
args = '["{\\"key\\": \\"x1\\", \\"dtype\\": \\"float64\\", \\"val\\": \\"id_cxxDjvW_19\\"}", "{\\"key\\": \\"x2\\", \\"dtype...uint64\\", \\"val\\": \\"id_cxxDjvW_20\\"}", "{\\"key\\": \\"side\\", \\"dtype\\": \\"str\\", \\"val\\": \\"left\\"}"]'
size = 3, request_id = None

    def send_string_message(
        self,
        cmd: str,
        recv_binary: bool = False,
        args: Optional[str] = None,
        size: int = -1,
        request_id: Optional[str] = None,
    ) -> Union[str, memoryview]:
        message = RequestMessage(
            user=username, token=self.token, cmd=cmd, format=MessageFormat.STRING, args=args, size=size
        )
        # Note - Size is a placeholder here because Binary msg not yet support json args and
        # request_id is a noop for now
        logger.debug(f"sending message {json.dumps(message.asdict())}")
    
        self.socket.send_string(json.dumps(message.asdict()))
    
        if recv_binary:
            frame = self.socket.recv(copy=False)
            view = frame.buffer
            if bytes(view[0 : len(b"Error:")]) == b"Error:":
            return view
        else:
            raw_message = self.socket.recv_string()
            try:
    
E                   RuntimeError: Error executing command: cannot cast class to type - runtime types not compatible

../../arkouda/client.py:532: RuntimeError
```

Fix this and add `uint64` back into `test_searchsorted`.

1RyanK:
I think `searchsorted` should work with uint types now. The problem was that somehow, `v = ak.array(v, dtype) + shift` would turn `v` into an array of `float64` if `v` was originally `uint64` and `shift` was 0. I changed it to be `v = ak.array(ak.array(v, dtype) + shift, dtype)` (and similarly where we do the same thing to `a`) and now tests pass. This should probably be fixed in `binopvs` or wherever it is as well, I imagine this is probably not intended behavior.

I also added a check that the data types match in `searchsorted`.

Closes #4235: searchsorted to work with uint type

### DIFF
--- a/arkouda/numpy/sorting.py
+++ b/arkouda/numpy/sorting.py
@@ -324,6 +324,7 @@ def searchsorted(
         If `a` has more than one dimension.
     TypeError
         If `a` has an unsupported dtype (i.e., not int64, uint64, bigint, or float64).
+        If the dtype of `a` and `v` does not match
 
 
     Examples
@@ -352,6 +353,9 @@ def searchsorted(
     else:
         scalar_input = True
         v_ = cast(pdarray, array([v]))
+
+    if a.dtype != v_.dtype:
+        raise TypeError(f"The dtype of a ({a.dtype}) and v ({v_.dtype}) must match.")
 
     repMsg = generic_msg(
         cmd=f"searchSorted<{a.dtype},{a.ndim},{v_.ndim}>",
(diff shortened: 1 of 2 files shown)


drculhane (approved):
Looks good.

jaketrookman (approved):
Looks good

PR #4447, opened 2025-05-23
Fixes ak.full behavior, Closes 4312

drculhane:
This makes the dtype argument optional, defaulting to None.  If it is None, then the function uses the type of the supplied fill_value parameter.

It also fixes an error in test_full_error.  Because its use of ak.full wasn't supplying a fill_value, it was causing an error, but not the one it was supposed to.


ajpotts (approved):
Thanks for fixing this!

1RyanK (approved):
Looks good!

PR #4454, opened 2025-05-29
Fix host-related issues re start_arkouda_server()

vasslitvinov:
Fix a bug that prevents `tests/optioned-server/conftest.py` added in #4391 from connecting to the server that's not on "localhost".

While there, clarify the handling of the server host and port in the test suites:
* "Host" that comes from a command-line argument or environment variable and defaults to `localhost` is used only in the CLIENT running mode.
* "Host" is obtained from `start_arkouda_server()`, which reads it from `ak-server-info` that's created by the server, in GLOBAL_SERVER and CLASS_SERVER modes.
* "Port" comes from a command-line argument or environment variable and defaults to 5555; the port produced by `start_arkouda_server()` is expected to be the same.

The testing frameworks store the host and the port for use in `ak.connect()` as follows:

| suite | tests/ | benchmarks/ | benchmark_v2/ |
| :-- | :-- | :-- | :-- |
| host | `pytest.server` | `get_server_info()` | `pytest.host` |
| port | `pytest.port` | `get_server_info()` | `pytest.port` | 
| file | conftest.py | run_benchmarks.py | conftest.py |

The `host` argument of `start_arkouda_server()` has not been used for its [intended purpose](https://github.com/Bears-R-Us/arkouda/pull/374/files#r427210331) since it was added in #374. I propose to remove it, which will increase the clarity of code.

### DIFF
--- a/benchmark_v2/conftest.py
+++ b/benchmark_v2/conftest.py
@@ -24,7 +24,7 @@ def pytest_addoption(parser):
         "--host",
         action="store",
         default=default_host,
-        help="arkouda server host",
+        help="arkouda server host for CLIENT running mode",
     )
 
     default_port = int(os.getenv("ARKOUDA_SERVER_PORT", 5555))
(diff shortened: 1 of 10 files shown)


ShreyasKhandekar (approved):
Looks great! Thanks for making these improvements, Vass!

ajpotts:
Thank you!

jaketrookman (approved):
Looks good

1RyanK (approved):
Looks good!

PR #4456, opened 2025-05-29
skip doctest unit tests for string, categorial on hpe systems

ajpotts:
Skips doctest unit tests for `string`, `categorial` modules on hpe systems due to hanging tests.


jaketrookman (approved):
Looks good

PR #4457, opened 2025-05-29
add check mypy version to CI

ajpotts:
Add check mypy version step to the mypy test in the CI.  This helps with debugging `mypy` issues.


1RyanK (approved):
Looks good!

jaketrookman (approved):

PR #4471, opened 2025-05-30
Closes #3964:  Refactor benchmark handling of running mode.

### LINKED ISSUE #3964: Refactor benchmark handling of running mode

ajpotts:
This is a clean-up item that comments out arguments that are not currently being used int `benchmark_v2/conftest.py`.  I decided to comment out rather than delete because I think they could be useful in the future.  However, leaving them in only partially implemented will be confusing to the user.

Closes #3964:  Refactor benchmark handling of running mode.


drculhane (approved):
I suggest adding a comment about the deleted code (so whoever looks at it next will know the background), but I don't need to hold up approval waiting for that.  Straightforward work, approved.

1RyanK (approved):
Seconding Andy's comment, but otherwise looks good.

1RyanK (approved):

PR #4472, opened 2025-05-30
Closes #4468 aligning vecdot function to numpy

### LINKED ISSUE #4468: Align ak.vecdot to np.vecdot
The ak.vecdot function precedes the numpy vecdot function, which was only added in numpy version 2.0.0.  Their behavior doesn't match.

drculhane:
Closes #4468 

-----

Update to the update

I rebased, and seem to have included changes that were already in the most recent master.

-----

Update

The array_api tests had a test of vecdot which included the comment "because numpy doesn't have a vecdot," and so it compared to something that performed the old computation.  That comment has been removed, and the test now compares the result to np.vecdot.

-----

Still a draft, for one more round of tests.

The old vecdot has been removed.  The new vecdot matches the performance of numpy vecdot, and includes an optional axis parameter which can be used when the input arrays have matching shape.  This didn't require any chapel-side code, just the use of broadcasting, binops, and ak.sum along the last axis.

I've added a comment in the Notes section of the docstring describing how to replicate the previous vecdot computation.

### DIFF
--- a/tests/numpy/numeric_test.py
+++ b/tests/numpy/numeric_test.py
@@ -1538,7 +1538,7 @@ class TestNumeric:
     def test_vecdot_1d(self, dtype, size):
         seed = pytest.seed if pytest.seed is not None else 8675309
         a = ak.randint(0, 100, size, dtype=dtype, seed=seed)
-        b = ak.randint(0, 100, size, dtype=dtype)
+        b = ak.randint(0, 100, size, dtype=dtype, seed=seed + 1)
         np_vecdot = np.vecdot(a.to_ndarray(), b.to_ndarray())
         ak_vecdot_f = ak.vecdot(a, b)
         ak_vecdot_r = ak.vecdot(a, b)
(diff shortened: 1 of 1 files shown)


1RyanK (approved):
Looks good! My personal opinion is that we should match numpy behavior, so assuming no one is currently using the old version of vecdot I'd be okay with this overwriting the old one.

tests/numpy/numeric_test.py
```diff
@@ -1561,3 +1532,114 @@ def test_take_multidim(self, dtype, axis):
         anp_taken = np.take(anp, indices_np, axis=axis)
 
         assert np.array_equal(a_taken.to_ndarray(), anp_taken)
+
+    @pytest.mark.parametrize("dtype", NO_BOOL)
+    @pytest.mark.parametrize("size", pytest.prob_size)
+    def test_vecdot_1d(self, dtype, size):
+        seed = pytest.seed if pytest.seed is not None else 8675309
+        a = ak.randint(0, 100, size, dtype=dtype, seed=seed)
+        b = ak.randint(0, 100, size, dtype=dtype)
```
ajpotts, line 1541: Why does `a` have a seed but not `b`?

ajpotts (approved):
Great work.  Thanks for fixing this.  I like all the tests.

tests/numpy/numeric_test.py
drculhane: You know ... that's a darn good question.  I thought the first seed would suffice, and that b would be the same each time.  I was wrong about that.  So I've added a seed to the generation of b in all of the vecdot tests.

But that prompted me to go through the whole of numeric_test, and it seems to me that we're inconsistent with our use of seeds.  I think the point of seeds is to ensure we're always running the same test.  So ... this might be a worthy issue.

ajpotts: I think they should all use `pytest.seed` and then we can always run the tests with a different` pytest.seed` value.  Thanks for fixing this.

PR #4484, opened 2025-06-02
Closes #4483: result_type function that also handles bigint

### LINKED ISSUE #4483: `result_type` function that also handles `bigint`
**Feature Description**
Numpy has a `result_type` function that takes in arrays or data types and returns a data type that would be the data type of the output of doing most binary operations on the two input types. It would be nice to have something in arkouda that could also handle things like `bigint` as well.

1RyanK:
Creates `ak.result_type` which enables users to check the output of a `binop` python-side. Useful for the reworking of the `binop` functions.

Closes #4483: result_type function that also handles bigint


arkouda/numpy/dtypes.py
```diff
@@ -228,6 +230,71 @@ def can_cast(from_, to) -> builtins.bool:
...
+
+    Determine the result dtype that would be returned by a NumPy-like operation
+    on the provided input arguments, accounting for Arkouda's extended types
+    such as ak.bigint.
+
+    This function mimics numpy.result_type, with support for Arkouda types.
+
+    Parameters
+    ----------
+    *args: pdarray, dtype, or type
```
ajpotts, line 245: This should be `Union[pdarray, np.dtype, type]`

arkouda/numpy/dtypes.py
```diff
@@ -228,6 +230,71 @@ def can_cast(from_, to) -> builtins.bool:
...
+        One or more input objects. These can be NumPy arrays, dtypes, Python
+        scalar types, or Arkouda pdarrays.
+
+    Returns
+    -------
+    dtype : numpy.dtype
```
ajpotts, line 251: This should be `Union[np.dtype, type]`, exactly matching the function description.

arkouda/numpy/dtypes.py
```diff
@@ -228,6 +230,71 @@ def can_cast(from_, to) -> builtins.bool:
...
+            dt = arg
+        elif hasattr(arg, "dtype"):
+            dt = arg.dtype
+        else:
+            dt = np.result_type(arg)
+
+        # Normalize Arkouda custom dtypes
+        if dt == bigint:
+            has_bigint = True
+        elif dt == float64 or dt == np.float64 or dt is float:
```
ajpotts, line 277: Please use `_is_dtype_in_union(float, float_scalars)` instead, it will include all supported float types including `np.float32`.

arkouda/numpy/dtypes.py
```diff
@@ -228,6 +230,71 @@ def can_cast(from_, to) -> builtins.bool:
...
+            has_float = True
+            np_dtypes.append(np.dtype(np.float64))
+        elif dt == int64 or dt == np.int64 or dt is int:
```
ajpotts, line 280: Similar to above, use `_is_dtype_in_union(float, int_scalars)`

ajpotts (approved):
Love this!  Thanks for putting it together.

arkouda/numpy/dtypes.py
1RyanK: Done!

arkouda/numpy/dtypes.py
1RyanK: Done!

arkouda/numpy/dtypes.py
1RyanK: I switched it to `_is_dtype_in_union(dt, Union[float, np.float64, float64])`. If it evaluates to `True`, we are then doing `np_dtypes.append(np.dtype(np.float64))`, which only makes sense if the input should be taken as a 64 bit float. If the code is handed something with a `np.float32`, that is handled a couple of lines down:
```
elif isinstance(dt, np.dtype):
    if dt.kind == "f":
        has_float = True
    np_dtypes.append(dt)
```
Similar for the `int` situation.

PR #4514, opened 2025-06-03
Closes #4293 add doctest to array api constants module

### LINKED ISSUE #4293: add doctest to array_api/_constants module

ajpotts:
Adds `doctest` unit test for the `array_api/_constants` module.

Closes #4293 add doctest to array api constants module

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,6 +3,7 @@ filterwarnings =
     ignore:Version mismatch between client .*
 testpaths =
     tests/array_api/typing_test.py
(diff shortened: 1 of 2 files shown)


drculhane (approved):
Downloaded and ran this, and it works as advertised.  Line 17 of array_api/array_object.py has a pre-existing typo ("local" where it means "locale").  You might want to correct that while you're in there.  In either case, approved.

jaketrookman (approved):

PR #4557, opened 2025-06-05
Closes #4497: Create a repartitionByLocaleString function

### LINKED ISSUE #4497: Create a `repartitionByLocaleString` function
**Feature Description**
Given a segmented String and an array of integers representing destination locales, return what basically amounts to a segmented String for each locale (e.g. using `[PrivateSpace] List(uint(8))` for the bytes and then something similar for the offsets)

**Example Behavior**
Ideally this can be ripped straight out of `concatenate_uniquely`

1RyanK:
Moved some of the logic from my `concatenateUniqueStrMsg` function into the newly created `Repartition` module. No new unit tests because this is still tested by `test_concatenate_uniquely`.

There are three coforalls in the `Repartition` version, but I think ultimately, it is a small price to pay for the convenience of a helper function in this context. Moreover, I don't think any of the coforalls are particularly cumbersome.

I also revamped a `for` into two `forall`s, which if there are at least two threads, I think that's a win.

Closes #4497: Create a repartitionByLocaleString function

### DIFF
--- a/src/ConcatenateMsg.chpl
+++ b/src/ConcatenateMsg.chpl
@@ -22,6 +22,8 @@ module ConcatenateMsg
     use Set;
     use List;
 
+    use Repartition;
+
     private config const logLevel = ServerConfig.logLevel;
     private config const logChannel = ServerConfig.logChannel;
     const cmLogger = new Logger(logLevel, logChannel);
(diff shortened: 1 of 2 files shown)


drculhane (approved):
Whew.  That's a ton of code, and it all looks good.  Nice work.

ajpotts (approved):
Thanks for this one!

PR #4563, opened 2025-06-09
Closes #4460: Simplify and extend logic in binopvs

### LINKED ISSUE #4460: Simplify and extend logic in `binopvs`
**Feature Description**
Part of issue #4448, but restricted to just `binopvs`

### LINKED ISSUE #4461: Simplify and extend logic in `binopsv`
**Feature Description**
Part of issue #4448, but restricted to just `binopsv`

1RyanK:
Applied mostly the same changes to `binopvs` and `binopsv` from `binopvv`. This could possibly benefit from being collapsed into one function that works regardless of vector or scalar.

Closes #4460: Simplify and extend logic in binopvs
Closes #4461: Simplify and extend logic in binopsv

### DIFF
--- a/arkouda/numpy/dtypes.py
+++ b/arkouda/numpy/dtypes.py
@@ -219,8 +219,13 @@ def can_cast(from_, to) -> builtins.bool:
 
     """
     if isSupportedInt(from_):
-        if (from_ < 2**64) and (from_ >= 0) and (to == dtype(uint64)):
-            return True
+        if isinstance(from_, int):
+            if (from_ < 2**64) and (from_ >= 0) and (to == dtype(uint64)):
+                return True
+            elif (from_ < 2**63) and (from_ >= -(2**63)) and (to == dtype(int64)):
+                return True
+    elif isSupportedFloat(from_):
+        return _is_dtype_in_union(to, float_scalars)
 
     if (np.isscalar(from_) or _is_dtype_in_union(from_, numeric_scalars)) and not isinstance(
         from_, (int, float, complex)
@@ -255,7 +260,7 @@ def result_type(*args: Union[pdarray, np.dtype, type]) -> Union[np.dtype, type]:
     Notes
     -----
     This function is meant to be a drop-in replacement for numpy.result_type
-    but includes logic to support Arkouda's bigint types.
+    but includes logic to support Arkouda's bigint type.
     """
     from numpy.typing import DTypeLike
 
--- a/arkouda/numpy/pdarrayclass.py
+++ b/arkouda/numpy/pdarrayclass.py
@@ -21,6 +21,7 @@ from arkouda.numpy.dtypes import get_byteorder, get_server_byteorder
 from arkouda.numpy.dtypes import int64 as akint64
 from arkouda.numpy.dtypes import (
     int_scalars,
+    isSupportedBool,
     isSupportedInt,
     isSupportedNumber,
     numeric_scalars,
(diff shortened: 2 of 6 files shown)


src/BinOp.chpl
```diff
@@ -297,695 +297,283 @@ module BinOp
...
-          }
+
+          when "<" { e = (l.a: castType) < (val: castType); }
+          when "<=" { e = (l.a: castType) <= (val: castType); }
+          when ">" { e = (l.a: castType) > (val: castType); }
+          when ">=" { e = (l.a: castType) >= (val: castType); }
+          when "==" { e = (l.a: castType) == (val: castType); }
+          when "!=" { e = (l.a: castType) != (val: castType); }
+          otherwise do return MsgTuple.error(nie); // Shouldn't happen
+
```
ajpotts, line 335: Throughout this function, in most cases everything is cast every time.  Will casting something that doesn't need to be cast cause it to be slower?

ajpotts (approved):

arkouda/numpy/dtypes.py
```diff
@@ -219,8 +219,13 @@ def can_cast(from_, to) -> builtins.bool:
...
     if isSupportedInt(from_):
-        if (from_ < 2**64) and (from_ >= 0) and (to == dtype(uint64)):
-            return True
+        if isinstance(from_, int):
+            if (from_ < 2**64) and (from_ >= 0) and (to == dtype(uint64)):
+                return True
+            elif (from_ < 2**63) and (from_ >= -(2**63)) and (to == dtype(int64)):
+                return True
+    elif isSupportedFloat(from_):
+        return _is_dtype_in_union(to, float_scalars)
```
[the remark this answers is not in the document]
1RyanK: If I revert it, `test_pdarray_and_scalar_ops` in `operator_test.py` fails. I think I had to change the logic in `_binop` and this seemed to be the easiest way to fix it.

src/BinOp.chpl
e-kayrakli: I wouldn't expect a noticeable hit for primitive types, but this would be good to confirm somehow. FWIW, the forall loop the compiler generates for these expressions look like the cast is being done element-wise, instead of creating a full-blown copy of the array with elements cast to `castType`, which would really kill performance. From that point, the compiler is doing what it supposed to do.

src/BinOp.chpl
```diff
@@ -297,695 +297,283 @@ module BinOp
...
-        }
-      } else if etype == real {
+
+      if lType == bool && rType == bool {
         select op {
-          // True division is the only integer type that would result in a
-          // resultant type of `real`
-          when "/" {
-            e = l.a:real / val:real;
-          }
```
e-kayrakli, line 459: I can't really follow the diff and see whether this special-case is lost or somewhere else.

src/BinOp.chpl
```diff
@@ -297,695 +297,283 @@ module BinOp
...
-        when ">>" {
-          if(val){
-            e = l.a:int >> val:int;
-          }else{
-            e = l.a:int;
-          }
+        when "*" { e = (l.a: etype * val: etype): etype; }
+        when "+" { e = (l.a: etype + val: etype): etype; }
+        when "-" { e = (l.a: etype - val: etype): etype; }
+        when "/" { e = ((l.a: etype) / (val: etype)): etype; }
```
e-kayrakli, line 375: Why did you need a different parentheses here? Is there something off with operator precedence?

src/BinOp.chpl
1RyanK: This is handled by `src/BinOp.chpl` line 375 in the new code. There are a few main ifs that determine what happens. They are:
```Chapel
if etype == bool {
...
else if lType == bool && rType == bool && etype == uint(8) {
...
else if etype == real(32) || etype == real(64) {
...
else {
...
```
If the operator is `/`, and the two types aren't both `bool`, then the output is guaranteed to be guaranteed to be `real(64)`. I included `real(32)` because if the list of supported data types gets expanded, then this will handle everything automatically.

arkouda/numpy/pdarrayclass.py
```diff
@@ -776,18 +767,20 @@ def _r_binop(self, other: pdarray, op: str) -> pdarray:
...
-            # do the cast so that return array will have same dtype
-            dt = self.dtype.name
-            other = self.dtype.type(other)
         if dt not in DTypes:
             raise TypeError(f"Unhandled scalar type: {other} ({type(other)})")
+
+        from arkouda.numpy.dtypes import can_cast as ak_can_cast
+
+        if self.dtype != "bigint" and ak_can_cast(other, self.dtype):
+            dt = str(self.dtype)
```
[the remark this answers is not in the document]
1RyanK: Done!

src/BinOp.chpl
1RyanK: I took them out. Sometimes when I'm writing code, I put in extra parentheses so I'm certain that it runs the way I want it to, and sometimes I don't. But I imagine if it's working for the three lines above this, it should work here as well.

e-kayrakli: I am definitely for using parentheses sometimes redundantly, I completely agree with that. But I do feel even more strongly for consistent coding style :) If you feel more comfortable with parentheses, by all means use them, but use them on the lines above, too. ;)

PR #4568, opened 2025-06-10
Closes-4567 aligns use of pytest.seed in numeric_test.py

### LINKED ISSUE #4567: Fix missing and non-standard uses of seed in numeric_test.py
Several tests in numeric_test.py use pytest.seed.  Of these, a few need attention.

test_cast and test_histogram set seed based on pytest.seed, but then don't use it (in both cases, the seed is passed into np.random.seed, but then only ak.randint is used to generate random numbers).  These should be corrected to use the seed with ak.randint, so that tests are reproducible.

test_histogram_multidim uses a seed hardcoded as 1.  test_precision uses seeds hardcoded as 1 and 2.  These should be corrected to pytest.seed and pytest.seed + 1, as we do elsewhere.

drculhane:
Closes #4567 

There were inconsistencies in the use of pytest.seed in tests/numpy/numeric_test.py.  This resolves the ones in test_cast, test_histogram, test_histogram_multidim, and test_precision.  That should be all.

Note: the np.random.seed(seed) call was removed from test_cast and test_histogram, because only ak.randint was being used for random number generation.  But it's been retained in test_histogram_multidim because that test uses both ak.randint and np.random.

### DIFF
--- a/tests/numpy/numeric_test.py
+++ b/tests/numpy/numeric_test.py
@@ -230,13 +230,12 @@ class TestNumeric:
     @pytest.mark.parametrize("prob_size", pytest.prob_size)
     def test_cast(self, prob_size, cast_to):
         seed = pytest.seed if pytest.seed is not None else 8675309
-        np.random.seed(seed)
         arrays = {
-            ak.int64: ak.randint(-(2**48), 2**48, prob_size),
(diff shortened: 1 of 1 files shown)


ajpotts (approved):
Looks good.  Thanks for fixing this!

ajpotts (approved):

PR #4591, opened 2025-06-11
Closes 4581 -- adds benchmark for where

### LINKED ISSUE #4581: Create a benchmark for the arkouda where function, to determine if a change is an improvement
I have a distributed version of the ak.where function, but we will need a benchmark for where before we can determine if it's performant.

drculhane:
Closes #4581 

We'll need a benchmark for where to see if the proposed change to it is an improvement.

### DIFF
--- a/benchmark_v2/where_benchmark.py
+++ b/benchmark_v2/where_benchmark.py
@@ -26,15 +26,8 @@ def bench_where(benchmark, dtype, v_or_s):
         N = pytest.prob_size * cfg["numLocales"]
         seed = pytest.seed
 
-        if dtype == "int64":
-            a = ak.randint(0, 2**32, N, seed=seed, dtype=ak.int64)
-            b = ak.randint(0, 2**32, N, seed=seed, dtype=ak.int64)
-        elif dtype == "uint64":
-            a = ak.randint(0, 2**32, N, seed=seed, dtype=ak.uint64)
-            b = ak.randint(0, 2**32, N, seed=seed, dtype=ak.uint64)
-        else:
-            a = ak.randint(0, 2**32, N, seed=seed, dtype=ak.float64)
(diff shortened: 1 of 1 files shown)


benchmark_v2/where_benchmark.py
```diff
@@ -0,0 +1,57 @@
...
+def bench_where(benchmark, dtype, v_or_s):
+    """
+    Measures the performance of where
+    """
+    if dtype in pytest.dtype:
+        cfg = ak.get_config()
+        N = pytest.prob_size * cfg["numLocales"]
+        seed = pytest.seed
+
+        if dtype == "int64":
```
ajpotts, line 29: Why have three separate cases instead of just using `dtype=dtype`?

benchmark.ini
```diff
@@ -25,6 +25,7 @@ testpaths =
     benchmark_v2/no_op_benchmark.py
     benchmark_v2/io_benchmark.py
     benchmark_v2/sort_cases_benchmark.py
+    benchmark_v2/where_benchmark.py
 python_functions = bench_* check_correctness*
```
ajpotts, line 29: Just checking, you're adding this to the benchmark.ini because you think it should be run every night?  If so, you will want to update `benchmark_v2/reformat_benchmark_results.py` so the results will display.

ajpotts (approved):
Looks good!

benchmark_v2/where_benchmark.py
drculhane:
```
            elif dtype == "float64":
                a = ak.randint(0, 1, N, dtype=ak.float64, seed=pytest.seed)
                b = ak.randint(0, 1, N, dtype=ak.float64, seed=pytest.seed)
```
I kinda copy/pasted from stream_benchmark.  But you're right, this can just be dtype = etc.

PR #4604, opened 2025-06-13
Fixes `pydoclint` error when using user conda environment

alvaradoo:
Moves `pydoclint` to be a `pip` dependence instead of a `conda` one to fix the `PackageNotFoundError` when creating a `conda` environment from `arkouda-env.yml`.


ajpotts (approved):
Thanks for fixing that!

1RyanK (approved):
Looks good!

1RyanK (approved):

PR #4613, opened 2025-06-13
Closes #3570:  Update groupby_benchmark

### LINKED ISSUE #3570: Update `groupby_benchmark`

ajpotts:
## PR: Refactor `groupby_benchmark.py` for Consistency and Correctness-Only Support

This PR refactors the `groupby_benchmark.py` benchmark to align with Arkouda’s standardized benchmarking style and enable correctness-only testing.

### ✅ Key Updates

- **Correctness-Only Mode Support (`pytest.correctness_only=True`)**  
  Automatically reduces problem size to `10**4` for fast validation and CI efficiency.

- **Accurate Byte Tracking for Throughput Reporting**  
  Benchmarked function now returns the number of bytes processed, enabling consistent and accurate `GiB/sec` transfer rate metrics.

- **Helper Function Enhanced**  
  `generate_arrays()` now accepts a target problem size and handles all dtype/structure combinations with accurate byte counting.

- **Consistent Benchmark Wrapping**  
  GroupBy creation logic is now wrapped in a function (`create_groupby`) that returns transfer volume, mirroring the approach used in `aggregate_benchmark.py`.

### 🔧 Structural Improvements

- Cleaner logic for mixed-type input
- Explicit `N` management using `cfg["numLocales"]`
- Unified naming (`numBytes`, `totalbytes`, `N`, etc.)

This update ensures that `ak.GroupBy` creation benchmarks are reliable, reproducible, and compatible with future correctness/NumPy testing workflows.

Closes #3570:  Update groupby_benchmark

### DIFF
--- a/benchmark_v2/groupby_benchmark.py
+++ b/benchmark_v2/groupby_benchmark.py
@@ -2,15 +2,11 @@ import pytest
 
 import arkouda as ak
 
-# mixed used for groupby of str and int64
 TYPES = ["int64", "bigint", "str", "mixed"]
 NUM_ARR = [1, 2, 8, 16]
 
 
-def generate_arrays(dtype, numArrays):
-    cfg = ak.get_config()
-    N = pytest.prob_size * cfg["numLocales"]
-
+def generate_arrays(dtype, numArrays, N):
     totalbytes = 0
     arrays = []
     for i in range(numArrays):
@@ -19,11 +15,7 @@ def generate_arrays(dtype, numArrays):
             b = ak.randint(0, 2**32, N // numArrays, dtype=ak.uint64, seed=pytest.seed)
             ba = ak.bigint_from_uint_arrays([a, b], max_bits=pytest.max_bits)
             arrays.append(ba)
-            # bytes per bigint array (N * 16) since it's made of 2 uint64 arrays
-            # if max_bits in [0, 64] then they're essentially 1 uint64 array
-            totalbytes += (
-                a.size * 8 if pytest.max_bits != -1 and pytest.max_bits <= 64 else a.size * 8 * 2
-            )
+            totalbytes += a.size * 8 if 0 < pytest.max_bits <= 64 else a.size * 16
         elif dtype == "int64" or (i % 2 == 0 and dtype == "mixed"):
             a = ak.randint(0, 2**32, N // numArrays, seed=pytest.seed)
             arrays.append(a)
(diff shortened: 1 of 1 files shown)


benchmark_v2/groupby_benchmark.py
```diff
@@ -19,11 +15,7 @@ def generate_arrays(dtype, numArrays):
             b = ak.randint(0, 2**32, N // numArrays, dtype=ak.uint64, seed=pytest.seed)
             ba = ak.bigint_from_uint_arrays([a, b], max_bits=pytest.max_bits)
             arrays.append(ba)
-            # bytes per bigint array (N * 16) since it's made of 2 uint64 arrays
-            # if max_bits in [0, 64] then they're essentially 1 uint64 array
-            totalbytes += (
-                a.size * 8 if pytest.max_bits != -1 and pytest.max_bits <= 64 else a.size * 8 * 2
-            )
+            totalbytes += a.size * 8 if pytest.max_bits != -1 and pytest.max_bits <= 64 else a.size * 16
```
vasslitvinov, line 18: How about checking `max_bits > 0` instead of `max_bits != -1` ?

Also, should this computation `totalbytes` also account for bigint metadata?

benchmark_v2/groupby_benchmark.py
```diff
@@ -45,11 +37,19 @@ def generate_arrays(dtype, numArrays):
 @pytest.mark.parametrize("dtype", TYPES)
 def bench_groupby(benchmark, numArrays, dtype):
     if dtype in pytest.dtype:
-        arrays, totalbytes = generate_arrays(dtype, numArrays)
-        benchmark.pedantic(ak.GroupBy, args=[arrays], rounds=pytest.trials)
+        N = 10**4 if pytest.correctness_only else pytest.prob_size * ak.get_config()["numLocales"]
```
vasslitvinov, line 40: Makes me wonder whether we should query `ak.get_config()` at server startup and stash it in `pytest.ak_config` or some such. (And clear it up when the server is shutdown.)

This is independent of this PR of course.

vasslitvinov (approved):
How about adding `skipif_if_numpy` and `    benchmark.extra_info["backend"] = "Arkouda"` ?

benchmark_v2/groupby_benchmark.py
ajpotts: I changed to `max_bits > 0`.

For the metadata what were you thinking?

vasslitvinov: > For the metadata what were you thinking?

My understanding is that Chapel's `bigint` and/or GMP's representation (used by `bigint`) uses memory in addition to the "limbs" that contain the bits for the number being represented. The latter is accounted for in `a.size * 8` (or 16). Does the former also need to be counted as memory being operated on?

jaketrookman (approved):
Looks good

PR #4617, opened 2025-06-13
Closes #4499: Create a repartitionByHashString function

### LINKED ISSUE #4499: Create a `repartitionByHashString` function
**Feature Description**
This should have each locale hash the data it is responsible for and use that to map it to a destination locale, likely using the `repartitionByLocaleString` function.

**Example Behavior**
Again this can probably be ripped straight from `concatenate_uniquely`.

1RyanK:
I moved up a coforall into a loop over coforalls, thereby saving a coforall but adding one in the Repartition function. Repartition function is rather simple this time.

Closes #4499: Create a repartitionByHashString function

### DIFF
--- a/src/ConcatenateMsg.chpl
+++ b/src/ConcatenateMsg.chpl
@@ -488,8 +488,12 @@ module ConcatenateMsg
             localeSets[here.id] = new set(string);
         }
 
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good

PR #4622, opened 2025-06-16
Closes #4621: Create a max_bits_list function

### LINKED ISSUE #4621: Create a `max_bits_list` function
**Feature Description**
There are a few functions (particularly in `pdarraymanipulation.py`) that take in a list of arrays and calculate the minimum `max_bits` of them, and whether or not one of them is a `bigint` array. It may be helpful to move this to a separate function.

1RyanK:
Decided to create a helper private function to do this. I don't think this really needs to be exposed to users, so I put an underscore in front of the name.

Closes #4621: Create a max_bits_list function


drculhane (approved):
Following our discussion, I understand the intent, and definitely approve.  Always better to have the code in one place than repeated multiple times.

PR #4637, opened 2025-06-17
Closes 2999, adds ak.logspace function

### LINKED ISSUE #2999: logspace function
Replicate the functionality of numpy.logspace:
https://numpy.org/doc/stable/reference/generated/numpy.logspace.html

drculhane:
Closes #2999

This adds a logspace function, similar to np.logspace.  Just like np.logspace, it uses the linspace function to generate a sequence of numbers, then raises it to the given power.

There is also a unit test for logspace.

Note that we really need to do some additional work.  My preference is to push this function for now, and then create a new issue to improve the alignment to numpy as described below (but I'm open to discussion).

- Neither ak.logspace nor ak.linspace allow non-scalar start, stop and base.  The numpy versions do.
- ak.logspace and ak.linspace use "length" to refer to the argument that numpy calls "num".
- Neither ak.logspace nor ak.linspace allow a dtype or axis argument.  Numpy does.

### DIFF
--- a/arkouda/__init__.py
+++ b/arkouda/__init__.py
@@ -223,6 +223,7 @@ from arkouda.numpy import (
     log10,
     log1p,
     log2,
+    logspace,
     longdouble,
     longlong,
     ma,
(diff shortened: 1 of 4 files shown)


ajpotts (approved):

ajpotts:
Thanks for putting this up!  I would recommend creating tickets for the improvements you suggested in the description.

jaketrookman (approved):
Looks good

PR #4644, opened 2025-06-18
Part of #2384: Checkpoint numeric arrays; improved framework

vasslitvinov:
Add the ability for checkpointing to save and load 1-d arrays of any integer or real or bool types specified in `registration-config.json`.

Implementation-wise, set up a framework for checkpointing of any subclasses of AbstractSymEntry, by using dynamic dispatch and starting the entry metadata file with a line that is consistent across all entry kinds.

Furthermore, set up a framework in the server code that allows `param` for-loops over all `param` dimensions and all element types specified in `"parameter_classes"["array"]` in `registration-config.json`.

In more implementation detail, the list of `"parameter_classes"["array"]["nd"]` is now available through `arrayDimensionsTy`; `"parameter_classes"["array"]["dtype"]` through `arrayElementsTy`. These are generated into `src/registry/Commands.chpl` by `src/registry/register_commands.py`, which handles `@arkouda.instantiateAndRegister` and similar attributes. Their use is exemplified by the implementations of `arrayDimIsSupported()`, `arrayElmTypeIsSupported()` or `loadSymEntry()`.

This framework will be leveraged to implement checkpointing of the remaining types of SymTab entries. The type-tuple-based approach is a workaround for Chapel's lack of iterators that yield `param` or `type`.

Other implementation changes:

Restricted `addEntry()` to handle only those types that are requested in `registration-config.json`. This reduces the amount of generated code because the compiler no longer needs to instantiate `SymEntry` for some of the unintended types, such as `[u]int`16 and 32.

Extended `SymTab.insert()` to set the new entry's name, which would otherwise be left empty.

Replaced uses of `MaxArrayDims` with `arrayDimIsSupported()`, except for one in getConfig().

Moved `getRegistrationConfig()` from `register_commands.py` to `MsgProcessing.chpl` for easier maintenance, and simplified it.

### DIFF
--- a/src/MultiTypeSymbolTable.chpl
+++ b/src/MultiTypeSymbolTable.chpl
@@ -55,6 +55,7 @@ module MultiTypeSymbolTable
             const name = nextName(),
                   response = MsgTuple.newSymbol(name, symbol.borrow());
             tab.addOrReplace(name, symbol);
+            symbol.setName(name);
             mtLogger.info(getModuleName(),getRoutineName(),getLineNumber(),response.msg);
             return response;
         }
@@ -73,6 +74,9 @@ module MultiTypeSymbolTable
             :returns: borrow of newly created `SymEntry(t)`
         */
         proc addEntry(name: string, shape: int ...?N, type t): borrowed SymEntry(t, N) throws {
+            if ! arrayDimIsSupported(N) then compilerWarning("arrays with rank ", N:string,
+              " are not included in the server's configured ranks: ", arrayDimensionsStr);
+
             // check and throw if memory limit would be exceeded
             // TODO figure out a way to do memory checking for bigint
             if t != bigint {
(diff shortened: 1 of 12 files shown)


ajpotts (approved):
Looks good.

src/MultiTypeSymbolTable.chpl
```diff
@@ -90,9 +94,8 @@ module MultiTypeSymbolTable
 
             tab.addOrReplace(name, entry);
             entry.setName(name);
-            // When we retrieve from table, it comes back as AbstractSymEntry so we need to cast it
-            // back to the original type. Since we know it already we can skip isAssignableTo check
-            return (tab[name]:borrowed GenSymEntry).toSymEntry(t, N);
+
+            return entry :borrowed :unmanaged :borrowed; // suppress lifetime checking
```
ShreyasKhandekar, line 98: Why is this needed?

ShreyasKhandekar (approved):

src/MultiTypeSymbolTable.chpl
vasslitvinov: Good question. First, this avoids a lookup in the symtable for the thing we just put there. Second, it works around the lifetime checker that would flag it an error if we just returned `entry :borrowed`.

ShreyasKhandekar: Thanks for explaining that. Resolved.

PR #4645, opened 2025-06-18
Closes #4489:  trapz instability

### LINKED ISSUE #4489: trapz instability
The example in the `trapz` function is gives unstable results:
```python
    >>> theta = xp.linspace(0, 2 * xp.pi, num=1000, endpoint=True)
    >>> xp.trapz(xp.cos(theta), x=xp.sin(theta))
    Arkouda Array ((), float64)3.1415719413758416

```
Sometimes instead gives:
```python
Arkouda Array ((), float64)3.141571941375841

```
instead.  It should give the same result every time.

ajpotts:
My best guess is the minor differences in the `trapz` return values are precision errors on the server.  I added ellipses to the docstring and additional `np.allclose` calls to the unit test to hopefully prevent CI errors from this.

Closes #4489:  trapz instability


1RyanK (approved):
Looks good!

PR #4652, opened 2025-06-20
Closes #3572:  Update io_benchmark

### LINKED ISSUE #3572: Update `io_benchmark`

ajpotts:
Closes #3572:  Update io_benchmark

### DIFF
--- a/benchmark_v2/io_benchmark.py
+++ b/benchmark_v2/io_benchmark.py
@@ -71,6 +71,7 @@ def _generate_df(N, dtype, returnDict=False):
 
 
 @pytest.mark.skip_correctness_only(True)
+@pytest.mark.skip_numpy(True)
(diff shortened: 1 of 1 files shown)


vasslitvinov:
@ajpotts any comments on why these tests skip correctness / what it would take to enable correctness checking if appropriate?

ajpotts:
The correctness testing for all functions in `arkouda` should be in the unit tests.  The only reason to put correctness testing in the benchmarks is to test the logic of the benchmark itself.  In this case, it didn't seem worth it.  For example, to test a write you would then need to read in the file and do a comparison to the original.  As `arkouda`'s output format is slightly different than `pandas` it would not be possible to directly compare the written files to `pandas`.  If you still think this is worth exploring I would recommend putting it in a separate ticket.

1RyanK (approved):
Looks good!

jaketrookman (approved):
Looks good

PR #4666, opened 2025-06-25
Increase unit test timeout to 30 minutes

ajpotts:
Increases unit test timeout to 30 minutes to avoid timeouts on HPE's Apollo.


bradcray (approved):
This seems worth the try to me.  If there's any way to inspect the last successful run to determine how much time it needs, let me know and I can poke into it or ask someone else to do so.

PR #4683, opened 2025-06-25
Closes #4682:  Refactor parquet-fixed-strings benchmark to use pytest…

### LINKED ISSUE #4682: Refactor parquet-fixed-strings benchmark to use pytest framework
Refactor the `parquet-fixed-strings` benchmark to use the pytest framework.

ajpotts:
## PR: Refactor `parquet-fixed-strings.py` for Pytest-Benchmark Framework

This PR refactors the original script-based benchmark into a parameterized `pytest-benchmark` test for standardized performance testing of fixed-length string reads from Parquet files.

### ✅ Key Improvements

- **Benchmark Group**: `ParquetFixedStrings`
- **Parameters**:
  - `nfiles`: number of files (1, 5, 10)
  - `scaling`: scale down file size per shard (True/False)
  - `fixed`: use `fixed_len` when reading Parquet files (True/False)

- **Benchmark Timing**:
  - Uses `benchmark.pedantic(...)` for consistent trial-based timing
  - Skips during `--numpy` and `--correctness_only` runs

- **Temporary Directory Isolation**:
  - Uses `tmp_path` fixture to write files for isolated test execution

- **Metadata**:
  - Problem size, number of files, fixed flag, scaling, and transfer rates are recorded

- **IO Benchmark File Cleanup**
  - `_remove_files` adds capability to remove directories as well as files.

This provides consistent, scalable performance benchmarking of Arkouda’s Parquet I/O read with fixed-length string options.

Closes #4682:  Refactor parquet-fixed-strings benchmark to use pytest framework

### DIFF
--- a/benchmark_v2/parquet-fixed-strings_benchmark.py
+++ b/benchmark_v2/parquet-fixed-strings_benchmark.py
@@ -1,0 +1,39 @@
+import pytest
+
+import arkouda as ak
+
+
+@pytest.mark.skip_numpy(True)
+@pytest.mark.skip_correctness_only(True)
+@pytest.mark.benchmark(group="Parquet_Fixed_Strings")
+@pytest.mark.parametrize("scaling", [False, True])
+@pytest.mark.parametrize("fixed_len", [1, 8])
+@pytest.mark.parametrize("nfiles", [1, 5, 10])
+def bench_parquet_fixed_strings(benchmark, tmp_path, scaling, fixed_len, nfiles):
+    cfg = ak.get_config()
+    N = pytest.prob_size * cfg["numLocales"] if scaling else pytest.prob_size
+    base_path = tmp_path / f"parq_{'scaled' if scaling else 'flat'}_{nfiles}"
+    base_path.mkdir(parents=True, exist_ok=True)
+
+    strings = ak.random_strings_uniform(fixed_len, fixed_len + 1, N, seed=pytest.seed)
+    numBytes = strings.nbytes
+
+    strings.to_parquet(base_path / "part_0.parquet")
+
+    def run():
+        ak.read_parquet(str(base_path / "part_0_LOCALE0000.parquet"), fixed_len=fixed_len)
+
+    benchmark.pedantic(run, rounds=pytest.trials)
+
+    benchmark.extra_info["description"] = (
+        f"Parquet read performance for fixed-length strings of size {fixed_len}, "
+        f"{'scaling' if scaling else 'non-scaling'} mode from {nfiles} files."
+    )
+    benchmark.extra_info["problem_size"] = N
+    benchmark.extra_info["fixed_len"] = fixed_len
+    benchmark.extra_info["scaling"] = scaling
+    benchmark.extra_info["nfiles"] = nfiles
+    benchmark.extra_info["backend"] = "Arkouda"
+    benchmark.extra_info["transfer_rate"] = "{:.4f} GiB/sec".format(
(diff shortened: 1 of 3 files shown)


benchmark_v2/parquet-fixed-strings_benchmark.py
```diff
@@ -0,0 +1,36 @@
...
+
+import arkouda as ak
+
+
+@pytest.mark.skip_numpy(True)
+@pytest.mark.skip_correctness_only(True)
+@pytest.mark.benchmark(group="Parquet_Fixed_Strings")
+@pytest.mark.parametrize("scaling", [False, True])
+@pytest.mark.parametrize("fixed_size", [1, 8])
+@pytest.mark.parametrize("nfiles", [1])
```
vasslitvinov, line 13: Please reconcile with the PR message that claims:
*  nfiles: number of files (1, 5, 10) [I see only `1`; if the other options are added, this should probably be incorporated into `extra_info["description"]`]
*  scaling: scale down file size per shard (True/False) [this matches]
*  fixed: use fixed_len when reading Parquet files (True/False) [instead I see `fixed_size`]

benchmark_v2/parquet-fixed-strings_benchmark.py
```diff
@@ -0,0 +1,36 @@
...
+    benchmark.pedantic(run, rounds=pytest.trials)
+
+    benchmark.extra_info["description"] = (
+        f"Parquet read performance for fixed-length strings of size {fixed_size}, "
+        f"{'scaling' if scaling else 'non-scaling'} mode."
+    )
+    benchmark.extra_info["problem_size"] = N
+    benchmark.extra_info["fixed_size"] = fixed_size
+    benchmark.extra_info["scaling"] = scaling
+    benchmark.extra_info["nfiles"] = nfiles
```
vasslitvinov, line 36: Transfer rate should also be recorded in `extra_info`, as well as `backend`.

Re `backend` -- depending on how it is used, we may indeed want to record it only in those cases where there are two backends, leaving it undefined when there is only the Arkouda backend. [In this case, my review feedback to the contrary on this and earlier PRs would be misguided.]

benchmark_v2/parquet-fixed-strings_benchmark.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.mark.skip_correctness_only(True)
+@pytest.mark.benchmark(group="Parquet_Fixed_Strings")
+@pytest.mark.parametrize("scaling", [False, True])
+@pytest.mark.parametrize("fixed_size", [1, 8])
+@pytest.mark.parametrize("nfiles", [1])
+def bench_parquet_fixed_strings(benchmark, tmp_path, scaling, fixed_size, nfiles):
+    cfg = ak.get_config()
+    N = pytest.prob_size * cfg["numLocales"] if scaling else pytest.prob_size
+    base_path = Path(pytest.io_path) / f"parq_{'scaled' if scaling else 'flat'}_{nfiles}"
+    base_path.mkdir(parents=True, exist_ok=True)
```
vasslitvinov, line 18: Should probably use the `tmp_path` fixture, according to the PR message?  Or adjust the PR message.

benchmark_v2/parquet-fixed-strings_benchmark.py
ajpotts: Fixed.  I left `backend` for now.

drculhane (approved):
This looks solid.

PR #4686, opened 2025-06-26
Temporarily don't timeout pytest for debugging

ajpotts:
Temporarily remove pytest timeout to help debug hanging tests on the HPE systems.


bradcray (approved):
This looks good to me, assuming that the lack of setting these values is sufficient to turn off the timeout.  Checking #4654 that introduced the timeout, this seems to be the case.

PR #4687, opened 2025-06-26
Closes 4657 aligns linspace and logspace to numpy

### LINKED ISSUE #4657: Align linspace and logspace to numpy
Arkouda's implementations of linspace and logspace differ significantly from numpy's.

- Neither ak.logspace nor ak.linspace allow non-scalar start, stop and base. The numpy versions do.
- ak.logspace and ak.linspace use "length" to refer to the argument that numpy calls "num".
- Neither ak.logspace nor ak.linspace allow a dtype or axis argument. Numpy does.

These should be addressed.

drculhane:
Closing because of the memory issue, and because I've posted an alternate python-only solution.

----------

Closes #4657 and #2999

----------

Still having issues with the memory usage.  In the meantime, I've done the simplifications I talked about (reducing the cases to scalar-scalar and vector-vector), and also simplified the calculation of the result shape (it no longer requires a separate proc).

----------

Update : reworked the for loops chapel-side into foralls, as Ryan suggested.

----------

Ready for review.  This supercedes PR #4637, which I've closed.  That PR added a logspace function, but it didn't include all the parameters of numpy logspace.  This one does.

Quick summary:

There's a lot of code here, but half of it (maybe more) is docstrings and tests.

adds logspace to the __ init __.py files, and to the __ all __ of pdarraycreation.py.
rewrites linspace python-side to use the same parameters and parameter names as numpy
adds logspace, which (just like numpy) computes base ** (results from linspace)
for both linspace and logspace, I've include a dtype parameter for compatibility with numpy, but I ignore it, because these are fundamentally float functions.  I'm open to discussion about that.
eliminates linspaceMsg from SequenceMsg.chpl, and instead has linspace_ss, linspace_sv, linspace_vs, linspace_vv, and a helper function linspace_shape.
includes unit tests for 1D, 2D and 3D versions.  The previous unit tests have been removed.

There is no logspace function chapel-side since it's not needed.

One topic for discussion: I didn't include multi-dim examples in the docstrings, because that made the upstream tests fail.  I could include them, and have numpy/pdarray_creation_test.py skip the docstring tests if the higher ranks aren't compiled (this is how numpy/numeric_test.py handles it).  I'm open to suggestions.

### DIFF
--- a/arkouda/__init__.py
+++ b/arkouda/__init__.py
@@ -222,6 +222,7 @@ from arkouda.numpy import (
     issubdtype,
     lib,
     linspace,
+    logspace,
     log,
     log10,
     log1p,
--- a/arkouda/numpy/__init__.py
+++ b/arkouda/numpy/__init__.py
@@ -259,6 +259,7 @@ from arkouda.numpy.pdarraycreation import (
     full,
     full_like,
     linspace,
+    logspace,
     ones,
     ones_like,
     promote_to_common_dtype,
--- a/arkouda/numpy/pdarrayclass.py
+++ b/arkouda/numpy/pdarrayclass.py
@@ -558,9 +558,9 @@ class pdarray:
 
     def __del__(self):
         try:
+            logger.debug(f"deleting pdarray with name {self.name}")
             from arkouda.client import generic_msg
 
-            logger.debug(f"deleting pdarray with name {self.name}")
             generic_msg(cmd="delete", args={"name": self.name})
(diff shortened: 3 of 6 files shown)


1RyanK (approved):
I think you should check some of the `for`s and see which of them can be `forall`s instead. You said something about docstring tests failing - is that because of precision issues or something? I think Amanda was doing some ... thing to get around those.

PR #4690, opened 2025-06-30
Skipif test_checkpoint()

vasslitvinov:
One of the testing machines does not support file system accesses across the login node (running the pytest client) and the compute nodes (running the server). Skip `test_checkpoint()` in that case to avoid failures.

While there, refactor slightly `test_checkpoint()` implementation.

P.S. We may need to skipif `auto_checkpoints.py` as well in this case.

### DIFF
--- a/tests/checkpoint_test.py
+++ b/tests/checkpoint_test.py
@@ -33,11 +33,19 @@ def cp_test_base_tmp(request):
 
 
 class TestCheckpoint:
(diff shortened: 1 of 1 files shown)


e-kayrakli (approved):

ajpotts (approved):
Thanks!

ajpotts (approved):

PR #4694, opened 2025-07-02
Closes-2915, adds isnumeric function

### LINKED ISSUE #2915: add isnumeric function for pdarrays
Add isnumeric function to pdarrays to match numpy function for ndarrays:
https://numpy.org/doc/stable/reference/generated/numpy.char.isnumeric.html

For each element, return True if there are only numeric characters in the element.

isnumeric may not be implemented yet for strings in chapel.

drculhane:
Update

--

The current revision (as of 7/10/25) moves isnumeric into the char folder.  It can now be addressed in three ways:

pda.isnumeric(), that is, as a method of the pdarray class
ak.char.isnumeric(pda), similar to np.char.isnumeric, and
ak.isnumeric(pda), and I'm not sure why that works.

arkouda/numpy/char.py has only the isnumeric function.  Its input is String, but when I import strings at the top of the file in order to use typechecking, I get a circular import.  The only way I found to avoid that was not to use @typecheck, and instead to import strings inside the function, and manually do the type check there.

--

I've implemented Ryan's suggestion to use a set instead of an array to store the known numeric unicodes.  I tried to search through all the letters in a string with a forall, but I kept getting this error message:

```
/home/drandy/projects/arkouda/src/SegmentedString.chpl:1776: error: A standalone or leader iterator is not found for the iterable expression in this forall loop
make: *** [Makefile:569: arkouda_server] Error 1

```
I understand that message, but it doesn't seem applicable, since it refers to zipped iterations, which I don't think this is.

--

I've done a major rewrite of this, such that it now covers all the cases that np.char.isnumeric() considers numeric (a total of 1922 characters, including decimal digits, numerical superscripts and subscripts, special characters with digits encircled or enclosed in parens, "vulgar fractions," non-Arabic numerals, etc).

The function is now a string method (in strings.py, not numeric.py), so it's addressed the same way the other string methods are, e.g.

stringName.isdigit()
stringName.isdecimal()
stringName.isnumeric()

It's also accessible as a standalone function, ak.isnumeric().

--

Closes #2915, adding an isnumeric function to match np.char.isnumeric.

Note:

This is presently in arkouda/numpy/numeric.py, but maybe it belongs in strings.py? [_That's where it is now._]

There is no multi-dim test, because arkouda, AFAIK, doesn't support multi-dim string pdarrays.  String pdarrays are actually segmented strings.

The SegString class already had a method isdigits, which I was able to use to implement isnumeric. [_outdated.  The class now has an isnumeric method._]


drculhane:
> I think this will work, but would it be too much to change the array of hex things into a set? Then instead of doing a forall comparing the things in the set to the current character, you can just see if the set contains the character, and do a forall over the strings or characters (or both? I think if you do both, then it works whether you have few strings with many characters each, or many strings with few characters each).

I really like this suggestion.  I'll push a revision later this morning.

1RyanK (commented):
Okay, so after looking into it, parallelizing over a single string is not really something the compiler likes to do, thus the compiler error.
However, one other thing - I'm concerned that every time the function is called, it creates that big var set and checks to see if the character is in it. Instead, could you try declaring it outside the function, and declaring as const? It doesn't change so that may help performance. Other than that, I don't think there's anything practical that can be done.

drculhane:
> Okay, so after looking into it, parallelizing over a single string is not really something the compiler likes to do, thus the compiler error. However, one other thing - I'm concerned that every time the function is called, it creates that big var set and checks to see if the character is in it. Instead, could you try declaring it outside the function, and declaring as const? It doesn't change so that may help performance. Other than that, I don't think there's anything practical that can be done.

Also an excellent suggestion.  Done.

arkouda/numpy/strings.py
```diff
@@ -757,6 +757,57 @@ def isdecimal(self) -> pdarray:
...
         )
 
+    @typechecked
+    def isnumeric(self) -> pdarray:
+        """
+        Return a boolean pdarray where index i indicates whether string i of the
+        Strings has all numeric characters. There are 1922 unicode characters that
+        qualify as numeric, including the digits 0 through 9, superscripts and
+        subscripted digits, special characters with the digits encircled or
+        enclosed in parens, "vulgar functions," and more.
```
ajpotts, line 767: I think you mean "vulgar fractions", not "vulgar functions".

tests/numpy/string_test.py
```diff
@@ -973,3 +974,28 @@ def test_concatenate_uniquely(self, size):
...
+        for i in range(prob_size):
+            heads_or_tails = random.choice([True, False])
+            if heads_or_tails:
+                prob_set.append("".join(random.choice(numerics) for j in range(4)))
+            else:
+                prob_set.append("".join(random.choice(alphanumerics) for j in range(4)))
+        nda = np.array(prob_set)
+        pda = ak.array(nda)
+        assert_arkouda_array_equivalent(np.char.isnumeric(nda), pda.isnumeric())
+        assert_arkouda_array_equivalent(np.char.isnumeric(nda), ak.isnumeric(pda))
```
ajpotts, line 1001: Maybe just compute `np.char.isnumeric(nda)` once, in case it's slow for large sizes?

ajpotts (approved):
Great work!

arkouda/numpy/strings.py
drculhane: Oops :)

ajpotts (approved):

PR #4697, opened 2025-07-02
Closes #4696:  set max-parallel for multi-dim tests in CI

### LINKED ISSUE #4696: set max-parallel for multi-dim tests in CI
Set the max-parallel for multi-dim tests in CI.  This will limit the number of runners in that group that can kick off at the same time and may reduce the chance of out-of-memory errors.

ajpotts:
Sets the `max-parallel` constraint for multi-dim tests in CI. This will limit the number of runners in that group that can kick off at the same time and may reduce the chance of out-of-memory errors.

Closes #4696:  set max-parallel for multi-dim tests in CI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -485,6 +485,7 @@ jobs:
     strategy:
       matrix:
         chpl-version: ['2.0.0','2.1.0','2.2.0','2.3.0','2.4.0']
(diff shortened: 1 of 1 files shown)


1RyanK (approved):
I'm happy to give this a shot, but I thought the amount of memory was per runner? Rather than some pile of memory for all runners?

Issue #4709, opened 2025-07-11
How to checkpoint ak.numpy.random.Generator ?


vasslitvinov:
This issue seeks a solution for the currently-incomplete checkpointing of `ak.numpy.random.Generator` objects.

#### Details

The [current checkpointing effort](https://github.com/Bears-R-Us/arkouda/milestone/18) is concerned with saving and restoring the SymTab entires _ON THE SERVER_. This is not sufficient / correct for preserving the complete state of an `ak.numpy.random.Generator` instance.  That is, in the following scenario:

```python
g = default_rng(...)
... use `g` ...
ak.save_checkpoint(someName)
... use `g` some more #1 ...
ak.load_checkpoint(someName)
... use `g` some more #2 ...
```

The user expects that the use of `g` in `#2` will generate the same sequence of numbers as its use in `#1`.

Currently this will not be the case, here is why. While the server's state related to `g` will be restored properly upon `ak.load_checkpoint(someName)`, `g._state` will not be restored while affecting the numbers generated by `g`.

In other words, these Generators rely on both server-side and client-side state, while the current checkpointing approach preserves only server-side state.

My initial recommendation is to eliminate `g._state`. It is used to fast-forward Chapel's RNGs, however it may not be necessary.

BTW if we keep `g._state`, we need to add its increment by 1 in cases like this:
```python
if size is None:
    # delegate to numpy when return size is 1
    return self._np_generator.uniform(low=low, high=high)
```

vasslitvinov:
Here is another proposal that will result in simpler implementation, thanks to @mppf:

* Store (seed, state) on the client side. Notes: "State", which is the current position in the random number sequence for the given seed, is already stored as Generator._state. "Seed" is also stored, we just need to "instantiate" it to an integer in the case that the generator is created with `seed=None`.

* Whenever a Generator needs to participate in a server-side operation, send its seed and state over to the server. The server will create a `randomStream` on the fly, with the desired element type.

This way we can remove `Generator._name_dict` in the client and `GeneratorSymEntry` on the server, as well as back out the additions in #4710 that are related to checkpointing of GeneratorSymEntrys.

[...]

PR #4710, opened 2025-07-11
Minor code improvements

vasslitvinov:
This PR initially added checkpointing of `GeneratorSymEntry`. However during testing I realized that, while checkpointing per se works, restoring from a checkpoint does not "roll back" Arkouda properly. This is because part of the Generator state, specifically its `state` field, is stored in the client and not on the server. Therefore I am backing out this checkpointing addition.

See #4709 for discussion on this.

This PR now contains minor code cleanups that I did while working on GeneratorSymEntry checkpointing.

User-facing: adds `ak.get_registration_config()`; populates a Generator in the client with GeneratorSymEntry instances on the server only for those types that are requested in `registration-config.json`.


vasslitvinov:
Moved to #4752 due to 'master' -> 'main' rename that invalidates this PR.

PR #4720, opened 2025-07-17
Prevent over instantiation of `opeqvs`

jabraham17:
Since pdarray.opeq enforces that the dtype of a scalar must match the dtype of the vector, this PR makes similar changes to the server to enforce that. This eliminates ~20 instantiations of `opeqvs`

Future work: given that `pdarray.opeq`/`opeqvs` outlaws `vector op= scalar`, can `pdarray._binop`/`pdarray._r_binop`/`binopsv`/`binopvs` do the same?


ajpotts (approved):
Looks good.

jaketrookman (approved):
Looks good

e-kayrakli (approved):

PR #4728, opened 2025-07-21
Refactor bitwise ops on booleans and remove duplicated `Set` from BinOp

jabraham17:
This PR makes a few changes to how bitwise and boolean ops are computed in the arkouda server. 

- Since the order of operands to bitwise ops does not matter, by using a helper function `doBoolBoolBitOp` and only calling it with operands in the same order (always `vector, scalar` or `vector, vector`) we can save some compilation time
- The same `Set.set` was being recreated all over the place to determine if an op was an operator. This PR refactors that set into a common place so it does not get instantiated for all the generic combinations.
- Makes a similar transformation with `Set.set` for `realOps`

I consider this a general cleanup and consolidation of logic in BinOp/OperatorMsg, it also saves a little bit of time while compiling

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -9,6 +9,7 @@ module BinOp
   use Message;
   use BitOps;
   use BigInteger;
+  use Set;
 
 
   private config const logLevel = ServerConfig.logLevel;
@@ -128,18 +129,11 @@ module BinOp
     return res;
   }
 
-  proc isRealOp(op: string): bool {
-    return op == "+" || op == "-" || op == "*" ||
-           op == "//" || op == "%" || op == "**";
-  }
+  const realOps = new set(string, ["+", "-", "*", "//", "%", "**"]);
 
   // All operations that involve one of these operations result in a `bool`
   // symbol table entry.
-  proc isBoolOp(op: string): bool {
-    return op == "<"  || op == "<=" ||
-           op == ">"  || op == ">=" ||
-           op == "==" || op == "!=";
-  }
+  const boolOps = new set(string, ["<", "<=", ">", ">=", "==", "!="]);
 
   proc doBoolBoolBitOp(
     op: string, ref e: [] bool, l: [] bool, r /*: [] bool OR bool*/
@@ -195,7 +189,7 @@ module BinOp
 
     if etype == bool {
 
-      if isBoolOp(op) {
+      if boolOps.contains(op) {
 
         select op {
 
@@ -328,7 +322,7 @@ module BinOp
 
     if etype == bool {
 
-      if isBoolOp(op) {
+      if boolOps.contains(op) {
 
         select op {
 
@@ -453,7 +447,7 @@ module BinOp
 
     if etype == bool {
 
-      if isBoolOp(op) {
+      if boolOps.contains(op) {
 
         select op {
 
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -129,6 +129,32 @@ module BinOp
     return res;
   }
 
+  //
+  // TODO: these checks sets are used only to check if a string matches
+  // i.e. realOps.contains(op) or boolOps.contains(op)
+  // however, a faster way to do this is to write a function that checks
+  // the ascii bytes
+  // for example, the boolOps can be written as:
+  //
+  // return (op.numBytes == 1 && (op.byte[0] == 60 ||
+  //                             op.byte[0] == 62)) ||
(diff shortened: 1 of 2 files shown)


src/BinOp.chpl
```diff
@@ -128,6 +128,36 @@ module BinOp
     return res;
   }
 
+  proc isRealOp(op: string): bool {
+    return op == "+" || op == "-" || op == "*" ||
+           op == "//" || op == "%" || op == "**";
+  }
```
1RyanK, line 134: I definitely agree on taking out all the instantiations of the set, but how does this compare to creating a single const set of these operators (outside of the procs) and using .contains()? Is it few enough things that comparing directly to each one beats hashing? I'm probably splitting hairs here.

1RyanK (approved):
Looks good! One small question.

src/BinOp.chpl
jabraham17: This is a good point. I was more focused on code size here than performance, but the set hashing beats the string comparison.

I used this little test program to validate this

```chapel
proc func1(op) {
  return op == "<" || op == ">" || op == "<=" ||
         op == ">=" || op == "==" || op == "!=";
}

  use Set;
  var ops = new set(string);
  ops.add("<");
  ops.add(">");
  ops.add("<=");
  ops.add(">=");
  ops.add("==");
  ops.add("!=");
proc func2(op) {
  return ops.contains(op);
}

proc main() {
  use Random;
  use Time;
  var ops: [1..100_000] string;
  var choices = ["<", ">", "<=", ">=", "==", "!=", "+", "-", "*", "/", "%", "&&", "||", "+=", "-=", "*=", "/=", "%=", "&&=", "||=", "**", "**=", "//", "//="];
  forall o in ops  with (var rs = new randomStream(int)) {
    o = choices[rs.choose(choices.domain)];
  }

  var ts = new stopwatch();
  ts.start();
  var count1 = 0;
  for o in ops {
    if func1(o) then count1 += 1;
  }
  ts.stop();
  writeln("Time taken for func1: ", ts.elapsed(), " seconds, count: ", count1);
  ts.reset();

  ts.start();
  var count2 = 0;
  for o in ops {
    if func2(o) then count2 += 1;
  }
  ts.stop();
  writeln("Time taken for func2: ", ts.elapsed(), " seconds, count: ", count2);
}
```

```
Time taken for func1: 0.003048 seconds, count: 24927
Time taken for func2: 0.001609 seconds, count: 24927
```

I think once `Set` is only being used once and not instantiated for every single type it can be still be used for the performance benefits here

jabraham17: Although I will note that an even faster version that beats `Set` is the following

```chapel
proc func3(op) {
  return (op.numBytes == 1 && (op.byte[0] == 60 ||
                              op.byte[0] == 62)) ||
         (op.numBytes == 2 && ((op.byte[0] == 60 && op.byte[1] == 61) ||
                               (op.byte[0] == 62 && op.byte[1] == 61) ||
                               (op.byte[0] == 61 && op.byte[1] == 61) ||
                               (op.byte[0] == 33 && op.byte[1] == 61)));
}
```

Although I am not sure the performance is worth the loss of code readability.

src/BinOp.chpl
```diff
@@ -128,6 +129,55 @@ module BinOp
...
+  // return (op.numBytes == 1 && (op.byte[0] == 60 ||
+  //                             op.byte[0] == 62)) ||
+  //        (op.numBytes == 2 && ((op.byte[0] == 60 && op.byte[1] == 61) ||
+  //                              (op.byte[0] == 62 && op.byte[1] == 61) ||
+  //                              (op.byte[0] == 61 && op.byte[1] == 61) ||
+  //                              (op.byte[0] == 33 && op.byte[1] == 61)));
+  //
+  // it should also be possible to use the `toByte` method at compile-time to
+  // improve the readability, although this has not been tested to see how it impacts
+  // compile-time and runtime performance:
```
e-kayrakli, line 148: If the worry about the compile time hit by this is because we will see this kind of `return`s all over the place, we can always have `param lessThan = "<".toByte();` etc. at module-scope to compile `toByte` functions only once.

In general, do you know empirically that either of these approaches makes things faster. I'd be surprised if `set.contains` is taking up too much time, even with small data sizes.

e-kayrakli (approved):

src/BinOp.chpl
e-kayrakli: Oh, I see that in the discussion above with Ryan you mention that they do indeed make the code go faster.

ajpotts (approved):

Issue #4701, opened 2025-07-07
Add pytest.ak_config to benchmark_v2


ajpotts:
Add `pytest.ak_config` to the `benchmark_v2/conftest.py` to store the config parameters and reduce some code duplication.

PR #4735, opened 2025-07-22
Closes #4704:  Create a pytest.N for benchmark_v2/conftest.py

### LINKED ISSUE #4704: Create a pytest.N for benchmark_v2/conftest.py
Create a pytest.N for benchmark_v2/conftest.py to reduce some code duplication in the benchmarks.

ajpotts:
Creates `pytest.N` for the size of the benchmark and `pytest.cfg` for the arkouda configuration map for `benchmark_v2/conftest.py`.

Closes #4704:  Create a pytest.N for benchmark_v2/conftest.py
#4701:  Add pytest.ak_config to benchmark_v2


drculhane (approved):
This looks excellent.  This being the first time I looked at this concept, I checked the other benchmark files, where the change hadn't been added.  Having done that, I think the same change should be made to these benchmarks:

array_transfer
bigint_bitwise_binops
dataframe_indexing
encoding
groupby
small_str_groupby
str_locality
substring_search

I can't tell if sort_cases needs it, and it looks like no-op and reformat don't.

This could easily be a separate issue.

drculhane (approved):
Nicely done.

PR #4748, opened 2025-07-28
Closes 4726, standardizing use of seed in unit tests

### LINKED ISSUE #4726: standardize seed for unit tests
We should use the same seed for every unit tests, and ideally it would be set in the `conftest.py`.

drculhane:
Closes #4726.

In almost all cases, the seed in the unit tests has been set to the standard pytest.seed.

There are a handful of exceptions where hardcoded seeds were retained, because the unit test specifically checks for exact known results.


tests/array_api/array_manipulation.py
```diff
@@ -4,14 +4,13 @@
 import arkouda as ak
 import arkouda.array_api as xp
 
-SEED = 12345
-s = SEED
+seed = pytest.seed if pytest.seed is not None else 8675309
```
ajpotts, line 7: Could you update conftest.py to set the default version of the seed, as discussed in the developers meeting?

tests/numpy/random_test.py
```diff
@@ -168,8 +170,8 @@ def test_choice(self):
         assert scalar in [0, 1, 2, 3, 4]
 
     def test_choice_flags(self):
-        # use numpy to randomly generate a set seed
-        seed = np.random.default_rng().choice(2**63)
+        # use numpy to randomly generate a set seed      --- maybe.  Or maybe use standard.
+        # seed = np.random.default_rng().choice(2**63)
```
ajpotts, line 174: Do we need these comments?  I find them confusing.

ajpotts (approved):
Great work!  Thanks for this!

tests/numpy/random_test.py
drculhane: Yep.  Leftover comment.  I've gotten better at deleting those, but I missed that one.

jaketrookman (approved):
Looks good

ajpotts (approved):

PR #4750, opened 2025-07-29
Add heartbeat on the server

vasslitvinov:
This heartbeat allows client to recognize that the connection to the server went away while the server was processing a message from the client. The client raises an exception in this case. The heartbeat is activated by passing a positive number to the `timeout` argument of `ak.connect()`.

While there, tidy up some code. Also, introduce `ak.client.server_sleep(seconds)` to enable testing of the heartbeat, and `ak.client.note_for_server_log(message)` to make it easier to track the tests being tested while examining `arkouda.log`.


arkouda/client.py
```diff
@@ -686,6 +697,73 @@ def disconnect(self) -> None:
...
+            self.heartbeatSocket = None
+            self.heartbeatInterval = -1
+            return
+
+        self.heartbeatInterval = timeout
+        timeout = max(int(timeout), 1)  # setsockopt() needs whole seconds
+
+        # Given the settings below, an exception will be raised within
+        # (4 * timeout) seconds. We can tune these.
+        # Note that these need to be enabled BEFORE self.socket.connect().
```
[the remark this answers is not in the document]
vasslitvinov: I moved it to the docstring.

I will let testing confirm that this function is invoked correctly.

tests/optioned-server/conftest.py
```diff
@@ -61,6 +62,16 @@ def _my_start_server(server_args, note):
     print(f"Started arkouda_server {note} -nl {pytest.nl} on {host}:{port}")
 
 
+def _my_stop_server():
+    if hasattr(pytest, "skip_stop_server_once"):
+        del pytest.skip_stop_server_once
+        # Without this, the parent fixture manage_connection() invokes
+        # ak.disconnect(), which throws a ZMQ error, which breaks testing.
+        ak.client.connected = False
```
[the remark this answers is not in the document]
vasslitvinov: Catching the error in `manage_connection` may be better indeed, however I do not see how to do so.

Instead, I renamed `skip_stop_server_once` to `server_already_stopped` to make it clear that the server is no longer running in this case, so we can't send any zmq messages to it. `server_already_stopped` is set in `heartbeat.py` after the server is forcefully killed.

tests/optioned-server/heartbeat.py
```diff
@@ -0,0 +1,36 @@
...
+        """
+        Ensure the client recognizes server disconnect reasonably soon.
+        """
+        hb_timeout = 1
+        kill_delay = 0.5
+        # reconnect with the specified timeout
+        ak.connect(server=pytest.server, port=pytest.port, timeout=hb_timeout)
+
+        def kill_server():
+            time.sleep(kill_delay)  # let the main thread
```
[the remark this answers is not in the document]
vasslitvinov: Indeed! Fixed.

arkouda/client.py
```diff
@@ -666,6 +673,7 @@ def connect(self, timeout: int = 0) -> None:
 
         context = zmq.Context()
         self.socket = context.socket(zmq.REQ)
+        self.setupHeartbeat(timeout)  # enables heartbeat if timeout > 0
```
ajpotts, line 676: If you add a socket in `connect()`, should it also be closed in `disconnect()`?

arkouda/client.py
```diff
@@ -686,6 +696,75 @@ def disconnect(self) -> None:
...
+            if self.socket.poll(self.heartbeatInterval * 1000, zmq.POLLIN):
+                return  # got a response from the server
+
+            # No server response within the timeout. So, check the monitor socket.
+            if cast(zmq.Socket, self.heartbeatSocket).poll(timeout=0):  # non-blocking
+                # We are monitoring only EVENT_CLOSED and EVENT_DISCONNECTED
+                # so no additional checks are needed here.
+                raise ConnectionError("connection to the server is closed or disconnected")
+
+            continue  # otherwise, keep waiting for a server response
```
ajpotts, line 766: Do you need the `continue` here?  I think the `while True` automatically continues.

Also, should we limit the number or re-try in case it never completes?

arkouda/client.py
```diff
@@ -686,6 +696,75 @@ def disconnect(self) -> None:
...
+        self.socket.setsockopt(zmq.TCP_KEEPALIVE_INTVL, timeout)  # interval between probes
+        self.socket.setsockopt(zmq.TCP_KEEPALIVE_CNT, 3)  # probes before dropping
+
+        # Attach a monitor socket.
+        self.heartbeatSocket = self.socket.get_monitor_socket(zmq.EVENT_CLOSED | zmq.EVENT_DISCONNECTED)
+
+        if not self.heartbeatIsEnabled():
+            raise RuntimeError("heartbeatIsEnabled() returned false unexpectedly")
+
+    def heartbeatIsEnabled(self) -> bool:
```
ajpotts, line 736: This needs a docstring.

arkouda/client.py
```diff
@@ -686,6 +696,75 @@ def disconnect(self) -> None:
...
 
+    def setupHeartbeat(self, timeout) -> None:
+        """
+        Turn on/off a server heartbeat with the given timeout (in seconds).
+
+        The timeout is for waiting for a response from the server.
+        It also doubles as the heartbeat frequency.
+
+        If timeout is zero or negative, heartbeat is deactivated.
+
```
ajpotts, line 707: The docstrings need a Parameter section like this:
```python
    Parameters
    ----------
    timeout : int
        Timeout in seconds. Also determines the heartbeat frequency.
        If timeout <= 0, heartbeat monitoring is disabled.
```

arkouda/client.py
```diff
@@ -686,6 +696,75 @@ def disconnect(self) -> None:
         except Exception as e:
             raise RuntimeError(e)
 
+    def setupHeartbeat(self, timeout) -> None:
```
ajpotts, line 699: Function names should be snake case : `setup_heartbeat`

arkouda/client.py
```diff
@@ -1079,26 +1159,66 @@ def generic_msg(
...
+
+def note_for_server_log(message) -> None:
+    """
+    Add an INFO entry in the server log with the given message.
+
+    Intended for testing.
+    The server will consider itself "idle" despite serving this message.
+
+    Cf. ak.write_log (client) and clientLogMsg (server) allow fine-tuning
+    the message and interrupt server "idle"-ness.
```
ajpotts, line 1217: This docstring also needs a Parameters section.

ajpotts (approved):
Looks good.

arkouda/client.py
vasslitvinov: Good catch. I added it.

arkouda/client.py
```diff
@@ -1079,26 +1176,72 @@ def generic_msg(
...
+    The server will consider itself "idle" despite serving this message.
+
+    """
+    generic_msg("wait_for_async_activity")
+
+
+def server_sleep(seconds) -> None:
+    """
+    Have the server "sleep" for the given number of seconds.
+
```
ajpotts, line 1212: This needs a Parameters section too.

vasslitvinov: Added, in the rebased commit 904c7c3

ajpotts (approved):

PR #4752, opened 2025-07-29
ak.get_registration_config() and tweaks

vasslitvinov:
* Add `ak.get_registration_config()`.
* Populate a Generator in the client with GeneratorSymEntry instances on the server only for those types that are requested in `registration-config.json`.
* Minor code improvements I did while working on the following.

While there, redirected the file naming for bigint checkpoint data to the standard-ish `generateFilename()`.

This PR initially added checkpointing of `GeneratorSymEntry`. However during testing I realized that, while checkpointing per se works, restoring from a checkpoint does not "roll back" Arkouda properly. This is because part of the Generator state, specifically its `state` field, is stored in the client and not on the server. Therefore backed out this checkpointing addition in the second commit in this PR.

See #4709 for discussion on this.

### DIFF
--- a/arkouda/client.py
+++ b/arkouda/client.py
@@ -119,6 +119,7 @@ __all__ = [
     "disconnect",
     "shutdown",
     "get_config",
+    "get_registration_config",
     "get_max_array_rank",
     "get_mem_used",
     "get_mem_avail",
@@ -878,7 +879,8 @@ def connect(
     with an existing connection, the socket will be re-initialized.
 
     """
-    global connected, serverConfig, regexMaxCaptures, channel, registrationConfig
+    global channel
+    global connected, serverConfig, regexMaxCaptures, registrationConfig
 
     # send the connect message
     cmd = "connect"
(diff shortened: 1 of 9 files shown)


lydia-duncan (approved):
Looks good to me!

arkouda/client.py
```diff
@@ -1120,11 +1126,34 @@ def get_config() -> Mapping[str, Union[str, int, float]]:
...
     return serverConfig
 
 
+def get_registration_config():
+    """
+    Get the registration settings that the server was built with.
+
+    These registration settings are defined in the file `registration-config.json`
+    and snapshot at server build time.
+
```
ajpotts, line 1140: This is the format that we use for the Returns section of our docstrings:
```python
    Returns
    -------
    dict
        A mapping from parameter names to nested dictionaries, reflecting the
        structure of the original JSON file.
```

arkouda/numpy/dtypes.py
```diff
@@ -202,6 +203,60 @@ def dtype(dtype):
...
+    "bool": "bool",
+    "bigint": "bigint",
+    "string": "str",
+}
+
+
+def dtype_for_chapel(type_name: str):
+    """
+    Returns dtype() for the given Chapel type.
+
```
ajpotts, line 234: Again, the format for our docstrings is something like this:
``` python
Parameters
----------
type_name : str
    The name of the Chapel type, with or without the bit width.

Returns
-------
dtype
    The corresponding Arkouda dtype object.
```

ajpotts (approved):
Looks great.  I noticed to fixed some other minor bugs, which is appreciated.  Thanks!

ajpotts (approved):

PR #4757, opened 2025-07-31
Closes #4372: Doc duplication issue for Reorged modules

### LINKED ISSUE #4372: Doc duplication issue for Reorged modules
It appears that all of the modules that have been reorged for the numpy/scipy/pandas reorg are now showin up in their respective file and their original locations. The duplicate needs to be removed.

jaketrookman:
This PR (Closes [#4372](https://github.com/Bears-R-Us/arkouda/issues/4372)) removed the duplicated modules that are generated twice in the docs after the reorg. They now generate correctly and only once in their designated modules.


ajpotts (approved):
Looks good.  Thanks!

PR #4773, opened 2025-08-07
Deletion of SegString's values and offsets + more

vasslitvinov:
Enable deletion of the "values" and "offsets" arrays of a `SegStringSymEntry` in the server when it is deleted itself.

Implementation Notes

A `SegStringSymEntry` is deleted after it is removed from the SymTab, which happens (I assume) when the corresponding Strings object is deleted in the client. Prior to this PR, "values" and "offsets" were thought to be handled / deleted by the client:

```
// NOTE: Clean-up, the client side pieces go out of scope and issue deletions on their own
// so it is not necessary to manually remove these pieces from the SymTab
// st.deleteEntry(offsetsName);
// st.deleteEntry(valuesName);
```

however I did not observe that happening. So I added the removal of "values" and "offsets" from the SymTab, which should cause them to be deleted if nothing else references them.

While there, I added an early `throw` to `arraySegString()` on non-uint8 invocations, since it did not work for those anyway. This means that the rest of the body will be ignored / dropped in such cases, reducing the generated code size. I also simplified `arraySegString()` significantly and made it return a `SegStringSymEntry`, avoiding the need for the client to combine "values" and "offsets" in a separate message to the server in `pdarraycreation.py`.

While there also changed the "values" and "offsets" fields of  `SegStringSymEntry` from `var` to `const`, to clarify their nature.

### DIFF
--- a/src/MultiTypeSymEntry.chpl
+++ b/src/MultiTypeSymEntry.chpl
@@ -491,10 +491,10 @@ module MultiTypeSymEntry
         }
 
         override proc removeDependents(st: borrowed SymTab) throws {
-            // getSegString([],[],st) creates unnamed SymEntries
-            // and does not store them in the SymTab
-            if offsetsEntry.name != "" then st.deleteEntry(offsetsEntry.name);
-            if bytesEntry.name   != "" then st.deleteEntry(bytesEntry.name);
+            if st.contains(offsetsEntry.name) then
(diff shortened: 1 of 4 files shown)


ajpotts (approved):
Thanks for fixing this!

1RyanK (approved):
Looks good!

PR #4776, opened 2025-08-08
4688 intermittent ci failures error invoking driver compilation phase

ajpotts:
The purpose of this PR is to reduce the RAM usage for the multi-dimensional unit tests running the CI.  It adds a script that will comment out a list of modules from the `ServerModules.cfg` that do not need to be tested in the multi-dimensional case.  It also adds a pytest skip to any unit tests that depend on these modules when arkouda is compiled for the multidimesional case.

### DIFF
--- a/arkouda/numpy/random/generator.py
+++ b/arkouda/numpy/random/generator.py
@@ -715,7 +715,6 @@ class Generator:
         array([5 9 7 3 0 2 1 6 4 8])
 
         """
-
         from arkouda.client import generic_msg
 
         if not isinstance(x, pdarray):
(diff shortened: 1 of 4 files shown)


ajpotts:
I ran an experiment where I was able to get this PR to compile in the multi-dimensional case with these changes: https://github.com/Bears-R-Us/arkouda/pull/4593

ajpotts:
@vasslitvinov Separating out `pytest -c pytest.opts.ini` is actually so the tests can run when `CheckpointMsg` is not compiled.

jaketrookman (approved):
Looks good

PR #4780, opened 2025-08-13
Closes #4779 strip ak connect from examples

### LINKED ISSUE #4779: strip ak.connect() from examples
Strip `ak.connect()` from Examples sections of all docstrings.  This should not be used because it can cause `doctest` to hang.

ajpotts:
This PR removes `>>> ak.connect()` from the docstrings which can cause `doctest` to hang.  It also adds a unit tests that will fail if any `>>> ak.connect()` are added to the docstrings in the future.

Closes #4779 strip ak connect from examples

### DIFF
--- a/arkouda/comm_diagnostics.py
+++ b/arkouda/comm_diagnostics.py
@@ -38,7 +38,7 @@ Examples
 --------
 >>> import arkouda as ak
 >>> import arkouda.comm_diagnostics as cd
->>> ak.connect()
+
 >>> from arkouda.comm_diagnostics import start_comm_diagnostics, stop_comm_diagnostics, \
 get_comm_diagnostics, print_comm_diagnostics_table
 >>> start_comm_diagnostics()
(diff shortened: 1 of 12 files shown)


arkouda/comm_diagnostics.py
```diff
@@ -38,7 +38,7 @@
 --------
 >>> import arkouda as ak
 >>> import arkouda.comm_diagnostics as cd
->>> ak.connect()
```
e-kayrakli, line 41: Removals here make me worried a bit. Because you have the proper `import` lines, making me (as a user) think that this is complete code. But then, `ak.connect()` is dropped without any warning sign that for this to be actually complete, you need to connect to the server.

My recommendation would be to skip these lines in testing instead. I don't know whether the `SKIP` directive is per-line or per-block, though.

ajpotts: We already had this discussion as a team and decided to drop the `ak.connect()` from all the examples.  Otherwise it would make sense to add them to all the examples in the entire project for consistency.

jaketrookman (approved):
Looks good

PR #4790, opened 2025-08-15
enable asynchronous Arkouda client requests

ajpotts:
TBD


ajpotts:
This one was too far out of date to easily rebase so I had to create a new PR: https://github.com/Bears-R-Us/arkouda/pull/2008

1RyanK (approved):
Looks good!

e-kayrakli (commented):
Unless I am missing something, the core feature added here feels too close to the [heartbeat feature added recently](https://github.com/Bears-R-Us/arkouda/pull/4750). The purpose of that was to get a clear indication in case the server goes down without giving any signal to the user. At a high-level this seems to be doing the same thing. Though this one has nice user-facing new things, too. Can the new niceties be tacked on the ZeroMQ-based solution instead of relying on asyncio?

PR #4819, opened 2025-08-25
Resolves Issue 4810, cumSum and cumProd performance

### LINKED ISSUE #4810: cumsum and cumprod performance regression
A performance regression was noticed in `cumsum` and `cumprod`:

- https://chapel-lang.org/perf/arkouda/16-node-xc/?startdate=2025/03/16&enddate=2025/08/21&graphs=scanperformance

- https://chapel-lang.org/perf/arkouda/chapcs/?startdate=2025/03/13&enddate=2025/08/21&graphs=scanperformance

- https://chapel-lang.org/perf/arkouda/16-node-hpe-apollo-hdr/?startdate=2025/02/21&enddate=2025/08/21&graphs=scanperformance

Most likely is related to : PR https://github.com/Bears-R-Us/arkouda/pull/4755

drculhane:
Closes #4810

Update to update:

Based on Engin's recent feedback, this should now be ready for review.  The final changes are very simple.  In the 1D case, it reverts to the original code.  The rest of the comments below are now out of date.

-------------

Update: I added plotting to the performance tool.  Enter any value as a 3rd argument and it will plot the measured time for all of the trials.  Still not sure this is helpful.

I commented out the console logging because that has more affect on runtime than the calculations do.

--------

Bottom Line Up Front: I'm trying to nail down the performance problem noted in Issue 4810.  So far I'm not succeeding.  This PR has both old and new versions, and some performance measuring code.  (The benchmark doesn't run the old ones, because there are no corresponding numpy names for them.)

The old versions are:  _ak.oldcumsum_ and _ak.oldcumprod_.  The new ones are: _ak.cumsum_ and _ak.cumprod_.

I've attached a python program, _pperformance.py_, meant to time the old vs. new.
- _python3 pperformance 100 2000_ will run 100 trials of a pdarray 2000 elements long, and display mean/std
- _python3 pperformance 100 2000 yes_ will do all of that, and also show the plot described above.

What do I test?  I test all four functions via their python interfaces (_ak.oldcumsum, ak.oldcumprod, ak.cumsum, ak.cumprod_) and I also do tests that directly invoke the four chapel procs via _generic_msg_ (these runs ignore the output).  This meant to reduce the effect of the python interpreter on the timing.

The input data is not random.  It's ak.arange(size) + 1.  Since it's always the same, the test results should be consistent.  They're not.  Run it and see.  According to this performance tool, the new versions are almost always worse than the old ones, even though the code is virtually identical.

Here's the test program: 
[pperformance.py](https://github.com/user-attachments/files/21992739/pperformance.py)

Some background:  when I wrote PR 4755, I updated the chapel proc _cumSum_ and _cumProd_ to implement the more complicated functions cumulative_sum and cumulative_prod.  These allow multi-dim, axis slicing, and the boolean _includeInitial_.  When the input rank is 1 and the flag is False, _cumulative_sum_ should give the same results as _cumsum_.  Similarly for _cumulative_prod_ and _cumprod_.

### DIFF
--- a/src/StatsMsg.chpl
+++ b/src/StatsMsg.chpl
@@ -119,17 +119,25 @@ module StatsMsg {
     }
 
     @arkouda.registerCommand()
-    proc cumSum(const ref x: [?d] ?t, axis: int, includeInitial: bool): [] t throws {
+    proc cumSum(const ref x: [?d] ?t, axis: int, includeInitial: bool): [] t throws
+        where t!= bool { // bool case was already converted to int python-side
+
       if d.rank == 1 {
-        var cs = makeDistArray(if includeInitial then x.size+1 else x.size, t);
-        cs[if includeInitial then 1.. else 0..] = (+ scan x):t;
-        return cs;
+
+          if !includeInitial {
+            return (+ scan x); 
+          } else {
+            var cs = makeDistArray(x.size+1, t);
+            cs[1..] = (+ scan x);
+            return cs;
+          }
+
       } else {
         var cs = makeDistArray(if includeInitial then expandedDomain(d, axis) else d, t);
 
         forall (slice, _) in axisSlices(d, new list([axis])) {
           const xSlice = removeDegenRanks(x[slice], 1),
-                csSlice = (+ scan xSlice):t;
+                csSlice = (+ scan xSlice); 
 
           for idx in slice {
             var csIdx = idx;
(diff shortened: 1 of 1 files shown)


src/StatsMsg.chpl
```diff
@@ -119,17 +119,32 @@ module StatsMsg {
...
+//      cs[if includeInitial then 1.. else 0..] = (+ scan x):t;
+//      return cs;
+
       } else {
         var cs = makeDistArray(if includeInitial then expandedDomain(d, axis) else d, t);
 
         forall (slice, _) in axisSlices(d, new list([axis])) {
           const xSlice = removeDegenRanks(x[slice], 1),
-                csSlice = (+ scan xSlice):t;
+                csSlice = (+ scan xSlice); //:t;
```
e-kayrakli, line 147: Do you remember why this cast was necessary to begin with? It will be an element-wise operation and could take some time.

e-kayrakli (commented):
Thanks, Andy! And I apologize for the delay. Looking at the diff, it is a good bet that this will improve performance. The parts that you are replacing has two potential performance issues:

- You create `cs` and return it after some edits. This will make compiler's job difficult removing an array out of the operation. Whereas just `+ scan x` can have better chance at the compiler avoiding a redundant array copy.
- Even with `includeInitial==false` which is the faster option no matter what, `cs[0..] = (+ scan x):t;` is an assignment into an array slice rather than an array (regardless of the slice being the full array, practically speaking). That will take a different codepath in the modules that can handle more complicated scenarios that slice assignment may bring. Imaginably, that will cost some time. Moreover, that line also has a cast, which should be unnecessary, I think.

In any case https://github.com/chapel-lang/chapel/pull/27828 will put this branch in the performance playground allowing us to collect some data. You can finalize the PR after we know for sure that this will recover performance.

PR #4821, opened 2025-08-25
Add pytest-benchmark dependency to setup.py

jabraham17:
This PR adds the missing pytest-* dependencies to setup.py, which is required to run the benchmark_v2 infrastructure


ajpotts (approved):
Good catch.  Thanks!

jaketrookman (approved):

Issue #4800, opened 2025-08-20
Standardize checks for axis validity and handling of negative axes


drculhane:
Checking of axis validity and handling of negative axes are things that pop up in several arkouda functions, especially those that are modeled on numpy.  There isn't a standard approach, and since this is such a common thing to do, there probably should be.

PR #4834, opened 2025-08-28
Closes 4831, standardizes axis checking for cases where axis can only be an integer

### LINKED ISSUE #4831: Standardize axis handling where the axis can only be an integer, not a tuple
Various functions allow the axis parameter to be an int or a tuple.  It will be simpler to address the integer axis handling separately.

drculhane:
Closes #4831, which is a subset of #4800

This covers the cases where the axis parameter can only be an integer.  This only applies to array_api and numpy functions, and not the pandas functions (where axis can be 0 or 1, "index" or "columns").

It doesn't cover the cases where the axis parameter can be a tuple.

Quick summary:

There is now a function in numpy/util.py called _integer_axis_validation.  It returns a bool and an int.  If the bool is false, the given axis wasn't valid, and in (I think) every case, an IndexError is raised.

There are several functions in array_api that just pass their parameters to arkouda/numpy functions with the same name.  In those cases, I don't bother to check at the array_api level.

Note that array_api's vecdot accepted an axis parameter, but didn't use it.  I changed that so that it passes the axis argument to the numpy function.  Also note that vecdot handles the axis validation differently than other functions, so it doesn't use the standard validation.

Also also note that expand_dims and stack are unique in that they check the validity against a rank 1 greater than the given input.

### DIFF
--- a/arkouda/array_api/linalg.py
+++ b/arkouda/array_api/linalg.py
@@ -42,4 +42,4 @@ def vecdot(x1: Array, x2: Array, /, *, axis: int = -1) -> Array:
 
     from .array_object import Array
 
-    return Array._new(ak_vecdot(x1._array, x2._array))
+    return Array._new(ak_vecdot(x1._array, x2._array, axis))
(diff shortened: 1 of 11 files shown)


1RyanK (approved):
Looks good!

jaketrookman (approved):
Looks good

ajpotts (approved):

PR #4854, opened 2025-09-04
Closes #4853: Problem in shuffle docstring

### LINKED ISSUE #4853: Problem in shuffle docstring
**Describe the bug**
Within the docstring for `shuffle` in `numpy/random/generator.py`, it has
```python
        >>> import arkouda as ak
        >>> rng = ak.random.default_rng(18)
        >>> pda = ak.arange(10)
        >>> pda
        array([0 1 2 3 4 5 6 7 8 9])
        >>> rng.shuffle(pda, method="FisherYates")
        >>> pda
        array([0 8 2 7 9 4 6 3 5 1])
        >>> rng.shuffle(pda, method="MergeShuffle")
        >>> pda
        array([5 6 9 3 8 2 7 0 4 1])
```
This is correct for 1 locale, but when running the server for different numbers of locales this doesn't work.

**To Reproduce**
1. Start a server with one locale, connect to it and run the lines above and the output matches.
2. Start a server with two locales, connect to it and run the lines above and instead get `array([4 7 9 0 6 3 1 8 5 2])`. Uh-oh...

**Expected behavior**
It'd be nice if it always came back the same, but failing that, the docstring test needs to be disabled.

**Error Message**
```
arkouda.numpy.random.generator.Generator.shuffle
Failed example:
    pda
Expected:
    array([5 6 9 3 8 2 7 0 4 1])
Got:
    array([4 7 9 0 6 3 1 8 5 2])
```

**Is this a Blocking Issue**
- Causes problems in doctest.

1RyanK:
This just uses the annotation to skip the doctest if the number of locales is not 1.

Closes #4853: Problem in shuffle docstring

### DIFF
--- a/tests/numpy/random_test.py
+++ b/tests/numpy/random_test.py
@@ -22,6 +22,7 @@ class TestRandom:
         result = doctest.testmod(random, optionflags=doctest.ELLIPSIS | doctest.NORMALIZE_WHITESPACE)
         assert result.failed == 0, f"Doctest failed: {result.failed} failures"
 
+    @pytest.mark.skip_if_nl_neq(1)
     def test_random_generator_docstrings(self):
         import doctest
 


jaketrookman (approved):

ajpotts (approved):
Thanks!

PR #4865, opened 2025-09-05
Closes #4863:  16-array Average rate always returns zero under benchm…

### LINKED ISSUE #4863: 16-array Average rate always returns zero under benchmark_v2 groupby
`16-array Average rate` always returns zero in `groupby.dat` under `benchmark_v2` benchmarks.

ajpotts:
This PR fixes a minor issue where the Average rate for `groupby.dat` and `coargsort.dat` was always zero when the number of arrays was `16` and the benchmark size was `10`.  This was due to integer division used to determine the individual array sizes, which will be zero when the number of arrays is smaller than the size (`pytest.N`) of the benchmark.

Closes #4863:  16-array Average rate always returns zero under benchmark_v2 groupby


1RyanK (approved):
Looks good!

PR #4873, opened 2025-09-08
Closes #4872: ak.randint doesn't behave the same for bool as numpy

### LINKED ISSUE #4872: `ak.randint` doesn't behave the same for `bool` as `numpy`
**Describe the bug**
Whenever `low` and `high` are passed to `ak.randint`, they are ignored and an array with 50/50 T/F values is generated. There's also some errors raised for invalid values.

1RyanK:
Fixed up the Chapel code for randint, and then also updated how it's handled in the Python-side. Then tests failed so updated several of them.

Closes #4872: ak.randint doesn't behave the same for bool as numpy

### DIFF
--- a/arkouda/numpy/random/legacy.py
+++ b/arkouda/numpy/random/legacy.py
@@ -112,7 +112,14 @@ def randint(
         Raised if dtype.name not in DTypes, size is not an int, low or high is
         not an int or float, or seed is not an int
     ValueError
-        Raised if size < 0 or if high < low
+        If ``size < 0`` or ``ndim < 1``.
+        For ``dtype=int64``, if ``low >= high``.
+        For ``dtype=bool`` (NumPy-compatible), one of:
+            * ``"high <= 0"`` if ``high <= 0``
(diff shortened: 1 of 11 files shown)


arkouda/numpy/random/legacy.py
```diff
@@ -157,21 +164,45 @@ def randint(
...
+        if cast(int, high) > 2:
+            raise ValueError("high is out of bounds for bool")
+        if cast(int, low) >= cast(int, high):
+            raise ValueError("low >= high")
+    elif dtype == akint64:
+        # General integer path: half-open interval requires high > low
+        if cast(int, high) <= cast(int, low):
+            raise ValueError("low >= high")
+    else:
+        # Float path (Arkouda-specific behavior): still require high >= low
```
drculhane, line 202: From the code and the context, I think you mean "still require high > low"

drculhane (approved):
Good work.  I noted a comment that I _think_ needs fixing, but other than that, it looks good and passes all unit tests.

drculhane (approved):
Looks good.

PR #4889, opened 2025-09-11
Fixes common type promotion in array_api concat and stack

### LINKED ISSUE #4880: Fix common type handling in array_api concat and stack
The array_api concat and stack functions produce a float output even if all inputs are int, because it uses np.common_type.  But numpy concat and numpy stack don't use that, and will produce an int output if all inputs are int.  This should be addressed.

drculhane:
Addendum

I neatened this a bit, with a function _has_type, so that the check against all  types would be a bit more readable and concise.

Closes #4880 

Both concat and stack were using numpy's promote_to_common_dtype to compute the result type.  But numpy's concat and stack don't use that.  They use the conventional numpy rule of:

any float -> output = float, else
any int AND any uint -> output = float, else
any int -> output = int, else
any uint -> output = uint, else
any bool -> output = bool, else
error

This PR implements the correct type promotion.

**Possible point of discussion**:

I've done the check (and type conversion) python-side.  The "where" function, implemented in EfuncMsg.chpl, does the same check chapel-side.  On the one hand, I would prefer to be consistent.  On the other hand, that would create multiple instantiations of the chapel-side proc, and we've recently been running up against instantiation limits.

I am entirely open to discussion on whether to move this chapel-side.

### DIFF
--- a/arkouda/array_api/manipulation_functions.py
+++ b/arkouda/array_api/manipulation_functions.py
@@ -130,7 +130,7 @@ def _resolve_common_dtype(pdarrayList):
     elif _has_type(akbool, pdarrayList):
         dt = akbool
     else:
-        raise TypeError("All arrays in concat must be float, int, uint, or bool.")
+        raise TypeError("All arrays in operation must be float, int, uint, or bool.")
 
     dt = akdtype(dt)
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good

PR #4892, opened 2025-09-11
Improve CI to automate builds (Part 2)

jabraham17:
Improves the CI by switching it to use the new auto-built images from https://github.com/Bears-R-Us/arkouda/pull/4891

This PR also drops Chapel versions 2.0-2.3 from the CI. A later PR will add 2.6

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,7 +4,8 @@ on: [pull_request, merge_group, workflow_dispatch]
 
 env:
   ARKOUDA_QUICK_COMPILE: true
-  
+  ARKOUDA_SKIP_CHECK_DEPS: true
+
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -259,6 +270,7 @@ jobs:
         dnf update -y
         dnf install -y zlib-devel gcc gcc-c++ make zlib-devel bzip2 bzip2-devel readline-devel sqlite sqlite-devel openssl openssl-devel libffi-devel xz-devel zlib-devel
 
+    # TODO: why does this need to build its own Python when the container has its own?
```
ajpotts, line 273: We test with the most recent version of python, but the almalinux container only comes with 3.9:
```
# /bin/python3 --version
Python 3.9.21
```

PR #4904, opened 2025-09-13
Closes #4903:  Resolve E271 Errors

### LINKED ISSUE #4903: Resolve E271 Errors
For example: 
```
pdarrayclass_test.py:167:34: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
```

ajpotts:
Resolves the `E271` errors.  Also remove error codes from the `setup.cfg` that are no longer causing an error.

Closes #4903:  Resolve E271 Errors


jaketrookman (approved):
Looks good, except the title of the PR does not match the error code

PR #4906, opened 2025-09-15
Improve parquet read implementation and multiple column read performance

e-kayrakli:
This PR has two main goals:

- improve the performance of parquet reads, particularly when all the columns of the file(s) are read
- start improving the implementation on C++ and Chapel sides

### User-facing Change:

- This PR deprecates `has_non_float_nulls: Bool`, in favor of `null_handling: {"none", "only floats", "all"}`
  - We need a 3-way switch to enable fast floating point reads when the user knows their data contains no nulls. Old boolean-based switch cannot capture that. The new `null_handling` formal to `read_parquet` defaults to "only floats" to match with the older behavior.
  - Right now, `null_handling` is effective only when reading all columns. `has_non_float_nulls` is used when only a subset of columns are read. It is a TODO to phase out `has_non_float_nulls`, among other things.

### Why do we need a new implementation for performance?

Parquet is a columnar format. However, that doesn't mean that all data for a column is stored consecutively. For large files (512M by default, I think), Parquet files will be written in consecutive _row groups_ instead of storing large columns one after another. Therefore, a more natural way of reading a parquet file is to read row-group-by-row-group while going through all the columns. Considering reading all columns as a common use case, it is important to read parquet files in that way for performance. Right now, if you have 10 columns, you open and read the entire file(s) 10 times, one for each column, skipping 9 columns during each read.

#### A small optimization along the way

With the new `null_handling` control flag, we can also enable batch reads of floating point types where prior to this PR we were reading them one-by-one to be able to catch nulls. I think even with nulls at play, we should be able to do better than that, but that's a post-merge TODO.

#### Performance Improvements

This is using ~400GBs of data in 5 columns partitioned into 128 files, read by 16 locales:

|           | `main` (s) | branch (s) | improvement (x)
|---------|-----------------|---------------|--------
| min    | 11.43           | 7.60          | 1.50
| max   | 22.58          | 14.96        | 1.51
| ave    |  16.69         | 9.38          | 1.78

This is using ~30GBs of data in 1000 columns partitioned into 128 files, read by 16 locales:

|           | `master` (s) | branch (s) | improvement (x)
|---------|-----------------|---------------|--------
| min    | 335.97           | 9.02          | 37.24
| max   | 335.97          | 15.64        | 21.48
| ave    | 335.97         | 9.98          | 33.67

### Implementation details:

- If the user doesn't pass any dataset names to `read_parquet`, we read all columns. This PR makes that case take an entirely new codepath with no modifications to the subset-of-columns code path. I intend to remove this old code path in favor of the new one even for subset-of columns. That's a post-merge TODO.
- This new code path introduces the following on the C++ side:
  - `namespace akcpp` on the C++ side. I want more code to live in this namespace going forward. Within that namespace:
  - `struct AkReadColOp` on the C++ side to represent type-agnostic information for reading a single column.
  - This type has a template `read()` function for reading a column with specific dynamic type
  - `read()` has some template specializations for handling some specific cases like short integrals that are not represented on the Chapel side and need to be cast to bigger ones
  - The `read()` family of functions has _a lot of code redundancy_ that I inherited from the earlier implementation. Refactoring those redundancies into helpers is an obvious TODO.  
  - `readAllCols` function to handle the operation. This is the function that selects the correct `read()` instantiation based on the dynamic type.
- This new code path introduces the following on the Chapel side:
  - `readAllColsParquetMsg` which is the main message handler here. The implementation is too much of a Frakenstein experiment where old style/structure is mixed with the new. Going forward, I want this function to be the general parquet read function, so it needs more improvements.
  - The implementation of this stores read metadata in `pqReadColOp`. This is deceptively similar to `AkReadColOp`. But it actually stores information for the whole read operation and it is not per-column. I want to consider whether we can have single metadata shared between C++ and Chapel.
  - This `pqReadColOp` determines whether we can read all columns at the same time based on the dynamic types of the columns. If we can't it will fall back to the old implementation. Currently that fallback is a bit rough as we call the message handler. I think that's a bad practice and the core logic of the read has to be available without the message handling. We'll need to reconcile the two different reads in any case, so that's part of post-merge TODOs.
  - I used this type to do column-per-column reads, too. That's the only change in the existing code path. I started off thinking that I can do all the refactoring in one go, but decided against it. Nevertheless, I left that part in. Currently that causes `pqReadColOp` to be more generic than I'd like. It will be fixed in the next refactor wave.
  - Simplified error location reporting
  - Several new small helpers and simplified helpers 

### Test Status:
- [x] local parquet unit testing
- [x] oversubscribed parquet unit testing

### Post-Merge TODO
- [ ] Refactor `read()` functions into having more helpers
  - [ ] Consider handling per-batch post-processing differently, maybe with functors.
- [ ] use the `akcpp::readAllCols` function for reading subset of columns, too and remove the old read implementation completely. Similarly for `readAllColsParquetMsg` on Chapel side
  - [ ] Fall back to col-per-col read in a more graceful way 
  - [ ] completely deprecate `has_non_float_nulls`
- [ ] Can we read floating point columns with nulls in them in batches and process nulls differently? 
- [ ] Can we use a single metadata type for C++ and Chapel?
- [ ] Use message registration attributes for Parquet message handlers
- [ ] Short integer and float reads use a temporary buffer. We can get rid of that by reading into the larger Chapel buffer and spacing things out in reverse order.
- [ ] Handle `ARROWERROR` and error messages populated by the C level better and more consistently
- [ ] https://github.com/Bears-R-Us/arkouda/issues/4925

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -17,6 +17,7 @@ module ParquetMsg {
   use Map;
   use SegmentedString;
   use IOUtils;
+  use ParquetSharedEnums;
 
   enum CompressionType {
     NONE=0,
@@ -29,7 +30,9 @@ module ParquetMsg {
 
 
   // Use reflection for error information
-  use Reflection;
+  import Reflection.{getModuleName as getM,
+                     getRoutineName as getR,
+                     getLineNumber as getL};
   require "ReadParquet.h";
   require "ReadParquet.o";
   require "WriteParquet.h";
@@ -125,7 +128,7 @@ module ParquetMsg {
       subdoms[i] = {offset..#lengths[i]};
       offset += lengths[i];
     }
-    return (subdoms, (+ reduce lengths));
+    return subdoms;
   }
 
   proc getRGSubdomains(bytesPerRG: [?D] ?t, maxRowGroups: int) {
@@ -151,6 +154,57 @@ module ParquetMsg {
                     hasNonFloatNulls, hasWhereNull=false);
   }
 
+  class ParquetReadError: ErrorWithContext {
+    proc init(msg: string, moduleName: string, routineName: string,
+              lineNumber: int(64)) {
+      super.init(msg, lineNumber, routineName, moduleName,
+                 errorClass="ParquetReadError");
+    }
+  }
+
+  proc processFilelist(Filelist: [] string) throws {
+    if Filelist.size == 1 {
+      if Filelist[0].strip().size == 0 then
+          throw new ParquetReadError("Filelist was empty", getM(), getR(), getL());
+
+      var GlobRes = glob(Filelist[0]);
+      pqLogger.debug(getM(), getR(), getL(),
+                     "glob expanded %s to %i files".format(Filelist[0],
+                                                           GlobRes.size));
+      if GlobRes.size == 0 {
+          throw new ParquetReadError("The wildcarded filename %s either " +
+                                     "corresponds to files inaccessible to " +
+                                     "Arkouda server or files of an invalid " +
+                                     "format".format(Filelist[0]),
+                                     getM(), getR(), getL());
+      }
+
+      // Glob returns filenames in weird order. Sort for consistency
+      sort(GlobRes);
+      return GlobRes;
+    } else {
+      return Filelist;
+    }
+  }
+
+  proc getCPtrToAkArrayEntry(e: borrowed GenSymEntry,
+                             ctype: c_int, off: int): c_ptr(void) throws {
+    var ret: c_ptr(void);
+    select typeFromCType(ctype) {
+      when ArrowTypes.int64  do
+        ret = c_ptrTo((e:(borrowed SymEntry(int(64),1))).a[off]);
+      when ArrowTypes.uint64 do
+        ret = c_ptrTo((e:(borrowed SymEntry(uint(64),1))).a[off]);
+      when ArrowTypes.double do
+        ret = c_ptrTo((e:(borrowed SymEntry(real(64),1))).a[off]);
+      otherwise do
+        throw new NotImplementedError("Unexpected column type while " +
+                                      "reading parquet file", getL(), getR(), getM());
+    }
+    return ret;
+  }
+
+
   /*
      whereNull will be populated by the CPP interface, where `true` would mean a
    0 (null) having been read.
@@ -161,7 +215,7 @@ module ParquetMsg {
                        param hasWhereNull=true) throws {
     extern proc c_readColumnByName(filename, arr_chpl, where_null_chpl, colNum, numElems, startIdx, batchSize, byteLength, hasNonFloatNulls, errMsg): int;
 
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
     var fileOffsets = (+ scan sizes) - sizes;
     
     coforall loc in A.targetLocales() with (ref A) do on loc {
@@ -184,7 +238,7 @@ module ParquetMsg {
                                   intersection.size, intersection.low - off,
                                   batchSize, byteLength, hasNonFloatNulls,
                                   c_ptrTo(pqErr.errMsg)) == ARROWERROR {
-              pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+              pqErr.parquetError(getL(), getR(), getM());
             }
           }
         }
@@ -194,7 +248,7 @@ module ParquetMsg {
 
   proc readStrFilesByName(ref A: [] ?t, filenames: [] string, sizes: [] int, dsetname: string) throws {
     extern proc c_readStrColumnByName(filename, arr_chpl, colname, numElems, batchSize, errMsg): int;
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
     
     coforall loc in A.targetLocales() do on loc {
       var locFiles = filenames;
@@ -211,7 +265,7 @@ module ParquetMsg {
             if c_readStrColumnByName(filename.localize().c_str(), c_ptrTo(col),
                                      dsetname.localize().c_str(), filedom.size,
                                      batchSize, c_ptrTo(pqErr.errMsg)) == ARROWERROR {
-              pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+              pqErr.parquetError(getL(), getR(), getM());
             }
             A[filedom] = col;
           }
@@ -222,7 +276,7 @@ module ParquetMsg {
 
   proc readListFilesByName(A: [] ?t, rows_per_file: [] int, seg_sizes: [] int, offsets: [] int, filenames: [] string, sizes: [] int, dsetname: string, ty) throws {
     extern proc c_readListColumnByName(filename, arr_chpl, colNum, numElems, startIdx, batchSize, errMsg): int;
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
     var fileOffsets = (+ scan sizes) - sizes;
     var segmentOffsets = (+ scan rows_per_file) - rows_per_file;
     
@@ -242,7 +296,7 @@ module ParquetMsg {
             if c_readListColumnByName(filename.localize().c_str(), c_ptrTo(col),
                                   dsetname.localize().c_str(), filedom.size, 0,
                                   batchSize, c_ptrTo(pqErr.errMsg)) == ARROWERROR {
-              pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+              pqErr.parquetError(getL(), getR(), getM());
             }
             A[filedom] = col;
           }
@@ -252,7 +306,7 @@ module ParquetMsg {
   }
 
   proc calcListSizesandOffset(seg_sizes: [] ?t, filenames: [] string, sizes: [] int, dsetname: string) throws {
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
 
     var listSizes: [filenames.domain] int;
     var file_offset: int = 0;
@@ -275,7 +329,7 @@ module ParquetMsg {
   }
 
   proc calcStrSizesAndOffset(offsets: [] ?t, filenames: [] string, sizes: [] int, dsetname: string) throws {
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
 
     var byteSizes: [filenames.domain] int;
 
@@ -298,7 +352,7 @@ module ParquetMsg {
   }
 
   proc calcStrListSizesAndOffset(offsets: [] ?t, filenames: [] string, sizes: [] int, dsetname: string) throws {
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
 
     var byteSizes: [filenames.domain] int;
 
@@ -322,7 +376,7 @@ module ParquetMsg {
 
   proc getNullIndices(A: [] ?t, filenames: [] string, sizes: [] int, dsetname: string, ty) throws {
     extern proc c_getStringColumnNullIndices(filename, colname, nulls_chpl, errMsg): int;
-    var (subdoms, length) = getSubdomains(sizes);
+    var subdoms = getSubdomains(sizes);
     
     coforall loc in A.targetLocales() do on loc {
       var locFiles = filenames;
@@ -337,7 +391,7 @@ module ParquetMsg {
             var col: [filedom] t;
             if c_getStringColumnNullIndices(filename.localize().c_str(), dsetname.localize().c_str(),
                                             c_ptrTo(col), pqErr.errMsg) {
-              pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+              pqErr.parquetError(getL(), getR(), getM());
             }
             A[filedom] = col;
           }
@@ -357,7 +411,7 @@ module ParquetMsg {
                                              c_ptrTo(pqErr.errMsg));
     
     if byteSize == ARROWERROR then
-      pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+      pqErr.parquetError(getL(), getR(), getM());
     return byteSize;
   }
 
@@ -372,7 +426,7 @@ module ParquetMsg {
                                              c_ptrTo(pqErr.errMsg));
     
     if byteSize == ARROWERROR then
-      pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+      pqErr.parquetError(getL(), getR(), getM());
     return byteSize;
   }
 
@@ -387,7 +441,7 @@ module ParquetMsg {
                                              c_ptrTo(pqErr.errMsg));
     
     if listSize == ARROWERROR then
-      pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+      pqErr.parquetError(getL(), getR(), getM());
     return listSize;
   }
   
@@ -398,9 +452,43 @@ module ParquetMsg {
     var size = c_getNumRows(filename.localize().c_str(),
                             c_ptrTo(pqErr.errMsg));
     if size == ARROWERROR {
-      pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+      pqErr.parquetError(getL(), getR(), getM());
     }
     return size;
+  }
+
+  proc typeFromCType(ctype: c_int) throws {
+    if ctype == ARROWINT64 then return ArrowTypes.int64;
+    else if ctype == ARROWINT32 then return ArrowTypes.int32;
+    else if ctype == ARROWUINT32 then return ArrowTypes.uint32;
+    else if ctype == ARROWUINT64 then return ArrowTypes.uint64;
+    else if ctype == ARROWBOOLEAN then return ArrowTypes.boolean;
+    else if ctype == ARROWSTRING then return ArrowTypes.stringArr;
+    else if ctype == ARROWDOUBLE then return ArrowTypes.double;
+    else if ctype == ARROWFLOAT then return ArrowTypes.float;
+    else if ctype == ARROWLIST then return ArrowTypes.list;
+    else if ctype == ARROWDECIMAL then return ArrowTypes.decimal;
+    throw getErrorWithContext(getL(), getM(), getR(),
+                              msg="Unrecognized Parquet data type",
+                              errorClass="ParquetError");
+    return ArrowTypes.notimplemented;
+  }
+
+  proc ctypeFromType(t: ArrowTypes) throws {
+    if      t == ArrowTypes.int64     then return ARROWINT64;
+    else if t == ArrowTypes.int32     then return ARROWINT32;
+    else if t == ArrowTypes.uint32    then return ARROWUINT32;
+    else if t == ArrowTypes.uint64    then return ARROWUINT64;
+    else if t == ArrowTypes.boolean   then return ARROWBOOLEAN;
+    else if t == ArrowTypes.stringArr then return ARROWSTRING;
+    else if t == ArrowTypes.double    then return ARROWDOUBLE;
+    else if t == ArrowTypes.float     then return ARROWFLOAT;
+    else if t == ArrowTypes.list      then return ARROWLIST;
+    else if t == ArrowTypes.decimal   then return ARROWDECIMAL;
+    throw getErrorWithContext(getL(), getM(), getR(),
+                              msg="Unrecognized Parquet data type",
+                              errorClass="ParquetError");
+    return ARROWERROR;
   }
 
   proc getArrType(filename: string, colname: string) throws {
@@ -409,27 +497,12 @@ module ParquetMsg {
     var arrType = c_getType(filename.localize().c_str(),
                             colname.localize().c_str(),
                             c_ptrTo(pqErr.errMsg));
+
     if arrType == ARROWERROR {
-      pqErr.parquetError(getLineNumber(), getRoutineName(), getModuleName());
+      pqErr.parquetError(getL(), getR(), getM());
     }
     
-    if arrType == ARROWINT64 then return ArrowTypes.int64;
-    else if arrType == ARROWINT32 then return ArrowTypes.int32;
-    else if arrType == ARROWUINT32 then return ArrowTypes.uint32;
-    else if arrType == ARROWUINT64 then return ArrowTypes.uint64;
-    else if arrType == ARROWBOOLEAN then return ArrowTypes.boolean;
-    else if arrType == ARROWSTRING then return ArrowTypes.stringArr;
-    else if arrType == ARROWDOUBLE then return ArrowTypes.double;
-    else if arrType == ARROWFLOAT then return ArrowTypes.float;
-    else if arrType == ARROWLIST then return ArrowTypes.list;
-    else if arrType == ARROWDECIMAL then return ArrowTypes.decimal;
-    throw getErrorWithContext(
-                  msg="Unrecognized Parquet data type",
-                  getLineNumber(),
-                  getRoutineName(),
-                  getModuleName(),
-                  errorClass="ParquetError");
-    return ArrowTypes.notimplemented;
+    return typeFromCType(arrType);
   }
 
   proc getListData(filename: string, dsetname: string) throws {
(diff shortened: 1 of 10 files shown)


src/ParquetMsg.chpl
```diff
@@ -871,6 +925,520 @@ module ParquetMsg {
...
+        return floatEntry: AbstractSymEntry;
+      }
+      else {
+        return se: AbstractSymEntry;
+      }
+    }
+
+    inline proc _postProcess(se: SymEntry(?t), colIdx) throws {
+      return se;
+    }
```
jabraham17, line 1438: consider having `_postProcess` also return `AbstractSymEntry` to prevent extra generics. I would suggest using an explicit return type of `AbstractSymEntry` for both, like `postProcess`

src/ParquetMsg.chpl
```diff
@@ -1064,41 +1589,53 @@ module ParquetMsg {
...
+    defer {
+      extern proc c_free_string(ptr);
+      c_free_string(res);
+    }
+    if c_getDatasetNames(filename.c_str(), c_ptrTo(res), false,
+                         c_ptrTo(pqErr.errMsg)) == ARROWERROR {
+      pqErr.parquetError(L(), R(), M());
+    }
+    var datasets: string;
+    try! datasets = string.createCopyingBuffer(res, strlen(res));
```
jabraham17, line 1633: avoid try! to prevent potential server crash

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+// specialization for storing 32-bit data in 64-bit arrays
+template<> struct ChplBuf<arrow::Int32Type>  { using ElemType = int64_t; };
+template<> struct ChplBuf<arrow::UInt32Type> { using ElemType = uint64_t; };
+template<> struct ChplBuf<arrow::FloatType>  { using ElemType = double; };
+
+// TODO The implementation seemingly reads 128-bit decimal values into
+// Chapel `real`s (or C `double`s). I don't see how this can work properly with
+// large enough data. For now, I think this keeps the implementation practically
+// same as the old, likely carrying the same bug into the new implementation,
+// unfortunately.
```
jabraham17, line 24: Is there an issue about this?

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+  using PqType = typename parquet::FixedLenByteArray;
+};
+} // end anonymous namespace
+
+// Engin: We could consider exposing this struct to Chapel. Note that in
+// ParquetMsg.chpl there is a somewhat similar type. However, this one is for a
+// single column read
+// Based on what we decide, the implementation of this struct can be mode more
+// C++-like or could stay more C-like.
+struct AkColReadOp {
```
jabraham17, line 49: nit, this is already inside a namespace called `akcpp`, no reason to further prefix the name with `Ak`

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+    // note that floats are the types that natively support nulls. This generic
+    // read() shouldn't do anything with floats
+    if (nullMode == noNulls || nullMode == onlyFloats) {
+      while (reader->HasNext() && row_idx < numElems) {
+        if((numElems - row_idx) < batchSize) // adjust batchSize if needed
+          batchSize = numElems - row_idx;
+        reader->ReadBatch(batchSize, nullptr, nullptr, &chpl_ptr[row_idx],
+                          &values_read);
+        row_idx+=values_read;
+        num_read+=values_read;
```
jabraham17, line 85: nit, formatting

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+      }
+    }
+    else {
+      int16_t definition_level; // nullable type and only reading single records in batch
+      while (reader->HasNext() && row_idx < numElems) {
+        (void)reader->ReadBatch(1, &definition_level, nullptr,
```
jabraham17, line 91: In the above block, you have `reader->ReadBatch` without the cast to void. Why is it needed here?

I assume its to silence unused return warnings? If so, please add the cast to void above

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+        reader->ReadBatch(batchSize, nullptr, nullptr, tmpArr, &values_read);
+
+        // promote to larger type
+        for (int64_t j = 0; j < values_read; j++) {
+          chpl_ptr[row_idx+j] = (ChplType)tmpArr[j];
+        }
+
+        row_idx+=values_read;
+        num_read+=values_read;
+      }
```
jabraham17, line 141: nit formatting

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+    ReaderType* reader = static_cast<ReaderType*>(column_reader.get());
+    *startIdx -= reader->Skip(*startIdx);
+
+    // TODO find better variable names. values_read and num_read are confusing
+    int64_t values_read = 0;
+
+    int64_t num_read = 0;
+    if (nullMode == noNulls) {
+      while (reader->HasNext() && row_idx < numElems) {
+        PqType* tmpArr = (PqType*)malloc(batchSize * sizeof(PqType));
```
jabraham17, line 129: use `new` instead of `malloc`

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+        row_idx+=values_read;
+        num_read+=values_read;
+      }
+    }
+    else {
+      int16_t definition_level; // nullable type and only reading single records in batch
+      while (reader->HasNext() && row_idx < numElems) {
+        PqType value = 0;
+        auto ret = reader->ReadBatch(1, &definition_level, nullptr, &value,
+                                     &values_read);
```
jabraham17, line 148: is `ret` used?

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+    // TODO find better variable names. values_read and num_read are confusing
+    int64_t values_read = 0;
+
+    int64_t num_read = 0;
+    if (nullMode == noNulls) {
+      while (reader->HasNext() && row_idx < numElems) {
+        PqType* tmpArr = (PqType*)malloc(batchSize * sizeof(PqType));
+        if((numElems - row_idx) < batchSize) // adjust batchSize if needed
+          batchSize = numElems - row_idx;
+        reader->ReadBatch(batchSize, nullptr, nullptr, tmpArr, &values_read);
```
jabraham17, line 132: same comment as above about ignoring the return result

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+        PqType value = 0;
+        auto ret = reader->ReadBatch(1, &definition_level, nullptr, &value,
+                                     &values_read);
+        // if values_read is 0, that means that it was a null value
+        if(values_read > 0) {
+          chpl_ptr[row_idx] = (ChplType) value;
+        }
+        else {
+          chpl_ptr[row_idx] = NAN;
+        }
```
jabraham17, line 155: nit, write this as a ternary

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+
+    // TODO find better variable names. values_read and num_read are confusing
+    int64_t values_read = 0;
+
+    int64_t num_read = 0;
+    if (nullMode == noNulls) {
+      while (reader->HasNext() && row_idx < numElems) {
+        if((numElems - row_idx) < batchSize) // adjust batchSize if needed
+          batchSize = numElems - row_idx;
+        reader->ReadBatch(batchSize, nullptr, nullptr, &chpl_ptr[row_idx],
```
jabraham17, line 181: same comment about return results

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+        PqType value = 0;
+        auto ret = reader->ReadBatch(1, &definition_level, nullptr, &value,
+                                     &values_read);
+        // if values_read is 0, that means that it was a null value
+        if(values_read > 0) {
+          chpl_ptr[row_idx] = (ChplType) value;
+        }
+        else {
+          chpl_ptr[row_idx] = NAN;
+        }
```
jabraham17, line 199: same comments as about this block

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+
+    // TODO find better variable names. values_read and num_read are confusing
+    int64_t values_read = 0;
+
+    int64_t num_read = 0;
+    if (not hasNonFloatNulls) {
+      // TODO we can read into the actual Chapel array, and then space the data
+      // out in reverse order. I don't think we need to have this temporary
+      // buffer
+      PqType* tmpArr = (PqType*)malloc(batchSize * sizeof(PqType));
```
jabraham17, line 268: use `new` instead of `malloc`

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+      while (reader->HasNext() && row_idx < numElems) {
+        if((numElems - row_idx) < batchSize) // adjust batchSize if needed
+          batchSize = numElems - row_idx;
+
+        // Can't read directly into chpl_ptr because it is int64
+        reader->ReadBatch(batchSize, nullptr, nullptr, tmpArr, &values_read);
```
jabraham17, line 274: same comments about ignoring the return

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+                                // records in batch
+      while (reader->HasNext() && row_idx < numElems) {
+        reader->ReadBatch(1, &definition_level, nullptr, &tmp, &values_read);
+        // if values_read is 0, that means that it was a null value
+        if(values_read == 0) {
+          where_null_chpl[row_idx] = true;
+        }
+        else {
+          chpl_ptr[row_idx] = (ChplType)tmp;
+        }
```
jabraham17, line 299: prefer ternary

src/parquet/ReadParquet.cpp
```diff
@@ -1,5 +1,407 @@
...
+          nrows_in_iter = nread;
+        }
+        else {
+          // we already now how many rows we are reading per column, which is
+          // stored in nrows_in_iter. But did we read the correct number of rows
+          // in this particular iteration?
+          if (nread != nrows_in_iter) {
+            std::cout << "\tUneven number of rows are read.";
+            std::cout << " Expected " << nrows_in_iter;
+            std::cout << " But read " << nread << " instead\n";
```
jabraham17, line 390: nit, these errors should go to `std::cerr`

src/parquet/ReadParquet.cpp
```diff
@@ -75,13 +478,14 @@ int64_t readColumnIrregularBitWidth(void* chpl_arr, int64_t *startIdx, std::shar
 
   int64_t num_read = 0;
   if (not hasNonFloatNulls) {
-    PqType* tmpArr = (PqType*)malloc(batchSize * sizeof(int32_t));
+    PqType* tmpArr = (PqType*)malloc(batchSize * sizeof(PqType));
```
jabraham17, line 481: same comment about new vs malloc

src/parquet/ReadParquet.cpp
jabraham17: also, this `tmpArr` is never freed. Consider wrapping in a `std::unique_ptr` (and applying that change to the rest of the code)

src/parquet/UtilParquet.cpp
```diff
@@ -94,6 +122,67 @@ int cpp_getPrecision(const char* filename, const char* colname, char** errMsg) {
...
+    *errMsg = strdup(e.what());
+    return ARROWERROR;
+  }
+
+  const int num_cols = sc->num_fields();
+  for(int i=0 ; i<num_cols ; i++) {
+    int ret = getTypeOfFieldAtIdx(sc, i);
+    if (ret == ARROWERROR) {
+      std::string msg = "Unsupported type";
+      *errMsg = strdup(msg.c_str());
```
jabraham17, line 176: strdup is unnecessary, `.c_str` returns new memory

jabraham17 (approved):

src/ParquetMsg.chpl
```diff
@@ -871,6 +925,520 @@ module ParquetMsg {
...
+      for (rawEntry, t, colIdx, colName) in zip(Entries, op.ty,
+                                                op.colDom, datasets) {
+        var entry = op.postProcess(rawEntry, t, colIdx);
+
+        var valName = st.nextName();
+        try {
+          // can't throw from a non-inlined iterator, I'll handle this myself
+          st.addEntry(valName, entry);
+        }
+        catch e {
```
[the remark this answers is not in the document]
e-kayrakli: Ah, I am upset that you are right. The module uses that style, so I concede. I am curious about your style preference in general though. I am a strong supporter of

```
}
else {
```

So,

```
}
catch {
```

is what I prefer and what my muscle reflex dictates.

src/ParquetMsg.chpl
```diff
@@ -871,6 +925,520 @@ module ParquetMsg {
...
+          when ARROWINT32   do entry = createSymEntry(this.len, int(64));
+          when ARROWUINT64  do entry = createSymEntry(this.len, uint(64));
+          when ARROWUINT32  do entry = createSymEntry(this.len, uint(64));
+          when ARROWDOUBLE  do entry = createSymEntry(this.len, real(64));
+          when ARROWFLOAT   do entry = createSymEntry(this.len, real(64));
+          when ARROWBOOLEAN do entry = createSymEntry(this.len, bool);
+          when ARROWDECIMAL do entry = createSymEntry(this.len, real(64));
+          otherwise do
+            throw new NotImplementedError("Unexpected column type while " +
+                                          "reading parquet file", L(), R(), M());
```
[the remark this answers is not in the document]
e-kayrakli: Good catch.

This whole duplication is unfortunate. I don't think we should have a specialization for a single-col read here. It doesn't give us a ton of benefit IMO. It is something I want to address, which will make this redundancy moot, in any case.

That being said, you are right. I wouldn't want us to wait for that effort to complete. I addressed this in a local commit.

src/ParquetMsg.chpl
```diff
@@ -871,6 +925,520 @@ module ParquetMsg {
...
+        when ARROWDECIMAL do return (e:(SE(real(64), 1))).a.domain;
+        otherwise do
+          throw new NotImplementedError("Unexpected column type while " +
+                                        "reading parquet file", L(), R(), M());
+      }
+    }
+
+    inline proc getDomain(e: [] GenSymEntry?): domain(?) throws {
+      // just assume that they all have the same domain
+      return getDomainImpl(e.first!, this.ty.first);
```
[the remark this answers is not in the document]
e-kayrakli: Good catch.

src/ParquetMsg.chpl
e-kayrakli: Can you elaborate why you think the current implementation would lead to more generics?

src/parquet/ReadParquet.cpp
e-kayrakli: That's a really nice nit. Fixed.

src/parquet/ReadParquet.cpp
e-kayrakli: I will postpone this to the next refactor. It is against my style preferences, too. But if I am going in that direction, I would want to change variable names first.

src/parquet/ReadParquet.cpp
e-kayrakli: Jade taught me about `std::ignore`. So, I used that in cases like this where I introduced this unused return values.

src/parquet/ReadParquet.cpp
e-kayrakli: I added removing this allocation to the list of followups.

src/parquet/ReadParquet.cpp
e-kayrakli: Considering leaving this as-is as per my previous comment.

src/parquet/ReadParquet.cpp
e-kayrakli: Leaving `malloc` as I want to remove this allocation altogether

src/parquet/ReadParquet.cpp
e-kayrakli: This is actually the snippet I was referring to above as something that's not very suitable for ternary. Note that the two branches assign to different things. I can make use of ternary operator here, but that would feel like an abuse of it. As, I think this should be kept as is, I prefer to keep other similar post-processing related conditionals the same for the time being.

src/parquet/ReadParquet.cpp
e-kayrakli: Good catch. We populate errMsg based on the exceptions's message at the end of the function. I just decided to throw `std::exception`s here and above to be captured at the end. I think we can consider tighter error handling schemes on the C++ side in the future. I added a TODO comment.

src/parquet/ReadParquet.cpp
e-kayrakli: Choosing to leave for future.

src/parquet/UtilParquet.cpp
e-kayrakli: This is inherited code, so I didn't write anything specific knowingly here, but https://en.cppreference.com/w/cpp/string/basic_string/c_str seems to suggest that it does _not_ return new memory.

src/parquet/SharedEnums.h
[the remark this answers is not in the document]
e-kayrakli: I haven't looked into yet, but the idea was appealing. I am planning to attempt that before my next test run.

e-kayrakli:
Thanks, @ajpotts and @jabraham17 

@ajpotts / @jabraham17 : I will revert the configuration files. I am keeping them as is to make my life easier testing the branch on different systems.

@jabraham17 : I resolved bunch of your comments with a new set of commits. Feel free to iterate through those to follow changes. Probably [the first new commit](https://github.com/Bears-R-Us/arkouda/pull/4906/commits/ead5a27a3c90e2db0e27ae641cd35882b1f6dbee) is the most consequential one.

I left some discussions unresolved because I either need some more clarifications, or prefer to do the changes you request in a subsequent PR out of paranoia + principle to not make nitty-gritty stylistic changes in this PR where big chunks of code being moved around and introduced. Let me know if you want to argue for urgency of those changes.

One big item for me to try tomorrow is your suggestion to replace the symlink-based enum interop with... something more principled (I need to remember the idea by going through our Slack convo). But feel free to chime in on the unresolved discussions in the meantime.

e-kayrakli:
I will do my testing after rebasing on `main`.

src/ParquetMsg.chpl
```diff
@@ -29,7 +30,9 @@ module ParquetMsg {
 
 
   // Use reflection for error information
-  use Reflection;
+  import Reflection.{getModuleName as M,
+                     getRoutineName as R,
+                     getLineNumber as L};
```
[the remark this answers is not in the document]
jabraham17: I think `gMN`, `gRN`, and `gLN` would be a improvement over the single letter idents while maintaining the short calls. I would also propose `getM`, `getR`, and `getL`

jabraham17:
> I left some discussions unresolved because I either need some more clarifications, or prefer to do the changes you request in a subsequent PR out of paranoia + principle to not make nitty-gritty stylistic changes in this PR where big chunks of code being moved around and introduced. Let me know if you want to argue for urgency of those changes.

I think many of my comments were about improving the quality of the code, both for developer readability and maintainability/future extensibility. I am perfectly happy with those comments being addressed in a future PR

src/parquet/UtilParquet.cpp
e-kayrakli: I started putting together a `ParquetCall` context manager type where calls to C layer can be made within a `manage` statement and error is handling while exiting the context. That has sprawled way too much and I reverted back. I think we need to contain calls to C, error message/`ARROWERROR` propagation to a well-defined piece of code rather than sprinkling the Chapel module with it. I just updated the TODO list in the OP to include that.

registration-config.json
```diff
@@ -7,8 +7,7 @@
                 "uint",
                 "uint(8)",
                 "real",
-                "bool",
-                "bigint"
```
ajpotts, line 11: Please revert this change before merging.

tests/io_test.py
```diff
@@ -307,7 +307,10 @@ def test_edge_case_read_write(self, par_test_base_tmp, dtype, comp):
                 assert (np_edge_case == pq_arr.to_ndarray()).all()
 
     @pytest.mark.parametrize("prob_size", pytest.prob_size)
-    def test_large_parquet_io(self, par_test_base_tmp, prob_size):
+    @pytest.mark.parametrize("use_str", [True, False])
+    @pytest.mark.parametrize("null_handling", ["none", "only floats", "all"])
+    def test_large_parquet_io(self, par_test_base_tmp, prob_size, use_str,
```
ajpotts, line 312: I'm not seeing any tests that read arrays with actual null values.  Is there a plan to add some?

src/ParquetMsg.chpl
```diff
@@ -871,6 +925,509 @@ module ParquetMsg {
...
+    var Filenames = processFilelist(filelist);
+
+    
+    var fileErrors: list(string);
+    var fileErrorCount:int = 0;
+    var fileErrorMsg:string = "";
+    var Sizes: [Filenames.domain] int;
+    var Offsets: [Filenames.domain] int;
+    
+    var rnames: list((string, ObjType, string)); // tuple (dsetName, item type, id)
```
ajpotts, line 983: Can this comment be deleted?

src/ParquetMsg.chpl
```diff
@@ -871,6 +925,509 @@ module ParquetMsg {
     }
   }
 
+  proc readAllColsParquetMsg(cmd: string, msgArgs: borrowed MessageArgs,
+                             st: borrowed SymTab): MsgTuple throws {
+    var repMsg: string;
+    var tagData: bool = msgArgs.get("tag_data").getBoolValue();
+    var strictTypes: bool = msgArgs.get("strict_types").getBoolValue();
```
ajpotts, line 932: `strictTypes` is never used.

src/ParquetMsg.chpl
```diff
@@ -871,6 +925,509 @@ module ParquetMsg {
...
+      if hadError {
+        // Keep running total, but we'll only report back the first 10
+        if fileErrorCount < 10 {
+          fileErrors.pushBack(fileErrorMsg.replace("\n", " ").replace("\r", " ").replace("\t", " ").strip());
+        }
+        fileErrorCount += 1;
+      }
+    }
+
+    const DataDom = makeDistDom(len);
```
ajpotts, line 1017: It doesn't seem like `DataDom` is used in this function?

ajpotts (approved):

tests/io_test.py
e-kayrakli: https://github.com/Bears-R-Us/arkouda/blob/main/tests/pandas/io_test.py#L441-L471 these tests use `np.nan`, which ends up taking the null-value path. They look pretty similar to what you have. So, I think we are already covered?

src/ParquetMsg.chpl
e-kayrakli: Good catch. It is something I inherited from the earlier iteration. I created https://github.com/Bears-R-Us/arkouda/issues/4925 to clean it up as a follow up.

src/parquet/ReadParquet.h
```diff
@@ -1,6 +1,12 @@
+#ifndef READ_PARQUET_H
+#define READ_PARQUET_H
+
 #include <stdint.h>
 #include <stdbool.h>
```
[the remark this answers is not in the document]
e-kayrakli: These errors were due to g++ 11 not supporting C++17's in-class template specialization. https://github.com/Bears-R-Us/arkouda/pull/4906/commits/907f25de2ab0658cfd52ebe4c71689c2c631ce59 fixed it.

tests/io_test.py
ajpotts: They don't cover all the `dtypes` or your new code paths that you're adding....

src/parquet/ReadParquet.h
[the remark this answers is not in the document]
e-kayrakli: @ajpotts could you give this another try?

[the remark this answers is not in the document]
ajpotts: This works now.  Thanks!

PR #4910, opened 2025-09-16
Fix paths in new CI image Dockerfile

jabraham17:
Fix paths in new CI image Dockerfile to actually add `chpl` to the PATH

Future work
- [ ] all of the work done to put stuff in a `~/.bashrc.local` is not being used at all. the CI runs with `/bin/sh -c ...`, so bash files will never be invoked

### DIFF
--- a/docker/ubuntu-with-arkouda-deps/Dockerfile
+++ b/docker/ubuntu-with-arkouda-deps/Dockerfile
@@ -4,7 +4,7 @@ ARG CHPL_VERSION
 
 USER root
 
-ENV PATH=/venv/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/venv/bin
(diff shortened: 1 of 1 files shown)


docker/ubuntu-with-arkouda-deps/Dockerfile
```diff
@@ -98,6 +98,7 @@ RUN rm -fr arkouda
 ###############################################################################
 
 #   Save build environment to the .bashrc.local
+# TODO: is this really needed?
```
ajpotts, line 101: We use this for debugging, and so that we can build the container for testing.  I use it a quite a bit for testing actually.

drculhane (approved):
Looks straightforward.

jaketrookman (approved):
Looks good

PR #4914, opened 2025-09-16
Closes #4565: MaxArrayDims is incorrect - fix or remove it

### LINKED ISSUE #4565: `MaxArrayDims` is incorrect - fix or remove it
### The Bug
The config param `MaxArrayDims` can easily be inconsistent with the set of functions cloned by the registration framework. For example, `registration-config.json` may request array dimensions 1, 2, 3 and `MaxArrayDims` may still have the default of 1.

Furthermore, `MaxArrayDims` cannot describe the situation where `registration-config.json` requests array dimensions 1 and 3.

Currently `MaxArrayDims` is used only for error reporting. It also appears in the output of a "get configuration" command. So removing it may cause issues with backward compatibility.

### Alternative Solution

Instead of removing it, we could connect it to the registration framework and require that it clones the functions for all consecutive array ranks.

1RyanK:
Decided to remove it. It doesn't look like it's actually used anywhere and in `ServerConfig.chpl` there already is a function `proc arrayDimIsSupported(param dim: int) param : bool {` which should continue to be used in place of `MaxArrayDims`.

Closes #4565: MaxArrayDims is incorrect - fix or remove it

### DIFF
--- a/src/ServerConfig.chpl
+++ b/src/ServerConfig.chpl
@@ -41,12 +41,6 @@ module ServerConfig
     };
 
     /*
-      maximum array dimensionality supported by the server
-      set in 'registration-config.json'
-    */
-    config param MaxArrayDims: int = 1;
-
-    /*
     Type of deployment, which currently is either STANDARD, meaning
     that Arkouda is deployed bare-metal or within an HPC environment, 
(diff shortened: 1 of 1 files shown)


ajpotts (approved):
Thanks for taking care of this one!

Issue #4835, opened 2025-08-28
Message registration: Same module appearing twice in the configuration file results in two copies to be stamped out


e-kayrakli:
I came across this while using `--saveUsedModules`, which happens to save modules that are core and not normally have to be included in `ServerModules.cfg`. That resulted in `GenSymIO` to appear both in my config file, and also a list somewhere in the registration infra where we list default modules. The infra was happy to go through `GenSymIO` twice, stamping out same functions twice quitely.

End result was a Chapel compiler error because of doubly-defined functions.

1RyanK:
I started a server with `--saveUsedModules` and ran the following:

```python
In [1]: a = ak.arange(5)

In [2]: a * 2
Out[2]: array([0 2 4 6 8])

In [3]: ak.tile(a, 2)
Out[3]: array([0 1 2 3 4 0 1 2 3 4])

In [4]: ak.shutdown()
```

Now my `UsedModules.cfg` looks like this:

```
(arkouda-dev) ryan@vbox:~/projects/arkouda\ $ cat UsedModules.cfg 
SequenceMsg
ManipulationMsg
OperatorMsg
```

I ran a `make` in that directory and no compiler error occurred. Did this issue somehow get fixed since you posted it? Or can you tell me how to better reproduce this?

e-kayrakli:
Not focusing on `--saveUsedModules`:

I can't compile Arkouda built with `MyModules.cfg`:

```
ApplyMsg
ApplyMsg
```

with error message

```
/home/engin/code/arkouda-clones/arkouda/src/registry/Commands.chpl:66: In module 'Commands':
/home/engin/code/arkouda-clones/arkouda/src/registry/Commands.chpl:83: error: cannot capture routine with name 'ark_reg_isPythonModuleSupported' because it is overloaded
/home/engin/code/arkouda-clones/arkouda/src/registry/Commands.chpl:76: note: declared here
/home/engin/code/arkouda-clones/arkouda/src/registry/Commands.chpl:465: note: declared here
```

That shows the redefinition.

The root issue here is that if you use the same module name twice in the module config file, and if that module uses attribute-based message registration, the framework stamps out functions twice.

I think this should be ignored, ideally with a warning issued just to make sure that the user is aware of it.

1RyanK:
I was also missing the switch to using a different modules file. That makes sense.

PR #4920, opened 2025-09-17
Closes #4836: --saveUsedModules should only save modules that are not compiled by default

### LINKED ISSUE #4836: `--saveUsedModules` should only save modules that are not compiled by default
Otherwise the same module is included by the registration framework twice causing errors (https://github.com/Bears-R-Us/arkouda/issues/4835).

Even if that issue is resolved, I think it'd be good for that flag to not save modules that are deemed "standard".

1RyanK:
In `CommandMap.chpl` modules are added rather indiscriminately to `usedModules`. The issue arises in that in `src/registry/register_commands.py`, this gets defined: `DEFAULT_MODS = ["MsgProcessing", "GenSymIO"]`. It seems that these automatically come along for the ride and don't need to be included in `UsedModules.cfg`.

Closes #4836: --saveUsedModules should only save modules that are not compiled by default

### DIFF
--- a/src/CommandMap.chpl
+++ b/src/CommandMap.chpl
@@ -23,6 +23,7 @@ module CommandMap {
   var moduleMap: map(string, (string, int));  // cmd-name => (module-name, line number)
   use Set;
   var usedModules: set(string);
+  const requiredModules = new set(string, ["GenSymIO", "MsgProcessing"]);
 
   /**
    * Register command->function in the CommandMap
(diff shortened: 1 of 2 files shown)


ajpotts:
I think this also closes #4835.

ajpotts (approved):
Thanks for fixing this!

jaketrookman (approved):
Looks good

PR #4922, opened 2025-09-18
Closes #4921: fabs

### LINKED ISSUE #4921: fabs
**Feature Description**
See the description of the function [here](https://numpy.org/doc/stable/reference/generated/numpy.fabs.html)

**Example Behavior**
```python
In [2]: a = np.array([-1, 0, 1])

In [3]: np.fabs(a)
Out[3]: array([1., 0., 1.])
```

**Additional context**
Essentially it is float absolute value. Be sure that testing like this works:
```python
In [5]: x = np.array([-1, -(2**63), 1])

In [6]: np.fabs(x)
Out[6]: array([1.00000000e+00, 9.22337204e+18, 1.00000000e+00])
```
I think it would be good to cast to float first, and then do absolute value. The other way around probably will fail on the above.

1RyanK:
Adds a floating point absolute value function.

Closes #4921: fabs

### DIFF
--- a/arkouda/__init__.py
+++ b/arkouda/__init__.py
@@ -171,6 +171,7 @@ from arkouda.numpy import (
     exp,
     expm1,
     eye,
+    fabs,
     finfo,
     flexible,
     flip,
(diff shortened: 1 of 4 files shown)


ajpotts (approved):
Looks great.  Thanks!

jaketrookman (approved):
Looks good

PR #4939, opened 2025-09-23
Closes #4157:  server can return a negative value when it should retu…

### LINKED ISSUE #4157: server can return a negative uint
Sometimes the server can return a negative `uint`.

ajpotts:
Fixes #4157 which was traced back to a string formatting error in `src/NumPyDType.chpl`.

Closes #4157:  server can return a negative value when it should return uint


1RyanK (approved):
Looks good!

jaketrookman (approved):
Looks good

PR #4952, opened 2025-09-25
Make the loop iterating over multiple parquet files parallel

e-kayrakli:
While working on https://github.com/Bears-R-Us/arkouda/pull/4906, I converted a parallel loop in to a serial one for better debugging, but then forgot to make it parallel again. This was observed as a performance regression in a use case. This PR adds the `forall` loop back instead of the `for` loop. Note that a similar part of code, that's used for col-by-col reads [already uses a parallel loop](https://github.com/Bears-R-Us/arkouda/blob/main/src/ParquetMsg.chpl#L1248-L1250).

I spot-checked the correctness.


drculhane:
This looks like it should be a no-brainer, but when I pulled it and ran tests, the test_poisson_seed_reproducibility test (which does use parquet) failed.

ajpotts (approved):

bradcray:
Closing this in favor of #5486 which the amazing Jade put together today.  Thanks Jade!!!

PR #5002, opened 2025-10-15
Closes #4227: Hitting Chapel's default instantiation limit when compiling for 3 or more dimensions

### LINKED ISSUE #4227: Hitting Chapel's default instantiation limit when compiling for 3 or more dimensions
**Describe the bug**
When compiling arkouda to support 1-3D arrays by modifying registration config, we run into an error with exceeding Chapel's instantiation limit.

**To Reproduce**
1. Set the number of dimensions to `"nd": [1,2,3],` in `registration-config.json`
2. `make`
3. See error

**Expected behavior**
Arkouda should compile without error

**Error Message**
```
             createTaskFunctions :   0.090 seconds
$CHPL_HOME/modules/dists/DSIUtil.chpl:485: error: Function 'chpl_allStridesArePositive' has been instantiated too many times
note:   If this is intentional, try increasing the instantiation limit from 512
make: *** [Makefile:567: arkouda_server] Error 1
```

**Is this a Blocking Issue** No
- Are there any known work-arounds?
   - Yes, increasing the instantiation limit is a good workaround
- If a known work around exists, please provide a code block demonstrating it.
   - `export CHPL_INSTANTIATION_LIMIT=1024`

1RyanK:
This allows users to override Chapel's instantiation limit from the make line.

As it happens, this is not easy to test that it "fixes" anything right now. I think recent PRs have done an outstanding job of fixing this issue from the perspective of not needing such a high instantiation limit (and, I might add, the memory issue). On my laptop with about 30 GB RAM free, I was able to set

```yaml
# Snippet from registration_config.json
"nd": [1, 2, 3, 4, 5],
```

(yes, five dimensions) and while it took a very long time to compile (around 40 minutes) it managed to compile with the default instantiation limit of 512. So how do we know that this works? It's actually not that hard to check that it breaks things. You can run

```bash
make instantiate_max=10
```

and you will see

```bash
$CHPL_HOME/modules/standard/CTypes.chpl:611: error: Function ':' has been instantiated too many times
note:   If this is intentional, try increasing the instantiation limit from 10
make: *** [Makefile:575: arkouda_server] Error 1
```

So clearly this code has an effect! That really ought to mean that if a user wants a higher instantiation limit, they should be able to get it.

Oh, and I updated `BUILD.md` so anyone who greps through our files for the error should be able to find documentation on how to fix it!

Closes #4227: Hitting Chapel's default instantiation limit when compiling for 3 or more dimensions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -371,6 +371,12 @@ CHPL_FLAGS += -lpython$(shell python3 -c "import sysconfig; print(sysconfig.get_
 # Ignore warnings from the Python headers. This is irrelevant for newer Python versions
 CHPL_FLAGS += --ccflags -Wno-macro-redefined
 
+# Allows the user to set a different instantiation limit
+ifneq ($(instantiate_max),)
+CHPL_USER_FLAGS += --instantiate-max=$(instantiate_max)
+endif
+
+
 PYTHON_VERSION := $(shell python3 -c "import sys; print(*sys.version_info[:2], sep='.')")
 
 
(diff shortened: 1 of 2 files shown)


jaketrookman (approved):
Looks good

PR #5005, opened 2025-10-16
Closes #4500: Create a repartitionByHash function

### LINKED ISSUE #4500: Create a `repartitionByHash` function
**Feature Description**
Like the `repartitionByHashString` function, but for other data types.

**Example Behavior**
Should hash the data it receives to determine a destination locale, and then use `repartitionbyLocale` function to send data where it belongs.

1RyanK:
Since deciding to switch over to the `innerArray` framework, I figured I would do this as an `innerArray` version. I also did the same for the `Strings` version.

I'm thinking I might change the way these work at some point. It'd be nice if I could do the hashing inside the first coforall of the function it moves on to. I've been thinking about how to make that work and I think it would maybe be a little tricky. But I think I might be able to do it with defining classes or records or something. And then that would allow users to define their own repartitioning schemes as well.

Closes #4500: Create a repartitionByHash function

### DIFF
--- a/src/Repartition.chpl
+++ b/src/Repartition.chpl
@@ -50,6 +50,52 @@ module Repartition
     return repartitionByLocaleString(destLocales, strOffsets, strBytes);
   }
 
+  // Note, the arrays passed here must have PrivateSpace domains.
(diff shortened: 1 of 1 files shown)


ajpotts (approved):
Looks good.

PR #5014, opened 2025-10-21
Closes 4737 - renames the registered name broadcast in BroadcastMsg.chpl to gbbroadcast

### LINKED ISSUE #4737: Multiple `broadcast` messages
Right now, there are two broadcast messages being registered for different purposes. Both of the client functions sending a `broadcast` message to the server are seemingly tested. I am not entirely sure how this works, but I think their names should be different.

These functions are being registered in `BroadcastMsg` and `ManipulationMsg` modules

drculhane:
Closes #4737, which noted that two different chapel functions were both being registered with the name "broadcast".

This changes the one associated with groupby to gbbroadcast.

### DIFF
--- a/ServerModules.cfg
+++ b/ServerModules.cfg
@@ -6,7 +6,6 @@ ArgSortMsg
 ArraySetopsMsg
 AryUtil
 BigIntMsg
-BroadcastMsg
 CastMsg
 CheckpointMsg
 CommDiagnosticsMsg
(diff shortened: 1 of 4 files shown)


src/BroadcastMsg.chpl
```diff
@@ -138,5 +138,5 @@ module BroadcastMsg {
   }
 
   use CommandMap;
-  registerFunction("broadcast", broadcastMsg, getModuleName());
+  registerFunction("gbbroadcast", broadcastMsg, getModuleName());
```
e-kayrakli: Could you consider changing `broadcastMsg` to `gbBroadcastMsg` for consistency?

ajpotts (approved):
Looks good!  Thanks for fixing this.

jaketrookman (approved):

PR #5037, opened 2025-11-05
Fix sort benchmark names in benchmark v2

e-kayrakli:
Probably due to a copy/paste oversight, all sort benchmarks were being reported as random_uniform. This PR adjusts them to have correct names.


ajpotts (approved):
Thanks for fixing this!

PR #5039, opened 2025-11-05
Use the Chapel standard sort for distributed sorting

e-kayrakli:
As of https://github.com/chapel-lang/chapel/pull/27277, Chapel's standard library distributed sort is faster than the undocumented one that Arkouda uses. This PR makes the switch to the new sort. In anecdotal runs, I see it to be much faster than the existing ones.


ajpotts (approved):
Good idea!

PR #5044, opened 2025-11-10
Closes #5043: Investigate performance loss from negative bigint changes

### LINKED ISSUE #5043: Investigate performance loss from negative `bigint` changes
It was noted that the following code caused performance loss:

```python
        neg_mask = array((a < 0) * -2 + 1)
        abs_a = array(abs(a), dtype, max_bits=max_bits)
        assert isinstance(abs_a, pdarray)  # This is for mypy reasons
        return neg_mask * abs_a
```
(from [here](https://github.com/Bears-R-Us/arkouda/pull/4985/files#diff-27f7215a2f35584282fb3e9598234b9d255b87a5f94eb222b22f56d77f72f751R305-R308))

It would be good to investigate if there's any better way to do this, or at least offset bad performance

1RyanK:
While PR #4985 technically fixed the issue of general negative bigint problems, it caused a performance regression. I believe this is primarily due to the overhead of creating the sign array. Instead, here's basically what happens in the code:

```python
    any_neg = np.any(flat < 0)
    req_bits: int
    if any_neg:
        req_bits = max(flat.max().bit_length(), (-flat.min()).bit_length()) + 1
    else:
        req_bits = flat.max().bit_length()
```
Then the code figures out how many times to pull off `uint64` limbs, rather than waiting until everything is zero (which doesn't happen in the negative case, it just continues to be -1).

The code runs out to ~~two~~ three separate Chapel functions depending on the case. If the input is just an `int64` or `uint64` array, it converts it directly to bigint (similar for `float64` or some other kind of floating point input). If the input is numpy's version of a bigint array, it goes to the multi-limb version (unfortunately, this is the case even if everything comes out to just one limb, but I think the performance loss here is not that bad). If the input data had any negative values, it treats the limbs as signed (all bits are positive and the top bit is negative). However, it's hard to create a bigint like this (AFAIK). So Chapel-side, it creates a signs array, strips off the top bit of every limb and treats it as a bool to reference later. Either way, it goes to the "Horner fold" step, which, as ChatGPT tells me, is possibly a faster way to create the bigints. Previously the code was bit shifting the limbs into the right spot and then adding them to the bigint value. The idea here is that you start with the highest limb of the data, then you bitshift it and add in the next limb, bitshift what you have and add in the next limb, and so on. You can read more about the generic version of this [here](https://en.wikipedia.org/wiki/Horner%27s_method) (take x = `2**64`). Then it adds in the signed bit as necessary.

It also handles the case where `max_bits` is not -1.

Hopefully any loss in performance is made up by a few factors:

1. Previously the Python code was stripping the limbs off by modding by `2**64` and then integer dividing by the same value. I think it could speed the code up to do a bitmask by `2**64 - 1` and then bitshift by 64.
2. Supposedly the Horner fold has better performance.
3. If the input is only one limb of `int64` or `uint64` data, it should go to the single limb version and that should run quicker.

~~This also handles all cases of numeric data to bigint output in a single function, so if the performance is back up, then I can cut out some code in the array function.~~

I went ahead and cut the old bigint code out.

Closes #5043: Investigate performance loss from negative bigint changes

### DIFF
--- a/arkouda/numpy/pdarraycreation.py
+++ b/arkouda/numpy/pdarraycreation.py
@@ -408,7 +408,7 @@ def array(
         except TypeError:
             raise RuntimeError(f"Unhandled dtype {a.dtype}")
 
-    if a.ndim != 1 and a.dtype.name not in NumericDTypes:
+    if a.ndim == 0 and a.dtype.name not in NumericDTypes:
         raise TypeError("Must be an iterable or have a numeric DType")
 
     # Return multi-dimensional pdarray if a.ndim in get_array_ranks()
@@ -416,6 +416,16 @@ def array(
 
     if a.ndim not in get_array_ranks():
         raise ValueError(f"array rank {a.ndim} not in compiled ranks {get_array_ranks()}")
+
+    if a.dtype == object and all(isinstance(x, (int, np.integer)) for x in a.ravel()) and dtype is None:
+        dtype = bigint
+    if a.dtype.kind in ("i", "u", "f", "O") and (dtype == bigint or dtype == "bigint"):
+        return _bigint_from_numpy(a, max_bits)
+    elif not (
+        np.issubdtype(a.dtype, np.str_)
+        or (a.dtype == np.object_ and a.size > 0 and isinstance(a[0], str))
+    ) and (dtype == bigint or dtype == "bigint"):
+        raise TypeError("Cannot convert type to bigint")
 
     # Check if array of strings
     # if a.dtype == numpy.object_ need to check first element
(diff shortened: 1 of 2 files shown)


arkouda/numpy/pdarraycreation.py
```diff
@@ -293,6 +293,8 @@ def array(
 
     if a.dtype == object and all(isinstance(x, (int, np.integer)) for x in a) and dtype is None:
         dtype = bigint
+    if a.dtype.kind in ("i", "u", "O") and (dtype == bigint or dtype == "bigint"):
+        return _bigint_from_numpy(a, max_bits)
```
ajpotts: Please subtract the old logic that is no longer used.

1RyanK: Done!

Issue #3945, opened 2024-12-13
`readcsv` does not handle quoted strings correctly


ShreyasKhandekar:
**Describe the bug**
CSV files can have quoted strings. This is usually used to escape the delimiter or other weird characters. Moreover, CSV also allows you to use quotes in your headers or values which can be escaped using a preceding quote (`""`)

I am referring to the standard laid out in RFC 4180: https://www.ietf.org/rfc/rfc4180.txt

**To Reproduce**
1. Have a small CSV file which has quotes in the header and/or in the values:
Ex:
```csv
"humidity","light","lpg","motion","smoke","temp","height (in inches)"
"51.0","false","0.00765082227055719","false","0.02041127012241292","22.7","14"""
"76.0","false","0.005114383400977071","false","0.013274836704851536","19.700000762939453","16"""
"50.9","false","0.007673227406398091","false","0.02047512557617824","22.6","25"""
```

Notice how the `height (in inches)` is expressed as `"14"""` where the `""` is used to represent an escaped quote so the actual value being represented in `14"` where the `"` stands for `inches`

2. Read it in in arkouda:
```python3
>>> import arkouda as ak
```

```python3
>>> ak.read_csv("dumdum.csv")
{'\\"humidity\\"': array(['"51.0"', '"76.0"', '"50.9"']), '\\"light\\"': array(['"false"', '"false"', '"false"']), '\\"lpg\\"': array(['"0.00765082227055719"', '"0.005114383400977071"', '"0.007673227406398091"']), '\\"motion\\"': array(['"false"', '"false"', '"false"']), '\\"smoke\\"': array(['"0.02041127012241292"', '"0.013274836704851536"', '"0.02047512557617824"']), '\\"temp\\"': array(['"22.7"', '"19.700000762939453"', '"22.6"']), '\\"height (in inches)\\"': array(['"14"""', '"16"""', '"25"""'])}
```
Notice the weird escape patterns in the header, and the unnecessary quotes being included in the values

Or see how it looks as a `DataFrame`

```python3
>>> ak.DataFrame.read_csv("dumdum.csv")
  \"humidity\" \"light\"  ...              \"temp\" \"height (in inches)\"
0       "51.0"   "false"  ...                "22.7"                 "14"""
1       "76.0"   "false"  ...  "19.700000762939453"                 "16"""
2       "50.9"   "false"  ...                "22.6"                 "25""" (3 rows x 7 columns)
```

**Expected behavior**
[...]

PR #5080, opened 2025-11-21
Fix CSV Parsing of Quoted Fields, Escaped Quotes, Embedded Delimiters, and Multi-Line Records

ShreyasKhandekar:
## Brief

This PR introduces a substantial improvement to Arkouda’s CSV parsing logic, bringing it closer to full [RFC 4180](https://www.ietf.org/rfc/rfc4180.txt) compatibility. Specifically, this PR fixes incorrect handling of:

- column delimiters inside quoted strings,
- escaped quotes (`""`) inside fields and headers,
- multi-line CSV records inside quoted fields, and
- quoted or escaped column headers that become mis-escaped when serialized through JSON.

These changes address multiple long-standing issues in Arkouda’s CSV reader and enable correct behavior for real-world CSV files.

Fixes:
- https://github.com/Bears-R-Us/arkouda/issues/3945
- https://github.com/Bears-R-Us/arkouda/issues/5081
- https://github.com/Bears-R-Us/arkouda/issues/5082

### 1. Correct handling of delimiters inside quoted strings
Arkouda previously treated any delimiter (e.g., `,`) as a field separator—even when it appeared inside a quoted field. This PR rewrites the low-level delimiter scanning logic so that:
- when inside a quoted field, delimiters are treated as literal characters,
- field boundaries are only recognized when the closing quote is reached.
This ensures correct behavior for fields such as:

```
"Hello, World!"
"Nested, commas, inside"
```

### 2. Proper parsing of escaped quotes (`""`)
RFC 4180 requires that "" inside a quoted field be interpreted as a literal ".
This PR adds correct recognition of escaped quotes and rewrites them appropriately during parsing.

For example:

```
"Quote ""Text"""
```

now parses to:

```
Quote "Text"
```

### 3. Support for multi-line records inside quoted fields

Arkouda previously treated physical newlines as record terminators regardless of quoting state. This PR:
- tracks quoting state across physical lines,
- continues reading until the closing quote is found,
- treats embedded newlines as part of the field value.

> [!IMPORTANT]
> https://github.com/Bears-R-Us/arkouda/issues/5083 is still left unresolved by this PR.

Example now works correctly:

```
"id","comment"
1,"Hello
World"
2,"Another
multi-line
comment"
```

## Tests Added / Updated

This PR adds several targeted tests to cover CSV edge cases that were previously unsupported.
- Escaped quotes and delimiters in header fields
- Delimiters inside quoted fields
- Multi-line CSV records
- Quoted values containing escaped quotes
- An `xfail` test for https://github.com/Bears-R-Us/arkouda/issues/5083

## Remaining Work / Future Improvements
While this PR significantly improves compliance and correctness, future enhancements may include:
- Streaming CSV parsing for extremely large multi-line quoted fields.
- A full RFC 4180-compliance validation mode.
- Optional type inference similar to Pandas (e.g., auto-convert true/false, numeric fields, etc.).
- A full revamp of the CSV reader because we still read each files once per column which is terrible for performance
- Adding parallel IO features to CSV readers for even faster reads

# Performance

The performance is mixed bag unfortunately. As mentioned in the issues related to this PR, there is a challenge to maintain correctness without sacrificing performance. Advancing through the column delimiter in Chapel is fast and efficient, but when we try to take into account maintain the "state" of being in quotes and managing them to ignore column delimiters makes this slower.
This becomes even slower when escaped quotes can exist which require marking and reverting the file pointer to manage reads.

I have added many "fast paths" where if we can determine that there are no quotes, or no multi line fields, we can be less careful and avoid doing some expensive operations. In one of the steps we also have the option to completely revert to the old implementation if we can determine that there are no quotes at all in the entire CSV file.

All the performance numbers here are collected by running:

```
❯ python3 -m pytest -c benchmark.ini benchmark_v2/csv_io_benchmark.py --size=1000000 
```

I have removed the write performance from the output for brevity

### 1. Here is what the performance looks like for the implementation in the PR:

```
----------------------------------------------------------------------------------- benchmark 'CSV_IO': 8 tests -----------------------------------------------------------------------------------
Name (time in ms)                    Min                 Max                Mean            StdDev              Median               IQR            Outliers      OPS            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_csv_io[read-uint64]       188.0938 (3.60)     189.4901 (3.37)     188.8971 (3.49)     0.5729 (1.0)      188.9925 (3.52)     0.9302 (1.15)          2;0   5.2939 (0.29)          5           1
bench_csv_io[read-int64]        188.1897 (3.60)     190.8679 (3.39)     190.0001 (3.51)     1.0423 (1.82)     190.3515 (3.55)     0.8081 (1.0)           1;1   5.2632 (0.28)          5           1
bench_csv_io[read-float64]      188.4753 (3.61)     190.0816 (3.38)     189.1288 (3.49)     0.7034 (1.23)     189.0157 (3.52)     1.2325 (1.53)          1;0   5.2874 (0.29)          5           1
bench_csv_io[read-str]          963.9651 (18.45)    970.8573 (17.27)    967.8028 (17.88)    3.1459 (5.49)     969.4045 (18.07)    5.4923 (6.80)          1;0   1.0333 (0.06)          5           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

<details> 

<summary> 
<h3>2.  What happens if you remove the "revert back to old implementation" fast path"? </h3>
</summary>

If you remove that fast path, you can argue it more accurately demonstrates the performance of the new implementation. While this is partly true, the new implementation still has a few "fast paths" for when there are some quotes in the file, but not perhaps on every line.

Nevertheless, since it does represent the performance of just the new implementation (which is worse) I have included it here:

```
------------------------------------------------------------------------------------ benchmark 'CSV_IO': 8 tests ------------------------------------------------------------------------------------
Name (time in ms)                    Min                 Max                Mean             StdDev              Median                IQR            Outliers      OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_csv_io[read-uint64]       414.5707 (7.49)     453.2336 (7.76)     423.7333 (7.44)     16.5427 (20.69)    416.9608 (7.33)     11.3597 (9.57)          1;1   2.3600 (0.13)          5           1
bench_csv_io[read-int64]        408.9339 (7.39)     410.8985 (7.04)     409.6753 (7.19)      0.7997 (1.0)      409.5734 (7.20)      1.1866 (1.0)           1;0   2.4410 (0.14)          5           1
bench_csv_io[read-float64]      451.7167 (8.16)     486.8253 (8.34)     459.3327 (8.06)     15.3819 (19.23)    452.8630 (7.96)      9.7451 (8.21)          1;1   2.1771 (0.12)          5           1
bench_csv_io[read-str]          957.3902 (17.30)    972.2829 (16.65)    964.4374 (16.93)     5.5556 (6.95)     965.3951 (16.98)     7.0409 (5.93)          2;0   1.0369 (0.06)          5           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
</details>

### 3. Compared to performance on main

```

---------------------------------------------------------------------------------------- benchmark 'CSV_IO': 8 tests ----------------------------------------------------------------------------------------
Name (time in ms)                      Min                   Max                  Mean             StdDev                Median                IQR            Outliers      OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_csv_io[read-uint64]         118.1343 (2.22)       119.1238 (2.13)       118.7053 (2.15)      0.3653 (1.02)       118.7932 (2.15)      0.4072 (1.0)           2;0   8.4242 (0.46)          5           1
bench_csv_io[read-int64]          119.2602 (2.24)       120.1037 (2.15)       119.7022 (2.17)      0.3596 (1.0)        119.6677 (2.16)      0.6308 (1.55)          2;0   8.3541 (0.46)          5           1
bench_csv_io[read-float64]        124.0562 (2.33)       125.7878 (2.25)       124.8458 (2.27)      0.6210 (1.73)       124.8347 (2.26)      0.5899 (1.45)          2;0   8.0099 (0.44)          5           1
bench_csv_io[read-str]          6,427.0379 (120.54)   6,478.8276 (115.86)   6,449.4896 (117.01)   20.9028 (58.12)    6,450.0977 (116.55)   32.9164 (80.84)         2;0   0.1551 (0.01)          5           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

So in general, performance is Much better for string reads (close to`7x` better), but worse for typed reads (about `1.5x` worse).

**Most CSV files are all "string reads" in Arkouda** because if Arkouda doesn't see the type headers (only available if arkouda wrote the csv file), then all the reads are string reads.
Given the above, this should improve performance in all cases when reading a CSV file written by some other tool.

If you're writing to CSV from Arkouda and reading it back, and you care about performance, then you should be using parquet anyway.

## Summary

This PR fixes several major issues with Arkouda's CSV parsing:
- Corrects the interpretation of embedded commas inside quoted fields
- Properly handles escaped quotes
- Supports multi-line quoted records
- Adds extensive tests to prevent regression

These improvements bring Arkouda's CSV behavior more in line with RFC 4180 and with tools such as Pandas, enabling more reliable ingestion of real-world CSV datasets.

### DIFF
--- a/tests/pandas/io_test.py
+++ b/tests/pandas/io_test.py
@@ -2508,7 +2508,7 @@ class TestCSV:
                     f.write(
                         f'"{delim_names[i]}"{delimiter}'
                         f'"{delim_descriptions[i]}"{delimiter}'
-                        f'{expected_values[i]}\n'
+                        f"{expected_values[i]}\n"
                     )
 
             data = ak.read_csv(file_name, column_delim=delimiter)


src/CSVMsg.chpl
```diff
@@ -487,38 +492,356 @@ module CSVMsg {
...
+            while true {
+                var hasQuotes = advanceCSVRecord(reader);
+                if hasQuotes {
+                    hasAnyQuotes = true;
+                }
+                rowCount += 1;
+            }
+        } catch e: EofError {
+            // Normal end of file
+        }
```
jabraham17: This kind of code really makes me nervous, relying on an eof thrown from advanceCSVRecord to break the loop is both inefficient and makes the control flow unclear

Two suggestions, one more invasive than the other
* `advanceCSVRecord` returns some kind of bool signal when its reached eof
*  make the `try`/`catch` more localized to `advanceCSVRecord` that then breaks the loop
    * there might be a perf cost for this though

src/CSVMsg.chpl
```diff
@@ -487,38 +492,356 @@ module CSVMsg {
...
+            throw e; // Re-throw EOF to signal end of file
+        } catch e: UnexpectedEofError {
+            throw e;
+        }
+
+        // Return single line if no multi-line record
+        if lines.size == 1 {
+            return lines[0];
+        } else {
+            return "\n".join(lines.toArray());
```
jabraham17: prefer `.join(lines.these())` to avoid allocation

src/CSVMsg.chpl
```diff
@@ -487,38 +492,356 @@ module CSVMsg {
...
+        }
+    }
```
jabraham17: as a nit, this code feels needlessly complex

```chapel
var lines = readCSVRecordLines(reader);
if lines.size == 1 { 
  return lines[0];
} else {
  return "\n".join(lines.toArray());
}
```

This should have the same behavior, no need to manually rethrow the exception. If you want to be more explicit you could do `var lines = try readCSVRecordLines(reader); `

src/CSVMsg.chpl
```diff
@@ -487,38 +492,356 @@ module CSVMsg {
...
+        // Slow path: use full parsing for quoted fields
+        var (boundaries, targetField) = parseCSVCore(csvRecord, colDelim, targetIdx);
+        return targetField;
+    }
+
+    // Helper function to find all field boundaries
+    proc findFieldBoundaries(csvRecord: string, colDelim: string, ref fieldBoundaries: list((int, int))) throws {
+        var (boundaries, targetField) = parseCSVCore(csvRecord, colDelim, -1);
+        fieldBoundaries = boundaries;
+    }
```
jabraham17: why does this need a ref arg instead of just returning?

src/CSVMsg.chpl
```diff
@@ -487,38 +492,356 @@ module CSVMsg {
...
+                            lines.pushBack(line);
+                            totalQuotes += line.count("\"");
+                        } catch e: EofError {
+                            break;
+                        } catch e: UnexpectedEofError {
+                            break;
+                        }
+                    }
+
+                    line = "\n".join(lines.toArray());
```
jabraham17: use `lines.these()`

src/CSVMsg.chpl
```diff
@@ -432,12 +431,16 @@ module CSVMsg {
...
+        // Use proper CSV parsing to handle quoted delimiters
+        var headerLine = readCSVRecord(fr);
+        var columnIndex = 0;
+        for column in parseCSVRecord(headerLine, colDelim) {
             if column == colName {
-                colIdx=i;
+                colIdx = columnIndex;
                 break;
             }
+            columnIndex += 1;
```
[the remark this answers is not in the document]
ShreyasKhandekar: Just cruft from debugging, I will change it back

PR #5090, opened 2025-11-28
Closes #5089: Add a `balancedSegStringRetrieval` function

### LINKED ISSUE #5089: Add a `balancedSegStringRetrieval` function
**Feature Description**
Create a function that does more local copying of string bytes when dealing with a `SegString`. Basically each locale should try to work with bytes that it already owns. Doing this correctly may consume less memory and work faster for lots of data.

1RyanK:
Previously it seemed like the way to divide bytes up appropriately in a SegString was to do something like this:

```
ref offsets = strOffsetInLocale[here.id].Arr;
offsets[0..#size] = strings.offsets.a.localSlice[offsetsDom];
const end = if offsetsDom.high >= strings.offsets.a.domain.high
            then strings.values.a.size
            else globalOffsets[offsetsDom.high + 1];
const start = offsets[0];
strBytesInLocale[here.id] = new innerArray({0..#(end - start)}, uint(8));
strBytesInLocale[here.id].Arr[0..#(end - start)] = strings.values.a[start..<end];
```

As you can see, each locale is just grabbing the strings corresponding to the offsets that are local. This is rather easy to do, but maybe not the best way to do things.

This PR makes it easy to do things a little bit differently - try to grab mostly the bytes that are on the current locale. ~~It also turns everything into an `innerArray` of strings. You can see how simple this makes the `concatenateUniqueStrMsg2` function.~~ Also added `segStringFromInnerArray` to convert back into a `SegString` easily.

The way that it figures out which strings end up on which locales is as follows: if a string is entirely on one locale, then it belongs to that locale. If it overlaps two or more locales, then the locale that ends up owning it is the locale with the most bytes. If there is a tie, then of the locales with the most bytes of that string, the one with the lowest ID gets it.

I have done some benchmarking on this PR and I found that an `innerArray` of strings was bad. I think this is largely because strings are probably objects which have a pointer to the bytes, along with some other useful things for the object. So when you try to bulk transfer an array of strings, it can't just transfer the bytes smoothly. So I reverted back to `innerArray`s of `uint(8)` bytes and the `int(64)` offsets.

I also added in my own hashing function. The reason for this is I can pass in the array, where to start hashing, and where to stop. I think this is not insanely better than using `borrow=true` in `interpretAsString` but there's some additional code surrounding that so maybe it helps performance just a bit. But who knows, I could be wrong.

I also changed it so that rather than doing something like `hash % numlocales` it does `((hash >> 32) * numLocales) >> 32`. The reason is that modulo relies on division, which can be slow. A multiply is fast and this basically gets the job done. Roughly speaking this is almost equivalent to `(hash * numLocales) >> 64`, except for the fact that this way would likely overflow the 64 bit size. So we sacrifice a tiny bit of precision for an easier way to do the calculation.

I opted not to use `balancedSegStringRetrieval` in the `concatenate_uniquely` code because I'm ultimately not really sure how much time it actually saves. That being said - I think the *idea* is useful. Essentially it lends itself really well to cutting out the initial bulk transfer into `innerArray`s. Basically just bulk transfer the surrounding bytes from other locales (according to the same rules as `balancedSegStringRetrieval`), hash, and then bulk transfer out of the local slice to other locales into their `innerArray`s.

Performance stacks up surprisingly well (at least to me) against `unique(concatenate(...))`. For smaller quantities of strings, `uniquely_concatenate` wins. At a certain point, `uniquely_concatenate` starts losing, but it takes a while to get there. Moreover I documented some ideas for improving performance in #5401 that may help get performance up there.

Closes #5089: Add a `balancedSegStringRetrieval` function


src/ConcatenateMsg.chpl
```diff
@@ -653,4 +653,100 @@ module ConcatenateMsg
         return new MsgTuple(repMsg, MsgType.NORMAL);
     }
     registerFunction("concatenateUniquely", concatenateUniqueStrMsg, getModuleName());
+
+    proc concatenateUniqueStrMsg2(cmd: string, msgArgs: borrowed MessageArgs, st: borrowed SymTab): MsgTuple throws {
```
e-kayrakli: ditto the name

src/ConcatenateMsg.chpl
```diff
@@ -653,4 +653,100 @@ module ConcatenateMsg
...
+                // Grab the strings by locale and throw them in the sets.
+                var stringArrays = balancedSegStringRetrieval(segString);
+                coforall loc in Locales do on loc {
+                    ref myStrings = stringArrays[here.id].Arr;
+                    var locSet = new set(string);
+                    
+                    // Reduce to unique strings within this locale
+                    forall str in myStrings with (+ reduce locSet) {
+                        locSet.add(str);
+                    }
```
e-kayrakli: This is an interesting pattern!

It looks like in some tests, at least, using `set(string, parSafe=true)` instead and not doing things through the `+ reduce` intent results in faster performance.

src/Repartition.chpl
```diff
@@ -24,6 +27,495 @@ module Repartition
...
+
+    // Last byte index in the global byte buffer
+    const endOfBytes = strBytes.domain.last;
+
+    // For each locale, the first string-start offset (global) that resides on that locale.
+    // -1 means that locale has no offsets.
+    var firstOffsetPerLoc: [0..#numLocales] int = -1;
+
+    // First pass: for each locale, record its local ranges in the global bytes and offsets.
+    forall loc in 0..#numLocales {
```
e-kayrakli: Right now, this loop runs entirely on locale 0 (though in parallel). It looks like this could actually be done in parallel by locales simultaneously, if I am not mistaken. If that's the case, `coforall loc in Locales.domain do on Locales[loc] {` could be a better loop+`on` combo here.

src/Repartition.chpl
```diff
@@ -24,6 +27,495 @@ module Repartition
...
+                           else myLastOffsetEnd;
+
+            // Full size of the string in bytes
+            const size = endStr - myOffsets[idx];
+            firstStringSizes[i] = size;
+
+            // Portion of this string that resides on locale i's byte range
+            // (truncate end to not go past that locale’s last byte + 1)
+            const endStrThisLoc = min(endStr, myLastIdxPerLoc[i] + 1);
+            firstStringNumBytesEachLocale[i] = endStrThisLoc - myFirstIdxPerLoc[i];
```
e-kayrakli: And another one.

These could be benign write-after-write hazards, though I still wanted to check.

e-kayrakli (approved):
Looks good, with the caveat that the logic was a bit hard to follow. That's not to say that you should change something in the logic, but just to say that I don't think I can catch logical issues and as long as testing is clean, it should be good enough.

There are some inline comments that are mostly stylistic and performance related. Otherwise, looks good!

PR #5094, opened 2025-12-01
Closes #5093: Fix mypy problems

### LINKED ISSUE #5093: Fix mypy problems
**Describe the bug**
mypy is causing issues in the CI. Should find a way to address all the problems.

1RyanK:
All the problems were in `series.py`, which I believe we are moving away from, so I think this is okay.

Closes #5093: Fix mypy problems


drculhane (approved):
Thanks for taking care of this.

PR #5121, opened 2025-12-05
Closes #5111:  numpy operations alignment tests

ajpotts:
# PR: Add Comprehensive NumPy-Alignment Test Suite for Unary, Binary, and Comparison Operators

This PR introduces a complete NumPy-alignment test suite for Arkouda’s `pdarray` operator semantics.  
It adds a new test module:

```
tests/numpy/alignment_verification/operators_alignment.py
```

and registers it in `pytest.ini`, ensuring it runs as part of the standard NumPy compatibility test suite.

---

### **1. Adds a full operator-alignment test suite**

The new module tests:

#### **Binary arithmetic operators**  
- `+`, `-`, `*`, `/`, `//`, `%`, `**`  
- Bitwise ops: `<<`, `>>`, `&`, `|`, `^`

#### **Scalar broadcasting**
- `array ⊗ scalar`  
- `scalar ⊗ array`

#### **Comparison operators**
- `==`, `!=`, `<`, `<=`, `>`, `>=`

#### **Unary operators**
- `+a` (pos)  
- `-a` (neg)  
- `~a` (invert)

This suite validates that Arkouda’s operator implementations **match NumPy semantics** exactly across `int64`, `uint64`, `float64`, and `bool_` dtypes.

---

### **Accurate dtype handling**
- Booleans excluded from inappropriate arithmetic combinations  
- Float-specific behaviors (`truediv`, exponent ranges, invert not allowed)  
- `uint64` and `int64` power and shift constraints

### **Shift and power safety**
The suite constrains RHS values for operators like:

- `pow` (limits exponent size)
- `lshift` / `rshift` (enforces nonnegative RHS)

### **NumPy-equivalent error handling**
Some operations are skipped when NumPy does not support them, ensuring test intent matches NumPy’s behavior.

---

### **pytest.ini**
Adds the new module to Arkouda’s NumPy test paths.  

### **tests/numpy/alignment_verification/operators_alignment.py**
A 227‑line comprehensive alignment suite covering:

- Operator correctness  
- Broadcasting  
- Unary behavior parity  
- Boolean handling  
- Edge-case domain constraints  
- Type-reflection between NumPy arrays and Arkouda pdarrays

---

## ✔ Why This Is Important

This suite provides Arkouda with:

### **1. A single source of truth for NumPy operator semantics**
All operator behavior is now validated rigorously and uniformly.

### **2. A foundation for correcting bugs**
The suite already uncovered several issues:
- Large-shift semantic mismatch for `rshift(int64)`  
[...]

PR #5124, opened 2025-12-05
Closes #5117: Unary Negation on ak.uint64 Does Not Match Numpy Semantics

### LINKED ISSUE #5117: Bug: Unary Negation (`-a`) on `ak.uint64` Does Not Match NumPy Semantics
# Bug: Unary Negation (`-a`) on `ak.uint64` Does Not Match NumPy Semantics

## Summary

Arkouda’s implementation of unary negation (`-a`) for unsigned 64-bit arrays (`ak.uint64`) does **not match NumPy behavior**.  
NumPy defines unary negation for unsigned integers as a **modular wrap-around** operation in the same unsigned dtype:

```
np.uint64(x) → (-x) % 2**64
```

Arkouda instead returns a **signed int64** result with negative values, breaking NumPy alignment tests and producing incorrect, dtype-inconsistent output.

This causes 100% mismatches in:

```
test_unary_ops_alignment[neg-neg-uint64]
```

---

## Expected NumPy Behavior

In NumPy, unary negation on `uint64`:

- Preserves dtype (`uint64`)
- Computes result modulo 2⁶⁴:

```python
>>> import numpy as np
>>> x = np.uint64(5)
>>> -x
18446744073709551611   # 2**64 - 5
>>> np.array([-x], dtype=np.uint64)
array([18446744073709551611], dtype=uint64)
```

---

## Arkouda’s Current Behavior

Arkouda produces **signed negative values** instead of `uint64` wraparound:

```python
>>> import arkouda as ak
>>> a = ak.array([5], dtype=ak.uint64)
>>> (-a).to_ndarray()
array([-5])   # WRONG — treated as signed int64
```

Across random arrays, every element differs from NumPy:

```
[left]:  array([-1830621171, -1957544328, -1935796560, ...])
[right]: [18446744071878930445, 18446744071752007288, ...]
```

These pairs differ by exactly `2**64`, proving Arkouda is performing **signed negation**, not `uint64` modulo negation.

This leads to:

- 100% mismatch rate in alignment tests  
- Incorrect dtype (signed instead of unsigned)  
- Incorrect numerical results  

---

### 1. NumPy Compatibility  
Arkouda aims to match NumPy semantics for distributed arrays.  
Unsigned negation is a well-defined operation in NumPy and widely relied upon.

### 2. Dtype Preservation  
Users expect unary operations to preserve dtype unless NumPy explicitly promotes.  
Arkouda violates this expectation.

### 3. Downstream Bitwise Logic  
Unsigned wraparound negation is used in hashing, bitboard logic, cryptographic ops, bloom filters, etc. Negation returning signed incorrect values breaks these algorithms.

---

## Root Cause

Arkouda appears to interpret `uint64` arrays as signed when applying unary minus:

- Likely calling a signed `-x` operator on the underlying Chapel array.
- No modulo operation or dtype preservation is performed.
- Result pdarray inherits the backend’s signed numerical interpretation.

NumPy, in contrast, uses **fixed-width modular arithmetic** for all unsigned integer operations.

---

### **Correct Implementation for Unary Minus on `ak.uint64`:**

```
(-x) → (2**64 - x) % 2**64
```

Backend logic should:

1. Detect dtype == `uint64`
2. Compute modular negation:

   ```chapel
   const mask:uint(64) = 0xFFFFFFFFFFFFFFFF;
   result = (~x + 1) & mask;   // equivalent to (-x) mod 2**64
   ```

3. Return a `uint64` pdarray, preserving dtype.

---

## Alternative (If Full Support Is Undesired)

If Arkouda chooses **not** to support unary negation for unsigned integers (less ideal for NumPy parity), then:

- Raise a `TypeError("Unary minus not supported for uint64")`
- Mirror NumPy error semantics for unsupported operations (if any)

But given NumPy *does* define the operation, best practice is to implement it correctly.

---

## Acceptance Criteria

- Unary negation on `ak.uint64` **matches NumPy exactly**:

  ```python
  assert_arkouda_array_equivalent(-a_ak, -a_np)
  ```

- Result dtype remains `uint64`  
- Modular negation works for all values, including large ones  
- Alignment test:

  ```
  test_unary_ops_alignment[neg-neg-uint64]
  ```

  passes without modification.

---

## Minimal Reproduction

```python
>>> import numpy as np, arkouda as ak

>>> a_np = np.random.randint(0, 2**64, 5, dtype=np.uint64)
>>> a_ak = ak.array(a_np)

>>> np_res = -a_np
>>> ak_res = (-a_ak).to_ndarray()

>>> print(np_res)
>>> print(ak_res)
```

Expected: arrays match exactly  
Actual: Arkouda produces negative signed values

---

## Conclusion

Unary negation on `uint64` is currently misimplemented.  
Arkouda must use **unsigned modular negation** rather than signed negation to achieve full NumPy compatibility and correct numeric behavior.

### LINKED ISSUE #5119: Bug: Unary Negation (`-a`) on `ak.bool_` Does Not Match NumPy Semantics
# Bug: Unary Negation (`-a`) on `ak.bool_` Does Not Match NumPy Semantics

## Summary

Arkouda currently **allows unary negation** (`-a`) on boolean arrays (`ak.bool_`), implicitly converting booleans to integers (`True → 1`, `False → 0`) and then negating them (`-1`, `0`).

NumPy **does not allow** unary negation on boolean arrays. Attempting `-np.array([True, False])` raises:

```
TypeError: The numpy boolean negative, the `-` operator, is not supported,
use the `~` operator or the logical_not function instead.
```

To maintain NumPy compatibility—as well as predictable semantics for boolean operations—Arkouda must be updated so that applying unary `-` to a boolean `pdarray` **raises a TypeError** rather than silently performing integer negation.

This mismatch currently causes the alignment test:

```
test_unary_ops_alignment[neg-neg-bool]
```

to error out on the NumPy side before comparisons can occur.

---

## Current Arkouda Behavior (Incorrect)

```python
>>> import arkouda as ak
>>> a = ak.array([True, False, True])

>>> -a
array([-1,  0, -1])    # Returned as int64
```

Arkouda performs **implicit dtype promotion** (`bool → int64`) and computes integer negation.  
NumPy *never* implicitly casts bool → int for unary minus.

---

## Expected NumPy-Compatible Behavior

Unary negation on boolean arrays should:

1. **Raise a TypeError**
2. Provide a message comparable to NumPy’s guidance

Example NumPy output:

```python
>>> import numpy as np
>>> -np.array([True, False])
TypeError: The numpy boolean negative, the `-` operator, is not supported,
use the `~` operator or the logical_not function instead.
```

Arkouda should raise something similar:

```python
TypeError("Unary negation (`-`) is not supported for boolean pdarrays; use `~a` or `ak.logical_not(a)` instead.")
```

---

### 🔹 NumPy Compatibility  
Arkouda’s boolean behavior should match NumPy exactly wherever possible.

### 🔹 Prevent Silent Type Promotion  
Silently converting boolean → int64 introduces errors that may go unnoticed in downstream computations.

### 🔹 Predictable Boolean Semantics  
Unary minus simply has no meaningful boolean interpretation; users should use:

- `~a` (bitwise invert)
- `ak.logical_not(a)` (boolean negation)

---

## Proposed Fix (Frontend)

Modify `pdarray.__neg__` in:

```
arkouda/numpy/pdarrayclass.py
```

to detect the boolean dtype:

```python
def __neg__(self):
    if self.dtype == ak.bool_.dtype:
        raise TypeError(
            "Unary negation (`-`) is not supported for boolean pdarrays; "
            "use `~a` or ak.logical_not(a) instead."
        )
    return <existing numeric negation logic>
```

No backend changes are needed; this is a client-side rule.

---

## Test Updates

Update the unary alignment test for `neg` + `bool`:

```python
if dtype is ak.bool_ and name == "neg":
        fn(_numpy_equivalent(a))
        fn(a)
    return
```

After this, the alignment suite will pass for this configuration.

---

## Acceptance Criteria

- `-ak.array([...], dtype=ak.bool_)` raises a `TypeError`  
- Error message suggests using `~` or `ak.logical_not`  
- No dtype coercion or negation occurs behind the scenes  
- NumPy-alignment tests confirm fully matched error behavior  
- No regression for numeric negation behavior (`ak.int64`, `ak.uint64`, `ak.float64`)

---

## Minimal Reproduction (Before Fix)

```python
>>> import numpy as np, arkouda as ak
>>> a_np = np.array([True, False])
>>> a_ak = ak.array([True, False])

>>> -a_np
TypeError: The numpy boolean negative, the `-` operator, is not supported...

>>> -a_ak
array([-1,  0])    # WRONG
```

---

## Conclusion

Unary minus on booleans is currently inconsistent with NumPy and can lead to subtle errors.  
Arkouda should raise a clear `TypeError` for `-a` when `a.dtype == ak.bool_`.

This ticket tracks implementing that behavior and updating the alignment test expectations accordingly.

1RyanK:
Since both of these are unary operators and in light of the testing framework in PR #5121, I thought I would bundle both the negation changes for `uint64` and `bool` together.

For unary negation on `uint64`, twos complement suggests that the way to handle this is to negate the bits and add 1.

For unary negation on `bool`, the solution suggested was to raise a `TypeError` and suggest using `~` instead or `logical_not`, but since `logical_not` did not exist, I added it and have it pointing at `~` since they seem to largely be the same thing. I think additional tests specific to `logical_not` would be unnecessary if `~` is being tested because it's literally a wrapper for `~`. However, I did grab the unit tests from PR #5121 that seem relevant to this PR.

Closes #5117: Unary Negation on ak.uint64 Does Not Match Numpy Semantics
Closes #5119: Bug: Unary Negation (-a) on ak.bool_ Does Not Match NumPy Semantics


1RyanK:
> From the looks of the operators_alignment.py, you have more in mind for this. As is, it looks good.

It's mostly just ripped out from the relevant parts of #5121

ajpotts (approved):

ajpotts (approved):

PR #5131, opened 2025-12-09
Closes 5113, fixing behavior of floor division operator (//)

### LINKED ISSUE #5113: Bug Report: Integer Floor Division (`//`) Misalignment Between Arkouda and NumPy
# Bug Report: Integer Floor Division (`//`) Misalignment Between Arkouda and NumPy

## Summary

Arkouda’s integer floor-division (`//`) between an `ak.pdarray` (dtype `int64`) and a scalar does **not** match NumPy’s results.  
This produces widespread discrepancies in alignment tests and breaks algorithms that rely on mathematically correct floor-division semantics, particularly when **negative values** are involved.

NumPy uses true mathematical floor division:

```
x // y = floor(x / y)
```

Arkouda currently appears to use **C-style truncation toward zero**, leading to incorrect results for negative integers.

---

## Example (no reference to unit tests)

```python
>>> import numpy as np
>>> import arkouda as ak

# Create values that cross zero
>>> arr_np = np.array([-9, -7, -5, -1, 0, 1, 5, 7, 9], dtype=np.int64)
>>> arr_ak = ak.array(arr_np)

>>> d = 3   # positive divisor
```

### Expected behavior (NumPy / Python)
NumPy uses true floor-division:

```python
>>> arr_np // d
array([-3, -3, -2, -1,  0,  0,  1,  2,  3])
```

Explanation:
- -7 / 3 ≈ -2.333 → floor → **-3**
- -1 / 3 ≈ -0.333 → floor → **-1**

---

### Arkouda’s current behavior (incorrect)
```python
>>> (arr_ak // d).to_ndarray()
array([-3, -2, -1,  0,  0,  0,  1,  2,  3])
```

Notice mismatches:

| Value | NumPy | Arkouda (today) | Error |
|-------|--------|------------------|--------|
| -7 // 3 | **-3** | **-2** | wrong |
| -5 // 3 | **-2** | **-1** | wrong |
| -1 // 3 | **-1** | **0** | wrong |

Arkouda is performing truncation toward zero rather than flooring.

---

## Why this matters

Python, NumPy, and pandas **all guarantee**:

```
x // y is the mathematical floor of the division
```

Many algorithms rely on this:

- bucketization  
- hashing  
- range partitioning  
- geometric tiling  
- coordinate transforms  

Incorrect floor division produces invalid bin IDs, negative bucket indices turning positive, and broken partition boundaries.

Example:

```python
value = -7
bucket_width = 3

# Python
>>> -7 // 3
-3

# Arkouda (current)
>>> ak.array([-7]) // 3
array([-2])   # WRONG
```

This shifts entire buckets one index off.

---

## Root cause

Arkouda is using C/C++-style division semantics:

```
result = trunc(x / y)
```

instead of:

```
result = floor(x / y)
```

For positive values they coincide; for negative values they diverge.

---

## Proposed Fix

Implement Python/NumPy-compatible floor division:

```
x // y = floor(x / y)
```

A correct scalar implementation for integer division is:

```python
def py_floor_div(x, y):
    q = x // y        # using Python semantics, not C
    return q
```

A C-style remainder-based correction:

```python
# Given q = trunc(x / y), r = x % y
# Adjust q when signs differ and remainder != 0
if (r != 0) and ((x < 0) != (y < 0)):
    q -= 1
```

Arkouda should implement this logic at the pdarray operation level.

---

## Acceptance Criteria

- Arkouda `ak.int64 // scalar` matches NumPy for all signs of numerator and denominator  
- No truncation-toward-zero for negative values  
- Roundtrip example above produces identical results between Arkouda and NumPy  
- Scalar–array and array–scalar both fixed  
- Behavior consistent across:  
  - `ak.int64`  
  - `ak.uint64` (where meaningful)  
  - future integer types  

---

## Example for documentation after fix

```python
>>> import arkouda as ak
>>> import numpy as np

>>> arr = ak.array([-9, -7, -5, -1, 0, 1, 5, 7, 9])
>>> (arr // 3).tolist()
[-3, -3, -2, -1, 0, 0, 1, 2, 3]
```

This should exactly match:

```python
>>> (np.array([-9, -7, -5, -1, 0, 1, 5, 7, 9]) // 3).tolist()
[-3, -3, -2, -1, 0, 0, 1, 2, 3]
```

---

## Additional context

Debug lines such as:

```
[pdarrayclass] DEBUG: created Chapel array ...
[pdarrayclass] DEBUG: deleting pdarray ...
```

are normal lifecycle logs and **not** related to the numerical mismatch.  
The underlying issue is the integer division semantics in Arkouda.

drculhane:
Closes #5113 

**Update** -- after a chat with Ryan, I realized that it wasn't really enough to correct the behavior of // without also distributing it.  So this version does that, with a new function called _floorDivision_, which calls a slightly changed _floorDivisionHelper_.

There are three versions of _floorDivision_, for the vv, vs and sv cases.

All of the calls to _floorDivisionHelper_ in BinOp, which were zippered foralls, now call _floorDivision_.

I also changed the calls to _floorDivisionHelper_ in TimeClassMsg.chpl and OperatorMsg.chpl.

Also at Ryan's suggestion, I removed "inline" from the declaration of _floorDivisionHelper_.  I haven't yet addressed _ModHelper_.  That will be in the PR for the behavior of the % operator.

---------

**Previously** -- I modified _floorDivisionHelper_ to handle more data types, and invoked it in places where chapel had been performing a simple (and incorrect) a/b division.

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -96,61 +96,115 @@ module BinOp
 
     }
 
-  /*
-    Helper function to ensure that floor division cases are handled in accordance with numpy
-  */
+  // vector-vector case of floor division      
 
-  inline proc floorDivisionHelper(numerator: real(64), denom: real(64)): real(64) {
-    if (numerator == 0 && denom == 0) || (isInf(numerator) && (denom != 0 || isInf(denom))){
-      return nan;
+  proc floorDivision (dividend: [?d] ?tn, divisor: [d] ?td, type etype) : [d] etype throws {
+    var quotient = makeDistArray(d, etype);
+    forall idx in d {
+        const numerator = dividend[idx]:etype ;
+        const denom     = divisor[idx]:etype;
+        quotient[idx] = floorDivisionHelper(numerator, denom, etype);
     }
-    else if (numerator > 0 && denom == -inf) || (numerator < 0 && denom == inf){
-      return -1:real;
+    return quotient;
+  }
+                
+  // vector-scalar case of floor division
+         
+  proc floorDivision (dividend: [?d] ?tn, divisor : ?td, type etype) : [d] etype throws {
+    var quotient = makeDistArray(d, etype);
+    const denom = divisor:etype;
+    forall idx in d {
+        const numerator = dividend[idx]:etype ;
+        quotient[idx] = floorDivisionHelper(numerator, denom, etype);
     }
-    else if denom == 0 || isInf(denom) {
-      return numerator / denom;
-    }
-
-    const q  = numerator / denom;
-    const fq = floor(q);
-    if q != fq then return fq;
-
-    // fma does (-q * denom) + numerator and stores that in r
-    // From https://en.cppreference.com/w/c/numeric/math/fma.html
-    // Computes (x * y) + z as if to infinite precision and rounded only once to fit the result type.
-    const r = fma(-q, denom, numerator);
-    if r == 0.0 then return q;
-
-    // This next part may seem a little weird, but if r = -q * denom + numerator
-    // Then numerator / denom = q + r / denom (This should remind you of the division algorithm)
-    // r should be small, it's basically the floating point error when we calculated q.
-    // At this point, q is an integer because we already returned if q != fq.
-    // If r / denom is negative then we need to actually round down. If r / denom is positive, then q
-    // is the correct floor division.
-    // Rather than actually dividing the values, we only need to check if the signs differ or are the
-    // same. If they differ, then r / denom is negative, otherwise it is positive.
-    const qIsBelow = ((r > 0.0) != (denom > 0.0));
-    return if qIsBelow then (q - 1.0) else q;
+    return quotient;
   }
 
-  /*
-    Helper function to ensure that mod cases are handled in accordance with numpy
-  */
-  inline proc modHelper(dividend: ?t, divisor: ?t2): real {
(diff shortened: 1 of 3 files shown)


src/BinOp.chpl
```diff
@@ -97,60 +97,93 @@ module BinOp
...
-    // when the dividend and divsor have opposite signs,
-    // we add the divsor into the result
-    // except for when res == 0 (divsor even divides dividend)
-    // see https://en.wikipedia.org/wiki/Modulo#math_1 for more information
-    if res != 0 && (((dividend < 0) && (divisor > 0)) || ((dividend > 0) && (divisor < 0))) {
-      // we do + either way because we want to shift up for positive divisors and shift down for negative
-      res += divisor;
-    }
-    return res;
+  inline proc floorDivisionHelper(dividend : ?tn, divisor : ?td) : floorDivisionType(tn,td) {
```
1RyanK: I feel like this is slightly heavier machinery than we need on int, uint, bool. I'm slightly concerned it may impact performance. Similar with modHelper I think

drculhane: I'm afraid I don't understand.  Everything from line 117 to 154 was already there.  The added lines from 154 to 167, which fix the problem, aren't computationally heavy.  As it happens, they also aren't invoked in the bool case, which is covered without calling floorDivisionHelper.

PR #5138, opened 2025-12-10
Remove typevar from sorting

ajpotts:
Removes `TypeVar` usage from `sorting` module.

#5139:  remove TypeVar usage from sorting module


drculhane (approved):
Straightforward.

PR #5145, opened 2025-12-11
Closes #5144: Clean up linking warning in make

### LINKED ISSUE #5144: Clean up linking warning in `make`
Every time I run `make` I get
```bash
clang++: warning: argument unused during compilation: '-L/home/ryan/miniconda3/envs/arkouda-dev/lib' [-Wunused-command-line-argument]
```
It'd be nice if it wasn't that way

1RyanK:
### **Fixed unused `-L` warnings during C++ compilation by separating compile and link flags**

This PR updates the `add-path` macro in the Makefile to avoid passing `-L` (library search paths) to compilation steps. Previously, `INCLUDE_FLAGS` contained both `-I` and `-L` paths, but `INCLUDE_FLAGS` is used for *compilation* (`clang++ -c`). Since the `-L` flag is only meaningful at *link* time, Clang emitted warnings like:

```
warning: argument unused during compilation: '-L<path>' [-Wunused-command-line-argument]
```

### **What changed**

* `INCLUDE_FLAGS` now contains only `-I` include paths.
* `CHPL_FLAGS` now contains the `-L` paths and corresponding `--ldflags="-Wl,-rpath,..."`.
* No functional changes to linked libraries; linking behavior remains the same.

### **Why this matters**

* Eliminates noisy compiler warnings during the Arrow helper builds.
* Cleans up the build system so compile/link flags are used correctly.
* Makes it easier to spot *actual* warnings in CI.

---

Closes #5144: Clean up linking warning in make

### DIFF
--- a/make/Prologue.mk
+++ b/make/Prologue.mk
@@ -71,12 +71,19 @@ endif
 
 # add-path: Append custom paths for non-system software.
 define add-path
+# Always add headers for both C compilation and Chapel-generated C compilation
+INCLUDE_FLAGS += -I$(1)/include
+CHPL_FLAGS    += -I$(1)/include
+
+# Add lib64 if present (independent of lib)
 ifneq ("$(wildcard $(1)/lib64)","")
-  INCLUDE_FLAGS += -I$(1)/include -L$(1)/lib64
-  CHPL_FLAGS    += -I$(1)/include -L$(1)/lib64 --ldflags="-Wl,-rpath,$(1)/lib64"
+CHPL_FLAGS += -L$(1)/lib64 --ldflags="-Wl,-rpath,$(1)/lib64"
 endif
-INCLUDE_FLAGS += -I$(1)/include -L$(1)/lib
-CHPL_FLAGS    += -I$(1)/include -L$(1)/lib --ldflags="-Wl,-rpath,$(1)/lib"
+
+# Add lib if present
+ifneq ("$(wildcard $(1)/lib)","")
+CHPL_FLAGS += -L$(1)/lib --ldflags="-Wl,-rpath,$(1)/lib"
+endif
 endef
 # Usage: $(eval $(call add-path,/home/user/anaconda3/envs/arkouda))
 #                               ^ no space after comma


make/Prologue.mk
```diff
@@ -71,12 +71,18 @@ endif
...
 # add-path: Append custom paths for non-system software.
 define add-path
-ifneq ("$(wildcard $(1)/lib64)","")
-  INCLUDE_FLAGS += -I$(1)/include -L$(1)/lib64
-  CHPL_FLAGS    += -I$(1)/include -L$(1)/lib64 --ldflags="-Wl,-rpath,$(1)/lib64"
-endif
-INCLUDE_FLAGS += -I$(1)/include -L$(1)/lib
-CHPL_FLAGS    += -I$(1)/include -L$(1)/lib --ldflags="-Wl,-rpath,$(1)/lib"
+  # Compile-time include path
+  INCLUDE_FLAGS += -I$(1)/include
```
jabraham17: this should not be indented

jaketrookman (approved):
Looks good

make/Prologue.mk
```diff
@@ -71,12 +71,19 @@ endif
...
 define add-path
+# Always add headers for both C compilation and Chapel-generated C compilation
+INCLUDE_FLAGS += -I$(1)/include
+CHPL_FLAGS    += -I$(1)/include
+
+# Add lib64 if present (independent of lib)
 ifneq ("$(wildcard $(1)/lib64)","")
-  INCLUDE_FLAGS += -I$(1)/include -L$(1)/lib64
-  CHPL_FLAGS    += -I$(1)/include -L$(1)/lib64 --ldflags="-Wl,-rpath,$(1)/lib64"
+CHPL_FLAGS += -L$(1)/lib64 --ldflags="-Wl,-rpath,$(1)/lib64"
```
e-kayrakli: This line and the one below inside the other `ifneq` _should_ be indented though, right? I don't write a ton of Makefile code, so I would be happy to be pointed otherwise.

I see @jabraham17's [comment](https://github.com/Bears-R-Us/arkouda/pull/5145#discussion_r2620525033) mentioned unindenting something, but it is hard to be certain what they referred to with squashed+rebased history. I am guessing based on the diff that they were referring to `INCLUDE_FLAGS` that's in the main body of the file? I don't think we should be unindenting everything.

1RyanK: If you look at the examples in [this link](https://www.gnu.org/software/make/manual/make.html#Multi_002dLine), you can see that they don't seem to indent between `define` and `endef`.

e-kayrakli: I don't mean to be stickler, but that documentation is not consistent. There are `ifeq` blocks in that documentation with mixed indentation styles.

My C reflexes expect `define` blocks to _not_ be indented in any case. But whether my C reflexes apply here or not is a different question. :)

In any case, I see that there is no agreed upon style, with some suggesting 0 indentation to avoid the notorious tab vs space issues in Makefile. I'm OK with the style as is. Thanks for entertaining the thought!

Issue #4766, opened 2025-08-06
Compilation Error Due to Deprecated `OpenFile` API in Arrow 21.0.0+


ShreyasKhandekar:
**Describe the bug**
 Arkouda fails to compile when using Apache Arrow 21.0.0 or later due to a removed API: `parquet::arrow::OpenFile`. The version currently used in `UtilParquet.cpp` accepts three arguments and returns an `arrow::Status`, which was deprecated in Arrow 19.0.0 and removed entirely in Arrow 21.0.0.

**To Reproduce**
1. Build Arkouda with Arrow/pyarrow version 21.0.0 or higher
2. See error

**Expected behavior**
 Compilation should complete successfully, regardless of Arrow version, or fail gracefully with a clear version requirement.

**Error Message**
Multiple instances of the same error:
```
/Users/name/arkouda/src/parquet/UtilParquet.cpp:464:20: error: too many arguments to function call, expected 2, have 3; did you mean 'parquet::ParquetFileReader::OpenFile'?
  464 |     ARROWSTATUS_OK(parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
      |                    ^~~~~~~
      |                    parquet::ParquetFileReader::OpenFile 
```

**Is this a Blocking Issue**
 Yes, this prevents Arkouda from building with newer versions of Arrow.
- Are there any known work-arounds? Yes
- If a known work around exists, please provide a code block demonstrating it.
    - Restrict pyarrow and Arrow C++ versions to `<21.0.0`. 

**`ak.get_config()` Output**
 N/A (build failure occurs before Arkouda can be run)

**Proposed Fix**
The deprecated `OpenFile` method was removed in Arrow 21.0.0. Its usage in `UtilParquet.cpp` needs to be updated to the new API (which returns an `arrow:Result` rather than `arrow:Status`). Until then, we should restrict Arrow/pyarrow versions to a range that includes the deprecated method. See [Arrow 20.0.0 API docs](https://arrow.apache.org/docs/20.0/cpp/api/formats.html#_CPPv48OpenFileNSt10shared_ptrIN5arrow2io16RandomAccessFileEEEPN5arrow10MemoryPoolEPNSt10unique_ptrI10FileReaderEE) for reference.

[...]

PR #5147, opened 2025-12-11
Closes #5146: Update Parquet file reading to use Arrow Result-returning OpenFile API

### LINKED ISSUE #5146: Update Parquet file reading to use Arrow Result-returning OpenFile API
Arkouda’s Parquet utility code (`src/parquet/UtilParquet.cpp`) still uses the deprecated overload of:

```cpp
parquet::arrow::OpenFile(...)
```

Arrow 19.0.0 deprecates this version and requires using the `arrow::Result<T>`-returning API instead. This produces build warnings such as:

```
'OpenFile' is deprecated: Deprecated in 19.0.0. Use arrow::Result version instead.
```

Updating to the Result-returning API requires two changes:

1. Replace all uses of

   ```cpp
   ARROWSTATUS_OK(parquet::arrow::OpenFile(..., &reader));
   ```

   with

   ```cpp
   ARROWRESULT_OK(parquet::arrow::OpenFile(...), reader);
   ```

2. Update the `ARROWRESULT_OK` macro to correctly move the returned `unique_ptr`, since `unique_ptr` is non-copyable.

This issue tracks modernizing the Parquet reading code and removing the deprecation warnings.

1RyanK:
### **Modernize Parquet file reading to Arrow’s Result-returning API**

This PR updates all uses of `parquet::arrow::OpenFile` in `UtilParquet.cpp` to the non-deprecated Arrow 19+ API. The old API wrote into a `FileReader*` via an out-parameter, while the new API returns:

```cpp
arrow::Result<std::unique_ptr<parquet::arrow::FileReader>>
```

### **Key changes**

* Replaced all instances of:

  ```cpp
  ARROWSTATUS_OK(parquet::arrow::OpenFile(..., &reader));
  ```

  with:

  ```cpp
  ARROWRESULT_OK(parquet::arrow::OpenFile(...), reader);
  ```

* Updated the `ARROWRESULT_OK` macro to use:

  ```cpp
  res = std::move(result.ValueOrDie());
  ```

  to correctly extract the move-only `unique_ptr` returned by Arrow.

### **Why this change is needed**

* Removes Arrow 19 deprecation warnings from the build.
* Aligns Arkouda’s Parquet integration with the current Arrow API.
* Makes future Arrow upgrades simpler.
* No changes to behavior or performance; this is purely an API modernization.

### **Notes**

* All existing error-handling behavior remains unchanged.
* No impact to downstream Chapel code or server behavior.

---

Closes #5146: Update Parquet file reading to use Arrow Result-returning OpenFile API

### DIFF
--- a/src/parquet/UtilParquet.cpp
+++ b/src/parquet/UtilParquet.cpp
@@ -22,7 +22,7 @@
       *errMsg = strdup(result.status().message().c_str());      \
       return ARROWERROR;                                        \
     }                                                           \
-    res = result.ValueOrDie();                                  \
+    res = std::move(result.ValueOrDie());                       \
   }
 
   // The `ARROWSTATUS_OK` macro should be used when calling an
@@ -60,7 +60,7 @@ int64_t cpp_getNumRows(const char* filename, char** errMsg) {
                    infile);
 
     std::unique_ptr<parquet::arrow::FileReader> reader;
-    ARROWSTATUS_OK(parquet::arrow::OpenFile(infile, arrow::default_memory_pool(), &reader));
+    ARROWRESULT_OK(parquet::arrow::OpenFile(infile, arrow::default_memory_pool()), reader);
   
     return reader -> parquet_reader() -> metadata() -> num_rows();
   } catch (const std::exception& e) {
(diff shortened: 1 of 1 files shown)


ShreyasKhandekar:
This is great!

If I understand correctly, since we've moved to the new OpenFile API, this should also resolve https://github.com/Bears-R-Us/arkouda/issues/4766 and therefore we no longer have to limit the Arrow version to be less than 21. Which means we can revert https://github.com/Bears-R-Us/arkouda/pull/4769.

Does that match your understanding @1RyanK?

PR #5148, opened 2025-12-11
Adds "bool" to ak.isnan argument types

### LINKED ISSUE #5130: isnan to support bool arrays
`ak.isnan` should support boolean arrays, to match NumPy, but right now I'm seeing the following:

```
In [10]: x = ak.array([True, False])

In [11]: ak.isnan(x)
TypeError                                 Traceback (most recent call last)
Cell In[11], line 1
----> 1 ak.isnan(x)

File ~/anaconda3/envs/arkouda-dev/lib/python3.13/site-packages/typeguard/__init__.py:891, in typechecked.<locals>.wrapper(*args, **kwargs)
    889 memo = _CallMemo(python_func, _localns, args=args, kwargs=kwargs)
    890 check_argument_types(memo)
--> 891 retval = func(*args, **kwargs)
    892 check_return_type(retval, memo)
    894 # If a generator is returned, wrap it if its yield/send/return types can be checked

File ~/git/arkouda/arkouda/numpy/numeric.py:707, in isnan(pda)
    705     return full(pda.size, False, dtype=bool)
    706 elif not is_numeric(pda):
--> 707     raise TypeError("isnan only supports pdarray of numeric type.")
    709 repMsg = generic_msg(
    710     cmd=f"isnan<{pda.ndim}>",
    711     args={
    712         "pda": pda,
    713     },
    714 )
    715 return create_pdarray(repMsg)

TypeError: isnan only supports pdarray of numeric type.

In [12]: np.isnan(x.to_ndarray())
Out[12]: array([False, False])


```

drculhane:
Closes #5130.

### Changes:

Inside ak.isnan, the check that was `is_numeric(pda)` is now `_datatype_check(pda.dtype,NUMERIC_TYPES,"isnan')`, thus including bool in the check.

If the datatype is not float (but is int, uint, or bool), `full` is now invoked with `pda.shape` instead of `pda.size`.  `pda.size` was a remnant from when we were only using 1D pdarrays.

There were comments in pdarraycreation.py that also referred to `pda.size`, which have been changed to `pda.shape`.  The actual code was already using `pda.shape`.  Only the comments were out of date.

In numeric_test.py, `test_isnan` now includes uint and bool types in the test.

### Possible future work:

numpy isnan, like many numpy functions, includes a "where" parameter.  ak.isnan does not include that.  We may want to add an additional issue.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -57,7 +57,7 @@ testpaths =
     tests/numpy/char_test.py
     tests/numpy/dtypes_test.py
     tests/numpy/manipulation_functions_test.py
-    tests/numpy/array_manipulation_tests.py
+    tests/numpy/pdarraymanipulation_tests.py
     tests/numpy/numeric_test.py
     tests/numpy/numpy_test.py
     tests/numpy/random_test.py


jaketrookman (approved):
Looks good

PR #5160, opened 2025-12-15
Closes #5159: Create a benchmark for uniquely concatenating strings

### LINKED ISSUE #5159: Create a benchmark for uniquely concatenating strings
**Feature Description**
Currently there are roughly two obvious ways of concatenating strings together without repeats. We can do `Strings.concatenate_uniquely` or `ak.unique(ak.concatenate(...))`. It would be good to know if there's anything gained by the first method, or if the hopeful improvement in #5090 is better than either (or further, if there is any way to improve at all).

1RyanK:
In order to test performance in PR #5090 I thought it would be good to have a benchmark. This sets up the framework, and then in that PR I will modify the benchmark to include the new function introduced there.

Closes #5159: Create a benchmark for uniquely concatenating strings

### DIFF
--- a/benchmark_v2/optional/concatenate_uniquely_benchmark.py
+++ b/benchmark_v2/optional/concatenate_uniquely_benchmark.py
@@ -1,0 +1,96 @@
+import pytest
+
+import arkouda as ak
+
+from arkouda import Strings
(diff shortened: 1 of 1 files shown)


drculhane (approved):
Looks straightforward.

PR #5169, opened 2025-12-17
Closes #5168:  gh-pages set cancel-in-progress=true

### LINKED ISSUE #5168: gh-pages set cancel-in-progress=true
In gh-pages.yml set `cancel-in-progress=true`, so that when a PR is resubmitted, any documentation build in progress will be canceled, since it is being replaced by the more recent run.

ajpotts:
Cancel in-progress gh-pages documentation builds when a newer run starts, ensuring only the lated PR update is built and deployed.

Closes #5168:  gh-pages set cancel-in-progress=true

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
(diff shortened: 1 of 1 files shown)


jaketrookman (approved):
Looks good

PR #5184, opened 2025-12-17
Closes #5183:  Increase Literals typing through project

### LINKED ISSUE #5183: Increase Literals typing through project.
Identify type hints that should be Literals instead of str throughout the project.

ajpotts:
Increases the usage of Literals as type hints throughout the project.  In general, `Literal` is preferred over `str` for type hints in cases where only a finite set of values is allowed.

Closes #5183:  Increase Literals typing through project


1RyanK (approved):
Looks good!

PR #5212, opened 2025-12-22
Closes 5170 adding overloads to reshape

### LINKED ISSUE #5170: overloads to reshape function
Add overloads to `reshape` function for more ergonomic type checking.

drculhane:
Closes #5170.

This adds overloads to _reshape_.

It also fixes a bug in _test_flatten_multidim_.  The expression `a.reshape(2,2,size/4)` has been changed to `a.reshape(2,2,size//4)`.  The reason is that size/4 is interpreted as a float, and floats aren't allowed in a shape.  This is consistent with numpy.

### DIFF
--- a/arkouda/numpy/pdarrayclass.py
+++ b/arkouda/numpy/pdarrayclass.py
@@ -6,7 +6,7 @@ import json
 from functools import reduce
 from math import ceil
 from sys import modules
-from typing import TYPE_CHECKING, List, Literal, Optional, Tuple, Union, cast, overload
+from typing import TYPE_CHECKING, List, Literal, Optional, Sequence, Tuple, Union, cast, overload
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


arkouda/numpy/pdarrayclass.py
```diff
@@ -2644,7 +2644,17 @@ def bigint_to_uint_arrays(self) -> List[pdarray]:
...
 
-    def reshape(self, *shape):
+    @overload
+    def reshape(self, shape: Sequence[int_scalars]) -> pdarray: ...
+
+    @overload
+    def reshape(self, *shape: int_scalars) -> pdarray: ...
+
+    @overload
+    def reshape(self, shape: pdarray) -> pdarray: ...
```
1RyanK: Should we add an `@overload` for numpy arrays?

```python
In [18]: shape_np = np.array([6, 4], dtype=np.int64)
    ...: try:
    ...:     a.reshape(shape_np)
    ...: except Exception as e:
    ...:     print(type(e).__name__, e)
    ...: 
TypeError type of argument "shape"[0] must be one of (int, int8, int16, int32, int64, uint8, uint16, uint32, uint64, Sequence, pdarray); got numpy.ndarray instead
```

drculhane: Done.  I had most of this code in there, but got sidetracked.

PR #5223, opened 2025-12-26
Adds overloads to ak.full

### LINKED ISSUE #5173: overloads to full function
Add overloads to the `full` function for more ergonomic type checking.

drculhane:
Closes #5173 

This one required a fair amount of work to satisfy mypy once the overloads were written.  I'm going to keep it in draft initially.

Notes:

In addition to the work on the overloads, mypy found an error in the order of calling parameters to full in my recent PR that changed ak.minimum and ak.maximum.  So that's been fixed.

Once the overloads were in place, mypy flagged an error in the use of full in isna function in the recent pandas extensions.  I addressed that by converting the result of ak.full to an ArkoudaArray before returning.

### DIFF
--- a/arkouda/numpy/pdarraycreation.py
+++ b/arkouda/numpy/pdarraycreation.py
@@ -32,9 +32,11 @@ from arkouda.numpy.dtypes import (
     int_scalars,
     is_supported_int,
     is_supported_number,
+    numeric_and_bool_scalars,
     numeric_scalars,
     resolve_scalar_dtype,
     str_,
+    str_scalars,
 )
 from arkouda.numpy.dtypes import dtype as akdtype
 from arkouda.numpy.dtypes import int64 as akint64
@@ -664,7 +666,7 @@ def ones(
     size: Union[int_scalars, Tuple[int_scalars, ...], str],
     dtype: Union[np.dtype, type, str, bigint] = float64,
     max_bits: Optional[int] = None,
-) -> pdarray:
+) -> Union[pdarray, Strings]:
     """
     Create a pdarray filled with ones.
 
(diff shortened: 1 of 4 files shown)


arkouda/numpy/pdarraycreation.py
```diff
@@ -757,11 +757,53 @@ def ones(
     return full(size=size, fill_value=1, dtype=dtype, max_bits=max_bits)
 
 
+SizeArg = Union[int_scalars, Tuple[int_scalars, ...], str]
+NumericFill = Union[numeric_scalars, np.bool]
+FillValue = Union[NumericFill, str]
+NumericDTypeArg = Union[np.dtype, type, bigint]
```
ajpotts: Can you use the type aliases in `arkouda.numpy._typing` if possible?  If not, then add the new definitions there?

arkouda/numpy/pdarraycreation.py
```diff
@@ -757,11 +757,53 @@ def ones(
...
+
+
+@overload
+def full(
+    size: SizeArg,
+    fill_value: str,
```
ajpotts: I would recommend using `StringDTypeTypes` instead of `str`, in case someone uses a `np.str_`.

Same for the other ones.

arkouda/numpy/pdarraycreation.py
```diff
@@ -813,32 +856,44 @@ def full(
...
+        raise TypeError(f"Unsupported dtype {dtype!r}")
+
+    #  The only string cases.
+
+    if dtype == Strings or akdtype(dtype) == str_:
+        if ndim != 1:
+            raise ValueError(f"Size parameter {size} incompatible with Strings.")
+        else:
+            if isinstance(fill_value, float):  # needed to match numpy, which
+                fill_value = int(fill_value)  # truncates floats to ints before "str"
```
ajpotts: This line leads to this error:
```
In [8]: ak.full(n, np.nan, dtype=ak.str_)
Cell In[8], line 1
----> 1 ak.full(n, np.nan, dtype=ak.str_)

File ~/anaconda3/envs/arkouda-dev/lib/python3.13/site-packages/typeguard/__init__.py:891, in typechecked.<locals>.wrapper(*args, **kwargs)
    889 memo = _CallMemo(python_func, _localns, args=args, kwargs=kwargs)
    890 check_argument_types(memo)
--> 891 retval = func(*args, **kwargs)
    892 check_return_type(retval, memo)
    894 # If a generator is returned, wrap it if its yield/send/return types can be checked

File ~/git/arkouda/arkouda/numpy/pdarraycreation.py:882, in full(size, fill_value, dtype, max_bits)
    880     else:
    881         if isinstance(fill_value, float):  # needed to match numpy, which
--> 882             fill_value = int(fill_value)  # truncates floats to ints before "str"
    883         return _full_string(full_size, str(fill_value))
    885 #  Remaining cases are all numeric.  Check dtype for error

ValueError: cannot convert float NaN to integer
```

Also consider what happens for `np.inf` and `-np.inf`.

arkouda/numpy/pdarraycreation.py
```diff
@@ -813,32 +856,44 @@ def full(
...
+            return _full_string(full_size, str(fill_value))
+
```
ajpotts: Perhaps this should go in a separate PR, but I'm getting this error:
```python
In [10]: ak.full(3, 1.9, dtype=ak.str_)
Out[10]: array(['1', '1', '1'])

```

ajpotts (approved):
Thank you!

arkouda/numpy/pdarraycreation.py
drculhane: I've removed these aliases.  They were meant to make the code clearer, but I don't think they were even doing that.

arkouda/numpy/pdarraycreation.py
drculhane: That breaks these overloads substantially.

Would you consider making this a separate issue? This file doesn't import StringDTypeTypes at all, and there are multiple functions that accept an argument of type "str".

arkouda/numpy/pdarraycreation.py
```diff
@@ -719,11 +721,47 @@ def ones(
...
+) -> Strings: ...
+
+
+@overload
+def full(
+    size: Union[int_scalars, Tuple[int_scalars, ...], str],
+    fill_value: Union[numeric_and_bool_scalars, str_scalars],
+    dtype: str,
+    max_bits: Optional[int] = ...,
+) -> Strings: ...
```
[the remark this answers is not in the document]
ajpotts: @1RyanK  is correct, the way it is written now, any dtype entered as a string, (i.e. "int64"), mypy will expect a `Strings` returned.  The correct way to do it is something like this:
```
from arkouda.numpy._typing._typing import  StringDTypeTypes,NumericDTypeTypes

@overload
def full(
    size: Union[int_scalars, Tuple[int_scalars, ...], str],
    fill_value: Union[numeric_and_bool_scalars, str_scalars],
    dtype:  StringDTypeTypes,
    max_bits: Optional[int] = ...,
) -> Strings: ...

```

I'm OK with fixing this in a separate PR, b/c as `ones` calls `full`, that may need to be overloaded first.

PR #5277, opened 2026-01-09
Simplify `deepcopy` for potential performance improvements

jabraham17:
Simplifies the `deepcopy` implementation to just use array assignment.

The previous implementation in the 1-node cases would result in fine-grained array accesses and in the multi-node case would do manual slicing/bulk copies. However, the builtin array assignment operator is already optimized for this. In the single node case for a trivial array it will result in a single `memcpy` and in the multi-node case it will result in a single memcpy per locale (assuming the same distribution).


drculhane (approved):
Simplicity is a beautiful thing.

PR #5296, opened 2026-01-10
Closes #5294:  alignment tests for arkouda.numpy.random.legacy

### LINKED ISSUE #5294: alignment tests for arkouda.numpy.random.legacy
Create alignment tests for `arkouda.numpy.random.legacy` module, comparing existing functions to the NumPy equivalents.

### LINKED ISSUE #5318: Legacy RNG global generator becomes stale across server restart, causing UnknownSymbolError in seed()
Problem:
CI fails with:
RuntimeError: Error executing command delGenerator: UnknownSymbolError ... unknown symbol

This occurs when ak.random.seed() attempts to delete the legacy global RNG generator using a symbol name that no longer exists on the server. The client-side global (theGenerator) persists across Arkouda server restarts / reconnects, but the server symbol table does not.

Trigger:
Recent changes to test ordering (adding random_legacy_alignment tests) expose this bug by causing ak.random.seed() to be called after a server reset.

Expected Behavior:
Calling ak.random.seed() should be robust even if a previous global generator was already deleted or lost due to a server restart.

Actual Behavior:
ak.random.seed() calls theGenerator.destructor(), which unconditionally sends delGenerator for a stale symbol name, causing CI failure.

Root Cause:
Legacy RNG uses a module-level global generator handle that is not reset when the Arkouda server restarts or the client reconnects.

Proposed Fixes (preferred order):
1. Make ak.random.seed() tolerant of missing generators by ignoring UnknownSymbolError during destructor.
2. Reset legacy RNG global state on Arkouda connect/disconnect.
3. (Less ideal) Test-side mitigation to enforce consistent server lifecycle.

Files Involved:
- arkouda/numpy/random/legacy.py
- arkouda/numpy/random/generator.py
- tests/numpy/random_test.py

Acceptance Criteria:
- ak.random.seed() no longer fails after server restarts.
- CI passes regardless of test ordering.
- No regression in RNG reproducibility tests.

ajpotts:
# Add NumPy-alignment tests for legacy random API (bool `randint` discrepancy tracked)

## Summary
This PR adds a NumPy-alignment test suite covering Arkouda’s **legacy random API**, with an emphasis on:
- shape semantics
- value bounds
- dtype behavior
- **error-message alignment** with NumPy’s legacy `RandomState`

A known discrepancy in `randint(dtype=bool)` error messages is intentionally captured and marked as `xfail`, so the suite documents the divergence without breaking CI.

## What’s included
- New test module:
  - `tests/numpy/alignment_verification/random_legacy_alignment.py`
- Added to `pytest.ini` test discovery
- Coverage for:
  - `ak.random.rand` (scalar and multi-dimensional shapes)
  - `ak.random.randint` (int64 bounds and shape semantics)
  - `ak.random.uniform`
  - `ak.random.standard_normal` (distribution sanity checks)
  - `ak.random.random`

## NumPy alignment strategy
For `randint(dtype=bool)`, expected error messages are derived dynamically from:
```python
numpy.random.RandomState.randint(..., dtype=bool)
```
This avoids hard-coding fragile strings and ensures the tests track NumPy’s actual behavior.

## Known discrepancy (xfail)
For `randint(dtype=bool)` with negative `low`:
- **NumPy** raises:
  ```
  ValueError: low is out of bounds for bool
  ```
- **Arkouda** currently raises:
  ```
  ValueError: low < 0
  ```

This mismatch is explicitly detected and marked with:
```python
pytest.xfail(... Issue #5295)
```
so CI remains green while preserving a clear signal that behavior differs from NumPy.

## Why this matters
- Establishes a reusable pattern for NumPy-alignment testing
- Makes random API differences explicit and actionable
- Prevents silent divergence in user-facing behavior

## Follow-up
Once Arkouda’s `randint(dtype=bool)` validation is updated to match NumPy’s error message, the `xfail` can be removed and the test will pass automatically.

Closes #5294:  alignment tests for arkouda.numpy.random.legacy
Closes #5318:  Legacy RNG global generator becomes stale across server restart, causing UnknownSymbolError in seed()


1RyanK (approved):
Looks good!

PR #5350, opened 2026-01-27
Closes #5349: Clean up indents in OperatorMsg.chpl and BinOp.chpl and remove unused data types

### LINKED ISSUE #5349: Clean up indents in `OperatorMsg.chpl` and `BinOp.chpl` and remove unused data types
**Feature Description**
These files have a mix of two space indents and four space indents. I'm transitioning them to entirely two space indents.

I had previously implemented what amounted to as much numpy-style dtype promotion as possible. However, our `registration-config.json` only uses `bool`, `uint(8)`, `uint(64)`, `int(64)`, `real(64)`, and `bigint`. Given that adding more data types would likely put even more pressure on compiler memory, I don't think it's realistic to support several more data types than we currently do. We can also always add the support back in if we get there. Hopefully this also allows me to do some cleanup of the functions, but for now, let's just remove references to these data types.

1RyanK:
Probably a lot of lines changed but most of it is fixing indents. Mostly just removed any references to unused data types.

Closes #5349: Clean up indents in OperatorMsg.chpl and BinOp.chpl and remove unused data types

### DIFF
--- a/src/BinOp.chpl
+++ b/src/BinOp.chpl
@@ -135,92 +135,57 @@ module BinOp
   inline proc isSmallOp(op: Operator): bool { return smallOpSet.contains(op); }
 
   proc splitType(type dtype) param : int {
-      // 0 -> bool, 1 -> uint, 2 -> int, 3 -> real
+    // 0 -> bool, 1 -> uint, 2 -> int, 3 -> real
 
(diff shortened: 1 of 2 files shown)


ajpotts:
Closing b/c I think we need more discussing before removing any dtypes.

PR #5354, opened 2026-01-27
Closes #5353:  Fix Index.to_pandas string/object handling to use pandas "str" dtype when appropriate

### LINKED ISSUE #5353: Fix `Index.to_pandas` string/object handling to use pandas `"str"` dtype when appropriate
## Description
This ticket tracks a change to `Index.to_pandas` that improves compatibility with pandas expectations for string-like indexes, particularly around missing values.

Previously, `Index.to_pandas` always constructed a `pd.Index` using the NumPy dtype inferred from the underlying array. For object, unicode, or bytes dtypes (`dtype.kind in {"O", "U", "S"}`), this could result in a pandas `Index` that does **not** match pandas’ `"string"` semantics, especially with respect to `NA` handling.

The new logic detects when the converted NumPy array:
- Has an object / unicode / bytes dtype, **and**
- Is empty **or**
- Contains only `str` values or `None`

In those cases, `Index.to_pandas` now constructs the result via:

```python
pd.Index(pd.array(val, dtype="str"), name=self.name)

ajpotts:
## Summary
This PR refactors `Index.to_pandas()` to better align Arkouda’s pandas interop with pandas’ own dtype semantics, while avoiding unintended dtype inference and unsupported server-side paths. It also modernizes and strengthens the associated unit tests.

### `Index.to_pandas()` behavior
- **Explicit dtype preservation for stable dtypes**  
  Numeric, boolean, datetime, and timedelta data now round-trip to pandas with their original dtype explicitly preserved, avoiding pandas re-inference.

- **Improved string handling**
  - Unicode string data is converted to pandas’ `string` dtype to match pandas NA semantics.
  - Arkouda `Strings` objects are materialized as pandas string dtype indexes.
  - No implicit decoding or coercion of bytes is attempted.

- **Categorical handling**
  - Arkouda categoricals are converted to `CategoricalIndex`, preserving categories, codes, and name.
  - If `Categorical.to_pandas()` already returns an Index, it is reused directly.

- **Cleaner materialization path**
  - Centralized logic for converting Arkouda-backed values into concrete NumPy arrays before pandas construction.

### Tests
- Replaced a monolithic `test_to_pandas` with smaller, focused tests:
  - Dtype preservation for int, float, and bool
  - String conversion for Python lists and Arkouda `Strings`
  - Rejection of mixed-type lists when `allow_list=True`
  - Empty index handling
  - Categorical round-trip correctness
  - MultiIndex correctness, including categorical levels and value round-trips
- Tests now reflect *supported* Arkouda construction paths and avoid unreachable object/bytes cases.

### Docs & cleanup
- Expanded `Index.to_pandas()` docstring with clearer semantics and examples.
- Fixed doctest import path for pandas index module.
- Minor test cleanup (imports, readability).

## Motivation
Previously, `to_pandas()` always passed `dtype=val.dtype`, which could:
- Force inappropriate pandas dtypes
- Mask pandas’ native string handling
- Lead to brittle behavior across pandas versions

This refactor makes dtype behavior explicit where it matters, defers to pandas where appropriate, and ensures tests match real Arkouda usage.

## Backward Compatibility
- No change to public APIs.
- Behavior changes are limited to pandas materialization and are covered by explicit tests.

Closes #5353:  Fix Index.to_pandas string/object handling to use pandas "str" dtype when appropriate

### DIFF
--- a/arkouda/pandas/index.py
+++ b/arkouda/pandas/index.py
@@ -795,15 +795,12 @@ class Index:
         Examples
         --------
         >>> import arkouda as ak
-        >>> import pandas as pd
+        >>> import pandas
         >>> idx = ak.Index(ak.array([1,2,3]))
         >>> pidx = idx.to_pandas()
         >>> pidx.dtype
         dtype('<i8')
         """
-        import numpy as np
-        import pandas as pd
-
         from arkouda.pandas.categorical import Categorical
 
         def _materialize(values):
@@ -836,7 +833,7 @@ class Index:
         if kind in ("b", "i", "u", "f", "c", "M", "m"):
             return pd.Index(val, dtype=dtype, name=self.name)
 
-        # 3) Fallback: let pandas decide (covers unusual/extension-ish cases).
+        # 4) Fallback: let pandas decide (covers unusual/extension-ish cases).
         return pd.Index(val, name=self.name)
 
     def to_ndarray(self):


jaketrookman (approved):
Looks good

PR #5377, opened 2026-02-02
Fixes bug 5298, so that generator.integers gives an error if asked for float

### LINKED ISSUE #5298: Bug: Generator.integers does not reject float dtypes
# Bug: Generator.integers does not reject float dtypes

## Summary
`ak.random.Generator.integers` fails to raise a `TypeError` when a floating-point dtype (e.g. `ak.float64`) is passed via the `dtype` argument. This diverges from NumPy behavior and allows invalid calls to proceed silently.

## Expected Behavior
Passing a floating-point dtype to `integers` should raise a `TypeError`, consistent with NumPy:

```python
import numpy as np

rng = np.random.default_rng(123)
rng.integers(0, 10, size=10, dtype=np.float64)  # TypeError
```

Arkouda should match this behavior.

## Actual Behavior
The following call succeeds without raising:

```python
import arkouda as ak

rng = ak.random.default_rng(123)
rng.integers(0, 10, size=10, dtype=ak.float64)  # no error
```

## Root Cause
Inside `Generator.integers`, the `dtype` argument is normalized using `to_numpy_dtype(dtype)`, producing a `numpy.dtype` object. The subsequent guard uses an identity check:

```python
if dtype is akfloat64:
    raise TypeError(...)
```

After normalization, this identity check fails even when the dtype represents float64, so the error is never raised.

## Proposed Fix
Replace the identity check with a semantic dtype check. For example:

```python
if dtype.kind == "f":
    raise TypeError(f"Unsupported dtype {dtype!r} for integers")
```

This correctly rejects all floating-point dtypes (`float32`, `float64`, etc.) and matches NumPy semantics.

## Regression Test
```python
def test_integers_rejects_float64_dtype():
    rng = ak.random.default_rng(123)
        rng.integers(0, 10, size=10, dtype=ak.float64)
```

## Impact
- Brings Arkouda RNG API into alignment with NumPy
- Prevents silent misuse of `integers` with invalid dtypes
- Improves correctness and test coverage in alignment test suite

drculhane:
Closes #5298

This is pretty simple.  It implements the fix suggested in the bug, with one small change that the test function requires a (self) argument.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -395,7 +395,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        chpl-version: ['2.4.0','2.5.0','2.6.0','2.7.0']
+        chpl-version: ['2.5.0','2.6.0','2.7.0']
     container:
       image: ghcr.io/bears-r-us/ubuntu-with-arkouda-deps-chpl-${{ matrix.chpl-version }}:latest
       credentials:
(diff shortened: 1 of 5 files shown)


ajpotts (approved):
Thanks for fixing this!  Great work!

PR #5378, opened 2026-02-02
Adds special cases to ak.linspace

### LINKED ISSUE #5267: Bug: `ak.linspace` Fails for `num == 1` (Scalar Case)
# Bug: `ak.linspace` Fails for `num == 1` (Scalar Case)

## Summary
`ak.linspace` raises a `ZeroDivisionError` when called with `num == 1` and scalar `start`/`stop` values.  
NumPy defines this as a valid case and returns a length‑1 array containing `start`.

This is a behavior mismatch with NumPy and breaks NumPy‑alignment tests.

### NumPy
```python
import numpy as np
np.linspace(0, 1, 1, endpoint=True)
# array([0.])
```

### Arkouda
```python
import arkouda as ak
ak.linspace(0, 1, 1, endpoint=True)
# ZeroDivisionError
```

## Root Cause
In the scalar path, Arkouda computes:

```python
divisor = num - 1 if endpoint else num
delta = (stop - start) / divisor
```

When `num == 1` and `endpoint == True`, `divisor == 0`, leading to a divide‑by‑zero.

NumPy treats `num == 1` as a special case and does **not** compute a step size.

## Expected Behavior (NumPy‑Compatible)
- `num == 0` → empty array
- `num == 1` → array containing `start`
- Works consistently for:
  - scalar `start` / `stop`
  - vector / broadcasted `start` / `stop`
  - both `endpoint=True` and `endpoint=False`

## Scope
- `arkouda.numpy.pdarraycreation.linspace`
- Scalar and vector code paths

## Proposed Fix
Add an early return before computing `divisor`:

```python
if num == 0:
    return ak.array([], dtype=ak.float64)

if num == 1:
    if isinstance(start_, pdarray):
        return start_.reshape((1,) + start_.shape)
    else:
        return ak.array([start_], dtype=ak.float64)
```

## Test Plan
- Convert NumPy‑alignment test for `num == 1` from `xfail` to passing
- Add explicit tests:
  - `ak.linspace(0, 1, 1)`
  - `ak.linspace(0, 1, 1, endpoint=False)`
  - Vector start/stop with `num == 1`

## Acceptance Criteria
- `ak.linspace(..., num=1)` matches NumPy output
- No `ZeroDivisionError`
- Alignment tests pass without xfail

drculhane:
Closes #5267 

The proposed fix required slight editing: changing uses of ak.float64 to just float64 (since pdarraycreation.py already imports it by that name), and changing uses of ak.array to just array (since that function is within this file).

Also added a unit test for these cases.  It isn't parameterized by data type, which is a judgment call, since imo, that wouldn't add anything.  But it would be simple to add, if the team so chooses.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -395,7 +395,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        chpl-version: ['2.4.0','2.5.0','2.6.0','2.7.0']
+        chpl-version: ['2.5.0','2.6.0','2.7.0']
     container:
       image: ghcr.io/bears-r-us/ubuntu-with-arkouda-deps-chpl-${{ matrix.chpl-version }}:latest
       credentials:
@@ -436,7 +436,7 @@ jobs:
       ARKOUDA_CONFIG_FILE: .configs/MultiDimTestServerModules.cfg
     strategy:
       matrix:
-        chpl-version: ['2.4.0','2.5.0','2.6.0','2.7.0']
+        chpl-version: ['2.7.0']
       max-parallel: 3
     container:
       image: ghcr.io/bears-r-us/ubuntu-with-arkouda-deps-chpl-${{ matrix.chpl-version }}:latest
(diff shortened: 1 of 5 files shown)


ajpotts (approved):
Thanks for fixing this one!

PR #5400, opened 2026-02-11
More benchmark v2 fixes

jabraham17:
More fixes to get benchmark_v2 working

In this PR
* includes https://github.com/Bears-R-Us/arkouda/pull/5370. That can be merged separately and this PR can then be rebased
* uses far less data for the array transfer benchmark. this matches what the old array transfer does
* uses a global server by default for running benchmarks. this matches what the old benchmark runner does
* implement an env var to control numtrials (which for some reason can't be set via the command line)

### DIFF
--- a/benchmark_v2/array_transfer_benchmark.py
+++ b/benchmark_v2/array_transfer_benchmark.py
@@ -2,8 +2,6 @@ import gc
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


ajpotts (approved):
Thanks!

PR #5416, opened 2026-02-18
Closes #5415:  Performance regresion from divide by zero check

### LINKED ISSUE #5415: Performance regresion from divide by zero check
The chapel team reported a performance regression in the benchmarks from PR https://github.com/Bears-R-Us/arkouda/pull/5315, which appears to be due to the check always being executed, even when the error mode is "ignore".

ajpotts:
## Summary
This PR fixes a performance regression introduced by divide-by-zero handling in `pdarray` binary operations. The previous implementation eagerly scanned denominators (e.g., `any(x2 == 0)`) on every vector/vector divide and reverse-divide, adding an extra full-array pass (and typically allocating a temporary boolean array) even when the divide error mode is set to `ignore`.

## Problem
- Divide-by-zero checks were performed **eagerly** at the Python level for array denominators.
- In the common/default configuration (`divide="ignore"`), these checks should be a no-op, but the eager scan still occurred, causing unnecessary work and a measurable slowdown.

## Solution
- Rewrite `_dbz_check` to:
  - **Return early** when `divide` mode is `ignore` (fast path).
  - Accept either a boolean or a **callable** (lazy evaluation) so expensive checks are only computed when needed.
  - Use centralized error handling via `arkouda.numpy.err.handle` to support all configured behaviors (`warn`, `raise`, `call`, `print`, `log`) consistently.

## Key changes
- `arkouda/numpy/pdarrayclass.py`
  - Replace eager `_dbz_check(op, flag)` with lazy-capable `_dbz_check(op: str, has_zero)`.
  - Remove direct `warnings` usage and dispatch via `akerr.handle("divide", msg)`.
  - Update call site for vv division to pass `lambda: any(x2 == 0)` so the scan only runs when mode != `ignore`.

## Why this is safe
- When `divide="ignore"`: behavior remains a no-op, now without extra scanning.
- When `divide!="ignore"`: behavior is preserved (and broadened) through `err.handle`, which is the standard place for error-mode semantics.
- The lazy callable is only invoked for division operations (`/` and `//`) and only when a check is required.

## Testing
- Existing test suite should cover divide behavior; this change is primarily performance-oriented.

## Notes
This PR intentionally avoids backend changes. The improvement comes from eliminating unnecessary Python-level scans in the common `ignore` mode while preserving correct behavior for `warn`/`raise` and other configured error actions.

Closes #5415:  Performance regresion from divide by zero check

### DIFF
--- a/arkouda/numpy/dtypes.py
+++ b/arkouda/numpy/dtypes.py
@@ -7,6 +7,7 @@ import warnings
 from enum import Enum
 from typing import (  # noqa: F401
     Literal,
+    Optional,
     TypeAlias,
     TypeGuard,
     Union,
@@ -70,6 +71,8 @@ from numpy.dtypes import (
 
 from ._bigint import bigint, bigint_
 
+
+CastingKind: TypeAlias = Literal["no", "equiv", "safe", "same_kind", "unsafe"]
 
 __all__ = [
     "_datatype_check",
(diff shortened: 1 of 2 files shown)


arkouda/numpy/pdarrayclass.py
```diff
@@ -224,26 +224,40 @@ def stats_reduce(
...
+        Either a precomputed flag, or a callable that computes whether the
+        denominator contains any zeros. Use a callable to avoid scanning
+        when divide mode is 'ignore'.
+    """
+    if op not in ("/", "//"):
         return
 
-    mode = ak_geterr()["divide"]
+    # Fast path: avoid any extra work for the common/default mode.
+    mode = ak_geterr().get("divide", "ignore")
```
e-kayrakli: Is there a way to test `divide!=ignore` today in today's test infra?

PR #5435, opened 2026-02-24
Closes #5428:  BUG: Pandas Series with dtype=ak Returns NumPy Index Values

### LINKED ISSUE #5428: BUG: Pandas Series with dtype="ak" Returns NumPy Index Values
## Description

When creating a pandas `Series` backed by the Arkouda extension dtype
(`dtype="ak"`), the index values are returned as a NumPy array instead
of an Arkouda-backed structure.

### Reproducer

``` python
>>> import arkouda as ak
>>> import pandas as pd

In [6]: s = pd.Series(["a","b","c"], dtype="ak")

In [7]: s.index.values
Out[7]: array([0, 1, 2])

In [8]: type(s.index.values)
Out[8]: numpy.ndarray
```

## Expected Behavior

For consistency with Arkouda-backed data structures, the index values
should be represented using an Arkouda-backed index (e.g., an
ArkoudaArray or Arkouda Index type), not materialized as a NumPy
`ndarray`.

Returning a NumPy array:

-   Forces client-side materialization
-   Breaks scalability expectations
-   Is inconsistent with Arkouda-backed column behavior

## Actual Behavior

`s.index.values` returns:

``` python
numpy.ndarray
```

## Impact

-   Violates scalability guarantees
-   Introduces silent data transfer to the client
-   Creates inconsistent behavior between data and index representations

## Proposed Fix

Ensure that Series created with `dtype="ak"` use an Arkouda-backed index
type by default, or override the `.values` accessor to return an Arkouda
structure instead of a NumPy array.

## Environment

-   Arkouda pandas extension
-   pandas Series with dtype="ak"

ajpotts:
## Summary

This PR improves Arkouda’s pandas-extension interoperability in two areas:

1. **RangeIndex conversion is now server-side.**  
   `_pandas_index_to_ak` detects `pandas.RangeIndex` and constructs the corresponding index directly on the Arkouda server via `ak_arange`, avoiding client-side materialization for very large ranges.

2. **`Series.ak.is_arkouda` is now “fully Arkouda-backed”.**  
   The predicate now returns `True` only when **both**:
   - the Series values are stored in an `ArkoudaExtensionArray`, and
   - the Series index is Arkouda-backed (and for a `MultiIndex`, **all levels** are Arkouda-backed).

## Motivation

- **Scalability / safety:** `RangeIndex` is “virtual” (start/stop/step). Any conversion path that materializes its values on the client can blow up memory for huge ranges. Creating the range directly on the server avoids this footgun.
- **Correctness / clarity:** Previously, `Series.ak.is_arkouda` could return misleading results by relying on `.index.values`, which is frequently a NumPy array even when the index is conceptually virtual or otherwise not truly Arkouda-backed. The updated logic checks `idx.array` (and MultiIndex levels) and better matches the meaning of “Arkouda-backed.”

### `_pandas_index_to_ak`
- `pd.RangeIndex` is converted via server-side `ak_arange(...)` and wrapped in `ak.Index`.
- Defensive check: a `RangeIndex` with `step == 0` raises `ValueError`.

### `Series.ak.is_arkouda`
- Now returns `True` only when both data and index are Arkouda-backed.
- For `pd.MultiIndex`, requires each level to be Arkouda-backed.

## Implementation notes

- `_pandas_index_to_ak`:
  - handles `pd.MultiIndex` as before (delegates to `ak.MultiIndex`).
  - special-cases `pd.RangeIndex` to build values server-side using `ak_arange(stop)` for the common `(start=0, step=1)` case, otherwise `ak_arange(start, stop, step)`.

- `Series.ak.is_arkouda`:
  - early-exits `False` if the Series values are not an `ArkoudaExtensionArray`.
  - checks index backing via `.array` (and each MultiIndex level `.array`) using a defensive `getattr(..., None)`.

### Index accessor tests (`tests/pandas/extension/index_accessor.py`)
- `test_pandas_index_to_ak_rangeindex_default_roundtrips`
- `test_pandas_index_to_ak_rangeindex_nontrivial_roundtrips`
- `test_pandas_index_to_ak_rangeindex_negative_step_roundtrips_or_errors`

### Series accessor tests (`tests/pandas/extension/series_accessor.py`)
- `test_is_arkouda_true_for_to_ak_series_with_default_rangeindex`
- `test_is_arkouda_true_for_to_ak_series_with_nontrivial_index`
- `test_is_arkouda_true_for_to_ak_series_with_multiindex`

## How to validate

- Run the pandas-extension test suite (or at least the two modified test modules):
  - `pytest -q tests/pandas/extension/index_accessor.py`
  - `pytest -q tests/pandas/extension/series_accessor.py`

## Notes / follow-ups

- The negative-step RangeIndex test is intentionally permissive: if the backend does not support negative-step `arange`, the test accepts a clear exception. If negative steps *should* be supported, we can tighten the expectation to require round-tripping.

Closes #5428:  BUG: Pandas Series with dtype=ak Returns NumPy Index Values

### DIFF
--- a/arkouda/pandas/extension/_index_accessor.py
+++ b/arkouda/pandas/extension/_index_accessor.py
@@ -86,12 +86,32 @@ def _pandas_index_to_ak(index: Union[pd.Index, pd.MultiIndex]) -> Union[ak_Index
     Returns
     -------
     Union[ak_Index, ak_MultiIndex]
+
+    Raises
+    ------
+    ValueError
+        If a pandas RangeIndex has an invalid step (e.g., step == 0).
     """
+    from arkouda.numpy.pdarraycreation import arange as ak_arange
+
     # MultiIndex: delegate directly to ak.MultiIndex, which handles
     # pandas.MultiIndex in its constructor.
     if isinstance(index, pd.MultiIndex):
         # Preserve names when constructing the ak.MultiIndex
         return ak_MultiIndex(index, names=index.names)
+
+    # IMPORTANT: RangeIndex is virtual (start/stop/step). Converting it via any path
+    # that materializes values can blow up the client for huge sizes.
+    # Create it directly on the Arkouda server instead.
+    if isinstance(index, pd.RangeIndex):
+        start, stop, step = index.start, index.stop, index.step
+        if step == 0:
+            raise ValueError("RangeIndex step cannot be 0")
+        if start == 0 and step == 1:
+            arr = ak_arange(stop)
+        else:
+            arr = ak_arange(start, stop, step)
+        return ak_Index(arr, name=index.name)
 
     # Single-level Index: ak.Index already knows how to consume pandas.Index,
     # including CategoricalIndex.
--- a/arkouda/pandas/extension/_series_accessor.py
+++ b/arkouda/pandas/extension/_series_accessor.py
@@ -354,14 +354,18 @@ class ArkoudaSeriesAccessor:
     @property
     def is_arkouda(self) -> bool:
         """
-        Return whether the underlying Series is Arkouda-backed.
+        Return True if this Series is fully Arkouda-backed.
 
-        A Series is Arkouda-backed if its underlying storage uses
-        :class:`ArkoudaExtensionArray`.
+        A Series is considered Arkouda-backed when both:
+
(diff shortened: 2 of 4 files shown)


jaketrookman (approved):
Looks good

PR #5486, opened 2026-04-17
Make the loop iterating over multiple parquet files parallel

jabraham17:
This makes the loop over multiple parquet files parallel

This PR is the same as https://github.com/Bears-R-Us/arkouda/pull/4952, except it fixes the bug exposed by that PR. `CPtrsToData` and `CPtrsToWhereNulls` need to be local for each parallel task, so they must be declared inside of the forall loop. Technically, they will be instantiated more than they really need to be, but it is correct.

Resolves https://github.com/Bears-R-Us/arkouda/pull/4952

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -1281,20 +1281,24 @@ module ParquetMsg {
 
       var subdoms = getSubdomains(this.sizes);
 
+      pqLogger.debug(getM(), getR(), getL(),
+        "readInto with e.domain=%?, ty=%?, offsets=%?, subdoms=%?, filenames=%?"
+        .format(e.domain, this.ty, this.offsets, subdoms, this.filenames));
+
       coforall loc in Locales with (ref this) do on loc {
         const LocTypes = this.ty;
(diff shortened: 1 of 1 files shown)


benharsh (approved):

bradcray (approved):

bradcray:
@jabraham17 : Belated thought: Can/should this be propagated to the standalone Parquet package?  (Am I correct in remembering that it lives here: https://github.com/e-kayrakli/Parquet ?).

jabraham17:
> Belated thought: Can/should this be propagated to the standalone Parquet package? 

No idea. ParquetMsg and the standalone Parquet module have diverged enough that its not immediately obvious. Note the Parquet mason package does not even use c_readAllCols, so may not suffer from this issue at all.

bradcray:
OK, thanks.  Tagging @e-kayrakli and @benharsh to see if they know whether it has alternate/equivalent functionality or whether it's focused only on features of use to FastOTF2 so is missing other features we'd want for Arkouda.

benharsh:
Right now the package is focused on features to be used in FastOTF2, and is not at a point where we could replace the Arkouda implementation.

PR #5495, opened 2026-06-09
Update `.gitignore` to exclude `.akdata/` and `*.tmp`

eric-vo:
Remove the temp files accidentally added under `tests/pandas/.akdata` in #5444, and exclude `.akdata` and `*tmp` in .gitignore to prevent new temp files from being committed.


bradcray (approved):
This looks good to me, thanks @eric-vo !

PR #5505, opened 2026-07-29
Bundle of performance improvements

jabraham17:
A bundle of performance improvements motivated by analytics workflow

* improved the parallelization of `diff`, 1.5x speedup on micro benchmark using a COMM=none server on my mac.
    * it looks like previous refactors accidentally removed the parallelization, so I added it back
    * I also found that the outer loop is usually a single trip count, so that being parallel is not very useful. the inner loop being parallel is much faster (at least in my benchmarks)
* improved the performance of array slices, up to a 2x speedup on a micro benchmark using a COMM=none server on my mac
    * by specializing for non-strided ranges and using whole-array operations, we take advantage of builtin chapel optimizations for slicing

### DIFF
--- a/src/IndexingMsg.chpl
+++ b/src/IndexingMsg.chpl
@@ -130,22 +130,56 @@ module IndexingMsg
     }
 
     @arkouda.registerCommand("[slice]")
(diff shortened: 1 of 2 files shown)


jabraham17:
I split this PR into https://github.com/Bears-R-Us/arkouda/pull/5508 and https://github.com/Bears-R-Us/arkouda/pull/5509

PR #5506, opened 2026-07-30
Implement `tag_data` for `ak.read_parquet` on the whole data set

jabraham17:
Implements `tag_data` for `ak.read_parquet` for the whole dataset, which was missing before

I also added a whole bunch of tests for `tag_data`, since there were none before

### DIFF
--- a/src/ParquetMsg.chpl
+++ b/src/ParquetMsg.chpl
@@ -1030,19 +1030,21 @@ module ParquetMsg {
       pqErr.parquetError(getL(), getR(), getM());
     }
 
-    pqLogger.debug(getM(),getR(),getL(), "will try tagging data");
-
-    // If tagging is turned on, tag the data
-    if tagData {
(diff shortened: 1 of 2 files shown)


src/ParquetMsg.chpl
```diff
@@ -1030,19 +1030,21 @@ module ParquetMsg {
       pqErr.parquetError(getL(), getR(), getM());
     }
 
-    pqLogger.debug(getM(),getR(),getL(), "will try tagging data");
-
-    // If tagging is turned on, tag the data
-    if tagData {
-      throw new NotImplementedError("Reading all columns while tagging " +
```
benharsh, line 1037: Should we recreate this error in the non-optimizable case?

jabraham17: No need. In the non-optimized case we fall back to an implementation that tags the data

benharsh (approved):

PR #5510, opened 2026-07-31
Migrate Parquet support to Mason package

eric-vo:
Replaces Arkouda's Parquet implementation with the external Mason `Parquet` package.

Arkouda now retains only server-specific responsibilities: message handling, symbol-table conversion, client responses, and Arkouda-specific null processing. File, schema, reader, writer, and distributed Parquet mechanics are delegated to the package.

## Changes

- Remove Arkouda's vendored Parquet C++ sources and legacy Arrow build shims.
- Add package acquisition and build integration.
- Re-export the package API through `ParquetMsg` for existing server modules.
- Preserve distributed, string, list, multi-column, append, and null-handling behavior.
- Read the runtime Arrow version through the package's C++ API.
- Store checkpoint `uint8` arrays directly without temporary widening.
- Add `uint8` support to `ak.zeros`, including the `"uint8"` dtype string.
- Make embedded Python interpreter initialization idempotent after testing exposed repeated initialization in CLIENT mode.
- Expand checkpoint and Parquet round-trip coverage.

## Dependency

This PR depends on the companion Mason Parquet PR and should be merged after it.


arkouda/numpy/pdarraycreation.py
```diff
@@ -762,7 +762,7 @@ def bigint_from_uint_arrays(arrays, max_bits=-1):
 @overload
 def zeros(
     size: Union[int_scalars, Tuple[int_scalars, ...], str],
-    dtype: Union[NumericDTypeTypes, type[bigint]] = ...,
+    dtype: Union[NumericDTypeTypes, type[np.uint8], type[bigint]] = ...,
```
jabraham17: What motivated this change? `NumericDTypeTypes` should already include `uint8`

make/deps/parquet.mk
```diff
@@ -0,0 +1,15 @@
...
+# module source path. We use this instead of `mason modules`, which does not
+# work reliably with Chapel 2.4. Override PARQUET_INSTALL_DIR to relocate the
+# clone; set ARKOUDA_PARQUET_SRC_DIR to reuse an existing checkout.
+PARQUET_PACKAGE_SCRIPT := $(ARKOUDA_PROJECT_DIR)/scripts/get_parquet_package.sh
+PARQUET_INSTALL_DIR ?= $(DEP_BUILD_DIR)/Parquet
+
+# Keep this as a shell command rather than Make's `$(shell ...)`: Make discards
+# the command's exit status, which would let a failed package build continue
+# into a misleading Chapel compile failure.
+PARQUET_PACKAGE_FLAGS_CMD = env CHPL_HOME="$(ARKOUDA_CHPL_HOME)" \
```
jabraham17: unnecessary comment, or at least its overly verbose

scripts/get_parquet_package.sh
```diff
@@ -0,0 +1,94 @@
...
+    log "ERROR: CHPL_HOME is unset and chpl is not available on PATH"
+    exit 1
+  fi
+  CHPL_HOME="$(chpl --print-chpl-home)"
+fi
+
+if [[ ! -x "${CHPL_HOME}/util/config/compileline" ]]; then
+  log "ERROR: invalid CHPL_HOME: ${CHPL_HOME}"
+  exit 1
+fi
```
jabraham17: This script should just take the already-computed `ARKOUDA_CHPL_HOME` as argument

scripts/get_parquet_package.sh
```diff
@@ -0,0 +1,94 @@
...
+# Gather the chpl flags the package needs.
+RAW_FLAGS="$(make -s -C "${PREREQ_DIR}" ARKOUDA_CHPL_HOME="${CHPL_HOME:-}" printchplflags)"
+if [[ -z "${RAW_FLAGS}" ]]; then
+  log "ERROR: the Parquet prerequisite build returned no Chapel flags"
+  exit 1
+fi
+
+# `printchplflags` emits `-I./src/util/` relative to the package root; rewrite it
+# to an absolute path so chpl (invoked from the Arkouda root) resolves it.
+FLAGS="$(printf '%s' "${RAW_FLAGS}" | sed "s|-I\./|-I${PARQUET_ROOT}/|g")"
```
jabraham17: This is something that should be fix in `printchplflags` of Parquet, not worked around here

scripts/get_parquet_package.sh
```diff
@@ -0,0 +1,94 @@
...
+
+# Emit the flags followed by the Parquet module source. This is the only line
+# written to stdout.
+printf '%s %s\n' "${FLAGS}" "${PARQUET_MODULE}"
```
jabraham17, line 74: overly commented

scripts/get_parquet_package.sh
```diff
@@ -0,0 +1,94 @@
...
+  log "ERROR: expected C++ prerequisites at ${PREREQ_DIR}, but they are missing"
+  exit 1
+fi
+
+if [[ ! -f "${PARQUET_MODULE}" ]]; then
+  log "ERROR: expected Chapel module at ${PARQUET_MODULE}, but it is missing"
+  exit 1
+fi
+
+# Build the C++ prerequisite objects (no-op if already up to date).
```
jabraham17: overly commented

scripts/get_parquet_package.sh
jabraham17, line 74: this looks overly complex. whats wrong with `echo "${FLAGS} ${PARQUET_MODULE}"`?

src/ApplyMsg.chpl
```diff
@@ -28,49 +26,78 @@ module ApplyMsg {
...
+      applyLogger.debug(
+        getModuleName(),
+        getRoutineName(),
+        getLineNumber(),
+        "Initializing Python interpreters");
+
+      forall interp in interpreters do
+        if interp == nil then
+          interp = new Interpreter();
+    }
```
jabraham17: Whats the reason for this change?

Regardless, this should go in its own PR to keep this a focused improvement

src/CheckpointMsg.chpl
```diff
@@ -168,21 +168,29 @@ module CheckpointMsg {
...
 
-      st.addEntry(entry.name, entry);
-      cpLogger.debug(M(), R(), L(), "Added entry with metadata ", mdName);
+          st.addEntry(entry.name, entry);
+          cpLogger.debug(M(), R(), L(), "Added entry with metadata ", mdName);
+        }
+        otherwise do
+          cpLogger.debug(M(), R(), L(), cpNotImplementedMsg("Loading",
+            entryMD.entryName, entryType:string, mdName));
+      }
```
jabraham17: Whats the reason for this change?

tests/numpy/pdarray_creation_test.py
```diff
@@ -585,12 +585,20 @@ def test_uniform(self, size):
...
+    @pytest.mark.parametrize("dtype", [ak.int64, ak.uint8, float, ak.float64, bool, ak.bool_, ak.bigint])
     def test_zeros_dtype(self, size, dtype):
         zeros = ak.zeros(size, dtype)
         assert isinstance(zeros, ak.pdarray)
         assert dtype == zeros.dtype
-        assert (0 == zeros).all()
+        if dtype == ak.uint8:
+            assert zeros.tolist() == [0] * size
+        else:
+            assert (0 == zeros).all()
```
jabraham17: why is the conditional needed?

tests/numpy/pdarray_creation_test.py
```diff
@@ -623,9 +631,6 @@ def test_zeros_misc(self):
         zeros = ak.ones("5")
         assert 5 == len(zeros)
 
-        with pytest.raises(TypeError):
-            ak.zeros(5, dtype=ak.uint8)
-
         with pytest.raises(TypeError):
             ak.zeros(5, dtype=str)
```
jabraham17, line 630: why is this removed?

tests/pandas/io_test.py
jabraham17, line 904: I take it you have now implemented delta encoding? Nice!

tests/pandas/parquet_edge_test.py
jabraham17, line 120: If the test suite is run with Arrow v9, this will flag as a failure, correct?

jabraham17 (commented):
Review part 1, I have not yet looked at the Parquet server code changes

arkouda/numpy/pdarraycreation.py
eric-vo: In `tests/numpy/pdarray_creation_test.py`, I passed in `ak.uint8` to `ak.zeros()` to be consistent with the other types instead of `"uint8"` as a string literal, and `NumericDTypeTypes` doesn't look like it has it yet. Should I add `type[ak_uint8]` to `ArkoudaNumericTypes` so I don't have to do this workaround?

jabraham17: Yes that would be a better change, but has a bigger scope and should then go in its own PR

scripts/get_parquet_package.sh
eric-vo: So no need for validation?

scripts/get_parquet_package.sh
eric-vo: `echo` apparently parses special characters weirdly, so `printf` could be [a slightly safer option](https://unix.stackexchange.com/a/58358)? Seems like that's not such a big deal for our case, so should I switch to `echo`?

jabraham17: no need, the printf is fine. I just want to understand the rationale

But the comment is superfluous and should be removed

scripts/get_parquet_package.sh
jabraham17: This script is intended to only be called by the Arkouda Makefile, assume if `ARKOUDA_CHPL_HOME` is set that its set correctly.

src/ParquetMsg.chpl
```diff
@@ -11,149 +11,53 @@ module ParquetMsg {
...
-    ZSTD=4,
-    LZ4=5
-  };
-
+  // The core Parquet I/O primitives now live in the standalone Mason `Parquet`
+  // package. ParquetMsg delegates to that API and only retains Arkouda-server
+  // specific plumbing (the message handlers, the optimized read-all path, and
+  // the multi-column writer) plus the few helpers Mason does not provide.
+  // These symbols are re-exported so existing importers (e.g. CheckpointMsg)
+  // keep resolving them through ParquetMsg.
```
jabraham17: unnecessary comment

src/ParquetMsg.chpl
```diff
@@ -11,149 +11,53 @@ module ParquetMsg {
...
                      getLineNumber as getL};
-  require "ReadParquet.h";
-  require "ReadParquet.o";
-  require "WriteParquet.h";
-  require "WriteParquet.o";
-  require "UtilParquet.h";
-  require "UtilParquet.o";
+  // The C++ Parquet prerequisites (headers and objects) are supplied on the
+  // `chpl` command line by scripts/get_parquet_package.sh via the Makefile, so
+  // no `require` statements are needed here.
```
jabraham17: unnecessary comment

src/ParquetMsg.chpl
```diff
@@ -187,569 +91,23 @@ module ParquetMsg {
...
-                 errorClass='IllegalArgumentError');
-    }
-    return filesExist;
+    return writeStringsColumn(filename, dsetName, ss.offsets.a, ss.values.a,
+                              compression: CompressionType, mode);
   }
 
+  // `dtype` is retained for backward compatibility with existing callers
+  // (e.g. CheckpointMsg and pdarray_toParquetMsg); the Mason writer infers the
+  // Arrow type from the Chapel array element type, so it is no longer needed.
```
jabraham17: overly verbose llm comment

src/ParquetMsg.chpl
```diff
@@ -1742,192 +902,27 @@ module ParquetMsg {
...
-        var endValIdx = if (lastOffset == localSegments[locDom.high]) then lastValIdx else segments[locDom.high + 1] - 1;
-              
-        var valIdxRange = startValIdx..endValIdx;
-        writeSegArrayComponent(myFilename, dsetName, olda, valIdxRange, segments, locDom, extraOffset, lastOffset, lastValIdx, c_dtype, compression);
-      }
-    }
-    return filesExist; // trigger warning if overwrite occuring
+    // Delegates to the Mason Parquet package. `c_dtype` is retained for the
+    // caller's dispatch but is unused here: writeListColumn infers the Arrow
+    // type from the value array's Chapel element type.
```
jabraham17: same nit, overly verbose llm comment

src/ParquetMsg.chpl
```diff
@@ -1742,192 +902,27 @@ module ParquetMsg {
...
-            pqErr.parquetError(getL(), getR(), getM());
-          }
-        }
-        
-      }
-    }
-    return filesExist; // trigger warning if overwrite occuring
+    // Delegates to the Mason Parquet package. For a SegArray of strings,
+    // `segments` indexes into the string offsets, `offsets` into the raw byte
+    // values, matching Arkouda's SegString layout.
```
jabraham17: llm comment doesn't add anything

src/ParquetMsg.chpl
```diff
@@ -2031,446 +1030,158 @@ module ParquetMsg {
...
+    op.distributed = true;
+    registerParquetColumns(op, colNames, symNames, colObjTypes, st);
+    op.write();
     return filesExist;
   }
 
-  proc identifyTargetLocales(name: string, objType: string, st: borrowed SymTab) throws {
-    var targetLocales;
-    select objType.toUpper(): ObjType {
+  private proc writeMultiColWithMason(filename: string,
```
jabraham17: The `WithMason` suffix sorta bugs me, is there a better name that should be used

src/ApplyMsg.chpl
eric-vo: Python interpreters kept getting reinitialized due to the `is_initialized` attr getting deleted for `ak.apply` between tests, which caused tests (`series_accessor.py`) to fail. I thought this might have been from my migration at first, but turns out it was from using `CLIENT` mode while running tests when I had a server already running (since creating a new server between tests doesn't cause this issue).

jabraham17: Ok, thats definitely something that should go in its own PR with more description about how the error occurred and the fix.

src/CheckpointMsg.chpl
eric-vo: Type checking + skipping known but unimplemented symbol types; looking again, it's unrelated, so I'll remove it + its related test

tests/numpy/pdarray_creation_test.py
eric-vo: Binary operators weren't available to `uint8`s yet, so I went ahead and did that now

jabraham17: This is definitely something that should go in a separate PR, because changing the registration config is likely to significantly increase compilation times. I would strongly encourage you to split the uint8 feature addition into its own PR

tests/numpy/pdarray_creation_test.py
eric-vo: `uint8` is now supported for `zeros()`

tests/pandas/parquet_edge_test.py
eric-vo: Yeah, I thought the current version we're using though is 24 (which causes the test to fail since it'll have correct data); should I preserve compatibility with 9?

tests/pandas/io_test.py
eric-vo: Delta encoding appears to work with the current Arrow, so it seems like expecting an error here is stale? It's somewhat unrelated to this PR though in my opinion, so should I remove this edit?

tests/pandas/parquet_edge_test.py
jabraham17: The latest version that can be used is 24, but the test suite can (and is) run with older versions. I think the main system where this test runs on a nightly basis is with arrow 9. Sadly, this test does not yet run in the CI.

I would suggest reverting this, or guarding it with a version check

tests/pandas/io_test.py
jabraham17: this is dependent on arrow version? If you are confident it will work on all supported versions of arrow, I agree the comment is stale and you can leave this in. But if this is arrow version dependent I think you should leave it alone

eric-vo: I believe it should work with Arrow 9, I think CI will show whether it does for sure though

scripts/get_parquet_package.sh
```diff
@@ -0,0 +1,82 @@
...
+  exit 1
+fi
+
+if [[ -z "${CHPL_HOME:-}" ]]; then
+  if ! command -v chpl >/dev/null 2>&1; then
+    log "ERROR: CHPL_HOME is unset and chpl is not available on PATH"
+    exit 1
+  fi
+  CHPL_HOME="$(chpl --print-chpl-home)"
+fi
```
jabraham17: this whole block should not be needed. Just use `ARKOUDA_CHPL_HOME` which is already computed

src/CheckpointMsg.chpl
```diff
@@ -386,7 +386,7 @@ module CheckpointMsg {
         return;
     }
 
-    if hasPrimitiveElements(entry) {
+    if hasPrimitiveElements(entry) && entry.etype != uint(8) {
```
jabraham17, line 389: should be removed since uint8 has been removed from this PR

src/ParquetMsg.chpl
```diff
@@ -2697,5 +1344,5 @@ module ParquetMsg {
   registerFunction("writeParquet", toparquetMsg, getM());
   registerFunction("lspq", lspqMsg, getM());
   registerFunction("getnullparquet", nullIndicesMsg, getM());
-  ServerConfig.appendToConfigStr("ARROW_VERSION", getVersionInfo());
+  ServerConfig.appendToConfigStr("ARROW_VERSION", try! getVersionInfo());
```
jabraham17: Avoid the use of `try!`, we never want the arkouda server to hard crash. I'd wrap this in a `try/catch`, log the error, and return an empty/null/invalid value

jabraham17:
I believe the CI is failing because you have not yet merged https://github.com/chapel-lang/Parquet/pull/22 correct? Lets focus on getting that merged first (that PR looks almost good to go) so we can look at the CI here

src/CheckpointMsg.chpl
eric-vo: I think we need to explicitly block `uint(8)`, otherwise it'll cause a compile error (since `hasPrimitiveElements()` includes `uint(8)`)

scripts/get_parquet_package.sh
jabraham17: You can't just rely on CHPL_HOME, it may not be set. You need to use ARKOUDA_CHPL_HOME

eric-vo: If I try to use `ARKOUDA_CHPL_HOME` without passing it to this script as an argument, I get an unbound variable error

scripts/get_parquet_package.sh
```diff
@@ -0,0 +1,74 @@
...
+  log "ERROR: expected Chapel module at ${PARQUET_MODULE}, but it is missing"
+  exit 1
+fi
+
+log "Building C++ prerequisites in ${PREREQ_DIR}"
+make -s -C "${PREREQ_DIR}" ARKOUDA_CHPL_HOME="${ARKOUDA_CHPL_HOME:-}" >&2
+
+# Gather the chpl flags the package needs.
+FLAGS="$(make -s -C "${PREREQ_DIR}" ARKOUDA_CHPL_HOME="${ARKOUDA_CHPL_HOME}" printchplflags)"
+if [[ -z "${FLAGS}" ]]; then
```
jabraham17: Should be `CHPL_HOME="$ARKOUDA_CHPL_HOME"`

jabraham17: Both `make` calls need this adjusted :)

eric-vo: Ooops 😳

jabraham17 (approved):

